PR #106, opened 2018-10-16
Clang Format All Files

dustin-lunarg:
Apply clang-format to all .h and .cpp files in the repository.


andrew-lunarg:
We are overdue doing this again on the current Clang-Format version in use on the github runners. See: https://github.com/LunarG/gfxreconstruct/pull/1096

PR #202, opened 2019-02-22
Update Android Usage Documentation

dustin-lunarg:


layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
-
-To work around this, you can use the following command to stop the
-debugger from breaking on SIGSEGVs generated by the capture layer's
-memory tracking:
-
+It may also be possible to grant the external storage permission to
+an installed application through the device Settings.
+
+**Failure to enable external storage** can cause the layer to crash if
+if it tries and fails to create a capture file on the external storage
```
lunarpapillo: "if" is repeated on the line before this and this line

layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
-running.
-
-To work around this, you can use the following command to stop the
-debugger from breaking on SIGSEGVs generated by the capture layer's
-memory tracking:
-
+It may also be possible to grant the external storage permission to
+an installed application through the device Settings.
+
+**Failure to enable external storage** can cause the layer to crash if
```
lunarpapillo: "crash" here looks bad.  Is it possible to exit more gracefully, and just say "exit" instead of "crash"?

layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
-Then choosing:
-
-```
-Debugger > LLDP Post Attach Commands
-```
-
-Where you can click the `+` and add it manually.
-
-## Globally Enabling the Layer
+It may also be set as a post launch command in the project configuration:
```
lunarpapillo: below, "post launch" is defined as "post attach".  Can we use "post attach" in both places?

layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
+ * Click on "Run" in the menu
+ * Choose "Edit Configurations..."
+ * In the dialog box, select the "Debugger" tab
+ * In the "Debugger" tab, select the "LLDB Post Attach Commands" tab
+ * Click the `+` to add the command to the command list
+ * Click "Apply"
```
lunarpapillo: I'm not terribly familiar with Android, but isn't there a missing step here?  E.g., actually entering the `process` command given above?

layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
 
-For example, to set the log_level to "warning", you would call:
+For example, to set the log_level to "warning", specify:
 
 ```
 adb shell "setprop debug.gfxrecon.log_level 'warning'"
 ```
 
+Options with the BOOL type accept the following values:
+* A case insensitive string value 'true' or a non-zero integer value indicate true.
```
lunarpapillo: "case insensitive" => "case-insensitive" (here and below)

layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
+* A case insensitive string value 'false' or a zero integer value indicate false.
+
```
lunarpapillo: What happens on all other values, e.g. empty string, or an unrecognized string?

layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
+Log Break On Error | debug.gfxrecon.log_break_on_error | BOOL | Trigger a debug break when logging an error.
+Log Detailed | debug.gfxrecon.log_detailed | BOOL | Include name and line number from the file responsible for the log message.
+Log File | debug.gfxrecon.log_file | STRING | When set, log messages will be written to a file at the specified path.
+Log File Create New | debug.gfxrecon.log_file_create_new | BOOL | Specifies that log file initialization should overwrite an existing file when true, or append to an existing file when false.
+Log File Flush After Write | debug.gfxrecon.log_file_flush_after_write | BOOL | Flush the log file to disk after each write when true.
+Log File Keep Open | debug.gfxrecon.log_file_keep_open | BOOL | Keep the log file open between log messages when true, close and reopen the log file for each message when false.
+Log Level | debug.gfxrecon.log_level | STRING | Specify the highest level message to log.  Options are: "debug", "info", "warning", "error", and "fatal".  The specified level and all levels listed after it will be enabled for logging.  For example, choosing the "warning" level will also enable the "error" and "fatal" levels.
```
lunarpapillo: What is the default value?

layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
+Log Output To Console | debug.gfxrecon.log_output_to_console | BOOL | Log messages will be written to Logcat.
```
lunarpapillo: Default is false, I presume?

layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
+Capture File Timestamp | debug.gfxrecon.capture_file_timestamp | BOOL | Add a timestamp to the capture file with the form: "gfxrecon_capture_yyyymmddThhmmss.gfxr".  Default is: true.
+Log Allow Indents | debug.gfxrecon.log_allow_indents | BOOL | Apply additional indentation formatting to log messages.
+Log Break On Error | debug.gfxrecon.log_break_on_error | BOOL | Trigger a debug break when logging an error.
+Log Detailed | debug.gfxrecon.log_detailed | BOOL | Include name and line number from the file responsible for the log message.
+Log File | debug.gfxrecon.log_file | STRING | When set, log messages will be written to a file at the specified path.
+Log File Create New | debug.gfxrecon.log_file_create_new | BOOL | Specifies that log file initialization should overwrite an existing file when true, or append to an existing file when false.
+Log File Flush After Write | debug.gfxrecon.log_file_flush_after_write | BOOL | Flush the log file to disk after each write when true.
+Log File Keep Open | debug.gfxrecon.log_file_keep_open | BOOL | Keep the log file open between log messages when true, close and reopen the log file for each message when false.
```
lunarpapillo: What is the default?

layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
+Capture File Compression Type | debug.gfxrecon.capture_compression_type | STRING | Compression format to use with the capture file.  Valid values are: "LZ4", "ZLIB", and "NONE". Default is: "LZ4".
+Capture File Timestamp | debug.gfxrecon.capture_file_timestamp | BOOL | Add a timestamp to the capture file with the form: "gfxrecon_capture_yyyymmddThhmmss.gfxr".  Default is: true.
+Log Allow Indents | debug.gfxrecon.log_allow_indents | BOOL | Apply additional indentation formatting to log messages.
+Log Break On Error | debug.gfxrecon.log_break_on_error | BOOL | Trigger a debug break when logging an error.
+Log Detailed | debug.gfxrecon.log_detailed | BOOL | Include name and line number from the file responsible for the log message.
+Log File | debug.gfxrecon.log_file | STRING | When set, log messages will be written to a file at the specified path.
+Log File Create New | debug.gfxrecon.log_file_create_new | BOOL | Specifies that log file initialization should overwrite an existing file when true, or append to an existing file when false.
+Log File Flush After Write | debug.gfxrecon.log_file_flush_after_write | BOOL | Flush the log file to disk after each write when true.
```
lunarpapillo: what is the default?

layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
+Capture File Name | debug.gfxrecon.capture_file | STRING | Path to use when creating the capture file.  Default is: "/sdcard/gfxrecon_capture.gfxr".
+Capture File Compression Type | debug.gfxrecon.capture_compression_type | STRING | Compression format to use with the capture file.  Valid values are: "LZ4", "ZLIB", and "NONE". Default is: "LZ4".
+Capture File Timestamp | debug.gfxrecon.capture_file_timestamp | BOOL | Add a timestamp to the capture file with the form: "gfxrecon_capture_yyyymmddThhmmss.gfxr".  Default is: true.
+Log Allow Indents | debug.gfxrecon.log_allow_indents | BOOL | Apply additional indentation formatting to log messages.
+Log Break On Error | debug.gfxrecon.log_break_on_error | BOOL | Trigger a debug break when logging an error.
```
lunarpapillo: The default values here and below seem "obvious"... but I'd prefer they be explicitly indicated.

layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
+Log Output To Console | debug.gfxrecon.log_output_to_console | BOOL | Log messages will be written to Logcat.
+Memory Tracking Mode | debug.gfxrecon.memory_tracking_mode | STRING | Specifies the memory tracking mode to use for detecting modifications to mapped Vulkan memory objects. Available options are: "page_guard", "assisted" and "unassisted". Default is "page_guard". <ul><li>"page_guard" tracks modifications to individual memory pages, which are written to the capture file on calls to `vkFlushMappedMemoryRanges`, `vkUnmapMemory`, and `vkQueueSubmit`. Tracking modifications requires allocating shadow memory for all mapped memory.</li><li>"assisted" expects the application to call `vkFlushMappedMemoryRanges` after memory is modified; the memory ranges specified to the `vkFlushMappedMemoryRanges` call will be written to the capture file during the call.</li><li>"unassisted" writes the full content of mapped memory to the capture file on calls to `vkUnmapMemory` and `vkQueueSubmit`. It is very inefficient and may be unusable with real world applications that map large amounts of memory.</li></ul>
```
lunarpapillo: Is there a reason HTML list construction is used instead of markdown list construction here?

layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
 
 ## Capture Files
+By default, capture files are written to `/sdcard/gfxrecon_capture.gfxr`
```
lunarpapillo: This is technically redundant with the Capture File Name description at line 97.
I don't mind it appearing in both places, but both should have the same formatting (at line 97, reverse apostrophes are not used).

If we keep this here, it really should end in a period.

layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
 
-By default, capture files are dumped out to the `/sdcard`, and will
-use the name `gfxrecon_capture.gfxr`:
+Capture files are created on the first call to vkCreateInstance, when the Vulkan
```
lunarpapillo: Elsewhere, references to Vulkan calls are formatted with back-quotes.  We should be consistent.  (I prefer back-quotes myself, but I'm more concerned about consistency.)

layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
- ```
-/sdcard/gfxrecon_capture.gfxr
-```
-
-Enabling timestamps may be needed to prevent apps that create multiple files from
-overwriting existing files.
-
-### Overriding the Default Name
+If multiple instances are created concurrently, only one capture file will be created.
+If multiple instances are created consecutively, each instance will create a new file.
```
lunarpapillo: How can this be?  There is only one setting for the file name.  If multiple instances are created, how can we create a new file for each?

layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
 
- ```
-/sdcard/gfxrecon_capture.gfxr
-```
-
-Enabling timestamps may be needed to prevent apps that create multiple files from
-overwriting existing files.
-
-### Overriding the Default Name
+If multiple instances are created concurrently, only one capture file will be created.
```
lunarpapillo: I'm not certain how to interpret "concurrently" and "consecutively" in this context.

If the application creates one Vulkan instance, and several minutes later (while the first is still open) creates a second, are these concurrent or consecutive?

layer/USAGE_android.md
```diff
@@ -1,143 +1,151 @@
...
-This may be done by performed by using `adb shell` in the following way:
+### Specifying Capture File Location
+The capture file's save location can be specified by setting the
+`debug.gfxrecon.capture_file` system property, described above in
+the [GFXRecon Debug Options](#gfxrecon-debug-options) section.
 
+The property may be set through ADB with the following command:
 ```
-adb shell "setprop debug.gfxrecon.capture_file '<your_path_and_file_name_here>.gfxr'"
+adb shell "setprop debug.gfxrecon.capture_file '<path_to_file>.gfxr'"
```
lunarpapillo: This implies the .gfxr suffix is required.  I'm not sure it is - is it?

layer/USAGE_android.md
dustin-lunarg: The only way to embed a list in a table seems to be a mix of Markdown + HTML.

layer/USAGE_android.md
dustin-lunarg: It isn't strictly required.

layer/USAGE_android.md
dustin-lunarg: Multiple instances active at the same time are concurrent.  I have made some changes to clarify.

layer/USAGE_android.md
dustin-lunarg: I am just going to document the valid values for now, and not the invalid values.  Invalid values should generate a log message, but I need to check on the current state of the settings error handling.

layer/USAGE_android.md
dustin-lunarg: The layer's influence over the application are limited, but there are some future changes that can be made to prevent the layer from crashing if initialization fails.  This is documenting the current behavior.

layer/USAGE_android.md
lunarpapillo: Ah, okay, you're right.

lunarpapillo:
Checked both of the forced pushes, and all look good to me.

PR #203, opened 2019-02-22
Utility Code Cleanup

dustin-lunarg:
* Set default log level to info
* Add null checks to platform wrapper functions that dereference pointers
* FileOpen returns 0 on success instead of errno
* Convert some memcpy calls to assignments

### DIFF
--- a/framework/util/platform.h
+++ b/framework/util/platform.h
@@ -356,6 +356,10 @@ inline int32_t FileVprintf(FILE* stream, const char* format, va_list vlist)
 
 inline int32_t LocalTime(tm* local_time, const time_t* timer)
 {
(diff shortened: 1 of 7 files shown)


framework/util/platform.h
```diff
@@ -343,30 +356,50 @@ inline int32_t FileVprintf(FILE* stream, const char* format, va_list vlist)
...
         return 0;
     }
     else
     {
         return errno;
     }
+#endif
 }
 
 inline int32_t GMTime(tm* gm_time, const time_t* timer)
```
davidlunarg: Remove this?

dustin-lunarg: This is actually used when creating the time string for the timestamped filenames, where there is an option to use either the local timezone or GMT.

PR #206, opened 2019-02-26
Fail vkCreateInstance When Layer Initialization Fails

dustin-lunarg:
When layer initialization fails, the layer will skip the call to the next vkCreateInstance function in its call chain and return VK_ERROR_INITIALIZATION_FAILED.
- Moves layer initialization from a custom pre-call command to the layer's dispatch_CreateInstance function.
- Adds a CreateInstance post-call command to check the result of the vkCreateInstance call and handle failure.


dustin-lunarg:
Force pushed a correction to the commit message.

PR #249, opened 2019-06-12
Add test script, test build directives, and test stubs

xykivo:
This change adds tests, and test build targets.

1. Add test build directives to CMake
2. Add option to skip tests - by default tests are executed as part of the build
3. Add test targets to all non android GFX reconstruct libraries - stubs only
4. Add test script that can runs specific tests, or all the tests
5. Add an option to generate a test archive as part of the build - contains all binaries and the test script


xykivo:
I'll recreate this pull request after the build script pull request is approved.
I'll manually cherry pick the test changes to a new branch creates from dev so it will be easier to rebase.

PR #252, opened 2019-06-25
Add tests to GFXReconstruct.

xykivo:
* Tests are based on Catch2
* Add test script
* Add test cmake build directives
* Add Catch2 header
* Add test binaries for every non android component

Change-Id: I4d6c520d195417e064abb90f2975418a003a5c8b

### DIFF
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -89,11 +89,8 @@ def parse_args():
         action='store_true', default=False,
         help='Apply C++ code style before compiling')
     arg_parser.add_argument(
-        '--skip-tests', dest='skip_tests',
-        action='store_true', default=False,
-        help='Skip running tests')
-    arg_parser.add_argument('--test-archive', dest='test_archive',
-                            action='store_true', default=False,
+        '--skip-tests', action='store_true', help='Skip running tests')
+    arg_parser.add_argument('--test-archive', action='store_true',
                             help='Generate a test archive package')
     arg_parser.add_argument(
         '--static-analysis', dest='static_analysis',
@@ -148,22 +145,17 @@ def cmake_generate_options(args):
         generate_options.append('-DRUN_TESTS=OFF')
         generate_options.append('-DGENERATE_TEST_ARCHIVE=OFF')
     else:
-        if args.code_style:
-            generate_options.append('-DAPPLY_CPP_CODE_STYLE=ON')
-        else:
-            generate_options.append('-DAPPLY_CPP_CODE_STYLE=OFF')
-        if args.skip_tests:
-            generate_options.append('-DRUN_TESTS=OFF')
-        else:
-            generate_options.append('-DRUN_TESTS=ON')
-        if args.test_archive:
-            generate_options.append('-DGENERATE_TEST_ARCHIVE=ON')
-        else:
-            generate_options.append('-DGENERATE_TEST_ARCHIVE=OFF')
-        if args.static_analysis:
-            generate_options.append('-DRUN_STATIC_ANALYSIS=ON')
-        else:
-            generate_options.append('-DRUN_STATIC_ANALYSIS=OFF')
+        generate_options.append(
+            '-DAPPLY_CPP_CODE_STYLE={}'.format(
+                'ON' if args.code_style else 'OFF'))
+        generate_options.append(
+            '-DRUN_TESTS={}'.format('OFF' if args.skip_tests else 'ON'))
+        generate_options.append(
+            '-DGENERATE_TEST_ARCHIVE={}'.format(
+                'ON' if args.test_archive else 'OFF'))
+        generate_options.append(
+            '-DRUN_STATIC_ANALYSIS={}'.format(
+                'ON' if args.static_analysis else 'OFF'))
     return generate_options
 
 
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -35,6 +35,7 @@ def is_windows():
 
 
 ARCHITECTURES = ['x64', 'x86']
+DEFAULT_ARCHITECTURE = ARCHITECTURES[0]
 BUILD_ROOT = os.path.abspath(
     os.path.join(os.path.split(os.path.abspath(__file__))[0], '..'))
 BUILD_CONFIGS = {'debug': 'dbuild', 'release': 'build'}
@@ -42,6 +43,7 @@ if is_windows():
     BUILD_CONFIGS['debug'] = 'build'
 CMAKE_VERSION_3_13 = distutils.version.StrictVersion('3.13.0')
 CONFIGURATIONS = ['release', 'debug']
+DEFAULT_CONFIGURATION = CONFIGURATIONS[0]
 VERSION = distutils.version.StrictVersion('0.0.0')
 
 
@@ -63,13 +65,13 @@ def parse_args():
     arg_parser.add_argument(
         '-a', '--arch', dest='architecture',
         metavar='ARCH', action='store',
-        choices=ARCHITECTURES, default=ARCHITECTURES[0],
+        choices=ARCHITECTURES, default=DEFAULT_ARCHITECTURE,
         help='Build target architecture. Can be one of: {0}'.format(
                 ', '.join(ARCHITECTURES)))
     arg_parser.add_argument(
         '-c', '--config', dest='configuration',
         metavar='CONFIG', action='store',
-        choices=CONFIGURATIONS, default=CONFIGURATIONS[0],
+        choices=CONFIGURATIONS, default=DEFAULT_CONFIGURATION,
         help='Build target configuration. Can be one of: {0}'.format(
             ', '.join(CONFIGURATIONS)))
     arg_parser.add_argument(
@@ -86,6 +88,13 @@ def parse_args():
         '--code-style', dest='code_style',
         action='store_true', default=False,
         help='Apply C++ code style before compiling')
+    arg_parser.add_argument(
+        '--skip-tests', dest='skip_tests',
+        action='store_true', default=False,
+        help='Skip running tests')
+    arg_parser.add_argument('--test-archive', dest='test_archive',
+                            action='store_true', default=False,
+                            help='Generate a test archive package')
     arg_parser.add_argument(
         '--static-analysis', dest='static_analysis',
         action='store_true', default=False,
@@ -136,11 +145,21 @@ def cmake_generate_options(args):
     if args.clean or args.clobber:
         generate_options.append('-DAPPLY_CPP_CODE_STYLE=OFF')
         generate_options.append('-DRUN_STATIC_ANALYSIS=OFF')
+        generate_options.append('-DRUN_TESTS=OFF')
+        generate_options.append('-DGENERATE_TEST_ARCHIVE=OFF')
     else:
         if args.code_style:
             generate_options.append('-DAPPLY_CPP_CODE_STYLE=ON')
         else:
             generate_options.append('-DAPPLY_CPP_CODE_STYLE=OFF')
+        if args.skip_tests:
+            generate_options.append('-DRUN_TESTS=OFF')
+        else:
+            generate_options.append('-DRUN_TESTS=ON')
+        if args.test_archive:
+            generate_options.append('-DGENERATE_TEST_ARCHIVE=ON')
+        else:
+            generate_options.append('-DGENERATE_TEST_ARCHIVE=OFF')
         if args.static_analysis:
             generate_options.append('-DRUN_STATIC_ANALYSIS=ON')
         else:
--- a/scripts/test.py
+++ b/scripts/test.py
@@ -18,6 +18,8 @@ GFX reconstruct build script
 '''
 
 import argparse
+import collections
+import copy
 import importlib
 import os
 import platform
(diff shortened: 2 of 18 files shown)


scripts/test.py
```diff
@@ -0,0 +1,146 @@
...
+}
+
+
+class TestError(Exception):
+    '''
+    Raised on a test error
+    '''
+
+
+def import_build_script():
```
lunarpapillo: I'm not a fan of unusual `import` workarounds; but there are situations where such are needed.  I don't think this is one of them...

Can you help me explain why this approach was taken over something more familiar like this in the `import` section above:
```
sys.path.insert(...)
import build
```
which would have allowed `build.XXX` to be used throughout the script, instead of having to pass `build_script` around as a parameter, simplifying the logic and making it more familiar... if there are good reasons, I'm all for it; but if there aren't, I prefer the more familiar construction.

scripts/test.py
```diff
@@ -0,0 +1,146 @@
...
+        metavar='CONFIG', action='store', choices=build_script.CONFIGURATIONS,
+        default=build_script.DEFAULT_CONFIGURATION,
+        help='Build target configuration. Can be one of: {0}'.format(
+            ', '.join(build_script.CONFIGURATIONS)))
+    return arg_parser.parse_args()
+
+
+def run_test(test_exe, test_args):
+    '''
+    Run a single test
```
lunarpapillo: Please add to the docstring here how the working directory is calculated and what will happen on test failure or exception.

scripts/test.py
```diff
@@ -0,0 +1,146 @@
...
+    '''
+    run_test_args = [test_exe]
+    if args.test_args is not None:
+        run_test_args.extend(args.test_args)
+    run_test_work_dir = os.path.split(test_exe)[0]
+    if '' == run_test_work_dir:
```
lunarpapillo: This construction has some strange effects..
- if `test_exe` is an absolute path, the working directory is the directory containing the executable
- if `test_exe` is just an executable name (presumably in the current directory), the working directory is also the directory containing the executable
- if `test_exe` is a relative path, the working directory is *not* the directory containing the executable, but is instead the current directory

Is this intended?  It seems strange.

scripts/test.py
```diff
@@ -0,0 +1,146 @@
...
+        run_test_work_dir = os.getcwd()
+    try:
+        test_result = subprocess.run(run_test_args, cwd=run_test_work_dir)
+        if 0 != test_result.returncode:
+            raise TestError('Running test', ' '.join(run_test_args), 'failed')
```
lunarpapillo: It's helpful to include the working directory and the returncode in the raised exception.

scripts/test.py
```diff
@@ -0,0 +1,146 @@
...
+    except Exception:
+        raise TestError('Failed to run test', test_exe, ' '.join(test_args))
```
lunarpapillo: As above, include the working directory and the returncode.  Also, to avoid losing information from the original exception, do something like:
```
except Exception as e:
    raise TestError('Failed to run "{}" in "{}", returncode={}'.format(
        run_test_args, run_test_work_dir, test_result.returncode
    )) from e
```

scripts/test.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+# Main entry point
+if '__main__' == __name__:
+    try:
+        build_script = import_build_script()
```
lunarpapillo: as above, avoid unusual `import` games unless there's a good reason.

scripts/test.py
```diff
@@ -0,0 +1,146 @@
...
+def is_windows():
+    '''
+    Check if the system is Windows
+    '''
+    return 'windows' == platform.system().lower()
+
+
+BUILD_ROOT = os.path.abspath(os.path.join(
+    os.path.split(os.path.abspath(__file__))[0], '..'))
+ALL_TESTS = {
```
lunarpapillo: Please document here what the value field represents.  It wasn't clear until I'd gotten all the way through the code.

Also, since `run_test()` accepts a value of `None` to indicate no test arguments, why don't each of these have a value of `None`?

scripts/test.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+# Main entry point
+if '__main__' == __name__:
+    try:
+        build_script = import_build_script()
+        args = parse_args(build_script)
+        tests = []
+        if args.test_exe is None:
+            for test in ALL_TESTS.items():
```
lunarpapillo: For deterministic behavior, the tests should be executed in a well-defined order (particularly if this is executed on different OSes).  For Python 3.6 and higher, I believe insertion order is guaranteed for `dict` variables; but for earlier version of Python 3 you'll need to either use `OrderedDict` for `ALL_TESTS` or do some type of sorting based on key here.

scripts/test.py
```diff
@@ -0,0 +1,146 @@
...
+                test_exe_dir = args.build_dir
+                if test_exe_dir is None:
+                    test_exe_dir = os.path.join(
+                        build_script.BUILD_CONFIGS[args.configuration],
+                        platform.system().lower(),
+                        args.architecture,
+                        'bin')
+                test_exe = os.path.join(test_exe_dir, test[0])
+                test_args = copy.deepcopy(test[1])
```
lunarpapillo: Why `copy.deepcopy()`?  I can't think of any reason it would be necessary.

scripts/test.py
```diff
@@ -0,0 +1,146 @@
...
+    '''
+    arg_parser = argparse.ArgumentParser(
+        description="gfxreconstruct build script",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    arg_parser.add_argument('--version', dest='version',
+                            action='version', version=str(build_script.VERSION))
+    arg_parser.add_argument(
+        '--test-exe', dest='test_exe', metavar='TEST_EXE',
+        action='store', default=None,
+        help='Test executable to run')
```
lunarpapillo: There's only a small number of valid test executables.  Add a `choices=` parameter here to ensure that, if `test_exe` is set, it's valid.

scripts/test.py
```diff
@@ -0,0 +1,146 @@
...
+                test_args = copy.deepcopy(test[1])
+                if args.test_args is not None:
+                    test_args.extend(args.test_args)
+                tests.append((test_exe, test_args))
+        else:
+            tests = [(args.test_exe, args.test_args)]
+        for test_exe, test_args in tests:
+            run_test(test_exe, test_args)
+    except Exception as error:
+        print('Error', *(error.args))
```
lunarpapillo: This loses information about the exception that caused the error, preventing a debugging engineer from seeing the stack trace.  Is there a good reason?  **Not** catching the exception at all will allow the normal Python error handling, which will include a non-zero exit code and a helpful stack trace.

scripts/test.py
```diff
@@ -0,0 +1,146 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+'''
+GFX reconstruct build script
+'''
+
+import argparse
+import copy
+import distutils.version
```
lunarpapillo: doesn't seem to be used

scripts/test.py
```diff
@@ -0,0 +1,146 @@
...
+import importlib
+import os
+import platform
+import re
```
lunarpapillo: doesn't seem to be used

scripts/test.py
```diff
@@ -0,0 +1,146 @@
...
+import shutil
```
lunarpapillo: doesn't seem to be used

scripts/build.py
```diff
@@ -86,6 +88,13 @@ def parse_args():
         '--code-style', dest='code_style',
         action='store_true', default=False,
         help='Apply C++ code style before compiling')
+    arg_parser.add_argument(
+        '--skip-tests', dest='skip_tests',
+        action='store_true', default=False,
```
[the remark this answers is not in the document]
xykivo: I agree that it's not needed, but it makes it clear what's going on, which is good for script maintenance.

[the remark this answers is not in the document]
xykivo: Unless you feel strongly about this, I would prefer to leave it as is.

scripts/build.py
```diff
@@ -86,6 +88,13 @@ def parse_args():
...
+        help='Skip running tests')
+    arg_parser.add_argument('--test-archive', dest='test_archive',
+                            action='store_true', default=False,
```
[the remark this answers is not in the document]
xykivo: Same as above.

scripts/build.py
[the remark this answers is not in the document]
lunarpapillo: Just for the record... what you find "makes it clear what's going on", I find "misleading because it makes me feel something is deliberately different from the default" and "misleading because it's an unusual usage of parameters".  I don't memorize all the arguments to every library function, so when I see unexpected ones being used, I have to go look them up.  So for me this is more effort spent every time someone reads the code, just to discover that the parameters are all equal to defaults.

But admittedly this doesn't affect code function and, all things considered, this is relatively minor.  Your preference is acceptable.

As an alternative suggestion (which you may also ignore :-), a small comment above all the `add_argument()` calls stating your intention, e.g.:
```
# Many of the arguments here have default values; there are provided to
# make the action explicitly clear.
```
would make it clear that this was deliberately done and why, and would lessen the confusion of code readers like me.  It makes it clear you know these are defaults (so I don't have to point that out in case it was an oversight) and doesn't make me wonder why the defaults are being provided.  (I'd still have to look up the library function to make sure the arguments are being used correctly.)

But if you don't want to provide such a comment, that's acceptable too.

lunarpapillo:
One more question that wasn't clear to me... the code test files are still empty, right?  So no tests are currently defined; this just provides the structure to easily add tests.

Your intent is to add unit-level tests, yes?  Testing functions at a low level to ensure they behave the way they're expected, yes?

Are you also considering using the `catch2` structure to provide user-level tests (testing normal use cases at the command line) to ensure that they also behave correctly?  Or would you expect a separate test structure for this type of testing?

scripts/test.py
xykivo: This provides a nice output for QA and QE - the plan is for AMD QA/QE run the same test script.
I'll update this to include the exception name, but the feedback we got from QA is that the stack trace is of no interest to them.

xykivo:
This code change just provides the framework for testing.
Initially we plan on adding tests for new code and any bug fixes we make. We'll try to add tests to existing code if we have the bandwidth. 

The plan is to use Catch2 to write
1. Unit tests - short term
2. Integration/system tests - short term

For user level testing we'll probably need to split the tests between Catch2 and scripts or another framework more suited for user level testing.
We want to add user level testing, but this is currently long term.

What we would like to see is that the tests are run for every pull request.
Internally we run tests for every code review, so if we merge changes from LunarG master we won't get test failures in internal reviews.

scripts/test.py
xykivo: There are a few of problems with adding choices
1. The test exe includes the local system path, which cannot easily be added to the choice
2. It limits the user from adding tests which are not built as part of GFXReconstruct - we have such tests intenrally
3. It makes this argument required. The current behavior is that if this argument is omitted, all tests are executed - this can fixed by adding an 'all' test exe, but I find this to be ugly.

scripts/test.py
lunarpapillo: QA includes not only driver QA, but test harness QA.  I understand why a test harness problem is of no interest to driver QA.  But it is of interest to whoever is maintaining the harness.  Without the normal stack trace, some classes of problems (particularly problems that seem to appear randomly) are difficult to debug.

scripts/test.py
lunarpapillo: Ah, all right.  There's confusion here because the code implies that all the tests are listed in `ALL_TESTS`.  Please add script comments and help text to make it clear that `--test-exe` is not required to be a member of `ALL_TESTS`.  I think it would be helpful to rename `ALL_TESTS` to `ALL_INTERNAL_TESTS` to help make it clear that external tests are also supported.

I'd like the help text to be much more complete in this case, e.g.:
```
help='the name of a single internal test to run (one of: {}), or a path '
     'to an external test to run.  If not specified, all internal tests are run.'
     .format(', '.join(ALL_INTERNAL_TESTS))

scripts/test.py
xykivo: How would prefer this to be handled?
1. Completely remove the try/except?
2. Add a printout of the callstack in except?

scripts/test.py
xykivo: Unused imports were removed.

scripts/test.py
xykivo: Unused imports were removed.

scripts/test.py
xykivo: Unused imports were removed.

xykivo:
I'll try to upload a new update early next week.
What unaddressed questions do you have?

scripts/test.py
xykivo: The sys.path.insert changes import paths for any script called from this one, or any other python script called in a subprocess.
I'm not a fan of changing the environment for the whole process tree if not needed.
This approach allows specific imports in the repo file tree + it is cleaner, since all import statements are grouped together at the top of the file.

scripts/test.py
xykivo: Same reply as above.
This keeps the system path clean for any sub process, or scripts launched from this one.

scripts/test.py
xykivo: These are the tests to run when no test exe is passed to the script, why do we need to add None here?

scripts/test.py
xykivo: I don't want to modify the original test arguments, since they might be used later.

xykivo:
Apologies for the late response - was busy with other projects.
I hope I addressed/answered all concerns.

Thanks

xykivo:
I ran pylint each time i save the script file - part of VSCode python plugin.
However I agree we should have a pylint check as part of the pull request tests for all Python scripts.

Thanks

scripts/test.py
lunarpapillo: I don't entirely agree (IMO there is only negligible potential for impacts to subprocesses or other imported scripts, while confusion over the unusual import has already impacted at least one reviewer), but your motives are defensible.  In general, I'd like to see doc strings and/or comments explaining unusual constructions like this, so the reader doesn't have to guess at why they're being done this way.

dustin-lunarg:
Closed with #270

PR #258, opened 2019-07-22
Code style: add code style checking before compiling

xooiamd:
- Use clang-format-diff.py which is distributed part of
LLVM project to perform code style checking
- Add check_code_style.py script to be called from CodeStyle.cmake
in order to perform the check
- Code style check can be turned off with option
--skip-check-code-style, it is on by default

SWLGFXREC-7

Change-Id: I374ea940bb81151d6fb3b6451d09524dd521e264

### DIFF
--- a/cmake/CodeStyle.cmake
+++ b/cmake/CodeStyle.cmake
@@ -19,7 +19,7 @@
 
 cmake_minimum_required(VERSION 3.1)
 
-option(APPLY_CPP_CODE_STYLE "Apply C++ code style using clang format" ON)
+option(APPLY_CPP_CODE_STYLE "Apply C++ code style using clang format" OFF)
 option(CHECK_CPP_CODE_STYLE "Check C++ code style using clang format" ON)
 
 find_program(CLANG_FORMAT clang-format DOC "Clang format executable")
--- a/cmake/CodeStyle.cmake
+++ b/cmake/CodeStyle.cmake
@@ -19,9 +19,27 @@
 
 cmake_minimum_required(VERSION 3.1)
 
-option(APPLY_CPP_CODE_STYLE "Apply C++ code style using clang format" OFF)
+option(APPLY_CPP_CODE_STYLE "Apply C++ code style using clang format" ON)
+option(CHECK_CPP_CODE_STYLE "Check C++ code style using clang format" ON)
 
 find_program(CLANG_FORMAT clang-format DOC "Clang format executable")
+
+# Python
+if(CMAKE_HOST_WIN32)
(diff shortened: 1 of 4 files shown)


cmake/CodeStyle.cmake
```diff
@@ -19,23 +19,52 @@
 
 cmake_minimum_required(VERSION 3.1)
 
-option(APPLY_CPP_CODE_STYLE "Apply C++ code style using clang format" OFF)
+option(APPLY_CPP_CODE_STYLE "Apply C++ code style using clang format" ON)
+option(CHECK_CPP_CODE_STYLE "Check C++ code style using clang format" ON)
```
lunarpapillo: Does this make sense to default to `ON`?  If I'm reading the below correctly, if `APPLY_CPP_CODE_STYLE` is `ON`, `CHECK_CPP_CODE_STYLE` will be forced to `OFF`.

xooiamd: hi, it should be off, I have changed the default to off. thanks

dustin-lunarg: Can the code style check be disabled by default in the CMake configuration files, and then enabled by default by the build script?  The intent would be for the default CMake output to perform a  minimum number of build steps, with a minimum number of tool dependencies, for the end user that just wants to check out and build the code.  The build script would then be used by developers contributing to the project to verify changes before they are submitted.

cmake/CodeStyle.cmake
```diff
@@ -20,22 +20,51 @@
 cmake_minimum_required(VERSION 3.1)
 
 option(APPLY_CPP_CODE_STYLE "Apply C++ code style using clang format" OFF)
+option(CHECK_CPP_CODE_STYLE "Check C++ code style using clang format" ON)
 
 find_program(CLANG_FORMAT clang-format DOC "Clang format executable")
 
+# Python
```
dustin-lunarg: Can the Python check be conditional on the code style check being enabled?

cmake/CodeStyle.cmake
xooiamd: sure. it is already enabled by default in the build script, and I just made a commit 
[cedbfcb](https://github.com/LunarG/gfxreconstruct/pull/258/commits/cedbfcba02d9e340e0d7c1a8e4d128497ab353cc) to disable it by default in the CMake.

cmake/CodeStyle.cmake
xooiamd: yes, the check can be conditional, it is updated in the commit [cedbfcb](https://github.com/LunarG/gfxreconstruct/pull/258/commits/cedbfcba02d9e340e0d7c1a8e4d128497ab353cc)

dustin-lunarg:
Closed with #270

PR #290, opened 2019-09-24
Mapped Memory Tracking Updates

### LINKED ISSUE #205: Add Options for Page Guard
Add support for obtaining page guard options from a configuration file or environment variable/system property.

Settings should be applied in TraceManager::Initialize, where the page guard manager is currently initialized with hard coded constants as `Create(true, false, true, true, true, true)`

dustin-lunarg:
Updates for the "page guard" memory tracking mode:
* Remove unused experimental and debug options
* Add an experimental option using VK_EXT_external_memory_host to eliminate the need for shadow memory to detect write access to mapped memory.  The current implementation only works reliably on Windows with AMD GPUs.
* Add settings file/environment variable options for controlling page guard behavior:
  - GFXRECON_PAGE_GUARD_COPY_ON_MAP to enable/disable copy from mapped
    memory to shadow memory after map.
  - GFXRECON_PAGE_GUARD_LAZY_COPY to change the copy on map behavior
    such that the copy is performed for individual pages on first access.
  - GFXRECON_PAGE_GUARD_EXTERNAL_MEMORY to enable the experimental
    feature that eliminates the need for shadow allocations.

Closes #205

### DIFF
--- a/framework/encode/capture_settings.cpp
+++ b/framework/encode/capture_settings.cpp
@@ -67,6 +67,12 @@ GFXRECON_BEGIN_NAMESPACE(encode)
 #define MEMORY_TRACKING_MODE_UPPER          "MEMORY_TRACKING_MODE"
 #define CAPTURE_FRAMES_LOWER                "capture_frames"
 #define CAPTURE_FRAMES_UPPER                "CAPTURE_FRAMES"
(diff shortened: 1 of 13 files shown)


illia-a-polishchuk:
Seems it's broken now
@dustin-lunarg 
https://github.com/LunarG/gfxreconstruct/issues/802

PR #311, opened 2019-11-20
Tell clang-format to ignore generated directory

mikes-lunarg:
This should allow running `git clang-format` without it modifying the generated files.  .clang-format file copied from KhronosGroup/Vulkan-ValidationLayers which has a similar need.


mikes-lunarg:
Tested against another codegen change and works as intended, removed WIP

PR #315, opened 2019-12-09
Warn for VK_INCOMPLETE on replay

dustin-lunarg:
- For replay, change log level for VK_INCOMPLETE return values from Debug to Warning.
- For capture, add assertion checks for index values used with the vector of swapchain image info structs maintained by the trimming state tracker. When vkGetSwapchainImagesKHR returns VK_INCOMPLETE, the index can be outside of the expected range. Primarily intended to detect the case where trimming a replay will fail when the image count read from the capture file is smaller than the image count generated for replay.

### DIFF
--- a/framework/encode/vulkan_state_tracker.cpp
+++ b/framework/encode/vulkan_state_tracker.cpp
@@ -932,11 +932,11 @@ void VulkanStateTracker::TrackSemaphoreSignalState(uint32_t           wait_count
 void VulkanStateTracker::TrackAcquireImage(
     uint32_t image_index, VkSwapchainKHR swapchain, VkSemaphore semaphore, VkFence fence, uint32_t deviceMask)
 {
-    assert(swapchain != VK_NULL_HANDLE);
-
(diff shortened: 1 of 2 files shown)


framework/encode/vulkan_state_tracker.cpp
```diff
@@ -932,12 +932,12 @@ void VulkanStateTracker::TrackSemaphoreSignalState(uint32_t           wait_count
 void VulkanStateTracker::TrackAcquireImage(
     uint32_t image_index, VkSwapchainKHR swapchain, VkSemaphore semaphore, VkFence fence, uint32_t deviceMask)
 {
-    assert(swapchain != VK_NULL_HANDLE);
```
davidlunarg: Are there cases now where swapchain might VK_NULL_HANDLE?

dustin-lunarg: The swapchain handle is still expected to be a non-null value here.  The assert wasn't completely removed, it was moved below the type cast and combined with a second check.

Issue #369, opened 2020-05-19
Build directory overlap with build.py


lunarpapillo:
When building with `scripts/build.py`, there are eight flavors of desktop build, all combinations of (`32-bit`|`64-bit`) * (`Release`|`Debug`) * (`cmake`|`script`).  Ideally, they should all build to different directories, so that a developer (or CI) can build and test all versions out of the same source tree without conflict (allowing one CI process to build all at once, and another to test all at once).

Right now there are two issues.  The first is that the 64-bit release cmake build directory `build` conflicts with the 32/64-bit release script build directory `build`.

The second is that on Windows, the script release and debug builds also share that same directory `build`.

The problem for developers is that they will overwrite a build of another flavor, requiring rebuilds that would otherwise not be necessary; and that a developer who doesn't realize that they've overwritten a previous build may not be testing the build they think they are.

The problem for CI is that this blocks the ability to build all flavors in a dedicated build machine and then test as a follow-on step.  (CI doesn't do this currently, but for other reasons we may be moving toward this in the future.)

lunarpapillo:
If and when this is fixed, the CI test scripts for gfxreconstruct will have to be adjusted.

xykivo:
I will take a look at this.

lunarpapillo:
Thanks...

dsmolars:
I think I was able to repro this on my system.
Just to so we're in sync - you want separate output for the cmake build files, and the build output files for each configuration.
This means the build directory structure will be

Release builds:
build/
  <os_name>/
    x64|x86/
      cmake_output/ <- will contains all CMake generated build files
      output/  <- will contains all build output files (exe, dll, .so ...)

Debug build:
dbuild/
  <os_name>/
    x64|x86/
      cmake_output/ <- will contains all CMake generated build files
      output/  <- will contains all build output files (exe, dll, .so ...)

Is this OK?

Note that internally at AMD the debug/release dir resides under the os_dir - e.g. build/linux/debug/x64/...

Thanks

lunarpapillo:
Your proposed structure (build|dbuild)/<os_name>(x64|x86)/(cmake_output|output) does solve the problem, and I'm fine with that.

[...]

PR #417, opened 2020-08-10
Changes to the build output directories

dsmolars:
* Each configuration + platform + architecture configuration now has
  it's own output directory
* Each configuration + platform + architecture configuration now has
  it's own cmake files output directory

Fixes: #369

Change-Id: Id6ee2a227c4cdfb1ede55cd3ef2a129b0ea1b0b1

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,7 @@ endif()
 configure_file("${CMAKE_SOURCE_DIR}/project_version.h.in" "${CMAKE_BINARY_DIR}/project_version.h")
 
 # Code checks
+include("Utils")
 include("CodeStyle")
 include("Cpp")
 include("Lint")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,12 +85,14 @@ configure_file("${CMAKE_SOURCE_DIR}/project_version.h.in" "${CMAKE_BINARY_DIR}/p
 
 # Code checks
 include("CodeStyle")
-include("Test")
+include("Cpp")
 include("Lint")
+include("Test")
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -85,12 +85,17 @@ configure_file("${CMAKE_SOURCE_DIR}/project_version.h.in" "${CMAKE_BINARY_DIR}/p
...
-include("Test")
+include("Cpp")
 include("Lint")
+include("Test")
 include("FindVulkanVersion")
 
 # Apply misc build directives to the given target
 macro(common_build_directives TARGET)
+    if(MSVC)
+        target_compile_options(${TARGET} PRIVATE $<$<CONFIG:Debug>:/MTd> $<$<CONFIG:Release>:/MT>)
```
dustin-lunarg, line 96: This appears to be generating linker warnings with Release builds and errors with Debug builds.  If you remove the target_compile_options addition, I think this should be good to merge.

dsmolars:
I'll try to repro this, but I don't remember seeing this on the amd-builds.

CMakeLists.txt
dsmolars: I'll try to repro this, but I don't remember seeing this on the amd-builds.

dsmolars: OK, found the issue.
For some reason I can only repro on my test system - I suspect it's related to what is installed on the system.
The issue is that LZ4 static lib is linked against the dynamic runtime (/MD).

dsmolars: Removed the /MT and /MTd options when compiling with MSVC

PR #440, opened 2020-10-15
Add a replay option to disable unsupported extensions and device features

dustin-lunarg:
Adds a new option named --remove-unsupported to adjust instance and device creation parameters to remove unsupported extensions and features.  This can cause replay to fail if the extensions/features are required, but can in some cases make replay work on different devices/drivers for applications that enable more features than they use.

### DIFF
--- a/USAGE_android.md
+++ b/USAGE_android.md
@@ -287,7 +287,8 @@ usage: gfxrecon.py replay [-h] [-p LOCAL_FILE] [--version] [--pause-frame N]
                           [--paused] [--screenshot-all] [--screenshots RANGES]
                           [--screenshot-format FORMAT] [--screenshot-dir DIR]
                           [--screenshot-prefix PREFIX] [--sfa] [--opcd]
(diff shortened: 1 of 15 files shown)


per-mathisen-arm:
Just curious, how does it know which extensions and features are actually unused?

dustin-lunarg:
This change is disabling the extensions and features that are not supported by the replay device without attempting to determine if they are used.  So basically a trial and error or use at your own risk scenario.  Warnings will be printed with the names of the extensions and features that were disabled.

I've been able to use this option to debug a replay crash with a file that was captured on a different Android device, where the replay was not fully correct, but it worked enough to debug and fix the issue.

per-mathisen-arm:
I have been thinking that a tracer could record the usage of extensions and features while tracing, and record this into the trace as some extra metadata. And then the replayer could use this metadata to figure out which extensions/features to actually enable. I don't know how feasible this would be for gfxreconstruct?

PR #459, opened 2020-11-03
cmake: Add BUILD_WERROR.

orbea:
The user's compiler may not be the same as the developer's compiler and its useful to allow disabling `-Werror` in such cases. By default it will be enabled to avoid changing the current behavior, but can be disabled by passing `-DBUILD_WERROR=OFF` as a cmake argument.


dustin-lunarg (approved):
Looks good, thanks

PR #496, opened 2021-01-21
Build fix for macOS

kvark:
Sub-tasks:
- [x] build fixes
- [x] CI checks
- [ ] WSI hookup

I don't want to invest into writing a test application, WSI hookup, README update, and other related things, at this point.
Would you be open to merging the PR without these?

### DIFF
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -104,6 +104,54 @@ jobs:
         name: ${{ matrix.config.artifact }}
         path: .\gfxreconstruct-dev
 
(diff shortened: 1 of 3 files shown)


dustin-lunarg:
Sorry, I received an email notification when this PR was created, but not when the subject was edited to add the question, so just seeing it now.

Yes, we are open to accepting a PR that enables the macOS build, without support for the items you listed, as an initial step toward macOS support.

bradgrantham-lunarg:
Changes pulled into #841

PR #501, opened 2021-02-05
Use host coherent as fallback when creating a staging buffer

infapi00:
By spec only COHERENT is mandatory, so it is possible finding a driver
not supporting CACHED. The current caller for CreateStagingBuffer
already checks if the buffer is coherent or not.

====

As a specific example, right now the v3dv driver (Mesa vulkan driver for V3D Broadcom GPUs, developed specifically for the rpi4) doesn't support CACHED. This change is needed when capturing with a initial frame different to 1 (that is useful when you want to skip intros, menu UI, etc).

The contribution guide recommends to test this change with DOTA, but that seems too much for the embedded device. I have tested it with Quake3e (a vulkan port of quake3 [1]), with vkQuake2 [2], and Unreal 4 Shooter Game demo [3]

[1] https://github.com/ec-/Quake3e
[2] https://github.com/kondrak/vkQuake2/
[3] https://docs.unrealengine.com/en-US/Resources/SampleGames/ShooterGame/index.html

### DIFF
--- a/framework/encode/vulkan_state_writer.cpp
+++ b/framework/encode/vulkan_state_writer.cpp
@@ -2870,6 +2870,16 @@ VkResult VulkanStateWriter::CreateStagingBuffer(const DeviceWrapper*    device_w
                                          &memory_type_index,
                                          memory_property_flags,
                                          state_table);
+        if (!found)
+        {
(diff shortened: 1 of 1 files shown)


dustin-lunarg:
Looks good, thanks!

PR #514, opened 2021-03-04
Include state load time in post-replay timing info

davidd-lunarg:
Add console output for trimmed file state load time. Also, when computing
FPS, account for trimmed state load time (if any) and report as "Replay
FPS".

Example output with trimmed capture file:

```
Load time:  1.435010 seconds
Total time: 7.576557 seconds
Replay FPS: 285.107319 fps, 6.141547 seconds, 1751 frames, framerange 250-2000
```

Example output with complete (not trimmed) capture file:

```
Total time: 6.758087 seconds
Replay FPS: 254.361922 fps, 6.758087 seconds, 1719 frames, framerange 1-1719
```

### DIFF
--- a/android/framework/graphics/CMakeLists.txt
+++ b/android/framework/graphics/CMakeLists.txt
@@ -2,6 +2,8 @@ add_library(gfxrecon_graphics STATIC "")
 
 target_sources(gfxrecon_graphics
                PRIVATE
+                    ${GFXRECON_SOURCE_DIR}/framework/graphics/fps_info.h
+                    ${GFXRECON_SOURCE_DIR}/framework/graphics/fps_info.cpp
                     ${GFXRECON_SOURCE_DIR}/framework/graphics/vulkan_device_util.h
                     ${GFXRECON_SOURCE_DIR}/framework/graphics/vulkan_device_util.cpp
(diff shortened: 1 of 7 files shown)


lunarpapillo:
CI looks for output lines with `fps` in them for special processing.  Functional **gfxrecon** is unaffected, as the line is logged as-is.  Performance **vvltraces** is a little fragile, relying on the exact syntax of the line.

The test suite can easily be updated to work in a way that can accept either the original or the new format.  I'll try to get that done immediately so this change can be merged without impacting current tests.

Tagging @Tony-LunarG for his awareness.

lunarpapillo:
@davidd-lunarg question: the original output also referred to a `loop`, e.g.:
```
31.570753 fps, 198.284784 seconds, 6260 frames, 1 loop, framerange 1-6260
```
Is this variable now gone from the output, or shows up conditionally?

davidd-lunarg:
The loop variable output is gone now. Previously it was always 1.

lunarpapillo:
Associated test change: https://github.com/LunarG/VulkanTests/pull/96

lunarpapillo:
Test change has merged.  You should be able to merge this PR without affecting testing.  (In theory, at least. ;-)

PR #516, opened 2021-03-05
VK_EXT_headless_surface support

tomped01:
This adds support for the VK_EXT_headless_surface extension to gfxrecon-replay, allowing us to replay on platforms without a windowing system.

This is very useful for testing and performance tracking on development platforms!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,14 @@ set(GFXRECONSTRUCT_PROJECT_VERSION_PATCH 7)
 set(GFXRECON_PROJECT_VERSION_DESIGNATION "-unknown")
 set(GFXRECON_PROJECT_VERSION_SHA1 "unknown-build-source")
 
+if (NOT DEFINED NO_XCB)
+    set(NO_XCB FALSE)
+endif()
+
+if (NOT DEFINED HEADLESS)
+    set(HEADLESS TRUE)
+endif()
+
 include(GetGitRevisionDescription)
 get_git_head_revision(GIT_REFSPEC GIT_SHA1)
 git_get_exact_tag(GIT_TAG)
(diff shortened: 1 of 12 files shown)


CMakeLists.txt
```diff
@@ -138,7 +150,8 @@ target_compile_definitions(linux_specific INTERFACE _FILE_OFFSET_BITS=64 PAGE_GU
     $<$<BOOL:${X11_FOUND}>:VK_USE_PLATFORM_XLIB_KHR>
     $<$<BOOL:${X11_Xrandr_FOUND}>:VK_USE_PLATFORM_XLIB_XRANDR_EXT>
     $<$<BOOL:${XCB_FOUND}>:VK_USE_PLATFORM_XCB_KHR>
-    $<$<BOOL:${WAYLAND_FOUND}>:VK_USE_PLATFORM_WAYLAND_KHR>)
+    $<$<BOOL:${WAYLAND_FOUND}>:VK_USE_PLATFORM_WAYLAND_KHR>
+    $<$<BOOL:${HEADLESS}>:VK_USE_PLATFORM_HEADLESS>)
```
dustin-lunarg, line 154: Does this need to be restricted to Linux, or can it also be enabled for Windows?  Currently, HEADLESS is set to true for both Windows and Linux, so the headless files are included in the Windows build, but the `--wsi headless` option is unavailable because VK_USE_PLATFORM_HEADLESS isn't defined for Windows.

framework/application/headless_window.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+
+void HeadlessWindow::SetVisibility(bool show)
+{
+    GFXRECON_UNREFERENCED_PARAMETER(show);
+}
+
+void HeadlessWindow::SetForeground() {}
+
+bool HeadlessWindow::GetNativeHandle(HandleType type, void** handle)
```
dustin-lunarg, line 86: This function is missing the GFXRECON_UNREFERENCED_PARAMETER macros, but otherwise everything looks good.

tomped01: My bad, fixed now!

CMakeLists.txt
tomped01: > Does this need to be restricted to Linux
No, updated now, hopefully the new solution is ok (let me know if you want me to change it).

dustin-lunarg:
Looks good, thanks!

PR #517, opened 2021-03-05
Add arm and arm64 build targets

tomped01:
This enables cross compiling for arm architectures using cmake.
This is useful for driver development and application testing on arm linux platforms!


dustin-lunarg:
Looks good, thanks!  Change was rebased with latest dev branch to resolve a conflict and pushed directly to dev.

PR #522, opened 2021-03-22
Use host coherent as fallback when creating a screenshot copy resource

infapi00:
By spec only COHERENT is mandatory, so it is possible finding a driver
not supporting CACHED. The current caller for CreateCopyResource
already checks if the buffer is coherent or not.

This is equivalent to #501 but for screenshot handler.


dustin-lunarg:
Looks good, thanks!

PR #524, opened 2021-03-31
Add annotations to capture file format

dustin-lunarg:
- Adds a new annotation block type to the capture file format.  The annotation consists of a string with a name/label for the annotation and a string with the annotation data.  The data can be represented as plain text, XML, and JSON.
- Adds a new AnnotationHandler interface.  Custom handlers can be created for processing annotation data.
- Adds support to the FileProcessor for processing annotation blocks that are read from the capture file, passing the annotation name and data to an AnnotationHandler.


per-mathisen-arm:
This looks like a really good start!

PR #532, opened 2021-04-27
Vulkan layer JSON fixes

mikes-lunarg:
2 fixes for the layer json file
- Add `@only` to configure_file so it doesn't do substitution on the
vkconfig `${LOCAL}` variable in VkLayer_gfxreconstruct.json.in
- Update the CMake script that extracts version numbers from vulkan_core.h
to work with the new macros in version 176 of the header.


christophe-lunarg (approved):
Looks good to me, I made the same change in LunarG/VulkanTools

mikes-lunarg:
Cherry-picked to sdk-1.2.176

christophe-lunarg:
Thanks!

PR #538, opened 2021-05-21
manifest: some setting are only available on Windows

christophe-lunarg:
Some gfxreconstruct settings are only available on Windows. The layer manifest schema support this as a result the Windows only settings won't be visible on other OS.


dustin-lunarg:
Looks good, thanks!

PR #542, opened 2021-06-04
Limit the number of vkGetQueryPoolResults retries

werman:
Follows the recent change to `vkGetEventStatus` - https://github.com/LunarG/gfxreconstruct/pull/523

When replaying `vkGetQueryPoolResults`, limit the number of retries
performed when attempting to get the replay result to match the capture
result, to avoid a potential infinite loop.

This happened in a number of D3D11 games running with DXVK.

I'm a bit hesitant to do the same for `vkGetFenceStatus` since I don't have 
any real-world app to suffer from it. So I just left a comment.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -45,6 +45,7 @@ GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
 const size_t kMaxEventStatusRetries = 16;
+const size_t kMaxQueryPoolResultsRetries = 16;
 
 const int32_t  kDefaultWindowPositionX = 0;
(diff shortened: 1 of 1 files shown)


bradgrantham-lunarg (approved):

bradgrantham-lunarg:
Hi, @werman !  If the application had used the query results to alter rendering parameters, then it wouldn't matter whether the query succeeded or exceeded the retry count on replay because no one is reading the query results during replay.  Is it possible in these traces that  the query is always exceeding the retry count?  Or did you replay some captures where it retried some smaller number of times?  Thanks!

werman:
Hi @bradgrantham-lunarg, 

> no one is reading the query results during replay.

Yes, in the original d3d11 trace there is no usage of predicates, so all query results are read on cpu and therefor are baked into the trace.

> Is it possible in these traces that the query is always exceeding the retry count?

Some are not exceeding the count, some are. It depend on how fast the GPU is.

PR #547, opened 2021-06-21
Update header to 1.2.182

locke-lunarg:
1. Skip `VK_NVX_binary_import` and `VK_HUAWEI_subpass_shading` extensions for now. They are not ready in vkspec.

2. Add custom code for `union VkAccelerationStructureMotionInstanceDataNV`.
     `VkAccelerationStructureSRTMotionInstanceNV srtMotionInstance` data size is the largest, so encode and decode it.


bradgrantham-lunarg (approved):

mikes-lunarg (changes requested):
Looks good, just one concern about file format portability and `VkAccelerationStructureMotionInstanceDataNV`

framework/encode/custom_vulkan_struct_encoders.cpp
```diff
@@ -90,6 +90,11 @@ void EncodeStruct(ParameterEncoder* encoder, const VkDeviceOrHostAddressConstKHR
     encoder->EncodeVkDeviceAddressValue(value.deviceAddress);
 }
 
+void EncodeStruct(ParameterEncoder* encoder, const VkAccelerationStructureMotionInstanceDataNV& value)
+{
+    EncodeStruct(encoder, value.srtMotionInstance);
```
mikes-lunarg, line 95: I don't think this is portable. It relies on unpacking bit fields at encode time and repacking them at decode time (with the same exact alignment) to reconstruct the other union members.

What we do in similar cases is create a custom encode/decode handler for the *parent* structure (`VkAccelerationStructureMotionInstanceNV` in this case), and then explicitly encode/decode the union members based on the `type` value. Look at `Decoded_VkPerformanceValueINTEL` and `Decoded_VkAccelerationStructureGeometryKHR` for examples.

framework/decode/custom_vulkan_struct_decoders.cpp
```diff
@@ -110,6 +110,45 @@ size_t DecodeStruct(const uint8_t* buffer, size_t buffer_size, Decoded_VkAcceler
...
+
+    bytes_read += ValueDecoder::DecodeEnumValue((buffer + bytes_read), (buffer_size - bytes_read), &(value->type));
+    bytes_read += ValueDecoder::DecodeFlagsValue((buffer + bytes_read), (buffer_size - bytes_read), &(value->flags));
+
+    switch (value->type)
+    {
+        case VK_ACCELERATION_STRUCTURE_MOTION_INSTANCE_TYPE_STATIC_NV:
+            wrapper->staticInstance = DecodeAllocator::Allocate<Decoded_VkAccelerationStructureInstanceKHR>();
+            wrapper->staticInstance->decoded_value =
+                reinterpret_cast<VkAccelerationStructureInstanceKHR*>(&value->data);
```
mikes-lunarg, line 128: You can skip the cast by reaching further into `data`, ie:
```
            wrapper->staticInstance->decoded_value = &value->data.staticInstance;
```
Totally up to you, it shouldn't change the behavior

framework/encode/custom_vulkan_struct_encoders.cpp
locke-lunarg: Thanks. Fixed it.

mikes-lunarg (approved):
Updates look great, thanks!

framework/decode/custom_vulkan_struct_decoders.cpp
locke-lunarg: Thanks. You're right.

PR #550, opened 2021-07-13
Add toolchain to make a static build for HPC clusters

per-mathisen-arm:
Note that this will not work for the tracer layer, as it would require building
it and all used libraries with -fPIC, so skipping the layer from this type of
build. Also make sure it compiles with older gcc versions.

Change-Id: I421e23af507a367033d9cafd93e022af05b09676


CMakeLists.txt
```diff
@@ -194,7 +203,9 @@ if (${RUN_TESTS})
 endif()
 
 add_subdirectory(framework)
-add_subdirectory(layer)
+if(!BUILD_STATIC)
```
lunarpapillo, line 206: This seems incorrect.  The `layer` directory was built before if `BUILD_STATIC` was off.  Now it's not?

lunarpapillo (changes requested):
This is failing LunarG CI because the `layer/` directory is not getting built any more.

per-mathisen-arm:
As far as I can tell, it builds fine now. Anything else that needs to be done to merge this?

lunarpapillo:
I'll add the curator as a reviewer and ping the team.

CMakeLists.txt
```diff
@@ -148,7 +150,16 @@ if(UNIX)
 endif(UNIX)
 
 # CMake provided find modules
-find_package(ZLIB)
+if(BUILD_STATIC)
+    message(STATUS "Making a static build")
```
bradgrantham-lunarg, line 154: These add verbosity to a CMakeLists.txt that otherwise doesn't have any message() directives.  I'd like these and the other three message() directives omitted, please.

cmake/toolchain/linux_x86_64_hpc.cmake
```diff
@@ -0,0 +1,7 @@
+set(BUILD_SHARED_LIBS OFF)
```
bradgrantham-lunarg, line 1: Can this be called "linux_x86_64_static.cmake"?  This PR doesn't look like it has anything in it that's specific to HPC, rather just adding static support.

framework/graphics/fps_info.cpp
```diff
@@ -26,6 +26,7 @@
 #include "util/date_time.h"
 #include "util/logging.h"
 
+#define __STDC_FORMAT_MACROS
```
bradgrantham-lunarg, line 29: We use the format macros everywhere with no errors without this #define.  Is this specific to a compiler or system headers version?  It appears from this one web page https://www.gnu.org/software/gnulib/manual/html_node/inttypes_002eh.html that you'd only need this on C++ compilers before C++11, and we specify C++14 support in BUILD.md.
Can you tell me more about why you had to add this?

bradgrantham-lunarg (changes requested):

framework/graphics/fps_info.cpp
per-mathisen-arm: This is actually the only place in the gfxreconstruct code where inttypes.h is included. Elsewhere #include <cinttypes> is used, which seems to work fine without adding the extra define line above. So I'll change it to use that instead.

per-mathisen-arm:
I believe this is ready to be merged now?

CMakeLists.txt
```diff
@@ -148,7 +150,14 @@ if(UNIX)
 endif(UNIX)
 
 # CMake provided find modules
-find_package(ZLIB)
+if(BUILD_STATIC)
+    find_library(ZLIB NAMES libz.a REQUIRED PATH_SUFFIXES lib lib/x86_64-linux-gnu lib64)
+    if(ZLIB)
+        message(STATUS "Found ZLIB: ${ZLIB}")
```
bradgrantham-lunarg, line 156: Is this just a debugging statement left in the CMakeLists.txt?  There are no other message()s in CMakeLists.txt.  Perhaps this should be a FATAL_ERROR if ZLIB is not found.

framework/graphics/fps_info.cpp
```diff
@@ -26,7 +26,7 @@
 #include "util/date_time.h"
 #include "util/logging.h"
 
-#include <inttypes.h>
+#include <cinttypes>
```
bradgrantham-lunarg, line 29: I'll put in a comment similar to the one I put before: We use the format macros everywhere with no errors without this #define. Is this specific to a compiler or system headers version?  Can you tell me more about why you had to add this?

bradgrantham-lunarg (changes requested):

framework/graphics/fps_info.cpp
per-mathisen-arm: The reason it works everywhere else is because cinttypes is used everywhere else. So I removed the #define and fixed the #include instead.

CMakeLists.txt
per-mathisen-arm: Ok, removed.

framework/graphics/fps_info.cpp
bradgrantham-lunarg: Sorry, let me try to be more clear.  This **file** compiles on other platforms, including Linux x86.  So why do you need to alter it to compile on ARM linux?

per-mathisen-arm: This actually works fine on the ARM platforms I've tried. The problem occurs on our x86-based HPC platforms, which can have fairly old toolchains. I believe the root cause is that traditionally, inttypes.h has required __STDC_FORMAT_MACROS to enable these macros for C++, while cinttypes removed this requirement earlier than inttypes.h. See https://cplusplus.github.io/LWG/issue984

If you want to enable using these format macros on the very oldest of toolchains, you should add __STDC_FORMAT_MACROS globally, but that is probably overkill as affected toolchains are now very old. Since changing to cinttypes works for us and makes things more internally consistent in gfxreconstruct, I see this as a reasonable solution in this case.

bradgrantham-lunarg (approved):

PR #553, opened 2021-07-16
Add measurement framerange support

tomped01:
This patch adds the option to specify start and stop frames for
secondary FPS measurements. In addition, options for aborting replay
after the end of the measurement range is reached and for flushing all
work at range boundaries have been added.

The framerange support is very useful when investigating performance for
large traces as creating new trimmed variants for every scene in the
trace can be very time consuming (and quickly becomes unmaintainable for
large sets of traces).
The abort option is very handy on slower development platforms where
replay of relatively small traces can take hours, this allows for
skipping work one might not be interested in for a particular run.

Change-Id: Ic3e81a53974d7955732e65337c2b8669f14ff887


tools/replay/replay_settings.h
```diff
@@ -75,11 +76,16 @@ const char kScreenshotRangeArgument[]          = "--screenshots";
 const char kScreenshotFormatArgument[]         = "--screenshot-format";
 const char kScreenshotDirArgument[]            = "--screenshot-dir";
 const char kScreenshotFilePrefixArgument[]     = "--screenshot-prefix";
+const char kMeasurementRangeArgument[]         = "--measurement-frame-range";
+const char kQuitAfterMeasurementRangeOption[]  = "--quit-after-measurement-range";
+const char kFlushMeasurementRangeOption[]      = "--flush-measurement-range";
 
 const char kOptions[] = "-h|--help,--version,--log-debugview,--no-debug-popup,--paused,--sync,--sfa|--skip-failed-"
-                        "allocations,--opcd|--omit-pipeline-cache-data,--remove-unsupported,--screenshot-all";
+                        "allocations,--opcd|--omit-pipeline-cache-data,--remove-unsupported,--screenshot-all"
```
lunarpapillo, line 84: There needs to be a `,` after `--screenshot-all` so it is recognized as distinct from the next switch.

Right now this is interpreted as a single switch `--screenshot-all--qamr`.

I'm also concerned that testing didn't exercise the new `--qamr` switch (which would have caught this issue).

lunarpapillo (changes requested):
The below is why the LunarG CI test run failed.

I'm also concerned about test coverage for the new options.  Neither LunarG CI nor the original tester exercised the new `--qamr` switch (which would have discovered the error).

tools/replay/replay_settings.h
tomped01: Sorry about that, fixed now!

Want me to look into the testing as well (i guess at the very least, raising a new issue would be a good idea)?

lunarpapillo: I agree.  Certainly automated testing needs to be added.  But the implementer here should have manually tested with the new options to make sure they work.  It's alarming to realize that this didn't occur.

I'd like an issue created that includes the manual test plan and the results of executing that plan.  The manual test plan so documented can then be used on both our side and yours to add automated tests.

tomped01: > But the implementer here should have manually tested with the new options to make sure they work.
Absolutely, i forgot about the --qamr and --fmr shortened options. Wont happen again.

Here is the issue, this should cover all the edge cases: https://github.com/LunarG/gfxreconstruct/issues/561

tomped01: Any more feedback on this? Its a pretty important feature on our end so would be nice if we could get it merged (or get some more pointers on what needs to be done to get it merged)

per-mathisen-arm:
Any chance that this could get merged soon?

tools/replay/replay_settings.h
bradgrantham-lunarg: I appreciate the issue created.  I think @lunarpapillo 's concern is that it didn't look like you had tested the changes you had made.  It's the responsibility of the PR submitter to test the code before we'll accept it.  Please note @lunarpapillo  said "I'd like an issue created that includes the manual test plan **and the results of executing that plan**." (my emphasis)  We will put a request for automated testing into our plan, but more important is whether you tested yourself with those options you've added, as Bob suggests?

framework/application/application.h
```diff
@@ -68,18 +73,27 @@ class Application
...
+    decode::FileProcessor*       file_processor_;           ///< The FileProcessor object responsible for decoding and processing
+                                                            ///< capture file data.
+    bool                         running_;                  ///< Indicates that the application is actively processing system
+                                                            ///< events for playback.
+    bool                         paused_;                   ///< Indicates that the playback has been paused.  When paused the
+                                                            ///< application will stop rendering, but will continue processing
+                                                            ///< system events.
+    std::string                  name_;                     ///< Application name to display in window title bar.
+    uint32_t                     pause_frame_;              ///< The number for a frame that replay should pause after.
+    int64_t                      measurement_start_time;    ///< The time when the measurement range started
```
bradgrantham-lunarg, line 95: These two should end in "_" to match the style guide.

bradgrantham-lunarg:
Could this functionality be accomplished just by replaying the capture with trimming, to make a cut-down capture, and then replay that capture with fps?

bradgrantham-lunarg (changes requested):

tools/replay/replay_settings.h
tomped01: > but more important is whether you tested yourself with those options you've added, as Bob suggests?
I did yes, it works fine now :)

I ran all the tests listed in the issue before submitting the pull request, but i always used the full length arguments, eg. --measurement-frame-range instead of --mfr, the bug was only exposed when you used the short version (eg. --mfr would break, --measurement-frame-range would not) so i assumed everything to be ok which was a mistake (essentially, i forgot to test one part of the argument parser code, i was too focused on the "internal" code i changed).

framework/application/application.h
tomped01: Fixed now!

bradgrantham-lunarg:
Changes applied in #697

PR #554, opened 2021-07-19
vk-to-ascii

patrick-lunarg:
This PR sets up "to ascii" functionality for Vulkan .gfxr captures.

A handful of generator/generated files are added to convert Vulkan structures and enums to string as well as some hand written files with string formatting utilities and handlers for the Vulkan structures and functions that need custom handling to stringify.

Note that this PR doesn't handle non-playback decoding that will be necessary to output certain handle ids...this logic may require some modifications to the base decoders that I'd prefer to put into its own review.

Note that the majority of formatting is handled in `to_string.h` in the functions `ToString()`, `ObjectToString()`, `FieldToString()`, and `ArrayToString()`.  As it stands the string output is formatted as JSON.

Example output:
```
"[217]vkCreateBuffer": {
    "return": "VK_SUCCESS",
    "device": "0x0000000000000004",
    "pCreateInfo": {
        "sType": "VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO",
        "pNext": null,
        "flags": 0,
        "size": 262144,
        "usage": 1,
        "sharingMode": "VK_SHARING_MODE_EXCLUSIVE",
        "queueFamilyIndexCount": 0,
        "pQueueFamilyIndices": []
    },
    "pAllocator": null,
    "[out]pBuffer": "0x0000000000000025"
},
"[218]vkGetBufferMemoryRequirements": {
    "device": "0x0000000000000004",
    "buffer": "0x0000000000000025",
    "[out]pMemoryRequirements": {
        "size": 262144,
        "alignment": 64,
        "memoryTypeBits": 7
    }
},
"[219]vkAllocateMemory": {
    "return": "VK_SUCCESS",
    "device": "0x0000000000000004",
    "pAllocateInfo": {
        "sType": "VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO",
        "pNext": null,
        "allocationSize": 262144,
        "memoryTypeIndex": 0
    },
    "pAllocator": null,
    "[out]pMemory": "0x0000000000000026"
},
"[220]vkBindBufferMemory": {
    "return": "VK_SUCCESS",
    "device": "0x0000000000000004",
    "buffer": "0x0000000000000025",
    "memory": "0x0000000000000026",
    "memoryOffset": 0
},
"[221]vkMapMemory": {
    "return": "VK_SUCCESS",
    "device": "0x0000000000000004",
    "memory": "0x0000000000000026",
    "offset": 0,
    "size": 18446744073709551615,
    "flags": 0,
    "[out]ppData": "0x0000000BA4EFF2D0"
},
"[222]vkUnmapMemory": {
    "device": "0x0000000000000004",
    "memory": "0x0000000000000026"
}
```


lunarpapillo:
The output format is curious, certainly not what I was expecting.  Before this goes public, we should be absolutely certain that this format meets the needs of its intended use patterns (as after it goes public we will likely have to support whatever our selected format was forever).

Is there a list of use cases somewhere, preferably someplace (like GitHub or Google Docs) where we can offer commentary?

If not, we can do a video meeting...

patrick-lunarg:
> The output format is curious, certainly not what I was expecting. Before this goes public, we should be absolutely certain that this format meets the needs of its intended use patterns (as after it goes public we will likely have to support whatever our selected format was forever).
> 
> Is there a list of use cases somewhere, preferably someplace (like GitHub or Google Docs) where we can offer commentary?
> 
> If not, we can do a video meeting...

Absolutely...FWIW, nearly (there's a handful of structs and function that need custom handling) all of the formatting is isolated into the functions `ObjectToString()`, `FieldToString()`, and `ArrayToString()`...those can be adjusted to output whatever we need.  As we discussed before, I'm a big fan of JSON, but if there's a different format that this should output it should be very straightforward to modify the formatting logic.

lunarpapillo:
> I'm a big fan of JSON, but if there's a different format that this should output it should be very straightforward to modify the formatting logic.

We may even want to (eventually) support multiple formats for multiple use cases - but we can add such a setting later.

patrick-lunarg:
> > I'm a big fan of JSON, but if there's a different format that this should output it should be very straightforward to modify the formatting logic.
> 
> We may even want to (eventually) support multiple formats for multiple use cases - but we can add such a setting later.

It would actually be pretty straight forward to do...everything passes a ToStringFlags bitmask through, when it hits the formatting functions we can read the bitmask and do the thing.

locke-lunarg:
You might need to merge or clean up some commits. 49 commits seem too many. Some commits could be merged into one.

patrick-lunarg:
> You might need to merge or clean up some commits. 49 commits seem too many. Some commits could be merged into one.

I was planning on squashing all the commits once it's passed review

framework/decode/custom_vulkan_ascii_consumer.h
```diff
@@ -0,0 +1,215 @@
...
+{
+    VkWriteDescriptorSet                         writeDescriptorSet{};
+    VkWriteDescriptorSetAccelerationStructureKHR writeDescriptorSetAccelerationStructure{};
+    if (pObj)
+    {
+
+        writeDescriptorSet.sType = VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET;
+        if (pObj->GetImageInfoCount())
+        {
+            writeDescriptorSet.descriptorCount = (uint32_t)pObj->GetImageInfoCount();
```
locke-lunarg, line 111: For converting, we prefer C++ style, like static_cast<uint32_t>.

framework/decode/custom_vulkan_to_string.cpp
```diff
@@ -0,0 +1,443 @@
...
+    );
+}
+
+template <>
+std::string ToString<VkAccelerationStructureGeometryDataKHR>(const VkAccelerationStructureGeometryDataKHR& obj, ToStringFlags toStringFlags, uint32_t tabCount, uint32_t tabSize)
+{
+    return ObjectToString(toStringFlags, tabCount, tabSize,
+        [&](std::stringstream& strStrm)
+        {
+            FieldToString(strStrm, true, "triangles", toStringFlags, tabCount, tabSize, ToString(obj.triangles, toStringFlags, tabCount, tabSize));
```
locke-lunarg, line 72: VkAccelerationStructureGeometryDataKHR is union. Here, it prints every member. In Dx12, it only prints the used member.  I just wonder which format style is our prefer. @bradgrantham-lunarg

framework/decode/custom_vulkan_ascii_consumer.h
patrick-lunarg: Done

framework/decode/custom_vulkan_to_string.cpp
patrick-lunarg: I chose to print every member for all unions (except for `VkPerformanceValueDataINTEL` because one of them is a `char*`).  In my opinion it's simpler to print them all so that unions can be stringified without context, but I'm definitely willing to setup code that will drill into unions from structures that provide that context and only print the necessary field, if that's needed.

locke-lunarg:
It needs to present if the argument is output.
It might look like this or some way.
```
"[217]vkCreateBuffer": {
     ...
    "pBuffer": "&0x0000000000000025"
},
```

patrick-lunarg:
> It needs to present if the argument is output.
> It might look like this or some way.
> 
> ```
> "[217]vkCreateBuffer": {
>      ...
>     "pBuffer": "&0x0000000000000025"
> },
> ```

I updated it so it looks like...
```
"[217]vkCreateBuffer": {
    ...
    "[out]pBuffer": "0x0000000000000025"
},
```

locke-lunarg:
Looks good to me except for cleaning up commits. Plus, we might need to discuss what output format is our preference. Dx12's is not JSON format. But we could do it later.

patrick-lunarg:
> Looks good to me except for cleaning up commits. Plus, we might need to discuss what output format is our preference. Dx12's is not JSON format. But we could do it later.

Whoever merges it should be able to select squash merge, if not I can do it manually before the PR is accepted

locke-lunarg:
> > Looks good to me except for cleaning up commits. Plus, we might need to discuss what output format is our preference. Dx12's is not JSON format. But we could do it later.
> 
> Whoever merges it should be able to select squash merge, if not I can do it manually before the PR is accepted

Usually, the PR's owner merges it after the PR gets approved, so you need to do it on your own. One PR usually needs two approved, or one approved if it's a small change.

PR #562, opened 2021-08-11
Fix compiling with gcc11

Samsuper12:


patrick-lunarg:
I think if you `#include <iterator>` the Windows builds will succeed.

PR #543, opened 2021-06-09
Arm1

per-mathisen-arm:
Fix Arm linux builds, allow setting a vendor-specific version suffix when building, and add a Travis-CI build automation file to help prevent further breakage of less commonly tested linux platforms. Hopefully the main repo can add integration with Travis-CI (free for open source projects) so this check is run for every commit.

PR #566, opened 2021-08-13
Fix Arm linux builds, allow adding version suffix, and add automatic linux build target testing

bradgrantham-lunarg:
Fix Arm linux builds, allow setting a vendor-specific version suffix when building, and add a Travis-CI build automation file to help prevent further breakage of less commonly tested linux platforms. Hopefully the main repo can add integration with Travis-CI (free for open source projects) so this check is run for every commit.

This is a resubmit of #543, which I merged prematurely and then had to revert.


CMakeLists.txt
```diff
@@ -42,13 +42,14 @@ set(GFXRECONSTRUCT_PROJECT_VERSION_MAJOR 0)
...
 set(GFXRECONSTRUCT_PROJECT_VERSION_PATCH 8)
 
-set(GFXRECON_PROJECT_VERSION_DESIGNATION "-unknown")
 set(GFXRECON_PROJECT_VERSION_SHA1 "unknown-build-source")
 
 if (NOT DEFINED HEADLESS)
     set(HEADLESS TRUE)
 endif()
 
+option(BUILD_ZSTD_SUPPORT "Build zstd support" OFF)
```
bradgrantham-lunarg, line 51: @per-mathisen-arm I think I said incorrectly that this was default OFF, so I think you had this right the first time.  If you could please add a commit to this PR with the default being ON, we'll re-run CI.  It might be that you have to commit that to your branch and I'll have to update the PR; not sure how that will work but I'll check on it in the morning.

per-mathisen-arm: Done. I've also split out the other changes as their own PRs to make this easier to handle. Although the automation PR will depend on this PR, which is less than optimal.

CMakeLists.txt
```diff
@@ -49,6 +49,8 @@ if (NOT DEFINED HEADLESS)
     set(HEADLESS TRUE)
 endif()
 
+option(BUILD_ZSTD_SUPPORT "Build zstd support" ON)
```
bradgrantham-lunarg, line 52: I sent this as an email to you last week, but maybe it went into spam: The consensus internally is that if you pass ZSTD_ROOT for CMake on the command line to the root dir of a ZSTD download, FindZSTD.cmake should find your include and lib and use those in the build.  That's actually what our internal CI does.  So I don't think you need to specify anything related to ZSTD in this PR.

CMakeLists.txt
```diff
@@ -125,15 +127,13 @@ elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
...
-elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "arm")
-    set(CMAKE_PREFIX_PATH
-        ${CMAKE_PREFIX_PATH}
-       "${CMAKE_SOURCE_DIR}/external/precompiled/linux/lib/arm")
 endif()
 
 # GFXReconstruct provided find modules
 find_package(LZ4)
-find_package(ZSTD)
+if(BUILD_ZSTD_SUPPORT)
```
bradgrantham-lunarg, line 134: I sent this as an email to you last week, but maybe it went into spam: The consensus internally is that if you pass ZSTD_ROOT for CMake on the command line to the root dir of a ZSTD download, FindZSTD.cmake should find your include and lib and use those in the build.  That's actually what our internal CI does.  So I don't think you need to specify anything related to ZSTD in this PR.

external/precompiled/linux/include/zconf.h
```diff
@@ -0,0 +1,534 @@
+/* zconf.h -- configuration of the zlib compression library
```
bradgrantham-lunarg, line 1: I'm so sorry to keep dragging out your PR, but shouldn't the build find zconf.h for your build like it finds zconf for the 32-bit and 64-bit x86 builds?

per-mathisen-arm:
Sorry about the slow response time. I am a bit confused about this. For cross-compiled environments, the problem is generally that you do not have many installed libraries available. The general way to deal with this is either to include them as submodules and compile them as needed, or (far more ugly) to include binary library blobs in git. gfxreconstruct seems to have opted for binaries, so we followed suit adding that also for the arm arch. In order to compile against these binary blobs, you also need the relevant headers added (you do not want to use random library headers from the host system, as these might have a different version). So that is why I added the zconf header, and if we want to use zstd, we'll need to add a binary blob + header for that as well. You can, of course, leave it up to the user to create a build environment that contains these libraries already installed for the target architectures, but unless you also provide instructions on how to do that, it is often a fairly large hurdle to overcome. Please advise which type of solution you would prefer, and we'll redo the patch to match to match.

panos-lunarg:
Hi Per
I'll attempt to resurrect this PR in order to get it merged.
We agree with you that including binaries in the repo is not a nice solution. However we don't have a strong feel about it so feel free to add it however you like, either as a binary, or as a sub-module, or as something completely external that cmake can be pointed at it by setting `ZSTD_ROOT`.

per-mathisen-arm:
Hi, can this please be merged so that arm linux builds can be done with less hassle? Currently the precompiled support for this platform is not working correctly. I agree it is not a great solution but it is a small change to remove a big annoyance for end users in that it won't cross-compile out of the box.

bradgrantham-lunarg:
Hey, @per-mathisen-arm , this looks like it now rebases to `dev` with a couple of small changes to `CMakeLists.txt`.  I don't yet know how to rebase locally but make a PR to your PR that adds one commit, nor does merge seem to do what I want.  Instead, I know I can create a new PR from your change, keeping you as the commit author, or you can fix conflicts yourself.  Let me know what you'd like!

bradgrantham-lunarg:
Oh, this is already a PR by me with your commit.  :laughing:  Still, it's from your branch, so let me know what you'd like to do.

per-mathisen-arm:
I'm off traveling for a few days, so if you want to merge it soon, feel free to commit it whichever way and under whichever name suits you :-)

PR #570, opened 2021-08-17
Add automatic build testing for linux arm build targets

per-mathisen-arm:


per-mathisen-arm:
This PR depends on PR 566 to work.

lunarpapillo (commented):
This code will work, but I'm curious about why the choice was made to run three architecture-specific builds in sequence, instead of adding the architectures to the `matrix`, e.g.:
```
      matrix:
        config:
        - {
            name: "ARM64 Release",
            os: ubuntu-latest,
            artifact: "gfxreconstruct-dev-arm64-release",
            type: "release",
            build_dir: "build",
            cc: "gcc", cxx: "g++",
            arch: "arm64"
          }
        - {
            name: "ARM32 Debug",
            os: ubuntu-latest,
            artifact: "gfxreconstruct-dev-arm32-debug",
            type: "debug",
            build_dir: "dbuild",
            cc: "gcc", cxx: "g++",
            arch: "arm32"
          }
```
and pass `--arch {{ matrix.config.arch }}` to `build.py` (you'll of course have to add `arch: "x64"` to the existing Ubuntu matrix configurations).

Using separate matrix configurations can allow for all the builds to happen in parallel, rather than in sequence, and offers more control of which exact configurations are built and tested.  It also makes it easier to identify what builds failed, and to keep their artifacts separate.

Of course, ignore this if there's a good reason to run the build in groups.

bradgrantham-lunarg:
Superceded by #596

PR #579, opened 2021-09-23
wsi-context

patrick-lunarg:
This PR breaks the `Application` class into two parts.  The first part remains `Application` and is responsible for reading the captured stream file, advancing the stream, preparing API calls for decoding, etc.  The second part is `WsiContext`, this is responsible for handling message procedures, window callbacks, etc.  In addition, each 0f the WSI specific extended `Application` classes now extend `WsiContext`.

This change is necessary due to the need for `Application` to be created before we're reading the captured stream...this meant that the WSI specific `Application` object had to be created before we were aware of what WSI system was actually used at capture time.  Now we're able to create a `WsiContext` after creating `Application`.

### DIFF
--- a/framework/application/application.cpp
+++ b/framework/application/application.cpp
@@ -22,8 +22,26 @@
 */
 
 #include "application/application.h"
-
 #include "util/logging.h"
+
+#if defined(VK_USE_PLATFORM_WIN32_KHR)
+#include "application/win32_context.h"
+#endif
+#if defined(VK_USE_PLATFORM_WAYLAND_KHR)
+#include "application/wayland_context.h"
+#endif
+#if defined(VK_USE_PLATFORM_XCB_KHR)
+#include "application/xcb_context.h"
+#endif
+#if defined(VK_USE_PLATFORM_XLIB_KHR)
+#include "application/xlib_context.h"
+#endif
+#if defined(VK_USE_PLATFORM_ANDROID_KHR)
+#include "application/android_context.h"
+#endif
+#if defined(VK_USE_PLATFORM_HEADLESS)
+#include "application/headless_context.h"
+#endif
 
 #include <algorithm>
 #include <cassert>
@@ -31,30 +49,16 @@
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(application)
 
-Application::Application(const std::string& name) :
-    file_processor_(nullptr), running_(false), paused_(false), name_(name)
-{}
-
-Application::~Application()
+ApplicationEx::ApplicationEx(const std::string& name, decode::FileProcessor* file_processor) :
+    name_(name), file_processor_(file_processor), running_(false), paused_(false), pause_frame_(0)
 {
-    if (!windows_.empty())
-    {
-        GFXRECON_LOG_INFO(
-            "Application manager is destroying windows that were not previously destroyed by their owner");
-
-        for (auto window : windows_)
-        {
-            delete window;
-        }
-    }
 }
 
-void Application::SetFileProcessor(decode::FileProcessor* file_processor)
+ApplicationEx ::~ApplicationEx()
 {
-    file_processor_ = file_processor;
 }
 
-void Application::Run()
+void ApplicationEx::Run()
 {
     running_ = true;
 
@@ -70,9 +74,8 @@ void Application::Run()
     }
 }
 
-void Application::SetPaused(bool paused)
+void ApplicationEx::SetPaused(bool paused)
 {
-
     paused_ = paused;
 
     if (paused_ && (file_processor_ != nullptr))
@@ -85,7 +88,7 @@ void Application::SetPaused(bool paused)
     }
 }
 
-bool Application::PlaySingleFrame()
+bool ApplicationEx::PlaySingleFrame()
 {
     bool success = false;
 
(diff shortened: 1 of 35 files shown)


johnzupin:
I was testing this PR out on Ubuntu 21.04 using Wayland. I'm able to capture and replay traces but I'm seeing something funny.  After re-capturing Sacha Willem's 'computecullandlod' trace,  I'm seeing flashes of what looks like junk in the buffer from a previous run of another sample. Most of the time it looks like some sort corruption for one frame length, but if I run the 'computecloth' sample (note I'm running the sample, not replaying a trace of it), sometime images can clearly be seen of that 'computecloth' sample during replay of the captured trace.

So in short, I'm seeing problems replaying a trace of SW's compute cull and lod trace.

patrick-lunarg:
Very interesting,

Thanks for the feedback...
I'll try to repro on Monday and get that squared away before we merge it.

-Patrick

bradgrantham-lunarg (changes requested):

framework/application/application.h
```diff
@@ -24,31 +24,38 @@
...
+#include <memory>
 #include <string>
+#include <utility>
 #include <vector>
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(application)
 
-class Application
+class Application final
```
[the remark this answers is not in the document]
PatrickPurcell: There's no "bad" behavior, per say, but I was breaking the functionality of this class (that was used as base) into a different object so that it could be allocated later.  The `final` is really just there so I can be sure that nothing was inheriting from this class any longer.  I can remove it if you prefer.

framework/application/xcb_context.cpp
```diff
@@ -21,32 +21,22 @@
...
-}
-
-bool XcbApplication::Initialize(decode::FileProcessor* file_processor)
+XcbContext::XcbContext(Application* application) : WsiContext(application)
 {
+    xcb_loader_.Initialize();
     if (!xcb_loader_.Initialize())
     {
-        return false;
+        GFXRECON_LOG_DEBUG("Failed initialize XCB loader");
```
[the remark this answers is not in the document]
PatrickPurcell: Done

framework/application/application.h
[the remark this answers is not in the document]
bradgrantham-lunarg: I think using it to debug the PR development is a great tactic!  I can't think of a good reason for Application to be subclassed at this time, since the WSI part has been factored out.  It just made me wonder if there was something in there that must not be overridden.  I think it's okay to leave it and we'll figure it out later if we really need to start subclassing Application.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2084,8 +2089,16 @@ VkResult VulkanReplayConsumerBase::CreateSurface(InstanceInfo*
     if ((options_.surface_index == -1) || (options_.surface_index == create_surface_count_))
     {
         // Create a window for our surface.
-        Window* window = window_factory_->Create(
-            kDefaultWindowPositionX, kDefaultWindowPositionY, kDefaultWindowWidth, kDefaultWindowHeight);
+        assert(application_);
```
bradgrantham-lunarg, line 2092: Is it actually possible to get to this point in this function (and the others in vulkan_replay_consumer_base.cpp that check wsi_context and window_factory with assert and/or ternary operator) without wsi_context and window_factory being valid?

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_replay_consumer_base.cpp
patrick-lunarg: It _shouldn't_ be possible to get there with nulls...since null derefs are guarded here, if they are null it'll cause errors later down the line (still shouldn't crash though, just spit out an error message and bail).  The `assert()` is just to make sure that if I made any mistakes in the logic leading up to there it will be instantly obvious.

framework/application/application.cpp
```diff
@@ -117,37 +117,65 @@ bool Application::PlaySingleFrame()
...
-    auto pos = std::find(windows_.begin(), windows_.end(), window);
-
-    if (pos == windows_.end())
+    if (!wsi_context_)
     {
-        GFXRECON_LOG_INFO(
-            "A remove window request was made for an window that was never registered with the application");
-        return false;
+#if defined(VK_USE_PLATFORM_WIN32_KHR)
+        if (!strcmp(surfaceExtensionName, VK_KHR_WIN32_SURFACE_EXTENSION_NAME))
```
davidd-lunarg, line 133: There is a GFXR utility function for string compare `gfxrecon::util::platform::StringCompare`.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2377,17 +2390,25 @@ VulkanReplayConsumerBase::OverrideCreateInstance(VkResult original_result,
...
+            assert(application_);
+            auto wsi_context       = application_->GetWsiContext();
+            auto window_factory    = wsi_context ? wsi_context->GetWindowFactory() : nullptr;
+            auto cli_wsi_extension = window_factory ? window_factory->GetSurfaceExtensionName() : nullptr;
+            if (cli_wsi_extension)
+            {
+                filtered_extensions.push_back(cli_wsi_extension);
+            }
+
             // Swap the surface extension supported by platform the replay is running on if different from trace
```
davidd-lunarg, line 2404: Is this comment still accurate? It looks like the swap may have been removed.

patrick-lunarg: Removed

PR #581, opened 2021-09-28
Add an option to the capture script use a capture layer in a given directory

dsmolars:
This is useful for
1. Make it easier to use GFXR even when the SDK is not installed
2. Make it easier to debug and test versions of GFXR during development.

### DIFF
--- a/tools/capture/gfxrecon-capture.py
+++ b/tools/capture/gfxrecon-capture.py
@@ -216,7 +216,7 @@ def ValidateArgs(args):
                     path)
             if not valid_capture_layer:
                 PrintErrorAndExit(
-                    'Failed to find valid capture layer in VK_LAYER_PATh environment variable'
+                    'Failed to find valid capture layer in VK_LAYER_PATH environment variable'
                 )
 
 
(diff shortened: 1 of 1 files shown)


bradgrantham-lunarg:
I couldn't verify the author had signed the CLA at the time.  Although the author had been working with a client and had submitted changes under that agreement, it's now not clear (after the author has changed companies) that the CLA is valid.  @xykivo I would like to merge this if you'll take a second to click through the CLA.

xykivo:
CLA signed.

Good luck.

per-mathisen-arm:
I think this change is a regression and is doing the wrong thing. First, it requires an environment variable to be present or a command line option to be used in order to use the tool, so it makes usage harder. Second, it requires the .so file and the .json file to be in the same directory, but the default 'make install' for gfxreconstruct puts them in separate directories. I think VK_LAYER_PATH should just point to the .json directory, and the .so file's location should be read from the JSON, rather than making any hardcoded assumptions about the .so file's location. Note that the default 'make install' for gfxreconstruct does _not_ set this JSON path correctly, either, as it again mistakenly assumes that the .so file will end up side-by-side with the JSON.

PR #591, opened 2021-10-08
Add thread support for multi-threaded drivers

tomek-brcm:
The libstdc++ must be initialised in multi-threaded mode otherwise shared
lib added with dlopen() won't have access to std::thread.


tomek-brcm:
Fixes #592

patrick-lunarg (approved):

PR #604, opened 2021-11-03
Update header to 1.2.196

locke-lunarg:


framework/decode/custom_vulkan_to_string.h
```diff
@@ -159,7 +159,11 @@ inline std::string BitmaskToString(VkFlags vkFlags)
             {
                 str.append("|");
             }
+#if defined(WIN32)
+            str.append(ToString(static_cast<VkBitmaskType>(1i64 << index)));
+#else
             str.append(ToString(static_cast<VkBitmaskType>(1 << index)));
+#endif
```
locke-lunarg, line 166: The new flag, VkFormatFeatureFlagBits2KHR, value is like `0x00000001ULL`. It caused Windows had a warning
>  warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?) 

so it needs to use `1i64`, instead of `1`. But Linux and Android don't have this warning and they don't support `1i64`.

patrick-lunarg (approved):

PR #607, opened 2021-11-10
Update header to 1.2.198

locke-lunarg:


bradgrantham-lunarg (approved):
Approved but it would be a good idea to make sure we can capture, capture with trimming, playback, and toascii a program that uses dynamic rendering.

locke-lunarg:
@bradgrantham-lunarg  Fixed VK_KHR_dynamic_rendering issue

patrick-lunarg (approved):

Issue #583, opened 2021-09-28
WSI Selection May Result In Incorrect WSI During Playback


patrick-lunarg:
GFX Reconstruct currently only supports a single WSI system during playback.  The WSI system used is selected using the first WSI extension requested at `VkInstance` creation time.  If an application loads multiple WSI extensions, there is no guarantee that playback will select the correct one.

This can be addressed by maintaining a collection of WsiContext objects with an entry for each WSI extension in the `VkInstanceCreateInfo`.

patrick-lunarg:
This issue was cleared with PR #619

PR #608, opened 2021-11-11
[patrick-lunarg][wsi-selection]

patrick-lunarg:
This PR enables fallback WSI selection for gfxr replay.

The logic in this PR results in the following behavior...

If there is a WSI selection on the CLI, gfxr-replay will attempt to use that extension.
If the CLI WSI selection fails, gfxr-replay will attempt to fallback to the WSI extension used at capture time.
If the capture time WSI selection fails, gfxr-replay will attempt to fallback to an available WSI extension.

Note that there is no guarantee that WSI configuration will be successful if the playback WSI selection is different from the capture time selection.

This should address https://github.com/LunarG/gfxreconstruct/issues/583


framework/application/android_window.cpp
```diff
@@ -111,6 +111,12 @@ bool AndroidWindow::GetNativeHandle(HandleType type, void** handle)
     }
 }
 
+const std::string& AndroidWindow::GetWsiExtension() const
+{
+    static std::string sWsiExtension { VK_KHR_ANDROID_SURFACE_EXTENSION_NAME };
+    return sWsiExtension;
```
bradgrantham-lunarg, line 117: Can this just return `VK_KHR_ANDROID_SURFACE_EXTENSION_NAME` and let the return construct the std::string?

bradgrantham-lunarg (changes requested):

framework/application/application.cpp
```diff
@@ -54,11 +54,114 @@ GFXRECON_BEGIN_NAMESPACE(gfxrecon)
...
+    }
+
+    // If auto_select is enabled and we still don't have a valid WSI context, loop
+    //  through loaded extensions and return the first one that _isn't_ the one
+    //  specified on the CLI or requested by the current API call
+    if (auto_select && itr == wsi_contexts_.end())
+    {
+        for (const auto& wsi_context_itr : wsi_contexts_)
+        {
+            if (wsi_context_itr.first != cli_wsi_extension_ && wsi_context_itr.first != wsi_extension &&
```
bradgrantham-lunarg, line 144: Do the previous two if-statements guarantee that you'd never find `cli_wsi_extension_` or `wsi_extension` in this loop?  If so, I suggest dropping the if-statement.

bradgrantham-lunarg (changes requested):

bradgrantham-lunarg (approved):

framework/application/application.cpp
patrick-lunarg: You're correct...will remove

framework/application/android_window.cpp
patrick-lunarg: done

framework/application/application.cpp
patrick-lunarg: done

framework/application/application.cpp
```diff
@@ -54,11 +54,109 @@ GFXRECON_BEGIN_NAMESPACE(gfxrecon)
...
+    if (itr == wsi_contexts_.end())
+    {
+        itr = wsi_contexts_.find(wsi_extension);
+    }
+
+    // If auto_select is enabled and we still don't have a valid WSI context, loop
+    //  through loaded extensions and return the first one we have
+    if (auto_select && itr == wsi_contexts_.end())
+    {
+        for (const auto& wsi_context_itr : wsi_contexts_)
```
bradgrantham-lunarg, line 141: Sorry to iterate, I didn't think about what this would end up looking like.  If the map is empty, I think this falls through, otherwise it returns `wsi_contexts_.begin()->second.get()`, right?   Could you just replace this entire for loop with `itr = wsi_contexts_.begin();`?  Then it will take advantage of and look like the previous logic and use your return statement below.

bradgrantham-lunarg (changes requested):

framework/application/application.cpp
patrick-lunarg: For sure

patrick-lunarg: Done

bradgrantham-lunarg (approved):

Issue #551, opened 2021-07-15
Assertion hit when trying to trace vkd3d-proton


Oschowa:
gfxrecon version: v0.9.8
OS: Arch Linux
Driver: Mesa recent main

When trying to trace content running with [vkd3d-proton v2.4](https://github.com/HansKristian-Work/vkd3d-proton) with gfxrecon, I hit the following assertion:  
```
src/gfxreconstruct/framework/encode/vulkan_state_tracker.cpp:275: void gfxrecon::encode::VulkanStateTracker::TrackEndRenderPass(VkCommandBuffer): Assertion `attachment_count <= render_pass_wrapper->attachment_final_layouts.size()' failed.
fish: Job 1, 'VK_INSTANCE_LAYERS="VK_LAYER_LU…' terminated by signal SIGABRT (Abort)
```

This happens both with d3d12 games (tested DEATH STRANDING and Scarlet Nexus on Proton Experimental), as well as with the `gears` demo app from vkd3d-proton.
  
To reproduce:
- clone the vkd3d-proton: `git clone --recurse-submodules https://github.com/HansKristian-Work/vkd3d-proton`
- configure with `meson -D enable_extras=true`
- build and then try to trace the resultion `gears` program with gfxrecon

mohanku:
Seeing same assert and crash in some android apps like:
https://play.google.com/store/apps/details?id=com.MoodGames.QueenBee&hl=en_US&gl=US
https://play.google.com/store/apps/details?id=com.linear.beltit&hl=en_US&gl=US
 https://play.google.com/store/apps/details?id=com.nagastudio.giant.monster.run&hl=en_US&gl=US

patrick-lunarg:
Currently I cannot repro this issue with the `gears` demo included with `vkd3d-proton`.

`gfxrecon` latest `dev` branch (as of 9 Sept 2021)
Arch Linux 2021.09.01
AMD RX 6800 XT
Mesa RADV driver

Tried with both `vkd3d-proton` `v2.4` and the latest `master` branch (as of 9 Sept 2021)...couldn't repro with either.

@Oschowa do you have any more info that might help me repro?

Thanks

Oschowa:
It turns out i'm also only hitting this assert when using `GFXRECON_CAPTURE_TRIGGER="F12"`, without the capture trigger it seems to work fine.

patrick-lunarg:
That's interesting...Ok, I'll take a look at that.
Thanks

[...]

PR #625, opened 2022-01-05
patrick-trim-capture-proton

patrick-lunarg:
This PR repairs trimmed capture of workloads utilizing VKD3D/Proton as reported in issue #551 .

This is achieved by improving trimmed capture handling the extensions `VK_KHR_create_renderpass2` and `VK_KHR_push_descriptor`.


framework/encode/custom_vulkan_command_buffer_util.cpp
```diff
@@ -0,0 +1,195 @@
...
+GFXRECON_BEGIN_NAMESPACE(encode)
+
+void TrackCmdPushDescriptorSetKHRHandles(CommandBufferWrapper*       wrapper,
+                                         VkPipelineLayout            layout,
+                                         uint32_t                    descriptorWriteCount,
+                                         const VkWriteDescriptorSet* pDescriptorWrites)
+{
+    assert(wrapper);
+
+    if (layout)
```
bradgrantham-lunarg, line 40: Can this and the other tests of a handle or pointer in a boolean context compare against the null object?  Then it will be similar to the generated `TrackCmd` functions and the hand-written `Track` commands in `framework/encode/vulkan_state_tracker.cpp`.

bradgrantham-lunarg (changes requested):

framework/encode/custom_vulkan_command_buffer_util.cpp
patrick-lunarg: Done

bradgrantham-lunarg (approved):

PR #626, opened 2022-01-05
exit with failure if FileProcessor can't be initialized

### LINKED ISSUE #567: gfxrecon-compress fails with 0 exit code when using an unsupported compression format
We recently saw an issue with a build compiled inadvertently without ZSTD support.  When a test attempted to convert the compression of a trace from ZLIB to ZSTD, it provoked error messages on `stderr`:
```
[gfxrecon] ERROR - Failed to initialize compression module: Zstandard compression is disabled.
[gfxrecon] ERROR - Failed to initialized file compression module (type = 3); processing of compressed data will not be possible
```
and failed to produce an output file - but still exited with a 0 exit code, which universally means success.

In any case where `gfxrecon-compress` fails to create an output file, it should report a nonzero exit code.  Ideally, the exit code would be unique for each potential error situation and could be used to figure out what went wrong without having to parse the `stderr`.  But I can live with just a nonzero exit code.

bradgrantham-lunarg:
Fixes #567 for the narrow case of a non-zero error code when `gfxrecon-compress` fails, see #627 for an extended task to review exit codes.

### DIFF
--- a/tools/compress/main.cpp
+++ b/tools/compress/main.cpp
@@ -141,8 +141,6 @@ static std::string GetCompressionTypeName(uint32_t type)
 
 int main(int argc, const char** argv)
 {
-    int return_code = 0;
-
     gfxrecon::util::Log::Init();
 
     gfxrecon::util::ArgumentParser arg_parser(argc, argv, kOptions, "");
@@ -249,16 +247,18 @@ int main(int argc, const char** argv)
         else
         {
             GFXRECON_WRITE_CONSOLE("Capture file %s could not be converted.", input_filename.c_str());
-            return_code = -1;
+            gfxrecon::util::Log::Release();
+            exit(-1);
         }
     }
     else
     {
         GFXRECON_WRITE_CONSOLE("CompressionConverter could not be initialized.");
+        gfxrecon::util::Log::Release();
         exit(-1);
     }
 
     gfxrecon::util::Log::Release();
 
-    return return_code;
+    return 0;
 }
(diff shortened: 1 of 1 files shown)


tools/compress/main.cpp
```diff
@@ -252,6 +252,11 @@ int main(int argc, const char** argv)
             return_code = -1;
         }
     }
+    else
+    {
+        GFXRECON_WRITE_CONSOLE("CompressionCoverter could not be initialized.");
+        exit(-1);
```
lunarpapillo, line 258: Would it be better/more consistent/allow for proper cleanup to have `return_code = -1;` like in the failure case above?

lunarpapillo (changes requested):
Simple and to the point, but I think a slight difference might be better?

tools/compress/main.cpp
bradgrantham-lunarg: That's a subtle one - I was going to do `return_code` but `exit(-1)` is already used in 3 other places in `main` so I judged `exit` to be more consistent.  I'll dwell on it today; it's slightly different in that the `exit` is in the `else` very far away from the function that precipitated it.

bradgrantham-lunarg: I surveyed the other tools; they use `exit` for failure handling except the last two failure possibilities in `optimize`.  So I went the other direction, using exit() for failures through the code in `compress`.  Does that work for you @lunarpapillo ?

bradgrantham-lunarg: Additionally I realized I didn't close the Log on exit while making this change, so fixed that.

bradgrantham-lunarg: @lunarpapillo

lunarpapillo (approved):
Woot!  Thanks for this...

PR #641, opened 2022-02-04
pass the layer name downchain to EnumerateDeviceExtensionProperties

bradgrantham-lunarg:
and fix clang-format for lines I didn't touch just in case CI will kick it back


patrick-lunarg (approved):

charles-lunarg (approved):
Fun how some bugs are submarines, hiding for years before surfacing and causing much confusion.

Issue #157, opened 2019-01-04
gfxrecon-toascii needs to support arbitrary path


lunarpapillo:
gfxrecon-toascii currently creates a .txt file in the same directory as the trace.

It should take command-line options to allow the .txt file to go to any desired location (including stdout, for use in pipes).

patrick-lunarg:
> @patrick-lunarg can you assign me for creating that feature?

Hey @Samsuper12 ,

Just checking in to see if you've made any progress on this issue.
If not, that's fine...just trying to determine where this issue is at for bug/feature prioritization.

Thanks,
-Patrick

Samsuper12:
@patrick-lunarg Hi!

Now I work with some ASM optimizations in Mesa and allocate 100% of my time to resolve that. 

If you have any tasks for me, I'll be happy in the near future to close those after main work.

About that issue: progress is insignificant, so we can forget about this issue.

Thanks,
- Michael

patrick-lunarg:
> @patrick-lunarg Hi!
> 
> Now I work with some ASM optimizations in Mesa and allocate 100% of my time to resolve that.
> 
> If you have any tasks for me, I'll be happy in the near future to close those after main work.
> 
> About that issue: progress is insignificant, so we can forget about this issue.
> 
> Thanks, -Michael

Thanks for the update.
I'll reassign this task to myself, not sure exactly when we'll get to it.

bradgrantham-lunarg:
Fixed by #647

PR #644, opened 2022-02-08
Add output path option to toascii

patrick-lunarg:
This PR adds a command line option to `toascii` that enables a user specified output path.

This addresses issue #157


tools/toascii/main.cpp
```diff
@@ -58,53 +48,19 @@ static void PrintUsage(const char* exe_name)
...
     GFXRECON_WRITE_CONSOLE("  --version\t\tPrint version information and exit.");
+    GFXRECON_WRITE_CONSOLE("  --output <filepath>\tFilepath, relative to the current working directory, to write JSON "
+                           "output to. Default is <.gfxr filepath/filename>.txt.");
 #if defined(WIN32) && defined(_DEBUG)
     GFXRECON_WRITE_CONSOLE("  --no-debug-popup\tDisable the 'Abort, Retry, Ignore' message box");
     GFXRECON_WRITE_CONSOLE("        \t\tdisplayed when abort() is called (Windows debug only).");
 #endif
 }
 
-static bool CheckOptionPrintUsage(const char* exe_name, const gfxrecon::util::ArgumentParser& arg_parser)
```
patrick-lunarg, line 67: Everything removed from this file was already defined in `tool_settings.h` which is now included in this .cpp file.

framework/decode/vulkan_ascii_consumer_base.cpp
```diff
@@ -39,28 +39,31 @@ VulkanAsciiConsumerBase::~VulkanAsciiConsumerBase()
 
 bool VulkanAsciiConsumerBase::Initialize(const std::string& filename)
 {
-    bool success = false;
-
-    if (m_file == nullptr)
+    assert(!file_);
+    if (filename == "stdout")
```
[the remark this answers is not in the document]
patrick-lunarg: Done

framework/decode/vulkan_ascii_consumer_base.cpp
```diff
@@ -39,28 +39,38 @@ VulkanAsciiConsumerBase::~VulkanAsciiConsumerBase()
 
 bool VulkanAsciiConsumerBase::Initialize(const std::string& filename)
```
bradgrantham-lunarg, line 40: I suggest Initialize take a `FILE*`, then the management of the filename can be completely elevated to toascii main.

patrick-lunarg: I agree, but if it's not a major preference for you, I'd rather leave it as-is on this particular branch.

tools/toascii/main.cpp
```diff
@@ -137,18 +94,42 @@ int main(int argc, const char** argv)
...
-        file_processor.ProcessAllFrames();
+        if (output_file)
+        {
+            gfxrecon::decode::VulkanAsciiConsumer ascii_consumer;
+            ascii_consumer.Initialize(output_file);
+            gfxrecon::decode::VulkanDecoder decoder;
+            decoder.AddConsumer(&ascii_consumer);
+            file_processor.AddDecoder(&decoder);
+            file_processor.ProcessAllFrames();
+            if (output_file != stdout)
```
bradgrantham-lunarg, line 124: Add a call to ascii_consumer.Destroy() here, or the closing `}` will only be printed to stdout, and will be lost otherwise because the file is closed before the AsciiConsumer is destructed.  I tested this by making output with a filename and with "stdout" and diffing the results.

tools/toascii/main.cpp
```diff
@@ -58,53 +49,19 @@ static void PrintUsage(const char* exe_name)
     GFXRECON_WRITE_CONSOLE("\nOptional arguments:");
     GFXRECON_WRITE_CONSOLE("  -h\t\t\tPrint usage information and exit (same as --help).");
     GFXRECON_WRITE_CONSOLE("  --version\t\tPrint version information and exit.");
+    GFXRECON_WRITE_CONSOLE("  --output <'stdout' or filepath>\t'stdout' or a working directory relative filepath to "
```
bradgrantham-lunarg, line 52: The filepath doesn't have to be relative, the way you've coded it, it can be absolute as well.  Also this doesn't line up with the other options in the usage output and the output line is very long.  I suggest, like in `replay_options.h` to make this line up with the other options to do something like:
```
    GFXRECON_WRITE_CONSOLE("  --output file\t\t'stdout' or a path to a file to write JSON output");
    GFXRECON_WRITE_CONSOLE("               \t\tto. Default is the input filepath with \"gfxr\" replaced by \"txt\".");
```

bradgrantham-lunarg (approved):

locke-lunarg (approved):

PR #653, opened 2022-02-18
Trimming Ray Tracing(deferred operation)

locke-lunarg:
https://github.com/nvpro-samples/vk_raytracing_tutorial_KHR
This PR could run trimming for these samples. AMD GPU might not support these samples.

`vk_ray_tracing_animation` and `vk_ray_tracing_instances` take very long for capturing because they loads many models. 

`vk_ray_tracing_advanced_compilation` is a complex case. It work well for replay and trimming. It just needs to wait for deferred operations. It uses `std::future` for waiting. It referred to `vk_ray_tracing_advanced_compilation` . But after waiting, it still has low chance to happen the rendering problem during replaying. Also
`[gfxrecon] FATAL - API call vkQueueSubmit returned error value VK_ERROR_DEVICE_LOST that does not match the result from the capture file: VK_SUCCESS.  Replay cannot continue.` or 
`[gfxrecon] FATAL - API call vkWaitForFences returned error value VK_ERROR_DEVICE_LOST that does not match the result from the capture file: VK_SUCCESS.  Replay cannot continue.` sometimes happen randomly during replaying.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -6125,16 +6125,6 @@ void VulkanReplayConsumerBase::OverrideGetDeferredOperationMaxConcurrencyKHR(
 }
 
 VkResult
-VulkanReplayConsumerBase::OverrideDeferredOperationJoinKHR(PFN_vkDeferredOperationJoinKHR  func,
-                                                           VkResult                        original_result,
-                                                           const DeviceInfo*               device_info,
-                                                           const DeferredOperationKHRInfo* deferred_operation_info)
-{
-    // Replay this in VulkanReplayConsumerBase::ProcessDeferredOperation
-    return original_result;
-}
-
-VkResult
 VulkanReplayConsumerBase::OverrideGetDeferredOperationResultKHR(PFN_vkGetDeferredOperationResultKHR func,
                                                                 VkResult                            original_result,
                                                                 const DeviceInfo*                   device_info,
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -41,6 +41,7 @@
 #include <cstdint>
 #include <limits>
 #include <unordered_set>
+#include <future>
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
@@ -5920,6 +5921,10 @@ VkResult VulkanReplayConsumerBase::OverrideCreateRayTracingPipelinesKHR(
     }
     else
     {
+        GFXRECON_LOG_ERROR_ONCE(
+            "The replayed application used vkCreateRayTracingPipelinesKHR, which may require the "
+            "rayTracingPipelineShaderGroupHandleCaptureReplay feature for accurate capture and replay. The replay "
+            "device does not support this feature, so replay may fail.");
         result = device_table->CreateRayTracingPipelinesKHR(device,
                                                             in_deferredOperation,
                                                             in_pipelineCache,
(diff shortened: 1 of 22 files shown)


framework/decode/vulkan_replay_consumer_base.h
```diff
@@ -1013,6 +1028,8 @@ class VulkanReplayConsumerBase : public VulkanConsumer
     // Used to track if any shadow sync objects are active to avoid checking if not needed
     std::unordered_set<VkSemaphore> shadow_semaphores_;
     std::unordered_set<VkFence>     shadow_fences_;
+
+    std::vector<std::future<void>> deferred_operation_joins;
```
bradgrantham-lunarg, line 1032: Should this have `_` at the end?

framework/encode/vulkan_state_info.h
```diff
@@ -68,6 +68,7 @@ struct DescriptorInfo
     VkDescriptorType                              type;
     const void*                                   write_pnext{ nullptr };
     HandleUnwrapMemory                            write_pnext_memory;
+    std::vector<VkAccelerationStructureKHR>       record_write_set_acce_structs;
```
bradgrantham-lunarg, line 71: `acce` -> `accel`

framework/decode/vulkan_replay_consumer_base.h
locke-lunarg: fixed

framework/encode/vulkan_state_info.h
locke-lunarg: fixed

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5732,6 +5733,10 @@ VkResult VulkanReplayConsumerBase::OverrideCreateRayTracingPipelinesKHR(
     }
     else
     {
+        GFXRECON_LOG_ERROR_ONCE(
+            "The replayed application used vkCreateRayTracingPipelinesKHR, which may require the "
+            "rayTracingPipelineShaderGroupHandleCaptureReplay feature for accurate capture and replay. The replay "
+            "device does not support this feature, so replay may fail.");
```
locke-lunarg, line 5739: Add error info for that the device doesn't support rayTracingPipelineShaderGroupHandleCaptureReplay in replaying CreateRayTracingPipelinesKHR.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5741,9 +5746,66 @@ VkResult VulkanReplayConsumerBase::OverrideCreateRayTracingPipelinesKHR(
...
+        ProcessDeferredOperation(device, in_deferredOperation);
+    }
     return result;
 }
 
+void VulkanReplayConsumerBase::ProcessDeferredOperation(VkDevice device, VkDeferredOperationKHR deferred_operation)
+{
+    GFXRECON_LOG_ERROR_ONCE("The replayed application used VkDeferredOperationKHR. It might fail for rendering "
+                            "incorrect or VK_ERROR_DEVICE_LOST. The reason might be the device doesn't support "
+                            "rayTracingPipelineShaderGroupHandleCaptureReplay feature.");
```
locke-lunarg, line 5764: Add error info for that the device doesn't support rayTracingPipelineShaderGroupHandleCaptureReplay if replaying use VkDeferredOperationKHR.

locke-lunarg:
Add issue
https://github.com/LunarG/gfxreconstruct/issues/659

locke-lunarg:
Added a commit to `Fix capturing DeferredOperationJoinKHR crash`

locke-lunarg:
Added a new commit `ProcessDeferredOperation in OverrideCreateRayTracingPipelinesKHR` to fix capturing failure for DOOMEternal(ray tracing). But it still fails during replaying. See https://github.com/LunarG/gfxreconstruct/issues/662

locke-lunarg:
Tested with AMD driver: 22.5.2. It didn't happen `VK_ERROR_DEVICE_LOST` again, but it failed in index 245149.
` [gfxrecon] FATAL - API call vkAllocateMemory returned error value VK_ERROR_OUT_OF_HOST_MEMORY that does not match the result from the capture file: VK_SUCCESS.  Replay cannot continue.`

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -1137,24 +1170,27 @@ VulkanCaptureManager::OverrideCreateRayTracingPipelinesKHR(VkDevice
...
     }
 
     if (((result == VK_SUCCESS) || (result == VK_OPERATION_DEFERRED_KHR) ||
          (result == VK_OPERATION_NOT_DEFERRED_KHR)) &&
         (pPipelines != nullptr))
     {
+        if (deferredOperation != VK_NULL_HANDLE &&
+            ((result == VK_OPERATION_DEFERRED_KHR) || (result == VK_OPERATION_NOT_DEFERRED_KHR)))
+        {
+            ProcessDeferredOperation(device, deferredOperation);
```
mizhen, line 1183: This changed target title behavior because the deferred operation join call come from GFXR, not original application.

PR #655, opened 2022-02-18
Add API call index and plumb ApiCallInfo down to consumers

patrick-lunarg:
This PR adds an `ApiCallInfo` member to `FileProcessor` that can be updated and passed to api call handlers.

### DIFF
--- a/framework/decode/file_processor.cpp
+++ b/framework/decode/file_processor.cpp
@@ -37,7 +37,7 @@ GFXRECON_BEGIN_NAMESPACE(decode)
 
 FileProcessor::FileProcessor() :
     file_header_{}, file_descriptor_(nullptr), current_frame_number_(0), bytes_read_(0),
-    error_state_(kErrorInvalidFileDescriptor), annotation_handler_(nullptr), compressor_(nullptr)
+    error_state_(kErrorInvalidFileDescriptor), annotation_handler_(nullptr), compressor_(nullptr), call_index_(0)
 {}
 
 FileProcessor::~FileProcessor()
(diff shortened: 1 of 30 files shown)


framework/decode/file_processor.cpp
```diff
@@ -443,6 +443,7 @@ bool FileProcessor::ProcessFunctionCall(const format::BlockHeader& block_header,
                 {
                     DecodeAllocator::Begin();
                     decoder->DecodeFunctionCall(call_id, call_info, parameter_buffer_.data(), parameter_buffer_size);
+                    ++call_info.index;
```
bradgrantham-lunarg, line 446: I'd prefer this looked like the source PR, with a member variable in the FileProcessor and then just make a local stack ApiCallInfo, assign it the index, and pass that one down.

bradgrantham-lunarg (changes requested):

framework/decode/file_processor.cpp
bradgrantham-lunarg: I realize I didn't really justify that, so here's my thinking.  Right now an ApiCallInfo contains only `thread_id` and `index`.  And if the only thing being used in callees is `index`, well, that seems safe enough to just have a prototype around all the time in `FileProcessor` and have it hold the current index that's incremented every time.

But it has a kind of code smell to me.  There's no guarantee more stuff won't go into `ApiCallInfo` later.  The `ApiCallInfo` in my mind is a kind of collection point for data needed by the decoders and by the consumers the decoders call.  I like the idea of the `ApiCallInfo` being constructed on the stack each time to pass down, meaning no state will secretly leak from one invocation to the next.  If the downchain functions have asserts on unexpected default values on ApiCallInfo, we'll catch errors in those members not being updated from one API call to the next.  But if there's just one hanging around in `FileProcessor`, then the data from the previous API call is just sitting there and if we don't write code to update all the members there may be an unexpected previous value.

So, there's no bug in your PR.  However, I think if we had `api_call_index_` in `FileProcessor` and put that into a newly-constructed `ApiCallInfo` before calling `decoder->DecodeFunctionCall`, we may prevent a new bug later.

patrick-lunarg: That makes sense...I'll make that change

patrick-lunarg: Done

bradgrantham-lunarg (approved):

locke-lunarg (approved):

PR #656, opened 2022-02-21
Make IsArgumentSet() Behave Like IsOptionSet()

patrick-lunarg:
This PR makes `IsArgumentSet()` behave like `IsOptionSet()`.

`IsOptionSet()` returns `true` when a given option has been set by the user on the cmd line and returns `false` when a given option has not be set by the user. 

`IsArgumentSet()` currenlty returns `true` for any valid argument, regardless of whether or not the argument is set.

This PR adds a member variable called `arguments_present_` and gives it the same treatment as the existing `options_present_` member variable.

### DIFF
--- a/framework/util/argument_parser.cpp
+++ b/framework/util/argument_parser.cpp
@@ -185,6 +185,7 @@ void ArgumentParser::Init(std::vector<std::string> command_line_args,
             argument_index++;
         }
     }
+    arguments_present_.resize(argument_index);
     argument_values_.resize(argument_index);
 
     for (size_t cur_arg = 0; cur_arg < command_line_args.size(); ++cur_arg)
@@ -245,7 +246,8 @@ void ArgumentParser::Init(std::vector<std::string> command_line_args,
                             }
                             argument_values_[cur_argument.second] = argument_value;
                         }
-                        is_argument = true;
+                        arguments_present_[cur_argument.second] = true;
+                        is_argument                             = true;
                         break;
                     }
                 }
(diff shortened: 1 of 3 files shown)


framework/util/argument_parser.cpp
```diff
@@ -245,7 +246,8 @@ void ArgumentParser::Init(std::vector<std::string> command_line_args,
                             }
                             argument_values_[cur_argument.second] = argument_value;
                         }
-                        is_argument = true;
+                        arguments_present_[cur_argument.second] = true;
```
bradgrantham-lunarg, line 249: Should this be "false" for when the argument is invalid, like checked above on line 266?

bradgrantham-lunarg (changes requested):

framework/util/argument_parser.cpp
bradgrantham-lunarg: I think this might need to be bumped up one line so it's inside the `else`

patrick-lunarg: It seems to me that since we can check for validity with `IsInvalid()`, `IsArgumentSet()` should indicate whether or not the user passed the given argument regardless of validity.

bradgrantham-lunarg: I think I'd argue the opposite.  Was `ArgumentSet` (i.e. is `_arguments_present_`) if the option occurred at the end of the command line and there was no filename for `--output`?

In any case, It doesn't have to be in `arguments_present_` if `is_invalid_` since all the tools exit if `IsInvalid()` before checking the arguments and `GetInvalidArgumentOrOptions()` would return the invalid arguments.  (No tool calls `GetInvalidArgumentOrOptions `currently, they just print Usage and exit, but I think let's leave that for another PR!)

bradgrantham-lunarg: It doesn't look like you made the change I requested?

patrick-lunarg: My bad...it's pushed now

bradgrantham-lunarg (approved):

locke-lunarg (approved):

PR #666, opened 2022-03-15
temporarily disable toascii

### LINKED ISSUE #667: current documentation discusses gfxrecon-toascii
Current **GFXReconstruct** documentation includes references to `gfxrecon-toascii`:

https://vulkan.lunarg.com/doc/view/1.3.204.1/windows/capture_tools.html#user-content-to-ascii

toascii

`gfxrecon-toascii` is not currently shipped in any SDK, so the documentation is at best misleading.  We know at the moment `gfxrecon-toascii` is not always valid JSON and some fields are incorrect. This should be temporarily removed from our documentation.

Note that this may be a transitory issue (as gfxrecon-toascii should enter SDKs "any day now", once the output format that we intend to support indefinitely is formalized), and it may not be worth the effort to make special SDK-motivated changes to the documentation process.

bradgrantham-lunarg:
Because of identified issues with the current `gfxrecon-toascii` implementation, `gfxrecon-toascii `is now temporarily removed from the build as we discussed in December.  Until resolved, build `gfxrecon-toascii`, provide the option `BUILD_TOASCII` with a non-false value during `cmake`, e.g.
```
cmake -Bbuild -DBUILD_TOASCII=true
```

Fixes #667

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -1,8 +1,12 @@
 add_subdirectory(replay)
-add_subdirectory(toascii)
 add_subdirectory(compress)
 add_subdirectory(info)
 add_subdirectory(extract)
 add_subdirectory(optimize)
 add_subdirectory(capture)
 add_subdirectory(gfxrecon)
+
+# TOASCII build is currently disabled pending resolution of some issues
+if(BUILD_TOASCII)
+add_subdirectory(toascii)
+endif()
--- a/USAGE_desktop.md
+++ b/USAGE_desktop.md
@@ -547,27 +547,6 @@ Optional arguments:
   --version             Print version information and exit.
 ```
 
-### To ASCII
-
-The `gfxrecon-toascii.exe` converts GFXReconstruct capture files to text.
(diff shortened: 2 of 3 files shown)


locke-lunarg (approved):

tools/CMakeLists.txt
```diff
@@ -1,8 +1,12 @@
...
-add_subdirectory(toascii)
 add_subdirectory(compress)
 add_subdirectory(info)
 add_subdirectory(extract)
 add_subdirectory(optimize)
 add_subdirectory(capture)
 add_subdirectory(gfxrecon)
+
+# TOASCII build is currently disabled pending resolution of some issues
+if(BUILD_TOASCII)
```
lunarpapillo, line 10: I'm mildly concerned that this will impact developers.  I agree that this is clean, but the SDK builds don't currently require this; so it's debatable whether a change that helps no one and inconveniences some people is worth it...

lunarpapillo (approved):
This will more than address the issue with **gfxreconstruct** documentation.  I'm pleased it looks like it wasn't much effort, as I would have recommended against a change that was harder.

locke-lunarg (approved):

Issue #671, opened 2022-03-20
Option to capture limited number of frames on key trigger


perlfu:
It's not possible to capture a consistent number of frames using the trigger key.
Ideally I'd like to be able to capture single frames for later analysis.

bradgrantham-lunarg:
Fixed by #672

PR #672, opened 2022-03-20
Add option to limit hotkey captures to a specific frame count

perlfu:
Add CAPTURE_TRIGGER_FRAMES option to specify a number a limit
on the number of frames that are captured when hotkey triggering
is used.  This automatically stops the current capture when
the specified number of frames is reached.

Implements LunarG/gfxreconstruct#671


framework/encode/capture_settings.cpp
```diff
@@ -697,5 +708,20 @@ std::string CaptureSettings::ParseTrimKeyString(const std::string& value_string)
     return trim_key;
 }
 
+uint32_t CaptureSettings::ParseTrimKeyFramesString(const std::string& value_string)
+{
+    uint32_t frames = 0;
+    int      value  = std::stoi(value_string);
```
bradgrantham-lunarg, line 714: `std::stoi` will throw an exception if the string contains something unparseable (`invalid_argument` I think).

You can test this in your branch with `GFXRECON_CAPTURE_TRIGGER_FRAMES=a1` or some such.

CaptureSettings::ParseTrimRangeString validates that frame numbers contain only digits; would you please add code similar to that to your new function `ParseTrimKeyFramesString`?

(`GFXRECON_CAPTURE_TRIGGER_FRAMES=1a` doesn't throw, but that probably should also print an error.)

bradgrantham-lunarg (changes requested):

framework/encode/capture_settings.cpp
perlfu: Thanks!
I had missed that `ParseTrimKeyFramesString` was validating the string was digits.

bradgrantham-lunarg:
This is great, thanks!

In general, using `static_cast ` is more consistent with the existing codebase.

However, that's a tiny nit and we'll merge this PR in the name of expediency and then we'll add that cast in a separate commit.

Thanks again!

Issue #611, opened 2021-11-19
handles are prefixed with "0x0x" on linux in toascii


ghost:
In this toascii output, handles are prepended with "0x0x" instead of just "0x".
This issue is observed on Linux only.
The output should have a single "0x".

```
"[50]vkBindImageMemory": {
    "return": "VK_SUCCESS",
    "device": "0x0x4",
    "image": "0x0x15",
    "memory": "0x0x16",
    "memoryOffset": Unhandled VkFormatFeatureFlagBits2KHR
}
```

bradgrantham-lunarg:
Fixed by #680

PR #680, opened 2022-03-25
gfxrecon-toascii: Remove extra 0x when printing handles and pointers

davidlunarg:
Fix for issue https://github.com/LunarG/gfxreconstruct/issues/611

### DIFF
--- a/framework/decode/custom_vulkan_ascii_consumer.h
+++ b/framework/decode/custom_vulkan_ascii_consumer.h
@@ -50,7 +50,7 @@ inline std::string HandleIdToString(format::HandleId handleId)
     std::stringstream strStrm;
     if (handleId)
     {
-        strStrm << "\"0x" << reinterpret_cast<const void*>(handleId) << "\"";
+        strStrm << "\"0x" << std::hex << handleId << "\"";
     }
     else
     {
--- a/framework/util/to_string.h
+++ b/framework/util/to_string.h
@@ -65,11 +65,18 @@ inline std::string ToString(uint32_t      apiFlags,
 }
 
 template <typename PtrType>
+inline std::string PtrToString(PtrType* ptr)
+{
+    std::stringstream strStrm;
+    strStrm << "0x" << std::hex << reinterpret_cast<uintptr_t>(ptr);
+    return strStrm.str();
+}
+
+template <typename PtrType>
 inline std::string PtrToString(PtrType ptr)
 {
     std::stringstream strStrm;
-    const void*       val = reinterpret_cast<const void*>(ptr);
-    strStrm << "0x" << std::hex << reinterpret_cast<uint64_t>(val);
+    strStrm << "0x" << std::hex << ptr;
     return strStrm.str();
 }
 
--- a/framework/util/to_string.h
+++ b/framework/util/to_string.h
@@ -68,7 +68,8 @@ template <typename PtrType>
 inline std::string PtrToString(PtrType ptr)
 {
     std::stringstream strStrm;
-    strStrm << "0x" << reinterpret_cast<const void*>(ptr);
+    const void*       val = reinterpret_cast<const void*>(ptr);
+    strStrm << "0x" << std::hex << reinterpret_cast<uint64_t>(val);
     return strStrm.str();
 }
 
--- a/framework/encode/handle_unwrap_memory.h
+++ b/framework/encode/handle_unwrap_memory.h
@@ -25,6 +25,7 @@
 
 #include "util/defines.h"
 
+#include <iterator>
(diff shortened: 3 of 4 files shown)


framework/decode/custom_vulkan_ascii_consumer.h
```diff
@@ -50,7 +50,8 @@ inline std::string HandleIdToString(format::HandleId handleId)
     std::stringstream strStrm;
     if (handleId)
     {
-        strStrm << "\"0x" << reinterpret_cast<const void*>(handleId) << "\"";
+        const void* val = reinterpret_cast<const void*>(handleId);
```
bradgrantham-lunarg, line 53: HandleId is typedef'd uint64_t, so will this `reinterpret_cast` truncate the handle on 32-bit?  Do you think it would work for us to just print the handleID? Like
`strStrm << "\"0x" << std::hex << handleId << "\"";`

framework/util/to_string.h
```diff
@@ -68,7 +68,8 @@ template <typename PtrType>
 inline std::string PtrToString(PtrType ptr)
 {
     std::stringstream strStrm;
-    strStrm << "0x" << reinterpret_cast<const void*>(ptr);
+    const void*       val = reinterpret_cast<const void*>(ptr);
```
bradgrantham-lunarg, line 71: I think on 32-bit `val` may end up being sign-extended, depending on the compiler.  https://godbolt.org/z/v6qro1a3z
Could we `reinterpret_cast` `ptr` directly to `uintptr_t` and then `<<` that?  Like
`strStrm << "0x" << std::hex << reinterpret_cast<uintptr_t>(ptr);`

bradgrantham-lunarg (changes requested):

framework/decode/custom_vulkan_ascii_consumer.h
davidlunarg: It works as well as the old code did on 32 bit. :-)
More seriously, do we support 32 bit, or have plans to? If so, this would be an issue.
I tried just printing the handleId as you suggested. The gcc compiler and VS compiler produce different results - one preceeds the output with a "0x", the other doesn't. I'm not sure why they produce different results. I tried to come up with the simplest code that produces the same output, this is what I found worked after lots of experiments.

framework/util/to_string.h
davidlunarg: I don't think sign extension would be an issue. None of the inputs, outputs, or intermediate values are signed. A conversion from a pointer to an unsigned 64 bit int shouldn't do sign extension.

framework/decode/custom_vulkan_ascii_consumer.h
bradgrantham-lunarg: I dunno; maybe we shouldn't, but right now we do support 32-bit as far as I can tell.  :shrug:  It may be necessary for capturing some older apps that run in a translation layer.

The code
```
    uint64_t handleId = 0x50;
    strStrm << "\"0x" << std::hex << handleId << "\"";
```
results in `0x50` for me in Compiler Explorer for MSVC and gcc and clang, 32-bit and 64-bit.

Do you mean the pointer version may or may not print an "0x"?  For that one I suggested a cast to uintptr_t.

framework/util/to_string.h
bradgrantham-lunarg: The godbolt link shows that it does.  I haven't looked into it; I imagine maybe whether or not sign-extension happens may be undefined or something.

bradgrantham-lunarg: Oh!  I had a typo in my Compiler Explorer example.  https://godbolt.org/z/7hq55vn56  But the result is the same.

framework/decode/custom_vulkan_ascii_consumer.h
davidlunarg: When I tried this in the PtrToString function in the to_string.h change below, I get an extra '0x' - the original bug.  I couldn't get the gcc compiler not to prepend an 0x when outputting a ptr. So given the temp var and casting I had to do in the PtrToString, I decided to have HandleIdToString be the same. I can make them different if you want, but I would prefer they be similar.

davidlunarg:
Suggested changes have been made.

framework/util/to_string.h
bradgrantham-lunarg: Please also change this instance of these two lines to `strStrm << "0x" << std::hex << reinterpret_cast<uintptr_t>(ptr);`

bradgrantham-lunarg (changes requested):

bradgrantham-lunarg (approved):

PR #697, opened 2022-05-16
Framerange support

panos-lunarg:
@bradgrantham-lunarg @tomped01

I created a new PR for the framerange support PR submitted by @tomped01 
This PR incorporates all commits by Tom, the changes from Brad and a couple of fixes by me that fix a few issues I discovered:

1. There was no way to distinguish a measurement for frameranges that began from frame 1 and when the `--mfr` parameter was not specified.
2. Traces with no frames (i.e. vulkaninfo) were reported as traces with 1 frame, which was confusing.
3. Brad's changes were causing the report of the total frames replayed when the `--mfr` option was not specified to be off-by-1

The PR by @bradgrantham-lunarg that I tried to piggyback was causing some strange conflicts when trying to merge back to dev so I created a new one.
If you all agree with the changes I propose to submit this PR.

### DIFF
--- a/framework/application/application.cpp
+++ b/framework/application/application.cpp
@@ -132,30 +132,33 @@ void Application::Run()
         // Only process the next frame if a quit event was not processed or not paused.
         if (running_ && !paused_)
         {
+            // Add one to match "trim frame range semantic"
+            uint32_t frame_number = file_processor_->GetCurrentFrameNumber() + 1;
 
             if (fps_info_ != nullptr)
             {
-                if (fps_info_->ShouldQuit(file_processor_->GetCurrentFrameNumber()))
+                if (fps_info_->ShouldQuit(frame_number))
                 {
                     running_ = false;
                     break;
                 }
 
-                if (fps_info_->ShouldWaitIdleBeforeFrame(file_processor_->GetCurrentFrameNumber()))
+                if (fps_info_->ShouldWaitIdleBeforeFrame(frame_number))
                 {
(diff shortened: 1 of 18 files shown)


bradgrantham-lunarg:
@tomped01 - I couldn't find the time to finish this feature up.  It looks like @panos-lunarg has done a good job of following through on the details of which frames are rendered and also corner cases.  Panos should be able to give this a little more timely attention.  Could you verify that this PR does what you expect?  It will include your commits in your name, so you'll get acknowledgement for the initial feature; will that be acceptable?

bradgrantham-lunarg (approved):

tomped01:
It seems to work as expected with one exception, the --qamr argument has been removed (maybe during merge conflict resolution?)

If you look here: https://github.com/LunarG/gfxreconstruct/pull/553/conflicts

--quit-after-measurment-range is as it should be in the framerange_support block, but here: https://github.com/LunarG/gfxreconstruct/pull/697/files#diff-69084bf7fc48983c236d81343dea587d162f905c4b5d6376f76ce8bd6d5a9f8b

It has been partially removed and partially merged with the --onhb argument string.

It still works if you specify the full --quit-after-measurment-range option and not the shortened one, however it brings back the issue that lunarpapillo highlighted here: https://github.com/LunarG/gfxreconstruct/pull/553 (comment from June 19. 2021), which is that --qamr cannot be used (since the kOption parsing breaks silently at runtime when the option string is wrong)!

Other than that though, it looks good :)

PR #701, opened 2022-05-24
Virtual swapchain

locke-lunarg:
Added an option `--virtual-swapchain` to enable it.

Testing:

Using vkCube for those tests. It could remove `Replace virtual images in QueuePresent` commit to ensure it really use virtual swapchain because the screen should become black. It only tested in Windows, not Android for now.

- Force a different min number of real swapchain images in replay, maybe +1 and verify number of images is different than capture (pCreateInfo->minImageCount)
  - Adding `modified_create_info.minImageCount = 2` in `VulkanVirtualSwapchain::CreateSwapchainKHR` for replay. Setting `desiredNumOfSwapchainImages = 3` in vkCube
  or
    Adding `modified_create_info.minImageCount = 4` in `VulkanVirtualSwapchain::CreateSwapchainKHR` for replay. Setting `desiredNumOfSwapchainImages = 3` in vkCube
  or
    Adding `modified_create_info.minImageCount = 3` in `VulkanVirtualSwapchain::CreateSwapchainKHR` for replay. Setting `desiredNumOfSwapchainImages = 2` in vkCube to test.

- In replay, hand edit AcquireNextImageKHR so the first call actually Acquires two images (discarding the first one) to deliberately get out of order
  - Adding more `func(device, swapchain, timeout, semaphore, fence, image_index)` in `VulkanVirtualSwapchain::AcquireNextImageKHR` to test it.

- Force a different swap mode on capture, like MAILBOX, make sure captures are still replayed correctly
  - Adding `modified_create_info.presentMode = VK_PRESENT_MODE_MAILBOX_KHR;` or other present mode in `VulkanVirtualSwapchain::CreateSwapchainKHR` to test it.

- Capture on Android R(11), replay on S(12) or the other way around - verify swapchain count has changed
  -  Android 10 can replay Android 12 capture file with `--remove-unsupported` to remove huge extensions that Android 12 added.

      Android12 can’t replay Android 10 capture file. The failure reason isn’t related to swapchain. It might not be able to run tests for virtual swapchain.
      Log:
      Skipping memory fill for VkDeviceMemory object (ID = 25) that is not mapped
      Skipping memory fill for VkDeviceMemory object (ID = 33) that is not mapped
      Skipping memory fill for VkDeviceMemory object (ID = 31) that is not mapped
      Skipping memory fill for VkDeviceMemory object (ID = 29) that is not mapped


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4424,6 +4452,65 @@ void VulkanReplayConsumerBase::OverrideGetImageSubresourceLayout(
...
+    PFN_vkCreateRenderPass2                                      func,
+    VkResult                                                     original_result,
+    const DeviceInfo*                                            device_info,
+    const StructPointerDecoder<Decoded_VkRenderPassCreateInfo2>* pCreateInfo,
+    const StructPointerDecoder<Decoded_VkAllocationCallbacks>*   pAllocator,
+    HandlePointerDecoder<VkRenderPass>*                          pRenderPass)
+{
+    GFXRECON_UNREFERENCED_PARAMETER(original_result);
+
+    return func(device_info->handle,
```
bradgrantham-lunarg, line 4482: Does this need to call `swapchain_->CreateRenderPass2`?

locke-lunarg: fixed

locke-lunarg:
Linux AMD device passed test with vkcube.

locke-lunarg:
Added two commits to fix VkQueue and InvalidImageLayout issuses.

Dustin's set ImageLayout of virtual swapchain image was VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL. But it caused some InvalidImageLayout issuses by VVL. Set VK_IMAGE_LAYOUT_PRESENT_SRC_KHR could avoid it. But I'm not sure if it has any down side. At least, VVL didn't complaint it.

per-mathisen-arm:
No, sorry, the comment was meant for the whole pull request in general, not that particular commit.

bradgrantham-lunarg:
> If this is not VK_SHARING_MODE_EXCLUSIVE then the following line where queueFamilyIndexCount is set to zero would be invalid. Perhaps worth checking this assumption?

@per-mathisen-arm , do you mean for the line 
```
image_create_info.sharingMode           = swapchain_info->image_sharing_mode;
```
if `image_sharing_mode` is `CONCURRENT` then the `queueFamilyIndexCount` and `pQueueFamilyIndices` should be copied from the original swapchain?

per-mathisen-arm:
> > If this is not VK_SHARING_MODE_EXCLUSIVE then the following line where queueFamilyIndexCount is set to zero would be invalid. Perhaps worth checking this assumption?
> 
> @per-mathisen-arm , do you mean for the line
> 
> ```
> image_create_info.sharingMode           = swapchain_info->image_sharing_mode;
> ```
> 
> if `image_sharing_mode` is `CONCURRENT` then the `queueFamilyIndexCount` and `pQueueFamilyIndices` should be copied from the original swapchain?

Yes, that would be my understanding.

bradgrantham-lunarg:
> > if `image_sharing_mode` is `CONCURRENT` then the `queueFamilyIndexCount` and `pQueueFamilyIndices` should be copied from the original swapchain?
> 
> Yes, that would be my understanding.

locke-lunarg:
fixed for adding queue family indices for swapchain images.

locke-lunarg:
Moved VulkanReplayConsumerBase::ProcessSetSwapchainImageStateCommand to virtual swapchain.

Only default swapchain does what `VulkanReplayConsumerBase::ProcessSetSwapchainImageStateCommand` does.

Virtual swapchain does nothing about it.

locke-lunarg:
Set virtual swapchain as default. option `--use-cpatured-swapchain-indices` could disable virtual swapchain

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -0,0 +1,789 @@
...
+
+        VkSubmitInfo submit_info         = { VK_STRUCTURE_TYPE_SUBMIT_INFO };
+        submit_info.waitSemaphoreCount   = present_info->waitSemaphoreCount;
+        submit_info.pWaitSemaphores      = present_info->pWaitSemaphores;
+        submit_info.pWaitDstStageMask    = &wait_stage;
+        submit_info.commandBufferCount   = 1;
+        submit_info.pCommandBuffers      = &command_buffer;
+        submit_info.signalSemaphoreCount = 1;
+        submit_info.pSignalSemaphores    = &semaphore;
+
```
per-mathisen-arm, line 586: I think there is a cleaner way to do this. Instead of overloading commands before this to change the used layouts, you could use kImageMemoryBarriers here in the vkCmdPipelineBarrier calls to change the layouts.

android/scripts/gfxrecon.py
```diff
@@ -79,6 +79,7 @@ def CreateReplayParser():
     parser.add_argument('--sync', action='store_true', default=False, help='Synchronize after each queue submission with vkQueueWaitIdle (forwarded to replay tool)')
     parser.add_argument('--remove-unsupported', action='store_true', default=False, help='Remove unsupported extensions and features from instance and device creation parameters (forwarded to replay tool)')
     parser.add_argument('--onhb', '--omit-null-hardware-buffers', action='store_true', default=False, help='Omit Vulkan calls that would pass a NULL AHardwareBuffer* (forwarded to replay tool)')
+    parser.add_argument('--virtual-swapchain', action='store_true', default=False, help='Enable virtual swapchain.')
```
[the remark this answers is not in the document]
locke-lunarg: Fixed . Thanks.

android/scripts/gfxrecon.py
```diff
@@ -134,6 +135,9 @@ def MakeExtrasString(args):
     if args.onhb:
         arg_list.append('--onhb')
 
+    if args.virtual_swapchain:
+        arg_list.append('--virtual-swapchain')
```
bradgrantham-lunarg, line 139: I think this needs to be updated to `--use-captured-swapchain-indices`.

android/scripts/gfxrecon.py
```diff
@@ -79,6 +79,7 @@ def CreateReplayParser():
     parser.add_argument('--sync', action='store_true', default=False, help='Synchronize after each queue submission with vkQueueWaitIdle (forwarded to replay tool)')
     parser.add_argument('--remove-unsupported', action='store_true', default=False, help='Remove unsupported extensions and features from instance and device creation parameters (forwarded to replay tool)')
     parser.add_argument('--onhb', '--omit-null-hardware-buffers', action='store_true', default=False, help='Omit Vulkan calls that would pass a NULL AHardwareBuffer* (forwarded to replay tool)')
+    parser.add_argument('--use-captured-swapchain-indices', action='store_true', default=False, help='Enable captured swapchain indices.')
```
bradgrantham-lunarg, line 82: Could this be something more descriptive of what's actually happening, like "Use the captured indices for swapchain images"?

bradgrantham-lunarg (changes requested):

android/scripts/gfxrecon.py
locke-lunarg: fixed. Thanks.

android/scripts/gfxrecon.py
locke-lunarg: Fixed. Thanks.

framework/decode/vulkan_virtual_swapchain.cpp
locke-lunarg: Good idea! Fixed. Added a new commit `Combine two CmdPipelineBarrier`. Thanks,

bradgrantham-lunarg (approved):

PR #702, opened 2022-05-24
Fix for multiple android surfaces workaround

panos-lunarg:
Commit 4ad06dc added a way to workaround the lack of support for
multiple Android surfaces, by giving the ability to choose which one to
create and render on, when a trace creates multiple surfaces/swapchains.

This commit adds a couple of checks for VK_NULL_HANDLE passed around as
surface handles that were missing causing assertion failures.

### DIFF
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -360,6 +360,7 @@ struct SurfaceKHRInfo : public VulkanObjectInfo<VkSurfaceKHR>
 {
     Window*                              window{ nullptr };
     std::unordered_map<uint32_t, size_t> array_counts;
+    bool                                 surface_creation_skipped{ false };
 
     std::unordered_map<VkPhysicalDevice, VkSurfaceCapabilitiesKHR> surface_capabilities;
 };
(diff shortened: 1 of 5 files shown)


panos-lunarg:
I am just realizing that in some cases surface **can** be `VK_NULL_HANDLE`
These cases are:

1. `vkGetPhysicalDeviceSurfaceFormats(2)KHR` and `vkGetPhysicalDeviceSurfacePresentModes(2)KHR` because of the [VK_GOOGLE_surfaceless_query](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VK_GOOGLE_surfaceless_query)
2. `vkDestroySurfaceKHR`

I guess I'll need to add a check if the extension is active

panos-lunarg:
Decided to take a different approach. I gathered that it would be impossible to determine if we need to skip a function call because a surface is null because we skipped its creation or whether the application passed it when that google extension is active.
Please let me know if the change is too disruptive

bradgrantham-lunarg (approved):

locke-lunarg (approved):

bradgrantham-lunarg (approved):

PR #705, opened 2022-05-31
WIP: Track protected queues creation when capturing and replaying trimmed

panos-lunarg:
The problem is that in a couple of places we are calling `vkGetDeviceQueue` while we should be calling `vkGetDeviceQueue2`
This is because ANGLE creates the queues with `VkDeviceQueueCreateInfo->flags == VK_DEVICE_QUEUE_CREATE_PROTECTED_BIT`
In that case the spec says that `vkGetDeviceQueue2` should be called instead of `vkGetDeviceQueue`.
We always call `vkGetDeviceQueue` which is returning null hence the segfault (Haven't looked into why this doesn't happen in the untrimmed version).
This happens both when recording and when playing the trimmed traces.
This patch is still in progress as it doesn't properly fix the issue when replaying the trimmed trace.
Basically I'm trying to store the flags passed at device creation in the `struct DeviceInfo` and access later but unfortunately this doesn't seem to work as the `DeviceInfo` struct allocated when the device is created is different when `vkGetDeviceQueue` is called so it holds different information.


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -969,8 +970,19 @@ void VulkanReplayConsumerBase::ProcessSetSwapchainImageStateQueueSubmit(
     pool_create_info.flags                   = VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT;
     pool_create_info.queueFamilyIndex        = queue_family_index;
 
-    // TODO: Improved queue selection?
-    table->GetDeviceQueue(device, queue_family_index, 0, &queue);
+    const auto queue_family_flags = device_info->queue_family_creation_flags.find(queue_family_index);
+    assert(queue_family_flags != device_info->queue_family_creation_flags.end());
+    if (queue_family_flags->second)
```
bradgrantham-lunarg, line 975: Since this is a bitmask, can this be more explicit, like `if (queue_family_flags->second != 0)`?

framework/decode/vulkan_resource_initializer.cpp
```diff
@@ -316,7 +318,21 @@ VkResult VulkanResourceInitializer::GetCommandExecObjects(uint32_t         queue
 
             if (result == VK_SUCCESS)
             {
-                device_table_->GetDeviceQueue(device_, queue_family_index, 0, queue);
+                const auto queue_family_flags = device_info_->queue_family_creation_flags.find(queue_family_index);
+                assert(queue_family_flags != device_info_->queue_family_creation_flags.end());
+                if (queue_family_flags->second)
```
bradgrantham-lunarg, line 323: I have the same thought here about an explicit comparison with 0.

bradgrantham-lunarg (changes requested):
Looks good to me with the two nits about an explicit comparison

framework/decode/vulkan_replay_consumer_base.cpp
panos-lunarg: Sure

PR #707, opened 2022-06-01
Provide a GetPhysicalDeviceProcAddr function [Fixes #688]

### LINKED ISSUE #688: vkGetInstanceProcAddr with unknown name crashes on linux
When capturing on linux, and if my app calls `vkGetInstanceProcAddr` with an unknown name, I get

```
/home/olv/projects/Vulkan-Loader/loader/unknown_function_handling.c:315: loader_phys_dev_ext_gpa_term_no_check: Assertion `NULL != inst' failed.
```

gdb says

```
(gdb) bt
#4  0x00007ffff7eeebd9 in loader_phys_dev_ext_gpa_term_no_check (inst=0x0, 
    funcName=0x5555555607d8 "vkFoo")
    at /home/olv/projects/Vulkan-Loader/loader/unknown_function_handling.c:315
#5  0x00007ffff7edc0c2 in loader_gpdpa_instance_terminator (inst=0x55555568faf0, 
    pName=0x5555555607d8 "vkFoo") at /home/olv/projects/Vulkan-Loader/loader/loader.c:3953
#6  0x00007ffff7eee63f in loader_check_layer_list_for_phys_dev_ext_address (inst=0x5555555712b0, 
    funcName=0x5555555607d8 "vkFoo")
    at /home/olv/projects/Vulkan-Loader/loader/unknown_function_handling.c:187
#7  0x00007ffff7eee73e in loader_phys_dev_ext_gpa_impl (inst=0x5555555712b0, 
    funcName=0x5555555607d8 "vkFoo", is_tramp=true)
    at /home/olv/projects/Vulkan-Loader/loader/unknown_function_handling.c:222
#8  0x00007ffff7eeeb6e in loader_phys_dev_ext_gpa_tramp (inst=0x5555555712b0, 
    funcName=0x5555555607d8 "vkFoo")
    at /home/olv/projects/Vulkan-Loader/loader/unknown_function_handling.c:307
#9  0x00007ffff7ebb5e6 in trampoline_get_proc_addr (inst=0x5555555712b0, funcName=0x5555555607d8 "vkFoo")
    at /home/olv/projects/Vulkan-Loader/loader/gpa_helper.c:262
#10 0x00007ffff7ee6527 in vkGetInstanceProcAddr (instance=0x5555555712b0, pName=0x5555555607d8 "vkFoo")
    at /home/olv/projects/Vulkan-Loader/loader/trampoline.c:106
```

I noticed this because [volk](https://github.com/zeux/volk.git) can call `vkGetInstanceProcAddr` with unknown names.

andrew-lunarg:
Provides a GetPhysicalDeviceProcAddr function which unwraps the
instance and passes it through to the next layer.

Fixes #688

Hoping this is ready to go in and looking for feedback on that.

The Loader PR that this previously depended on has been re-implemented and merged to Loader main.

### DIFF
--- a/layer/trace_layer.cpp
+++ b/layer/trace_layer.cpp
@@ -98,7 +98,7 @@ static VkInstance get_instance_handle(const void* handle)
     return (entry != instance_handles.end()) ? entry->second : VK_NULL_HANDLE;
 }
 
-// The vk_layerGetPhysicalDeviceProcAddr of the next layer in the chain.
+// The GetPhysicalDeviceProcAddr of the next layer in the chain.
 // Retrieved during instance creation and forwarded to by this layer's
 // GetPhysicalDeviceProcAddr() after unwrapping its VkInstance parameter.
 static std::mutex                                                    gpdpa_lock;
@@ -426,7 +426,7 @@ extern "C"
         return VK_SUCCESS;
     }
 
-    // The following three functions are not directly invoked by the desktop loader, which instead uses the function
+    // The following two functions are not directly invoked by the desktop loader, which instead uses the function
     // pointers returned by the negotiate function.
     VK_LAYER_EXPORT VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL vkGetInstanceProcAddr(VkInstance  instance,
                                                                                    const char* pName)
@@ -437,12 +437,6 @@ extern "C"
     VK_LAYER_EXPORT VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL vkGetDeviceProcAddr(VkDevice device, const char* pName)
     {
         return gfxrecon::GetDeviceProcAddr(device, pName);
-    }
-
-    VK_LAYER_EXPORT VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL vk_layerGetPhysicalDeviceProcAddr(VkInstance  instance,
-                                                                                               const char* pName)
-    {
-        return gfxrecon::GetPhysicalDeviceProcAddr(instance, pName);
     }
 
     // The following four functions are not invoked by the desktop loader, which retrieves the layer specific properties
--- a/layer/trace_layer.cpp
+++ b/layer/trace_layer.cpp
@@ -99,7 +99,7 @@ static VkInstance get_instance_handle(const void* handle)
 }
 
 // The vk_layerGetPhysicalDeviceProcAddr of the next layer in the chain.
-// Retrieved during instance creation and used to forward this layer's
+// Retrieved during instance creation and forwarded to by this layer's
 // GetPhysicalDeviceProcAddr() after unwrapping its VkInstance parameter.
 static std::mutex                    gpdpa_lock;
 static std::unordered_map<VkInstance, PFN_GetPhysicalDeviceProcAddr> next_gpdpa;
@@ -107,20 +107,20 @@ static std::unordered_map<VkInstance, PFN_GetPhysicalDeviceProcAddr> next_gpdpa;
 static void set_next_gpdpa(const VkInstance instance, PFN_GetPhysicalDeviceProcAddr p_next_gpdpa)
 {
     assert(instance != VK_NULL_HANDLE);
-    assert(p_next_gpdpa != nullptr);
     std::lock_guard<std::mutex> lock(gpdpa_lock);
-    fprintf(stderr, "[gfxreconstruct] - set_next_gpdpa(%p, %p)\n", instance,  p_next_gpdpa);
     next_gpdpa[instance] = p_next_gpdpa;
 }
 
 static PFN_GetPhysicalDeviceProcAddr get_next_gpdpa(const VkInstance instance)
 {
+    assert(instance != VK_NULL_HANDLE);
     std::lock_guard<std::mutex> lock(gpdpa_lock);
     auto it_gpdpa = next_gpdpa.find(instance);
-    assert(it_gpdpa != next_gpdpa.end());
+    if (it_gpdpa == next_gpdpa.end())
+    {
+        return nullptr;
+    }
     auto gpdpa = it_gpdpa->second;
-    assert(gpdpa != VK_NULL_HANDLE);
-    fprintf(stderr, "[gfxreconstruct] - get_next_gpdpa(%p) -> %p\n", instance, gpdpa);
     return gpdpa;
 }
 
@@ -128,7 +128,6 @@ VKAPI_ATTR VkResult VKAPI_CALL dispatch_CreateInstance(const VkInstanceCreateInf
                                                        const VkAllocationCallbacks* pAllocator,
                                                        VkInstance*                  pInstance)
 {
-    fprintf(stderr, "[gfxreconstruct] ENTER - dispatch_CreateInstance()\n");
     VkResult result = VK_ERROR_INITIALIZATION_FAILED;
 
     VkLayerInstanceCreateInfo* chain_info =
@@ -150,7 +149,6 @@ VKAPI_ATTR VkResult VKAPI_CALL dispatch_CreateInstance(const VkInstanceCreateInf
                 chain_info->u.pLayerInfo = chain_info->u.pLayerInfo->pNext;
 
                 result = fpCreateInstance(pCreateInfo, pAllocator, pInstance);
-                fprintf(stderr, "[gfxreconstruct] fpCreateInstance() -> %p\n", (void*)*pInstance);
 
                 if ((result == VK_SUCCESS) && pInstance && (*pInstance != nullptr))
                 {
@@ -167,7 +165,6 @@ VKAPI_ATTR VkResult VKAPI_CALL dispatch_CreateInstance(const VkInstanceCreateInf
         }
     }
 
-    fprintf(stderr, "[gfxreconstruct] EXIT - dispatch_CreateInstance(), *pInstance = %p\n", (void*) *pInstance);
     return result;
 }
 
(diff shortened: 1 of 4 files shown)


layer/trace_layer.cpp
```diff
@@ -241,6 +268,40 @@ VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL GetDeviceProcAddr(VkDevice device, cons
...
+    PFN_vkVoidFunction result = nullptr;
+
+    if (upstreamInstance != VK_NULL_HANDLE)
+    {
+        const VkInstance instance = encode::GetWrappedHandle<VkInstance>(upstreamInstance);
+        fprintf(stderr, "[gfxreconstruct] GetPhysicalDeviceProcAddr(%p, %s), unwrapped instance: %p \n", upstreamInstance, pName, instance);
+        PFN_GetPhysicalDeviceProcAddr next_gpdpa = get_next_gpdpa();
+        if(next_gpdpa)
+        {
+            result = next_gpdpa(instance, pName);
```
andrew-lunarg, line 292: Simply calls through to next gpdpa in the layer chain after unwrapping.

layer/trace_layer.cpp
```diff
@@ -98,10 +98,33 @@ static VkInstance get_instance_handle(const void* handle)
     return (entry != instance_handles.end()) ? entry->second : VK_NULL_HANDLE;
 }
 
+// The vk_layerGetPhysicalDeviceProcAddr of the next layer in the chain.
+// Retrieved during instance creation and used to forward this layer's
+// GetPhysicalDeviceProcAddr() after unwrapping its VkInstance parameter.
+static std::mutex                    gpdpa_lock;
+static PFN_GetPhysicalDeviceProcAddr next_gpdpa = nullptr;
```
bradgrantham-lunarg, line 105: Is it possible for an app to create two instances with different layer settings and thus a different GPDPA per instance?

layer/trace_layer.cpp
```diff
@@ -98,6 +98,32 @@ static VkInstance get_instance_handle(const void* handle)
     return (entry != instance_handles.end()) ? entry->second : VK_NULL_HANDLE;
 }
 
+// The vk_layerGetPhysicalDeviceProcAddr of the next layer in the chain.
+// Retrieved during instance creation and forwarded to by this layer's
+// GetPhysicalDeviceProcAddr() after unwrapping its VkInstance parameter.
+static std::mutex                                                    gpdpa_lock;
+static std::unordered_map<VkInstance, PFN_GetPhysicalDeviceProcAddr> next_gpdpa;
+
+static void set_next_gpdpa(const VkInstance instance, PFN_GetPhysicalDeviceProcAddr p_next_gpdpa)
```
bradgrantham-lunarg, line 107: My preference here, if it's okay with you, is to also mention "instance" in the function name (and below for "get") to echo that it's keyed from the instance, like `get_instance_handle` above.  Maybe `{get,set}_instance_next_gpdpa`?

layer/trace_layer.cpp
```diff
@@ -98,6 +98,32 @@ static VkInstance get_instance_handle(const void* handle)
...
+{
+    assert(instance != VK_NULL_HANDLE);
+    std::lock_guard<std::mutex> lock(gpdpa_lock);
+    auto                        it_gpdpa = next_gpdpa.find(instance);
+    if (it_gpdpa == next_gpdpa.end())
+    {
+        return nullptr;
+    }
+    auto gpdpa = it_gpdpa->second;
+    return gpdpa;
```
bradgrantham-lunarg, line 124: Very clear, but in the future for stuff this short, I'm okay with just `return it_dpdpa->second`.

layer/trace_layer.cpp
```diff
@@ -241,6 +271,28 @@ VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL GetDeviceProcAddr(VkDevice device, cons
...
+ * but we do need to unwrap the instance before the downstream layer
+ * sees it.
+ */
+VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL GetPhysicalDeviceProcAddr(VkInstance ourInstanceWrapper, const char* pName)
+{
+    PFN_vkVoidFunction result = nullptr;
+
+    if (ourInstanceWrapper != VK_NULL_HANDLE)
+    {
+        const VkInstance              nextLayersInstance = encode::GetWrappedHandle<VkInstance>(ourInstanceWrapper);
```
bradgrantham-lunarg, line 285: I doubt that this is the cause of the recursive stack overflow in CI job 1686 but since there's already a function `get_instance_handle` above that does this lookup through the map created in `CreateInstance`, I suggest using that.

layer/trace_layer.cpp
```diff
@@ -388,6 +440,12 @@ extern "C"
         return gfxrecon::GetDeviceProcAddr(device, pName);
     }
 
+    VK_LAYER_EXPORT VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL vk_layerGetPhysicalDeviceProcAddr(VkInstance  instance,
```
bradgrantham-lunarg, line 443: Do you definitely need this?  It looks like the loader will get `gfxrecon::GetPhysicalDeviceProcAddr` through the `pVersionStruct` or through GetProcAddr directly because you added it into `framework/generated/generated_layer_func_table.h` through the generator.

layer/trace_layer.cpp
```diff
@@ -241,6 +271,28 @@ VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL GetDeviceProcAddr(VkDevice device, cons
...
+ */
+VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL GetPhysicalDeviceProcAddr(VkInstance ourInstanceWrapper, const char* pName)
+{
+    PFN_vkVoidFunction result = nullptr;
+
+    if (ourInstanceWrapper != VK_NULL_HANDLE)
+    {
+        const VkInstance              nextLayersInstance = encode::GetWrappedHandle<VkInstance>(ourInstanceWrapper);
+        PFN_GetPhysicalDeviceProcAddr next_gpdpa         = get_next_gpdpa(ourInstanceWrapper);
+        if (next_gpdpa)
```
bradgrantham-lunarg, line 287: Although I personally prefer to leave off the `!= nullptr` in my own projects because I think it looks better, we've established that we use an explicit comparison with `!= nullptr` in this codebase.  Would you take a moment to add `!= nullptr` here?

bradgrantham-lunarg (changes requested):

andrew-lunarg:
This PR depends on the Loader PR 949 https://github.com/KhronosGroup/Vulkan-Loader/pull/949 which has been reverted https://github.com/KhronosGroup/Vulkan-Loader/pull/951.
The loader issue that 949 addresses is a real problem requiring a fix: https://github.com/KhronosGroup/Vulkan-Loader/issues/948.
The [loader PR]( (https://github.com/KhronosGroup/Vulkan-Loader/pull/949) that fixed their issue 948 triggered a Crash in Vulkan-ValidationLayers CI:
https://github.com/KhronosGroup/Vulkan-Loader/issues/950.
This PR is therefore logjammed behind all that. And yet we must have a `GetPhysicalDeviceProcAddr` to not be a bad layer.

layer/trace_layer.cpp
andrew-lunarg: That's probably an artefact of me debugging this where I had a nullptr assert after this line 123  and could single step and see what I'm dealing with in the locals. Changed.

layer/trace_layer.cpp
andrew-lunarg: I don't think we should go through a mutex protected global variable map when we have the value we want sitting here in a member of the structure we have a pointer to. The line is just `const VkInstance nextLayersInstance =                  ourInstanceWrapper->handle;` with some casting and template foo that boils away in the compile. Without the template layering this is what the function we call is doing in full:

```cpp
inline VkInstance GetWrappedHandle(const VkInstance& handle)
{
    if (handle == VK_NULL_HANDLE)
    {
        return VK_NULL_HANDLE;
    }
    HandleWrapper<VkInstance>* wrapper = reinterpret_cast<HandleWrapper<VkInstance>*>(handle);
    VkInstance                 wrapped = wrapper->handle;
    return wrapped;
}
```

layer/trace_layer.cpp
andrew-lunarg: I don't know for sure. I did it for consistency with the other get proc functions. The comment above the other two actually already mentioned a third, but there was no third:
```
    // The following three functions are not directly invoked by the desktop loader, which instead uses the function
    // pointers returned by the negotiate function.
```
Looking at the Loader / Layer interaction docs on `vk_layerGetPhysicalDeviceProcAddr()` It is talked about as a function "added to the layer".
Our validation layers also seem to expose it as a symbol, albeit not extern "C":
```
layers/generated/chassis.cpp
14035:VK_LAYER_EXPORT VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL vk_layerGetPhysicalDeviceProcAddr(VkInstance instance,
```

layer/trace_layer.cpp
bradgrantham-lunarg: Nice inclusion of the called code.  vk_layerGetPhysicalDeviceProcAddr hopefully isn't called frequently, so I'm not very worried about the cost of the mutex, but I accept it's doing much the same thing and we probably should prefer a lock-less indirection.

charles-lunarg (commented):
Does GFXReconstruct allow querying of GPDPA through GIPA? I just recently uncovered a bug in the loader tests which were giving nonsense results because of that.

charles-lunarg:
After discussion elsewhere, yes this change looks to be correct.

charles-lunarg (approved):

layer/trace_layer.cpp
charles-lunarg: I do not believe `vk_layerGetPhysicalDeviceProcAddr` needs to be exported, eg available to the OS dynamic linker. Only drivers need to be doing that and they should be exporting `vk_icdGetPhysicalDeviceProcAddr`, a differently named function.

bradgrantham-lunarg: I think @andrew-lunarg had said in an exchange last week that he would remove it.

andrew-lunarg: Done

bradgrantham-lunarg:
Made a new PR #713 with small required change for loader issue

Issue #677, opened 2022-03-23
GFXR has different behavior compared with target app and validation layer in the following case


mizhen:
If target app deletes a resource but reference to the resource still exist later, GFXR may lead to different behavior compared with target app and validation layer. For example when an image view (let’s call it IMV_A) has already been deleted, but in a vkUpdateDescriptorSets call later, the image view still shows up in an element of a VkDescriptorImageInfo array in VkWriteDescriptorSet  struct.

So when GFXR get the call vkUpdateDescriptorSets, when it tries to get the unwrapped handle of IMV_A, it leads to the following two cases:

1. The pointer to the wrapper struct of IMV_A is still a valid pointer. This case occurs when some other resource is created and when new the resource wrapper struct, the return wrapper pointer happens to be the same with the IMV_A previous wrapper struct. So GFXR will send the wrong resource handle which is later created (the handle is valid, but it may not be same type resource) to driver.

2. The pointer to the wrapper struct of IMV_A is an invalid pointer, it will crash for this case.

Let’s compare the behavior in the same situation for the title running by itself and running with validation layer:

1.  For the title with validation layer, the driver will get a VK_NULL_HANDLE which is Doc allowed in the Doc : https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkDescriptorImageInfo.html , this is because validation layer use global ID (assume we doesn’t set VK_VALIDATION_FEATURE_DISABLE_UNIQUE_HANDLES_EXT).

2. For the title running by itself, the driver will get a handle which was released before. the caused behavior depend on driver implementation, same as title running by itself.

PR #715, opened 2022-06-08
Fix Wolf2 texture missing issue

mizhen:
Currently, when target app call creates any non-dispatchable Vulkan handle , GFXR will replace the real handle with the wrapper pointer and return a fake handle to the target app. So within target app, the wrapper pointer is used as Vulkan non-dispatch handle. When target app calls some API function that uses such handle, GFXR will convert the fake handle to real Vulkan handle and pass it to driver.

Some title shows texture missing issues when running with GFXR capture layer, however the title has no issue when running with validation + capture layer. Further investigation verified that the issue is caused by the target title behaves differently when the fake handle using wrapper pointer. By changing the fake handle to global ID, the texture missing issue is resolved. It’s also the same case for validation layer which use global Id as Vulkan handle, if it's changed to this which make validation layer use pointer as fake handle, same issue will occur. So for the title, the fake handle needs to be the global ID instead of handle wrapper pointer.

The commit changes the fake handle from using wrapper pointer to global ID to fix the texture missing issue for the title, the change also fixes the issue descripted in https://github.com/LunarG/gfxreconstruct/issues/677 .

The change in the commit composite of the following 2 parts:

1. Utility functions used to handle wrapper pointer to global ID conversion. 

        These functions are all in the files vulkan_handle_wrapper_util.h/.cpp, they provide the conversion between wrapper 
    pointer, global ID and Vulkan real non-dispatch handle.
        A class WrapperManager is used for these conversions which manage a map from global ID to handle wrapper pointer.
    For any global ID, the class can get its wrapper pointer and further get its real Vulkan handle through wrapper pointer. 
    Before the commit, handle wrapper handling already provide the mechanism for similar conversion, the commit extend
     the conversion following the existing mechanism and current unified interface.

2. Change all conversion from wrapper pointer to Global Id. 

    All these changes happen in capture side, the commit doesn’t change replay side because replay already use global ID. 
When GFXR write recorded data to trace file, the handle is already changed to global ID. For capture side including capture
and trim handling, the commit switch all wrapper pointer to global ID if the wrapper pointer is used as a  faked Vulkan
non-dispatchable handle.

The commit already test all Sascha welliem samples on Win64 and Win32, also tested Wolf2 and Doom ET, Linux side tested vkcube and all passed. For Doom ET testing, because the title needs deferred operation pull request, so GFXR binary come from merging result from deferred operation fix pull request + this pull request.

The pull-request includes changes of vulkan_api_call_encoders_body_generator.py and gencode.py. Please note that the change in gencode.py is only for adding additional copyright info to auto generated files.

### DIFF
--- a/framework/encode/vulkan_capture_manager.cpp
+++ b/framework/encode/vulkan_capture_manager.cpp
@@ -1,7 +1,7 @@
 /*
 ** Copyright (c) 2018-2021 Valve Corporation
 ** Copyright (c) 2018-2021 LunarG, Inc.
-** Copyright (c) 2019 Advanced Micro Devices, Inc. All rights reserved.
+** Copyright (c) 2019-2022 Advanced Micro Devices, Inc.
 **
 ** Permission is hereby granted, free of charge, to any person obtaining a
 ** copy of this software and associated documentation files (the "Software"),
@@ -109,7 +109,7 @@ void VulkanCaptureManager::InitVkInstance(VkInstance* instance, PFN_vkGetInstanc
     CreateWrappedHandle<NoParentWrapper, NoParentWrapper, InstanceWrapper>(
         NoParentWrapper::kHandleValue, NoParentWrapper::kHandleValue, instance, GetUniqueId);
 
-    auto wrapper = reinterpret_cast<InstanceWrapper*>(*instance);
+    auto wrapper = getWrapperPointerFromHandle<InstanceWrapper*>(*instance);
     LoadInstanceTable(gpa, wrapper->handle, &wrapper->layer_table);
 }
 
@@ -120,7 +120,7 @@ void VulkanCaptureManager::InitVkDevice(VkDevice* device, PFN_vkGetDeviceProcAdd
     CreateWrappedHandle<PhysicalDeviceWrapper, NoParentWrapper, DeviceWrapper>(
         VK_NULL_HANDLE, NoParentWrapper::kHandleValue, device, GetUniqueId);
 
-    auto wrapper = reinterpret_cast<DeviceWrapper*>(*device);
+    auto wrapper = getWrapperPointerFromHandle<DeviceWrapper*>(*device);
     LoadDeviceTable(gpa, wrapper->handle, &wrapper->layer_table);
 }
 
@@ -408,8 +408,9 @@ void VulkanCaptureManager::SetDescriptorUpdateTemplateInfo(VkDescriptorUpdateTem
 
     if (create_info->descriptorUpdateEntryCount > 0)
     {
-        DescriptorUpdateTemplateWrapper* wrapper = reinterpret_cast<DescriptorUpdateTemplateWrapper*>(update_template);
-        UpdateTemplateInfo*              info    = &wrapper->info;
+        DescriptorUpdateTemplateWrapper* wrapper =
+            getWrapperPointerFromHandle<DescriptorUpdateTemplateWrapper*>(update_template);
+        UpdateTemplateInfo* info = &wrapper->info;
 
         for (size_t i = 0; i < create_info->descriptorUpdateEntryCount; ++i)
         {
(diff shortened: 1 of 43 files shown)


per-mathisen-arm:
What is the performance impact of doing so many additional mutex checks? When I implemented my own unreleased Vulkan tracer, I used Intel's TBB concurrent unordered multimap to look up internal handles. Nowadays it would perhaps be best to switch entirely to using VK_EXT_private_data, I think.

mizhen:
> What is the performance impact of doing so many additional mutex checks? When I implemented my own unreleased Vulkan tracer, I used Intel's TBB concurrent unordered multimap to look up internal handles. Nowadays it would perhaps be best to switch entirely to using VK_EXT_private_data, I think.

I didn't see any performance drop during all tests for the pull-request.
Regarding the extension of VK_EXT_private_data, it's a good idea to attach private data to a Vulkan handle. However for this pull-request, it is important to ensure that the fake handle implementation can be used in any hardware/driver implementation.

bradgrantham-lunarg:
@mizhen we just in the last few days fixed a couple of small bugs in the layer implementation.  Could you try 82e04fabc1b92852354e4d379017a7744435410b (or anything later, like `master`) and verify that you still see the issue with Wolf2 with those small fixes?

mizhen:
> @mizhen we just in the last few days fixed a couple of small bugs in the layer implementation. Could you try [82e04fa](https://github.com/LunarG/gfxreconstruct/commit/82e04fabc1b92852354e4d379017a7744435410b) (or anything later, like `master`) and verify that you still see the issue with Wolf2 with those small fixes?

yea, I will do the test.

mizhen:
> > @mizhen we just in the last few days fixed a couple of small bugs in the layer implementation. Could you try [82e04fa](https://github.com/LunarG/gfxreconstruct/commit/82e04fabc1b92852354e4d379017a7744435410b) (or anything later, like `master`) and verify that you still see the issue with Wolf2 with those small fixes?
> 
> yea, I will do the test.

I tried current dev branch a993c4da43545a2dab0aee5a58a24c2b5b3b0eb7 and still saw the issue.

locke-lunarg:
I tested `Wolfenstein II The New Colossus` with `branch dev`. But it didn't happen problem. Could you give more information for `Wolfenstein II The New Colossus` issue. Thank you.

This PR changed invalid handle to null. But it could keep it as invalid handle, it would be better. invalid handle and null have different results for rendering. I think we could just not to delete those wrap objects. Keep those until end, so it could get invalid handles even if they are destroyed.  Also, add an active flag in wrap objects if necessary.

bradgrantham-lunarg:
This may be a good PR to return to when considering moving to mapped handles.  In our testing we couldn't cause a problem in Wolf2:Colossus with GFXR with or without VVL; the problem may have been fixed.  We don't have steps to reproduce so we can't test.  We've merged #885 for solving the handle problem.  So we're closing this at this time.  If the issue can be reproduced, we can reopen this PR with a new issue for Wolf2.

mizhen:
Hi @bradgrantham-lunarg, I just downloaded Steam Wolfenstein II The New Colossus and tested the current dev branch (commit: e4b90928a54d02c534721a6a83ed589815fd4253) on the local machine with RX6700XT, found the issue still exist with dev branch. In the following attached images, screenshot_frame_3352 is the location of game scene where the issue show up, creenshot_frame_4473 is the game setting for the issue, to reproduce the issue, GPU culling should be on. Thanks.

screenshot_frame_4473
screenshot_frame_3352
screenshot_frame_3360_issue

bradgrantham-lunarg:
Thanks for the update, Ming.  We reverted #885, so the version of `dev` you built didn't contain a fix.  We're going to reinvestigate in January.

mizhen:
Thanks, Brad.

I tried same fix as https://github.com/LunarG/gfxreconstruct/pull/885 when I was working on this PR. At the time it didn't change the Wolf2 issue.

locke-lunarg:
@mizhen This PR look good to me. But could you tell me what's wrong with wrapper address and how does global Id fix it? Also, why do VkInstance, VkPhysicalDevice, VkDevice, VkQueue, VkCommandBuffer still use wrapper address?

locke-lunarg:
But for non-dispatch handles, I don't get why they have to use `global ID` as a fake handles. What's wrong with `wrapper pointer address`? I removed the whole `delete` wrappers code, so the wrapper pointer addresses should be an unique number as global ID. I also added a `std::set` to record the whole wrapper pointer addresses in `CreateWrappedDispatchHandle` and `CreateWrappedNonDispatchHandle` to ensure that no duplicated wrapper pointer addresses exist, but texture losing still happens....

locke-lunarg:
https://github.com/LunarG/gfxreconstruct/pull/1048
This PR could fix that issue, so closed.

PR #717, opened 2022-06-09
Correct check of mmap's return value

panos-lunarg:
On error mmap returns the value MAP_FAILED (that is, (void *) -1), not
NULL

### DIFF
--- a/framework/util/page_guard_manager.cpp
+++ b/framework/util/page_guard_manager.cpp
@@ -293,8 +293,6 @@ void* PageGuardManager::AllocateMemory(size_t aligned_size, bool use_write_watch
 {
     assert(aligned_size > 0);
 
-    void* memory = nullptr;
-
     if (aligned_size > 0)
     {
 #if defined(WIN32)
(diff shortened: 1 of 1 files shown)


framework/util/page_guard_manager.cpp
```diff
@@ -317,7 +317,11 @@ void* PageGuardManager::AllocateMemory(size_t aligned_size, bool use_write_watch
 #endif
     }
 
+#if defined(WIN32)
     if (memory == nullptr)
+#else
+    if (memory == MAP_FAILED)
```
bradgrantham-lunarg, line 323: Can the `if (aligned_size > 0)` ever be false above?  Then `memory` would be `nullptr` on Linux, so I think the `#else# clause probably has to check both.  And it would be great to also print `errno` on failure on Linux.  What do you think about refactoring this so that there are failure checks inside the platform clauses inside the `if`, and then a failure print as a new `else` for `if (aligned_size > 0)`?

bradgrantham-lunarg (changes requested):

PR #719, opened 2022-06-16
Clean up for clang build.

bradgrantham-lunarg:
Clean up for clang.
Mark a bunch of functions `inline` so they aren't compiled multiple times in one unit.
Mark a bunch of variables and functions maybe unused.
Delete some unused variables.

### DIFF
--- a/framework/decode/vulkan_enum_util.h
+++ b/framework/decode/vulkan_enum_util.h
@@ -41,7 +41,7 @@ GFXRECON_BEGIN_NAMESPACE(enumutil)
  * @param value VkResult code to process.
  * @return String describing the specified VkResult code.
  */
-static const char* GetResultDescription(VkResult result)
+inline static const char* GetResultDescription(VkResult result)
 {
     // clang-format off
     switch (result)
--- a/framework/format/platform_types.h
+++ b/framework/format/platform_types.h
@@ -190,9 +190,9 @@ extern "C"
         VkDevice device, const VkMemoryGetAndroidHardwareBufferInfoANDROID* pInfo, struct AHardwareBuffer** pBuffer);
 
     inline static VKAPI_ATTR VkResult VKAPI_CALL vkCreateAndroidSurfaceKHR(VkInstance,
-                                                                    const VkAndroidSurfaceCreateInfoKHR*,
-                                                                    const VkAllocationCallbacks*,
-                                                                    VkSurfaceKHR*)
+                                                                           const VkAndroidSurfaceCreateInfoKHR*,
+                                                                           const VkAllocationCallbacks*,
+                                                                           VkSurfaceKHR*)
     {
         GFXRECON_LOG_ERROR("Calling unsupported platform extension function vkCreateAndroidSurfaceKHR");
         return VK_ERROR_EXTENSION_NOT_PRESENT;
@@ -292,10 +292,8 @@ typedef VkResult(VKAPI_PTR* PFN_vkGetSemaphoreZirconHandleFUCHSIA)(
 
 extern "C"
 {
-    inline static VKAPI_ATTR VkResult VKAPI_CALL vkCreateImagePipeSurfaceFUCHSIA(VkInstance,
-                                                                          const VkImagePipeSurfaceCreateInfoFUCHSIA*,
-                                                                          const VkAllocationCallbacks*,
-                                                                          VkSurfaceKHR*)
+    inline static VKAPI_ATTR VkResult VKAPI_CALL vkCreateImagePipeSurfaceFUCHSIA(
+        VkInstance, const VkImagePipeSurfaceCreateInfoFUCHSIA*, const VkAllocationCallbacks*, VkSurfaceKHR*)
     {
         GFXRECON_LOG_ERROR("Calling unsupported platform extension function vkCreateImagePipeSurfaceFUCHSIA");
         return VK_ERROR_EXTENSION_NOT_PRESENT;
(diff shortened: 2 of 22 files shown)


panos-lunarg:
That's a lot of unused variables (maybe parameters too) and functions.
Maybe we should make things a bit more strict and enable things like `-Wall -Wextra` etc?
Making the compiler stricter, although some times bothersome, can be beneficial in the long run

andrew-lunarg:
> Maybe we should make things a bit more strict and enable things like `-Wall -Wextra` etc? Making the compiler stricter, although some times bothersome, can be beneficial in the long run

Makes sense. Perhaps we could do this for internal builds without imposing it on users who may be compiling on compilers/compiler versions we haven't used and would thus have to deal with warnings as errors that we haven't seen.

bradgrantham-lunarg:
> > Maybe we should make things a bit more strict and enable things like `-Wall -Wextra` etc? Making the compiler stricter, although some times bothersome, can be beneficial in the long run
> 
> Makes sense. Perhaps we could do this for internal builds without imposing it on users who may be compiling on compilers/compiler versions we haven't used and would thus have to deal with warnings as errors that we haven't seen.

I thought we already were compiling with `-Wall` because of the number of times I've been scolded by the compiler!  But we are using `-Werror`, so `-Wall` could be painful.  Still, if the team agrees to it, I also think it would be a good idea.

framework/decode/vulkan_enum_util.h
```diff
@@ -41,7 +41,7 @@ GFXRECON_BEGIN_NAMESPACE(enumutil)
  * @param value VkResult code to process.
  * @return String describing the specified VkResult code.
  */
-static const char* GetResultDescription(VkResult result)
+inline static const char* GetResultDescription(VkResult result)
```
andrew-lunarg, line 44: I think using `inline static` will generate a copy of `GetResultDescription()` in every compilation unit using the header, and those copies will be preserved through linking.
If we use plain `inline`, the linker is free to choose an arbitrary one of those copies and eliminate the others.

framework/format/platform_types.h
```diff
@@ -1139,16 +1139,17 @@ typedef VkBool32(VKAPI_PTR* PFN_vkGetPhysicalDeviceScreenPresentationSupportQNX)
...
+                             const VkScreenSurfaceCreateInfoQNX* pCreateInfo,
+                             const VkAllocationCallbacks*        pAllocator,
+                             VkSurfaceKHR*                       pSurface)
     {
         GFXRECON_LOG_ERROR("Calling unsupported platform extension function vkCreateScreenSurfaceQNX");
         return VK_ERROR_EXTENSION_NOT_PRESENT;
     }
 
-    static VKAPI_ATTR VkBool32 VKAPI_CALL vkGetPhysicalDeviceScreenPresentationSupportQNX(
+    inline static VKAPI_ATTR VkBool32 VKAPI_CALL vkGetPhysicalDeviceScreenPresentationSupportQNX(
```
andrew-lunarg, line 1152: `inline` rather than `inline static`?

framework/format/platform_types.h
```diff
@@ -1139,16 +1139,17 @@ typedef VkBool32(VKAPI_PTR* PFN_vkGetPhysicalDeviceScreenPresentationSupportQNX)
 
 extern "C"
 {
-    static VKAPI_ATTR VkResult VKAPI_CALL vkCreateScreenSurfaceQNX(VkInstance                          instance,
-                                                                   const VkScreenSurfaceCreateInfoQNX* pCreateInfo,
-                                                                   const VkAllocationCallbacks*        pAllocator,
-                                                                   VkSurfaceKHR*                       pSurface)
+    inline static VKAPI_ATTR VkResult VKAPI_CALL
```
andrew-lunarg, line 1142: `inline` rather than `inline static`?

framework/format/platform_types.h
```diff
@@ -189,24 +189,24 @@ extern "C"
     typedef VkResult(VKAPI_PTR* PFN_vkGetMemoryAndroidHardwareBufferANDROID)(
         VkDevice device, const VkMemoryGetAndroidHardwareBufferInfoANDROID* pInfo, struct AHardwareBuffer** pBuffer);
 
-    static VKAPI_ATTR VkResult VKAPI_CALL vkCreateAndroidSurfaceKHR(VkInstance,
-                                                                    const VkAndroidSurfaceCreateInfoKHR*,
-                                                                    const VkAllocationCallbacks*,
-                                                                    VkSurfaceKHR*)
+    inline static VKAPI_ATTR VkResult VKAPI_CALL vkCreateAndroidSurfaceKHR(VkInstance,
+                                                                           const VkAndroidSurfaceCreateInfoKHR*,
```
andrew-lunarg, line 193: `inline` rather than `inline static`?

framework/format/platform_types.h
```diff
@@ -189,24 +189,24 @@ extern "C"
...
+                                                                           const VkAllocationCallbacks*,
+                                                                           VkSurfaceKHR*)
     {
         GFXRECON_LOG_ERROR("Calling unsupported platform extension function vkCreateAndroidSurfaceKHR");
         return VK_ERROR_EXTENSION_NOT_PRESENT;
     }
 
-    static VKAPI_ATTR VkResult VKAPI_CALL vkGetAndroidHardwareBufferPropertiesANDROID(
+    inline static VKAPI_ATTR VkResult VKAPI_CALL vkGetAndroidHardwareBufferPropertiesANDROID(
```
andrew-lunarg, line 201: `inline` rather than `inline static`?
There are loads of these in this file. I won't label them all.

framework/generated/generated_vulkan_dispatch_table.h
```diff
@@ -51,7 +51,7 @@ GFXRECON_BEGIN_NAMESPACE(encode)
 typedef const void* DispatchKey;
 
 // Retrieve a dispatch key from a dispatchable handle
-static DispatchKey GetDispatchKey(const void* handle)
+inline static DispatchKey GetDispatchKey(const void* handle)
```
andrew-lunarg, line 54: `inline` rather than `inline static`? Couple more follow in file.

andrew-lunarg (commented):
Looks good if I'm wrong about the `static inline` thing.

bradgrantham-lunarg:
In general I think we should be more strict, passing options equivalent to `-Wall` and `-Wpedantic`.  But this PR was extracted from a MacOS PR and is too specific.  Closing it now in favor of a later pass of making the whole code base more strict.

PR #721, opened 2022-06-20
Store architecture type used when capturing

panos-lunarg:
Store architecture type when capturing a trace in the file's header.
gfxrecon-info prints that information


panos-lunarg:
On a second thought this PR doesn't add anything of value...

PR #733, opened 2022-07-07
Andy fix 722 screen unsupported extensions

### LINKED ISSUE #722: Fail to replay a trace from Hitman 3
I have tried to capture a trace of Hitman 3 with c127490 (dev branch) but then it failed to replay with the following output:

```
[gfxrecon] WARNING - Skipping object handle mapping for unsupported external object type processed by vkGetDescriptorSetHostMappingVALVE
[gfxrecon] WARNING - Skipping object handle mapping for unsupported external object type processed by vkGetDescriptorSetHostMappingVALVE
[gfxrecon] WARNING - Skipping object handle mapping for unsupported external object type processed by vkGetDescriptorSetHostMappingVALVE
[gfxrecon] WARNING - Skipping object handle mapping for unsupported external object type processed by vkGetDescriptorSetHostMappingVALVE
[gfxrecon] WARNING - Skipping object handle mapping for unsupported external object type processed by vkGetDescriptorSetHostMappingVALVE
[gfxrecon] WARNING - Skipping unrecognized file block with type 32679
Total time: 1.500958 seconds
Replay FPS: 0.666241 fps, 1.500958 seconds, 1 frame, framerange 1-1
```

The trace is 4.9GiB, so not sure if it's worth uploading it.

andrew-lunarg:
Screen out unsupported device extensions.

This was made to screen out `VK_VALVE_descriptor_set_host_mapping`, since their `vkd3d-proton` fallback is likely faster and more robust than our attempt to trace it, but I've thrown a few more in there to see how CI reacts to their removal.

Perhaps instead of the blacklist used by this PR we should move in time to a whitelist and screen out all unknown extensions. If we do that we should have a setting to turn the screening off for people who have a new Vulkan implementation but are stuck with a stale gfxr binary for some reason.
Perhaps we should load the black or white list from a file at runtime, also named in a setting.

Fix #722.


layer/trace_layer.cpp
```diff
@@ -53,6 +53,58 @@ const VkLayerProperties kLayerProps = {
...
+    // decisions based on low level PCI bus info about the capture device.
+    "VK_EXT_pci_bus_info",
+    // Looks problematic:
+    "VK_EXT_physical_device_drm",
+    // Look problematic:
+    "VK_KHR_external_memory_fd",
+    "VK_KHR_external_memory",
+    // Supporting the CPU moving around descriptor set data directly has too many
+    // perf/robustness tradeoffs to be worth it.
+    "VK_VALVE_descriptor_set_host_mapping",
```
andrew-lunarg, line 72: This is the only one that absolutely has to be here to fix #722 .

layer/trace_layer.cpp
```diff
@@ -53,6 +53,58 @@ const VkLayerProperties kLayerProps = {
 
 const std::vector<VkExtensionProperties> kDeviceExtensionProps = { VkExtensionProperties{ "VK_EXT_tooling_info", 1 } };
 
+/// An alphabetical list of device extensions which we do not report upstream if
+/// other layers or ICDs expose them to us.
+const char* const kDeviceExtensionBlacklist[] = {
+    // Look problematic:
```
andrew-lunarg, line 59: I added a few more extensions just to exercise the code that runs through this list but these ones I chose do genuinely seem a problem for trace/replay to me.
@bradgrantham-lunarg Should I cull them for now and maybe make a separate issue for expanded the list back out with all the extensions we don't support the tracing of?

bradgrantham-lunarg: Yes, I think make a single separate "task" issue listing these and we can then go address each of these in turn if they become a priority.  It looks like most of these are very special-case and we probably will see them go by as a capture issue if they become more important.

layer/trace_layer.cpp
```diff
@@ -53,6 +53,58 @@ const VkLayerProperties kLayerProps = {
 
 const std::vector<VkExtensionProperties> kDeviceExtensionProps = { VkExtensionProperties{ "VK_EXT_tooling_info", 1 } };
 
+/// An alphabetical list of device extensions which we do not report upstream if
+/// other layers or ICDs expose them to us.
+const char* const kDeviceExtensionBlacklist[] = {
```
bradgrantham-lunarg, line 58: I know this is going to sound PC but would you change "Blacklist" to "Blocklist" in your change?

layer/trace_layer.cpp
bradgrantham-lunarg, line 72: Can this be `VK_VALVE_DESCRIPTOR_SET_HOST_MAPPING_EXTENSION_NAME` ?

layer/trace_layer.cpp
```diff
@@ -337,9 +389,40 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumerateDeviceExtensionProperties(VkPhysicalDevi
...
+        auto     instance_table            = encode::GetInstanceTable(physicalDevice);
+        auto     wrapped_device            = encode::GetWrappedHandle(physicalDevice);
+        uint32_t downstream_property_count = 0;
+
+        result = instance_table->EnumerateDeviceExtensionProperties(
+            wrapped_device, pLayerName, &downstream_property_count, nullptr);
+        if (result != VK_SUCCESS)
+            return result;
+
+        std::vector<VkExtensionProperties> downstream_properties;
```
bradgrantham-lunarg, line 403: Can this and the next line be combined into one of the constructor forms that takes the size?

bradgrantham-lunarg (changes requested):

layer/trace_layer.cpp
andrew-lunarg: I've done something equivalent.

layer/trace_layer.cpp
andrew-lunarg: I'll do that for now. It's my assertion that we should block any extension we can see won't capture right rather than letting users stumble over them, report issues that may not reference the extension in question, and then have to back-track to the root cause. Better for users to see some "required extension missing message" up-front during capture.

layer/trace_layer.cpp
```diff
@@ -337,9 +389,40 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumerateDeviceExtensionProperties(VkPhysicalDevi
...
+        downstream_properties.resize(downstream_property_count);
+        result = instance_table->EnumerateDeviceExtensionProperties(
+            wrapped_device, pLayerName, &downstream_property_count, &downstream_properties[0]);
+        if (result != VK_SUCCESS)
```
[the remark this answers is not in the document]
andrew-lunarg: done and done.

layer/trace_layer.cpp
andrew-lunarg: done.

layer/trace_layer.cpp
andrew-lunarg: Done. Bad habit.

layer/trace_layer.cpp
```diff
@@ -53,6 +53,58 @@ const VkLayerProperties kLayerProps = {
...
+    "VK_KHR_external_memory_fd",
+    "VK_KHR_external_memory",
+    // Supporting the CPU moving around descriptor set data directly has too many
+    // perf/robustness tradeoffs to be worth it.
+    "VK_VALVE_descriptor_set_host_mapping",
+};
+
+static void remove_unsupported_extensions(std::vector<VkExtensionProperties>& extensionProps)
+{
+    std::sort(extensionProps.begin(),
```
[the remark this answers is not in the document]
andrew-lunarg: What I have done is slightly different. I removed what I guess you objected to which was the `.erase()` inside a loop, but I've kept the algorithm in-place, working on an array and a vector, without building a tree or hash table. I have the sort of the extensions at the top which is O(n logn), but expected linear time as the input is usually already sorted. Then I've documented the linear character (caveat I'm treating short string compares as constant time when talking about complexity) of the rest of the function.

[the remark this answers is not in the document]
andrew-lunarg: @bradgrantham-lunarg Let me know if this is resolved when you have time next week or if you want the set of strings.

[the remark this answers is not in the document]
andrew-lunarg: I pushed a simpler version based on `std::remove_if(std::find_if())`. Since the list of extensions to screen out is currently length 1, we have a ways to go before the linear search of it is going to be a problem.

layer/trace_layer.cpp
```diff
@@ -53,6 +53,29 @@ const VkLayerProperties kLayerProps = {
...
+
+static void remove_extensions(std::vector<VkExtensionProperties>& extensionProps,
+                              const char* const                   screenedExtensions[],
+                              const size_t                        screenedCount)
+{
+    auto new_end = std::remove_if(
+        extensionProps.begin(),
+        extensionProps.end(),
+        [&screenedExtensions, screenedCount](const VkExtensionProperties& extension) {
+            return std::find_if(screenedExtensions, &screenedExtensions[screenedCount], [&extension](auto screened) {
```
bradgrantham-lunarg, line 72: As a future consideration I personally have trouble reading nested lambdas, but I think this is fine for getting this in and closing the issue

layer/trace_layer.cpp
```diff
@@ -53,6 +53,29 @@ const VkLayerProperties kLayerProps = {
 
 const std::vector<VkExtensionProperties> kDeviceExtensionProps = { VkExtensionProperties{ "VK_EXT_tooling_info", 1 } };
 
+/// An alphabetical list of device extensions which we do not report upstream if
+/// other layers or ICDs expose them to us.
+const char* const kUnsupportedDeviceExtensions[] = {
```
bradgrantham-lunarg, line 58: As a future consideration I think a searchable list of strings in C++ reads better for me as an `unordered_set` or `set` of `string` but I think this is fine for getting this in and closing the issue.

bradgrantham-lunarg (approved):
Looks great to me!  I left some comments on directions I'd like to see in the future.  Do with those comments what you will.  Note that CI is still failing in `layer/test/main.cpp` and I'd like you to resolve that before merging.  Thank you!

layer/trace_layer.cpp
```diff
@@ -53,6 +53,29 @@ const VkLayerProperties kLayerProps = {
...
+    // Supporting the CPU moving around descriptor set data directly has too many
+    // perf/robustness tradeoffs to be worth it.
+    VK_VALVE_DESCRIPTOR_SET_HOST_MAPPING_EXTENSION_NAME,
+};
+
+static void remove_extensions(std::vector<VkExtensionProperties>& extensionProps,
+                              const char* const                   screenedExtensions[],
+                              const size_t                        screenedCount)
```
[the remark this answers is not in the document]
andrew-lunarg: I tend to make everything const that possibly can be. I could list reasons but it is probably because Carmack started doing that years ago and I'm just a fanboy. I should stop doing that in this codebase?

layer/trace_layer.cpp
andrew-lunarg: Understood.

andrew-lunarg:
> CI gfxreconstruct build # 1779 failed.

These should be big and bold and red and maybe use some '90s style flashing html style.

layer/trace_layer.cpp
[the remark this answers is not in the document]
bradgrantham-lunarg: This one in particular is in a parameter, and const in a parameter to me usually says something about the contract between the caller and callee.  Since this is a value parameter, const or not doesn't matter to the caller.  It will error throughout the function itself if you alter it but that's not something I frequently see.  So to me it is a signal I'll parse while reading this code and wonder why the parameter is marked that way.  However, it does mean that you get error if you alter it by mistake!  So I'm convinced.  Never mind!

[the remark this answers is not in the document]
andrew-lunarg: Fun fact: A function prototype elsewhere can omit the const and users will still find an implementation like this one with const in the parameter list and link successfully. Putting it in a header would actually be pretty meaningless.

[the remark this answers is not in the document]
bradgrantham-lunarg: Yes.  I just don't remember frequently seeing the `const` added in implementations for value parameters and I never add it myself.  Maybe that's a nice habit to pick up!

PR #743, opened 2022-07-18
[DRAFT] Don't record swapchain images setup while replaying+recording

panos-lunarg:
Experimental

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -962,6 +962,9 @@ void VulkanReplayConsumerBase::ProcessSetSwapchainImageStateQueueSubmit(
     auto table  = GetDeviceTable(device);
     assert(table != nullptr);
 
+    if (table->GFXRStopRecording != encode::noop::GFXRStopRecording)
+        table->GFXRStopRecording();
+
     VkResult        result             = VK_SUCCESS;
     VkQueue         queue              = VK_NULL_HANDLE;
     VkCommandPool   pool               = VK_NULL_HANDLE;
@@ -1261,6 +1264,9 @@ void VulkanReplayConsumerBase::ProcessSetSwapchainImageStateQueueSubmit(
     {
         table->DestroyFence(device, wait_fence, nullptr);
     }
+
+    if (table->GFXRResumeRecording != encode::noop::GFXRResumeRecording)
+        table->GFXRResumeRecording();
 }
 
 void VulkanReplayConsumerBase::ProcessBeginResourceInitCommand(format::HandleId device_id,
--- a/framework/encode/capture_manager.h
+++ b/framework/encode/capture_manager.h
@@ -128,6 +128,20 @@ class CaptureManager
 
     virtual CaptureSettings::TraceSettings GetDefaultTraceSettings();
 
+    void stop_recording()
+    {
+        old_capture_mode_ = capture_mode_;
+        capture_mode_ &= ~kModeWrite;
+    }
+
+    void resume_recording()
+    {
+        if (old_capture_mode_ & kModeWrite)
+        {
+            capture_mode_ |= kModeWrite;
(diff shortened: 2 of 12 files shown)


panos-lunarg:
Sorry forgot to create this as a draft so I added a note in the title

bradgrantham-lunarg:
Great, Panos, thank you!

I assume the generated code changes would be added to the generators if this moved forward.

It might be nice to know what happens with these getprocaddrs if other layers are active upchain of the GFXR layer during replay.  That is to say, if 

Also any other app could getprocaddr and call these functions, right?  If GFXR becomes more widespread but game developers want to forcibly disable capturing in their app, I think they might call these functions.  Could this functionality be more abstract, like a way for replay to note to capture that it is beginning and ending a replay state snapshot (trimmed range state setup) command, and then the capture layer can recreate the kBeginMarker and kEndMarker that are in the source capture in the new capture?  Then a subsequent replay can know to not screenshot during state snapshot replay.

panos-lunarg:
> It might be nice to know what happens with these getprocaddrs if other layers are active upchain of the GFXR capture layer during replay. That is to say, if VVL is between replay and the capture layer.

Good question. I haven't thought of that. Need to check.

>I think if recording is just disabled/re-enabled by gfxrecon-replay, then information is doubly lost. Won't the recaptured capture's first Acquire expect a different index?

From what I understand/have seen from the runs I did locally, `SetSwapchainImageStateCommand` is only called when replaying a trimmed capture. This is done in order to set the state of the swapchain images in such a manner that is in agreement with the state at of the images on the first captured frame. If the replay is re-captured again then I guess that information does not matter since:
- If capturing un-trimmed then swapchain images start from scratch.
- If capturing trimmed then a new `SetSwapchainImageStateCommandHeader` will be recorded.
So I think we don't lose any information that needs to be preserved in the new capture. These are all guesses/thoughts on my side so feel free to straighten me out in case I am wrong.

bradgrantham-lunarg:
> > I think if recording is just disabled/re-enabled by gfxrecon-replay, then information is doubly lost. Won't the recaptured capture's first Acquire expect a different index?
> 
> From what I understand/have seen from the runs I did locally, `SetSwapchainImageStateCommand` is only called when replaying a trimmed capture. This is done in order to set the state of the swapchain images in such a manner that is in agreement with the state at of the images on the first captured frame. If the replay is re-captured again then I guess that information does not need to be recapruted since:
> 
> * If capturing un-trimmed then swapchain images start from scratch.
> * If capturing trimmed then a new `SetSwapchainImageStateCommandHeader` will be created and recorded from the layer.

When replay processes `SetSwapchainImageStateCommandHeader`, if all the images can't be acquired up front (https://github.com/LunarG/gfxreconstruct/blob/6bf1a6a89a80ef7e490f0b9ca4259d8386e1da21/framework/decode/vulkan_replay_consumer_base.cpp#L720), the replay issues the initial `Present`s to get to the correct swapchain index.  In current `dev` applied, on some platforms that becomes unexpected black/junk initial images in replay of the recapture but at the end of the initial images the swapchain index will be set correctly.  That's the whole point of this PR if I understand it correctly.

But with this PR applied the **replay of the recapture** won't make those initial image `Present`s *and* the swapchain index won't be set correctly.  I think this will be true for both trimmed and un-trimmed recapture.  I talked this through with David and we're pretty sure as well as not recording the `Present` calls a recapture will need to correctly reset the swapchain index.  So when replay calls `StopRecording`, the capture layer will need to set a flag to package up and write a new `SetSwapchainImageStateCommand` when `ResumeRecording` is called.

bradgrantham-lunarg:
I think `virtual-swapchain` handles this.  What do you think?

bradgrantham-lunarg:
@panos-lunarg I think `virtual-swapchain` handles this.  What do you think?

panos-lunarg:
Yeah, sounds good

PR #747, opened 2022-07-25
Keep same usage flags for AHB when replaying

panos-lunarg:
When replaying the AndroidHardwareBuffer usage flags were altered
compared to the recorded flags. This caused vkAllocateMemory to fail
when importing an AHB on Android 12.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -421,9 +421,6 @@ void VulkanReplayConsumerBase::ProcessCreateHardwareBufferCommand(
     desc.usage                = usage;
     desc.width                = width;
 
-    // Make sure we can write to the buffer.
-    desc.usage |= AHARDWAREBUFFER_USAGE_CPU_WRITE_OFTEN;
-
     AHardwareBuffer* buffer = nullptr;
     int              result = AHardwareBuffer_allocate(&desc, &buffer);
     if ((result == 0) && (buffer != nullptr))
@@ -432,8 +429,7 @@ void VulkanReplayConsumerBase::ProcessCreateHardwareBufferCommand(
         ahb_info.memory_id           = memory_id;
         ahb_info.hardware_buffer     = buffer;
 
-        void* data = nullptr;
-        result     = -1;
+        result = -1;
 
         std::vector<format::HardwareBufferPlaneInfo> replay_plane_info;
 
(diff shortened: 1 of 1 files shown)


panos-lunarg:
I believe this is ready to be reviewed.

I also tried out building with `__ANDROID_API__ >= 29` and seems to work fine.

In the recording layer the locking that takes places makes sense as the AHB content is accessed and stored in the capture file (although I am suspecting that the code will not work as expected when the AHB is a true multiplanar/YUV image but I don't know whether we have a scenario to test out any changes)

PS. To get to compile for `__ANDROID_API__ >= 29`:
```
diff --git a/android/tools/replay/build.gradle b/android/tools/replay/build.gradle
index b054e8aa..a19dbc6d 100644
--- a/android/tools/replay/build.gradle
+++ b/android/tools/replay/build.gradle
@@ -1,12 +1,12 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 27
+    compileSdkVersion 29
     ndkVersion '21.3.6528147'
     defaultConfig {
         applicationId "com.lunarg.gfxreconstruct.replay"
-        minSdkVersion 26
-        targetSdkVersion 27
+        minSdkVersion 29
+        targetSdkVersion 29
         versionCode 1
         versionName "1.0"
         ndk {
```

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -432,108 +429,93 @@ void VulkanReplayConsumerBase::ProcessCreateHardwareBufferCommand(
...
+            }
+            else
             {
-                format::HardwareBufferPlaneInfo ahb_plane_info;
-                ahb_plane_info.offset =
-                    reinterpret_cast<uint8_t*>(ahb_planes.planes[i].data) - reinterpret_cast<uint8_t*>(data);
-                ahb_plane_info.pixel_stride = ahb_planes.planes[i].pixelStride;
-                ahb_plane_info.row_pitch    = ahb_planes.planes[i].rowStride;
-                replay_plane_info.emplace_back(std::move(ahb_plane_info));
+                GFXRECON_LOG_WARNING("AHardwareBuffer_lockPlanes failed: AHardwareBuffer_lock will be used instead");
```
bradgrantham-lunarg, line 469: Is this warning obsolete now?

bradgrantham-lunarg (approved):
Looks great but check out that warning about "`_lock` will be used instead"

framework/decode/vulkan_replay_consumer_base.cpp
panos-lunarg: No it should still be valid.
In the `__ANDROID_API__ >= 29` case I left the AHB lock as is. In this case it actually kinda accesses the buffer's data (`ahb_planes.planes[i].data`) although just to calculate offsets.
I wonder if there's a better way to calculate those offsets without locking but I am a bit reluctant to tamper with the `__ANDROID_API__ >= 29` code as it deals with multi-planar images and I don't know if we have a test case to test potential changes.

bradgrantham-lunarg: It looks like you deleted the call to AHardwareBuffer_lock.  Am I misreading the diff?

bradgrantham-lunarg: I don't see an instance of `_lock` in ProcessCreateHardwareBufferCommand in framework/decode/vulkan_replay_consumer_base.cpp in your branch.

panos-lunarg: Oops, sorry, you're right.

bradgrantham-lunarg (approved):

PR #751, opened 2022-08-02
warn if replay image_index differs from capture

bradgrantham-lunarg:


bradgrantham-lunarg:
Functionality absorbed into Virtual Swapchain

PR #756, opened 2022-08-09
Add timestamp metadata and use it to replay at same speed

JnCrMx:
This adds a timestamp command before every VkQueuePresent and syncs the relative delay between timestamp commands in replay.

Therefore, the replay will be around the same speed as the original program and e.g. animations do not look too fast.

This is just something I needed and found useful for myself. I would probably make this optional for replay and maybe for capture as well. The changes required for that will be added later on.

I would appreciate any kind of feedback on this, but it would of course also be fine if this kind of feature is not desired for GFXreconstruct.

### DIFF
--- a/USAGE_android.md
+++ b/USAGE_android.md
@@ -305,7 +305,7 @@ usage: gfxrecon.py replay [-h] [-p LOCAL_FILE] [--version] [--pause-frame N]
                           [--screenshot-format FORMAT] [--screenshot-dir DIR]
                           [--screenshot-prefix PREFIX] [--sfa] [--opcd]
                           [--surface-index N] [--sync] [--remove-unsupported]
-                          [-m MODE]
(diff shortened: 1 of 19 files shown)


andrew-lunarg:
I think there would have to be documentation edits to let users know how to turn it on:
https://github.com/LunarG/gfxreconstruct/blob/dev/USAGE_desktop.md
https://github.com/LunarG/gfxreconstruct/blob/dev/USAGE_android.md
It seems to always be enabled currently, no?
The best way to get feedback on a feature idea would be to raise an issue, and then the discussion in any corresponding PRs can be restricted to the readiness of the actual code implementing the idea.

JnCrMx:
Yes, it currently is always enabled. I still need to implement an option for enabling/disabling it.
And thank you very much for the advice, I will open an issue to discuss the idea!

PR #773, opened 2022-08-24
Deferred operation for CreateRayTracingPipelinesKHR

locke-lunarg:
The old PR https://github.com/LunarG/gfxreconstruct/pull/653 changes order of both capture and replay. It runs `DeferredOperationJoinKHR` immediately after `CreateRayTracingPipelinesKHR`.

This PR keeps the original order, so it needs to record parameters and their pointer's data of CreateRayTracingPipelinesKHR to avoid those parameters be modified or invalid. Replay is more complicated than capture.  Now, it has a problem for replay. It's hard to deep copy pNext and pNext's pointer data. So if the application uses pNext, the replay would fail.

The second solution is to record decode data, like `StructPointerDecoder<Decoded_VkRayTracingPipelineCreateInfoKHR> pCreateInfos`, example: https://github.com/locke-lunarg/gfxreconstruct/commits/deferred3. But the problem is that it uses `DecodeAllocator::Allocate` to allocate memory and save data. The memory could be overwrite in the following function call, so the data might be overwrite before `DeferredOperationJoinKHR`. It can't guarantee safety. We have two ways to fix it. Firstly, we could add a option to save data in another container, instead of `DecodeAllocator::Allocate`. But it might take a week. Secondly, we could add a switch in `DecodeAllocator::End()` to avoid it to clear memory before `DeferredOperationJoinKHR`.

The third solution is the most safety and easy that is to run `DeferredOperationJoinKHR` immediately after `CreateRayTracingPipelinesKHR`. It's just like the old PR, but only for replay. But it changes the order.

### DIFF
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -199,7 +199,6 @@ typedef VulkanObjectInfo<VkDebugUtilsMessengerEXT>        DebugUtilsMessengerEXT
 typedef VulkanObjectInfo<VkAccelerationStructureKHR>      AccelerationStructureKHRInfo;
 typedef VulkanObjectInfo<VkAccelerationStructureNV>       AccelerationStructureNVInfo;
 typedef VulkanObjectInfo<VkPerformanceConfigurationINTEL> PerformanceConfigurationINTELInfo;
-typedef VulkanObjectInfo<VkDeferredOperationKHR>          DeferredOperationKHRInfo;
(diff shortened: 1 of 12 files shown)


locke-lunarg:
I think `adding a switch in DecodeAllocator::End() to avoid it to clear memory before GetDeferredOperationResultKHR` is the best solution. It could replay a lot of functions between `CreateRayTracingPipelinesKHR` and `GetDeferredOperationResultKHR`. The whole functions' data will be kept and not be released. But I think the data size should be fine.

PR #776, opened 2022-08-24
Deferred operation for CreateRayTracingPipelinesKHR

locke-lunarg:
The old PRs: https://github.com/LunarG/gfxreconstruct/pull/653, https://github.com/LunarG/gfxreconstruct/pull/773

This PR keeps the original order, so it needs to record parameters and their pointer's data of `CreateRayTracingPipelinesKHR` to avoid those parameters be modified or invalid.

Replay is more complicated than capture. It is to record decode data, like `StructPointerDecoder<Decoded_VkRayTracingPipelineCreateInfoKHR> pCreateInfos`. It added a switch in `DecodeAllocator::End()` to avoid it to clear memory before `DeferredOperationJoinKHR`. It might replay a lot of functions between CreateRayTracingPipelinesKHR and GetDeferredOperationResultKHR. Those functions' data will not be released until `DeferredOperationJoinKHR`. But I think the data size should be not big. A potential problem is if an application uses deferred operation on `CreateRayTracingPipelinesKHR`, but it doesn't run `DeferredOperationJoinKHR` for some unknown reasons, it will keep the whole data until end.

It's normal if it prints a lot of `[gfxrecon] WARNING - API call vkDeferredOperationJoinKHR returned value VK_SUCCESS that does not match return value from capture file: VK_THREAD_IDLE_KHR.` During capture, it might run tens or hundreds times `DeferredOperationJoinKHR` until it gets VK_SUCCESS. The whole `DeferredOperationJoinKHR` are written into files. During replay, it runs loop in first `DeferredOperationJoinKHR` until it gets VK_SUCCESS. Since the first `DeferredOperationJoinKHR` has gotten VK_SUCCESS, the rest will get VK_SUCCESS, too. So the replay results are different to the capture results.

### DIFF
--- a/framework/decode/vulkan_decoder_base.h
+++ b/framework/decode/vulkan_decoder_base.h
@@ -178,6 +178,10 @@ class VulkanDecoderBase : public ApiDecoder
                                                        const uint8_t*     parameter_buffer,
                                                        size_t             buffer_size);
 
+    size_t Decode_vkCreateRayTracingPipelinesKHR(const ApiCallInfo& call_info,
+                                                 const uint8_t*     parameter_buffer,
+                                                 size_t             buffer_size);
+
   private:
     std::vector<VulkanConsumer*> consumers_;
 };
--- a/framework/decode/vulkan_decoder_base.h
+++ b/framework/decode/vulkan_decoder_base.h
@@ -35,6 +35,7 @@
 
 #include <algorithm>
 #include <vector>
+#include <unordered_map>
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
@@ -182,8 +183,20 @@ class VulkanDecoderBase : public ApiDecoder
                                                  const uint8_t*     parameter_buffer,
                                                  size_t             buffer_size);
 
+    size_t Decode_vkDeferredOperationJoinKHR(const ApiCallInfo& call_info,
+                                             const uint8_t*     parameter_buffer,
+                                             size_t             buffer_size);
+
   private:
     std::vector<VulkanConsumer*> consumers_;
+
+    struct DeferredOperationFunctionCallData
+    {
+        StructPointerDecoder<Decoded_VkRayTracingPipelineCreateInfoKHR> pCreateInfos;
+        StructPointerDecoder<Decoded_VkAllocationCallbacks>             pAllocator;
+        HandlePointerDecoder<VkPipeline>                                pPipelines;
+    };
+    std::unordered_map<format::HandleId, DeferredOperationFunctionCallData> record_deferred_operation_function_call;
 };
 
 GFXRECON_END_NAMESPACE(decode)
--- a/framework/encode/vulkan_handle_wrappers.h
+++ b/framework/encode/vulkan_handle_wrappers.h
@@ -308,6 +308,7 @@ struct PipelineWrapper : public HandleWrapper<VkPipeline>
     // Ray tracing pipeline's shader group handle data
     format::HandleId     device_id{ format::kNullHandleId };
     std::vector<uint8_t> shader_group_handle_data;
+    CreateDependencyInfo deferred_operation;
 
     // TODO: Base pipeline
     // TODO: Pipeline cache
--- a/framework/encode/vulkan_handle_wrappers.h
+++ b/framework/encode/vulkan_handle_wrappers.h
@@ -78,7 +78,6 @@ struct DebugUtilsMessengerEXTWrapper        : public HandleWrapper<VkDebugUtilsM
 struct ValidationCacheEXTWrapper            : public HandleWrapper<VkValidationCacheEXT> {};
 struct IndirectCommandsLayoutNVWrapper      : public HandleWrapper<VkIndirectCommandsLayoutNV> {};
 struct PerformanceConfigurationINTELWrapper : public HandleWrapper<VkPerformanceConfigurationINTEL> {};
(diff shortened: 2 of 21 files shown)


locke-lunarg:
@bradgrantham-lunarg 
Support VK_NVX_binary_import for some games. But it blocks vkCmdCuLaunchKernelNVX  and VkCuLaunchInfoNVX since  pParams and pExtras couldn't be encoded correctly. Nvidia device could capture and replay as AMD device. I couldn't find a way to support some games without support VK_NVX_binary_import.

framework/encode/custom_vulkan_api_call_encoders.cpp
```diff
@@ -348,5 +348,12 @@ VKAPI_ATTR VkResult VKAPI_CALL CopyAccelerationStructureKHR(VkDevice
     return GetDeviceTable(device)->CopyAccelerationStructureKHR(device, deferredOperation, pInfo);
 }
 
+VKAPI_ATTR void VKAPI_CALL CmdCuLaunchKernelNVX(VkCommandBuffer commandBuffer, const VkCuLaunchInfoNVX* pLaunchInfo)
+{
+    // TODO
+    GFXRECON_LOG_ERROR("CmdCuLaunchKernelNVX encoding is not supported");
```
locke-lunarg, line 354: Skip encoding CmdCuLaunchKernelNVX since pParams and pExtras  in VkCuLaunchInfoNVX couldn't be encoded correctly.

framework/encode/custom_vulkan_api_call_encoders.cpp
```diff
@@ -348,5 +348,12 @@ VKAPI_ATTR VkResult VKAPI_CALL CopyAccelerationStructureKHR(VkDevice
...
+    return GetDeviceTable(commandBuffer)->CmdCuLaunchKernelNVX(commandBuffer, pLaunchInfo);
+}
```
locke-lunarg, line 356: Skip encoding CmdCuLaunchKernelNVX since pParams and pExtras in VkCuLaunchInfoNVX couldn't be encoded correctly.

framework/decode/vulkan_decoder_base.h
```diff
@@ -178,8 +179,24 @@ class VulkanDecoderBase : public ApiDecoder
...
   private:
     std::vector<VulkanConsumer*> consumers_;
+
+    struct DeferredOperationFunctionCallData
+    {
+        StructPointerDecoder<Decoded_VkRayTracingPipelineCreateInfoKHR> pCreateInfos;
+        StructPointerDecoder<Decoded_VkAllocationCallbacks>             pAllocator;
+        HandlePointerDecoder<VkPipeline>                                pPipelines;
+    };
+    std::map<format::HandleId, DeferredOperationFunctionCallData> record_deferred_operation_function_call;
```
bradgrantham-lunarg, line 199: Can this be an `unordered_map`?

framework/encode/vulkan_handle_wrappers.h
```diff
@@ -309,11 +308,22 @@ struct PipelineWrapper : public HandleWrapper<VkPipeline>
...
+    CreateDependencyInfo deferred_operation;
 
     // TODO: Base pipeline
     // TODO: Pipeline cache
 };
 
+struct DeferredOperationKHRWrapper : public HandleWrapper<VkDeferredOperationKHR>
+{
+    // Record CreateRayTracingPipelinesKHR parameters for safety.
+    HandleUnwrapMemory                             record_handle_unwrap_memory;
```
bradgrantham-lunarg, line 320: None of the other Wrappers have `record_` in the variable name.  Does `record` mean something special in this case that the other Wrappers don't do?

framework/encode/vulkan_state_tracker.cpp
```diff
@@ -597,6 +597,23 @@ void VulkanStateTracker::TrackUpdateDescriptorSets(uint32_t                    w
...
                     binding.write_pnext = TrackPNextStruct(write->pNext, &binding.write_pnext_memory);
+                    auto* pnext         = reinterpret_cast<const VkBaseInStructure*>(binding.write_pnext);
+                    switch (pnext->sType)
+                    {
+                        case VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET_ACCELERATION_STRUCTURE_KHR:
+                        {
+                            auto* acc_struct = reinterpret_cast<VkWriteDescriptorSetAccelerationStructureKHR*>(
+                                const_cast<void*>(binding.write_pnext));
+                            binding.record_write_set_accel_structs.resize(acc_struct->accelerationStructureCount);
+                            std::memcpy(binding.record_write_set_accel_structs.data(),
```
bradgrantham-lunarg, line 608: could this use std::copy?

bradgrantham-lunarg (changes requested):

framework/encode/vulkan_handle_wrappers.h
locke-lunarg: It's not necessary. I can remove it.

framework/encode/vulkan_state_tracker.cpp
locke-lunarg: the copy source is VkAccelerationStructureKHR*, not a STL container. std::copy doesn't work on it.

locke-lunarg: fixed

framework/encode/vulkan_handle_wrappers.h
locke-lunarg: fixed

PR #784, opened 2022-08-30
Fix build.py error if --clean or --clobber used

### LINKED ISSUE #782: build.py Error Exits if `--clean` or `--clobber` Used
### Platform
Ubuntu 22.04. 1 LTS, 64 Bit
Python 3.10.4

### Reproduction

```

rm -rf ./dbuild/ ./build/ ; python3  ./scripts/build.py --clean
.../scripts/build.py:28: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  import distutils.version

rm -rf ./dbuild/ ./build/ ; python3  ./scripts/build.py --clobber

..././scripts/build.py:28: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  import distutils.version
```

JonathanPlasse:
Fix #782

### DIFF
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -25,7 +25,7 @@ GFXReconstruct build script
 '''
 
 import argparse
-import distutils.version
+import packaging.version
 import os
 import platform
 import re
@@ -46,10 +46,10 @@ DEFAULT_ARCHITECTURE = ARCHITECTURES[0]
 BUILD_ROOT = os.path.abspath(
     os.path.join(os.path.split(os.path.abspath(__file__))[0], '..'))
 BUILD_CONFIGS = {'debug': 'dbuild', 'release': 'build'}
-CMAKE_VERSION_3_13 = distutils.version.StrictVersion('3.13.0')
+CMAKE_VERSION_3_13 = packaging.version.Version('3.13.0')
 CONFIGURATIONS = ['release', 'debug']
 DEFAULT_CONFIGURATION = CONFIGURATIONS[0]
-VERSION = distutils.version.StrictVersion('0.0.0')
+VERSION = packaging.version.Version('0.0.0')
 
 
 class BuildError(Exception):
@@ -172,7 +172,7 @@ def cmake_version():
         r'cmake version (?P<version>[\d\.]+)', cmake_version_output)
     if match is None:
         raise BuildError('failed to get CMake version')
-    cmake_version = distutils.version.StrictVersion(match.group('version'))
+    cmake_version = packaging.version.Version(match.group('version'))
     return cmake_version
 
 
--- a/BUILD.md
+++ b/BUILD.md
@@ -96,6 +96,14 @@ be used to update Git submodule dependencies, configure, and build the GFXRecons
 software.  The script works with both Windows and Linux, and requires Python 3.5 or
 above.
 
+The python module `packaging` must be installed before running the build script.
+Here is how to install it on Debian/Ubuntu.
+
+```bash
+sudo apt install python3-pip
+pip install packaging
+```
+
 By default, the script performs some optional build steps that are intended for
 developer builds.  If the target build system has not been configured for
 developer builds as described in [CONTRIBUTING.md](CONTRIBUTING.md), the script


andrew-lunarg:
Thank you for the fix Jonathan. I think it would at least need some docs* updated to mention installing the `packaging` module as that doesn't seem to be installed by default on my Ubuntu 22.04 box.

```
  git remote add ext-JonathanPlasse git@github.com:JonathanPlasse/gfxreconstruct.git
  git pull ext-JonathanPlasse 
  git checkout fix-build-py 
  rm -rf ./dbuild/ ./build/ ;  python3  ./scripts/build.py -a x64 --clean --clobber --config debug --skip-tests --lint</pre>
Traceback (most recent call last):
  File ".../scripts/build.py", line 28, in <module>
    import packaging.version
ModuleNotFoundError: No module named 'packaging'
```

After installing pip and packaging with it, `--clean` and `--clobber` seem to work. So for the Debian/Ubuntu case:
```
sudo apt install python3-pip
pip install packaging
```
* https://github.com/LunarG/gfxreconstruct/blob/dev/BUILD.md

scripts/build.py
```diff
@@ -172,7 +172,7 @@ def cmake_version():
         r'cmake version (?P<version>[\d\.]+)', cmake_version_output)
     if match is None:
         raise BuildError('failed to get CMake version')
-    cmake_version = distutils.version.StrictVersion(match.group('version'))
+    cmake_version = packaging.version.Version(match.group('version'))
```
lunarpapillo, line 175: This parses the version string twice; and the `cmake version` pattern will fail on many inputs legal to `packaging.version`.  `packaging.version.VERSION_PATTERN` can be used to construct a pattern that will correctly match all legal inputs and only require parsing once. 

But this may not be important if we don't need to support the full version syntax... but then, if we only need to support `X.Y.Z` versions, a tuple comparison is easier than adding a whole module just for version comparison.

scripts/build.py
```diff
@@ -46,10 +46,10 @@ def is_windows():
 BUILD_ROOT = os.path.abspath(
     os.path.join(os.path.split(os.path.abspath(__file__))[0], '..'))
 BUILD_CONFIGS = {'debug': 'dbuild', 'release': 'build'}
-CMAKE_VERSION_3_13 = distutils.version.StrictVersion('3.13.0')
+CMAKE_VERSION_3_13 = packaging.version.Version('3.13.0')
```
lunarpapillo, line 49: This is only used to compare CMake version at one point.  It's not clear to me why using the full `Version` implementation is superior to a simple integer list comparison:
```
CMAKE_VERSION_3_13 = [3, 13, 0]
```
with corresponding changes to `cmake_version`:
```
cmake_version = [int(x) for x in match.group('version').split('.')]
```

lunarpapillo (changes requested):
This is fine; it will do what it says it will, though the code is a little clumsy.

There are easier ways to do limited version parsing and comparison that do not require installing the `packaging` module on all build systems, as we're not using `packaging` for anything other than simple version parsing and comparison.

If we need to support full `packaging` version syntax, then the `cmake_version` function will need to be updated.  If we're not, and we can use the simple integer versions that `cmake_version` parses for, we don't need to use `packaging`.

There are a lot of other issues here:
- The script does not pass `pylint`.
- The project doesn't even have a defined `.pylintrc`.  (Is some other syntax checker defined for this project?)
- The `--version` switch doesn't have a `help=` string, so I can't tell what it's intended to do.
- `args.version` is never used or referenced, so I can't reverse-engineer what `--version` is supposed to do.  (Is it even necessary at all?)

But all of those are pre-existing problems that aren't directly relevant to this PR.  I recommend:
- a tracking issue be created to establish a linter and appropriate settings for this project's scripts
- *either* this PR changes to fully support `packaging.version` syntax, *or* changes to not require `packaging` and to do list comparisons for version comparisons
- `--version` be removed as unnecessary.

andrew-lunarg:
While it is great that Jonathan has jumped in here with an immediate fix for a "bug" I raised, I'm embarrassed to have to admit that the issue was raised in error: I described it as causing an error exit incorrectly based on my misunderstanding of the normal operation of the script. The normal function of `--clean` is not to clean _and then build_, but merely to clean and exit. Similarly for `---clobber`. I have updated #782 to say as much. We still have until Python 3.12 is the default on platforms that we care about before needing to address the deprecation warning.

PR #785, opened 2022-08-30
Usage: --use-captured-swapchain-indice

locke-lunarg:


bradgrantham-lunarg (approved):

andrew-lunarg:
As someone who is new to this feature, it would be useful to have some text describing what the flag does in greater detail than the name of it already does. The descriptions on the flags just add a few redundant words like 'the' and 'for' and 'images'  and replace some hyphens with spaces compared to the name of the flag.

Perhaps a follow-up enhancement issue could remind us to add that documentation later, and perhaps review docs for all flags if others are similar?

USAGE_desktop.md
```diff
@@ -448,6 +455,10 @@ Key(s) | Action
 Space, p | Toggle pause/play.
 Right arrow, n | Advance to the next frame when paused.
 
+### Virtual Swapchain
+
+During replay, swapchain indices for present could be different from captured indices. For example, the swapchain image count is different between capture and replay, or vkAcquireNextImageKHR gets a different pImageIndex to capture. These issues could cause unexpected rendering or even crashes. Virtual swapchain creates some images to replace real swapchain images to follow captured file's process. Before present, it copies the virtual image to the target swapchain image for present. Since this issue could happen in many situations, virtual swapchain is the default setup. If the user wants to use the captured indices for present, use `--use-captured-swapchain-indices` option.
```
andrew-lunarg, line 460: During replay, swapchain indices for present can be different from captured indices. Causes for this can include the swapchain image count differing between capture and replay, and `vkAcquireNextImageKHR` returning a different `pImageIndex` at replay to the one that was captured. These issues can cause unexpected rendering or even crashes.
   Virtual Swapchain insulates higher layers in the Vulkan stack from these problems by creating a set of images, exactly matching the swapchain configuration at capture time, which it exposes for them to render into.  Before a present, it copies the virtual image to a target swapchain image for display. Since this issue can happen in many situations, virtual swapchain is the default setup. If the user wants to bypass the feature and use the captured indices to present directly on the swapchain of the replay implementation, they should add the `--use-captured-swapchain-indices` option when invoking `gfxrecon-replay`.

andrew-lunarg (commented):
This is much more understandable. I've added some polish / wording suggestions. Feel free to ignore them or use them as you wish :).

locke-lunarg:
@andrew-lunarg Thanks. Fixed.

andrew-lunarg (approved):
Looks great to me.

Issue #646, opened 2022-02-11
wrong pointer is printed for output parameters


bradgrantham-lunarg:
toascii is printing the consumer output pointer, rather than the pointer that was stored in the capture file.

### Repro

By diffing the output of two runs of of the same build of `toascii` on the same capture file, a bunch of pointers show up with different values. The offending pointers include struct members as well as function output parameters. If their values came from the trace files they would not differ.

![Screenshot from 2022-08-30 15-56-03](https://user-images.githubusercontent.com/105252583/187473273-c45ac81a-7728-4386-b0b3-e2bab72211c4.png)

### Investigation

The `VulkanAsciiConsumer` process function for Vulkan calls using void data pointers plumbs through to this `ToString` function in `custom_vulkan_ascii_consumer.h`:

```
template <typename T>
inline std::string DataPointerDecoderToString(const T& pObj)
{
    // NOTE : Currently we just output the integer value of data void pointers, for
    //  fields that use a PointerDecoder<> we output the address of the decoder
    //  itself...this isn't terribly useful, but the capture time address wouldn't
    //  be a terribly useful thing anyway.  There are structures that have data
    //  fields that can be interpreted correctly...if we want to output interpreted
    //  void pointer data we can implement custom handlers for those structures.
    return "\"" + util::PtrToString(pObj) + "\"";
}
```
`const T& pObj` in this case is a reference to a const pointer to a `decode::PointerDecoder<unsigned char, unsigned char>`.   _Note, that's a lot of `NOTE`._

That calls through to:
```
template <typename PtrType>
inline std::string PtrToString(PtrType* ptr)
{
    std::stringstream strStrm;
    strStrm << "0x" << std::hex << reinterpret_cast<uintptr_t>(ptr);
    return strStrm.str();
}
```
This function then outputs the address of the `decode::PointerDecoder<unsigned char, unsigned char>` pointed to by `ptr` rather than any representation of the pointer from the trace file stored inside it.

Since the `PointerDecoder` instances whose addresses are output are sitting on the stack in the local frame of a `Decode_vk<SOME_FUNCTION>` function, there are lots of duplicates in a trace of addresses of objects which happen to to be at the same location on the stack at the time they are output.

andrew-lunarg:
Fixed by PR #787

PR #787, opened 2022-09-02
Output Addresses From Capture for void* and Opaque Struct* Parameters

andrew-lunarg:
Fixes issue #646

Adds three versions of DataPointerDecoderToString() making it clear what
types they service, each one converting the address from the capture
file for the pointer in question rather than using the address of an
object on the stack at decode time as the old template did.


framework/decode/custom_vulkan_ascii_consumer.h
```diff
@@ -80,18 +80,66 @@ inline std::string EnumPointerDecoderToString(PointerDecoder<VkEnumType>* pObj)
...
+/// `vkCmdSetCheckpointNV`, `IDirectFB*` from
+/// `vkGetPhysicalDeviceDirectFBPresentationSupportEXT`, `struct _screen_window*` from
+/// `kGetPhysicalDeviceScreenPresentationSupportQNX`, `Display*` from
+/// `vkGetPhysicalDeviceXlibPresentationSupportKHR`, and from
+/// `vkAcquireXlibDisplayEXT`, and from
+/// `vkGetRandROutputDisplayEXT`, `struct wl_display*` from
+/// `vkGetPhysicalDeviceWaylandPresentationSupportKHR`, `xcb_connection_t*` from
+/// `vkGetPhysicalDeviceXcbPresentationSupportKHR`, Win32 memory `HANDLE` from
+/// `vkGetMemoryWin32HandlePropertiesKHR`.
+/// The uint64_t is treated as an address in each case.
```
bradgrantham-lunarg, line 111: Really great and helpful comment, thanks

framework/decode/custom_vulkan_ascii_consumer.h
```diff
@@ -80,18 +80,66 @@ inline std::string EnumPointerDecoderToString(PointerDecoder<VkEnumType>* pObj)
...
+{
+    std::string str{ "\"" + util::PtrToString(buffer) + "\"" };
+    return str;
+}
+
+/// Convert pointers to arrays of void pointers to a string of the captured address.
+template <typename T, typename U>
+inline std::string DataPointerDecoderToString(PointerDecoder<T, U>* pDecoder)
+{
+    if (nullptr == pDecoder)
```
bradgrantham-lunarg, line 110: I suggest just a GFXRECON_ASSERT() here and don't bother returning a NULL.

framework/decode/custom_vulkan_ascii_consumer.h
```diff
@@ -80,16 +80,52 @@ inline std::string EnumPointerDecoderToString(PointerDecoder<VkEnumType>* pObj)
...
+        return GFXRECON_TOJSON_NULL;
+    }
+    std::string str{ "\"" + util::PtrToString(pDecoder->GetAddress()) + "\"" };
+    return str;
+}
+
+/// Convert void pointers into a string representation of the address in the capture file.
+inline std::string DataPointerDecoderToString(PointerDecoder<uint8_t>* pDecoder)
+{
+    if (nullptr == pDecoder)
```
bradgrantham-lunarg, line 122: Would this be called for `pData` parameter to `vkGetPipelineCacheData`?  That can legally be `nullptr` so this probably should just print "0x0" or some such for that case.  I wonder if you could just omit this `if` altogether.

bradgrantham-lunarg (changes requested):

framework/decode/custom_vulkan_ascii_consumer.h
andrew-lunarg: In that case the null would be returned from the `PointerDecoder<uint8_t>::GetAddress()`. `pDecoder` here should never be null, This parameter should really be a reference but the code generation delivers it as a pointer so I handle the null case just to be defensive.

framework/decode/custom_vulkan_ascii_consumer.h
andrew-lunarg: Done

framework/decode/custom_vulkan_ascii_consumer.h
andrew-lunarg: Done

bradgrantham-lunarg (approved):

PR #788, opened 2022-09-05
Documented Parallel Build via build.py

andrew-lunarg:
Documented use of `CMAKE_BUILD_PARALLEL_LEVEL` with our Python build script.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -105,6 +105,17 @@ can be run with the following options to disable the developer build steps:
 python scripts/build.py --skip-check-code-style --skip-tests
 ```
 
+The above invocation will run on a single core.
+To speed up the build with parallel processes,
+the standard CMake environment variable
+[`CMAKE_BUILD_PARALLEL_LEVEL`](https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html)
+may be used.
+For example, to use eight processes from within Bash:
+
+```bash
+CMAKE_BUILD_PARALLEL_LEVEL=8 python scripts/build.py --skip-check-code-style --skip-tests
+```
+
 Run the script with the `-h` option for additional usage information.
 
 ## Building for Windows


BUILD.md
```diff
@@ -105,6 +105,17 @@ can be run with the following options to disable the developer build steps:
 python scripts/build.py --skip-check-code-style --skip-tests
 ```
 
+The above invocation will run on a single core.
+To speed up the build with more processes,
+the standard CMake environment variable
+[`CMAKE_BUILD_PARALLEL_LEVEL`](https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html)
+may be used.
+For example, to use all hardware threads of a machine from within Bash:
```
bradgrantham-lunarg, line 113: The CMake docs say "If this variable is defined empty the native build tool's default number is used", which may not mean it uses all hardware threads.

If I use `CMAKE_BUILD_PARALLEL_LEVEL="8"`, then I get 8 jobs when I interrupt the build, which sounds like what I expect.

On my VM, which reports one physical core in /proc/cpuinfo, if I run with `CMAKE_PARALLEL_LEVEL=""`, I get what looks like all the objects for a given module compiled in parallel.  
```
grantham@grantham-VirtualBox:~/trees/gfxreconstruct$ cmake --build  build/ --target clean
grantham@grantham-VirtualBox:~/trees/gfxreconstruct$ CMAKE_BUILD_PARALLEL_LEVEL="" cmake --build  build/ 
[  1%] Building CXX object framework/util/CMakeFiles/gfxrecon_util.dir/argument_parser.cpp.o
[lots of "Building" lines, then I hit CTRL-C]
^Cmake[2]: *** [framework/util/CMakeFiles/gfxrecon_util.dir/build.make:63: framework/util/CMakeFiles/gfxrecon_util.dir/argument_parser.cpp.o] Interrupt
[22 other .o's are interrupted, implying they were all being built]
make: *** [Makefile:130: all] Interrupt
```
So 22 jobs were running in parallel.  That doesn't seem catastrophic, but I think it does look like `cmake` passed `-j` to `make` without a parameter, and `make` ran all the compiles at once, which could cause thrashing.  I lack the cmake-Fu to figure out what's going on there to verify that `-j` was passed without a parameter but it's not limited to the number of hardware threads in the machine.

I suggest if we're going to mention `CMAKE_BUILD_PARALLEL_LEVEL` then we should be sure we're saying what's actually happening which seems hard to glean from the CMake docs, or just be intentionally vague and say "use CMAKE_BUILD_PARALLEL_LEVEL and that's documented at https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html" and not give an example.

bradgrantham-lunarg (changes requested):

BUILD.md
andrew-lunarg: Good points. I think running builds in parallel is useful enough, and telling CMake to go parallel with a build from an environment variable rather than a commandline switch is obscure and non-obvious enough for it to be worth telling our users about this mechanism. I will make a change to get rid of the untrue statements I made about using all cores.

bradgrantham-lunarg: Yes, running builds in parallel is useful, I agree!  Thanks for documenting the environment variable!

bradgrantham-lunarg (approved):

Issue #683, opened 2022-03-29
The wrong value is printed for some input pointer parameters


bradgrantham-lunarg:
E.g. `pCode` in the `VkShaderModuleCreateInfo` passed to `vkCreateShaderModule`  appears to be printing the address of the buffer allocated at decode time to contain the code BLOB read from file.  It should print the original address at capture time.

### Investigation
`VkShaderModuleCreateInfo` provides a good target as it has custom ToString() code rather than generated. Below, I've modified `std::string ToString<decode::Decoded_VkShaderModuleCreateInfo>()` to output some options, starting with the current version which accesses the raw Vulkan struct's `pCode` pointer.
I believe the version we need is the second or third one, depending on whether a decimal number or hex-format number in a string is preferred, since the `address_` member of the base of the `StructPointerDecoder` that these access is decoded directly from the capture file.
```json
"[89]vkCreateShaderModule": {
    "return": "VK_SUCCESS",
    "device": "0x3",
    "pCreateInfo": {
        "sType": "VK_STRUCTURE_TYPE_SHADER_MODULE_CREATE_INFO",
        "pNext": null,
        "flags": 0,
        "codeSize": 1512,
        "pCode": "0x5555563230d0",
        "pCode": "0x6e129708f0",    <--- Choice A  [PtrToString(decoded_obj.pCode.GetAddress())]
        "pCode": 472758290672,      <--- Choice B [ToString(decoded_obj.pCode.GetAddress())]
        "pCode": "0x5555563230d0",  [PtrToString(decoded_obj.pCode.GetPointer())]
        "pCode": "0x0"              [PtrToString(decoded_obj.pCode.GetOutputPointer())]
    },
    "pAllocator": null,
    "[out]pShaderModule": "0x18"
},
```

andrew-lunarg:
Fixed by PR https://github.com/LunarG/gfxreconstruct/pull/789

PR #789, opened 2022-09-06
Output Addresses From Capture for void* and Opaque* Struct Members

andrew-lunarg:
Fixes issue #683

Adds versions of DataPointerDecoderToString(), each one converting
the address from the capture file for the pointer in question,
and plumbs them in during code generation for Vulkan struct members
that are void pointers or pointers to opaque structs like platform
window handles or external memory.


tools/CMakeLists.txt
```diff
@@ -6,7 +6,4 @@ add_subdirectory(optimize)
 add_subdirectory(capture)
 add_subdirectory(gfxrecon)
 
-# TOASCII build is currently disabled pending resolution of some issues
-if(BUILD_TOASCII)
 add_subdirectory(toascii)
-endif()
```
andrew-lunarg, line 12: I'm planning to back the commit with this out before merging.

framework/decode/custom_vulkan_ascii_consumer.h
```diff
@@ -103,20 +103,21 @@ inline std::string DataPointerDecoderToString(const uint64_t buffer)
...
-    std::string str{ "\"" + util::PtrToString(pDecoder->GetAddress()) + "\"" };
+    GFXRECON_ASSERT(nullptr != pBuffer &&
+                    "This should only ever come through as a pointer through an unavoidable weaknesses in "
+                    "the code gen: it should never be null.");
+    std::string str{ "\"" + util::PtrToString(*pBuffer) + "\"" };
     return str;
 }
 
-/// Convert void pointers into a string representation of the address in the capture file.
-inline std::string DataPointerDecoderToString(PointerDecoder<uint8_t>* pDecoder)
```
andrew-lunarg, line 117: I really liked having this concrete function to drop a breakpoint in during debugging and know exactly what types I was dealing with but the template version deals with this and also the `uint32_t`s pointed to by spirv code pointers, so the sensible choices were to lose the template and have concrete versions for all types or just have the template function to cover all PointerDecoder<> instantiations. Less code won.

PR #797, opened 2022-09-09
create and use CreateHardwareBufferCommandHeader2

### LINKED ISSUE #745: CreateHardwareBufferCommandHeader.usage is of the wrong type
`CreateHardwareBufferCommandHeader.usage` is declared as a `uint32_t` while `AHardwareBuffer_Desc.usage` is `uint64_t` (https://developer.android.com/ndk/reference/struct/a-hardware-buffer-desc)

Discovered while digging around our AHB code (haven't seen any case that this is causing any failures yet) and I'm not sure if this is a bug or done so on purpose. From what I see [in the documentation](https://developer.android.com/ndk/reference/group/a-hardware-buffer#group___a_hardware_buffer_1ga47c480435db00c8e88023eea28ccce90) the higher 32bit seem to be used for some non trivial use cases so might not be used that often/at all?

If this is a bug and we want to fix it then this might a be a good opportunity to enforce some compiler flag to catch such kind of errors (narrowing down conversions) 

While the fix should be trivial I guess it will break existing android captures (or apply some logic to detect the difference)

bradgrantham-lunarg:
Make a `CreateHardwareBufferCommandHeader2` with uint64_t, transition all encoding to the 2 version, and update `DispatchCreateHardwareBufferCommand` and `ProcessCreateHardwareBufferCommand` to take uint64_t usage, add an additional `FileProcessor` clause for the new `CreateHardwareBufferCommandHeader2`.

Fixes #745.

### DIFF
--- a/framework/decode/file_processor.cpp
+++ b/framework/decode/file_processor.cpp
@@ -628,7 +628,7 @@ bool FileProcessor::ProcessMetaData(const format::BlockHeader& block_header, for
             HandleBlockReadError(kErrorReadingBlockHeader, "Failed to read display message meta-data block header");
         }
     }
-    else if (meta_data_type == format::MetaDataType::kCreateHardwareBufferCommand)
+    else if (meta_data_type == format::MetaDataType::kCreateHardwareBufferCommand_deprecated)
     {
         format::CreateHardwareBufferCommandHeader_deprecated header;
 
@@ -699,7 +699,7 @@ bool FileProcessor::ProcessMetaData(const format::BlockHeader& block_header, for
                                  "Failed to read create hardware buffer meta-data block header");
         }
     }
-    else if (meta_data_type == format::MetaDataType::kCreateHardwareBufferCommand2)
+    else if (meta_data_type == format::MetaDataType::kCreateHardwareBufferCommand)
     {
         format::CreateHardwareBufferCommandHeader header;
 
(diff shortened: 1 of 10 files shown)


framework/format/format.h
```diff
@@ -312,6 +313,8 @@ struct ResizeWindowCommand2
     uint32_t         pre_transform;
 };
 
+// This command header's "usage" member is sized incorrectly.  Existing decoding must remain in place in order to
+// process captures that already exist.  Do not use this command header in any new code.
```
panos-lunarg, line 317: Just an idea with no strong opinion on adapting it:
Perhaps the _new_ should be named `CreateHardwareBufferCommandHeader` and the _existing_ one `CreateHardwareBufferCommandHeader_deprecated` or something along those lines?

bradgrantham-lunarg: Yes, I'll do that.

locke-lunarg (approved):

framework/decode/file_processor.cpp
```diff
@@ -628,6 +628,77 @@ bool FileProcessor::ProcessMetaData(const format::BlockHeader& block_header, for
             HandleBlockReadError(kErrorReadingBlockHeader, "Failed to read display message meta-data block header");
         }
     }
+    else if (meta_data_type == format::MetaDataType::kCreateHardwareBufferCommand_deprecated)
+    {
```
locke-lunarg, line 632: Could it add a wrong message for not using this type, the user should capture it again?

bradgrantham-lunarg: Yes, that's a good idea and I'll see about adding that!

bradgrantham-lunarg: Added directly to `dev` as 9e2f5ecd

PR #798, opened 2022-09-10
[Docs] Add Clang to Fedora dependencies

Evernow:
At least on my Fedora 36 machine I needed clang to be installed, without it I would get `No CMAKE_CXX_COMPILER could be found`

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -233,7 +233,7 @@ command:
 
 ```bash
 sudo dnf install git cmake libxcb-devel libX11-devel xcb-util-keysyms-devel \
-        libXrandr-devel wayland-devel zlib-devel lz4-devel libzstd-devel
+        libXrandr-devel wayland-devel zlib-devel lz4-devel libzstd-devel clang
 ```
 
 For arm64 builds (cross compilation):


BUILD.md
```diff
@@ -233,7 +233,7 @@ command:
 
 ```bash
 sudo dnf install git cmake libxcb-devel libX11-devel xcb-util-keysyms-devel \
-        libXrandr-devel wayland-devel zlib-devel lz4-devel libzstd-devel
+        libXrandr-devel wayland-devel zlib-devel lz4-devel libzstd-devel clang
```
andrew-lunarg, line 236: This is for `clang-format`? Poking around* it looks like the packaging on Fedora changed at version 32, meaning this wouldn't pull in `clang-format` for those versions. Are you on Fedora <= 31?

If the situation is as I am surmising, perhaps the dependency should be on the Fedora `git-clang-format` package which should pull in `clang-format` on all recent Fedora releases, or on `clang-tools-extra` which bundles it on versions from 32 onwards (i.e. including the current release).

Not running Fedora here so unable to test locally.

\* https://bugzilla.redhat.com/show_bug.cgi?id=1815913
   https://stackoverflow.com/a/51719891/506073

andrew-lunarg (dismissed):
.

BUILD.md
andrew-lunarg, line 236: This looks like the project is choosing clang over gcc explicitly. Is there no way to ask Fedora to install whatever the system's default compiler is as we do for Debian-derived distributions with the `build-essential` package?

bradgrantham-lunarg: https://fedoraproject.org/wiki/Releases/36/ChangeSet implies `gcc` is the default compiler.  https://unix.stackexchange.com/a/1344  says the command to install the development system is `sudo dnf groupinstall "Development Tools" "Development Libraries"`.  Although the source should compile under `clang` just fine, if there's a generic "install the compiler" target, we should require that, but if there isn't one, it looks like `gcc` is the compiler we should add here.

bradgrantham-lunarg: https://fedoraproject.org/wiki/Changes/CompilerPolicy#:~:text=The%20default%20compiler%20for%20Fedora,the%20default%20compiler%20is%20clang appears to have a stronger prescription that `gcc` is the default.

bradgrantham-lunarg:
After a workstation live CD install of Fedora 36, gcc is installed.  `sudo dnf install g++` allows cmake to progress past that message.  I'll add g++ to the installation dependencies.

bradgrantham-lunarg:
Hey, @Evernow , I want to apologize for being so abrupt.  I appreciate you posting this PR which prompted me to dig into Fedora a little further and install a fresh Fedora to see what the defaults are.  Thank you!  Please let me know if the change I made, which should now be in `dev` branch, works for you.

Evernow:
> Hey, @Evernow , I want to apologize for being so abrupt. I appreciate you posting this PR which prompted me to dig into Fedora a little further and install a fresh Fedora to see what the defaults are. Thank you! Please let me know if the change I made, which should now be in `dev` branch, works for you.

Don't worry one bit, I greatly appreciate you doing it, and it does work on my end.

Thank you!

andrew-lunarg:
I would say: "won't fix". The general requirements above include a C++14
compiler. If the platform has no clean way to add a package for the default
C++ compiler, we can rely on users to know they need a compiler and pick
their favourite.

Issue #802, opened 2022-09-13
Cyberpunk2077 Replay failed with VkImportMemory - vkAllocateMemory returned error value


illia-a-polishchuk:
Reproduced when I tried to replay Cyberpunk 2077 capture

- Mesa main branch with 6c0215eb09e commit
- Ubuntu 20.04.5 LTS
- Proton Experimental
- Intel® Core™ i5-10300H CPU @ 2.50GHz × 8
- Mesa Intel® UHD Graphics (CML GT2)

Steps to reproduce:
1. Download Cyberpunk 2077 from Steam
2. Enable Proton Experimental in Steam settings 
3. Build latest release mesa from main branch (the commit 6c0215eb09e)
4. export VK_ICD_FILENAMES=/home/illia/issues/cyberpunk-7211/mesa/build/share/vulkan/icd.d/intel_icd.x86_64.json
5. export VK_INSTANCE_LAYERS=VK_LAYER_LUNARG_gfxreconstruct
6. export VK_LAYER_PATH=/home/illia/sources/gfxreconstruct/build/layer/:$VK_LAYER_PATH
7. export GFXRECON_CAPTURE_FILE=/home/illia/traces/gfxrecon_capture.gfxr

Try to replay captured file:
`./gfxrecon-replay /home/illia/traces/gfxrecon_capture_20220913T144728.gfxr`

The error appears:
`[gfxrecon] FATAL - API call vkAllocateMemory returned error value VK_ERROR_INVALID_EXTERNAL_HANDLE that does not match the result from the capture file: VK_SUCCESS. Replay cannot continue.`

Vulkan validation error messages:

```
Objects: 1
        [0] 0x555cba6d96d0, type: 3, name: NULL
        VUID_Undefined(ERROR / SPEC): msgNum: 2044605652 - Validation Error: 
        [ VUID_Undefined ] Object 0: handle = 0x555cba6d96d0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x79de34d4 | 
            VkImportMemoryHostPointerInfoEXT: required parameter pHostPointer specified as NULL.
Objects: 1
        [0] 0x555cba6d96d0, type: 3, name: NULL
        VUID-VkMemoryAllocateInfo-pNext-03332(ERROR / SPEC): msgNum: 980427729 - Validation Error: 
        [ VUID-VkMemoryAllocateInfo-pNext-03332 ] Object 0: handle = 0x555cba6d96d0, 
[...]

PR #803, opened 2022-09-13
Add VkImportMemoryHostPointerInfoEXT replaying support

effolkronium:
Fixes issue: https://github.com/LunarG/gfxreconstruct/issues/802

Replay which use VkImportMemoryHostPointerInfoEXT fails with error
` [gfxrecon] FATAL - API call vkAllocateMemory returned error value VK_ERROR_INVALID_EXTERNAL_HANDLE that does not match the result from the capture file: VK_SUCCESS. Replay cannot continue.`

Vulkan validation layers output:
```
Objects: 1
        [0] 0x555cba6d96d0, type: 3, name: NULL
        VUID_Undefined(ERROR / SPEC): msgNum: 2044605652 - Validation Error: 
        [ VUID_Undefined ] Object 0: handle = 0x555cba6d96d0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x79de34d4 | 
            VkImportMemoryHostPointerInfoEXT: required parameter pHostPointer specified as NULL.
Objects: 1
        [0] 0x555cba6d96d0, type: 3, name: NULL
        VUID-VkMemoryAllocateInfo-pNext-03332(ERROR / SPEC): msgNum: 980427729 - Validation Error: 
        [ VUID-VkMemoryAllocateInfo-pNext-03332 ] Object 0: handle = 0x555cba6d96d0, 
        type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x3a7023d1 | 
        If the pNext chain includes a VkImportMemoryHostPointerInfoEXT structure, opaqueCaptureAddress must be zero. 
        The Vulkan spec states: If the pNext chain includes a VkImportMemoryHostPointerInfoEXT structure, 
        VkMemoryOpaqueCaptureAddressAllocateInfo::opaqueCaptureAddress must be zero 
        (https://vulkan.lunarg.com/doc/view/1.3.216.0/linux/1.3-extensions/vkspec.html#VUID-VkMemoryAllocateInfo-pNext-03332)
```

1 - Implemented allocation for external memory
2 - Resetted opaqueCaptureAddress to 0 if replay uses VkImportMemoryHostPointerInfoEXT

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -37,6 +37,7 @@
 #include "util/file_path.h"
 #include "util/hash.h"
 #include "util/platform.h"
+#include "util/logging.h"
 
 #include "generated/generated_vulkan_enum_to_string.h"
 
@@ -3209,8 +3210,14 @@ VkResult VulkanReplayConsumerBase::OverrideAllocateMemory(
 
         // Check if this allocation was captured with an opaque address
         bool                uses_address   = false;
+        bool                uses_import_memory = false;
         uint64_t            opaque_address = 0;
         VkBaseOutStructure* current_struct = reinterpret_cast<const VkBaseOutStructure*>(replay_allocate_info)->pNext;
+
+        size_t                                            host_pointer_size = 0;
+        std::unique_ptr<void, std::function<void(void*)>> external_memory_guard(
+            nullptr, [&](void* memory) { util::platform::FreeRawMemory(memory, host_pointer_size); });
+
         while (current_struct != nullptr)
         {
             if (current_struct->sType == VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_FLAGS_INFO)
@@ -3232,12 +3239,49 @@ VkResult VulkanReplayConsumerBase::OverrideAllocateMemory(
                 }
                 break;
             }
+
+            if (current_struct->sType == VK_STRUCTURE_TYPE_IMPORT_MEMORY_HOST_POINTER_INFO_EXT)
+            {
+                auto import_info = reinterpret_cast<VkImportMemoryHostPointerInfoEXT*>(current_struct);
+
+                GFXRECON_CHECK_CONVERSION_DATA_LOSS(size_t, replay_allocate_info->allocationSize);
+
+                size_t allocation_size = static_cast<size_t>(replay_allocate_info->allocationSize);
+
+                host_pointer_size =
+                    util::platform::GetAlignedSize(allocation_size, util::platform::GetSystemPageSize());
+
+                // VkAllocateMemory fails when memory was allocated with default malloc func, probably because of extra
+                // memory bytes allocated for malloc private info
+                import_info->pHostPointer = util::platform::AllocateRawMemory(host_pointer_size);
+
+                if (import_info->pHostPointer == nullptr)
+                {
+                    GFXRECON_LOG_ERROR("Failed to allocate raw memory with size = %" PRIuPTR " with error code: %u",
+                                       host_pointer_size,
+                                       util::platform::GetSystemLastErrorCode());
+                    std::abort();
+                }
+                external_memory_guard.reset(import_info->pHostPointer);
+
+                uses_import_memory = true;
+            }
+
             current_struct = current_struct->pNext;
         }
 
         if (uses_address)
         {
             // Insert VkMemoryOpaqueCaptureAddressAllocateInfo into front of pNext chain before allocating
+
+            // The Vulkan spec states: If the pNext chain includes a VkImportMemoryHostPointerInfoEXT structure,
+            // VkMemoryOpaqueCaptureAddressAllocateInfo::opaqueCaptureAddress must be zero
+            // (https://vulkan.lunarg.com/doc/view/1.3.216.0/linux/1.3-extensions/vkspec.html#VUID-VkMemoryAllocateInfo-pNext-03332)
+            if (uses_import_memory)
+            {
+                opaque_address = 0;
+            }
+
             VkMemoryAllocateInfo                     modified_allocate_info = (*replay_allocate_info);
             VkMemoryOpaqueCaptureAddressAllocateInfo address_info           = {
                 VK_STRUCTURE_TYPE_MEMORY_OPAQUE_CAPTURE_ADDRESS_ALLOCATE_INFO,
@@ -3271,6 +3315,12 @@ VkResult VulkanReplayConsumerBase::OverrideAllocateMemory(
             // When memory allocation fails at replay, but succeeded at capture, check for memory incompatibilities and
             // recommend enabling memory translation.
             allocator->ReportAllocateMemoryIncompatibility(replay_allocate_info);
+        }
+
+        if (result == VK_SUCCESS && uses_import_memory)
+        {
+            external_memory_.emplace(*replay_memory,
+                                     std::make_pair(external_memory_guard.release(), host_pointer_size));
         }
     }
     else
@@ -3400,6 +3450,13 @@ void VulkanReplayConsumerBase::OverrideFreeMemory(PFN_vkFreeMemory  func,
     {
         memory         = memory_info->handle;
         allocator_data = memory_info->allocator_data;
+
+        auto findIt = external_memory_.find(memory);
+        if (findIt != external_memory_.end())
+        {
+            util::platform::FreeRawMemory(findIt->second.first, findIt->second.second);
+            external_memory_.erase(findIt);
+        }
 
         memory_info->allocator_data = 0;
     }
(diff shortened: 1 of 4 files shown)


effolkronium:
Fixed formatting

effolkronium:
@bradgrantham-lunarg @panos-lunarg  Any feedback ?

bradgrantham-lunarg:
Hi, @effolkronium !  Thanks for your submission!

`PageGuardManager` is designed for use in capture for detecting changes to mapped pages for writing to the capture file.  I don't think it's the right thing to use in the replay consumer.  Can you do what you're doing to keep track of those allocations another way?  Like store them in the replay consumer class?  (I haven't had as much time to look into this as I'd like, so I'm sorry I can't provide a more specific suggestion.)

The "generated" files are written by `framework/generated/generate_vulkan.py`.  The next time that's run (typically on a new SDK version but also when the generated code needs to be altered) your changes will be overwritten.  If you still need to change the generated code after addressing the above, can you instead change the generators to add your change, or alternately pull it out of generation and make a custom handler?

Thanks again!

framework/util/platform.h
```diff
@@ -455,8 +490,51 @@ inline int32_t MakeDirectory(const char* filename)
...
+inline void* AllocateRawMemory(size_t aligned_size, bool use_write_watch = false)
+{
+    assert(aligned_size > 0);
+
+    if (aligned_size > 0)
+    {
+        void* memory = mmap(nullptr, aligned_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+
+        if (memory == MAP_FAILED)
+        {
```
andrew-lunarg, line 508: `GFXRECON_LOG_ERROR()` To get a message output before the segfault in the caller?

andrew-lunarg:
The build script will run the style checks. E.g., from the root: `CMAKE_BUILD_PARALLEL_LEVEL="16"  python3 ./scripts/build.py -a x64 -c debug`

framework/util/platform.h
effolkronium: I thought it's a caller responsibility now
But maybe better log here, but it's harder to inject logging because the logging.h depends on platform.h
Will try to handle this

andrew-lunarg: Okay, just a thought. maybe at the call sites then. ... even though if this fails, things are going very badly.

effolkronium: Yep, gdb will point to a failed line

Added extra check to the client side in replay_consumer

effolkronium:
Fixed formatting, please, run CI again
@bradgrantham-lunarg

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -3232,12 +3239,43 @@ VkResult VulkanReplayConsumerBase::OverrideAllocateMemory(
                 }
                 break;
             }
+
+            if (current_struct->sType == VK_STRUCTURE_TYPE_IMPORT_MEMORY_HOST_POINTER_INFO_EXT)
+            {
+                auto import_info = reinterpret_cast<VkImportMemoryHostPointerInfoEXT*>(current_struct);
+
+                host_pointer_size = util::platform::GetAlignedSize(replay_allocate_info->allocationSize,
```
panos-lunarg, line 3247: Why does the size have to rounded to multiples of the system's page size?

effolkronium: Hi @panos-lunarg 

It was implemented in the same way inside `VulkanCaptureManager::OverrideAllocateMemory` function 3 years ago
With comment explaining why:
```
// TODO: This should be aligned to minImportedHostPointerAlignment, but there is a currently a loader bug
// that prevents the layer from querying for that value when a 1.0 application does not explicitly enable
// physical_device_properties2.  For now we align to system page size.
```

Looks stable for now, but if you don't mind I can reimplement with minImportedHostPointerAlignment rounding, based on VK-GL-CTS examples. VK-GL-CTS uses `posix_memalign` api for that

framework/util/platform.h
```diff
@@ -50,6 +50,7 @@
 #include <sys/types.h>
 #include <unistd.h>
 #include <signal.h>
+#include <sys/mman.h>
```
panos-lunarg, line 53: I guess the same include from framework/util/page_guard_manager.cpp: 103 can be removed now

framework/decode/vulkan_replay_consumer_base.cpp
panos-lunarg: I see. makes sense

panos-lunarg (approved):
lgtm

framework/util/platform.h
effolkronium: Yep, removed it

effolkronium:
Resolved conflicts with the https://github.com/LunarG/gfxreconstruct/commit/9622292187b1a9dd3aeef38f0492f3d8bcda018c commit.

@panos-lunarg can it be merged to dev branch now ?

locke-lunarg (approved):

locke-lunarg:
@panos-lunarg  @effolkronium  I think it's ready to be merged.

bradgrantham-lunarg:
Give me a moment and I'll merge it.

bradgrantham-lunarg:
It looks like our internal CI never ran (oversight on our part) so I'm running it now.  If that passes, I'll merge.

bradgrantham-lunarg:
`replay_allocate_info->allocationSize` is `VkDeviceSize`, which is always `uint64_t`.  But on 32-bit Windows, at least, `size_t` is 32 bits, so this produces a warning and we take warnings as errors.  You can test the 32-bit build on your machine with `python3 scripts/build.py -c debug -a x86 --build-dir dbuild` (set `build-dir` as you desire).

Having an allocation that would exceed `size_t` would be invalid under https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImportMemoryHostPointerInfoEXT-handleType-01750, so I just added a check (which should compile out in Release) and added a cast to get rid of the warning (which we take as error).

effolkronium:
I see. Thanks for the check. Will do the x86 builds in case of another PR's from me : )

bradgrantham-lunarg:
> I see. Thanks for the check. Will do the x86 builds in case of another PR's from me : )

No problem.  Thanks very much for the PR!

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -3401,6 +3451,12 @@ void VulkanReplayConsumerBase::OverrideFreeMemory(PFN_vkFreeMemory  func,
         memory         = memory_info->handle;
         allocator_data = memory_info->allocator_data;
 
+        auto findIt = external_memory_.find(memory);
+        if (findIt != external_memory_.end())
+        {
+            util::platform::FreeRawMemory(findIt->second.first, findIt->second.second);
+        }
+
```
locke-lunarg, line 3459: Does it need to erase memory from external_memory_?

effolkronium: Well, didn't found any clear information about this in vulkan spec's

At least gfxreconstruct already do so in the vulkan_capture_manager.cpp VulkanCaptureManager::PostProcess_vkFreeMemory function

Also KhronosGroup VK-GL-CTS test erases the memory as well, so I think it's needed action

https://github.com/KhronosGroup/VK-GL-CTS/blob/60be5c346e5a63d45ca55a938e1025c0c5d32432/external/vulkancts/modules/vulkan/memory/vktMemoryExternalMemoryHostTests.cpp

```
ExternalMemoryHostBaseTestInstance::~ExternalMemoryHostBaseTestInstance (void)
{
	deAlignedFree(m_hostMemoryAlloc);
}
```

effolkronium: Done

effolkronium: I can squash all into a single commit if you'd like

locke-lunarg: Squash would be better. Thank you very much.

effolkronium: Done

locke-lunarg (approved):

bradgrantham-lunarg:
> CI gfxreconstruct build # 2064 failed.

I think we may have had a CI machine failure.  I'm running the case that failed again by hand.

bradgrantham-lunarg:
> CI gfxreconstruct build queued with queue ID 13346.

I'm reasonably sure your PR will pass if we can keep the CI machine running, but it will probably be tomorrow because a few other jobs were also backed up on that machine.  If it passes overnight I'll merge it tomorrow.

effolkronium:
💓

PR #805, opened 2022-09-14
just warn instead of fatal on different image index

bradgrantham-lunarg:
The message was warning but the function was LOG_FATAL.  This caused an otherwise potentially workable capture to fail.  Instead, just LOG_WARNING.


locke-lunarg:
VulkanCapturedSwapchain::AcquireNextImage2KHR need to be modified, too.

locke-lunarg (changes requested):
VulkanCapturedSwapchain::AcquireNextImage2KHR need to be modified, too.

bradgrantham-lunarg:
> VulkanCapturedSwapchain::AcquireNextImage2KHR need to be modified, too.

Good catch, thanks!  Fixed.

locke-lunarg (approved):

PR #813, opened 2022-09-23
Introducing a signal handler watcher thread

panos-lunarg:
Some applications install their own signal handlers. If this takes place after GFXR installs the signal handler that is part of the page_guard memory tracking method, then it effectively renders the mechanism useless and cause segmantation faults.

This patch introduces a which constantly monitors the active SIGSEG handler, and it is not the one installed by the page guard mechanism then replaces it back.

This feature is disabled by default and can be enabled by setting the appropriate enviroment variable, depending on the OS.


USAGE_desktop.md
```diff
@@ -177,6 +177,7 @@ Page Guard External Memory | GFXRECON_PAGE_GUARD_EXTERNAL_MEMORY | BOOL | When t
...
 Page guard unblock SIGSEGV | GFXRECON_PAGE_GUARD_UNBLOCK_SIGSEGV | BOOL | When the `page_guard` memory tracking mode is enabled and in the case that SIGSEGV has been marked as blocked in thread's signal mask, setting this enviroment variable to `true` will forcibly re-enable the signal in the thread's signal mask. Default is `false`
+Page guard signal handler watcher | GFXRECON_PAGE_GUARD_SIGNAL_HANDLER_WATCHER | BOOL | When the `page_guard` memory tracking mode is enabled, setting this enviroment variable to `true` will spawn a thread which will periodically make sure that the `SIGSEGV` handler is not replaced. Default is `false`
```
andrew-lunarg, line 180: > periodically make sure that the `SIGSEGV` handler is not replaced.

"will periodically reinstall the `SIGSEGV` if it has been replaced by the application being traced."

framework/util/page_guard_manager.h
```diff
@@ -225,9 +235,18 @@ class PageGuardManager
...
 
     // Only applies to WIN32 builds and Linux/Android builds with PAGE_GUARD_ENABLE_UCONTEXT_WRITE_DETECTION defined.
     const bool enable_read_write_same_page_;
+
+#if !defined(WIN32)
+    pthread_t    signal_handler_watcher_thread;
+    static void* signal_handler_watcher_f(void*);
+
+    // Signal watcher thread period in miliseconds
+    static constexpr int64_t signal_watcher_thread_period{ 100 };
```
andrew-lunarg, line 248: Should we default to 0ms (AKA spin)? 100ms is enough time to lose a whole frame. There is always going to be a race that generates rare failures, even if the monitor thread spins, but maybe minimising the duration by default is the best we can do?

framework/util/page_guard_manager.cpp
```diff
@@ -230,15 +235,58 @@ PageGuardManager::~PageGuardManager()
...
+
+        struct sigaction current_handler;
+        int              result = sigaction(SIGSEGV, nullptr, &current_handler);
+
+        assert(instance_);
+        if (result != -1 && instance_->exception_handler_ && current_handler.sa_sigaction != PageGuardExceptionHandler)
+        {
+            assert(instance_->exception_handler_count_);
+
+            GFXRECON_LOG_WARNING("Signal handler has been removed. Re-installing.")
```
andrew-lunarg, line 253: `GFXRECON_LOG_WARNING_ONCE`?

framework/util/page_guard_manager.cpp
```diff
@@ -230,15 +235,58 @@ PageGuardManager::~PageGuardManager()
...
+                                         signal_handler_watcher);
+
+#if !defined(WIN32)
+        if (signal_handler_watcher)
+        {
+            int ret =
+                pthread_create(&instance_->signal_handler_watcher_thread, nullptr, signal_handler_watcher_f, nullptr);
+            if (ret)
+            {
+                GFXRECON_LOG_WARNING("Page guard manager failed spawning thread (%s)", strerror(ret));
```
andrew-lunarg, line 286: `GFXRECON_LOG_ERROR`?

andrew-lunarg (commented):
Such a cool workaround.

framework/util/page_guard_manager.h
panos-lunarg: We had a similar discussion with Brad concerning the frequency the check is performed.
I guess it depends on how you look at it. The way I see it, for something that will only happen once or twice and most probably at init time, perhaps 100ms is too frequent. Fortnite seems to work with 100ms. We can make this configurable by env variables

framework/util/page_guard_manager.cpp
panos-lunarg: I think it it would be useful to know how often an application install its own handlers

USAGE_desktop.md
panos-lunarg: fixed, thanks!

framework/util/page_guard_manager.cpp
andrew-lunarg: True. Especially for us as the maintainers. If it is only at init time this is fine. If some other app does it every frame and the user has to lower the sleep period to catch it, this logging will slow gfxr down. I'm not sure if the user would then have to rebuild gfxr to kill the perf cost of the logging or whether a `grep -v` on the output would be enough.Anyway, this is just a comment.

framework/util/page_guard_manager.h
```diff
@@ -218,16 +228,31 @@ class PageGuardManager
...
     const bool               enable_copy_on_map_;
     const bool               enable_separate_read_;
     const bool               unblock_sigsegv_;
+    bool                     enable_signal_handler_watcher_;
 
     // Only applies to WIN32 builds and Linux/Android builds with PAGE_GUARD_ENABLE_UCONTEXT_WRITE_DETECTION defined.
     const bool enable_read_write_same_page_;
+
+#if !defined(WIN32)
+    pthread_t signal_handler_watcher_thread;
```
bradgrantham-lunarg, line 245: I think these members should have `_` appended to be consistent with the other members.

framework/util/page_guard_manager.cpp
```diff
@@ -230,15 +233,102 @@ PageGuardManager::~PageGuardManager()
...
 
+#if !defined(WIN32)
+constexpr uint32_t PageGuardManager::signal_handler_watcher_max_restores;
+uint32_t           PageGuardManager::signal_handler_watcher_restores = 0;
+
+void PageGuardManager::MarkAllTrackedMemoryAsDirty()
+{
+    std::lock_guard<std::mutex> lock(tracked_memory_lock_);
+
+    for (auto entry = memory_info_.begin(); entry != memory_info_.end(); ++entry)
```
bradgrantham-lunarg, line 244: Could this be a range-for?  (e.g. `for (auto& entry : memory_info_) { ... }` and then you'd change `entry->second` to `entry.second` )

framework/util/page_guard_manager.cpp
```diff
@@ -230,15 +233,102 @@ PageGuardManager::~PageGuardManager()
...
+
+    return false;
+}
+
+void* PageGuardManager::SignalHandlerWatcher(void* args)
+{
+    while (instance_->enable_signal_handler_watcher_ &&
+           signal_handler_watcher_restores < signal_handler_watcher_max_restores)
+    {
+        signal_handler_watcher_restores += CheckSignalHandler();
```
bradgrantham-lunarg, line 301: I know `true` will evaluate to `1` here, but for clarity if `signal_handler_watcher_restores` is a `uint32_t` then I'd prefer `CheckSignalHandler` to return a `uint32_t` or do `? 1 : 0` here.

bradgrantham-lunarg (approved):

framework/util/page_guard_manager.cpp
andrew-lunarg: For my own learning of the house style, would the most explicit option also be the most preferred? Like this:
```cpp
if(true == CheckSignalHandler())
{
   ++signal_handler_watcher_restores;
}
```

bradgrantham-lunarg: That would work, too, but I don't feel like comparing a value with `true` is required.

bradgrantham-lunarg (approved):

bradgrantham-lunarg (approved):

PR #830, opened 2022-09-30
Toascii - Header Line with source, version, commit.

### LINKED ISSUE #809: toascii - Add a Metadata Header to JSON Lines Format
It would be useful to have some metadata in every JSON file output.

### Fields to Include
* Version Number [_of JSON Schema?_]
* Commit-id tool was built against 
  * Also a bool for uncommitted changes? (Only of interest to gfxr devs)
* Source capture filename

### Maybes
* is_trimmed: bool [Requires seeking back to fix it]
* Frame range captured [Do we have this info soon enough?]
* Number of lines [optional, requires seeking back and leaving ugly reserved spaces ]
  * ! It probably can't be added at the front of the file until the whole file is generated so perhaps not worthwhile.
* List of functions in capture [Requires long insert at front of file, so not great ]

### Laters
* Set of flags passed to tool (any command-line switches and env vars affecting output). [_There are currently none._]

### Metadata Footer
Is it worth gathering the data we only pick up during processing and presenting it in a footer?

andrew-lunarg:
Toascii - Header Line with source, version, commit.
    
Fixes #809

Toascii is disabled in the build, but this PR previously passed CI with it temporarily enabled. 

The headers look like:

```json
    {
      "header": {
        "source-path": "<long string>.gfxr",
        "json-version": "0.8.0",
        "gfxrecon-version": "0.9.15-dev (branch:c13c3fe*)",
        "vulkan-version": "1.3.224"
      }
    }
```

### DIFF
--- a/tools/toascii/main.cpp
+++ b/tools/toascii/main.cpp
@@ -1,6 +1,6 @@
 /*
-** Copyright (c) 2018-2020 Valve Corporation
-** Copyright (c) 2018-2020 LunarG, Inc.
+** Copyright (c) 2018-2022 Valve Corporation
+** Copyright (c) 2018-2022 LunarG, Inc.
 **
 ** Permission is hereby granted, free of charge, to any person obtaining a
 ** copy of this software and associated documentation files (the "Software"),
@@ -22,7 +22,6 @@
 */
 
 #include "project_version.h"
-
 #include "tool_settings.h"
 #include "format/format.h"
 #include "generated/generated_vulkan_ascii_consumer.h"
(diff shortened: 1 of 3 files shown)


tools/toascii/main.cpp
```diff
@@ -57,8 +56,10 @@ static void PrintUsage(const char* exe_name)
 #endif
 }
 
-static std::string GetOutputFileName(const gfxrecon::util::ArgumentParser& arg_parser,
-                                     const std::string&                    input_filename)
+namespace
```
bradgrantham-lunarg, line 59: Why did you add this namespace around this function?

bradgrantham-lunarg (changes requested):

tools/toascii/main.cpp
bradgrantham-lunarg: TIL about anonymous namespace in a compilation unit!  It will be the first time it's used in any of our tools, but I suppose there's no harm!

bradgrantham-lunarg (approved):

tools/toascii/main.cpp
andrew-lunarg: This was leftover from when I had my path split function in here. I can follow-up with a tiny PR to put things back to `static` or widen the anon namespace up to cover the constants and other static linkage function depending on your preference.
TIL that there was a backlash against the deprecation of this use of static in favour of anon namespaces and that the current consensus is _use whichever your prefer as we'll never be taking this use of static out of the standard due to C compatibility and too much old C++ using it_.

PR #838, opened 2022-10-07
Bug surface format & AHardwareBuffer

locke-lunarg:


framework/encode/vulkan_state_writer.cpp
```diff
@@ -1062,10 +1062,16 @@ void VulkanStateWriter::ProcessHardwareBuffer(format::HandleId memory_id,
...
         WriteCreateHardwareBufferCmd(memory_id, hardware_buffer, plane_info);
-        WriteFillMemoryCmd(memory_id, 0, allocation_size, data);
+
+        if (data != nullptr)
+        {
+            WriteFillMemoryCmd(memory_id, 0, allocation_size, data);
+        }
+        else
+        {
+            GFXRECON_LOG_WARNING("AHardwareBuffer_unlock warning: data is nullptr");
```
bradgrantham-lunarg, line 1073: Should this warn about `AHardwareBuffer_lock`, since I think that's responsible for setting `data`?

bradgrantham-lunarg (approved):
Looks great other than my question about the print!

framework/encode/vulkan_state_writer.cpp
bradgrantham-lunarg: That is to say, this LOG_WARNING mentions `..._unlock`, but the function call is `..._lock`.

locke-lunarg: Honestly, I have no idea what the good log is here. Could you write that for me? Thanks.

bradgrantham-lunarg (approved):

PR #848, opened 2022-10-13
Always get swapchain images

panos-lunarg:


panos-lunarg:
Turns out this is a bad idea after all

Issue #847, opened 2022-10-12
convert/toascii needs to escape special characters in strings


mikes-lunarg:
I noticed newline characters in `VkPhysicalDeviceDriverProperties::driverInfo` are making it into the jsonlines output, breaking up the json entry into multiple lines and confusing jq. 

```
{"index":8,"vkFunc":{"name":"vkGetPhysicalDeviceProperties2KHR","args":{"physicalDevice":2,"pProperties":{"sType":"VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2","pNext":{"sType":"VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SUBGROUP_PROPERTIES","pNext":{"sType":"VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DRIVER_PROPERTIES","pNext":{"sType":"VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DEPTH_STENCIL_RESOLVE_PROPERTIES","pNext":{"sType":"VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VERTEX_ATTRIBUTE_DIVISOR_PROPERTIES_EXT","pNext":null,"maxVertexAttribDivisor":65535},"supportedDepthResolveModes":1,"supportedStencilResolveModes":1,"independentResolveNone":0,"independentResolve":0},"driverID":"VK_DRIVER_ID_QUALCOMM_PROPRIETARY","driverName":"Qualcomm Technologies Inc. Adreno Vulkan Driver","driverInfo":"Driver Build: 85da404, I46ff5fc46f, 1606794520
Date: 11/30/20
Compiler Version: EV031.31.04.01
Driver Branch: promo490_3_Google
[...]

PR #850, opened 2022-10-13
JSON String Escaping

andrew-lunarg:
JSON String Escaping
    
Fixes Issue #847

Made JSONEscape() and unit test.
Centralised C-String conversions in one function CStrToString() which
calls the escape function.
Plumbed it in to generated and custom code.


bradgrantham-lunarg (approved):
Assuming it passes CI

PR #855, opened 2022-10-14
Lazily resize tracked swapchain images vector

panos-lunarg:
Calling vkAcquireNextImage before vkGetSwapchainImages is a valid use
case but was not being handled properly when capturing (tracking)
outside of a trimmed range.

This commit fixes this issue for capturing and replaying


panos-lunarg:
Capturing and replaying antutu's Terracotta trimmed was failing on Pixel 6 and S22. This should fix capturing for both and replaying to the Pixel 6. S22 failure to replay seems to be because of other unrelated reasons.

bradgrantham-lunarg (approved):

PR #859, opened 2022-10-15
scripts: add keyframes.py script

lunarpapillo:
This script is based upon the script of the same name that we have been using internally to process capture files for testing use.  This version has had several improvements added, to simplify and automate newly identified capture processing requirements.  All such improvements are controlled by command-line arguments, so the script should behave in a backward-compatible way if the new command-line arguments are not used.

The improved keyframes.py script differs in these ways from its predecessor:

- improved documentation
- gfxrecon-convert tool instead of gfxrecon-toascii tool
- support a .keyframesrc configuration file for user-customized command line arguments
- explicitly pass text encoding (to satisfy modern pylint)
- allow creation of a command_counts.json file during capture file processing
- allow creation of stub reference images during capture file processing

I've also added a .pylintrc file.  It's typical for Python development to use such a file to encapsulate script style preferences.  This fairly standard instance is intended to be used as a baseline; as the project makes other style decisions, the .pylintrc can be updated to match.

### DIFF
--- a/scripts/.pylintrc
+++ b/scripts/.pylintrc
@@ -1,0 +1,89 @@
+[REFACTORING]
+
+# Maximum number of nested blocks for function / method body
+max-nested-blocks=10
+
+# Complete name of functions that never returns. When checking for
+# inconsistent-return-statements if a never returning function is called then
+# it will be considered as an explicit return statement and no message will be
+# printed.
+never-returning-functions=optparse.Values,sys.exit
+
+[BASIC]
+
+# Naming style matching correct argument names
+argument-naming-style=any
+
+# Naming style matching correct attribute names
+attr-naming-style=any
+
+# Naming style matching correct class attribute names
+class-attribute-naming-style=any
+
+# Naming style matching correct class names
+class-naming-style=PascalCase
+
+# Naming style matching correct function names
+function-naming-style=any
+
+# Naming style matching correct method names
+method-naming-style=camelCase
+
+# Naming style matching correct module names
+module-naming-style=any
+
+# Naming style matching correct variable names
+variable-naming-style=snake_case
+# snake_case complains about single-word entries; this fixes that.
+# also want to specifically allow things like env_HOME
+variable-rgx=^[a-zA-Z][a-zA-Z0-9]*((_[a-zA-Z0-9]+)*)?$
+
+# Colon-delimited sets of names that determine each other's naming style when
+# the name regexes allow several styles.
+name-group=argument:attr:class-attribute:function:module
+
+# Good variable names which should always be accepted, separated by a comma
+good-names=i,j,k,e,ex,Run,_,log
+
+[FORMAT]
+
+# Maximum number of characters on a single line.
+max-line-length=120
+
+# Maximum number of lines in a module
+max-module-lines=2000
+
+[DESIGN]
+
+# Maximum number of arguments for function / method
+max-args=10
+
+# Maximum number of attributes for a class (see R0902).
+max-attributes=20
+
+# Maximum number of boolean expressions in a if statement
+max-bool-expr=10
+
+# Maximum number of branch for function / method body
+max-branches=30
+
+# Maximum number of locals for function / method body
+max-locals=50
+
+# Maximum number of parents for a class (see R0901).
+max-parents=7
+
+# Maximum number of public methods for a class (see R0904).
+max-public-methods=50
+
+# Maximum number of return / yield for function / method body
+max-returns=20
+
+# Maximum number of statements in function / method body
+max-statements=1000
+
+# Minimum number of public methods for a class (see R0903).
+# Permit the use of a class as a simple data container
+# with no public methods (over some of the other suggested
+# alternatives)
+min-public-methods=0
--- a/scripts/keyframes.py
+++ b/scripts/keyframes.py
@@ -1,0 +1,422 @@
+#!/usr/bin/env python3
+"""Given a gfxreconstruct command file, identify "interesting"
+keyframes based on distinct maximal function sets used in a frame.  Keyframes
+selected in this way tend to exercise more distinct parts of the trace,
+as many frames will use the same "footprint" of functions as other frames;
+by collapsing to maximal function sets, keyframes can be selected that more
+thoroughly exercise the graphics API.
+
+The selected keyframes can be further refined through the use of filter
+operations.
+
+To use on a trace captured by gfxreconstruct, the trace must be converted
+to an ASCII command file containing one function or method name per line, e.g.:
+
+    ...
+    vkDestroyImageView
+    vkBeginCommandBuffer
+    vkCmdPipelineBarrier
+    vkCmdWriteTimestamp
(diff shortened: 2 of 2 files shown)


scripts/keyframes.py
```diff
@@ -0,0 +1,422 @@
+#!/usr/bin/env python3
+"""Given a gfxreconstruct command file, identify "interesting"
```
andrew-lunarg, line 2: Not sure a "gfxreconstruct command file" is a thing that exists. On first pass this script seemed completely independent of GFXReconstruct as a whole and gfxrecon-convert in particular.
Shouldn't this read, _"Given a text file of Vulkan function names, one per line, in order of execution, ..."_?

scripts/keyframes.py
```diff
@@ -0,0 +1,422 @@
...
+    vkCmdEndRenderPass
+    vkCmdResetQueryPool
+    vkEndCommandBuffer
+    vkFlushMappedMemoryRanges
+    vkResetFences
+    vkQueueSubmit
+    vkQueuePresentKHR
+    ...
+
+The command file must include frame marker functions (e.g. vkQueuePresentKHR
```
andrew-lunarg, line 39: I note we have lost the information on where images are being presented (a gui 3d modelling app may present to lots of surfaces in odd orders and we only care about the main viewport presents for example). Perhaps we should use annotations or a metacommand to emit frame markers that are converted to json and then preserved in the text file input to this script. Not a priority: this way is fine for typical games.

scripts/keyframes.py
```diff
@@ -0,0 +1,422 @@
...
+for Vulkan); if it does not, no frame boundaries can be established, and keyframes
+cannot be calculated.
+
+The gfxrecon-convert tool can be used to create such a file from a trace
+file, using an editor or a script to extract only the function or method
+name associated with each captured call.
```
andrew-lunarg, line 45: It might be useful to give the sed one-liner that does that.
```
gfxrecon-convert --output stdout vkcube.f1.gfxr | sed "s/.*\"name\":\"vk\([^\"]*\).*/vk\1/" > commandfile.txt
python3 keyframes.py commandfile.txt
```
Assuming this script can read from stdin:
```
gfxrecon-convert --output stdout vkcube.f1.gfxr | sed "s/.*\"name\":\"vk\([^\"]*\).*/vk\1/" | python3 keyframes.py
```

scripts/keyframes.py
```diff
@@ -0,0 +1,422 @@
...
+    $ python3 keyframes.py commandfile.txt
+    INFO: processing the command file commandfile.txt...
+    INFO: sorting 70 function sets by cardinality...
+    INFO: collapsing into maximal distinct sets...
+    INFO: filtering 10 keyframes from 22 maximal sets...
+    INFO: Keyframes in priority order: 0,1908,1051,1910,2086,4280,5329,1616,2088,4281
+    INFO: Keyframes in sorted order:   0,1051,1616,1908,1910,2086,2088,4280,4281,5329
+
+This script can also create a command_counts.json file in the same directory, and
+create stub reference images, each of which is helpful in trace processing.
```
andrew-lunarg, line 58: It isn't clear to me what a stub reference image is or why it is useful.

scripts/keyframes.py
```diff
@@ -0,0 +1,422 @@
...
+    # have the count we want or until we run out of footprints.
+    keyframes = getKeyframes(maximal_footprints, footprints_to_frames, args.n, filters)
+
+    # Give our output.
+    log.info("Keyframes in priority order: %s", ",".join([str(x) for x in keyframes]))
+    keyframes.sort()
+    log.info("Keyframes in sorted order:   %s", ",".join([str(x) for x in keyframes]))
+
+    # Create stub reference images if requested.  The stub reference images essentially
+    # encode the keyframe numbers, which can be used with a normal CI run to generate the
```
andrew-lunarg, line 415: Seems very coupled to details of internal CI processes.

scripts/keyframes.py
```diff
@@ -0,0 +1,422 @@
...
+
+    filter_descriptions = [f'{x.name} ({x.description})' for x in Filter.__subclasses__()]
+    filter_help = ('comma-separated list of keyframe filters to use [none]; available filters are: '
+                   + '; '.join(filter_descriptions))
+    common.add_argument('-f', '--filters', help=filter_help)
+
+    common.add_argument('--stub-reference-images', metavar='DIRECTORY',
+                        help='if set, names a directory where stub reference images will be created')
+
+    common.add_argument('--save-command-counts', metavar='JSONFILE',
```
andrew-lunarg, line 349: Thanks!

andrew-lunarg (commented):
I mostly took on faith that the Python works. I'm not up to date on features past about Python 1.5. Looks fine to me.

scripts/.pylintrc
```diff
@@ -0,0 +1,89 @@
+[REFACTORING]
```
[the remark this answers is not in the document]
lunarpapillo: I'd be happy to do so as a second commit.  They do not pass it currently.

scripts/keyframes.py
lunarpapillo: That's a better description, thanks.

scripts/keyframes.py
lunarpapillo: I note that `gfxrecon-replay` has a `--surface-index` switch.  Perhaps `gfxrecon-convert` should also have a similar switch, or perhaps the surface index should be part of the `gfxrecon-convert` output JSON.

scripts/keyframes.py
lunarpapillo: My ultimate goal here is to automatically create and parse the output from `gfxrecon-convert`; I'm saving that for a second PR after/if this one is accepted.  But agreed that this revision could have a helpful command line.

scripts/keyframes.py
lunarpapillo: Added more commentary.

scripts/keyframes.py
lunarpapillo: This whole script is tied to internal CI processes - the original point of it was to let an outside vendor create packages that we could "just use" in internal CI.

I'm not at all against making this an internal script - we used such an internal-only script for a long time before a need came for outside vendors to process traces like we do.  If we're going to support outside contributions of test workloads, it'd be nice if we used the same process that we advertise.

scripts/keyframes.py
lunarpapillo: This was actually a driving reason.  We want command count files for every trace; they're useful for developer analysis, and I also want to use them for `gfxrecon-convert` testing.  So I definitely want any script that an outside vendor might use to contribute a workload to include this in its basic processing.

andrew-lunarg:
I see this has conflicts now. Are reviews still wanted? (sorry they have not been forthcoming previously)

bradgrantham-lunarg:
I think this is coupled enough to our internal processes that this should be part of our internal suite.  We can hand this out from our internal tool suite to anyone who we are asking to engage in our internal CI process.

PR #861, opened 2022-10-18
Hush compiler warning

panos-lunarg:


panos-lunarg:
Compiler is complaining about that line when building for Android (clang)

framework/util/to_string.h
```diff
@@ -240,7 +240,7 @@ inline void JSONEscape(const char* cstr, std::string& escaped)
     if (cstr)
     {
         char c;
-        while (c = *cstr++)
+        while ((c = *cstr++))
```
bradgrantham-lunarg, line 243: How about moving `char c` inside the while, like `while(char c = *str++)`?  That doesn't give me a warning for clang in Compiler Explorer and it moves `c` into a tighter scope.

bradgrantham-lunarg (approved):

framework/util/to_string.h
panos-lunarg: sounds good

PR #862, opened 2022-10-18
Flush After Presents

andrew-lunarg:
**Try the approach in Issue #876 before falling back to something like this**
If 876 is successful, this still might be useful but only to flush the last frame to account for application hard crashes between frame end and next frame start.

Flush after presents to avoid capture files with incomplete final blocks. The last block in a capture is often a present since the logic to quit is evaluated between a present and the next Vulkan call that would be made. By selctively flushing after presents we can take care of these cases without the cost (if any) of the option to flush after every block is written.

The warning seen when running many captures through `convert`, `compress`, and `optimize` is:

    [gfxrecon] WARNING - Incomplete block at end of file

It might be worth investigating whether something like this was also needed to get data all the way out of even OS disk cache to storage on clean shutdown:
```cpp
inline int32_t FileClose(FILE* stream)
{
#if defined(__USE_POSIX)
    fflush(stream);
    fsync(fileno(stream));
#endif
    return fclose(stream);
}
```

### DIFF
--- a/framework/encode/capture_manager.cpp
+++ b/framework/encode/capture_manager.cpp
@@ -444,6 +444,13 @@ void CaptureManager::EndApiCallCapture()
             WriteToFile(parameter_buffer->GetHeaderData(),
                         parameter_buffer->GetHeaderDataSize() + parameter_buffer->GetDataSize());
         }
+
+        // Flush after presents to avoid capture files with incomplete final
+        // blocks.
+        if (thread_data->call_id_ == format::ApiCall_vkQueuePresentKHR)
+        {
+            file_stream_->Flush();
+        }
     }
 }
 


bradgrantham-lunarg:
> It might be worth investigating whether something like this was also needed to get data all the way out of even OS disk cache to storage on clean shutdown:

This shouldn't be necessary.  `fwrite` does have a local buffer (see https://man7.org/linux/man-pages/man0/stdio.h.0p.html) but that buffer is flushed on normal exit (see https://www.gnu.org/software/libc/manual/html_node/Flushing-Buffers.html) and should then be the operating system's I/O cache.  If the data is in the operating system's hands, the data will be written to the file unless the system itself has a hard crash.

So this PR is really a solution for getting data into the operating system's hands in the case of an abnormal shutdown like killing through Task Manager or an unhandled signal on Linux or Android.

I think it's good, and as we discussed we just need a little measurement to make sure the performance characteristics are as we expect.

andrew-lunarg:
Food for thought! I'm still concerned about whether the stdlib doing the
exit for the app is the same std lib we are linked against. If not, how's
it going to know to close our files? Most extreme version off the top of my
head, app is a windows game running through steam proton on linux. we live
on the linux side. Less extreme, app static links a std lib and we dynamic
link the system std lib. I think the safest bet is to go belt and braces:
flush, maybe fsync, fclose, both our files through any shutdown mechanism
we can trap.

bradgrantham-lunarg:
As noted in the other issue, all instances of libc will make an `atexit` to close all files and that flushes all files.  I'm not worried about it in the slightest.  I think this PR's content, flush at Present is a good idea.

Issue #585, opened 2021-09-29
gfxrecon-info produces no helpful information for captures that don't present a frame


lunarpapillo:
This was observed on a Linux Mesa AMD machine using a trace of `vulkaninfo`, which does Vulkan calls but does not produce any Vulkan output.  I imagine the problem could also occur with a compute-only trace.
```
$ gfxrecon-info --version
gfxrecon-info version info:
  GFXReconstruct Version 0.9.9-dev (dev:7f6b3c6)
  Vulkan Header Version 1.2.189
$ gfxrecon-info /media/nas/smb_traces/vulkan/GFXR//windows-nvidia/vulkaninfo/vulkaninfo_20201110T115442.gfxr
[gfxrecon] WARNING - Pointer decoder's external memory capacity (8) is smaller than the decoded array size (16); data will be truncated
File did not contain any frames
```
I expected all the information about the file that's normally available to output traces, e.g.:
```
$ gfxrecon-info /media/nas/smb_traces/vulkan/GFXR//windows-nvidia/baldursgate3/baldursgate3_20201023T150706.gfxr 
[gfxrecon] WARNING - Incomplete block at end of file
File info:
	Compression format: LZ4
	Total frames: 8855

Application info:
	Application name: Baldur's Gate 3
	Application version: 1096026085
	Engine name: The Divinity Engine
	Engine version: 1096026085
	Target API version: 4198400 (1.1.0)

Physical device info:
	Device name: GeForce RTX 2080 SUPER
	Device ID: 0x1e81
	Vendor ID: 0x10de
	Driver version: 1913765888 (0x7211c000)
	API version: 4202638 (1.2.142)

Device memory allocation info:
	Total allocations: 80
	Min allocation size: 256
	Max allocation size: 877050880

Pipeline info:
	Total graphics pipelines: 5917
	Total compute pipelines: 27
```
## Other nits
In addition to `Total frames`, the tool should also output `Total commands` (which is more interesting for traces that have no identifiable frames)

bradgrantham-lunarg:
Agreed, replay should print out information for vulkaninfo and compute-only captures just like visual captures.  However, is it possible this particular capture showing the same issue as https://github.com/LunarG/gfxreconstruct/issues/584 ?  I ask because it outputs 
`[gfxrecon] WARNING - Pointer decoder's external memory capacity (8) is smaller than the decoded array size (16); data will be truncated`

bradgrantham-lunarg:
[...]

PR #864, opened 2022-10-19
Print gfxr info even if frame number is 0

locke-lunarg:
https://github.com/LunarG/gfxreconstruct/issues/585
sample: `lunarg@tcubumes:/media/nas/smb_traces/vulkan/gfxr/windows-nvidia/vulkaninfo-new/vulkaninfo_20211005T094235.gfxr`
```
Info:
File info:
        Compression format: LZ4
        Total frames: 0

Application info:
        Application name: vulkaninfo
        Application version: 1
        Engine name:
        Engine version: 0
        Target API version: 4194304 (1.0.0)

Physical device info:
        Device name: GeForce RTX 2070
        Device ID: 0x1f07
        Vendor ID: 0x10de
        Driver version: 1917288448 (0x72478000)
        API version: 4202638 (1.2.142)

        Total allocations: 0
        Min allocation size: 0
        Max allocation size: 0

Pipeline info:
        Total graphics pipelines: 0
        Total compute pipelines: 0

File did not contain any frames
```
The engine name and engine version is null and 0 in the captured file. It looks no problem to print the info even if the frame number is 0. I don't know why the original code requires frame number has to be larger than 0.

### DIFF
--- a/tools/info/main.cpp
+++ b/tools/info/main.cpp
@@ -536,8 +536,7 @@ int main(int argc, const char** argv)
         file_processor.AddDecoder(&decoder);
         file_processor.ProcessAllFrames();
 
-        if ((file_processor.GetCurrentFrameNumber() > 0) &&
-            (file_processor.GetErrorState() == gfxrecon::decode::FileProcessor::kErrorNone))
+        if (file_processor.GetErrorState() == gfxrecon::decode::FileProcessor::kErrorNone)
         {
             GFXRECON_WRITE_CONSOLE("File info:");
 
@@ -618,10 +617,15 @@ int main(int argc, const char** argv)
                 }
             }
 
+            auto alocation_count = stats_consumer.GetAllocationCount();
             GFXRECON_WRITE_CONSOLE("\nDevice memory allocation info:");
-            GFXRECON_WRITE_CONSOLE("\tTotal allocations: %" PRIu64, stats_consumer.GetAllocationCount());
-            GFXRECON_WRITE_CONSOLE("\tMin allocation size: %" PRIu64, stats_consumer.GetMinAllocationSize());
-            GFXRECON_WRITE_CONSOLE("\tMax allocation size: %" PRIu64, stats_consumer.GetMaxAllocationSize());
+            GFXRECON_WRITE_CONSOLE("\tTotal allocations: %" PRIu64, alocation_count);
+
+            if (alocation_count > 0)
+            {
+                GFXRECON_WRITE_CONSOLE("\tMin allocation size: %" PRIu64, stats_consumer.GetMinAllocationSize());
+                GFXRECON_WRITE_CONSOLE("\tMax allocation size: %" PRIu64, stats_consumer.GetMaxAllocationSize());
+            }
 
             GFXRECON_WRITE_CONSOLE("\nPipeline info:");
             GFXRECON_WRITE_CONSOLE("\tTotal graphics pipelines: %" PRIu64, stats_consumer.GetGraphicsPipelineCount());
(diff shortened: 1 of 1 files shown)


tools/info/main.cpp
```diff
@@ -620,7 +619,10 @@ int main(int argc, const char** argv)
 
             GFXRECON_WRITE_CONSOLE("\nDevice memory allocation info:");
             GFXRECON_WRITE_CONSOLE("\tTotal allocations: %" PRIu64, stats_consumer.GetAllocationCount());
-            GFXRECON_WRITE_CONSOLE("\tMin allocation size: %" PRIu64, stats_consumer.GetMinAllocationSize());
+            GFXRECON_WRITE_CONSOLE("\tMin allocation size: %" PRIu64,
+                                   (stats_consumer.GetMinAllocationSize() == std::numeric_limits<uint64_t>::max()
+                                        ? 0
+                                        : stats_consumer.GetMinAllocationSize()));
```
lunarpapillo, line 625: When will `GetMinAllocationsSize()` return the same as `max()`?

Comments here to explain the unexpected code would be helpful.

tools/info/main.cpp
```diff
@@ -632,14 +634,17 @@ int main(int argc, const char** argv)
             // GFXRECON_WRITE_CONSOLE("\nDraw/dispatch call info:");
             // GFXRECON_WRITE_CONSOLE("\tTotal draw calls: %" PRIu64, stats_consumer.GetDrawCount());
             // GFXRECON_WRITE_CONSOLE("\tTotal dispatch calls: %" PRIu64, stats_consumer.GetDispatchCount());
+
+            GFXRECON_WRITE_CONSOLE("\n");
```
lunarpapillo, line 638: Is this extra newline intentional?

lunarpapillo (commented):
I find a couple of lines of code confusing, but I think it's much better than it was (thanks!).

I should probably not be considered a primary reviewer, as I'm not familiar with the code base, so I'm not clicking **Approve**.

tools/info/main.cpp
locke-lunarg: https://github.com/LunarG/gfxreconstruct/blob/dev/tools/info/main.cpp#L494

tools/info/main.cpp
locke-lunarg: https://github.com/LunarG/gfxreconstruct/blob/dev/tools/info/main.cpp#L494

locke-lunarg: `min_allocation_size_`  initialize `std::numeric_limits<uint64_t>::max()`, so if there is no memory allocation, GetMinAllocationsSize() would be the same as max().

tools/info/main.cpp
locke-lunarg: I just replaced this to
```
if (file_processor.GetCurrentFrameNumber() == 0)
{
    GFXRECON_WRITE_CONSOLE("\nFile did not contain any frames");
}
```

locke-lunarg: https://github.com/LunarG/gfxreconstruct/pull/864/files#diff-8b3e734e213d80820573b5bdb3c821349c5387e09b96ad221f1dee67f000070fR638

andrew-lunarg:
What is output for a file that is valid (has no IO errors) but ends after the `vkCreateInstance()` that triggers creation of the capture file? Perhaps the original author used the framecount as a heuristic for _"the capture file is likely to be fairly substantial"_.
Attached is a truncated vkcube trace that has that property and a second one I corrupted in a hex editor.

[vkcube.vscode-launch_frames_1_through_10_20221020T133041.quit_after_vkCreateInstance.contains_annotations.zip](https://github.com/LunarG/gfxreconstruct/files/9830165/vkcube.vscode-launch_frames_1_through_10_20221020T133041.quit_after_vkCreateInstance.contains_annotations.zip)
[vkcube.vscode-launch_frames_1_through_10_20221020T133041.quit_after_vkCreateInstance.contains_annotations.corrupted.zip](https://github.com/LunarG/gfxreconstruct/files/9830167/vkcube.vscode-launch_frames_1_through_10_20221020T133041.quit_after_vkCreateInstance.contains_annotations.corrupted.zip)

tools/info/main.cpp
andrew-lunarg: I think Bob is suggesting adding a comment above that line like:
```
// If there were no allocations, report zero rather than the default value:
GFXRECON_WRITE_CONSOLE("\tMin allocation size: %" PRIu64, ...
```
Best practice would be to define a constant somewhere and have the initialisation and this test both use it so the two places can't get out of sync and questions like this are answerable by grepping the name of the constant:
```
constexpr uint64_t GFXR_MIN_ALLOC_INITIAL = std::numeric_limits<uint64_t>::max();
```

locke-lunarg: Thank you. Added.

bradgrantham-lunarg: Could we just not output `Min allocation size` and `Max allocation size` if there were no allocations?

bradgrantham-lunarg (approved):
Looks good, but do allocation min and max need to be printed if there weren't any?

locke-lunarg:
@bradgrantham-lunarg  Good idea! It doesn't print min and max allocation size when allocation count is 0.

tools/info/main.cpp
locke-lunarg: @bradgrantham-lunarg Good idea! It won't print min and max allocation size when allocation count is 0.

andrew-lunarg (approved):

PR #867, opened 2022-10-21
Trigger trimming by env variable

panos-lunarg:
Control trimmed capture by setting an env variable. More usefull in Android where setting a hot key is not as simple as on desktop

### DIFF
--- a/USAGE_android.md
+++ b/USAGE_android.md
@@ -148,6 +148,8 @@ Option | Property | Type | Description
 ------| ------------- |------|-------------
 Capture File Name | debug.gfxrecon.capture_file | STRING | Path to use when creating the capture file.  Default is: `/sdcard/gfxrecon_capture.gfxr`
 Capture Specific Frames | debug.gfxrecon.capture_frames | STRING | Specify one or more comma-separated frame ranges to capture.  Each range will be written to its own file.  A frame range can be specified as a single value, to specify a single frame to capture, or as two hyphenated values, to specify the first and last frame to capture.  Frame ranges should be specified in ascending order and cannot overlap. Note that frame numbering is 1-based (i.e. the first frame is frame 1).  Example: `200,301-305` will create two capture files, one containing a single frame and one containing five frames.  Default is: Empty string (all frames are captured).
+Capture triggered for Android | debug.gfxrecon.capture_android_use_trigger | BOOL | Specifies that `debug.gfxrecon.capture_android_trigger` will be used to star/stop trimmed captures. Default is: `false`
+Capture trigger for Android | debug.gfxrecon.capture_android_trigger | BOOL | Set to `true` to start capturing and to `false` to stop. Default is: `false`
 Capture File Compression Type | debug.gfxrecon.capture_compression_type | STRING | Compression format to use with the capture file.  Valid values are: `LZ4`, `ZLIB`, `ZSTD`, and `NONE`. Default is: `LZ4`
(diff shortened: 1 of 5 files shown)


panos-lunarg:
Been experimenting with some android games today and realized that this could be something useful on Android.
It should simplify skipping through a game's menus and capturing actual gameplay

andrew-lunarg:
The hotkey doesn't always work on Linux for me IIRC. This could be generally useful.

bradgrantham-lunarg:
Processes on the desktop inherit their environment variables and have a local copy, so waiting on env var values to change won't work on desktop.  But it's a nice thing to have on Android!  

See #870 for some further thoughts.

andrew-lunarg:
So its doing `getprop` underneath to read "environment variables" on Android.

bradgrantham-lunarg:
> So its doing getprop underneath to read "environment variables" on Android.

Right, right, environment variables are inherited into the child process on Android also.  We're just using Android properties on Android and those are (I think?) system-global.

panos-lunarg:
It works on Android. I just enter the command in an adb shell `setprop debug.gfxrecon.capture_env_var_trigger 1/0`
Nothing groundbreaking, just something to have until we find a better solution. No problem rejecting it either.

panos-lunarg:
Right, good point

USAGE_android.md
```diff
@@ -148,6 +148,8 @@ Option | Property | Type | Description
 ------| ------------- |------|-------------
 Capture File Name | debug.gfxrecon.capture_file | STRING | Path to use when creating the capture file.  Default is: `/sdcard/gfxrecon_capture.gfxr`
 Capture Specific Frames | debug.gfxrecon.capture_frames | STRING | Specify one or more comma-separated frame ranges to capture.  Each range will be written to its own file.  A frame range can be specified as a single value, to specify a single frame to capture, or as two hyphenated values, to specify the first and last frame to capture.  Frame ranges should be specified in ascending order and cannot overlap. Note that frame numbering is 1-based (i.e. the first frame is frame 1).  Example: `200,301-305` will create two capture files, one containing a single frame and one containing five frames.  Default is: Empty string (all frames are captured).
+Capture triggered for Android | debug.gfxrecon.capture_android_triggered | BOOL | Specifies that `debug.gfxrecon.capture_android_trigger` will be used to star/stop trimmed captures. Default is: `false`
```
bradgrantham-lunarg, line 151: Can this be `use_trigger`?

bradgrantham-lunarg (approved):
Other than my one comment, I think this looks good.

lunarpapillo:
Last failure looks like a hung device.  I'm trying to get it up again.

framework/encode/capture_manager.cpp
```diff
@@ -497,7 +507,8 @@ void CaptureManager::CheckContinueCaptureForWriteMode()
         }
     }
     else if (IsTrimHotkeyPressed() ||
-             ((trim_key_frames_ > 0) && (current_frame_ >= (trim_key_first_frame_ + trim_key_frames_))))
+             ((trim_key_frames_ > 0) && (current_frame_ >= (trim_key_first_frame_ + trim_key_frames_))) ||
+             !IsTrimEnvVarEnabled())
```
ishitatsuyuki, line 511: This breaks desktop range capture. Desktop hotkey captures now only last one frame long because `!IsTrimEnvVarEnabled()` always evaluates to true.

panos-lunarg: Thanks for reporting this. Indeed this line causes a regression with the desktop hot key

PR #869, opened 2022-10-21
Annotate Captures With Layer Version

### LINKED ISSUE #831: Add an Annotation Block With the Build Version of the Capture Layer
Inject the same information reported by the `--version` parameter of the command line tools.
Put it early in the capture file, before any API calls.
* Extend `gfxrecon-info` to report the version info.
* Extend `gfxrecon-convert` to store the info in the header object of its json lines output (https://github.com/LunarG/gfxreconstruct/issues/809).
* Extend `gfxrecon-optimize` to not strip the annotations.

We can expand the header with more information: https://github.com/LunarG/gfxreconstruct/issues/868

andrew-lunarg:
convert outputs the annotations.
```
{"index":0,"annotation":{"type":"kJson","label":"operation","data":"{\n    \"tool\": \"capture\",\n    \"gfxrecon-version\": \"0.9.15-dev (andy-fix-831-annotate-captures-with-layer-version:621b79a*)\",\n    \"vulkan-version\": \"1.3.231\"\n}\n"}}
```
info counts them and outputs the operation ones.
```
Annotation info:
        Total annotations: 1
        Operation annotations: 1
    
        {
            "tool": "capture",
            "gfxrecon-version": "0.9.15-dev",
            "vulkan-version": "1.3.231"
        }

```
The other tools pass them through unchanged or ignore them as appropriate.

Fixes #831.

### DIFF
--- a/framework/decode/file_processor.cpp
+++ b/framework/decode/file_processor.cpp
@@ -1290,7 +1290,9 @@ bool FileProcessor::ProcessAnnotation(const format::BlockHeader& block_header, f
         {
             std::string label;
             std::string data;
-            size_t      total_length = size_t(label_length + data_length);
+            const auto  size_sum = label_length + data_length;
+            GFXRECON_CHECK_CONVERSION_DATA_LOSS(size_t, size_sum);
+            const size_t total_length = static_cast<size_t>(size_sum);
 
             success = ReadParameterBuffer(total_length);
             if (success)
@@ -1304,7 +1306,8 @@ bool FileProcessor::ProcessAnnotation(const format::BlockHeader& block_header, f
                 if (data_length > 0)
                 {
                     auto data_start = std::next(parameter_buffer_.begin(), label_length);
-                    data.assign(data_start, std::next(data_start, size_t(data_length)));
+                    GFXRECON_CHECK_CONVERSION_DATA_LOSS(size_t, data_length);
+                    data.assign(data_start, std::next(data_start, static_cast<size_t>(data_length)));
                 }
 
                 assert(annotation_handler_ != nullptr);
--- a/framework/decode/file_processor.cpp
+++ b/framework/decode/file_processor.cpp
@@ -1,6 +1,6 @@
 /*
-** Copyright (c) 2018-2020 Valve Corporation
-** Copyright (c) 2018-2020 LunarG, Inc.
+** Copyright (c) 2018-2020,2022 Valve Corporation
+** Copyright (c) 2018-2020,2022 LunarG, Inc.
 **
 ** Permission is hereby granted, free of charge, to any person obtaining a
 ** copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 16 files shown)


framework/format/format.h
```diff
@@ -260,7 +260,7 @@ struct AnnotationHeader
     BlockHeader    block_header;
     AnnotationType annotation_type;
     uint32_t       label_length;
-    uint32_t       data_length;
+    uint64_t       data_length;
```
andrew-lunarg, line 263: @bradgrantham-lunarg  I kept the label length 32 bit and just boosted the data one.

andrew-lunarg:
I think I'll change the name (label) of the annotation. `"header"` is generic, the json lines output of convert already has a header, and there can be multiple of these annotations to document all operations on a capture, so header doesn't really fit.  Shortlist:
1. `"tool-meta"`: Represents the fact this is metadata like build version, time, paths, users, settings, etc. added by each tool to have processed it from capture onwards.
1. `"process"`: ... Metadata from a process applied to the capture file.
1. `"stage"`: 
1. `"operation"`: 
1. `"step"`: 
1. `"transformation"`: (-) Capture isn't really a transformation.
1. `"kernel"`: Like shader kernel in a data-flow system,
1. `"tool"`: (-) Duplicates the name of one of the fields inside the json.

framework/decode/file_processor.cpp
```diff
@@ -1290,7 +1290,7 @@ bool FileProcessor::ProcessAnnotation(const format::BlockHeader& block_header, f
         {
             std::string label;
             std::string data;
-            auto        total_length = label_length + data_length;
+            size_t      total_length = size_t(label_length + data_length);
```
bradgrantham-lunarg, line 1293: Please make this a C++ cast

andrew-lunarg: Will do. I'd mislearned that these were construction, not casting, so thanks!

All this casting is to supress warnings as errors in 32 bit windows builds. This reminds me, our 32 bit builds are in trouble if any 64 bit fields in our capture format come in with values over 4GB. I don't know that we try to handle that anywhere, even to log and fail gracefully.

andrew-lunarg: Applied the macro and changed the casts.

framework/encode/capture_manager.cpp
```diff
@@ -700,6 +712,26 @@ void CaptureManager::WriteDisplayMessageCmd(const char* message)
     }
 }
 
+void CaptureManager::WriteAnnotation(const format::AnnotationType type, const char* label, const char* data)
+{
+    if ((capture_mode_ & kModeWrite) == kModeWrite)
+    {
+        const auto label_length = decltype(format::AnnotationHeader::label_length)(util::platform::StringLength(label));
```
bradgrantham-lunarg, line 719: Sorry I didn't catch this previously; should this also be a `static_cast` and use `GFXRECON_CHECK_CONVERSION_DATA_LOSS`?

bradgrantham-lunarg (approved):

framework/encode/capture_manager.cpp
andrew-lunarg: I got rid of a lot of this mess and made it much simpler. Hopefully I didn't also mess up the 32 bit win build again.

bradgrantham-lunarg:
> CI gfxreconstruct build # 2133 failed.

Spurious failure on Android; re-running shows that replay succeeds.  (As expected since this PR shouldn't affect replay.)

PR #875, opened 2022-10-28
Run AcquireNextImage Before QueuePresent

locke-lunarg:
Some applications' trimming capture don't track the first AcquireNextImage. It causes SURFACE_LOST when replaying QueuePresent. Validation VUID-VkPresentInfoKHR-pImageIndices-01296 error.

This issue only happens on virutal swapchain because captured swapchain runs AcquireNextImage in ProcessSetSwapchainImageStateCommand.


panos-lunarg (approved):
Tried it out and indeed I can replay a trimmed capture of Sacred Path without specifying `--use-captured-swapchain-indices`
LGTM

bradgrantham-lunarg (approved):

PR #881, opened 2022-11-02
Concrete ToString() Functions for Primitive Types

andrew-lunarg:
Advantages:

1. Can put a breakpoint on a particular type's ToString()
2. We can move the functions out of header later.

Once we remove the vestiges of the old form of formatted JSON output, all the boilerplate below will go from each and they'll be four lines each, including two lines that are just curlies.

```
                       ..., ToStringFlags toStringFlags = kToString_Default,
                            uint32_t      tabCount      = 0,
                            uint32_t      tabSize       = 4)
{
    GFXRECON_UNREFERENCED_PARAMETER(toStringFlags);
    GFXRECON_UNREFERENCED_PARAMETER(tabCount);
    GFXRECON_UNREFERENCED_PARAMETER(tabSize);
```

### DIFF
--- a/framework/util/to_string.h
+++ b/framework/util/to_string.h
@@ -26,6 +26,7 @@
 #define GFXRECON_TO_STRING_H
 
 #include "util/defines.h"
+#include "util/logging.h"
 
 #include <sstream>
 #include <string>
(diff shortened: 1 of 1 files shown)


framework/util/to_string.h
```diff
@@ -43,29 +44,176 @@ enum ToStringFlagBits
...
+}
+
+inline std::string
+ToString(const bool& obj, ToStringFlags toStringFlags = kToString_Default, uint32_t tabCount = 0, uint32_t tabSize = 4)
+{
+    GFXRECON_UNREFERENCED_PARAMETER(toStringFlags);
+    GFXRECON_UNREFERENCED_PARAMETER(tabCount);
+    GFXRECON_UNREFERENCED_PARAMETER(tabSize);
+    return std::to_string(obj);
+}
```
andrew-lunarg, line 107: I know it looks like a lot of code but after a couple more refactorings of this part of Convert these will reduce down to one-liners like::
```
inline std::string ToString(bool obj) { return std::to_string(obj); }

andrew-lunarg:
After sleeping on it, I'll make the case for this change post 1.0 and post some other changes that will make it a lower line count diff and allow the template to be deleted rather than deprecated.

PR #884, opened 2022-11-04
Special treatment for unknown AHB format

panos-lunarg:
Treat HAL_PIXEL_FORMAT_EXYNOS_YCbCr_420_SP_M_SBWC as AHARDWAREBUFFER_FORMAT_Y8Cb8Cr8_420


panos-lunarg:
This is just a workaround that helps get the AHBs created during replay. The S22 vulkan driver seems not bothered with the format substitution when it creates the `VkImageView`s and doesn't crash. 
Surfaces that actually use that format do not get rendered correctly though

PR #885, opened 2022-11-04
Keep HandleWrappers for invalid handles

locke-lunarg:
The application might use invalid handles that have destroyed. If
HandleWrappers are deleted when the handles are destroyed, it will get
invalid pointer address of HandleWrappers to crash when the application uses
invalid handles. So don't delete HandleWrappers when the handles are
destroyed. Delete HandleWrappers in the ending.

closed: https://github.com/LunarG/gfxreconstruct/issues/677


locke-lunarg:
Test code:
https://github.com/locke-lunarg/Vulkan-Tools/tree/invalid-handle

Handle is VK_NULL_HANDLE and invalid handle are different cases. The program could still run with VK_NULL_HANDLE handle if VK_NULL_HANDLE is legal. But it might crash with invalid handle on Nvidia device. The interesting thing is not to crash on AMD device.

bradgrantham-lunarg (approved):

PR #887, opened 2022-11-05
Optimize generated ToString compilation performance

ishitatsuyuki:
By not calling std::string constructor on every struct field ever.

generated_vulkan_struct_to_string: 79s -> 65s
generated_vulkan_struct_decoders_to_string.cpp: 73s -> 57s

### DIFF
--- a/framework/util/to_string.h
+++ b/framework/util/to_string.h
@@ -116,7 +116,7 @@ ObjectToString(ToStringFlags toStringFlags, uint32_t& tabCount, uint32_t tabSize
 
 inline void FieldToString(std::stringstream& strStrm,
                           bool               firstField,
-                          const std::string& fieldName,
+                          const char*        fieldName,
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
@andrew-lunarg Mind taking a look? This should be fairly trivial to review.

framework/util/to_string.h
```diff
@@ -116,7 +116,7 @@ ObjectToString(ToStringFlags toStringFlags, uint32_t& tabCount, uint32_t tabSize
 
 inline void FieldToString(std::stringstream& strStrm,
                           bool               firstField,
-                          const std::string& fieldName,
+                          const char*        fieldName,
```
andrew-lunarg, line 119: I always like the explicit `const char* const` to tell the compiler we won't be modifying the pointer, but this is fine too.

andrew-lunarg (approved):
Looks good to me.

PR #893, opened 2022-11-09
test Update VMA

locke-lunarg:


locke-lunarg:
VMA fixed 32-bit issue officially. I will update VMA when it release a new version.

PR #897, opened 2022-11-16
disable win dbg popup if GFXRECON_NO_DEBUG_POPUP has value

bradgrantham-lunarg:
Add a new linker section in MSVC containing a variable with an initializer function that disables debug popups if GFXRECON_NO_DEBUG_POPUP has a non-empty value

Add some usage information for GFXRECON_NO_DEBUG_POPUP

### DIFF
--- a/USAGE_desktop_Vulkan.md
+++ b/USAGE_desktop_Vulkan.md
@@ -36,6 +36,7 @@ Vulkan API calls on Desktop systems.
     4. [Trimmed File Optimization](#trimmed-file-optimization)
     5. [JSON Lines Conversion](#json-lines-conversion)
     6. [Command Launcher](#command-launcher)
+    7. [Options Common To All Tools](#common-options)
 
 ## Capturing API calls
 
@@ -658,3 +659,17 @@ gfxrecon.py capture -o vkcube.gfxr vkcube
 
 On Windows, after installing Python3, be sure to associate the `.py` file extension with
 the Python3 interpreter before you run the script.
+
+### Options Common To all Tools
+
+If the environment variable `GFXRECON_NO_DEBUG_POPUP` has any non-zero
+number or non-empty, non-numeric string value when running any of
+of the file processing tools, the tool will attempt to disable the
+'Abort, Retry, Ignore' message box displayed when `assert()` fails on
+Windows in a Debug build.  This behavior is slightly different than
+`--no-debug-popup` in that the message box is disabled before any other
+variable initialization.  This is probably most useful in headless or
+"Continuous Integration" builds when an on-screen message box that
+can't be automatically dismissed may hang scripts or cause directories
+to be locked.  (Note that "FALSE" and "no", as examples, are non-empty,
+non-numeric string values and will be interpreted as enabling the option.)
--- a/tools/compress/CMakeLists.txt
+++ b/tools/compress/CMakeLists.txt
@@ -33,7 +33,18 @@ target_sources(gfxrecon-compress
                    ${CMAKE_CURRENT_LIST_DIR}/main.cpp
                    ${CMAKE_CURRENT_LIST_DIR}/compression_converter.h
                    ${CMAKE_CURRENT_LIST_DIR}/compression_converter.cpp
+                   ${CMAKE_CURRENT_LIST_DIR}/../platform_debug_helper.cpp
 )
+
+if (MSVC)
+    # Force inclusion of "gfxrecon_disable_popup_result" variable in linking.
+    # On 32-bit windows, MSVC prefixes symbols with "_" but on 64-bit windows it doesn't.
+    if(NOT CMAKE_CL_64)
+      target_link_options(gfxrecon-replay PUBLIC "LINKER:/Include:_gfxrecon_disable_popup_result")
+    else()
+      target_link_options(gfxrecon-replay PUBLIC "LINKER:/Include:gfxrecon_disable_popup_result")
+    endif()
+endif()
 
 target_include_directories(gfxrecon-compress PUBLIC ${CMAKE_BINARY_DIR})
 
--- a/tools/convert/CMakeLists.txt
+++ b/tools/convert/CMakeLists.txt
@@ -32,7 +32,18 @@ target_sources(gfxrecon-convert
                PRIVATE
                     ${CMAKE_CURRENT_LIST_DIR}/../tool_settings.h
                     ${CMAKE_CURRENT_LIST_DIR}/main.cpp
+                    ${CMAKE_CURRENT_LIST_DIR}/../platform_debug_helper.cpp
               )
+
+if (MSVC)
+    # Force inclusion of "gfxrecon_disable_popup_result" variable in linking.
+    # On 32-bit windows, MSVC prefixes symbols with "_" but on 64-bit windows it doesn't.
+    if(NOT CMAKE_CL_64)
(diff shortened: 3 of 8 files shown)


USAGE_desktop.md
```diff
@@ -646,3 +647,15 @@ gfxrecon.py capture -o vkcube.gfxr vkcube
 
 On Windows, after installing Python3, be sure to associate the `.py` file extension with
 the Python3 interpreter before you run the script.
+
+### Options Common To all Tools
+
+If the environment variable GFXRECON_NO_DEBUG_POPUP has any non-empty
```
lunarpapillo, line 653: Recommend exact computing terms in prose have backticks, e.g. `` `GFXRECON_NO_DEBUG_POPUP` `` here and `` `abort()` `` below.

framework/util/CMakeLists.txt
```diff
@@ -98,6 +99,11 @@ target_link_libraries(gfxrecon_util platform_specific vulkan_registry ${CMAKE_DL
 if (MSVC)
     # These files may fail to compile with VS2017 and older, requiring the default section limit of 2^16 to be increased.
     set_source_files_properties(${CMAKE_SOURCE_DIR}/framework/generated/generated_vulkan_struct_to_string.cpp PROPERTIES COMPILE_FLAGS /bigobj)
+    if(NOT CMAKE_CL_64)
```
lunarpapillo, line 102: It's not clear from context why `CMAKE_CL_64` affects the underscore in the linker option... a comment for this wizardry is sure to be appreciated later.

framework/util/platform_debug_settings.cpp
```diff
@@ -0,0 +1,54 @@
...
+DWORD DisableDebugPopup(void)
+{
+    DWORD variable_length = GetEnvironmentVariableA("GFXRECON_NO_DEBUG_POPUP", nullptr, 0);
+
+    bool variable_exists = (variable_length > 0);
+
+    if (variable_exists)
+    {
+        _set_abort_behavior(0, _WRITE_ABORT_MSG | _CALL_REPORTFAULT);
+    }
```
lunarpapillo, line 43: Vague memory and the [**LunarG Suggested Code Development Workflow** document](https://docs.google.com/document/d/1W0W4Sm8hLC9Sh9xBwFJthPk7ZZzPxjqU68xDLZ6LyQY/edit#bookmark=id.4yvy0kvdgglg) suggest that other calls may also be of interest:
```
SetErrorMode(SEM_FAILCRITICALERRORS | SEM_NOGPFAULTERRORBOX);
_CrtSetReportMode(_CRT_ASSERT, _CRTDBG_MODE_FILE);
_CrtSetReportFile(_CRT_ASSERT, _CRTDBG_FILE_STDERR);
```
I'm actually not sure whether any of these are relevant or will work in this context; I'm just trying to be complete.  If investigating such is not something you have time for, I'll still happily take just disabling debug popups.

framework/util/platform_debug_settings.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+    return variable_length;
+}
+
+extern "C"
+{
+#pragma section(".CRT$XCT", read)
+    __declspec(allocate(".CRT$XCT")) DWORD gfxrecon_disable_popup_result = DisableDebugPopup();
+}
```
lunarpapillo, line 52: That is an impressive incantation... A comment here pointing to the documentation you used to figure this out would likely be appreciated by your successors. ;-)

lunarpapillo (approved):
I love this and will accept it in any form you are willing to provide it (though my excitement has not precluded me adding suggestions ;-).

USAGE_desktop.md
bradgrantham-lunarg: Yes, good recommendation.  Fixed!

framework/util/CMakeLists.txt
bradgrantham-lunarg: Yes, good recommendation.  I've added some clarifying comments.  Verify change has been made and resolve if satisfied, please.

framework/util/platform_debug_settings.cpp
bradgrantham-lunarg: Yes, very good point!  My successors or me!  I've added some clarifying comments.  Verify change has been made and resolve if satisfied, please.

framework/util/platform_debug_settings.cpp
bradgrantham-lunarg: In this case I've simply copied the code that is called for `--no-debug-popup`.  Those seem like good suggestions and we should come back to those.

lunarpapillo (approved):
Astounding wizardry!  If this works as well for GFXR as I think it will, I may press for it to become company standard - we've had other executables with `abort()` calls that hanged CI in the past.

bradgrantham-lunarg:
I'm sorry to say I have another round of work with this.  @davidd-lunarg points out that this will load in the capture layer, too, potentially conflicting with settings from the running app.  So I'll need to make this specific to the tools somehow, maybe by putting it in the tools settings parser.  That's where `--no-debug-popup` lives, anyway, so maybe that's reasonable.

USAGE_desktop.md
```diff
@@ -646,3 +647,15 @@ gfxrecon.py capture -o vkcube.gfxr vkcube
 
 On Windows, after installing Python3, be sure to associate the `.py` file extension with
 the Python3 interpreter before you run the script.
+
+### Options Common To all Tools
+
+If the environment variable `GFXRECON_NO_DEBUG_POPUP` has any non-empty
+value (e.g. "1" or "true" or "foo") when running any of of the file processing
```
locke-lunarg, line 654: Excuse me.  A silly question. Why is `foo`? What is this?

bradgrantham-lunarg: Not silly at all.  It's an awkward option.  The other options are compared against "1" or "true" or "foo".  In order to reduce the number of operations and thus the chance of failure or exception I just turn on the option if `GFXRECON_NO_DEBUG_POPUP` exists at all.  Should I just drop the statement in parentheses?

lunarpapillo: On Windows, there's no way to unset an environment variable, only a way to set it to empty, right?  So we definitely need some check on the value of the environment variable.

I've also seen situations where someone accidentally (on Windows) set an environment variable to nothing but blank spaces, which was treated as `true` because it was not empty...   And I've seen confusion when someone `set GFXRECON_NO_DEBUG_POPUP=0` and expected it to behave as unset...

For simple comparisons, I prefer that the value is `false` if:
- the environment variable doesn't exist
- the environment variable is empty
- the environment variable is just spaces
- the environment variable contains a number that evaluates to 0
and is `true` in all other cases.

If we want to get fancy, I'd add to the `false` evaluation:
- case-insensitive `no`, `off`, and `false`

bradgrantham-lunarg: As I understand it, an environment variable can be unset on Windows (`ERROR_ENVVAR_NOT_FOUND`) and that's different from "" (succeeded but length 0).

I probably could extend this in a simple way so that the variable is only true for "true", "True", "TRUE", or a number not 0.

locke-lunarg: I got it. `foo` means anything else.

locke-lunarg (approved):

USAGE_desktop.md
locke-lunarg: We might need to add `number not 0`.

bradgrantham-lunarg: I think this does all the above.  Tested it in December.

PR #899, opened 2022-11-16
Documented some ToString Templates

andrew-lunarg:
This is just a documentation quickie.

Added an assertion to the non-obvious template for flag sets
that should never be called. It isn't currently called.

### DIFF
--- a/framework/util/to_string.h
+++ b/framework/util/to_string.h
@@ -20,12 +20,14 @@
 ** FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 ** DEALINGS IN THE SOFTWARE.
 */
-/// @file Facilities for the conversion of types to strings.
+/// @file Facilities for the conversion of types to strings, oriented  towards
+/// JSON Lines output.
 
 #ifndef GFXRECON_TO_STRING_H
 #define GFXRECON_TO_STRING_H
 
 #include "util/defines.h"
+#include "util/logging.h"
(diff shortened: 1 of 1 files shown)


framework/util/to_string.h
```diff
@@ -63,6 +73,9 @@ inline std::string ToString(uint32_t      apiFlags,
     GFXRECON_UNREFERENCED_PARAMETER(toStringFlags);
     GFXRECON_UNREFERENCED_PARAMETER(tabCount);
     GFXRECON_UNREFERENCED_PARAMETER(tabSize);
+    GFXRECON_LOG_ERROR("std::string ToString(uint32_t apiFlags,..) should never be called see: " __FILE__ ", line %d ",
```
bradgrantham-lunarg, line 76: Could this be LOG_FATAL and then omit the ASSERT?

andrew-lunarg: I might have missed it but I don't see `LOG_FATAL` actually killing the process. It just seems to be a verbosity level between _error_ and _always_. 

My intent here is to log under any build, and only kill the process with an assert in debug builds.

I can actually delete all this logging and asserting though if it is a distraction as I see no way this function will ever be called. The log and assert are just sanity checks on that.

andrew-lunarg: I went ahead and removed them. The function isn't going to be called without one of use going out of our way to do so.

bradgrantham-lunarg (approved):

PR #901, opened 2022-11-17
rebind for coping no layouts info of images

locke-lunarg:
If a image doesn't have layouts info, it can't be copied correctly on BindImageMemory. But the layouts info is recorded on GetImageSubresourceLayout. The application could run BindImageMemory first, and then GetImageSubresourceLayout. It causes it can't copy image correctly and get warning `Image subresource layout info is not available for mapped memory write; capture/replay memory alignment differences will not be handled properly`, so give it an init layouts on CreateImage.

### DIFF
--- a/framework/decode/vulkan_rebind_allocator.cpp
+++ b/framework/decode/vulkan_rebind_allocator.cpp
@@ -251,6 +251,17 @@ VkResult VulkanRebindAllocator::CreateImage(const VkImageCreateInfo*     create_
             {
                 resource_alloc_info->uses_extensions = true;
             }
+
+            VkImageSubresource subresource;
+            subresource.aspectMask = VK_IMAGE_ASPECT_NONE;
+            subresource.mipLevel   = 0;
+            subresource.arrayLayer = 0;
+            VkSubresourceLayout layout;
+            functions_.get_image_subresource_layout(device_, *image, &subresource, &layout);
+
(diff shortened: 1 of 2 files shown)


framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -909,43 +920,37 @@ void VulkanRebindAllocator::WriteBoundResource(ResourceAllocInfo* resource_alloc
             }
             else
             {
-                if (!resource_alloc_info->layouts.empty())
+                if (resource_alloc_info->layouts.size() > 1)
                 {
+                    // TODO: multi-plane image format support when strides do not match.
```
panos-lunarg, line 925: Should we file an issue for that?

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -909,43 +920,37 @@ void VulkanRebindAllocator::WriteBoundResource(ResourceAllocInfo* resource_alloc
...
+                    GFXRECON_LOG_ERROR("Skipping mapped memory write for image with multiple subresources: support "
+                                       "not yet implemented");
+                }
+                else
+                {
+                    SubresourceLayouts* layouts = nullptr;
```
panos-lunarg, line 931: Does not need to be initialized

locke-lunarg:
It's not a correct solution.

PR #902, opened 2022-11-18
Encode pNext fields without valid extensions as null when encoding

ishitatsuyuki:
At least one game in the wild passes (partially) uninitialized memory to vkEnumeratePhysicalDeviceGroups; as a result, gfxrecon crashes immediately on startup trying to traverse the invalid pNext chain. This prevents using gfxrecon for driver regression tests, which is one of the acceptable use cases.

Even though the Vulkan spec requires these to be NULL, such errors tend to be overlooked because drivers often just ignore pNext fields where values are not expected.

Even if a spec update adds extension to a struct, it's unlikely that the extension struct is known to us, so just skip encoding pNext if we don't expect one.

### DIFF
--- a/framework/generated/generated_vulkan_struct_encoders.cpp
+++ b/framework/generated/generated_vulkan_struct_encoders.cpp
@@ -73,7 +73,7 @@ void EncodeStruct(ParameterEncoder* encoder, const VkRect2D& value)
 void EncodeStruct(ParameterEncoder* encoder, const VkBufferMemoryBarrier& value)
 {
     encoder->EncodeEnumValue(value.sType);
-    EncodePNextStruct(encoder, value.pNext);
+    EncodePNextStruct(encoder, nullptr);
     encoder->EncodeFlagsValue(value.srcAccessMask);
     encoder->EncodeFlagsValue(value.dstAccessMask);
     encoder->EncodeUInt32Value(value.srcQueueFamilyIndex);
@@ -133,7 +133,7 @@ void EncodeStruct(ParameterEncoder* encoder, const VkImageMemoryBarrier& value)
 void EncodeStruct(ParameterEncoder* encoder, const VkMemoryBarrier& value)
 {
     encoder->EncodeEnumValue(value.sType);
-    EncodePNextStruct(encoder, value.pNext);
+    EncodePNextStruct(encoder, nullptr);
     encoder->EncodeFlagsValue(value.srcAccessMask);
     encoder->EncodeFlagsValue(value.dstAccessMask);
 }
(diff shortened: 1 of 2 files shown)


andrew-lunarg (dismissed):
Notionally this makes sense to me and it seems right that we be robust to such mistakes from apps. The code looks correct.

andrew-lunarg:
Are you able to share the names of some of the titles and platforms you've seen this on so we can evaluate whether we need to expand our test coverage?

ishitatsuyuki:
The app is Quake II RTX (Linux native). You'll need to disable Steam runtime to attach gfxrecon, instructions are as follows:

Right click Steam Linux Runtime - Solider in steam -> browse local file -> edit _v2-entry-point and add:
```
shift 2
exec "$@"
```

I'm not aware of any other games that has similar pNext bug. When I wrote the commit message I was under the impression that other tools (Fossilize, RenderDoc) goes through some length to avoid stuff like this but looking at their commit history again I couldn't find anything significant so I might be wrong.

andrew-lunarg:
Great that it is open source.

I can confirm that they do not zero out their `pNext` in this case:
```cpp
       VkDeviceGroupDeviceCreateInfo device_group_create_info;
	VkPhysicalDeviceGroupProperties device_group_info; // Whoops, struct is not zeroed. <-----------------------

	if(num_device_groups > 0) {
		// we always use the first group
		num_device_groups = 1;
		_VK(vkEnumeratePhysicalDeviceGroups(qvk.instance, &num_device_groups, &device_group_info)); // Reference that struct with undefined value in its pNext <---------------
```
`src/refresh/vkpt/main.c`


This must fail validation and also trigger warnings in some C++ compilers.

I raised an issue on their repo but that doesn't mean we shouldn't be robust against such errors.

bradgrantham-lunarg:
@ishitatsuyuki Nice diagnosis!

This is definitely invalid Vulkan (https://registry.khronos.org/vulkan/specs/1.3-khr-extensions/html/vkspec.html#VUID-VkPhysicalDeviceGroupProperties-pNext-pNext).  Do the Vulkan Validation Layers print anything?

This PR solves the problem for all implementations based on `vk.xml` with the **same version that GFXR used in its codegen**.  But this means we won't try to encode (and skip over) any valid extension structs from newer `vk.xml` than the built GFXR.  That seems really unlikely to happen in practice, but this is also a lot of change for one broken app (as nice as it is to test Quake II RTX).

I think I'm unlikely to approve this PR in its current form.  Instead, could we wrap this with a runtime variable (something like `GFXRECON_DONT_DESCEND_INVALID_PNEXT` or something more descriptive and less awkward) and if that's `false` but a struct with no extensions has a non-nullptr `pNext`, print a warning describing the situation and that we're likely to crash and if so try again with the variable set to `true`?

ishitatsuyuki:
Yeah, after rechecking it does trip validation. Hence such kind of violation in practice might be quite more rare than I thought.

I'll try make this parameterized as you mentioned. Since all options are namespaced right now, maybe a prefix like `GFXRECON_WORKAROUND_` would make sense? Don't have a good idea for the concrete name yet, though.

bradgrantham-lunarg:
@ishitatsuyuki - is this still something you have time to work on so we can close it out?

PR #904, opened 2022-11-18
flush before closing window for possible XCB issue

### LINKED ISSUE #903: Linux AMD captures are unreliable
````
gfxrecon-replay: ../../src/xcb_conn.c:215: write_vec: Assertion `!c->out.queue_len' failed.
```
This may be a bug in XCB or a bug in the way we invoke X11.  If it is not a bug in GFXR, we may need to workaround it with repeated flushes or some such on `XCloseDisplay` or possibly reach into XCB structures to determine when it is safe to do so.

bradgrantham-lunarg:
Attempt to fix #903


andrew-lunarg:
Not looked into any of this but shouldn't any chane be in `bool XcbWindow::Destroy()`, file: `framework/application/xcb_window.cpp` ?

bradgrantham-lunarg:
Weirdly, at least in the issues we've seen, the crashes are registered through libX11, which I think on some platforms now calls through XCB.

bradgrantham-lunarg:
> Looks low risk. Perhaps we should merge just after an SDK release so we have a full cycle to see if it helps intermittent problems go away and revert it if not.

I did test this and it didn't actually solve the failure case I was testing, so I don't think there's much of a reason to commit it.

Apparently this is a long-standing problem in `libxcb` that was fixed in 2013 but apparently reintroduced or not quite fully fixed:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1293474
* https://cgit.freedesktop.org/xcb/libxcb/patch/src/xcb_out.c?id=be0fe56c3bcad5124dcc6c47a2fad01acd16f71a

Issue #734, opened 2022-07-08
Expand list of blocked device extensions


andrew-lunarg:
It's my assertion that we should block any extension we can see won't capture right rather than letting users stumble over them, report issues that may not reference the extension in question, and then have to back-track to the root cause. Better for users to see some "required extension missing message" up-front during capture.

`trace_layer.cpp` has (or will have when [PR 733](https://github.com/LunarG/gfxreconstruct/pull/733) is in) a list of blocked device extensions, `kUnsupportedDeviceExtensions`.
Evaluate all extensions out in the ecosystem to see whether they can trace and replay on our current codebase and whether they make sense to trace and replay at all. External memory extensions and ones that expose low-level details of the device captured on seem a good place to start blocking.

```cpp
    // Look problematic:
    "VK_EXT_external_memory_dma_buf",
    "VK_EXT_external_memory_host",
    // No app can expect this to always be present and we wouldn't want them to make
    // decisions based on low level PCI bus info about the capture device.
    "VK_EXT_pci_bus_info",
    // Looks problematic:
    "VK_EXT_physical_device_drm",
    // Look problematic:
    "VK_KHR_external_memory_fd",
    "VK_KHR_external_memory",
``` 

Perhaps instead of the block-list used by this PR we should move in time to a _known-good_-list and screen out all unknown extensions by default. If we do that we should have a setting to turn the screening off for people who have a new Vulkan implementation but are stuck with a stale gfxr binary for some reason.
Perhaps we should load the bad or good list from a file at runtime, also named in a setting.

ishitatsuyuki:
I'm in favor of an extension allowlist. There are way too much extensions that can't just be handled by serializing the structs alone (from EXTs like GPL #889 to undocumented ones like VK_VALVE_descriptor_host_mapping).

For the reference, RenderDoc is using an allowlist (https://github.com/baldurk/renderdoc/blob/de2d6e7be9aed1b20c5c9f624f1ae790fed387b2/renderdoc/driver/vulkan/vk_core.cpp#L816) and also maintains a (partially) machine-updateable support list (https://github.com/baldurk/renderdoc/blob/v1.x/renderdoc/driver/vulkan/extension_support.md).

ishitatsuyuki:
Looks like vkd3d-proton capture is borked unless VK_EXT_shader_module_identifier is disabled so you might want to add that to the list as a bandaid.

perim:
[...]

PR #906, opened 2022-11-20
block EXT_shader_module_identifier

bradgrantham-lunarg:
As noted in #734, `VK_EXT_shader_module_identifier` allows an app to query an identifier for a shader module which then can be used in place of a `VkShaderModule` object.  This causes a crash in capture.  @ishitatsuyuki , will you give this a try?


ishitatsuyuki:
Already have a change locally to block this extension and confirms that it fixes the crash on startup with vkd3d-proton (others fixes are going to come as I narrow things down and make sure you can repro). Thanks!

andrew-lunarg (approved):
Looks good to me.

PR #908, opened 2022-11-21
Don't write signal for zero valued timeline semaphores

ishitatsuyuki:
The Vulkan spec says in VkSemaphoreSignalInfo:
value must have a value greater than the current value of the semaphore.

The initial value is zero, so don't signal zero-valued semaphores.

### DIFF
--- a/framework/encode/vulkan_state_writer.cpp
+++ b/framework/encode/vulkan_state_writer.cpp
@@ -408,7 +408,11 @@ void VulkanStateWriter::WriteSemaphoreState(const VulkanStateTable& state_table)
             }
 
             // Write command to set semaphore value on replay
-            WriteSignalSemaphoreValue(signal_call_id, device_wrapper->handle_id, wrapper->handle_id, semaphore_value);
+            if (semaphore_value != 0)
+            {
+                WriteSignalSemaphoreValue(
+                    signal_call_id, device_wrapper->handle_id, wrapper->handle_id, semaphore_value);
+            }
         }
         else if (wrapper->signaled)
         {


framework/encode/vulkan_state_writer.cpp
```diff
@@ -408,7 +408,9 @@ void VulkanStateWriter::WriteSemaphoreState(const VulkanStateTable& state_table)
             }
 
             // Write command to set semaphore value on replay
-            WriteSignalSemaphoreValue(signal_call_id, device_wrapper->handle_id, wrapper->handle_id, semaphore_value);
+            if (semaphore_value)
```
bradgrantham-lunarg, line 411: Please wrap if and else blocks with braces.

bradgrantham-lunarg, line 411: Since `semaphore_value` is an increasing `uint64_t` value for `VK_SEMAPHORE_TYPE_TIMELINE`, as you note, would you please make this an explicit comparison against `!= 0`?  (We've been tending towards explicit comparisons for clarity.)

bradgrantham-lunarg (approved):
Looks good to me with the additions I noted!  I'll merge with those changes.  I figure after this PR every Vulkan capture will likely be shorter.  Thanks!

ishitatsuyuki:
> I figure after this PR every Vulkan capture will likely be shorter.

Violating the signal "greater" constraint results in an immediate device loss on Mesa so this probably hasn't been encountered before. As with other PRs of mine, this is a case that commonly happens with vkd3d though.

Issue #909, opened 2022-11-21
SharedMutex is buggy


ishitatsuyuki:
SharedMutex is broken. It tracks `has_read_lock_` to allow upgrading a read lock to write lock except it never unsets the flag.

https://github.com/LunarG/gfxreconstruct/blob/2a2c74f9d0b2fead65a6442aa9492685948d0e99/framework/util/shared_mutex.cpp#L69-L72

Then it will happily allow a writer lock to be taken with readers alive. This leads to corrupt entries getting inserted inside initialization blocks.

I suggest that we replace this with `std::shared_mutex`. Some reasons:
- We only take the mutex on each Vulkan function entrypoint. Vulkan has no re-entrancy guarantee AFAIK so we'll never run into a situation where upgrades are required.
- Most of the time it's just reader lock, the writer lock is only taken on vkQueuePresent(). There's nearly no contention and lock implementation should have no effect on performance. (I don't see why we should use spinning here either.)
- The current implementation performs more atomic operations than necessary.
- There is another questionable synchronization handling:
  https://github.com/LunarG/gfxreconstruct/blob/2a2c74f9d0b2fead65a6442aa9492685948d0e99/framework/util/shared_mutex.cpp#L44-L48
  `wait_for_writer_` is stored while mutex is unlocked, so another thread could come in trying to acquire the writer lock, sets the flag but ends up getting it wiped.

andrew-lunarg:
Thanks for the analysis. Do you have a reliable repro for it causing a problem in use? I do appreciate threading bugs can be indeterminate and hard to repro.

> I suggest that we replace this with std::shared_mutex

That's another point for raising our minimum from C++14 to 17: https://en.cppreference.com/w/cpp/thread/shared_mutex

panos-lunarg:
Perhaps @davidd-lunarg has something to suggest here.
I remember having a similar conversation with him in the past, when I tried to replace it with `std::shared_mutex` in a PR but he said that we have `SharedMutex` in place for performance reasons. I also remember mentioning something about a patch with fixes for `SharedMutex`.

perim:
FYI, I made a tracing implementation doing almost no mutexing (that is, none for the general case), where each thread instead streams to its own file, then the files are merged together on end of tracing. Hopefully I can open source it soon for study.

ishitatsuyuki:
[...]

PR #910, opened 2022-11-22
Fix accounting for a thread's shared lock state

panos-lunarg:
Change-Id: I7e9bfc77ddcd3f972468e59a38779dd5523f9412

### DIFF
--- a/framework/util/shared_mutex.cpp
+++ b/framework/util/shared_mutex.cpp
@@ -26,41 +26,45 @@
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(util)
 
-thread_local std::atomic_bool SharedMutex::has_read_lock_{ false };
+thread_local std::atomic_size_t SharedMutex::read_lock_depth_{ 0 };
 
 void SharedMutex::lock()
 {
+    // Acquire writer mutex and indicate to readers the writer is waiting.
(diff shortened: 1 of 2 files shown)


panos-lunarg:
I'm afraid that this patch needs some work. I just run into an assertion failure in https://github.com/LunarG/gfxreconstruct/pull/910/files#diff-86d287f9befc06931be97da47215184aaf84c64f41e94d12893623de00d5e833R36

ishitatsuyuki:
I can test this later but as noted in review this misses one of the point mentioned in #910. There's also the question whether we should retain the upgrading functionality at all (also mentioned in review comment).

ishitatsuyuki:
Just had the time to check MSVC's shared_mutex implementation and it seems at least recent versions use SRWLOCK, which should not require kernel roundtrip unless contended.

It might be interesting to redo the performance test David has done [here](https://github.com/LunarG/gfxreconstruct/issues/909#issuecomment-1322585920) and check if it's still worth rolling our own sync primitive.

davidd-lunarg:
@ishitatsuyuki Thanks for looking into this. Using `std::shared_mutex` appears to have similar capture-time performance as the internal `SharedMutex` implementation. I'll look into switching GFXR to use it.

davidd-lunarg:
#979 changes GFXR to use `std::shared_mutex`. Once that is merged I will close this PR.

PR #929, opened 2022-12-09
Add DirectX 12 support

davidd-lunarg:
This PR adds DirectX 12 support to GFXReconstruct for Windows x86+x64. This PR includes:
- Capture and replay support for DX12 including D3D12 and DXGI
- DX12 trimming and DXR support
- DX12-specific encoder, decoder, and consumer classes
- New capture, replay, and tool options
- DX12-specific modes for gfxrecon-optimize
- DX12-specific output for gfxrecon-info
 
See **USAGE_desktop_D3D12.md** for DX12 usage information.


bradgrantham-lunarg (approved):

framework/generated/dx12_generators/dx12_struct_unwrappers_header_generator.py
```diff
@@ -0,0 +1,195 @@
...
+    def write_struct_unwrap_def(self, struct, indent=''):
+        expr = indent + 'void UnwrapStructObjects({}* value, '\
+            'HandleUnwrapMemory* unwrap_memory);\n'.format(struct)
+
+        write(expr, file=self.outFile)
+
+    def write_include(self):
+        code = ''
+
+        code += '#include "encode/handle_unwrap_memory.h"\n'
```
andrew-lunarg, line 181: Could add:
```c++
code += '#include "encode/custom_dx12_struct_unwrappers.h"\n'
```
That would reflect that some instantiations of the templates in this generated header dispatch to custom functions in the header included by this line and rely on compilation units which include the generated header also including the custom header.
I'm not pushing strongly for this in case there are places this can be included that never instantiate the templates with types which require the custom header, in which case pulling in an extra header would be bloat.

framework/generated/dx12_generators/dx12_base_generator.py
```diff
@@ -0,0 +1,662 @@
...
+            'BreadcrumbContextsCount'
+        ],
+        [
+            'D3D12_FEATURE_DATA_PROTECTED_RESOURCE_SESSION_TYPES', 'pTypes',
+            'Count'
+        ],
+        [
+            'ID3D12GraphicsCommandList_OMSetRenderTargets',
+            'pRenderTargetDescriptors',
+            'RTsSingleHandleToDescriptorRange ? 1 : NumRenderTargetDescriptors'
```
andrew-lunarg, line 62: This generates code which is incorrect in the case that NumRenderTargetDescriptors is < 1 such as when the application has called the `ID3D12GraphicsCommandList::OMSetRenderTargets()` method with no render targets specified. In this case (seen in the wild), the code on this line overrides that zero count with a one, which causes a memory copy from a junk address to be initiated.

A more robust line would be something like the uglier:

```
NumRenderTargetDescriptors ? (RTsSingleHandleToDescriptorRange ? 1 : NumRenderTargetDescriptors) : 0
````

File: framework/format/api_call_id.h [hunk 3]
```diff
@@ -653,7 +655,519 @@ enum ApiCallId : uint32_t
...
+    // dxgi1_5.h
+    ApiCall_IDXGIOutput5_DuplicateOutput1                                                         = MakeApiCallId(ApiFamily_Dxgi, 0x108d),
+    ApiCall_IDXGISwapChain4_SetHDRMetaData                                                        = MakeApiCallId(ApiFamily_Dxgi, 0x108e),
+    ApiCall_IDXGIDevice4_OfferResources1                                                          = MakeApiCallId(ApiFamily_Dxgi, 0x108f),
+    ApiCall_IDXGIDevice4_ReclaimResources1                                                        = MakeApiCallId(ApiFamily_Dxgi, 0x1090),
+    ApiCall_IDXGIFactory5_CheckFeatureSupport                                                     = MakeApiCallId(ApiFamily_Dxgi, 0x1091),
+
+    // dxgi1_6.h
+    ApiCall_DXGIDeclareAdapterRemovalSupport                                                      = MakeApiCallId(ApiFamily_Dxgi, 0x1092),
+    ApiCall_IDXGIAdapter4_GetDesc3                                                                = MakeApiCallId(ApiFamily_Dxgi, 0x1093),
```
andrew-lunarg, line 819: Where is the DXGI 1.6 function `DXGIDisableVBlankVirtualization` or where is its generation suppressed in the Python codegen if capturing it is not wanted?
If it should be captured, wouldn't its logical call id be `0x1093`, and shouldn't an ID close to that be reserved for it somewhere around this line at least?

https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_6/nf-dxgi1_6-dxgidisablevblankvirtualization

PR #939, opened 2023-01-03
Check for Zero Render Target Descriptor Count

### LINKED ISSUE #938: feature-d3d12-release Can Memcpy From Junk Address When App Sets Zero NumRenderTargetDescriptors
See `framework/generated/dx12_generators/dx12_base_generator.py`:

```
        [
            'ID3D12GraphicsCommandList_OMSetRenderTargets',
            'pRenderTargetDescriptors',
            'RTsSingleHandleToDescriptorRange ? 1 : NumRenderTargetDescriptors'
```
This generates code which is incorrect in the case that NumRenderTargetDescriptors is < 1 but  RTsSingleHandleToDescriptorRange set to true.
Seen in a popular benchmarking app just as it finishes loading but before the first 3D rendered image is displayed when running on ARM64 but the issue is not platform-specific.
The application calls the ID3D12GraphicsCommandList::OMSetRenderTargets() method with no render targets specified but with RTsSingleHandleToDescriptorRange set to true. In this case (seen in the wild), the code on this line overrides that zero count with a one, which causes a memory copy from a junk address to be initiated.

andrew-lunarg:
Existing code seeks to reduce length of array copied to one in a special circumstance in which that makes sense but can lengthen size associated with a a zero-length array to one, leading to a read from a junk address.

This fix checks for a zero-length array before the extension and avoids doing it.
Fixes #938


framework/generated/generated_dx12_command_list_util.cpp
```diff
@@ -333,13 +333,9 @@ void Track_ID3D12GraphicsCommandList_OMSetRenderTargets(ID3D12CommandList_Wrappe
     GFXRECON_ASSERT(info != nullptr);
     if(pRenderTargetDescriptors != nullptr)
     {
-        if( NumRenderTargetDescriptors > 0 )
+        for (UINT i = 0; i < (NumRenderTargetDescriptors ? (RTsSingleHandleToDescriptorRange ? 1 : NumRenderTargetDescriptors) : 0); ++i)
```
bradgrantham-lunarg, line 336: By changing the `param_length` in `dx12_base_generator.py` for `ID3D12GraphicsCommandList_OMSetRenderTargets`/`pRenderTargetDescriptors`, the lookup in https://github.com/andrew-lunarg/gfxreconstruct/blob/eba1b72137e7940f96039abf7b26e3e12300a0f7/framework/generated/dx12_generators/dx12_command_list_util_body_generator.py#L138 no longer succeeds, which is why this generated code changes from an if-statement and a loop inside the body to just a loop.   To my eyes this seems fine, but since this removes any use of `SPECIAL_ARRAY_LENGTH` you could probably remove `SPECIAL_ARRAY_LENGTH` and any processing of `parameter_check_zero` in `add_loop_code()` in `framework/generated/dx12_generators/dx12_command_list_util_body_generator.py`.

bradgrantham-lunarg (approved):
Thank you!

davidd-lunarg (approved):

PR #942, opened 2023-01-05
Correct JSON Output of Handles Represented as uint64_t

andrew-lunarg:
Uses the correct function for handle output for four cases that debug-related structs have handles represented as uint64_t.

### DIFF
--- a/framework/generated/vulkan_generators/vulkan_struct_to_string_body_generator.py
+++ b/framework/generated/vulkan_generators/vulkan_struct_to_string_body_generator.py
@@ -87,6 +87,7 @@ class VulkanStructToStringBodyGenerator(BaseGenerator):
             #include "util/custom_vulkan_to_string.h"
             #include "generated_vulkan_struct_to_string.h"
             #include "generated_vulkan_enum_to_string.h"
+            #include "decode/custom_vulkan_ascii_consumer.h"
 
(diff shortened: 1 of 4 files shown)


andrew-lunarg:
Rebased on dev after the d3d12 feature merge.

framework/generated/vulkan_generators/vulkan_struct_to_string_body_generator.py
```diff
@@ -87,6 +87,7 @@ def beginFile(self, genOpts):
             #include "util/custom_vulkan_to_string.h"
             #include "generated_vulkan_struct_to_string.h"
             #include "generated_vulkan_enum_to_string.h"
+            #include "decode/custom_vulkan_ascii_consumer.h"
```
andrew-lunarg, line 90: Perhaps `HandleIdToString` should be refactored out of that header rather than taking the hit for a new include to get it here.

andrew-lunarg:
We should hold off on Convert PRs after this (or possibly before this) until Export is merged.

framework/generated/vulkan_generators/vulkan_struct_to_string_body_generator.py
andrew-lunarg: Investigating this led me to raising this issue and suggesting a fix for it in a PR: https://github.com/LunarG/gfxreconstruct/issues/959

PR #955, opened 2023-01-17
Update headers to 1.3.238

locke-lunarg:
Skip VK_KHR_video_decode_h264 VK_KHR_video_decode_h265


framework/generated/vulkan_generators/base_generator.py
```diff
@@ -71,6 +71,7 @@ def _make_re_string(list, default=None):
 _remove_extensions = [
     "VK_KHR_video_queue", "VK_KHR_video_decode_queue",
     "VK_KHR_video_encode_queue", "VK_EXT_video_encode_h264",
+    'VK_KHR_video_decode_h264', 'VK_KHR_video_decode_h265',
```
andrew-lunarg, line 74: We should have issues for adding them which reference this PR and the header version that introduced them.

andrew-lunarg (approved):
Looks good to me.

framework/generated/vulkan_generators/base_generator.py
locke-lunarg: Added https://github.com/LunarG/gfxreconstruct/issues/956

PR #957, opened 2023-01-19
Removed Subscript of Potentially Empty Vector

andrew-lunarg:
The old code was triggering a debug assertion in MS STL under rare circumstances.
Subscripting an empty vector is undefined behaviour but calling `data()` on one [is not](https://en.cppreference.com/w/cpp/container/vector/data).


andrew-lunarg:
We discussed adding an explicit check for zero length to be extra cautious but I think that would be unnecessarily messy with no practical benefit. as far as I am concerned, this is ready to be merged.

bradgrantham-lunarg (approved):

PR #960, opened 2023-01-21
Removed two Versions of HandleIdToString()

### LINKED ISSUE #959: Three Versions of HandleIdToString() Differ in Null Handling
Since Convert outputs handles as decimal JSON numbers,
the most regular output form would be a decimal number zero as a JSON number type.

Current duplicates output 1. `"VK_NULL_HANDLE"`, 2. "NULL", or 3. a single character ascii/utf-8 representation of numerical zero: 0.

Changes required:
1. Decide how to represent null handles.
2. Have one function without duplicates.

### Feedback Requested
I'd like null handles to be represented with the JSON number type as zeroes. This would allow a processing script to treat nulls and valid handles uniformly without having separate code paths to look for a either a number for valid handles or a string with a special value for null handles.

### Current Duplicates

`framework/decode/custom_vulkan_ascii_consumer.h`
```cpp
/// @param handleId An unsigned 64 bit int holding the encoded representation of a handle.
inline std::string HandleIdToString(format::HandleId handleId)
{
    if (0 != handleId)
    {
        return std::to_string(handleId); /// @todo Replace with to_chars when project moves to C++17.
    }
    /// <<-- Returns a JSON string: -->>
    return "\"VK_NULL_HANDLE\"";
}
```

`framework/util/to_string.h`
```cpp
template <typename HandleIdType>
inline std::string HandleIdToString(HandleIdType handleId)
{
    std::stringstream strStrm;
    if (handleId)
    {
        strStrm << "\"0x" << reinterpret_cast<const void*>(handleId) << "\"";
    }
    else
    {
        /// <<-- Returns a different JSON string: -->>
        strStrm << "\"NULL\"";
    }
    return strStrm.str();
}

template <>
inline std::string HandleIdToString(format::HandleId handleId)
{
    /// <<-- Returns a JSON number 0 for null handles: -->>
    return std::to_string(handleId);
}
```

andrew-lunarg:
Standardised on outputting null handle IDs as
a zero JSON number rather than some string that
downstream tools would need special handling for.

Fixes #959

### DIFF
--- a/framework/decode/custom_vulkan_ascii_consumer.h
+++ b/framework/decode/custom_vulkan_ascii_consumer.h
@@ -50,21 +50,11 @@ GFXRECON_BEGIN_NAMESPACE(decode)
 /// String representation of a null pointer.
 constexpr auto GFXRECON_TOJSON_NULL = "null";
 
-/// @param handleId An unsigned 64 bit int holding the encoded representation of a handle.
-inline std::string HandleIdToString(format::HandleId handleId)
-{
-    if (0 != handleId)
-    {
-        return std::to_string(handleId); /// @todo Replace with to_chars when project moves to C++17.
-    }
-    return "\"VK_NULL_HANDLE\"";
-}
-
 template <typename VkHandleType>
 inline std::string HandlePointerDecoderToString(HandlePointerDecoder<VkHandleType>* pObj)
 {
     auto pDecodedObj = pObj ? pObj->GetPointer() : nullptr;
-    return HandleIdToString(pDecodedObj ? *pDecodedObj : format::kNullHandleId);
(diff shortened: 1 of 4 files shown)


andrew-lunarg:
We should hold off on Convert PRs after this (or possibly before this) until Export is merged.

andrew-lunarg:
This PR will be redundant if work done on https://github.com/LunarG/gfxreconstruct/compare/dev...andrew-lunarg:gfxreconstruct:andy-export-tool-07-changes is merged.

andrew-lunarg:
Not needed due to https://github.com/LunarG/gfxreconstruct/pull/1050

PR #964, opened 2023-01-25
add a deprecated version of capture

bradgrantham-lunarg:


tools/capture/gfxrecon-capture.py
```diff
@@ -0,0 +1,50 @@
...
+argc = len(sys.argv)
+
+if __name__ == '__main__':
+
+    # We don't support running under Cygwin Python
+    if sys.platform == 'cygwin':
+        print(os.path.basename(__file__) + ' error: Cygwin Python not supported')
+        sys.exit(1)
+
+    print("Warning: the 'capture' command is deprecated and will be removed in a future update. Use 'capture-vulkan' instead.", flush=True)
```
davidd-lunarg, line 48: This script should only get run if it was called directly, `gfxrecon.py capture` will already redirect to `gfxrecon.py capture-vulkan` (#961), so maybe change the message to call out the script names? e.g.,:
"Warning: the 'gfxecon-capture.py' script is deprecated and will be removed in a future update. Use 'gfxrecon-capture-vulkan.py' instead."

bradgrantham-lunarg: Yes.  Done.

davidd-lunarg (approved):

bradgrantham-lunarg:
Added to sdk-1.3.239

PR #970, opened 2023-01-26
Registry driver version info

igorbrsn:
**The problem**
We need alternative solution to retrieve driver info in case AGS library would fail to retrieve it   

**The solution**
Read registry from `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\DirectX\*` and write it to file 

**Result**
We can use it as an alternative way to retrieve driver.

Example of registry info
NVIDIA driver info

### DIFF
--- a/framework/util/driver_info.cpp
+++ b/framework/util/driver_info.cpp
@@ -79,13 +79,13 @@ bool AMD_GetAGSInfo(std::string& driver_info)
         if (ags_gpu_info.radeonSoftwareVersion != "")
         {
             driver_info +=
-                "Radeon software version: " + static_cast<std::string>(ags_gpu_info.radeonSoftwareVersion) + "\n\t";
+                "AMD Radeon software version: " + static_cast<std::string>(ags_gpu_info.radeonSoftwareVersion) + "\n\t";
         }
 
         // Read driver version
         if (ags_gpu_info.driverVersion != "")
         {
-            driver_info += "Release version: " + static_cast<std::string>(ags_gpu_info.driverVersion) + "\n\t";
+            driver_info += "AMD driver version: " + static_cast<std::string>(ags_gpu_info.driverVersion) + "\n\t";
         }
 
         driver_info_read = true;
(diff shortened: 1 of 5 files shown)


framework/util/driver_info.cpp
```diff
@@ -172,6 +281,48 @@ bool GetDriverInfo(std::string& driver_info, format::ApiFamilyId api_family)
...
+std::string ConvertDWORDtoVersionNumber(uint64_t dword)
+{
+    uint32_t    version[4]  = {};
+    std::string str_version = "";
+    version[0]              = (unsigned int)((dword & 0xFFFF000000000000) >> 16 * 3);
+    version[1]              = (unsigned int)((dword & 0x0000FFFF00000000) >> 16 * 2);
+    version[2]              = (unsigned int)((dword & 0x00000000FFFF0000) >> 16 * 1);
+    version[3]              = (unsigned int)((dword & 0x000000000000FFFF));
+    str_version = std::to_string(version[0]) + "." + std::to_string(version[1]) + "." + std::to_string(version[2]) +
+                  "." + std::to_string(version[3]) + ".";
```
davidd-lunarg, line 315: Final period in version number could be removed: `Driver Version: 31.0.12029.4013.` -> `Driver Version: 31.0.12029.4013`

davidd-lunarg (changes requested):
If possible I think it would be good to capture any available driver info and include it in the output of `gfxrecon-info`. For example, when `AMD_IsDriverActive` is true include both AMD and system version info:
```
Driver info:
        AMD* Radeon software version: 22.10.3
        AMD* driver version: 22.20.29.04-221025a-385333C-AMD-Software-Adrenalin-Edition
        AMD* UMD version (amdxc64.dll): 8.18.10.0462 
        System driver version: 31.0.12029.4013
        System UMD version: 31.0.12029.4013
```
(\* Can use `AMD` or whatever desired prefix for these values.)

and when `AMD_IsDriverActive` is false:
```
Driver info:
        System driver version: 31.0.12029.4013
        System UMD version: 31.0.12029.4013
```

This gives any tools that might parse the output of `gfxrecon-info` a consistent field (e.g., `System driver version`) to look for. It also adds a new version # not currently included when `AMD_IsDriverActive` is true.

framework/util/driver_info.cpp
```diff
@@ -172,6 +281,48 @@ bool GetDriverInfo(std::string& driver_info, format::ApiFamilyId api_family)
...
+            driver_info_found = true;
+        }
+    }
+
+    return driver_info_found;
+}
+
+#endif
+
+std::string ConvertDWORDtoVersionNumber(uint64_t dword)
```
andrew-lunarg, line 306: I think a DWORD is still 32 bits on 64 bit Windows. Does this function and the argument to it need tweaked names?
https://learn.microsoft.com/en-gb/windows/win32/winprog/windows-data-types
https://learn.microsoft.com/en-us/windows/win32/winprog64/the-new-data-types
https://godbolt.org/z/4sbjG85nb

igorbrsn: fixed

davidd-lunarg (approved):

PR #971, opened 2023-01-26
Upgrade C++ standard to C++17

davidd-lunarg:


andrew-lunarg:
Is this a minimal PR to switch versions? The second and third commits look unrelated, unless they are working around consequences of the compiler switch.

davidd-lunarg:
> Is this a minimal PR to switch versions?

That's right. fb62856 updates the standard and suppresses warnings for deprecated `codecvt` use, a7adc4f removed the warning suppression and the deprecated code, and b28775f fixes an issue that caused an exception in debug-build Win64 DX12 CI tests after the C++17 update.

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -1986,14 +1986,17 @@ HRESULT Dx12ReplayConsumerBase::OverrideGetBuffer(DxObjectInfo*                r
         auto swapchain_info = GetExtraInfo<DxgiSwapchainInfo>(replay_object_info);
         if (swapchain_info != nullptr)
         {
-            if (swapchain_info->images[buffer] == nullptr)
+            if ((swapchain_info->image_ids.size() > buffer) &&
```
bradgrantham-lunarg, line 1989: Is `swapchain_info->image_ids.size() <= buffer` an exceptional condition that should have an assert, or considered so unlikely we shouldn't bother checking?

bradgrantham-lunarg (approved):
LGTM except my one question

framework/decode/dx12_replay_consumer_base.cpp
davidd-lunarg: The check `(swapchain_info->image_ids.size() > buffer)` was added here to handle potentially bad application input for the buffer index in the call to `GetBuffer()`, but because `SUCCEEDED(replay_result)` is checked above the buffer index should have already been validated. I've switched instead to an assert.

bradgrantham-lunarg (approved):

PR #972, opened 2023-01-27
More Robust Unit Tests for UtcString() Function

andrew-lunarg:
Rather than just looking at the punctuation and length of the output, these compare the results for semi-manually inverse-converted second counts for mornings and evening on different dates out past multiple leap years to manual reference strings.
E.g. "2000-08-31T23:47:54Z"

### DIFF
--- a/framework/util/test/main.cpp
+++ b/framework/util/test/main.cpp
@@ -205,37 +205,45 @@ TEST_CASE("UtcString", "[datetime]")
 {
     gfxrecon::util::Log::Init(gfxrecon::util::Log::kDebugSeverity);
 
-    auto zero         = UtcString(0);
-    auto one          = UtcString(1);
-    auto two          = UtcString(2);
+    // Looking for clean output like zero: 1970-01-01T00:00:00Z
+    REQUIRE(UtcString(0) == "1970-01-01T00:00:00Z");
+    REQUIRE(UtcString(1) == "1970-01-01T00:00:01Z");
+    REQUIRE(UtcString(60 * 60) == "1970-01-01T01:00:00Z");
+    REQUIRE(UtcString(60 * 60 + 1) == "1970-01-01T01:00:01Z");
+    REQUIRE(UtcString(60 * 60 + 61) == "1970-01-01T01:01:01Z");
+    REQUIRE(UtcString(60 * 60 + 61) == "1970-01-01T01:01:01Z");
+
+    const unsigned min       = 60;
+    const unsigned hour      = min * 60;
+    const unsigned day       = hour * 24;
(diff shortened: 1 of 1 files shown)


framework/util/test/main.cpp
```diff
@@ -205,37 +205,45 @@ TEST_CASE("UtcString", "[datetime]")
...
+            "2000-08-31T23:00:00Z");
+    REQUIRE(UtcString(year * 23 + leap_year * 7 + 29 * day + 31 * day * 4 + 30 * day * 2 + 30 * day + 23 * hour +
+                      47 * min) == "2000-08-31T23:47:00Z");
+    REQUIRE(UtcString(year * 23 + leap_year * 7 + 29 * day + 31 * day * 4 + 30 * day * 2 + 30 * day + 23 * hour +
+                      47 * min + 54) == "2000-08-31T23:47:54Z");
+
     auto one_thousand = UtcString(1000);
     auto two_thousand = UtcString(2000);
-
-    // Looking for clean output like zero: 1970-00-01T00:00:00Z
```
andrew-lunarg, line 214: 1970-*00*-01T00:00:00Z Hah!

framework/util/test/main.cpp
```diff
@@ -205,37 +205,45 @@ TEST_CASE("UtcString", "[datetime]")
...
-
-    REQUIRE(two[18] == '2');
-    REQUIRE(two[19] == 'Z');
-
     REQUIRE(one_thousand[19] == 'Z');
     REQUIRE(two_thousand[19] == 'Z');
-
-    REQUIRE(zero.length() == 20);
-    REQUIRE(one.length() == 20);
-    REQUIRE(two.length() == 20);
```
andrew-lunarg, line 234: These old tests were just looking at length and punctuation but not checking that the actual correct dates were being output.

framework/util/test/main.cpp
```diff
@@ -205,37 +205,45 @@ TEST_CASE("UtcString", "[datetime]")
...
     REQUIRE(one_thousand.length() == 20);
     REQUIRE(two_thousand.length() == 20);
 
-    // Pump some random time points between -2'147'483'648 and 2'147'483'647 through conversion:
+    // Pump some random time points between -2'147'483'648 and 2'147'483'647 through conversion
+    // and check that they have the correct length and punctuation:
     const time_t time_points[]{
```
andrew-lunarg, line 247: The tests bellow still only do length and punctuation but are a useful smoke test since they throw a bunch of random numbers at the function that it should cope with turning into _some_ date.

bradgrantham-lunarg (approved):

PR #973, opened 2023-01-27
[Bugfix] DX12 Unknown layout texture save/restore

mizhen:
**The problem**
In current trim process, if a resource is CPU accessible texture with unknown layout, when saving its data to trimmed trace file during capture, or uploading data to restore resource state during playback, normal resource mapping will be used to save or upload the resource data. However, this shouldn't be the case because texture with unknown layout is a special case for mapping and mapping such resource to get a mapped pointer is invalid. Such trim handling error cause GFXR failed to save or upload the resource data and make trim trace of some title missing HUD text because the title use texture with unknown layout for rendering these HUD text.

**The solution**
The pull request is the fix for the trim handling error. It avoid using normal resource map/unmap when saving/uploading resource data for such resource, instead, using CPU non-accessible resource code path in trim handling which first copy the resource data to staging buffer then using map/unmap.

Using ```ID3D12Resource::WriteToSubresource / ID3D12Resource::ReadFromSubresource``` calls can also fix the issue and avoid using the staging buffer, after analyzing both ways for fixing the issue, using staging buffer can reuse existing code path and make less changes to existing code, so the pull-request uses existing CPU non-accessible resource code path in trim handling as the final solution.

**Result**
All tests passed for the target title and other titles include trim trace and full trace capture/playback, these test results show the pull-request fixed the trim issue for target title and doesn't affect other titles.

### DIFF
--- a/framework/decode/dx12_object_info.h
+++ b/framework/decode/dx12_object_info.h
@@ -278,6 +278,9 @@ struct D3D12ResourceInfo : DxObjectExtraInfo
 
     std::map<uint64_t, uint64_t>                       mapped_gpu_addresses;
     std::map<uint64_t, graphics::Dx12ShaderIdentifier> mapped_shader_ids;
+
+    D3D12_RESOURCE_DIMENSION dimension{ D3D12_RESOURCE_DIMENSION_UNKNOWN };
+    D3D12_TEXTURE_LAYOUT     layout{ D3D12_TEXTURE_LAYOUT_UNKNOWN };
 };
 
 struct D3D12CommandSignatureInfo : DxObjectExtraInfo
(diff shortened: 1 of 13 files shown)


framework/encode/custom_dx12_wrapper_commands.h
```diff
@@ -248,6 +248,8 @@ struct CustomWrapperPostCall<format::ApiCallId::ApiCall_ID3D12Device4_CreateComm
     }
 };
 
+// TODO: add CustomWrapperPostCall struct for ID3D12Device4::CreateReservedResource1
```
davidd-lunarg, line 251: Capture support for `ID3D12Device4::CreateReservedResource1` was added in commit: https://github.com/LunarG/gfxreconstruct/commit/5aa149eae301191c40e4dad468d76739e408e71c

Could you rebase onto the latest in `dev` and update this PR as needed?

davidd-lunarg (changes requested):

framework/encode/custom_dx12_wrapper_commands.h
mizhen: Done.

davidd-lunarg (approved):

PR #974, opened 2023-01-27
Rename script improvements

igorbrsn:
**The problem**
Python script dumps all gfxr messages at the end of replay.

**The solution**
Modify script and instead `subprocess.check_output` use `subprocess.Popen` and loop the process until it is active.

**The result**
Realtime messages available during replay


bradgrantham-lunarg (approved):
This looks good and will be a nice improvement.

PR #976, opened 2023-01-30
Merging decoders used by gfxrecon-info

igorbrsn:
**Problem**
Driver and exe file info split between 2 decoders and consumers.  

**Solution**
Create one new decoder (`InfoDeoder`) and consumer (`InfoConsumer`) to contain both driver and info logic.

**Results + testing** 
The performance remained unchanged as confirmed by tests that involved capturing and replaying traces and displaying info.


davidd-lunarg (changes requested):

framework/decode/info_consumer_base.h
```diff
@@ -0,0 +1,45 @@
...
+#ifndef GFXRECON_DECODE_EXE_INFO_CONSUMER_BASE_H
+#define GFXRECON_DECODE_EXE_INFO_CONSUMER_BASE_H
+
+#include "decode/api_decoder.h"
+#include "decode/handle_pointer_decoder.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(decode)
+
+class InfoConsumerBase
```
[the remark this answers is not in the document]
andrew-lunarg: That's a good point. We usually use the `XConsumer` and `XConsumerBase` idiom when the base is manual code and the derived class is generated.

framework/decode/info_decoder.h
```diff
@@ -34,86 +34,82 @@ GFXRECON_BEGIN_NAMESPACE(decode)
...
 
     virtual void DispatchExeFileInfo(format::ThreadId thread_id, format::ExeFileInfoBlock& info) override;
 
-    void AddConsumer(ExeInfoConsumerBase* consumer) { consumers_.push_back(consumer); }
+    void AddConsumer(InfoConsumer* consumer) { consumers_.push_back(consumer); }
 
-    virtual bool SupportsApiCall(format::ApiCallId id) override { return true; }
+    virtual bool SupportsApiCall(format::ApiCallId id) { return true; }
 
-    virtual bool SupportsMetaDataId(format::MetaDataId meta_data_id) override { return true; }
```
davidd-lunarg, line 56: Include `override` on virtual methods that override base class methods.

tools/info/main.cpp
```diff
@@ -687,33 +685,34 @@ int main(int argc, const char** argv)
...
-    if (exe_info_only == false)
+    if (exe_info_only == true)
     {
+        gfxrecon::decode::InfoConsumer info_consumer(true);
+        GatherExeInfo(input_filename, info_consumer);
+        PrintExeInfo(info_consumer);
+    }
+    else
+    {
+        gfxrecon::decode::InfoConsumer info_consumer;
```
davidd-lunarg, line 701: This `info_consumer` appears to be unused.

davidd-lunarg (changes requested):

davidd-lunarg (approved):

PR #977, opened 2023-01-30
Include missing adapters in trimmed traces

igorbrsn:
**The problem**
Trimmed traces often exclude adapter info. The problem is in clearing adapters in `TrackAdapters` method. For trimmed trace we have additional `TrackAdapters` call after `PostProcess_D3D12CreateDevice` (this is where we retrieve which adapter is active). As a result, we are losing track of active adapter. Once it happened no data recorded to file about it.   

**The solution**
The solution is to clear only not active adapter instead of clearing all. This way we are not losing track of what is active at any moment.   

**Results + testing**
As a result of this fix, we have D3D12 adapter info available in every case (including titles where adapter info was missing). Tests performed on normal and trimmed traces.  

Example before changes:
before

After:
after

### DIFF
--- a/framework/graphics/dx12_util.cpp
+++ b/framework/graphics/dx12_util.cpp
@@ -652,7 +652,7 @@ void TrackAdapters(HRESULT result, void** ppFactory, graphics::dx12::ActiveAdapt
         if (SUCCEEDED(factory1->QueryInterface(__uuidof(IDXGIFactory1), reinterpret_cast<void**>(&factory1))))
         {
             // Get a fresh enumeration, in case it was previously filled by 1.0 tracking
-            adapters.clear();
+            RemoveDeactivatedAdapters(adapters);
 
             // Enumerate 1.1 adapters and fetch data with GetDesc1()
             IDXGIAdapter1* adapter1 = nullptr;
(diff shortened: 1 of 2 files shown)


framework/graphics/dx12_util.cpp
```diff
@@ -703,6 +703,26 @@ void TrackAdapters(HRESULT result, void** ppFactory, graphics::dx12::ActiveAdapt
...
+    std::vector<int64_t> deactive_adapters;
+    for (auto itr = adapters.begin(); itr != adapters.end(); ++itr)
+    {
+        if (itr->second.active == false)
+        {
+            deactive_adapters.push_back(itr->first);
+        }
+    }
+
+    if (deactive_adapters.size() > 0)
```
bradgrantham-lunarg, line 717: You don't need this `if` - if the `size()` is 0 then the loop won't execute.

framework/graphics/dx12_util.cpp
```diff
@@ -703,6 +703,26 @@ void TrackAdapters(HRESULT result, void** ppFactory, graphics::dx12::ActiveAdapt
     }
 }
 
+void RemoveDeactiveAdapters(graphics::dx12::ActiveAdapterMap& adapters)
+{
+    std::vector<int64_t> deactive_adapters;
+    for (auto itr = adapters.begin(); itr != adapters.end(); ++itr)
```
bradgrantham-lunarg, line 709: I suggest just `for(const auto& adapter: adapters)` here, then you can use `adapter.second` and `adapter.first` in the loop body.

bradgrantham-lunarg (changes requested):
I suggest "Inactive" (or possibly "Deactivated" if the adapter was active and we deactivated it somehow).

bradgrantham-lunarg (approved):

PR #978, opened 2023-01-31
Logging if the D3D12 folder with DLLs do not exist

igorbrsn:
**The problem**
During replay some traces required agility SDK's D3D12 folder with dlls for it. We need a warning mechanism to notify the user.

**The solution**
I created `VerifyAgilitySDKRuntime` function in `dx12_util` that retrieves executable folder and checking that Agility SDK runtime is in place.

**Result**
Tested with replaying captures. Warning issues during replay if Agility SDK runtime is missing. Also optimizing process tested and warning issued when `--dxr` option was used.

### DIFF
--- a/framework/graphics/dx12_util.cpp
+++ b/framework/graphics/dx12_util.cpp
@@ -790,6 +790,32 @@ bool IsSoftwareAdapter(const format::DxgiAdapterDesc& adapter_desc)
     return software_desc;
 }
 
+bool VerifyAgilitySDKRuntime()
+{
+    char module_name[MAX_PATH] = {};
+    bool detected_runtime      = false;
+
+#if defined(D3D12_SUPPORT)
+    if (GetModuleFileNameA(nullptr, module_name, MAX_PATH))
+    {
+        const std::string tool_executable_path = module_name;
+        std::string       tool_working_dir     = "";
+        size_t            dir_location         = tool_executable_path.find_last_of(util::filepath::kAltPathLastSepStr);
+        if (dir_location >= 0)
+        {
+            tool_working_dir = tool_executable_path.substr(0, dir_location);
+        }
+        const std::string runtime_path = "\\D3D12\\D3D12Core.dll";
+        if (gfxrecon::util::filepath::IsFile(tool_working_dir + runtime_path))
+        {
+            detected_runtime = true;
+        }
+    }
+#endif
+
+    return detected_runtime;
+}
+
 bool GetAdapterAndIndexbyLUID(LUID                              luid,
                               IDXGIAdapter*&                    adapter_ptr,
                               uint32_t&                         index,
--- a/framework/graphics/dx12_util.cpp
+++ b/framework/graphics/dx12_util.cpp
@@ -803,21 +803,6 @@ bool VerifyAgilitySDKRuntime()
     {
         GFXRECON_LOG_ERROR("GetModuleFileNameA failed with error code %d", GetLastError());
     }
-    else if (ret > MAX_PATH)
-    {
-        module_name.resize(ret);
-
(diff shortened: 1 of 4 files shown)


framework/graphics/dx12_util.cpp
```diff
@@ -748,6 +748,32 @@ bool IsSoftwareAdapter(const format::DxgiAdapterDesc& adapter_desc)
     return software_desc;
 }
 
+bool VerifyAgilitySDKRuntime()
+{
+    char module_name[MAX_PATH] = {};
+    bool detected_runtime      = false;
+
+#if defined(D3D12_SUPPORT)
+    if (GetModuleFileNameA(nullptr, module_name, MAX_PATH))
```
bradgrantham-lunarg, line 757: If this fails it may return MAX_PATH.
 https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulefilenamea  Could you check the return value to make sure it's less than MAX_PATH?

tools/replay/desktop_main.cpp
```diff
@@ -174,74 +176,85 @@ int main(int argc, const char** argv)
...
-
-                dx12_replay_consumer.SetFatalErrorHandler(
-                    [](const char* message) { throw std::runtime_error(message); });
-                dx12_replay_consumer.SetFpsInfo(&fps_info);
-
-                // check for user option if first pass tracking is enabled
-                if (dx_replay_options.enable_d3d12_two_pass_replay)
+                if (gfxrecon::graphics::dx12::VerifyAgilitySDKRuntime() == false)
+                {
+                    GFXRECON_LOG_WARNING(
```
bradgrantham-lunarg, line 181: If this fails, execution doesn't continue, so should this be LOG_FATAL?

bradgrantham-lunarg (changes requested):

bradgrantham-lunarg (approved):

davidd-lunarg (changes requested):

davidd-lunarg (changes requested):

tools/replay/desktop_main.cpp
```diff
@@ -174,74 +176,85 @@ int main(int argc, const char** argv)
...
-                dx12_replay_consumer.SetFpsInfo(&fps_info);
-
-                // check for user option if first pass tracking is enabled
-                if (dx_replay_options.enable_d3d12_two_pass_replay)
+                if (gfxrecon::graphics::dx12::VerifyAgilitySDKRuntime() == false)
+                {
+                    GFXRECON_LOG_FATAL(
+                        "Did not find Agility SDK runtimes. Verify \\D3D12\\D3D12Core.dll exists in the same "
+                        "directory as gfxrecon-replay.exe.");
+                    return_code = -1;
```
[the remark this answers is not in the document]
rurra-amd: Having the Agility SDK runtimes is a hard requirement after we set 
`D3D12SDKVersion = 606` 
`D3D12SDKPath = u8".\\D3D12\\"`. 

Since we set that, I don't know that the app would have anywhere to find a runtime, so why continue? Also, the warning/error message would get lost in spam. Example, this fills the terminal window. So if we need to to continue we'd have to make sure subsequent errors don't drown out the warning/error.

`[gfxrecon] WARNING - Skipping memory fill for unrecognized mapped memory object (ID = 2496276791296)`

tools/replay/desktop_main.cpp
```diff
@@ -199,49 +207,51 @@ int main(int argc, const char** argv)
                 file_processor.AddDecoder(&dx12_decoder);
             }
 #endif
+            if (return_code == 0)
```
davidd-lunarg, line 210: Is this check still needed?

davidd-lunarg (approved):

PR #979, opened 2023-01-31
Fix CaptureManager's shared mutex

davidd-lunarg:
Unify DX12/Vulkan `CaptureManager` mutex use 
Replace `CaptureManager`'s `util::ShareMutex` with `std::shared_mutex`
Rename `CaptureManager::state_mutex_` to `api_call_mutex_`


bradgrantham-lunarg (approved):

framework/encode/capture_manager.h
```diff
@@ -148,6 +141,8 @@ class CaptureManager
     bool GetIUnknownWrappingSetting() const { return iunknown_wrapping_; }
 
   protected:
+    typedef std::shared_mutex ApiCallMutexT;
+
```
andrew-lunarg, line 145: Typedef the lock type too?
```c++
typedef std::shared_lock<util::SharedMutex> ApiCallLockT;
```
Just a thought. The uses of the lock are always `auto` anyway.

davidd-lunarg: Each lock type (`std::shared_lock<ApiCallMutexT>`, `std::unique_lock<ApiCallMutexT>`) is defined just once in each of the `Acquire*ApiCallLock` methods so I think their typedefs can be omitted.

PR #980, opened 2023-01-31
Support video extensions

locke-lunarg:
https://github.com/nvpro-samples/vk_video_samples
https://developer.nvidia.com/vulkan-driver
It could use this repo to test this PR. This repo requires Vulkan SDK 1.3.239, Vulkan Beta driver, and a `.yuv` video for test. https://engineering.purdue.edu/~reibman/ece634/Videos/YUV_videos/

`vk_video_encoder` could encode `.yuv` video and output a encrypted file, like `out.264`.  `out.264` could be played by `vk_video_decoder`, but not by regular video players.
[Running vk_video_encoder Sample](https://github.com/nvpro-samples/vk_video_samples/blob/main/vk_video_encoder/BUILD.md#running-the-video-encode-sample)
[Running vk_video_decoder Sample](https://github.com/nvpro-samples/vk_video_samples/blob/main/vk_video_decoder/BUILD.md#linux-decoder-tests)

Both `vk_video_encoder` and `vk_video_decoder` could be captured and replayed successfully by this PR. You would see it's playing video when it replays the captured file of `vk_video_decoder`. But it wouldn't generate `out.264` again when it replays the captured file of `vk_video_encoder` since the output process is Win32 code, not  Vulkan code.

close: https://github.com/LunarG/gfxreconstruct/issues/956

PR #981, opened 2023-02-01
Add capture support for VK_ANDROID_frame_boundary

mikes-lunarg:
Capture the vkFrameBoundaryANDROID function and treat it as a frame delimiter.

It is very simple extension. The most complicated parts are:
- There was an existing bug in the capture layer where it wouldn't always return an pointer for layer-implemented functions. This is handled in the first commit.
- VK_ANDROID_frame_boundary is not in the main spec so this change also adds a way to patch the vk.xml tree so the code generators can know about the extension.

Questions:
- **Is this useful?** I thinks so, HWUI is already using it. Ideally this private vendor extension would be replaced with a EXT or KHR extension and hosted in the main spec.
- **Should there be an option to enable it?** Probably yes. It makes sense to make undocumented/experimental functionality opt-in.
- **Should it be exposed on Android only?** I'm not sure. Exposing it on other platforms doesn't hurt, but also isn't very useful.
- **Should it be exposed in the layer manifest file?** Probably no. This is a private extension so I don't see much value in telling the desktop loader we support it.

Future work:
- It would be nice to hook this up to the virtual swapchain (and screenshot!) on replay. Replaying a blank screen isn't very exciting...


framework/format/VK_ANDROID_frame_boundary.h
```diff
@@ -0,0 +1,30 @@
...
+** FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+** DEALINGS IN THE SOFTWARE.
+*/
+
+#ifndef GFXRECON_VK_ANDROID_FRAME_BOUNDARY
+#define GFXRECON_VK_ANDROID_FRAME_BOUNDARY
+
+#include "vulkan/vulkan.h"
+
+typedef void(VKAPI_PTR* PFN_vkFrameBoundaryANDROID)(VkDevice device, VkSemaphore semaphore, VkImage image);
```
bradgrantham-lunarg, line 28: Is there a reason this couldn't live in framework/format/platform_types.h with the other Vulkan functions?

bradgrantham-lunarg: Oh, I see you included this from there.  :)  Interesting.  I'll let that soak for a day or two.

bradgrantham-lunarg (approved):
~You may have a conflict with api IDs added by #980~
Yeah, I think that's what the conflicts listed below are.

MarkY-LunarG:
Closing and re-opening as PR #1288

PR #984, opened 2023-02-09
Check Whether a Device is UMA During Staging

### LINKED ISSUE #954: [capture] GetAvailableGpuAdapterMemory Assumes Discrete GPU
**Describe the capture bug:**
When trimming is enabled by requesting a subset of frames to be captured, at the end of the last frame before the captured  range begins, a stack of calls beginning from `D3D12CaptureManager::ActivateTrimming()` results in `Dx12StateWriter::WriteResourceSnapshots()` deciding that the GPU (if it is of the integrated / unified memory architecture type) has no available memory to use as staging buffers for download even if it has plenty.

The error stems from `uint64_t GetAvailableGpuAdapterMemory(IDXGIAdapter3* adapter)` in `framework\graphics\dx12_util.cpp`, only checking for available `DXGI_MEMORY_SEGMENT_GROUP_NON_LOCAL` memory. This is the GPU-accessible system memory (staging memory / upload & readback memory) for a discrete GPU but is unused and always zero for a true UMA integrated GPU since they share memory with the CPU and have no concept of a separate pool that both can see.

* If the GPU is integrated / UMA, this function should examine `DXGI_MEMORY_SEGMENT_GROUP_LOCAL` instead of `...NON_LOCAL`.
* It should also be renamed to something like `GetAvailableGpu_Staging_Memory` to indicate it is designed to report the memory available for CPU<->GPU transfers.
* void Dx12StateWriter::WriteResourceSnapshots() should probably also be rewritten with a UMA path which avoids copying through staging buffers .
  * *We should probably do a review of the code for UMA incompatibilities and inefficiencies like copying resources through staging buffers unnecessarily* Efficiency matters for UMA machines since they tend to be integrated and mobile, and a part of systems with low bandwidth compared to high end discrete. 

Background info on the memory segment groups:
* https://github.com/GPUOpen-LibrariesAndSDKs/D3D12MemoryAllocator/blob/aae28a6fc98f309468d0046b5e41b5e0cfcd4239/include/D3D12MemAlloc.h#L403
* https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_4/ne-dxgi1_4-dxgi_memory_segment_group

*Speculation: will we also see problems with this in discrete GPU systems configured to use ReBar (Resizable BAR)? I.e. do or could implementations running on discrete GPUs configured that way also report zero `DXGI_MEMORY_SEGMENT_GROUP_NON_LOCAL`?*

**Build Environment:**
* Branch: andy-d3d12-arm64-build
* CI: f292b3d51d65a432adb491aedb0db796326218fa

**Capture Settings:**
```
set    GFXRECON_CAPTURE_FRAMES=300-500
```

**To Reproduce:**

Enable trimming:
    set  GFXRECON_CAPTURE_FRAMES=10-20

Start 3D Mark from same environment that trimming was enabled in:

    "C:\Program Files\UL\3DMark\3DMark.exe"

Launch Wild Life Extreme Arm64 / D3D12 benchmark from 3DMark app.

Replay the capture, e.g.:
    
    ./replay/Debug/gfxrecon-replay.exe --paused ../../../../temp/d3d12_3dmark_wildlife_extreme_arm64_trimmed_frames_300_through_500_20230117T180208.gfxr

This results in a black window for quite a while as `[gfxrecon] ERROR - Detected adapter memory oversubscription` scrolls by many times and then the following is printed while the first rendered frame prints `FRAME: 693` in the rendered window (see screenshot below):

    [gfxrecon] INFO - Finished loading state for captured frame 300
    [gfxrecon] INFO - Paused at frame 1

**System environment:**
 - GPU: Adreno 8cx Gen 3
 -  Driver version: 30.0.3542.2200, dated 25/07/2022.

**Title configuration:**
  - 3D Mark Wild Life Extreme
  - Version: ...
  - No DXR

**Additional information (optional):**
  - Does capture operate correctly on a different GPU or driver version?
    Untested, but expected to work on discrete GPUs.

andrew-lunarg:
This patch queries for whether each device created on encode and decode side is based on a Unified Memory Architecture (UMA) and stores that information alongside the device wrappers . The stored bool is used during trimming to transfer data via the correct `DXGI_MEMORY_SEGMENT_GROUP`.
Unified Memory Architecture devices lack the `DXGI_MEMORY_SEGMENT_GROUP` `..._NON_LOCAL` since all memory is device-visible for them. For UMA devices like those on mobile SOCs and PC integrated GPUs we therefore use the `DXGI_MEMORY_SEGMENT_GROUP` `..._LOCAL` for our own transfers.

Fixes #954.

### DIFF
--- a/framework/encode/dx12_state_writer.cpp
+++ b/framework/encode/dx12_state_writer.cpp
@@ -747,7 +747,9 @@ void Dx12StateWriter::WriteResourceSnapshots(
                 auto     device_info   = device_wrapper->GetObjectInfo();
 
                 const double max_cpu_mem_usage = 7.0 / 8.0;
-                if (!graphics::dx12::IsMemoryAvailable(size_in_bytes, device_info.get()->adapter3, max_cpu_mem_usage))
+
+                const bool isUma = graphics::dx12::IsUma(device_wrapper->GetWrappedObjectAs<ID3D12Device>());
+                if (!graphics::dx12::IsMemoryAvailable(size_in_bytes, device_info.get()->adapter3, max_cpu_mem_usage, isUma))
                 {
                     // If neither system memory or GPU memory are able to accommodate next resource,
                     // execute the existing Copy() calls and release temp buffer to free memory
--- a/framework/encode/dx12_state_writer.cpp
+++ b/framework/encode/dx12_state_writer.cpp
@@ -748,8 +748,8 @@ void Dx12StateWriter::WriteResourceSnapshots(
 
                 const double max_cpu_mem_usage = 7.0 / 8.0;
 
-                const bool isUma = graphics::dx12::IsUma(device_wrapper->GetWrappedObjectAs<ID3D12Device>());
-                if (!graphics::dx12::IsMemoryAvailable(size_in_bytes, device_info.get()->adapter3, max_cpu_mem_usage, isUma))
+                const bool is_uma = device_wrapper->GetObjectInfo()->is_uma;
+                if (!graphics::dx12::IsMemoryAvailable(size_in_bytes, device_info.get()->adapter3, max_cpu_mem_usage, is_uma))
                 {
                     // If neither system memory or GPU memory are able to accommodate next resource,
                     // execute the existing Copy() calls and release temp buffer to free memory
--- a/framework/graphics/dx12_util.cpp
+++ b/framework/graphics/dx12_util.cpp
@@ -817,14 +817,38 @@ bool GetAdapterAndIndexbyDevice(ID3D12Device*                     device,
     return success;
 }
 
-uint64_t GetAvailableGpuAdapterMemory(IDXGIAdapter3* adapter)
+bool IsUma(ID3D12Device* device)
+{
+    GFXRECON_ASSERT(nullptr != device && "Null device pointer is expected to have been checked by callers.");
+    bool                             isUma = false;
+    D3D12_FEATURE_DATA_ARCHITECTURE1 architecture_info{};
+    const auto                       result =
+        device->CheckFeatureSupport(D3D12_FEATURE_ARCHITECTURE1, &architecture_info, sizeof(architecture_info));
+    if (SUCCEEDED(result))
+    {
+        isUma = architecture_info.UMA;
+    }
+    else
+    {
+        GFXRECON_LOG_ERROR("CheckFeatureSupport(D3D12_FEATURE_ARCHITECTURE,...) failed with result: %ld. The GPU will be assumed to be non-UMA.",
+                           static_cast<long>(result));
+    }
+    return isUma;
+}
+
+uint64_t GetAvailableGpuAdapterMemory(IDXGIAdapter3* adapter, const bool is_uma)
 {
     uint64_t available_mem = 0;
 
     if (adapter != nullptr)
     {
         DXGI_QUERY_VIDEO_MEMORY_INFO video_memory_info = {};
-        if (SUCCEEDED(adapter->QueryVideoMemoryInfo(0, DXGI_MEMORY_SEGMENT_GROUP_NON_LOCAL, &video_memory_info)))
+        DXGI_MEMORY_SEGMENT_GROUP memory_segment       = DXGI_MEMORY_SEGMENT_GROUP_NON_LOCAL;
+        if (is_uma)
+        {
+            memory_segment = DXGI_MEMORY_SEGMENT_GROUP_LOCAL;
+        }
+        if (SUCCEEDED(adapter->QueryVideoMemoryInfo(0, memory_segment, &video_memory_info)))
         {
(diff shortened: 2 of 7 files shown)


framework/graphics/dx12_util.h
```diff
@@ -187,16 +187,22 @@ bool GetAdapterAndIndexbyDevice(ID3D12Device*                     device,
 // Get the adapter at specified index
 IDXGIAdapter* GetAdapterbyIndex(graphics::dx12::ActiveAdapterMap& adapters, int32_t index);
 
+// Returns whether the device passed represents a unified memory architecture
+// (UMA) GPU as with most integrated graphics / APUs and mobile SOCs.
+// If the underlying D3D call fails, the function will also return false and the
+// error will be logged.
+bool IsUma(ID3D12Device& device);
+
```
andrew-lunarg, line 195: @davidd-lunarg Is this a good place for this little helper function and are you okay with the bool I added to those below?

framework/graphics/dx12_util.h
```diff
@@ -187,16 +187,22 @@ bool GetAdapterAndIndexbyDevice(ID3D12Device*                     device,
...
 // This function is used to get available GPU virtual memory.
 // The input is current adapter which created current device.
-uint64_t GetAvailableGpuAdapterMemory(IDXGIAdapter3* adapter);
+uint64_t GetAvailableGpuAdapterMemory(IDXGIAdapter3* adapter, bool is_uma = false);
```
bradgrantham-lunarg, line 198: I think `is_uma` shouldn't have a default value.  I don't anticipate code calling this function without providing that parameter.

bradgrantham-lunarg (changes requested):
Modulo the change I requested, this looks good to me!

framework/graphics/dx12_util.cpp
```diff
@@ -817,14 +817,37 @@ bool GetAdapterAndIndexbyDevice(ID3D12Device*                     device,
     return success;
 }
 
-uint64_t GetAvailableGpuAdapterMemory(IDXGIAdapter3* adapter)
+bool IsUma(ID3D12Device& device)
```
bradgrantham-lunarg, line 820: Oh, this probably should be an `ID3D12Device*`

framework/encode/dx12_state_writer.cpp
```diff
@@ -747,7 +747,9 @@ void Dx12StateWriter::WriteResourceSnapshots(
                 auto     device_info   = device_wrapper->GetObjectInfo();
 
                 const double max_cpu_mem_usage = 7.0 / 8.0;
-                if (!graphics::dx12::IsMemoryAvailable(size_in_bytes, device_info.get()->adapter3, max_cpu_mem_usage))
+
+                const bool isUma = graphics::dx12::IsUma(*device_wrapper->GetWrappedObjectAs<ID3D12Device>());
```
davidd-lunarg, line 751: Can the call to `IsUma` be moved outside the inner loop, or its result be saved for each device? This would avoid making repeated API calls to `ID3D12Device::CheckFeatureSupport` for each device.

framework/graphics/dx12_util.h
davidd-lunarg: Yes looks good.

framework/graphics/dx12_util.h
davidd-lunarg: Agreed--and same thing for the `is_uma` argument to `IsMemoryAvailable`.

davidd-lunarg (changes requested):

framework/graphics/dx12_util.cpp
bradgrantham-lunarg: And, as you suggested, ASSERT on it not being nullptr.

framework/encode/dx12_state_writer.cpp
andrew-lunarg: Existing code gets the device wrapper from the snapshot which the loop yields. I'd have to cache at least the isUma bool as a device wrapper member. I did want to do that but thought it might be considered too intrusive. If you and Brad like that approach, I'm happy to go that way too.

Question, should the cached var go in the device wrapper or the reference counted `ID3D12DeviceInfo` that the wrapper points at? The latter seems to fit the pattern of the existing code but bouncing around in memory two hops to look up a cached bool seems _out of keeping with the aim of going fast_. In this case in the state writer we do touch the info object for its other contents anyway. Maybe we will always do so in other cases where we'd want the cached flag.

andrew-lunarg: I've implemented caching on encode and decode sides. This captures with trimming and the capture replays without logging the error message from issue #954 (`[gfxrecon] ERROR - Detected adapter memory oversubscription`).

If you two are happy and CI passes, I'll call that done.

davidd-lunarg (approved):

PR #987, opened 2023-02-13
add missing overrides

bradgrantham-lunarg:
Enable compilation on macOS with strict compiler flags


framework/decode/stat_decoder_base.h
```diff
@@ -84,32 +86,34 @@ class StatDecoderBase : public ApiDecoder
...
+                                        format::HandleId                                    memory_id,
+                                        uint64_t                                            buffer_id,
+                                        uint32_t                                            format,
+                                        uint32_t                                            width,
+                                        uint32_t                                            height,
+                                        uint32_t                                            stride,
+                                        uint64_t                                            usage,
+                                        uint32_t                                            layers,
+                                        const std::vector<format::HardwareBufferPlaneInfo>& plane_info) override
     {}
```
andrew-lunarg, line 110: Talking of clang-format versions elsewhere, this change in formatting seems a bit random 🤷 . Just sayin.

framework/decode/vulkan_decoder_base.h
```diff
@@ -57,7 +57,7 @@ class VulkanDecoderBase : public ApiDecoder
 
     virtual void WaitIdle() override;
 
-    virtual bool IsComplete(uint64_t block_index)
+    virtual bool IsComplete(uint64_t block_index) override
     {
```
andrew-lunarg, line 61: I take it functions like this were never implemented, never called, and previous compilers / flags caused no errors.

framework/decode/stat_decoder_base.h
bradgrantham-lunarg: Yup.

andrew-lunarg:
Looks beautiful. +1 from me.

PR #994, opened 2023-02-17
Update headers to 1.3.241

locke-lunarg:
1. VkBufferCreateInfo and VkDeviceBufferMemoryRequirements are removed from MapStructHandles and UnwrapStructHandles by the generation. The reason is that we skip VK_FUCHSIA_buffer_collection, so VkBufferCreateInfo's pNext doesn't include VkBufferCollectionBufferCreateInfoFUCHSIA, so VkBufferCreateInfo doesn't include handles, so it is removed from MapStructHandles and UnwrapStructHandles. And then, VkBufferCreateInfo doesn't include handles, so VkDeviceBufferMemoryRequirements doesn't include handles, too, so VkDeviceBufferMemoryRequirements also is removed. The weird is that we have skipped VK_FUCHSIA_buffer_collection so long, why were they not removed until now? I think the reason could be something was modified in Vulkan headers.

2. Add _remove_supported in base_generator.py to skip vulkansc for enumEnumerants in genGroup.

3. Remove VK_LAYER_EXPORT from trace_layer.cpp since it is removed from Vulkan Headers.


framework/generated/vulkan_generators/base_generator.py
```diff
@@ -565,7 +569,7 @@ def genGroup(self, groupinfo, group_name, alias):
             enumerants = dict()
             for elem in groupinfo.elem:
                 supported = elem.get('supported')
-                if not supported or not 'disabled' in supported:
+                if not supported or not supported in _remove_supported:
```
bradgrantham-lunarg, line 572: For some of these, `supported` is a string with multiple comma-separated values.  For example, `VK_KHR_display` has `supported="vulkan,vulkansc"`.

This works for now because (I think) the only values are `disabled`, `vulkan`,`vulkan,vulkansc` and `vulkansc` but if another subset is added, like `vulkanes`, then some support values in the XML might be e.g. `vulkansc,vulkanes` and may want that to be excluded.

Instead of the excluding list `_remove_supported`, can this instead have a list of supported API subsets, like `_supported_subsets=['vulkan']` and then check that at least one of them is in `supported`?

bradgrantham-lunarg (changes requested):
Looks great but I had one suggestion.

framework/generated/vulkan_generators/base_generator.py
locke-lunarg: fixed it

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -720,9 +720,8 @@ VkResult VulkanCaptureManager::OverrideCreateBuffer(VkDevice
     VkDevice                  device_unwrapped      = device_wrapper->handle;
     auto                      device_table          = GetDeviceTable(device);
     auto                      handle_unwrap_memory  = VulkanCaptureManager::Get()->GetHandleUnwrapMemory();
-    const VkBufferCreateInfo* pCreateInfo_unwrapped = UnwrapStructPtrHandles(pCreateInfo, handle_unwrap_memory);
 
-    VkBufferCreateInfo modified_create_info = (*pCreateInfo_unwrapped);
+    VkBufferCreateInfo modified_create_info = (*pCreateInfo);
```
andrew-lunarg, line 724: Is this change related to the header update?

andrew-lunarg (approved):
Looks good to me.

framework/encode/vulkan_capture_manager.cpp
andrew-lunarg: I guess it relates to item one in the pr description above.

locke-lunarg: Yes. That related to it.

bradgrantham-lunarg (approved):

locke-lunarg:
Modified the descriptions in commit. Didn't change the code.

PR #995, opened 2023-02-18
Using real handles, not fake.

locke-lunarg:
Why we should give up fake handle:
https://docs.google.com/document/d/1T2MpmhmeSJfn42ZNCUJcJZsMpuuWH3nrVqNJgtY5SJw/edit

For this PR:
1. Generate VulkanStateHandleTable to store those pair of VkHandle and
wrapper. It's similar with VulkanStateTable. The different is
VulkanStateHandleTable uses mutex. VulkanStateTable uses mutex
outside the class.

2. GetWrapper<Wrapper> replaces reinterpret_cast<Wrapper*>

3. wrapper.handle replaces reinterpret_cast<VkHandle>(wrapper)

4. Remove the code that changed handle to wrapper address, and Add the
new wrapper to VulkanStateHandleTable

5. Rmove data from VulkanStateHandleTable when the wrapper is deleted.

6. It doesn't need trasnform functnions between real and fake handles
any longer. Remove GetWrappedHandle and UnwrapHandles.

One issue: Multithreading DestroyHandle and CreateHandle. `Wolfenstein II The New Colossus` use multithreading to destroy massive handle and create massive handle at the same time. When it destroys a handle, and then create the same type new handle, the new handle value could be the same with the destroyed one.  It causes conflict when it removes the destroyed one and adds new one to the same map by multithreading. After I added mutex in vulkan_state_table_base.h, the issue was better, but it still has 10% chances to cause conflict. The conflict handle seems always to be VkRenderPass. Maybe I should add mutex in somewhere else.

The following PR: Rename UnwrapStructHandles, HandleUnwrapMemory, and everything about unwrap. They would look confused if we don't use unwrapped handles any longer. They should rename to something like, `Copy Pointer Memory`.


framework/encode/vulkan_state_table_base.h
```diff
@@ -71,6 +71,42 @@ class VulkanStateTableBase
...
+    template <typename Wrapper>
+    const Wrapper* GetWrapper(typename Wrapper::HandleType                            handle,
+                              const std::map<typename Wrapper::HandleType, Wrapper*>& map) const
+    {
+        const std::lock_guard<std::mutex> lock(mutex_);
+        auto                              entry = map.find(handle);
+        return (entry != map.end()) ? entry->second : nullptr;
+    }
+
+    mutable std::mutex mutex_;
```
locke-lunarg, line 109: Can anyone help me to use mutex lock correctly?

andrew-lunarg:
I wonder if the error is in the application's external synchronisation and
it just happens to affect us worse than it does the downstream driver.

It would be an interesting experiment to change the shared lock at the top
of our vk(Create|Delete)RenderPass() into an exclusive lock and see if that
makes any difference.

If that does make the errors go away we could add envars to either make all
vkcreate/vkdestroy functions exclusively locked or make a list of pairs of
then exclusively locked as a workaround for buggy apps, effectively
serialising their use of create/destroy from multiple threads.

andrew-lunarg:
I wonder if the error is in the application's external synchronisation and it just happens to affect us worse than it does the downstream driver.

It would be an interesting experiment to change the shared lock at the top of our `vk(Create|Delete)RenderPass()` wrappers and perhaps our `vkCmdBeginRenderPass()` wrapped into an exclusive lock and see if that makes any difference \*.

The general principal here would be to start with the widest lock possible and then experiment with narrowing the scope down if you need to until it covers only the region of conflict. The widest scope for the lock might not just be this function pair but it might actually be generating the change for all Vulkan functions that we intercept. I mean injecting  `auto api_call_lock = VulkanCaptureManager::AcquireExclusiveApiCallLock();` into all of them to make our layer effectively one thread at a time as a first step before narrowing down on the problem.

If that does make the errors go away we could add envars to either make all vkcreate/vkdestroy functions exclusively locked or make a list of pairs of then exclusively locked as a workaround for buggy apps, effectively serialising their use of create/destroy from multiple threads.

It might also be possible to narrow the scope of the lock back down by flipping `AcquireExclusiveApiCallLock` calls back to `Acquire`*Shared*`ApiCallLock` incrementally until only the relevant ones remain exclusive, and then debugging/eyeballing through to find ways to bring the exclusive api lock or another lock even tighter, inside the body of our code.

\* Like this:
```
diff --git a/framework/generated/generated_vulkan_api_call_encoders.cpp b/framework/generated/generated_vulkan_api_call_encoders.cpp
index ae64beff..80253f53 100644
--- a/framework/generated/generated_vulkan_api_call_encoders.cpp
+++ b/framework/generated/generated_vulkan_api_call_encoders.cpp
@@ -2520,7 +2520,7 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateRenderPass(
     const VkAllocationCallbacks*                pAllocator,
     VkRenderPass*                               pRenderPass)
 {
-    auto api_call_lock = VulkanCaptureManager::AcquireSharedApiCallLock();
+    auto api_call_lock = VulkanCaptureManager::AcquireExclusiveApiCallLock();
 
     bool omit_output_data = false;
 
@@ -2560,7 +2560,7 @@ VKAPI_ATTR void VKAPI_CALL DestroyRenderPass(
     VkRenderPass                                renderPass,
     const VkAllocationCallbacks*                pAllocator)
 {
-    auto api_call_lock = VulkanCaptureManager::AcquireSharedApiCallLock();
+    auto api_call_lock = VulkanCaptureManager::AcquireExclusiveApiCallLock();
 
     CustomEncoderPreCall<format::ApiCallId::ApiCall_vkDestroyRenderPass>::Dispatch(VulkanCaptureManager::Get(), device, renderPass, pAllocator);
```

framework/generated/vulkan_generators/vulkan_state_table_header_generator.py
```diff
@@ -108,6 +114,14 @@ def generate_all(self):
...
             map_code += '    std::map<format::HandleId, {0}*> {1};\n'.format(handle_wrapper, handle_map)
+            vk_insert_code += '    bool InsertWrapper({0}* wrapper) {{ return InsertEntry(wrapper->handle, wrapper, {1}); }}\n'.format(handle_wrapper, handle_map)
+            vk_remove_code += '    bool RemoveWrapper(const {}* wrapper) {{\n'.format(handle_wrapper)
+            vk_remove_code += '         if (wrapper == nullptr) return false;\n'
+            vk_remove_code += '         return RemoveEntry(wrapper->handle, {});\n'.format(handle_map)
+            vk_remove_code += '    }\n'
+            vk_get_code += 'template<> inline {0}* VulkanStateHandleTable::GetWrapper<{0}>({1} handle) {{ return VulkanStateTableBase::GetWrapper(handle, {2}); }}\n'.format(handle_wrapper, vkhandle_name, handle_map)
+            vk_const_get_code += 'template<> inline const {0}* VulkanStateHandleTable::GetWrapper<{0}>({1} handle) const {{ return VulkanStateTableBase::GetWrapper(handle, {2}); }}\n'.format(handle_wrapper, vkhandle_name, handle_map)
+            vk_map_code += '    std::map<{0}, {1}*> {2};\n'.format(vkhandle_name, handle_wrapper, handle_map)
```
andrew-lunarg, line 124: Don't we want a `std::unordered_map`? I believe other projects like VL go further and use a robinhood map for even better performance.

locke-lunarg: Thank you. Changed it.

locke-lunarg:
Add option GFXRECON_ENABLE_EXCLUSIVE_LOCK to set all Vulkan Capturing function using GetEnableExclusiveLock(). It actually avoid thread conflict for Create/Destroy. When conflicts happens, we can see logs,  like `Create a duplicated Handle`, because it couldn't remove the destroyed one from the map in time.

But crash still happens about one every 30 times. This link includes the crashes data, like .gfxr, .jsonl, VVL(Validation Layers info).txt. Replay could re-generate the issue. The crash reason is that it didn't run vkBeginCommandBuffer before vkCmdBeginRenderPass. I feel this might be the application's issue. We couldn't fix it.
https://drive.google.com/drive/folders/1Om2M5YT1R76yxRInB0ThgoZn3bTenTb2

andrew-lunarg:
On Ubuntu 22.04.1 I tried this:
```
python3 ./scripts/build.py -a x64 -c debug
```
And saw build errors in the encode unit tests\*. I think that `framework/encode/test/main.cpp` needs updating. Not sure how CI missed it.

\*
```
In file included from /home/andrew/dev/andrew-lunarg-gfxreconstruct-2/framework/encode/test/main.cpp:27:
/home/andrew/dev/andrew-lunarg-gfxreconstruct-2/framework/encode/test/main.cpp: In function ‘void C_A_T_C_H_T_E_S_T_0()’:
/home/andrew/dev/andrew-lunarg-gfxreconstruct-2/framework/encode/test/main.cpp:58:35: error: ‘GetWrappedHandle’ is not a member of ‘gfxrecon::encode’; did you mean ‘CreateWrappedHandle’?
   58 |         REQUIRE(gfxrecon::encode::GetWrappedHandle(buffer) == kBufferHandle);
      |                                   ^~~~~~~~~~~~~~~~
In file included from /home/andrew/dev/andrew-lunarg-gfxreconstruct-2/framework/encode/test/main.cpp:27:
/home/andrew/dev/andrew-lunarg-gfxreconstruct-2/framework/encode/test/main.cpp:58:35: error: ‘GetWrappedHandle’ is not a member of ‘gfxrecon::encode’; did you mean ‘CreateWrappedHandle’?
   58 |         REQUIRE(gfxrecon::encode::GetWrappedHandle(buffer) == kBufferHandle);
      |                                   ^~~~~~~~~~~~~~~~
/home/andrew/dev/andrew-lunarg-gfxreconstruct-2/framework/encode/test/main.cpp:58:35: error: ‘GetWrappedHandle’ is not a member of ‘gfxrecon::encode’; did you mean ‘CreateWrappedHandle’?
   58 |         REQUIRE(gfxrecon::encode::GetWrappedHandle(buffer) == kBufferHandle);
   ...
   ```

**Comment by locke-lunarg** [comment]
@andrew-lunarg  Thank you. Fixed `framework/encode/test/main.cpp` issue.

**Comment by andrew-lunarg** [comment]
Does the commit [Add option ENABLE_EXCLUSIVE_LOCK](https://github.com/LunarG/gfxreconstruct/pull/995/commits/3317768f96de626ce1e63c23a3b9ba27afb5ea65) mostly fix the threading issue all by itself? If so, perhaps we should give it its own PR and integrate it as a workaround for games that trigger threading issues irrespective of the approach taken to handle wrapping/management.

**Comment by andrew-lunarg** [comment]
Looking at the main part of the PR, I see one map per handle type, all protected by a single shared mutex. I can imagine some other ways of structuring things with different tradeoffs:

1. **One map** for all handle types with **one mutex**.
   * All the map instantiations are from a handle to a pointer and so will make bit-identical data layouts in memory and machine code. Some template get/set wrappers could hide the casting required to share the map.
   * Avoids duplication of template generated code for one map type per handle/wrapper type. We would only instantiate one map type <uint64_t,void*> not 43 duplicates.
   * This may have unforeseen (by me) interactions with code already using these maps though.
1. **Many maps**: one for each handle types but with **one mutex per map** for lower contention.
   * Multiple threads can access the different maps without getting in each others way.

Even with one map object per handle type it should be possible to instantiate only a single `map<uint64_t,void*>` type and have two level accessors: type erased that take the lock and call the map functions, and templated type safe wrappers that inline and are optimised away.

These are just thoughts and I don't know this part of the code as well as others in the team so I understand that I may be missing key points for why things are the way they are, especially since we already had multiple maps. In fact some of this may be out of scope of the PR, relating to how the code was already structured.

File: framework/generated/generated_vulkan_state_table.h [hunk 3]
```diff
@@ -267,52 +267,421 @@ class VulkanStateTable : VulkanStateTableBase
...
+    std::unordered_map<format::HandleId, RenderPassWrapper*> renderPass_map_;
+    std::unordered_map<format::HandleId, SamplerWrapper*> sampler_map_;
+    std::unordered_map<format::HandleId, SamplerYcbcrConversionWrapper*> samplerYcbcrConversion_map_;
+    std::unordered_map<format::HandleId, SemaphoreWrapper*> semaphore_map_;
+    std::unordered_map<format::HandleId, ShaderModuleWrapper*> shaderModule_map_;
+    std::unordered_map<format::HandleId, SurfaceKHRWrapper*> surfaceKHR_map_;
+    std::unordered_map<format::HandleId, SwapchainKHRWrapper*> swapchainKHR_map_;
+    std::unordered_map<format::HandleId, ValidationCacheEXTWrapper*> validationCacheEXT_map_;
+    std::unordered_map<format::HandleId, VideoSessionKHRWrapper*> videoSessionKHR_map_;
+    std::unordered_map<format::HandleId, VideoSessionParametersKHRWrapper*> videoSessionParametersKHR_map_;
```
andrew-lunarg, line 313: `WriteState()` might need adapting to use sort contents of `unordered_maps` if we use them since it iterates over maps and relied on the current order the `std::map` provides.

locke-lunarg: Thank you. Fixed it in `Change map to unordered_map` commit.

locke-lunarg:
@andrew-lunarg
1. Creating an individual PR for threading issue is a good idea. But we still have an issue for Wolf II. I can't identify if it's related to threading. 
There a ReadMe about the issue in this shared folder. 
https://drive.google.com/drive/u/3/folders/1Om2M5YT1R76yxRInB0ThgoZn3bTenTb2

2. Every type handles has its own map because I'm afraid of the different type handles could have the same handle value. 

According to vkspec:
```
non-dispatchable handles may encode object information directly in the handle rather than acting as a reference to an underlying object, and thus may not have unique handle values. 
```
https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#fundamentals-objectmodel-overview

If it's sure that active handle value of different handle type won't duplicate, I'm ok to use one map for the whole handles.

andrew-lunarg:
> According to vkspec

My bad then. Having multiple maps is a good solution.

locke-lunarg:
https://github.com/LunarG/gfxreconstruct/issues/1027
Moved a commit about thread and crash issue to this issue.

framework/generated/generated_vulkan_state_table.h
```diff
@@ -313,6 +313,375 @@ class VulkanStateTable : VulkanStateTableBase
     std::map<format::HandleId, VideoSessionParametersKHRWrapper*> videoSessionParametersKHR_map_;
```
andrew-lunarg, line 313: This is a very superficial comment and I don't understand the architecture here, but are we maintaining these original maps and the unordered ones below for every handle live in the system at any one time?

E.g. every image created will have its handle tracked in `std::map<format::HandleId, ImageWrapper*> image_map_` and also in `std::unordered_map<VkImage, ImageWrapper*> image_map_;` below at the same time for the lifetime of the image?

framework/generated/generated_vulkan_state_table.h
```diff
@@ -313,6 +313,375 @@ class VulkanStateTable : VulkanStateTableBase
...
+    std::unordered_map<VkRenderPass, RenderPassWrapper*> renderPass_map_;
+    std::unordered_map<VkSampler, SamplerWrapper*> sampler_map_;
+    std::unordered_map<VkSamplerYcbcrConversion, SamplerYcbcrConversionWrapper*> samplerYcbcrConversion_map_;
+    std::unordered_map<VkSemaphore, SemaphoreWrapper*> semaphore_map_;
+    std::unordered_map<VkShaderModule, ShaderModuleWrapper*> shaderModule_map_;
+    std::unordered_map<VkSurfaceKHR, SurfaceKHRWrapper*> surfaceKHR_map_;
+    std::unordered_map<VkSwapchainKHR, SwapchainKHRWrapper*> swapchainKHR_map_;
+    std::unordered_map<VkValidationCacheEXT, ValidationCacheEXTWrapper*> validationCacheEXT_map_;
+    std::unordered_map<VkVideoSessionKHR, VideoSessionKHRWrapper*> videoSessionKHR_map_;
+    std::unordered_map<VkVideoSessionParametersKHR, VideoSessionParametersKHRWrapper*> videoSessionParametersKHR_map_;
```
andrew-lunarg, line 592: Re my question above, are we maintaining this set of maps live in the system at the same time as the `std::map` equivalents above?

framework/generated/generated_vulkan_state_table.h
andrew-lunarg, line 313: This is a very superficial comment and I don't understand the architecture here, but are we maintaining these original maps and the unordered ones below for every handle live in the system at any one time?

E.g. every image created will have its handle tracked in `std::map<format::HandleId, ImageWrapper*> image_map_` and also in `std::unordered_map<VkImage, ImageWrapper*> image_map_;` below at the same time for the lifetime of the image?

framework/generated/generated_vulkan_state_table.h
andrew-lunarg, line 592: Re my question above, are we maintaining this set of maps live in the system at the same time as the `std::map` equivalents above?

framework/generated/generated_vulkan_state_table.h
andrew-lunarg: Okay, answering myself: The old set of `map`s live only as long as the system is tracking state to allow trimmed captures (deferred capture / capture from frame >1), but the new maps are static data that lives for the lifetime of the capture layer.

locke-lunarg: The lifetime of data in both maps are almost the same. They are inserted and removed at the same Api calls, but different place. `maps` are at `EndCreateApiCallCapture` and `EndDestroyApiCallCapture`. `unordered_maps`  are at `CreateWrappedHandles` and `DestroyWrappedHandle`. I attempted to keep only one maps, but it seems not to possible,

framework/encode/vulkan_handle_wrapper_util.h
```diff
@@ -140,7 +205,12 @@ void CreateWrappedNonDispatchHandle(typename Wrapper::HandleType* handle, PFN_Ge
         Wrapper* wrapper   = new Wrapper;
         wrapper->handle    = (*handle);
         wrapper->handle_id = get_id();
-        (*handle)          = reinterpret_cast<typename Wrapper::HandleType>(wrapper);
+        if (!state_handle_table_.InsertWrapper(wrapper))
```
bradgrantham-lunarg, line 208: Under what circumstance would this happen?  Would we have to have an internal error, like finding the wrapper for the handle failed in a parent function, but then inserting the wrapper here in this function failed because the handle was already in the map?

bradgrantham-lunarg (approved):

framework/encode/vulkan_handle_wrapper_util.h
locke-lunarg: It could happen in multi-threading. When it destroys and creates by mutli-threading at the same time, it could create the same value of the new handle, and might not be able to erase the destroyed data in time before inserting a new one.

PR #996, opened 2023-02-20
Rebased Export PR 918 on Latest dev

andrew-lunarg:
The alternate JSON conversion tool Export, brought up to date with the current state of the repo. This takes the original PR https://github.com/LunarG/gfxreconstruct/pull/918 which was created before the upheaval of the DX12 integration and some other changes, and gets it building again after a rebase on latest dev.

* Add new command tool to export a trace capture contents to JSON format.
* Output format supported include JSON and JSON Lines.
* Add dependency to Nlohmann's JSON library.
* JSON output includes internal metacommands.
* Allow dumping of binary data into external binary files.
* Allow outputting each frame into a separate file.
* Allow expanding Vulkan's flags into human readable / symbolic form instead of just a value.

The contribution in this PR has been further evolved to replace the old Convert tool here: https://github.com/LunarG/gfxreconstruct/pull/1050

PR #997, opened 2023-02-21
Export Top-level Schema: Nested Option

andrew-lunarg:
This PR proposes a change to the way the Export schema represents the different top-level objects such as functions, metacommands, and state markers. It follows #996.
Rather than having an explicit `name:value` pair in the top level object identifying its type such as `"gfxrecon_block":"FunctionCall"`, this approach opens a nested object to carry the body of the object, while its name identifies its type (see examples below). This should allow a natural nested expression in a formal JSON schema and is slightly less verbose, saving some space.
Shorter names are also used as a measure to save some more space ("return_value"->"return", "call_index"->"index", etc.).
This PR is most easily evaluated by looking at the top commit [here](https://github.com/LunarG/gfxreconstruct/pull/997/commits/99a7d83115c4474aa9df2843b5df578322e492e0).

```json
    {
        "state": {
            "marker_type": "BeginMarker",
            "frame_number": 25
        }
    },
    {
        "index": 118,
        "vkFunc": {
            "name": "vkEndCommandBuffer",
            "thread": 1,
            "return": "VK_SUCCESS",
            "args": {
                "commandBuffer": "0x000000000000002b"
            }
        }
    },
    {   
        "meta": {
            "meta": "SetSwapchainImageStateCommand",
            "args": {
                "device_id": "0x0000000000000004",
                "swapchain_id": "0x000000000000000e",
                "last_presented_image": 2,
                "image_state": "not available"
            }
        }
    },
    {
        "state": {
            "marker_type": "EndMarker",
            "frame_number": 25
        }
    }
```


andrew-lunarg:
Superseded by https://github.com/LunarG/gfxreconstruct/pull/1050

PR #1001, opened 2023-02-24
Export: Made Some Hex Types Variable Length

andrew-lunarg:
I'm suggesting we make this change since the JSON isn't some columnar, fixed-width format, so we may as well save space. I also find the long line of up to 15 zeros on the left of a number obfuscating. I squint and scan for eights hiding in there.

Putting this up for feedback.


framework/util/to_string.h
```diff
@@ -158,14 +158,6 @@ inline std::string Bool32ToString(const /* Don't take the header dependency for
...
 }
 
-template <typename T, typename std::enable_if<std::is_integral<T>::value>::type* = nullptr>
-std::string to_hex_string(T value)
-{
-    std::stringstream stream;
-    stream << "0x" << std::setfill('0') << std::setw(sizeof(T) * 2) << std::hex << value;
-    return stream.str();
-}
-
```
andrew-lunarg: This was dead code.

andrew-lunarg:
This was put here as a draft to solicit feedback.

PR #1013, opened 2023-02-28
Add ExecuteIndirect() CountBuffer Support

JerryAMD:
**The Problem**
ID3D12GraphicsCommandList::ExecuteIndirect method could pass the pararmeters' count by a buffer resource. Current code does not read the number from count buffer, so that the GFXR replay engine does not able to translated parameters at correct count.

**The Solution**
Added new ResourceValueType to represent CountBuffer data. Read CountBuffer data at first pass.
Push the number of arguments to the map, then second pass to translate all of the arguments.
Translated the parameters by CommandSignature.

**The Result**
So far only D3D12 sample ExecuteIndirect uses the countbuffer to pass the count number, which could verify the changes.


framework/decode/dx12_object_info.h
```diff
@@ -137,12 +140,33 @@ static size_t GetResourceValueSize(ResourceValueType type)
...
     ResourceValueType     type{ ResourceValueType::kUnknown };
     uint64_t              size{ 0 };
     D3D12StateObjectInfo* state_object{ nullptr }; ///< Used to map values in shader records.
+    ArgumentBufferExtraInfo arg_buffer_extra_info;
+
+    ResourceValueInfo(uint64_t                in_offset,
+                      ResourceValueType       in_type,
+                      uint64_t                in_size,
+                      D3D12StateObjectInfo*   in_state_object          = nullptr,
+                      ArgumentBufferExtraInfo in_arg_buffer_extra_info = { nullptr, nullptr, 0 })
```
davidd-lunarg, line 162: These should not have default values. The default arguments to the constructor can lead to errors where new `ResourceValueInfo`s are created and modified from existing `ResourceValueInfo`s.

For example in `CopyResourceValuesFromDstToSrc` below:
```
        src.insert({ static_cast<uint64_t>(src_offset),
                     (*dst_iter).type,
                     static_cast<uint64_t>((*dst_iter).size),
                     (*dst_iter).state_object });
```
`ResourceValueInfo::in_arg_buffer_extra_info` from `(*dst_iter)` should be set on the new `ResourceValueInfo` but currently it gets the default argument instead.

framework/decode/dx12_resource_value_mapper.cpp
```diff
@@ -1226,11 +1240,40 @@ void Dx12ResourceValueMapper::MapValue(const ResourceValueInfo& value_info,
...
+            for (const auto& resource_value_info :
+                 value_info.arg_buffer_extra_info.command_signature_info->resource_value_infos)
+            {
+                resource_value_infos.insert({ resource_value_info.offset + command_byte_offset,
+                                              resource_value_info.type,
+                                              resource_value_info.size,
+                                              value_info.state_object });
+            }
+            command_byte_offset += value_info.arg_buffer_extra_info.command_signature_info->byte_stride;
+        }
```
davidd-lunarg, line 1269: This code block is similar to the else block in `PostProcessExecuteIndirect`. Could it be refactored into a function such as `GetExecuteIndirectResourceValues(ResourceValueInfoMap& resource_value_info_map, uint32_t command_count, etc..)`?

See `GetDispatchRaysResourceValues` which is a helper for getting resource-values for `ResourceValueType::kIndirectArgumentDispatchRays` and `Dx12ResourceValueMapper::PostProcessDispatchRays()`

davidd-lunarg (changes requested):

framework/decode/dx12_resource_value_mapper.cpp
```diff
@@ -1226,11 +1251,30 @@ void Dx12ResourceValueMapper::MapValue(const ResourceValueInfo& value_info,
...
+    {
+        // Read command counts while CountBuffer contents is still visible in temp_resource_data
+        uint32_t command_count = 0;
+        util::platform::MemoryCopy(&command_count,
+                                   sizeof(command_count),
+                                   temp_resource_data.data() + value_info.offset,
+                                   sizeof(command_count));
+
+        // Insert new ArgumentBuffer RV infos, which will queue them for translation.
+        GetExecuteIndirectResourceValues(indirect_values_map[value_info.arg_buffer_extra_info.argument_buffer],
```
davidd-lunarg, line 1266: If `command_count == 0`, this call to `GetExecuteIndirectResourceValues` can be skipped. This should avoid triggering the warning `Did not find resource value info for the given object.` in `Dx12ResourceValueMapper::MapResources`

JerryAMD: `GetExecuteIndirectResourceValues` will go through each argument by the command_count, if `command_count == 0`, the functions will return immediately, won't add any value to the map.

> If `command_count == 0`, this call to `GetExecuteIndirectResourceValues` can be skipped. This should avoid triggering the warning `Did not find resource value info for the given object.` in `Dx12ResourceValueMapper::MapResources`

davidd-lunarg: The first argument `indirect_values_map[value_info.arg_buffer_extra_info.argument_buffer]` will add an entry to `indirect_values_map` for `value_info.arg_buffer_extra_info.argument_buffer` even though that entry won't contain any values.

davidd-lunarg (approved):

PR #1019, opened 2023-03-01
Check for EI workloads when optimizing with --dxr

davidd-lunarg:
The --dxr optimization option can optimize both DXR and ExecuteIndirect workloads, but the optimization was only run if DXR was present. This change enables the optimization for captures that contain just EI workloads without DXR.

### DIFF
--- a/tools/optimize/dx12_optimize_util.cpp
+++ b/tools/optimize/dx12_optimize_util.cpp
@@ -53,7 +53,7 @@ void CreateResourceValueTrackingConsumer(
     const decode::Dx12OptimizationOptions&                      options)
 {
     std::string app_string = "GFXReconstruct Optimizer - analyzing file";
-    if (options.optimize_dxr_experimental)
+    if (options.optimize_resource_values_experimental)
     {
         app_string += ", image may appear corrupt";
     }
@@ -68,10 +68,10 @@ void CreateResourceValueTrackingConsumer(
 
     // Create the replay consumer.
     dx12_replay_consumer = std::make_unique<decode::Dx12ResourceValueTrackingConsumer>(
-        application, dx_replay_options, options.optimize_dxr_experimental);
+        application, dx_replay_options, options.optimize_resource_values_experimental);
     dx12_replay_consumer->SetFatalErrorHandler([](const char* message) { throw std::runtime_error(message); });
 
-    if (options.optimize_dxr_experimental)
+    if (options.optimize_resource_values_experimental)
     {
         dx12_replay_consumer->EnableReplayOfResourceValueCalls(false);
     }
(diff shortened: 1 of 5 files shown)


tools/optimize/dx12_optimize_util.cpp
```diff
@@ -104,22 +104,23 @@ bool BypassDxrOptimization(gfxrecon::decode::Dx12ObjectScanningConsumer& resref_
 {
     bool bypass_dxr_optimization = false;
 
-    bool contains_dxr_workload = resref_consumer.ContainsDxrWorkload();
-    bool is_dxr_optimized      = resref_consumer.ContainsOptFillMem();
+    bool contains_dxr_ei_workload = resref_consumer.ContainsDxrWorkload() || resref_consumer.ContainsEiWorkload();
+    bool is_dxr_optimized         = resref_consumer.ContainsOptFillMem();
```
bradgrantham-lunarg, line 108: The optimization is both DXR and EI, is that right?  Is there a better name for this variable and `bypass_dxr_optimization` then?  Like `is_optimized_for_address_within_buffer`?  Or something less verbose, maybe even `is_optimized` and `bypass_optimization`?

bradgrantham-lunarg (commented):
LGTM, but I did wonder about the remaining var names.

bradgrantham-lunarg:
I think this is good to go except maybe a thought towards the variable names I mentioned in my review.  @rurra-amd do you want a chance to add suggestions to this PR?

tools/optimize/dx12_optimize_util.cpp
rurra-amd: `contains_dxr_ei_workload` could become `bypass_rv_workload`

`is_dxr_optimized` could become `is_rv_optimized`

rv = resource value

Resource values represent that type of work. Using `bypass_optimization` is a bit misleading because there's also PSO optimization step.

rurra-amd (approved):
Looks good, I think the only thing needed is `git clang-format`. I noticed some param misalignment in `x12_object_scanning_consumer`

rurra-amd:
> I think this is good to go except maybe a thought towards the variable names I mentioned in my review. @rurra-amd do you want a chance to add suggestions to this PR?

Yep just did

tools/optimize/dx12_optimize_util.cpp
davidd-lunarg: Optimizer variables have been renamed from **dxr** to **resource_value**. Resource value is how GFXR internally refers to GPU VAs, GPU descriptors, shader IDs that an application can write to data and must be mapped for repaly.

The user-facing argument option for `gfxrecon-optimize` has not changed. It is `--dxr`.

bradgrantham-lunarg (approved):

PR #1021, opened 2023-03-06
Fix Multiple Resource Sharing the Same GPUVA Range

JerryAMD:
**The problem**
There is a bug if multiple resources with different sizes share the same GPUVA range. When mapping a given GPUVA to resource, GFXR's GpuVaMap returns the resource whose starting GPUVA is closest to the given address. However, the returned resource might not cover the entire range we were looking for.

**The solution**
Search the GpuVaMap with size if it's available, and return the resource whose range is covered by that size.


framework/graphics/dx12_gpu_va_map.cpp
```diff
@@ -61,7 +62,7 @@ void Dx12GpuVaMap::Remove(format::HandleId resource_id, uint64_t old_start_addre
     }
 }
 
-uint64_t Dx12GpuVaMap::Map(uint64_t address, format::HandleId* resource_id, bool* found) const
+uint64_t Dx12GpuVaMap::Map(uint64_t address, format::HandleId* resource_id, bool* found, uint64_t old_end_address) const
```
davidd-lunarg, line 65: Could the new argument to `Map` and `FindMatch` be renamed to `minimum_old_end_address` since an exact end address match is not required? `FindMatch` checks `old_end_address <= info.old_end_address`.

davidd-lunarg (approved):

davidd-lunarg (approved):

PR #1022, opened 2023-03-06
Remove competing sufrace extensions with wsi option

### LINKED ISSUE #998: Remove competing surface extensions with --wsi option
Currently --wsi option forces the specified wsi surface, which is helpful for playback on different platforms.
With this option enabled, related surface extension is included in enabled extensions list in VkInstanceCreateInfo, right next to other surface extensions used originally. However, if the surface extension used originally in the trace is not supported on the device, it will still trigger vkCreateInstance VK_ERROR_EXTENSION_NOT_PRESENT error.
To workaround this issue we could use --remove-unsupported option. Still, that's not an ideal solution as it would remove all other unsupported extensions as well. 
Using --wsi option to force a specific surface type should also remove other surface extensions already enabled in vkCreateInstance calls, making it easier to replay on different platforms. This should not impact any other unsupported extensions.
We're planning to implement that feature, any suggestions or ideas?

bartosz-muszarski-arm:
Fixes #998 
**Problem**
--wsi replay option adds relevant surface extension on instance creation without removing other surface extensions used in the trace. That makes it impossible to replay on different platforms without --remove-unsupported option.
**Solution**
Surface extensions different than the one specified with --wsi option are removed on instance creation, --remove-unsupported is no longer required.

### DIFF
--- a/framework/application/application.h
+++ b/framework/application/application.h
@@ -58,6 +58,8 @@ class Application final
 
     WsiContext* GetWsiContext(const std::string& wsi_extension, bool auto_select = false);
 
+    const std::string& GetWsiCliContext() const { return cli_wsi_extension_; }
+
     bool IsRunning() const { return running_; }
 
     void Run();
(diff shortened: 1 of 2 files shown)


bradgrantham-lunarg:
Hi, @bartosz-muszarski-arm , this failed in testing on our Windows systems.  Did you verify this against (1) an NVIDIA Windows installation, and (2) a 32-bit AMD Windows installation?

bradgrantham-lunarg (approved):

PR #1023, opened 2023-03-07
Inject adapter ID into metadata and calculate submits per adapter

igorbrsn:
**The problem**
When we run gfxrecon-info tool on the trace that was captured on the system with 2 GPUs we do not have information which of the adapters were active.

**The solution**
To determine the active adapter, we can track the Object_id by reading the adapters that recorded during capture (based on 'capture_id'), mapping it based on D3D12CreateDevice, CreateCommandQueue, and finally counting with ExecuteCommandLists. The ID with the largest count would indicate the active adapter. The Object_id could be associated with the LUID that belongs to the active adapter.

**Result and Test**
Test on dual GPU:

before:
before

after:
after

### DIFF
--- a/framework/decode/dx12_stats_consumer.h
+++ b/framework/decode/dx12_stats_consumer.h
@@ -299,7 +299,6 @@ class Dx12StatsConsumer : public Dx12Consumer
                                            HandlePointerDecoder<void*>* ppDevice)
     {
         GFXRECON_ASSERT(ppDevice != nullptr);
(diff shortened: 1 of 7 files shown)


bradgrantham-lunarg:
This PR has changes that won't compile on Linux or Android.

davidd-lunarg (approved):
Looks good. This PR will need to be updated with the latest from `dev` to resolve merge conflicts.

PR #1026, opened 2023-03-09
Fix JSON generated on Array2DMatrixToString

ceyusa:
A 2D matrix in valid json is [[1,1],[1,1]], while now is represented as [1,1][1,1] which is invalid and jq fails to parse.

This patch fixes it.

### DIFF
--- a/framework/util/to_string.h
+++ b/framework/util/to_string.h
@@ -295,10 +295,16 @@ inline std::string Array2DMatrixToString(size_t m,
                                          uint32_t      tabSize       = 4)
 {
     std::stringstream strStrm;
+    strStrm << '[';
     for (size_t i = 0; i < m; ++i)
     {
+        if (i)
+        {
+            strStrm << ',';
+        }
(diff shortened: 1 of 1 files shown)


andrew-lunarg:
Thanks for the clear PR. You can leave out the newline handling (` << GetNewlineString(toStringFlags)`) since Convert's JSON generation is at a moment of transition right now where we had decided to be JSON Lines only and formatted JSON can no longer be enabled, but remnants remain.

If you will be making any more substantial commits to Convert, just a heads-up that we are currently integrating and bringing to parity an external contribution of a similar tool, currently named Export in the PR's related to it.

andrew-lunarg:
Manually started CI Build Manual/job/gfxreconstruct/2495/ without trace regression testing for speed of completion given the triviality here.

andrew-lunarg:
Patch applied locally and confirmed to fix the JSON output as verified by feeding it through `jq` and seeing an error before but not after.

Thanks again for catching this!

PR #1028, opened 2023-03-10
Removed VkPerformanceCounterResultKHR ToString

andrew-lunarg:
The VkPerformanceCounterResultKHR type never appears in the API because it is referred to by a void*, for example by vkGetQueryPoolResults(). Our layer can therefore never see it and have a need to convert it. We lack the local context when converting it to get the type out of the void pointer so adding back in this currently dead code would be a small part of a hypothetical future PR which achieved that feat.
The extension this is part of has also not seen broad support (VK_KHR_performance_query).

### DIFF
--- a/framework/util/custom_vulkan_to_string.h
+++ b/framework/util/custom_vulkan_to_string.h
@@ -84,12 +84,6 @@ std::string ToString<VkLayerProperties>(const VkLayerProperties& obj,
                                         uint32_t                 tabSize);
 
 template <>
-std::string ToString<VkPerformanceCounterResultKHR>(const VkPerformanceCounterResultKHR& obj,
-                                                    ToStringFlags                        toStringFlags,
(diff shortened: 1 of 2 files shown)


framework/util/custom_vulkan_to_string.h
```diff
@@ -83,12 +83,6 @@ std::string ToString<VkLayerProperties>(const VkLayerProperties& obj,
                                         uint32_t                 tabCount,
                                         uint32_t                 tabSize);
 
-template <>
-std::string ToString<VkPerformanceCounterResultKHR>(const VkPerformanceCounterResultKHR& obj,
```
andrew-lunarg, line 87: Never called since no struct has a member of this type\* and it never appears in a function signature.  \*`VkPerformanceCounterResultKHR`

bradgrantham-lunarg (approved):

PR #1038, opened 2023-03-15
Assign ID3D12Object names during replay to assist debugging

rurra-amd:
**The problem**
ID3D12 objects only have names captured if the app called SetName() on them. If the app did not assign names, then debugging replay becomes harder since GFXR does not have any names recorded for them.

**The solution**
Add a new switch  `--dx12-override-object-names` for the replayer. Default is **off**. This will:
- Generate a `"gfxr_obj_N (creator)"` string when an object is created during replay
 (i.e. `"gfxr_obj_8 (ID3D12Device8_CreateCommittedResource2"`)
- Call `SetName()` using that string on all objects
- Bypass any object naming performed by the app

**Testing**
Verified that object names are generated and markers are emitted with a UE4 app.

### DIFF
--- a/framework/decode/dx12_replay_consumer_base.h
+++ b/framework/decode/dx12_replay_consumer_base.h
@@ -1,6 +1,6 @@
 /*
 ** Copyright (c) 2021-2022 LunarG, Inc.
-** Copyright (c) 2021-2022 Advanced Micro Devices, Inc. All rights reserved.
+** Copyright (c) 2021-2023 Advanced Micro Devices, Inc. All rights reserved.
 **
 ** Permission is hereby granted, free of charge, to any person obtaining a
 ** copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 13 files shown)


framework/decode/dx12_replay_consumer_base.h
```diff
@@ -163,15 +163,51 @@ class Dx12ReplayConsumerBase : public Dx12Consumer
...
+                // See if this is a D3D12Object
+                if (SUCCEEDED(iunknown->QueryInterface(IID_ID3D12Object, reinterpret_cast<void**>(&object))))
+                {
+                    const std::wstring constructed_name = ConstructObjectName(*p_id, call_id);
+
+                    HRESULT res = object->SetName(constructed_name.c_str());
+                    GFXRECON_ASSERT(res == S_OK);
+
+                    // QueryInterface() increases the refcount on success, so release here
+                    iunknown->Release();
```
davidd-lunarg, line 185: Should `Release()` be called on `object` instead of `iunknown`? Or maybe they're guaranteed to be the same object so it doesn't matter?

Alternatively a smart com ptr typedef `ID3D12ObjectComPtr` could be added to `dx12_util.h` and used to release automatically.

tools/replay/replay_settings.h
```diff
@@ -29,7 +30,8 @@ const char kOptions[] =
     "-h|--help,--version,--log-debugview,--no-debug-popup,--paused,--sync,--sfa|--skip-failed-allocations,--"
     "opcd|--omit-pipeline-cache-data,--remove-unsupported,--validate,--debug-device-lost,--create-dummy-allocations,--"
     "screenshot-all,--onhb|--omit-null-hardware-buffers,--qamr|--quit-after-measurement-"
-    "range,--fmr|--flush-measurement-range,--use-captured-swapchain-indices,--dcp,--discard-cached-psos";
+    "range,--fmr|--flush-measurement-range,--use-captured-swapchain-indices,--dcp,--discard-cached-psos,"
+    "--dx12-override-object-names";
```
davidd-lunarg, line 34: Can you add documentation for the new argument to `USAGE_desktop_D3D12.md` and also to the usage output below in this file `replay_settings.h`?

framework/decode/dx12_replay_consumer_base.h
rurra-amd: Done

tools/replay/replay_settings.h
rurra-amd: Done

davidd-lunarg (approved):

PR #1045, opened 2023-03-20
Improvements for launcher tool

rurra-amd:
**The problem**
When using the launcher tool to capture a D3D12 app, it tries to look for GFXR capture libraries where they were built by Visual Studio. This prevents it from working in a deployment setting, where the capture binaries live in a `d3d12_capture` folder. Also, the launcher wasn't correctly forwarding along the app's custom command line params to `CreateProcess()`

**The solution**
This PR does the following:
1. Allows the launcher to dynamically know where it resides and find the capture libraries beside it
2. Validates the location of said libraries before trying to capture
3. Call `CreateProcess()` correctly such that an application launches with the intended param setup

**Testing**
Verified that capture works both in a Visual Studio setting and in a deployment setting


tools/launcher/desktop_main.cpp
```diff
@@ -130,25 +151,47 @@ int main(int argc, const char** argv)
...
-                    &pi,
-                    interceptor_path.c_str());
+                char working_dir[MAX_PATH] = {};
+                GetCurrentDirectory(MAX_PATH, working_dir);
+
+                // Save off GFXR location in an environment variable, which is read later as our libs are loaded in
+                SetEnvironmentVariableA(gfxrecon::util::interception::kGfxrInstallEnv, working_dir);
+
+                // If we're using the launcher, we can assume detours will be used
+                SetEnvironmentVariableA("GFXRECON_USE_DETOURS_HOOKING", "true");
```
davidd-lunarg, line 161: Could you add some error handling in case `GetCurrentDirectory` and `SetEnvironmentVariableA` fail? Maybe just check for success and log a warning/error message if not.

rurra-amd: fixed

davidd-lunarg (approved):

PR #1046, opened 2023-03-21
Skip call to vkAcquireNextImage if capture status wasn't VK_SUCCESS

davidlunarg:


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5036,7 +5036,7 @@ VkResult VulkanReplayConsumerBase::OverrideAcquireNextImageKHR(PFN_vkAcquireNext
     // If image acquire failed at capture, there is nothing worth replaying as the fence and semaphore aren't processed
     // and a successful acquire on replay of an image that does not have a corresponding present to replay can lead to
     // OUT_OF_DATE errors.
-    if (original_result < 0)
+    if (original_result != VK_SUCCESS)
```
andrew-lunarg, line 5039: I think I get why the original line did what it did and what this is trying for. However, I think you want to also accept `VK_SUBOPTIMAL_KHR` for cases when the user changed bitdepth or when the window got resized. So perhaps:
```c++
if (original_result != VK_SUCCESS && original_result != VK_SUBOPTIMAL_KHR)
```
`VK_TIMEOUT` and `VK_NOT_READY` are also considered "success" but I don't think we want to treat them as such here because we care about results that give us an image we can render into.

andrew-lunarg (changes requested):

framework/decode/vulkan_replay_consumer_base.cpp
bradgrantham-lunarg: I thought I had requested that change yesterday as well, I guess I closed that tab without clicking "Request".  But, yes, I think this needs to check for `VK_SUBOPTIMAL_KHR` as well.

locke-lunarg (approved):

PR #1049, opened 2023-03-22
Add option ENABLE_EXCLUSIVE_LOCK

### LINKED ISSUE #1027: Thread issue for capture
https://github.com/locke-lunarg/gfxreconstruct/tree/thread
I have a branch that added a option `GFXRECON_ENABLE_EXCLUSIVE_LOCK` to set all API call using `AcquireExclusiveApiCallLock` to avoid thread issues.

Now, we have a thread issue for this PR
https://github.com/LunarG/gfxreconstruct/pull/995
This branch added a map to store pairs of handle and wrapper. When the applications run create/destroy multithread, it might conflict when inserting and erasing on the map. You can see a log, like `Create a duplicated Handle:.....`, when this conflict happens. The exclusive API lock could help it.

Also, `AcquireExclusiveApiCallLock` could serialize processes to rule out threading issues to help debug. Should we also add this on d3d12?

Plus, we have a crash issue for Wolf II. `ReadMe` in this shared folder descript the issue.
https://drive.google.com/drive/u/3/folders/1Om2M5YT1R76yxRInB0ThgoZn3bTenTb2
Whatever enabling exclusive lock or not, sometimes, it crashes before it runes `vkCmdBeginRenderPass` without running `vkBeginCommandBuffer`. The chance very is low. Sometimes, it's even 1/40. I'm not sure if it's related to thread. Plus, it doesn't happen if I don't enable GFXR or enable ValidationLayers. It looks like GFXR does something unknown to cause it.

locke-lunarg:
Add option GFXRECON_ENABLE_EXCLUSIVE_LOCK to set all Vulkan Capturing function using GetEnableExclusiveLock() to avoid thread conflict

closed: https://github.com/LunarG/gfxreconstruct/issues/1027

### DIFF
--- a/framework/generated/generated_vulkan_api_call_encoders.cpp
+++ b/framework/generated/generated_vulkan_api_call_encoders.cpp
@@ -88,7 +88,17 @@ VKAPI_ATTR void VKAPI_CALL DestroyInstance(
     VkInstance                                  instance,
     const VkAllocationCallbacks*                pAllocator)
 {
-    auto api_call_lock = VulkanCaptureManager::AcquireSharedApiCallLock();
+    auto force_command_serialization = VulkanCaptureManager::Get()->GetForceCommandSerialization();
+    std::shared_lock<CaptureManager::ApiCallMutexT> shared_api_call_lock;
+    std::unique_lock<CaptureManager::ApiCallMutexT> exclusive_api_call_lock;
+    if (force_command_serialization)
+    {
+        exclusive_api_call_lock = VulkanCaptureManager::AcquireExclusiveApiCallLock();
+    }
+    else
+    {
+        shared_api_call_lock = VulkanCaptureManager::AcquireSharedApiCallLock();
+    }
 
     CustomEncoderPreCall<format::ApiCallId::ApiCall_vkDestroyInstance>::Dispatch(VulkanCaptureManager::Get(), instance, pAllocator);
 
@@ -112,7 +122,17 @@ VKAPI_ATTR VkResult VKAPI_CALL EnumeratePhysicalDevices(
     uint32_t*                                   pPhysicalDeviceCount,
     VkPhysicalDevice*                           pPhysicalDevices)
 {
-    auto api_call_lock = VulkanCaptureManager::AcquireSharedApiCallLock();
+    auto force_command_serialization = VulkanCaptureManager::Get()->GetForceCommandSerialization();
+    std::shared_lock<CaptureManager::ApiCallMutexT> shared_api_call_lock;
+    std::unique_lock<CaptureManager::ApiCallMutexT> exclusive_api_call_lock;
+    if (force_command_serialization)
+    {
+        exclusive_api_call_lock = VulkanCaptureManager::AcquireExclusiveApiCallLock();
+    }
+    else
+    {
+        shared_api_call_lock = VulkanCaptureManager::AcquireSharedApiCallLock();
+    }
 
     bool omit_output_data = false;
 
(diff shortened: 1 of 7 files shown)


framework/generated/vulkan_generators/vulkan_api_call_encoders_body_generator.py
```diff
@@ -233,7 +233,11 @@ def make_cmd_body(self, return_type, name, values):
         if name == "vkCreateInstance" or name == "vkQueuePresentKHR":
             body += indent + 'auto api_call_lock = VulkanCaptureManager::AcquireExclusiveApiCallLock();\n'
         else:
-            body += indent + 'auto api_call_lock = VulkanCaptureManager::AcquireSharedApiCallLock();\n'
+            body += indent + 'auto enable_exclusive_lock = VulkanCaptureManager::Get()->GetEnableExclusiveLock();\n'
+            body += indent + 'std::shared_lock<CaptureManager::ApiCallMutexT> shared_api_call_lock;\n'
+            body += indent + 'std::unique_lock<CaptureManager::ApiCallMutexT> exclusive_api_call_lock;\n'
+            body += indent + 'if (enable_exclusive_lock) exclusive_api_call_lock = VulkanCaptureManager::AcquireExclusiveApiCallLock();\n'
+            body += indent + 'else shared_api_call_lock = VulkanCaptureManager::AcquireSharedApiCallLock();\n'
```
andrew-lunarg, line 240: I know this is generated but shouldn't it follow the coding guidelines a bit and use curly braces and spread over eight lines?
At least putting the branch bodies on their own lines would let a person set a breakpoint on them.

framework/generated/generated_vulkan_api_call_encoders.cpp
```diff
@@ -88,7 +88,11 @@ VKAPI_ATTR void VKAPI_CALL DestroyInstance(
     VkInstance                                  instance,
     const VkAllocationCallbacks*                pAllocator)
 {
-    auto api_call_lock = VulkanCaptureManager::AcquireSharedApiCallLock();
+    auto enable_exclusive_lock = VulkanCaptureManager::Get()->GetEnableExclusiveLock();
+    std::shared_lock<CaptureManager::ApiCallMutexT> shared_api_call_lock;
+    std::unique_lock<CaptureManager::ApiCallMutexT> exclusive_api_call_lock;
+    if (enable_exclusive_lock) exclusive_api_call_lock = VulkanCaptureManager::AcquireExclusiveApiCallLock();
+    else shared_api_call_lock = VulkanCaptureManager::AcquireSharedApiCallLock();
```
andrew-lunarg, line 95: See my point in the python generator.

andrew-lunarg (commented):
It might be interesting to see some timings for capture to know if adding the small amount of extra code at the top of every entrypoint is having a noticeable impact when the option is not enabled.
Perhaps use a simple app like vkcube with vsync off and a tiny window so the GPU is not the bottleneck. Just as a sanity check.

framework/generated/vulkan_generators/vulkan_api_call_encoders_body_generator.py
locke-lunarg: fixed it. Thank you.

framework/generated/generated_vulkan_api_call_encoders.cpp
locke-lunarg: fixed it. Thank you.

andrew-lunarg (approved):
Looks good to me if it doesn't slow down the default shared lock case too much.

locke-lunarg:
Renamed the option to `FORCE_COMMAND_SERIALIZATION` and added in USAGE docs.

bradgrantham-lunarg (approved):

PR #1068, opened 2023-03-24
Regress Required CMake Version for Ubuntu 18.04

andrew-lunarg:
EDIT: don't merge this, SDK support starts at 20.04 which brings us up to 3.16.3

Since the Vulkan SDK supports 18.04 we can't require CMake beyond the version that it provides.
<https://launchpad.net/ubuntu/+source/cmake>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,7 @@
 # Description: CMake script for framework util target
 ###############################################################################
 
-cmake_minimum_required(VERSION 3.17.0)
+cmake_minimum_required(VERSION 3.10.2)
 project(GFXReconstruct LANGUAGES CXX)
 
 set_property(GLOBAL PROPERTY TEST_SCRIPT ${CMAKE_ROOT})


juan-lunarg (changes requested):
:-1:

There is no reason to do this. We have already been shipping libraries with `3.17.2` minimum since the last SDK release.

Also given you aren't explicitly testing your CMake minimum as part of your CI process this is also just untested.

https://stackoverflow.com/q/70898802/19739129

bradgrantham-lunarg:
> 👎
> 
> There is no reason to do this. We have already been shipping libraries with `3.17.2` minimum since the last SDK release.
> 
> Also given you aren't explicitly testing your CMake minimum as part of your CI process this is also just untested.
> 
> https://stackoverflow.com/q/70898802/19739129

@juan-lunarg CMake on Ubuntu 20.04 is 3.16.  I could not build gfxreconstruct on my Ubuntu installation last night without editing CMakeLists.txt.  So what do you suggest?

bradgrantham-lunarg:
> > @juan-lunarg CMake on Ubuntu 20.04 is 3.16. I could not build gfxreconstruct on my Ubuntu installation last night without editing CMakeLists.txt. So what do you suggest?
> 
> I'm fine with gfx-reconstruct having a different CMake minimum if it wants to. As long as it explicitly tests the CMake minimum it's advertising.

There's a regression in our tree right now which is a higher priority.

> But there is no reason to restrict yourself to the CMake version that comes with your LTS release. The tarballs provided by CMake have extreme portability guarantees for users. Stronger guarantees than our tarball in fact.

Our instructions for building on Ubuntu only require `apt-get` for build prerequisites.  I don't want to require developers to use a tarball to get a non-standard version of CMake.

@andrew-lunarg, please update this to 3.16.3 to match the SDK requirement for 20.04 minimum and I'll approve it.

juan-lunarg:
> Our instructions for building on Ubuntu only require `apt-get` for build prerequisites. I don't want to require developers to use a tarball to get a non-standard version of CMake.

If you are using the docker images provided by github actions you are using the latest versions of cmake without knowing it.

https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md

> CMake 3.26.0

That's why we use get-cmake to test our minimum.

bradgrantham-lunarg:
> > Our instructions for building on Ubuntu only require `apt-get` for build prerequisites. I don't want to require developers to use a tarball to get a non-standard version of CMake.
> 
> If you are using the docker images provided by github actions you are using the latest versions of cmake without knowing it.

*Our instructions for building on Ubuntu only require `apt-get` for build prerequisites.*

juan-lunarg:
> _Our instructions for building on Ubuntu only require `apt-get` for build prerequisites._

I'm trying to say your CI is using CMake 3.26 because the runners provided by GitHub use the tarballs.

So if you want to ensure you don't break your users you need to explicitly grab the CMake version `3.16.3`

juan-lunarg:
Here is an example where the vulkan-loader advertized a cmake minimum version of `3.10.2` but used newer CMake features. So it broke the Conan build.

https://github.com/KhronosGroup/Vulkan-Loader/issues/1095

Now this will never happen again since the loader explitly tests it's minimum:

https://github.com/KhronosGroup/Vulkan-Loader/pull/1112

bradgrantham-lunarg:
That's very helpful, thank you.  Is that something you know how to change
in the yaml?  Do you have time to try to make that PR?

juan-lunarg:
> Do you have time to try to make that PR?

Sure :+1: 

> What I mean by that is that we provide build instructions for users, and we don't require people to use our GH actions artifacts. So if someone followed our build instructions, like I did last night, on Ubuntu 20.04, the build would fail. That's a regression.

Also I didn't realize this is what you meant by a regression when I originally answered. I understand this PR a lot better now.

juan-lunarg:
https://github.com/LunarG/gfxreconstruct/pull/1069

andrew-lunarg:
We don't need this anymore. Situation much improved by https://github.com/LunarG/gfxreconstruct/pull/1069, thanks @juan-lunarg :raised_hands:.

PR #1085, opened 2023-04-07
screenshot: Add support to export PNG format

MarkY-LunarG:
Add the ability for grabbing PNG formatted screenshots during replay and D3D12 capture.  The PNG format is using STB headers pulled in as a submodule with the additional tweak of using the Zlib we already pull in to increase the compression ratio versus the original image.

The current implementation supports 2 modes: threaded, non-threaded. For desktop replay, the replay will be multithreaded to 4 threads (defined in the CMakelists.txt file using the MAX_SCREENSHOT_THREADS define which is passed into the compiler definitions). For Android replay, the replay will be single-threaded to not put any additional memory pressure on the Android device.


CMakeLists.txt
```diff
@@ -198,6 +201,7 @@ endif() # WIN32
 
 find_package(LZ4)
 find_package(ZSTD)
+find_package(PNG)
```
bradgrantham-lunarg, line 204: Is this used for anything?

MarkY-LunarG: Good catch.  This was for using libPNG which I dropped.

MarkY-LunarG:
NOTE: I myself have not tested to make sure that the D3D side of things is working properly.  This will need to be done before I move this off of "draft".

framework/decode/screenshot_handler.cpp
```diff
@@ -38,10 +38,121 @@ static constexpr size_t kUnormIndex = 0;
...
+    {
+        bool valid = false;
+        {
+            std::unique_lock<std::mutex> lock(g_screenshot_write_mutex);
+            valid = g_screenshot_thread_info[index].valid;
+        }
+        // Do it this way so we're not sleeping while holding onto the lock
+        if (!valid)
+        {
+            std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
panos-lunarg, line 114: I see you chose the consumers to periodically sleep when there's no work and the producer to busy wait if all consumers are busy. Do you think conditional variables would be better?
The consumers can wait on a conditional variable and be awaken when new work arrives.
Also the producer can wait on another one if all consumers are busy.

MarkY-LunarG: There would be a race on the conditional variable check then set without additional synchronization primitives  It's not a bad idea.  I'll look into it.  Save the CPU from always running.

MarkY-LunarG: @panos-lunarg , please look over the updated design and let me know what you think.

panos-lunarg: Sorry this was a bit more complicated than I expected when I just read the code the first time.
I have pushed some tentative changes in my repo so you can take a look.
There are two conditional variables and 1 mutex.
I suggest introducing a ring buffer where the producer puts work and the consumers remove from it.
When a work slot is removed the data needs to be `std::move`d so the slot is free to be reused and the mutex to be held for as short as possible

I haven't tested my changes so most likely they won't work as expected.
Perhaps we are spending too much time in a non performance critical path and because I am to be blamed for this I have no strong opinion on this. Feel free to use whichever version you think best.

The rest of the code lgtm so you have my +1

Tentative patch: https://github.com/LunarG/gfxreconstruct/commit/b7d537070013cadddead148d9ffb4d38f411da9f

framework/util/image_writer.cpp
```diff
@@ -25,26 +25,57 @@
...
+#if defined(ENABLE_ZLIB_COMPRESSION) && defined(ENABLE_PNG_SCREENSHOT)
+#include <zlib.h>
+
+// This function re-formats the call to the Zlib compress2 function so that we can
+// use it in the STB PNG generation code.  Using the default STB PNG compression
+// resulted in a 20% reduction versus the original image in tests, while using the
+// zlib compress2 function resulted in a 40+% reduction versus the original image.
+uint8_t* gfxr_zlib_compress2(uint8_t* data, int32_t data_len, int32_t* out_len, int32_t quality)
+{
+    uint8_t*      target  = reinterpret_cast<uint8_t*>(malloc(data_len));
```
andrew-lunarg, line 37: I know typical platforms fail allocation lazily but it would be easy to check malloc's return value here since we already have a return path returning a null that the caller is expected to deal with.
`compress2` doesn't seem to have an immediate null check of its own (https://github.com/luvit/zlib/blob/master/compress.c#L38).

framework/util/CMakeLists.txt
```diff
@@ -177,9 +177,13 @@ endif()
 if (TARGET ZLIB::ZLIB)
     target_compile_definitions(gfxrecon_util
                                PUBLIC
-                                   ENABLE_ZLIB_COMPRESSION
+                                   ENABLE_ZLIB_COMPRESSION ENABLE_PNG_SCREENSHOT
```
andrew-lunarg, line 180: I admit it would look ugly to have one of these with a project prefix and the other not, but we have talked before about doing a sweep and prefixing build sys vars and macros which are currently not prefixed.

tools/replay/replay_settings.h
```diff
@@ -104,6 +104,7 @@ static void PrintUsage(const char* exe_name)
     GFXRECON_WRITE_CONSOLE("          \t\tAvailable formats are:");
     GFXRECON_WRITE_CONSOLE("          \t\t    %s\t\tBitmap file format.  This is the default format.",
                            kScreenshotFormatBmp);
+    GFXRECON_WRITE_CONSOLE("          \t\t    %s\t\tProtable Network Format file format.", kScreenshotFormatPng);
```
andrew-lunarg, line 107: "Portable Network Graphics"

framework/decode/screenshot_handler.cpp
```diff
@@ -38,10 +39,130 @@ static constexpr size_t kUnormIndex = 0;
...
+            {
+                GFXRECON_LOG_ERROR("Screenshot could not be created: failed to write BMP file %s", filename.c_str());
+            }
+            break;
+#ifdef ENABLE_PNG_SCREENSHOT
+        case util::ScreenshotFormat::kPng:
+            // For PNG format, we have to force the alpha channel to opaque, otherwise we get
+            // image artifacts on most image viewers from the transparency for anything that was
+            // rendered last to the image using transparency (think tree leaves).
+            uint32_t* working_data = reinterpret_cast<uint32_t*>(data);
```
andrew-lunarg, line 71: Bakes in a little-endian assumption. Covers ARM, IA32, x64, RISC-V, etc. Do we care?

andrew-lunarg (changes requested):
There are some comments and rhetorical questions in my review comments that are offered to take or leave, but also some typos that need a fix. It should be obvious which are which.

I roughly grokked the threading logic in there, but not to the level that I am saying it is right or wrong. Wouldn't the threading logic be simpler if we had a single-writer, multiple-reader queue wrapping the synchronisation? Images are pumped into it. Workers poll it for next image to write, sleeping inside the queue's code when it is empty. Shutdown is indicated by a token pumped into the queue rather than a global variable.

andrew-lunarg:
Adding PNG support and threading screenshotting seem orthogonal to me (I guess PNG saves slow things down to a crawl which motivated the threading). It would be easy to approve a quick boilerplate PR for PNG on the main thread which could be useful for low-res framebuffers maybe, get that merged, and then we could review a smaller PR with only the threading side of things in it afterwards, since the threading is enough to think about / review in one PR.

framework/decode/screenshot_handler.cpp
MarkY-LunarG: The problem with a ring buffer with File I/O is that it itself dependent on other factors and may be completed out of order from what you define.  So just because you trigger a file write for offset 0 doesn't mean it will complete before offset 1.  So you could be waiting longer for threads than you have to with a ring buffer.  I'll look over your suggestions in the patch, though.  Thanks.

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "external/Vulkan-Headers"]
 	path = external/Vulkan-Headers
 	url = https://github.com/KhronosGroup/Vulkan-Headers.git
+[submodule "external/stb"]
+	path = external/stb
+	url = https://github.com/nothings/stb.git
```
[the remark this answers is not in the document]
MarkY-LunarG: I'm not against that, but we spend so much time generating copyrights and licenses for our products automatically, we'd have to at least do more than just the header.  We'd still want a separate directory with the headers of interest + the License info.  So I thought it was just easier to stick the submodule in.

tools/replay/replay_settings.h
MarkY-LunarG: Oops

MarkY-LunarG:
> Adding PNG support and threading screenshotting seem orthogonal to me (I guess PNG saves slow things down to a crawl which motivated the threading). It would be easy to approve a quick boilerplate PR for PNG on the main thread which could be useful for low-res framebuffers maybe, get that merged, and then we could review a smaller PR with only the threading side of things in it afterwards, since the threading is enough to think about / review in one PR.

Yes, the problem was performance.  There are large stutters when saving the PNG files.  But that being said, I will strip out the threading changes from this and create a separate PR.

framework/decode/screenshot_handler.cpp
bradgrantham-lunarg: No, I don't think we care.  If someone wants to capture and replay Vulkan on a Mac SE/30, we'll fix it then.

.gitmodules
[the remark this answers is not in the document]
bradgrantham-lunarg: What about having an `external/stb` tree and checking in just `stb_image.h` and any copyright or `LICENSE` in there?  Then we don't have to worry about one more submodule update and dealing with submodule versioning.  (I often forget to do `git submodule update` when we update `external/Vulkan-Headers`)

framework/decode/screenshot_handler.cpp
MarkY-LunarG: Good catch.  Yes, I think we do since some Android targets may be big-endian.

CMakeLists.txt
```diff
@@ -42,6 +42,9 @@ set(GFXRECONSTRUCT_PROJECT_VERSION_MAJOR 0)
 set(GFXRECONSTRUCT_PROJECT_VERSION_MINOR 9)
 set(GFXRECONSTRUCT_PROJECT_VERSION_PATCH 19)
 
+# Replay settings that can be modified
+set(MAX_SCREENSHOT_THREADS 4)
+
```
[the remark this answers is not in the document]
MarkY-LunarG: I replaced all the ENABLE_XXXX ones (ZLIB, LZ4, PNG, ZSTD) with GFXRECON_ENABLE_XXXX.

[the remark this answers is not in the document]
MarkY-LunarG: Screenshot threads removed.

framework/util/image_writer.cpp
MarkY-LunarG: Done.

framework/decode/CMakeLists.txt
```diff
@@ -227,6 +227,13 @@ if (MSVC)
     endif()
 endif()
 
+if (TARGET ZLIB::ZLIB)
+    target_compile_definitions(gfxrecon_decode
+                               PUBLIC
+                                   ENABLE_PNG_SCREENSHOT
```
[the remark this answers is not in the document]
MarkY-LunarG: Done.

framework/util/CMakeLists.txt
MarkY-LunarG: Done.

framework/decode/screenshot_handler.cpp
MarkY-LunarG: Done.  Added a new `IsBigEndian` check in `framework/util/platform.h` and use that then flip the value based on big-endianness.  I believe it's correct if you want to double check.

MarkY-LunarG:
Oops, header still talks about threads.  That commit message will be remove in my next push.

framework/decode/screenshot_handler.cpp
bradgrantham-lunarg: Actually, I don't see why this couldn't be a `uint8_t*` and just set the appropriate byte to 0xFF.  Also, does this behave the same way as BMP?

bradgrantham-lunarg: I think we want to output whatever was in the framebuffer for alpha, so I think you should just omit this loop.

bradgrantham-lunarg: (The output is primarily for comparison and regression testing, not for viewing, so it should represent the actual framebuffer.)

MarkY-LunarG: If you output with alpha enabled on PNG the screen capture format is messed up.  Anywhere that has alpha (like leaves on a tree) result in the PNG having an alpha'd pixel.  That means it blends with the background of the image viewer and doesn't look correct.  This is why I ended up masking it out.

MarkY-LunarG: talos_alphamasked_versus_not

MarkY-LunarG: I could do the loop to modify it per byte, it will just take potentially longer.  But, that should require an endianness check.

bradgrantham-lunarg: It's okay if the alpha results in a screenshot that looks messed up.
Like I said, it's not in there for looking at.
It's in there for regression testing, for image comparison with a program.
If a user needs to look at it, they can use a tool that disables alpha.
Setting the alpha to 0xFF removes information from the screenshot.

Why would modifying the data per byte take longer?

bradgrantham-lunarg: *also* note that the BMP writer that's already in there *does* write alpha.

framework/decode/screenshot_handler.cpp
MarkY-LunarG: Ok.  I built a release build and the PNG output didn't make that big of an impact.  So I think the multithreading isn't needed anyway.

bradgrantham-lunarg: > I think the multithreading isn't needed anyway.

Sounds good.  But keep it in your back pocket - I would assume it speeds up both BMP and PNG by moving encoding, optionally compression, and I/O to another thread.

MarkY-LunarG: Agreed.  Yes, it does speed it up, but it's only a big concern if you're dumping out a bunch of consecutive screenshots in a row (which most people won't do regularly).

bradgrantham-lunarg (changes requested):

framework/util/platform.h
```diff
@@ -632,6 +632,16 @@ inline T AlignValue(T original)
     return (original + (alignment_t - 1)) & ~(alignment_t - 1);
 }
 
+inline bool IsBigEndian()
```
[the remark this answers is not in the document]
MarkY-LunarG: I swore I already removed this...

MarkY-LunarG:
Fixed CI clang-format failure.  [It was my clang-format that apparently broke it, ugh].

tools/replay/replay_settings.h
```diff
@@ -106,6 +106,7 @@ static void PrintUsage(const char* exe_name)
     GFXRECON_WRITE_CONSOLE("          \t\tAvailable formats are:");
     GFXRECON_WRITE_CONSOLE("          \t\t    %s\t\tBitmap file format.  This is the default format.",
                            kScreenshotFormatBmp);
+    GFXRECON_WRITE_CONSOLE("          \t\t    %s\t\tPortable Network Graphics Format file format.", kScreenshotFormatPng);
```
andrew-lunarg, line 109: The "Format" came back?  "Portable Network Graphics Format file format"

andrew-lunarg (changes requested):

tools/replay/replay_settings.h
MarkY-LunarG: Ugh.  Will fix

MarkY-LunarG:
@andrew-lunarg / @bradgrantham-lunarg , I think all your concerns have been addressed.

framework/decode/screenshot_handler.cpp
```diff
@@ -38,10 +39,41 @@ static constexpr size_t kUnormIndex = 0;
...
+inline void WriteImageFile(const std::string&     filename,
+                           util::ScreenshotFormat file_format,
+                           uint32_t               width,
+                           uint32_t               height,
+                           uint64_t               size,
+                           void*                  data)
+{
+    switch (file_format)
+    {
+        default:
```
andrew-lunarg, line 57: Would it not be better to have no default and let this be a compile-time warning (promoted to an error by warnings as errors)? Same with the DX12 case in another file. Sorry not to mention earlier, I just noticed.

I am not blocking on this, as interested in learning from contrary opinions as in asserting my own here!

framework/graphics/dx12_util.cpp
```diff
@@ -104,14 +105,40 @@ void TakeScreenshot(std::unique_ptr<graphics::DX12ImageRenderer>& image_renderer
...
-                            util::imagewriter::WriteBmpImage(filename,
-                                                             static_cast<unsigned int>(fb_desc.Width),
-                                                             static_cast<unsigned int>(fb_desc.Height),
-                                                             datasize,
-                                                             std::data(captured_image.data),
-                                                             static_cast<unsigned int>(pitch));
+
+                            switch (screenshot_format)
+                            {
+                                default:
```
andrew-lunarg, line 111: `screenshot_format` can only be unknown if `ScreenshotFormat` is changed. Don't we want to shout about this at compile time so that whoever changes `ScreenshotFormat` comes here and plumbs-in a handler for it before merging their future PR?

andrew-lunarg (approved):
Looks great to me. Open the docking bay doors: we have a PR coming in hot!

framework/decode/screenshot_handler.cpp
bradgrantham-lunarg: `kPng` in `framework/util/options.h` would also need to be wrapped with `GFXRECON_ENABLE_PNG_SCREENSHOT` or this code would error without `default`

framework/graphics/dx12_util.cpp
```diff
@@ -104,14 +105,40 @@ void TakeScreenshot(std::unique_ptr<graphics::DX12ImageRenderer>& image_renderer
...
+                                                                          datasize,
+                                                                          std::data(captured_image.data),
+                                                                          static_cast<unsigned int>(pitch)))
+                                    {
+                                        GFXRECON_LOG_ERROR(
+                                            "Screenshot could not be created: failed to write BMP file %s",
+                                            filename.c_str());
+                                    }
+                                    break;
+                                case gfxrecon::util::ScreenshotFormat::kPng:
```
andrew-lunarg, line 128: Does this need wrapping in `#ifdef GFXRECON_ENABLE_PNG_SCREENSHOT`?

framework/decode/screenshot_handler.cpp
andrew-lunarg: Presumably kPng could never reach this function if the code which emits it and is wrapped in `GFXRECON_ENABLE_PNG_SCREENSHOT` is not compiled-in to a build. But I'm either/or on this and I see where you are coming from.

framework/graphics/dx12_util.cpp
MarkY-LunarG: That define is not used up in the DX code level.  The `WritePngImage` function call itself though should return false if the #define is not set.  So it will return the error on line 137.

Issue #949, opened 2023-01-12
Option to Report Only Family 0, Queue 0 at Capture


andrew-lunarg:
Some users would like slightly more portable traces than we capture by default.

Many implementations have a single queue family with often a single queue in it that is used for all types of work. 

When capturing on a restricted implementation like this, we have a reasonable chance of replay on a different implementation with a more diverse set of queue families as they tend to have their most capable family be at index zero, so a trace using family zero, queue zero maps across very well.

When capturing on an implementation with many queue families, where the engine is written to make use of them, the resulting capture cannot replay on an implementation with a different configuration of queues unless the capabilities of the queues used accidentally happen to be supported by the queue families and queues within them at the same indexes.

Since most implementations have at least one queue in queue family zero that has most capabilities that most applications depend on for non-optional features,  and since most applications have the ability to work with a single queue to be able to target implementations that only expose one queue, if gfxr's capture layer had an option to expose only the first queue of the first family, captures made on any implementation with that option enabled would have the greatest chance of cross-implementation portable replay without requiring a sophisticated queue remapping / virtualisation feature within gfxr.

The feature suggested here would be turned on by an environment variable:

* Name: `GFXRECON_QUEUE_ZERO_ONLY`
* Type: `BOOL`
* Default `false`

The traces that resulted from capture with this option would not be useful for workload characterisation of a particular application on particular hardware, but they could represent a substantial stream of Vulkan commands, including leading edge rendering algorithms with multiple passes and compute, that painted the expected colours to the final frame. 

Separating this out from #856 as the core of that is a more complex problem that can be pursued in parallel, while this is a one-day PR affecting only the capture layer that could be used in the meantime and would be no burden to support over time.

locke-lunarg:
The first idea is to modify queue family setting to only zero on CreateDevice and return only zero of queue family on GetPhysicalDeviceQueueFamilyProperties during capturing. This idea only works for the first time capture from a game. Recapture doesn't work.

andrew-lunarg:
> Recapture doesn't work.

Yes, when I wrote "recapture them" in issue #1090 I meant repeat the capture from the application. My mistake for not spotting that I'd used a term we have a specific technical meaning for in this project :disappointed:. I realise that's  not much help when debugging a trace pulled off a drive and made by somebody else :-/.

locke-lunarg:
Bard left a idea in PR #1129

PR #1090, opened 2023-04-14
Set queue family index to 0 on creating buffer and image during replaying

locke-lunarg:
When the trimming capturing initizes, it needs to deal with buffers and images on ProcessBufferMemory and ProcessImageMemory. But when the buffers and images are created with queue family index not 0, it might happen DeivceLost on QueueWaitIdle or QueueSubmit of ProcessBufferMemory  or ProcessImageMemory.

related: https://github.com/LunarG/Projects/issues/528

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -4155,23 +4155,32 @@ VulkanReplayConsumerBase::OverrideCreateBuffer(PFN_vkCreateBuffer               
 
     VkResult                              result = VK_SUCCESS;
     VulkanResourceAllocator::ResourceData allocator_data;
-    auto                                  replay_buffer      = pBuffer->GetHandlePointer();
-    auto                                  capture_id         = (*pBuffer->GetPointer());
-    auto                                  replay_create_info = pCreateInfo->GetPointer();
-
+    auto                                  replay_buffer               = pBuffer->GetHandlePointer();
+    auto                                  capture_id                  = (*pBuffer->GetPointer());
+    auto                                  replay_create_info          = pCreateInfo->GetPointer();
+    VkBufferCreateInfo                    modified_create_info        = *replay_create_info;
+    uint32_t                              modified_queue_family_index = 0;
     // Check for a buffer device address.
     bool                uses_address         = false;
     VkBufferCreateFlags address_create_flags = 0;
     VkBufferUsageFlags  address_usage_flags  = 0;
+
+    if ((modified_create_info.sharingMode == VK_SHARING_MODE_CONCURRENT) &&
+        (modified_create_info.queueFamilyIndexCount > 0) && (modified_create_info.pQueueFamilyIndices != nullptr))
+    {
+        modified_create_info.queueFamilyIndexCount = 1;
+        modified_create_info.pQueueFamilyIndices   = &modified_queue_family_index;
+    }
+
     if (device_info->property_feature_info.feature_bufferDeviceAddressCaptureReplay)
     {
-        if ((replay_create_info->usage & VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT) ==
+        if ((modified_create_info.usage & VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT) ==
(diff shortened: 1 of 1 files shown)


locke-lunarg:
https://github.com/LunarG/Projects/issues/528
For this issue, it always failed during trimming capturing. The reason is `Device_Lost` on `WaitForQueue`. The queue's family index is 1. I figured out if I changed it to 0, the trimming will success. Honestly, I don't exact know why it happens `Device_Lost`, and I'm not sure if it's a good idea to change it to 0.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4155,23 +4155,32 @@ VulkanReplayConsumerBase::OverrideCreateBuffer(PFN_vkCreateBuffer
...
     VkResult                              result = VK_SUCCESS;
     VulkanResourceAllocator::ResourceData allocator_data;
-    auto                                  replay_buffer      = pBuffer->GetHandlePointer();
-    auto                                  capture_id         = (*pBuffer->GetPointer());
-    auto                                  replay_create_info = pCreateInfo->GetPointer();
-
+    auto                                  replay_buffer               = pBuffer->GetHandlePointer();
+    auto                                  capture_id                  = (*pBuffer->GetPointer());
+    auto                                  replay_create_info          = pCreateInfo->GetPointer();
+    VkBufferCreateInfo                    modified_create_info        = *replay_create_info;
```
andrew-lunarg, line 4161: We used to have a null check on line 4220 but now we trust this isn't null. Is that safe?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4155,23 +4155,32 @@ VulkanReplayConsumerBase::OverrideCreateBuffer(PFN_vkCreateBuffer
...
+    uint32_t                              modified_queue_family_index = 0;
     // Check for a buffer device address.
     bool                uses_address         = false;
     VkBufferCreateFlags address_create_flags = 0;
     VkBufferUsageFlags  address_usage_flags  = 0;
+
+    if ((modified_create_info.sharingMode == VK_SHARING_MODE_CONCURRENT) &&
+        (modified_create_info.queueFamilyIndexCount > 0) && (modified_create_info.pQueueFamilyIndices != nullptr))
+    {
+        modified_create_info.queueFamilyIndexCount = 1;
```
andrew-lunarg, line 4171: I don't think we can assert that all buffers with concurrent usage only use the first queue family. This is a lie for traces in which buffers are used by multiple families and violates the Vulkan specification. 
https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkBufferCreateInfo

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4283,34 +4290,42 @@ VulkanReplayConsumerBase::OverrideCreateImage(PFN_vkCreateImage
...
+    auto                                  capture_id                  = (*pImage->GetPointer());
+    auto                                  replay_create_info          = pCreateInfo->GetPointer();
+    VkImageCreateInfo                     modified_create_info        = *replay_create_info;
+    uint32_t                              modified_queue_family_index = 0;
 
-    VkResult result = allocator->CreateImage(
-        pCreateInfo->GetPointer(), GetAllocationCallbacks(pAllocator), capture_id, replay_image, &allocator_data);
+    if ((modified_create_info.sharingMode == VK_SHARING_MODE_CONCURRENT) &&
+        (modified_create_info.queueFamilyIndexCount > 0) && (modified_create_info.pQueueFamilyIndices != nullptr))
+    {
```
andrew-lunarg, line 4301: Same point as for buffers.

andrew-lunarg (changes requested):
I don't think we can just lie to the Vulkan implementation and say all these buffers and images are only used on queue family zero. It might help one particular troublesome trace on a particular implementation but it will break other traces and other implementations.

At a minimum, I think it should be hidden behind a flag and only turned on optionally, but even then it really doesn't make sense to me.

For traces with queue family trouble, I'd suggest implementing my issue https://github.com/LunarG/gfxreconstruct/issues/949 and then recapturing them so that they really do only use queue family zero.

locke-lunarg:
@andrew-lunarg  You're right. I agreed with you.

PR #1097, opened 2023-04-19
Introducing GFXR plugins

panos-lunarg:
The idea behind GFXR plugins is that one can extend GFXR’s functionality by adding new function calls that will be called before and after each Vulkan API call.

The plugins should be kept in a separate repository and their implementation should be independent of GFXR code base. GFXR declares an interface that each plugin needs to implement in order to communicate. This gives the following advantages:
- The plugin’s functionality can be a black box (proprietary) as only the compiled binary needs to be shared
- GFXR sources will not get polluted with functionality that perhaps only a few users might find useful.

This can be done both for capture and replay.


plugins/includes/capture/plugins_func_table_general.h
```diff
@@ -0,0 +1,75 @@
...
+                             FuncP*                           funcp)
+{
+    FuncP result = reinterpret_cast<FuncP>(gpa(handle, name));
+    if (result)
+    {
+        *funcp = result;
+    }
+}
+
+// clang-format off
```
andrew-lunarg, line 61: Why are we turning off clang-format in a non-generated file? If we don't like our clang format definition, let's change it, not dot our code with exceptions.

andrew-lunarg (commented):
I'll get back to a full review late week if it isn't merged by then.

plugins/includes/capture/plugins_func_table_general.h
panos-lunarg: Sorry this was a copy-paste leftover. Will fix

panos-lunarg:
All these are interesting points.

When we were thinking about the plugins we had in mind something like the Vulkan layers. Something that one can write, share the .so and then don't have to maintain it, even if GFXR changes over time.
So this introduces a restriction: The ABI between GFXR and the plugins must be stable across compilers/platforms. That means GXFR can not directly share it's internal state (because most things are stored in c++ stl structures). Then on the other hand we didn't want to add the extra overhead of converting the internal state so that it can be shared with the plugins. At least not yet. 

Having said that:

> 1. Change argument values during capture/replay

I guess you mean the plugin alters argument values in order to affect what gfxr captures to the file?
The easiest, fastest but dirties way would be to const_cast everything but this doesn't sound great. We'll need to think about that.

> 2. Add new blocks to the trace on capture

I think this conflicts with the stable ABI part. GFXR would have to share with the plugin the encoder class. Need to think about that too.

> 3. Skip call execution in pre handler

You mean the plugin skip execution? Each plugin can choose which functions to implement. Those that are omitted will be just noops.

> 4.Handle blocks other than api calls
> 5. Read parsed capture/replay cli options

I think these should be straightforward to add

per-mathisen-arm:
If the plugin doesn't get access to internal gfxr state, then how is writing a gfxr plugin easier/better/more flexible than writing your own Vulkan layer to do the same thing instead?

panos-lunarg:
Turns out plugins have a considerable performance impact. Abandoning the idea in favor of separate Vulkan layers

PR #1099, opened 2023-04-20
Use shared_mutex in VulkanStateTable

davidd-lunarg:
Reduce mutex/lock overhead by switching from mutex to shared_mutex. Use a unique_lock when modifying the table and a shared_lock when reading from the table.

### DIFF
--- a/framework/encode/vulkan_state_table_base.h
+++ b/framework/encode/vulkan_state_table_base.h
@@ -32,6 +32,7 @@
 #include <cassert>
 #include <functional>
 #include <map>
+#include <shared_mutex>
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(encode)
@@ -77,8 +78,8 @@ class VulkanStateTableBase
                      Wrapper*                                                    wrapper,
                      std::unordered_map<typename Wrapper::HandleType, Wrapper*>& map)
     {
-        const std::lock_guard<std::mutex> lock(mutex_);
-        const auto&                       inserted = map.insert(std::make_pair(handle, wrapper));
+        const std::unique_lock<std::shared_mutex> lock(mutex_);
+        const auto&                               inserted = map.insert(std::make_pair(handle, wrapper));
         return inserted.second;
     }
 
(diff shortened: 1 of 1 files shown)


framework/encode/vulkan_state_table_base.h
```diff
@@ -77,38 +78,30 @@ class VulkanStateTableBase
...
     {
-        const std::lock_guard<std::mutex> lock(mutex_);
+        const std::unique_lock<std::shared_mutex> lock(mutex_);
         return (map.erase(handle) != 0);
     }
 
     template <typename Wrapper>
-    Wrapper* GetWrapper(typename Wrapper::HandleType                                handle,
-                        std::unordered_map<typename Wrapper::HandleType, Wrapper*>& map)
+    Wrapper* GetWrapper(typename Wrapper::HandleType                                      handle,
```
andrew-lunarg, line 95: Is this "read access" If any thread can modify the the `Wrapper` pointed-to, and is that thread-safe if there is no further synchronisation? (maybe there is synchronisation elsewhere for the accesses to the fields of `Wrapper`s? Sorry, I should grep this but late here.

davidd-lunarg: I have restored the const and non-const methods with the change that the non-const method now takes a `const` argument for `map`--it would not be thread-safe to modify `map` while it is protected by a `shared_lock` and not a `unique_lock`.

> maybe there is synchronisation elsewhere for the accesses to the fields of Wrappers?

In general I do not think GFXR synchronizes wrapper state updates and instead relies on the application to make changes to objects in a thread-safe way. For example an application cannot record commands to a single `VkCommandBuffer` from multiple threads simultaneously and so GFXR doesn't add additional synchronization when updating tracked command buffer state in `VulkanStateTracker::TrackCommandExecution`.

andrew-lunarg: Okay, that makes sense :).

andrew-lunarg (approved):
It looks great. I get the point of it to protect the contended map rather than to protect the wrappers in the map now.

I'm not asking for benchmarking to be done before it goes in as it makes logical sense in any case, but out of interest do we get a measurable performance bump on any captures with this?

davidd-lunarg:
> do we get a measurable performance bump on any captures with this?

Yes. In one case FPS of the application doubled during capture. Some other cases also showed big improvements and some didn't change much. This is on Windows. I didn't test other platforms, but I wouldn't necessarily expect the same kind of performance difference there depending on those platforms's implementation of `mutex` and `shared_mutex`.

PR #1100, opened 2023-04-21
Added SplitString and RemoveWhitespace Functions

andrew-lunarg:
Tidied `GetMeasurementFrameRange` using `SplitString()`.

Also refactored duplicate code into `RemoveWhitespace()` and provided a unit test and documented its current (weird) behaviour. It is used where you might imagine trimming whitespace from left and right of a token but it actually removes all whitespace inside the string it is given too. So `"  Hello World!  "` becomes `"HelloWorld!"` rather than `"Hello World!"`.

A following PR will use these functions for other purposes but this PR stands alone as a refactor that pulls an algorithm out of the body of a function and makes it reusable and de-duplicates and documents.

If reviewing, ignore the commits from https://github.com/LunarG/gfxreconstruct/pull/1101. Hopefully they'll be merged to dev before this goes in.

### DIFF
--- a/framework/decode/dx12_ascii_consumer_base.h
+++ b/framework/decode/dx12_ascii_consumer_base.h
@@ -130,7 +130,7 @@ class Dx12AsciiConsumerBase : public Dx12Consumer
                 const auto& parametersStr = ObjectToString(to_string_flags_, tabCount, tabSize, toStringFunction);
                 for (auto c : parametersStr)
                 {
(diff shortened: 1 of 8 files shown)


bradgrantham-lunarg (approved):
This looks like a fine PR once it's rebased on top of andrew-lunarg:andy-fix-disable-debug-assert-popups-in-unit-tests-on-windows

PR #1101, opened 2023-04-21
Disable Debug Assert Popups in Unit Tests

andrew-lunarg:
We run unit tests on CI so cannot have interactive UI elements blocking them.

### DIFF
--- a/framework/application/CMakeLists.txt
+++ b/framework/application/CMakeLists.txt
@@ -84,7 +84,7 @@ if (${RUN_TESTS})
     if (MSVC)
         # Force inclusion of "gfxrecon_disable_popup_result" variable in linking.
         # On 32-bit windows, MSVC prefixes symbols with "_" but on 64-bit windows it doesn't.
-        if(NOT CMAKE_CL_64)
+        if(CMAKE_SIZEOF_VOID_P EQUAL 4)
             target_link_options(gfxrecon_application_test PUBLIC "LINKER:/Include:_gfxrecon_disable_popup_result")
         else()
             target_link_options(gfxrecon_application_test PUBLIC "LINKER:/Include:gfxrecon_disable_popup_result")
--- a/framework/application/CMakeLists.txt
+++ b/framework/application/CMakeLists.txt
@@ -78,8 +78,18 @@ common_build_directives(gfxrecon_application)
 if (${RUN_TESTS})
     add_executable(gfxrecon_application_test "")
     target_sources(gfxrecon_application_test PRIVATE
-            ${CMAKE_CURRENT_LIST_DIR}/test/main.cpp)
+            ${CMAKE_CURRENT_LIST_DIR}/test/main.cpp
+            ${CMAKE_CURRENT_LIST_DIR}/../../tools/platform_debug_helper.cpp)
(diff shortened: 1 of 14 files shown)


lunarpapillo:
Please include @juan-lunarg in reviews that are primarily CMake.

This looks awesome to me, but also has quite a bit of redundancy; I suspect it can be made more compact and easier to use by defining and using a CMake macro.  I'm sure Juan will know the best practice.

bradgrantham-lunarg:
> Please include @juan-lunarg in reviews that are primarily CMake.
> 
> This looks awesome to me, but also has quite a bit of redundancy; I suspect it can be made more compact and easier to use by defining and using a CMake macro. I'm sure Juan will know the best practice.

The CMake is probably not good - I just pounded on my keyboard until the first pass worked.  But since this fixes an issue that potentially hangs Windows CI, can we follow this PR with a PR that fixes the CMake style for this file and linker option in the tools and test targets?

framework/application/CMakeLists.txt
```diff
@@ -78,8 +78,18 @@ common_build_directives(gfxrecon_application)
...
     add_executable(gfxrecon_application_test "")
     target_sources(gfxrecon_application_test PRIVATE
-            ${CMAKE_CURRENT_LIST_DIR}/test/main.cpp)
+            ${CMAKE_CURRENT_LIST_DIR}/test/main.cpp
+            ${CMAKE_CURRENT_LIST_DIR}/../../tools/platform_debug_helper.cpp)
     target_link_libraries(gfxrecon_application_test PRIVATE gfxrecon_application)
+    if (MSVC)
+        # Force inclusion of "gfxrecon_disable_popup_result" variable in linking.
+        # On 32-bit windows, MSVC prefixes symbols with "_" but on 64-bit windows it doesn't.
+        if(NOT CMAKE_CL_64)
```
juan-lunarg, line 87: Use [CMAKE_SIZEOF_VOID_P](https://cmake.org/cmake/help/latest/variable/CMAKE_SIZEOF_VOID_P.html) to distinguish between 32/64.

`CMAKE_CL_64` is discouraged for new projects.

```suggestion
        if(CMAKE_SIZEOF_VOID_P EQUAL 4)
```

juan-lunarg (commented):
Are you using `ctest` to run your tests?

Due to CTest's use of libuv to launch test processes, all system debug popup windows are always blocked.

framework/application/CMakeLists.txt
bradgrantham-lunarg: Nope, `catch2`, and we've been discussing today that popup windows are not blocked.

juan-lunarg: But given switching to ctest would be a big change I'd say this is fine.

andrew-lunarg: I'll make the `CMAKE_SIZEOF_VOID_P` change.

bradgrantham-lunarg: I suggest leaving that for another PR.

andrew-lunarg: Oh, I just did it.
Should I back it out?

bradgrantham-lunarg (approved):

PR #1107, opened 2023-04-27
Replace stringstream With in or out Versions

### LINKED ISSUE #810: toascii - Change std::stringstreams to std::ostringstreams
We use large numbers of temporary input/output `std::stringstream`s when all we want is to output into them.

The class `std:stringstream` uses multiple inheritance.

Multiple

The class `std:ostringstream` only uses single inheritance, with hopefully faster function dispatch and a simpler internal layout to each object.

Single

Since we create them all over the place, we should use the simplest option we can get away with.

Related:
https://github.com/LunarG/gfxreconstruct/issues/601
https://github.com/LunarG/gfxreconstruct/issues/761

andrew-lunarg:
We use temporary input/output `std::stringstream`s when all we want is to either input from them (split string type applications) or output into them (formatting into a string) but never both.

The class `std:stringstream` uses _virtual_ multiple inheritance.

Multiple

The classes `std:ostringstream` `std:istringstream` only use single inheritance, with hopefully faster function dispatch and a simpler internal layout to each object.

Single

This PR uses the simplest option we can get away with in each case.

Fixes #810.

### DIFF
--- a/framework/decode/custom_dx12_ascii_consumer.cpp
+++ b/framework/decode/custom_dx12_ascii_consumer.cpp
@@ -41,7 +41,7 @@ void Dx12AsciiConsumer::Process_ID3D12RootSignatureDeserializer_GetRootSignature
     writeApiCallToFileInfo.pFunctionName = "GetRootSignatureDesc";
     auto returnValue = ToString(*return_value->GetPointer(), to_string_flags_, tab_count, tab_size);
     writeApiCallToFileInfo.pReturnValue = !returnValue.empty() ? returnValue.c_str() : nullptr;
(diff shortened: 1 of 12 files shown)


andrew-lunarg:
Made draft because while it is "done" and there is no reason why it would have broken the D3D12 Convert and Windows in general, in the absence of automated tests for that, I should manually confirm all is well.

andrew-lunarg:
Some of this has been done on this branch\* where related code was being touched anyway so this PR should be reworked to sweep up the remaining cases after that branch has become a PR and been merged.

\* https://github.com/andrew-lunarg/gfxreconstruct/tree/andy-enhance-better-d3d12-convert

andrew-lunarg:
I snuck it into other PRs like a crafty, creepy, code sneaker :).

A recursive grep shows it only exists under `/external` now: `rg std::stringstream`.

PR #1116, opened 2023-05-05
Panos gfxr base layer

panos-lunarg:


framework/encode/vulkan_state_writer.h
```diff
@@ -50,7 +50,7 @@ class VulkanStateWriter
     ~VulkanStateWriter();
 
     // Returns number of bytes written to the output_stream.
```
andrew-lunarg, line 52: Returns number of *blocks* written.

PR #1119, opened 2023-05-05
Add list of Vulkan extensions to info output

jacobv-nvidia:
Currently, in order to get information about which Vulkan extensions are requested in a capture, the primary workflow is to use the `gfxrecon-convert` utility to obtain the JSON output, and then separately inspect each instance/device CreateInfo structure. This can become quite inconvenient, especially for larger capture files.

This PR resolves the issue by having the `gfxrecon-info` tool perform this collection automatically and include the list of requested Vulkan extensions as part of its normal output.

An example of the additional `gfxrecon-info` output:
```
...
	Target API version:

Extension info:
        Instance extensions:
                VK_KHR_get_physical_device_properties2
                VK_KHR_surface
                VK_KHR_win32_surface
                VK_KHR_portability_enumeration
        Device extensions:
                VK_KHR_portability_subset
                VK_KHR_swapchain

Physical device info:
...
```

### DIFF
--- a/framework/decode/vulkan_stats_consumer.h
+++ b/framework/decode/vulkan_stats_consumer.h
@@ -35,6 +35,7 @@
 
 #include <set>
 #include <unordered_map>
+#include <unordered_set>
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
@@ -57,6 +58,9 @@ class VulkanStatsConsumer : public gfxrecon::decode::VulkanConsumer, public gfxr
     uint64_t                        GetMaxAllocationSize() const { return max_allocation_size_; }
     uint64_t                        GetAnnotationCount() const { return annotation_count_; }
     const std::vector<std::string>& GetOperationAnnotationDatas() const { return operation_annotation_datas_; }
+
+    const std::unordered_set<std::string>& GetInstanceExtensions() const { return instance_extensions_; }
+    const std::unordered_set<std::string>& GetDeviceExtensions() const { return device_extensions_; }
 
     const std::set<gfxrecon::format::HandleId>& GetInstantiatedDevices() const { return used_physical_devices_; }
     const VkPhysicalDeviceProperties*           GetDeviceProperties(gfxrecon::format::HandleId id) const
@@ -102,6 +106,13 @@ class VulkanStatsConsumer : public gfxrecon::decode::VulkanConsumer, public gfxr
         {
             auto create_info = pCreateInfo->GetPointer();
             auto app_info    = create_info->pApplicationInfo;
+
+            for (uint32_t i = 0; i < create_info->enabledExtensionCount; i++)
+            {
+                std::string extension(create_info->ppEnabledExtensionNames[i]);
+                instance_extensions_.insert(extension);
+            }
+
             if (app_info != nullptr)
             {
                 if (app_info->pApplicationName != nullptr)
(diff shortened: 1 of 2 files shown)


bradgrantham-lunarg:
It looks like this PR collects requested device extensions for all devices in the capture file into the one `set` and so prints a merged set of extensions.  I think a `gfxrecon-info` user would probably want these separated by device.  

It may be helpful for you to know that `convert` in `dev` and the 1.3.250 SDK outputs JSON by default.

You or any other developer can convert a capture file to JSON, then search for `vkCreateDevice`, and see the device extensions enabled at that time.

```
#assuming SDK 1.3.250
gfxrecon.py convert your-capture-file.gfxr
# load your-capture-file.json into your favorite text editor and search around
```

To go one level deeper, here are commands to get the JSON and then run `jq` (JSON query tool) command line to print all the device extensions for physical devices for `vkCreateDevice` in the JSON output.  (On Linux you may need to install `jq`; on Windows if you install `jq` it may be available as `jq-win32`.)
```
#assuming SDK 1.3.250
gfxrecon.py convert your-capture-file.gfxr
jq-win64.exe '.[] | select(.function.name == "vkCreateDevice") | select(.function.args.pCreateInfo.enabledExtensionCount > 0) | [.function.args.physicalDevice,.function.args.pCreateInfo.ppEnabledExtensionNames]' < your-capture-file.json
```
E.g. for my RTX 2070, I get:
```
[
  3,
  [
    "VK_KHR_swapchain",
    "VK_KHR_device_group"
  ]
]
```
`3` is GFXR's internal HandleId for the `VkPhysicalDevice` Vulkan handle.  If one wanted to disambiguate which device that was, one can also use jq to print the device name returned for `vkGetPhysicalDeviceProperties` for that HandleId like this (replacing the `3` with the HandleId printed as the physical device used in `vkCreateDevice`:
```
jq-win64.exe -r '[ .[] | select(.function.name == "vkGetPhysicalDeviceProperties") | select(.function.args.physicalDevice == 3) | [.function.args.pProperties.deviceName] ] | unique' < your-capture-file.json
```
The result is:
```
[
  [
    "NVIDIA GeForce RTX 2070 SUPER"
  ]
]
```
(Someone smarter than I am probably knows how to extract just that string from the inner array! @andrew-lunarg ? :smile:)

jacobv-nvidia:
@bradgrantham-lunarg I am aware of the `jq` method. I just figured that if it's a common enough task, then the functionality might be worth including in `gfxreconstruct` itself, rather than requiring an external tool and workflow.

However, you're right, the fact that it groups all devices into one is not robust. If you think this is at all worth pursuing, let me know, and I can refactor to enumerate extensions per instance/device. Otherwise, if you still don't think the feature should be included, this PR can probably just be closed out.

Since such output would be more verbose, I also think it would be better if it's placed behind a CLI flag, being disabled by default.

bradgrantham-lunarg:
@jacobv-nvidia - I think it is worth pursuing.  And I think it can be printed directly, no need to put it behind a CLI flag.  Presumably for most captures there will be just one instance and device.

jacobv-nvidia:
Okay, after far too long, I've finally gotten around to separating the requested device extension list by device.

Example output:

```
...
        Target API version: 4198400 (1.1.0)

Extension info:
        Instance extensions:
                VK_KHR_get_physical_device_properties2
                VK_EXT_debug_utils
                VK_NV_external_memory_capabilities
                VK_KHR_surface
                VK_KHR_win32_surface
        Device extensions:
                Device 0:
                        VK_EXT_descriptor_indexing
                        VK_EXT_pageable_device_local_memory
                        VK_KHR_swapchain
                        VK_EXT_memory_priority
		Device 1:
                        VK_KHR_swapchain

...
```

framework/decode/vulkan_stats_consumer.h
```diff
@@ -85,6 +89,13 @@ class VulkanStatsConsumer : public gfxrecon::decode::VulkanConsumer
         {
             auto create_info = pCreateInfo->GetPointer();
             auto app_info    = create_info->pApplicationInfo;
+
+            for (uint32_t i = 0; i < create_info->enabledExtensionCount; i++)
+            {
+                std::string extension(create_info->ppEnabledExtensionNames[i]);
+                instance_extensions_.insert(extension);
```
MarkY-LunarG, line 96: NOTE: This is adding ALL instance extensions listed in all created instances.  So if an application creates 5 instances, it mashes them all together and might not properly list what you desire.  For example, in an Android capture, if it captures `skiavk` data as well, the `skiavk` instance will enable ALL visible instance extensions so your list will simply list all extensions.

PR #1128, opened 2023-05-18
Expose GetBlockIndexGFXR() from capture layer

panos-lunarg:
GetBlockIndexGFXR() returns the current block index being processed by the capture manager.

### DIFF
--- a/framework/encode/capture_manager.cpp
+++ b/framework/encode/capture_manager.cpp
@@ -55,11 +55,13 @@ uint32_t                                                 CaptureManager::instanc
 std::mutex                                               CaptureManager::instance_lock_;
 thread_local std::unique_ptr<CaptureManager::ThreadData> CaptureManager::thread_data_;
 CaptureManager::ApiCallMutexT                            CaptureManager::api_call_mutex_;
+std::atomic<uint64_t>                                    CaptureManager::block_index_ = 0;
 
 std::atomic<format::HandleId> CaptureManager::unique_id_counter_{ format::kNullHandleId };
 
 CaptureManager::ThreadData::ThreadData() :
-    thread_id_(GetThreadId()), object_id_(format::kNullHandleId), call_id_(format::ApiCallId::ApiCall_Unknown)
+    thread_id_(GetThreadId()), object_id_(format::kNullHandleId), call_id_(format::ApiCallId::ApiCall_Unknown),
+    block_index_(0)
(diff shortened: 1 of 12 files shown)


locke-lunarg:
The code looks good to me. Could you please give me some examples about where we will need the capture block index? Thank you. Plus, this capture block index is different with the decoder block index since the capture's incudes system code. The decoder's has only Vulkan code. Should we care about it? or it doesn't matter?

locke-lunarg:
System code I mean something aren't Vulkan function calls. It doesn't matter. LGTM.

locke-lunarg (approved):

locke-lunarg:
I'm thinking that could we write this block index to the file? So the decoder could just read the block index to ensure both have the same block index.

panos-lunarg:
I am not sure if it's worth it. We can't store the id in the block header without breaking backward compatibility, so we'll have to store it as a metadata? This would essentially double the blocks written to the file.
Can you think of a better way?

locke-lunarg:
You are right. It's not worth. We shouldn't break backward compatibility.

locke-lunarg (approved):

framework/encode/capture_manager.cpp
```diff
@@ -823,21 +828,26 @@ void CaptureManager::WriteDisplayMessageCmd(const char* message)
...
         message_cmd.meta_header.block_header.type = format::BlockType::kMetaDataBlock;
         message_cmd.meta_header.block_header.size = format::GetMetaDataBlockBaseSize(message_cmd) + message_length;
         message_cmd.meta_header.meta_data_id =
             format::MakeMetaDataId(api_family_, format::MetaDataType::kDisplayMessageCommand);
-        message_cmd.thread_id = GetThreadData()->thread_id_;
+        message_cmd.thread_id = thread_data->thread_id_;
 
         CombineAndWriteToFile({ { &message_cmd, sizeof(message_cmd) }, { message, message_length } });
+
+        ++block_index_;
```
bradgrantham-lunarg, line 843: Is there something causing these two lines to be atomic?

bradgrantham-lunarg (approved):
This looks pretty good to me.  Later we can consider whether this should be exposed through some kind of "LUNARG" Vulkan extension, maybe even make it generic like VK_LUNARG_get_tooling_marker or some such.

PR #1129, opened 2023-05-18
Add option GFXRECON_QUEUE_ZERO_ONLY

### LINKED ISSUE #949: Option to Report Only Family 0, Queue 0 at Capture

locke-lunarg:
It forces to using only QueueFamilyIndex: 0 and queueCount: 1 on capturing to avoid replay error for unavailble VkQueue.

closed: https://github.com/LunarG/gfxreconstruct/issues/949
related: https://github.com/LunarG/Projects/issues/528

### DIFF
--- a/framework/encode/vulkan_capture_manager.cpp
+++ b/framework/encode/vulkan_capture_manager.cpp
@@ -675,6 +675,41 @@ VkResult VulkanCaptureManager::OverrideCreateDevice(VkPhysicalDevice            
     pCreateInfo_unwrapped->enabledExtensionCount   = static_cast<uint32_t>(modified_extensions.size());
     pCreateInfo_unwrapped->ppEnabledExtensionNames = modified_extensions.data();
 
+    VkDeviceQueueCreateInfo modified_queue_ci = {};
+
+    if (GetQueueZeroOnly())
+    {
+        // GetPhysicalDeviceQueueFamilyProperties force to return only FamilyIndex: 0 and queueCount: 1, so it shouldn't
+        // need to modifiy. If modified happens, it means the application doesn't follow
+        // GetPhysicalDeviceQueueFamilyProperties to set DeviceQueueCreateInfo. This modified could break the
+        // application setting. It might cause error.
+        GFXRECON_ASSERT(pCreateInfo_unwrapped->pQueueCreateInfos != nullptr);
+
+        if (pCreateInfo_unwrapped->queueCreateInfoCount > 1)
+        {
+            GFXRECON_LOG_WARNING("Because QUEUE_ZERO_ONLY is enabled, force queueCreateInfoCount to 1 on CreateDevice. "
+                                 "It might cause error.");
+            pCreateInfo_unwrapped->queueCreateInfoCount = 1;
+        }
+
+        modified_queue_ci = pCreateInfo_unwrapped->pQueueCreateInfos[0];
+
+        if (modified_queue_ci.queueFamilyIndex > 0)
+        {
+            GFXRECON_LOG_WARNING("Because QUEUE_ZERO_ONLY is enabled, force queueFamilyIndex to 0 on CreateDevice. It "
+                                 "might cause error.");
+            modified_queue_ci.queueFamilyIndex = 0;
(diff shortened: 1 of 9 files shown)


framework/encode/vulkan_capture_manager.cpp
```diff
@@ -1235,6 +1270,66 @@ VulkanCaptureManager::OverrideCreateRayTracingPipelinesKHR(VkDevice
...
+
+    GetInstanceTable(physicalDevice)
+        ->GetPhysicalDeviceQueueFamilyProperties(physicalDevice, pQueueFamilyPropertyCount, pQueueFamilyProperties);
+
+    if (GetQueueZeroOnly())
+    {
+        *pQueueFamilyPropertyCount = 1;
+        if (pQueueFamilyProperties != nullptr)
+        {
+            pQueueFamilyProperties[0].queueCount = 1;
```
locke-lunarg, line 1288: Force to return pQueueFamilyPropertyCount = 1 and queueCount = 1 on OverrideGetPhysicalDeviceQueueFamilyProperties.

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -675,6 +675,41 @@ VkResult VulkanCaptureManager::OverrideCreateDevice(VkPhysicalDevice
...
+                                 "might cause error.");
+            modified_queue_ci.queueFamilyIndex = 0;
+        }
+
+        if (modified_queue_ci.queueCount > 1)
+        {
+            GFXRECON_LOG_WARNING(
+                "Because QUEUE_ZERO_ONLY is enabled, force queueCount to 1 on CreateDevice. It might cause error.");
+            modified_queue_ci.queueCount = 1;
+        }
```
locke-lunarg, line 709: Force queueCreateInfoCount = 1, queueFamilyIndex = 0. and queueCount = 1 if it needs. But If modified happens, it means the application doesn't follow GetPhysicalDeviceQueueFamilyProperties to set DeviceQueueCreateInfo. This modified could break the application setting. It might cause error.

bradgrantham-lunarg: I agree this could break but if the application catches GPDQF like the change later in the file, I think the application at worst will exit with a failure because it could not find a queue suitable for submissions it will make later. I think this is a hammer to try and solve a problem and we should continue to think of it as such.  So this is okay with me for now.

(A later solution could be to try to "spoof" queues in an intelligent way so we only present some "virtual" set of queues and route submissions and memory requests and such ourselves.)

bradgrantham-lunarg (approved):

PR #1138, opened 2023-05-25
Handle VR frame delimiters

davidd-lunarg:
When a command buffer containing a VR end of frame debug marker is submitted, treat that as a frame delimiter. During capture, write end of frame markers to the capture file when this special end of frame condition is encountered and handle those frame markers in the file processor.


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -3194,6 +3194,43 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit(PFN_vkQueueSubmit func,
         GetDeviceTable(queue_info->handle)->QueueWaitIdle(queue_info->handle);
     }
 
+    // Check whether any of the submitted command lists buffers are frame boundaries.
+    if (screenshot_handler_ != nullptr)
+    {
+        bool is_frame_boundary = false;
+        for (uint32_t i = 0; i < submitCount; ++i)
```
andrew-lunarg, line 3201: I saw a loop looking for this elsewhere outside this file. Can the several loops be refactored into a reusable function with a self-documenting name?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -3329,6 +3366,44 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit2(PFN_vkQueueSubmit2 func,
         GetDeviceTable(queue_info->handle)->QueueWaitIdle(queue_info->handle);
     }
 
+    // Check whether any of the submitted command buffers are frame boundaries.
+    if (screenshot_handler_ != nullptr)
+    {
```
andrew-lunarg, line 3371: This isn't the other place I saw a search of command buffers. I saw it in another file too.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -6283,6 +6358,50 @@ VkResult VulkanReplayConsumerBase::OverrideGetAndroidHardwareBufferPropertiesAND
...
+void VulkanReplayConsumerBase::OverrideCmdDebugMarkerInsertEXT(
+    PFN_vkCmdDebugMarkerInsertEXT                             func,
+    CommandBufferInfo*                                        command_buffer_info,
+    StructPointerDecoder<Decoded_VkDebugMarkerMarkerInfoEXT>* marker_info_decoder)
+{
+    const VkDebugMarkerMarkerInfoEXT* marker_info = marker_info_decoder->GetPointer();
+    func(command_buffer_info->handle, marker_info);
+
+    // Look for the debug marker that identifies this command buffer as a VR frame boundary.
+    if (util::platform::StringContains(marker_info->pMarkerName, "vr-marker,frame_end,type,application"))
```
andrew-lunarg, line 6399: This string literal is used in encode too. Should we define it once in a file seen by both? `"vr-marker,frame_end,type,application")`

Also, does the match need to account for case and whitespace?

andrew-lunarg (commented):
Looking good!

andrew-lunarg:
Is there anything in CI that tests this? I think CI is currently just going to test that it doesn't break anything when the markers are not in use?

layer/trace_layer.cpp
```diff
@@ -51,7 +52,16 @@ const VkLayerProperties kLayerProps = {
...
+const std::vector<VkExtensionProperties> kDeviceExtensionProps = {
+    VkExtensionProperties{ "VK_EXT_tooling_info", 1 },
+    VkExtensionProperties{ VK_EXT_DEBUG_MARKER_EXTENSION_NAME, VK_EXT_DEBUG_MARKER_SPEC_VERSION }
+};
+
+std::unordered_set<std::string> kProvidedDeviceFunctions = { "vkCmdDebugMarkerBeginEXT",
+                                                             "vkCmdDebugMarkerEndEXT",
+                                                             "vkCmdDebugMarkerInsertEXT",
+                                                             "vkDebugMarkerSetObjectNameEXT",
+                                                             "vkDebugMarkerSetObjectTagEXT" };
```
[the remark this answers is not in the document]
bradgrantham-lunarg: `find(...) != ...end()` on a `set` is idiomatic and seems fine to me.  This array could get longer, so we'd have to worry about where the crossover point is, and you'd have to verify clang and MSVC also behave the same way.  While I commend your thoroughness including writing a test case in QuickBench (I learned that exists today!), we have a lot of work to do and I don't think this change important enough to microbenchmark something that occurs once at application startup.

I agree it could be `const`.

My only thought about this new variable would be that in the functions below, `kProvidedDeviceFunctions.count(in) > 0` is slightly more readable to me until `unordered_set::contains` in C++20.

layer/trace_layer.cpp
```diff
@@ -51,7 +52,16 @@ const VkLayerProperties kLayerProps = {
         GFXRECON_PROJECT_VERSION_DESIGNATION
 };
 
-const std::vector<VkExtensionProperties> kDeviceExtensionProps = { VkExtensionProperties{ "VK_EXT_tooling_info", 1 } };
+const std::vector<VkExtensionProperties> kDeviceExtensionProps = {
```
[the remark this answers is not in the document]
bradgrantham-lunarg: > Sorry, if this is const, why was it ever a vector?

For that you'd have to ask the original author.

A helpful suggestion might be "if we're not taking advantage of any methods of `std::vector`, can this be a const array?"  The reply might be, "you'd have to change the references to `kDeviceExtensionProps` below.  That may not be a very big change but would require more code changes more than just adding values to this vector."

My thought would be, is `vector` so distasteful because of performance or size concerns that we need to change it?  I think the answer would be "no" because there's only one of these, and you know I'm a fan of not changing things from their existing style unless there's a compelling reason.  But if your thought makes sense to David then I'm also okay with him changing it.

framework/decode/vulkan_replay_consumer_base.cpp
andrew-lunarg: My bad for cramming my review into late night tired time. Great that you have it covered.

davidd-lunarg:
> Is there anything in CI that tests this? I think CI is currently just going to test that it doesn't break anything when the markers are not in use?

That's right, CI doesn't currently have any tests that contain the VR frame markers.

bradgrantham-lunarg (approved):

PR #1139, opened 2023-05-30
Convert: Find "." from end of filename

### LINKED ISSUE #1131: `convert` changes `.gfxr*` to `.json`, which may create the wrong filename.
I have a capture here in `D:/Users/grantham/Downloads/g2.gfxr/g2.gfxr`.

Running `convert` on that file results in `D:/Users/grantham/Downloads/g2.json`.  Note that file is in the parent directory.

`convert` should change the *last* `.gfxr` to `.json`

locke-lunarg:
closed: https://github.com/LunarG/gfxreconstruct/issues/1131

### DIFF
--- a/tools/convert/main.cpp
+++ b/tools/convert/main.cpp
@@ -91,10 +91,11 @@ static std::string GetOutputFileName(const gfxrecon::util::ArgumentParser& arg_p
     else
     {
         output_filename   = input_filename;
-        size_t suffix_pos = output_filename.find(GFXRECON_FILE_EXTENSION);
-        if (suffix_pos != std::string::npos)
+        auto ext_pos      = output_filename.find_last_of(".");
+        auto path_sep_pos = output_filename.find_last_of(gfxrecon::util::filepath::kPathSepStr);
+        if (ext_pos != std::string::npos && (path_sep_pos == std::string::npos || ext_pos > path_sep_pos))
         {
-            output_filename = output_filename.substr(0, suffix_pos);
+            output_filename = output_filename.substr(0, ext_pos);
         }
         switch (output_format)
         {


tools/convert/main.cpp
```diff
@@ -91,9 +91,10 @@ static std::string GetOutputFileName(const gfxrecon::util::ArgumentParser& arg_p
     else
     {
         output_filename   = input_filename;
-        size_t suffix_pos = output_filename.find(GFXRECON_FILE_EXTENSION);
+        size_t suffix_pos = output_filename.find_last_of(GFXRECON_FILE_EXTENSION);
         if (suffix_pos != std::string::npos)
         {
+            suffix_pos -= std::strlen(GFXRECON_FILE_EXTENSION);
             output_filename = output_filename.substr(0, suffix_pos);
```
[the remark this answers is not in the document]
locke-lunarg: Thank you . Fixed.

tools/convert/main.cpp
```diff
@@ -90,11 +90,12 @@ static std::string GetOutputFileName(const gfxrecon::util::ArgumentParser& arg_p
     }
     else
     {
-        output_filename   = input_filename;
-        size_t suffix_pos = output_filename.find(GFXRECON_FILE_EXTENSION);
-        if (suffix_pos != std::string::npos)
+        output_filename = input_filename;
+        auto dot_pos    = output_filename.find_last_of(".");
+        auto slash_pos  = output_filename.find_last_of("\\");
```
andrew-lunarg, line 95: This is windows-specific. Slash goes the other way on Linux.

There is `kPathSep` and also path functions in `framework/util/file_path.h` like:
```
std::string InsertFilenamePostfix(const std::string& filename, const std::string& postfix);
std::string GetBasedir(const std::string& path);
std::string GetFilename(const std::string& path);
std::string GetFilenameStem(const std::string& path);
std::string GetFilenameExtension(const std::string& path);
```

andrew-lunarg: Sorry I forgot about them when I read your previous commit but this slash reminded me.

locke-lunarg: Thank you. Fixed.

tools/convert/main.cpp
```diff
@@ -91,10 +91,11 @@ static std::string GetOutputFileName(const gfxrecon::util::ArgumentParser& arg_p
     else
     {
         output_filename   = input_filename;
-        size_t suffix_pos = output_filename.find(GFXRECON_FILE_EXTENSION);
-        if (suffix_pos != std::string::npos)
+        auto ext_pos      = output_filename.find_last_of(".");
+        auto path_sep_pos = output_filename.find_last_of(gfxrecon::util::filepath::kPathSepStr);
+        if (ext_pos != std::string::npos && ext_pos > path_sep_pos)
```
bradgrantham-lunarg, line 96: If there's no path separator in the filename, like I run`gfxrecon-convert gfxrecon_capture_20230601T134916.gfxr`, this `if` will fail because `path_sep_pos` will be `npos` and on MSVC, gcc, and clang that's 18446744073709551615.  I get `gfxrecon_capture_20230601T134916.gfxr.json` but the function should return `gfxrecon_capture_20230601T134916.json` (no ".gfxr")

bradgrantham-lunarg (changes requested):

tools/convert/main.cpp
locke-lunarg: fixed

bradgrantham-lunarg (approved):

PR #1144, opened 2023-06-08
Fix agsInitialize call capture

feimao-gfxr:
**Problem**
Some applications may call agsInitialize() with "config" parameter with callback pointer values. At replay, these callback pointer values can not be resolved and will cause replay crash. 

**Solution**
At the capture time, we will force the call with config=nullptr. The agsInitialize call will also be captured as config being nullptr. Implemented in the call wrapper, because adding log there is more straightforward. 

**Result**
Tested with an application passing in "config" parameter containing valid pointer values. The replay is still OK.

### DIFF
--- a/framework/encode/custom_ags_wrappers.cpp
+++ b/framework/encode/custom_ags_wrappers.cpp
@@ -52,25 +52,15 @@ AMD_AGS_API AGSReturnCode agsInitialize(int                     agsVersion,
             shared_api_call_lock = D3D12CaptureManager::AcquireSharedApiCallLock();
         }
 
-        if (config != nullptr && (config->allocCallback != nullptr || config->freeCallback != nullptr))
-        {
-            GFXRECON_LOG_WARNING_ONCE(
-                "The application calls agsInitialize function with a non-null 'config' parameter. "
-                "GFXR forces the parameter to be a nullptr value. "
-                "The behavior of the application may change.");
-        }
-
-        AGSConfiguration* forced_config = nullptr;
-
         CustomWrapperPreCall<format::ApiCallId::ApiCall_Ags_agsInitialize_6_0_1>::Dispatch(
-            manager, agsVersion, forced_config, context, gpuInfo);
+            manager, agsVersion, config, context, gpuInfo);
 
-        result = manager->GetAgsDispatchTable().agsInitialize(agsVersion, forced_config, context, gpuInfo);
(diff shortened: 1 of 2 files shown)


framework/encode/custom_ags_wrappers.cpp
```diff
@@ -52,15 +52,25 @@ AMD_AGS_API AGSReturnCode agsInitialize(int                     agsVersion,
             shared_api_call_lock = D3D12CaptureManager::AcquireSharedApiCallLock();
         }
 
+        if (config != nullptr && (config->allocCallback != nullptr || config->freeCallback != nullptr))
```
bradgrantham-lunarg, line 55: This changes application runtime behavior.  I recommend this be implemented instead in replay, by dropping any callback pointers passed into those functions there.

bradgrantham-lunarg (changes requested):

framework/encode/custom_ags_wrappers.cpp
feimao-gfxr: Hi Brad, we actually tested both solutions with a real title, dropping the two callback pointers in capture or in replay. Both solutions worked. We chose to fix it in capture because we wanted to make the capture and the replay behaviors consistent. If you still think implementing this fix in replay is better, please let me know. Thanks.

bradgrantham-lunarg:
I think we've always preferred recording and honoring application behavior on capture as much as possible, and then changing calls on replay as necessary.

If capture changes the behavior of the app or NULLs out parameters, that discards information about the application's use of the API.  As an example, we save the Vulkan allocator callbacks even though they aren't valid in the replay context - we still get the information that callbacks were provided, which could be useful in a debugging context.  We discard them on replay, in `GetAllocationCallbacks()`.

This may be a case where we don't care enough, and you really know more about what information is important from AGS, but I would prefer being consistent with the Vulkan approach.

I think @davidd-lunarg is out for a few days, but I would value his thoughts on this when he gets back.

framework/encode/custom_ags_wrappers.cpp
feimao-gfxr: Thanks for the explanation. Now it is fixed at the replay side.

feimao-gfxr:
The change in capture has been reverted. The fix at the replay side works.

davidd-lunarg (approved):

PR #1146, opened 2023-06-12
Add wait before present option

bartosz-muszarski-arm:
Force wait on completion of queue operations for all queues before calling Present. This is needed for accurate acquisition of instrumentation data on some platforms.

### DIFF
--- a/USAGE_android.md
+++ b/USAGE_android.md
@@ -833,7 +833,7 @@ usage: gfxrecon.py replay [-h] [--push-file LOCAL_FILE] [--version] [--pause-fra
                           [--measurement-file DEVICE_FILE] [--quit-after-measurement-range]
                           [--flush-measurement-range] [-m MODE]
                           [--swapchain MODE] [--use-captured-swapchain-indices]
-                          [--use-colorspace-fallback]
+                          [--use-colorspace-fallback] [--wait-before-present]
                           [file]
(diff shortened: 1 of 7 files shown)


andrew-lunarg (approved):
Looks like a nice, tight little change.

Only question I have is whether it needs to account for alternate frame delimiters for platforms where images get to eyeballs without going through presents. Doing so could be a later edition though if it doesn't matter for the platforms the author cares about / is able to test on easily.

bartosz-muszarski-arm:
> Looks like a nice, tight little change.
> 
> Only question I have is whether it needs to account for alternate frame delimiters for platforms where images get to eyeballs without going through presents. Doing so could be a later addition though if it doesn't matter for the platforms the author cares about / is able to test on easily.

We can add this later, as you said it's not something we can test easily now

bradgrantham-lunarg:
> We can add this later, as you said it's not something we can test easily now

Can you test this against https://github.com/LunarG/gfxreconstruct/pull/1138?

charles-lunarg (approved):
The contents of the PR needed rebasing/cleanup to work with current dev. I did have to redo how the device handle was used to get the dispatch table to make it function. 

The design is reasonable - force a vkDeviceWaitIdle on every vkQueuePresentKHR.

PR #1147, opened 2023-06-13
Android capture trigger now appears in logcat

panos-lunarg:
When debug.gfxrecon.capture_android_trigger is switched between
true/false should now be visible in logcat

### DIFF
--- a/framework/encode/capture_settings.cpp
+++ b/framework/encode/capture_settings.cpp
@@ -685,11 +685,12 @@ CaptureSettings::ParseMemoryTrackingModeString(const std::string&               
     return result;
 }
 
+#if defined(__ANDROID__)
 CaptureSettings::RuntimeTriggerState
 CaptureSettings::ParseAndroidRunTimeTrimState(const std::string&                   value_string,
                                               CaptureSettings::RuntimeTriggerState default_value)
 {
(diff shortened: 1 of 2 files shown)


USAGE_desktop_Vulkan.md
```diff
@@ -124,6 +124,31 @@ The following command would be executed from the command line to set the
...
+The layer can be configured to detect memory modifications by marking the mapped
+memory as write protected, triggering an access violation when the application
+writes to the memory.
+The layer then uses a signal handler to intercept the signal generated by the
+access violation, where it removes the write protection, marks the modified
+memory page as dirty, and allows the application to continue.
+
+##### Disabling Debug Breaks Triggered by the GFXReconstruct Layer
+
+When running an application in a debugger with the layer enabled, the
```
bradgrantham-lunarg, line 143: Is this only true for Linux?

bradgrantham-lunarg (approved):
The change to USAGE_desktop_Vulkan.md seems to be tagging along, and is Linux specific.

I think it's pretty harmless but probably should indicate that the access violations that require a debugger setting are only on Linux.  With that change I think this should be merged.

panos-lunarg:
Sorry the changes in USAGE_desktop_Vulkan.md were committed by accident. I'll remove them

PR #1149, opened 2023-06-13
Disable AGS build if D3D12 build is off

davidd-lunarg:


bradgrantham-lunarg (approved):

davidlunarg:
I don't fully get it. When I observed this problem, I removed the contents of the build directory, and then ran the cmake command with D3D12_SUPPORT=OFF. I saw errors regarding AGS from cmake.  Not sure where GFXRECON_AGS_SUPPORT could have been cached if I removed the contents of the build directory. Maybe @davidd-lunarg's comment above isn't completely accurate, and the problem shows up only when the Windows SDK and D3D12 support are installed? Otherwise, the PR looks good.

davidlunarg (approved):

davidd-lunarg:
> I don't fully get it. When I observed this problem, I removed the contents of the build directory, and then ran the cmake command with D3D12_SUPPORT=OFF. I saw errors regarding AGS from cmake. Not sure where GFXRECON_AGS_SUPPORT could have been cached if I removed the contents of the build directory. Maybe @davidd-lunarg's comment above isn't completely accurate, and the problem shows up only when the Windows SDK and D3D12 support are not installed? Otherwise, the PR looks good.

The comment about caching did not capture the scope of this PR so I've removed it. The intent of this change is to always disable AGS in the build when D3D12 is disabled regardless of the value of `GFXRECON_AGS_SUPPORT`. Previously `util/frame/CMakeLists.txt` linked against AGS `if (${GFXRECON_AGS_SUPPORT})`. Now it uses `if (${GFXRECON_AGS_SUPPORT_FINAL})` where `GFXRECON_AGS_SUPPORT_FINAL` is only `ON` if both `D3D12_SUPPORT` and `GFXRECON_AGS_SUPPORT` are `ON`. I've verified that configuring a fresh build with the current `dev` with `-DD3D12_SUPPORT=OFF` fails with CMake errors, and configuring the PR branch with `-DD3D12_SUPPORT=OFF` successfully builds.

PR #1150, opened 2023-06-14
Imported xHook sources

panos-lunarg:
Temporarily importing xHook sources in a subfolder inside external.

Page guard manager uses xHook to intercept all calls to sigaction().
If a call to sigaction regarding SIGSEGV is removing GFXR's signal
handler is detected ,then it is not forwarded to the real sigaction.


bradgrantham-lunarg:
@panos-lunarg Do you think this is still an approach we should fully investigate and try to merge or is it supplanted by UFFD?

panos-lunarg:
I believe we can abandon this in favor of uffd

PR #1153, opened 2023-06-16
Loosen up frame range check

panos-lunarg:
Loosen up frame range check by instead of doing a strict equality check a less than or equal check of the starting frame


panos-lunarg:
Turns out this commit does not solve any problem. Rejecting

PR #1156, opened 2023-06-20
Add option to skip blit of virtual swapchain

marius-pelegrin-arm:
Add an option during replay to skip blitting virtual swapchain to real swapchain to remove the overhead of the blit. This overhead is noticeable for microbenchmarks and impacts hardware counter analysis.
The option is "--vssb" or "--virtual-swapchain-skip-blit".

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -181,6 +181,7 @@ VulkanReplayConsumerBase::VulkanReplayConsumerBase(std::shared_ptr<application::
             break;
         default:
             swapchain_ = std::make_unique<VulkanVirtualSwapchain>();
+            swapchain_->SetPerformanceMode(options.virtual_swapchain_skip_blit);
     }
     if (options_.enable_debug_device_lost)
     {
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -181,8 +181,13 @@ VulkanReplayConsumerBase::VulkanReplayConsumerBase(std::shared_ptr<application::
             break;
         default:
             swapchain_ = std::make_unique<VulkanVirtualSwapchain>();
-            swapchain_->SetPerformanceMode(options.virtual_swapchain_skip_blit);
     }
+
+    VulkanSwapchainOptions swapchain_options;
+    swapchain_options.skip_additional_present_blts = options.virtual_swapchain_skip_blit;
+    swapchain_options.select_surface_index         = options_.surface_index;
+    swapchain_->SetOptions(swapchain_options);
+
     if (options_.enable_debug_device_lost)
     {
         GFXRECON_LOG_WARNING("This debugging feature has not been implemented for Vulkan.");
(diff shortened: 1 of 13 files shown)


framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -411,6 +415,11 @@ VkResult VulkanVirtualSwapchain::QueuePresentKHR(PFN_vkQueuePresentKHR
     }
     queue = queue_info->handle;
 
+    if (performance_mode_)
+    {
+        return func(queue, present_info);
+    }
+
```
andrew-lunarg, line 422: Hi, thanks for another interesting PR from ARM.

I'm wondering what appears on the screen when this PR is in effect. Do you get any valid images (but not necessarily all frames or in the right order) out of it?

per-mathisen-arm: There would be no visual output.

bradgrantham-lunarg:
Do I understand this change correctly to only call `Present` if `performance_mode_` is true?  So it's equivalent to wrapping all the Virtual Swapchain logic with `if(!performance_mode_)`?  I think calling `Present` also causes all the implicit barriers and transitions that you'd expect for the chosen swapchain image, but bypassing all the Virtual Swapchain logic means it will wait on the wrong resources, wouldn't it, if the swapchain image index returned during replay is not the same as the index returned during capture?  I'm not sure this is what I expect from "skip blit".  Can you give me more of a summary of what you expect from any end-of-frame synchronization that would normally happen in `Present` and have you tried this with present modes MAILBOX and IMMEDIATE?

marius-pelegrin-arm:
Hi ! To fix this, here's a little change, to skip ONLY the blit. This way the synchronizations and image layouts still behave the same.

bradgrantham-lunarg:
Probably this failure in 3694 is an internal CI issue related to the other failures today.

However, @marius-pelegrin-arm @per-mathisen-arm  have you tried out `--swapchain offscreen` for replay, which doesn't `QueuePresent` or do you require creating the onscreen surfaces?

per-mathisen-arm:
We would like to replay it with a performance workload that is as close as possible to the real app, and I think offscreen is quite a bit off that ideal. As far as I understood we'd be missing the final blit to the swapchain entirely with offscreen. On the other hand, using virtual swapchain with this patch approximates very closely to the workload of running without virtual swapchain.

MarkY-LunarG:
Rebased on latest `dev` tree and cleaned up because we're starting to have multiple swapchain options spread around.

lunarpapillo:
CI failure seems to be due to a busy machine.  Re-running.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -180,8 +180,13 @@ VulkanReplayConsumerBase::VulkanReplayConsumerBase(std::shared_ptr<application::
             break;
         default:
             swapchain_ = std::make_unique<VulkanVirtualSwapchain>();
-            swapchain_->SetPerformanceMode(options.virtual_swapchain_skip_blit);
     }
+
+    VulkanSwapchainOptions swapchain_options;
+    swapchain_options.skip_additional_present_blts = options.virtual_swapchain_skip_blit;
```
bradgrantham-lunarg, line 186: Could `options` be renamed `replay_options`?  I had trouble reading these two lines because `options_` is similar to to `options`.

framework/decode/vulkan_swapchain.h
```diff
@@ -40,22 +40,31 @@ GFXRECON_BEGIN_NAMESPACE(decode)
...
 {
   public:
     virtual ~VulkanSwapchain() {}
 
     virtual void Clean();
-    virtual void SetPerformanceMode(bool performance_mode) {}
+
+    void SetOptions(VulkanSwapchainOptions& options)
+    {
+        memcpy(&swapchain_options_, &options, sizeof(VulkanSwapchainOptions));
```
bradgrantham-lunarg, line 58: I think this should use `=` and not `memcpy`.

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_replay_consumer_base.cpp
MarkY-LunarG: `options_` is the global class member version of `options`.  And both `options` and `options_` are variables of type `VulkanReplayOptions`.  So I can rename just this one function's `options` to `replay_options` but `replay_options` will then be used to initialize `options_`.  And I don't want to globally change `options_` to `replay_options_`  since then it's going to flow into many other files. 

That all being said, I think it might just be better to leave it as is.  What do you think @bradgrantham-lunarg ?

charles-lunarg (approved):
Tested it locally, and while the output is indeed garbage, I believe that is intentional.

I rebased on dev and added a missing --virtual-swapchain-skip-blit option in the vulkan USAGE markdown.

PR #1159, opened 2023-06-21
Fix --log-file option for Android during replay

marius-pelegrin-arm:
Add the `--log-file` and `--log-level` options for android at replay time. It is needed in android environments where logcat isn't available.

### DIFF
--- a/tools/replay/android_main.cpp
+++ b/tools/replay/android_main.cpp
@@ -84,6 +84,12 @@ void android_main(struct android_app* app)
 
     if (run)
     {
+        // Reinitialize logging with values retrieved from command line arguments
+        gfxrecon::util::Log::Settings log_settings;
+        GetLogSettings(arg_parser, log_settings);
+        gfxrecon::util::Log::Release();
+        gfxrecon::util::Log::Init(log_settings);
+
         std::string filename = kDefaultCaptureFile;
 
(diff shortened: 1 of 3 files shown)


tools/replay/android_main.cpp
```diff
@@ -84,6 +84,12 @@ void android_main(struct android_app* app)
 
     if (run)
     {
+        // Reinitialize logging with values retrieved from command line arguments
+        gfxrecon::util::Log::Settings log_settings;
+        GetLogSettings(arg_parser, log_settings);
+        gfxrecon::util::Log::Release();
+        gfxrecon::util::Log::Init(log_settings);
+
```
andrew-lunarg, line 92: It feels a little clunky that the standard codepath on Android even without this feature enabled will always be to init/release/init:
```cpp
gfxrecon::util::Log::Init();
//...
gfxrecon::util::Log::Init(log_settings);
```

marius-pelegrin-arm: I have followed the exact same path as for desktop (I just copied these lines from `desktop_main.cpp` and pasted them at the equivalent same place)

andrew-lunarg: You make a convincing argument!

bradgrantham-lunarg (approved):

PR #1161, opened 2023-06-23
Using userfaultfd in page guard manager

panos-lunarg:


per-mathisen-arm:
Exciting! Any data on the performance of this solution compared with mprotect? Why is it not enabled for Android?

panos-lunarg:
@per-mathisen-arm 
This is still experimental. It's not enabled on Android because I haven't managed to get it to work there properly yet.

framework/util/page_guard_manager.cpp
```diff
@@ -199,6 +209,377 @@ void PageGuardManager::InitializeSystemExceptionContext(void)
...
+}
+
+std::atomic<bool> PageGuardManager::uffd_stop_handler_thread = { false };
+
+bool PageGuardManager::UffdResetEntryProtection(MemoryInfo* memory_info)
+{
+    UffdUnregisterMemory(memory_info->shadow_memory, memory_info->shadow_range);
+
+    void* shadow_memory = memory_info->shadow_memory;
+    if (munmap(memory_info->shadow_memory, memory_info->shadow_range))
```
per-mathisen-arm, line 277: I don't think you are supposed to unmap the memory first here. This introduces a race condition between the unmap and the new mmap.

panos-lunarg: I wasn't aware of that and I am dealing with random segfaults that may be explained by this?
Any ideas?

per-mathisen-arm: I think you can just remove the unmap call. Redoing the mmap call with MAP_FIXED should work and throw out the previously mapped in pages.

panos-lunarg: Seems you're right, he random segfaults are gone!

However I am seeing random pages that although they trigger a pagefault event, when populating them `ioctl/uffdio_copy` triggers a `File exists` error. Indeed checking beforehand the pages with `mincore` these pages are already allocated.

I am not sure if these are responsible for the random visual artifacts (randomly corrupted textures and meshes) I am observing.
Also noting that I don't see neither the already allocated pages not the visual artifacts when trying this on a desktop game (recapturing a trace of a commercial title)

panos-lunarg: Also on desktop the `munmap` + `mmap` combination also seems to work.
Perhaps there's a difference in the strategy the two OSes allocate pages

framework/util/page_guard_manager.cpp
```diff
@@ -199,6 +209,377 @@ void PageGuardManager::InitializeSystemExceptionContext(void)
...
+        {
+            GFXRECON_LOG_ERROR("Unexpected read size\n");
+            break;
+        }
+
+        for (int i = 0; i < readres / sizeof(struct uffd_msg); ++i)
+        {
+            if (msg[i].event == UFFD_EVENT_PAGEFAULT)
+            {
+                UffdHandleFault(uffd_fd_, msg[i].arg.pagefault.address, msg[i].arg.pagefault.flags);
```
per-mathisen-arm, line 490: I think if you handle multiple events here, you need to use the DONTWAKE flags in the mode field for all but the latest event, otherwise you will get more race conditions, as the calling code wakes up and can do reads/writes that we presumably cannot catch while still processing more events.

panos-lunarg: Ah, good catch.

framework/util/page_guard_manager.cpp
```diff
@@ -205,33 +205,83 @@ PageGuardManager::PageGuardManager() :
...
+                                   MemoryProtectionMode protection_mode) :
     exception_handler_(nullptr),
     exception_handler_count_(0), system_page_size_(util::platform::GetSystemPageSize()),
     system_page_pot_shift_(GetSystemPagePotShift()), enable_copy_on_map_(enable_copy_on_map),
     enable_separate_read_(enable_separate_read), unblock_sigsegv_(unblock_SIGSEGV),
     enable_signal_handler_watcher_(enable_signal_handler_watcher),
     signal_handler_watcher_max_restores_(signal_handler_watcher_max_restores),
-    enable_read_write_same_page_(expect_read_write_same_page)
+    enable_read_write_same_page_(expect_read_write_same_page), protection_mode_(protection_mode)
 {
```
bradgrantham-lunarg, line 248: This constructor and the previous constructor used to be one line.  I think these two may now be candidates to share a function which does the shared work.  Or could one use the other as a delegated constructor?  I think delegated constructors came in with C++11.

bradgrantham-lunarg (changes requested):

PR #1163, opened 2023-06-26
Pull swapchain handling code into separate library

MarkY-LunarG:
Pull all of the swapchain handling code into a separate library with the intent of having it be stand-alone for use by other GFXReconstruct utilities.

**NOTES**:
  1) This library is called `compatibility` because I intend to move some other universal functionality into it eventually alongside of the swapchain functionality currently in there.
  2) This is currently still using a C++ interface.  Going forward, it might make sense to convert this to a C interface if this library is going to be released in a compiled version for other usage scenarios.


MarkY-LunarG:
**NOTE**: This library is called `compatibility` because I intend to move some other universal functionality into it eventually alongside of the swapchain functionality currently in there.

andrew-lunarg:
> 2\. Going forward, it might make sense to convert this to a C interface if this library is going to be released in a compiled version for other usage scenarios.

A C-API would make sense if third-party tools were really going to use it without building from source (we are open source: why do they need a binary?) or if we were going to release it entirely outside of GFXR on a different cadence from our own releases, but just to reuse a library internally that is built at the same time using the same compilers as the tools that use it, or for external users who build from source, it seems like a lot of faff for us and some overhead to thunk through a `C` interrop layer.

I'm guessing you are thinking of something like the hourglass API idiom. 
https://www.youtube.com/watch?v=PVYdHDm0q6Y
https://github.com/CppCon/CppCon2014/blob/master/Presentations/Hourglass%20Interfaces%20for%20C%2B%2B%20APIs/Hourglass%20Interfaces%20for%20C%2B%2B%20APIs%20-%20Stefanus%20Du%20Toit%20-%20CppCon%202014.pdf

MarkY-LunarG:
@andrew-lunarg, my current plan is to leave it as a C++ interface because of the overhead of converting all the C++ items (vectors, etc) into C could be a pain.  Yes, it would allow us to have a more ABI-resistant interface than C++, but I figure that we can just require anyone dumping out C++ code to eventually need the GFXReconstruct code base to build their samples.  I'm not completely against eventually going the C route, but I figured at the very least, this was a good lower impact first step.

bradgrantham-lunarg (changes requested):

bradgrantham-lunarg:
Raised to P1 because `tocpp` work  can and should take advantage of this to prevent the code diverging.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5525,8 +5753,15 @@ VulkanReplayConsumerBase::OverrideQueuePresentKHR(PFN_vkQueuePresentKHR
     // Only attempt to find imported or shadow semaphores if we know at least one around.
     if ((!have_imported_semaphores_) && (shadow_semaphores_.empty()) && (modified_present_info.swapchainCount != 0))
     {
-        result = swapchain_->QueuePresentKHR(
-            func, capture_image_indices_, swapchain_infos_, queue_info, &modified_present_info);
+        if (queue_info == nullptr || queue_info->handle == VK_NULL_HANDLE)
```
locke-lunarg, line 5756: In the begin of `OverrideQueuePresentKHR`, it has checked `queue_info != nullptr`.  We could also check  `queue_info->handle != VK_NULL_HANDLE` there, so we don't need to check them here again.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5108,8 +5294,31 @@ VkResult VulkanReplayConsumerBase::OverrideAcquireNextImageKHR(PFN_vkAcquireNext
 
             assert(replay_index != nullptr);
 
-            result = swapchain_->AcquireNextImageKHR(
-                func, device_info, swapchain_info, timeout, semaphore_info, fence_info, captured_index, replay_index);
+            VkSwapchainKHR swapchain = VK_NULL_HANDLE;
+            VkSemaphore    semaphore = VK_NULL_HANDLE;
+            VkFence        fence     = VK_NULL_HANDLE;
+            if (swapchain_info != nullptr)
```
locke-lunarg, line 5300: In the begin of `OverrideAcquireNextImageKHR`, it has checked `swapchain_info != nullptr`. We don't need to check again.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4774,6 +4819,112 @@ VkResult VulkanReplayConsumerBase::OverrideCreateDebugUtilsMessengerEXT(
...
+    va_end(args);
+}
+
+void LogInfoMessage(const char* message, ...)
+{
+    va_list args;
+    va_start(args, message);
+    GFXRECON_LOG_INFO(message, args);
+    va_end(args);
+}
```
locke-lunarg, line 4923: GFXRECON_LOG is in utils\loggings.h. I feel gfxrecon_util.lib is the lowest level library. compatibility could just use gfxrecon_util.lib.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4813,30 +4964,62 @@ VkResult VulkanReplayConsumerBase::OverrideCreateSwapchainKHR(
...
+        compatibility::SwapchainCreationInfo swapchain_creation_info{
+            device_info->handle,
+            physical_device,
+            replay_create_info,
+            { allocator,
+              CreateImageResource,
+              DestroyImageResource,
+              AllocateDeviceMemoryResource,
+              FreeDeviceMemoryResource,
+              BindImageResourceToDeviceMemoryResource },
```
locke-lunarg, line 4979: Decode is upper level library. Compatibility is lower level library. I don't know if it's common that lower level library use the function of upper level library?

locke-lunarg:
My opinion about library levels is decode -> compatibility -> graphics -> util. Upper level could link lower level, but lower level couldn't not link upper level. `util' is that every library could use, but it doesn't include vk/d3d code. `graphics` is similar with util, but it includes vk/d3d code. I know now `util` has many vk/d3d code.  I feel those code should move to `graphics`, but that's not this PR's topic.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4774,6 +4819,112 @@ VkResult VulkanReplayConsumerBase::OverrideCreateDebugUtilsMessengerEXT(
...
+// Callback functions required to work with the compatibility library since it has no direct
+// knowledge of the VulkanResourceAllocator.
+// ------------------------
+VkResult CreateImageResource(void*                        allocator_class,
+                             const VkImageCreateInfo*     create_info,
+                             const VkAllocationCallbacks* allocation_callbacks,
+                             VkImage*                     image,
+                             uintptr_t*                   resource_id)
+{
+    gfxrecon::decode::VulkanResourceAllocator* resource_allocator =
```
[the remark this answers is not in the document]
MarkY-LunarG: This is going to require more work because the VulkanResourceAllocator uses objects out of the decode path.  (The various `Info` structures and the HandleId as examples).  The plan is to eventually port it over, but this change is already massive as it is.

framework/decode/vulkan_replay_consumer_base.cpp
MarkY-LunarG: @locke-lunarg : This is going to be compiled into non-GFXR code (i.e. the generated "toCpp" output).  My original design did not intend for that source to have to link back into GFXR source since they may not have it on the generating system.  Is this important to maintain?  If so, it will likely require more changes.

PR #1166, opened 2023-06-26
Add the option to enable address sanitizer

MarkY-LunarG:
Add a CMake build option `GFXRECON_ENABLE_ASAN` to enable using address sanitizer during a build.  The option can be used during the CMake generation step by adding the following to the command-line:
  -DGFXRECON_ENABLE_ASAN=ON

By default, this option is currently set to OFF.

**NOTE:** This was copied from the way VVL did their changes.  I did run this and it helped me track down a recent memory leak in my changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,6 +118,80 @@ if(${GFXRECON_ENABLE_RELEASE_ASSERTS})
     add_definitions(-DGFXRECON_ENABLE_RELEASE_ASSERTS)
 endif()
 
+# Taking the sanitizer enablement solution from the `cpp-best-practices`
+# "cmake_template" Github Repo (which is unencombered open source).
+option(GFXRECON_ENABLE_SANITIZER_ADDRESS   "Enable address sanitization"   OFF)
+option(GFXRECON_ENABLE_SANITIZER_LEAK      "Enable leak sanitization"      OFF)
+option(GFXRECON_ENABLE_SANITIZER_MEMORY    "Enable memory sanitization"    OFF)
+option(GFXRECON_ENABLE_SANITIZER_THREAD    "Enable thread sanitization"    OFF)
+option(GFXRECON_ENABLE_SANITIZER_UNDEFINED "Enable undefined sanitization" OFF)
+
+# Now check support for each sanitizer option
+set(SUPPORTS_SANITIZE_ADDRESS FALSE)
+set(SUPPORTS_SANITIZE_LEAK FALSE)
+set(SUPPORTS_SANITIZE_MEMORY FALSE)
+set(SUPPORTS_SANITIZE_THREAD FALSE)
+set(SUPPORTS_SANITIZE_UNDEFINED FALSE)
+if((CMAKE_CXX_COMPILER_ID MATCHES ".*Clang.*" OR CMAKE_CXX_COMPILER_ID MATCHES ".*GNU.*") AND NOT WIN32)
+    set(SUPPORTS_SANITIZE_ADDRESS TRUE)
+    set(SUPPORTS_SANITIZE_LEAK TRUE)
+    set(SUPPORTS_SANITIZE_MEMORY TRUE)
+    set(SUPPORTS_SANITIZE_THREAD TRUE)
+    set(SUPPORTS_SANITIZE_UNDEFINED TRUE)
+elseif(MSVC)
+    set(SUPPORTS_SANITIZE_ADDRESS TRUE)
+endif()
+
+# Set up a list of enabled sanitizers
+# Warn when one or more are not compatible with other sanitizers.
+set(SANITIZER_LIST "")
+if (GFXRECON_ENABLE_SANITIZER_ADDRESS AND SUPPORTS_SANITIZE_ADDRESS)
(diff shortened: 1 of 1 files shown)


per-mathisen-arm:
Would it be possible specify through this option which sanitizer to use instead of hard-coding the address sanitizer?

MarkY-LunarG:
@per-mathisen-arm , can you give me more information?  We've used this same address sanitizer in all our products and I wasn't aware there was another.  Are you referring to something like Valgrind?  If there's another mechanism, I'd like to see more info.  Also, this currently doesn't affect Android, just the standard Windows/Linux build based on it's placement in the CMakefileLists.txt file.

bradgrantham-lunarg:
Some thoughts:
* This can affect ARM (and others) if they want to turn it on because they build and run on ARM linux.
* The given options don't work on Windows (`-fomit-frame-pointer` isn't an option for MSVC, it's `-Oy`)
  * Why is omitting the frame pointer added to this mode?  To better match what happens when optimization is enabled?
* If this is turned on, should we really run it in CI?  (@lunarpapillo )  so maybe it should just be part of debug builds.
* I think this PR could include a lot more in it if we really want to do compiler-assisted checking for us.  I think just copying VVL is a good start but not sufficient.

MarkY-LunarG:
I disagree.

1. It is optional and not enabled by default
2. It must work on Windows because Microsoft themselves [document this capabiliy](https://learn.microsoft.com/en-us/cpp/build/reference/fsanitize?view=msvc-170).  We're probably just missing something.
3. While, yes, it can affect ARM, it has to be enabled first.  And @per-mathisen-arm  wasn't complaining about the enabling of this, but expanding the options to include more capabilities.

I believe the benefit of having this option available as a tool could save other developers in the future.  I know when I've used this in the Vulkan loader, it has saved us pre-commit when making changes to underlying infrastructure.

bradgrantham-lunarg:
> It must work on Windows because Microsoft themselves [document this capabiliy](https://learn.microsoft.com/en-us/cpp/build/reference/fsanitize?view=msvc-170). We're probably just missing something.
That's `-fsanitize=`, not `-fomit-frame-pointer`

per-mathisen-arm:
You also have -fsanitize=undefined and -fsanitize=thread at least, which are useful. New ones seem to be added regularly. I would also compile with -fsanitize-undefined-trap-on-error which makes both compile and debugging easier, in my opinion. I am not worried about running this on Arm Linux - it should work as is and if not we can fix it later.

MarkY-LunarG:
> You also have -fsanitize=undefined and -fsanitize=thread at least, which are useful. New ones seem to be added regularly. I would also compile with -fsanitize-undefined-trap-on-error which makes both compile and debugging easier, in my opinion. I am not worried about running this on Arm Linux - it should work as is and if not we can fix it later.

I just saw this.  I only knew what I was copying.  I did see several references to building a "hardened" mode in CMake from other projects.  I'm going to look into that and @bradgrantham-lunarg's concerns about a failure to build on WIndows and revisit this.  Thanks @per-mathisen-arm .

CMakeLists.txt
```diff
@@ -118,6 +118,12 @@ if(${GFXRECON_ENABLE_RELEASE_ASSERTS})
     add_definitions(-DGFXRECON_ENABLE_RELEASE_ASSERTS)
 endif()
 
+option(GFXRECON_ENABLE_ASAN "Use address sanitization")
+if (GFXRECON_ENABLE_ASAN)
+    add_compile_options(-fsanitize=address -fno-omit-frame-pointer)
+    add_link_options(-fsanitize=address)
```
andrew-lunarg, line 124: It seems like you'd want to have different versions of this for MS and other platforms like this example shows:
https://cmake.org/cmake/help/latest/command/add_compile_options.html#example

Something like this using the forward slash option notation for windows:
```
if (MSVC)
    add_compile_options(/fsanitize=address /Oy-)
else()
# as before ...
endif()
```
I'm the last person to speak confidently on build system stuff though and I probably missed something (does cmake translate the options to what MSVC expects or is it clang being fed these options even on windows by some other mechanism perhaps...).

andrew-lunarg: Although what we might reall want is the CMake presets from CMake 3.19:
https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html
E.g.:
https://github.com/cpp-best-practices/cmake_template/blob/main/CMakePresets.json

andrew-lunarg:
It is great to have the option but don't we want to use it in the Github matrix like VVL does: https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/1b23b771d8ba76084cf70a622d3528115933449b
* Test that the sanitiser option successfully compiles with our code.
* We could at least run our unit tests with it turned on. 
```
      matrix:
...
        flags: [{c: "-fsanitize=address", ld: "-fsanitize=address"}, {c: "-fsanitize=thread", ld: "-fsanitize=thread"}]
...
        with:
          key: ${{ matrix.os }}-${{ matrix.config }}-${{ matrix.compiler.cc }}-${{ matrix.compiler.cxx }}-${{ matrix.cpp_std }}-${{ matrix.flags.c }}-${{ matrix.flags.ld }}
```

PR #1167, opened 2023-06-28
Add used resolutions info

bartosz-muszarski-arm:
gfxrecon-info reports resolutions used in vulkan traces based on data provided in vkCreateSwapchainKHR and vkCreateSharedSwapchainsKHR calls.

Sample output:
```
Exe info:
        Application exe name: 
        Application version: 0.0.0.0
        Application Company name: 
        Product name: 
File info:
        Compression format: LZ4
        Total frames: 11

Application info:
        Application name: ray_queries
        Application version: 0
        Engine name: Vulkan Samples
        Engine version: 0
        Target API version: 4198400 (1.1.0)
        Used resolutions: 1280x720

Physical device info:
        Device name: NVIDIA GeForce RTX 2060
        Device ID: 0x1e89
        Vendor ID: 0x10de
        Driver version: 2181955968 (0x820e0180)
        API version: 4206797 (1.3.205)

        Total allocations: 12
        Min allocation size: 3932160
        Max allocation size: 134217728

Pipeline info:
        Total graphics pipelines: 2
        Total compute pipelines: 0

```

### DIFF
--- a/framework/decode/vulkan_stats_consumer.h
+++ b/framework/decode/vulkan_stats_consumer.h
@@ -30,8 +30,8 @@
 #include "generated/generated_vulkan_struct_decoders.h"
 #include "generated/generated_vulkan_consumer.h"
 #include "util/defines.h"
-#include "vulkan/vulkan.hpp"
 #include "vulkan/vulkan_hash.hpp"
+#include "vulkan/vulkan_structs.hpp"
 
 #include "vulkan/vulkan.h"
 
--- a/framework/decode/vulkan_stats_consumer.h
+++ b/framework/decode/vulkan_stats_consumer.h
@@ -30,11 +30,14 @@
 #include "generated/generated_vulkan_struct_decoders.h"
 #include "generated/generated_vulkan_consumer.h"
 #include "util/defines.h"
+#include "vulkan/vulkan.hpp"
+#include "vulkan/vulkan_hash.hpp"
 
 #include "vulkan/vulkan.h"
 
 #include <set>
 #include <unordered_map>
+#include <unordered_set>
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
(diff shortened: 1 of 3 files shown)


bradgrantham-lunarg:
Could this be accomplished by a script parsing the output of `convert`?

bradgrantham-lunarg:
As an example, here's a `jq` line that extracts the imageExtent dimensions from `vkCreateSwapchainKHR`.
```
$ jq-win64.exe '.[] | select(.function.name == "vkCreateSwapchainKHR") | [.function.args.pCreateInfo.imageExtent.width,.function.args.pCreateInfo.imageExtent.height]' < gfxrecon_capture_20230428T153938.json
[
  500,
  500
]
```

bartosz-muszarski-arm:
Yes, the same data can be generated from a json-converted trace. I would argue this is still more convenient to use on a daily basis. We use the resolution information along with application name and total frame count which we get from gfxrecon-info already.

bradgrantham-lunarg (changes requested):
Looks good to me except if there are no swapchains created (e.g. `vulkaninfo`) then the output is `Used resolutions` and then nothing else.  I think there should be no output if `resolution.size() == 0`.

bradgrantham-lunarg:
> Yes, the same data can be generated from a json-converted trace. I would argue this is still more convenient to use on a daily basis. We use the resolution information along with application name and total frame count which we get from gfxrecon-info already.

Okay, sounds good to me.  I added a note about empty `resolutions` but with that change we'll merge.

framework/decode/vulkan_stats_consumer.h
```diff
@@ -30,11 +30,14 @@
 #include "generated/generated_vulkan_struct_decoders.h"
 #include "generated/generated_vulkan_consumer.h"
 #include "util/defines.h"
+#include "vulkan/vulkan.hpp"
```
bradgrantham-lunarg, line 33: Sorry, I didn't catch this on first pass.  This would be the first inclusion of VulkanHpp and I want to be a little more deliberate about adding it.  Can you just use VkExtent2D?

bradgrantham-lunarg (changes requested):
I had the one thought about vulkan.hpp.

Also your branch is failing in our internal CI because it's missing commit 32496a2170c178b143b48d103054130087080ade committed to `dev` on June 29.  If you rebase on current `dev` I think it will probably pass our internal CI.  Sorry, I think these are the last two items; thanks for being patient and sticking with us!

framework/decode/vulkan_stats_consumer.h
bradgrantham-lunarg: @bartosz-muszarski-arm if you haven't done this already I see how it smooths out the feature you added and I'm not in principle against it.  I'm doing a straw poll to see if anyone has any issues with including VulkanHpp.

andrew-lunarg: Could we at least reduce this to just including the structs header and not taking in the entire C++ wrapper?

```cpp
#include <vulkan/vulkan_structs.hpp>
```

andrew-lunarg: And also turn off every feature we don't need in the headers that has a macro that defaults to enabled. e.g. in our CMake configs for all projects define `VULKAN_HPP_NO_STRUCT_SETTERS`. We can then turn stuff on when we need it, where we need it.

framework/decode/vulkan_stats_consumer.h
```diff
@@ -57,6 +60,7 @@ class VulkanStatsConsumer : public gfxrecon::decode::VulkanConsumer, public gfxr
     uint64_t                        GetMaxAllocationSize() const { return max_allocation_size_; }
     uint64_t                        GetAnnotationCount() const { return annotation_count_; }
     const std::vector<std::string>& GetOperationAnnotationDatas() const { return operation_annotation_datas_; }
+    const auto&                     GetResolutions() const { return resolutions_; }
```
andrew-lunarg, line 63: I'm not a fan of `auto` here where it forces the future contributors to scroll down the file to know what the function they might want to call actually returns (`std::unordered_set<vk::Extent2D>`).

PR #1174, opened 2023-06-29
set assert, abort, and _CrtDebugReport to not open a dialog

bradgrantham-lunarg:


andrew-lunarg (approved):
I wasn't asked to review but as I was passing: looks good.

bradgrantham-lunarg:
Thanks for the review!

PR #1176, opened 2023-06-29
Add options to pass through fence status queries

marius-pelegrin-arm:
Add the command line options `--sgfs` (`--skip-get-fence-status`) and `--sgfr` (`--skip-get-fence-ranges`) to allow to skip calls to `vkWaitForFences` and `vkGetFenceStatus` under certain conditions at replay time.

For GPU performance measurements of applications we know but cannot modify, we sometimes want to "pack frames" when replaying by removing fences that we know unnecessary. For this reason, an option to skip fences at replay time is necessary.

There are three possible status that can be set using `--sgfs`:
- 0 - Don't skip: The default status, nothing particular happens.
- 1 - Skip unsuccessful: `vkWaitForFences`/`vkGetFenceStatus` is called only if the result at capture time was `VK_SUCCESS`.
- 2 - Skip all: `vkWaitForFences`/`vkGetFenceStatus` is never called and `VK_SUCCESS` is returned directly instead.

The `--sgfr` option specify at which frames these conditions apply. If `--sgfr` is not specified, they apply to all frames.


framework/application/application.h
```diff
@@ -84,6 +84,8 @@ class Application final
 
     void StopRunning() { running_ = false; }
 
+    const decode::FileProcessor* GetFileProcessor() const { return file_processor_; }
```
panos-lunarg, line 87: I think it would be better if this returned directly the frame number instead of the file processor

marius-pelegrin-arm: Done !

marius-pelegrin-arm:
Hi !

So after a check with @bartosz-muszarski-arm it appears there's an issue with `clang-format`. I have `clang-format` version `9.0.1-12` which should be good according to the contribution guidelines and when I run it on my commits it doesn't detect what makes the run fail.

Could you kindly take a look at it ?

andrew-lunarg:
> Hi !
> 
> So after a check with @bartosz-muszarski-arm it appears there's an issue with `clang-format`. I have `clang-format` version `9.0.1-12` which should be good according to the contribution guidelines and when I run it on my commits it doesn't detect what makes the run fail.
> 
> Could you kindly take a look at it ?

Apologies! The Github runner is using clang-format version: `14.0.0-1`, and we haven't updated the guidelines accordingly.

andrew-lunarg:
`clang-format-14 -i framework/application/application.h` fixes up an unrelated function as a side effect so if you could manually fixup the commit that would be cleaner:

```
-    uint32_t GetCurrentFrameNumber() const { return file_processor_->GetCurrentFrameNumber(); }
+    uint32_t GetCurrentFrameNumber() const
+    {
+
+        return file_processor_->GetCurrentFrameNumber();
+    }
```

Sorry that this process is somewhat messy.

lunarpapillo:
The LunarG CI failure was a system problem (NAS didn't mount).  I've requeued.

andrew-lunarg (changes requested):
This does seem like a reasonably unobtrusive feature to add but a little obscure. I think a high level description should be added to our documentation with the aim of educating a new user as to why they would use it. You have a first draft ready in the description of the PR, but we shouldn't lose that information by leaving it there: https://github.com/LunarG/gfxreconstruct/pull/1176#issue-1780502195

My apologies for letting this languish so long. If you can get it rebased on dev (not merge dev into it), I think we can move it towards landing in dev quickly.

marius-pelegrin-arm:
So I have rebased the changes on top of `dev` as requested.
The next commit is to "fix" what was broken by rebasing (principally `FrameRange` struct not existing anymore.
Then I added the description of the command options to the USAGE files.

Let me know if you feel more explanation is necessary. In which case I can make a little paragraph at the end of `USAGE_desktop_Vulkan.md`

andrew-lunarg:
> So I have rebased the changes on top of `dev` as requested. The next commit is to "fix" what was broken by rebasing (principally `FrameRange` struct not existing anymore. Then I added the description of the command options to the USAGE files.
> 
> Let me know if you feel more explanation is necessary. In which case I can make a little paragraph at the end of `USAGE_desktop_Vulkan.md`

Thank you for updating. I really think we do need some top-level explanation of why the feature exists and what it is intended for. This will be useful for the core team at LunarG who may be asked to explain it as much as for any potential users who are not in your team at ARM who are trying to understand what our tool does and which options are relevant to their work. I think you have a good first draft here: https://github.com/LunarG/gfxreconstruct/pull/1176#issue-1780502195.

lunarpapillo:
FYI, the build failure on # 3693 seems to have been due to an unrelated crashed GPU.  I've reset the machine and restarted the test.

charles-lunarg (approved):
Rebased on dev, fixed a single typo, and looks good.

While a niche use case, it certainly is useful to skip fence waits on upon reason.

PR #1177, opened 2023-06-29
Enabled Existing Operation Annotations in Info

andrew-lunarg:
The code to dump them was there but one step of
plumbing had gone missing.


tools/info/main.cpp
```diff
@@ -376,6 +376,7 @@ void GatherVulkanStats(const std::string& input_filename)
         stat_decoder.AddConsumer(&stat_consumer);
         file_processor.AddDecoder(&stat_decoder);
         vulkan_decoder.AddConsumer(&vulkan_stats_consumer);
+        file_processor.SetAnnotationProcessor(&vulkan_stats_consumer);
```
davidd-lunarg, line 379: Could this also be added to `GatherD3D12Stats` to output annotations for DX12 captures?

andrew-lunarg: I don't see why not. I'd like to get this one liner in for Vulkan just so the rest of the code that is already in place does something. We could follow-up for D3D12.

andrew-lunarg: Annotations should really be dealt with in API agnostic code so perhaps some refactoring is required rather than just duplication.

bradgrantham-lunarg: One comment I was thinking about adding to #1167 was that it should also do DX12 but I feel like that's likely to make that PR stall.

You're in a unique place here @andrew-lunarg that it wouldn't take you much to add DX12 too here.  (would it?)

andrew-lunarg: I guess its a half day or less so I can add it to my list.

andrew-lunarg: Funnily enough I think it was probably the D3D12 merge that killed the  `file_processor.SetAnnotationProcessor(&stats_consumer);` line from the original PR.

andrew-lunarg: I would consider that an enhancement and this PR a bugfix though.

bradgrantham-lunarg: Okay to make this PR a bugfix and add an enhancement to do DX12.

bradgrantham-lunarg (approved):

tools/info/main.cpp
bradgrantham-lunarg: @davidd-lunarg are you okay with this PR being a bugfix and adding another PR to add bits to GatherD3D12Stats?  If so, would you approve this PR or request additional changes?

davidd-lunarg (approved):

tools/info/main.cpp
andrew-lunarg: I have annotation processing factored out of vulkan condumer for Convert on a branch. I think that makes sense here too rather than adding annotations to the d3d side. There is a reason our former architect didn't plumb annotations through the decode and consumer dispatch chain: They are not api-specific.

PR #1179, opened 2023-06-30
Log error message if the user tries to capture AGS v5

feimao-gfxr:
**Problem**
The user may not pay attention to the current user guide and try to capture AGS in an application using AGS v5 with current GFXR implementation supporting only AGS version 6.x.x. 

**Solution**
Expose agsInit() in the AGS capture DLL. This API function doesn't exist in AGS version 6.x.x. If this function is called by the application during the capture, we will log an error message to ask the user to exclude the AGS capture and return AGS_FAILURE to the application.   

**Result**
Tested with an applications using AGS v5.x.x. The error message was logged.


lunarpapillo:
FYI, the first LunarG CI failure was due to a system GPU reset.  We don't think this PR was to blame, as this happens to the machine occasionally.  I've requeued the job.

lunarpapillo:
Again, the failure on `tcubumes3` was a system failure, this time a NAS failing to mount.  I've restarted.

framework/encode/custom_ags_wrappers.cpp
```diff
@@ -469,5 +469,17 @@ AMD_AGS_API AGSReturnCode agsDriverExtensionsDX12_SetMarker(AGSContext*
...
+AMD_AGS_API void agsInitNotSupported()
+{
+    static bool check_version = true;
+    if (check_version)
+    {
+        GFXRECON_LOG_ERROR("AGS version mismatch for GFXR capture: GFXR doesn't support this version of "
+                           "amd_ags_x64.dll. Please remove the capture DLL and rename the original DLL back.");
+
+        check_version = false;
+    }
```
davidd-lunarg, line 481: `GFXRECON_LOG_ERROR_ONCE` can be used to log an error only once. It uses the same `static bool` mechanism.

layer/ags_capture/dll_main.cpp
```diff
@@ -115,6 +115,27 @@ EXTERN_C AGSReturnCode AMD_AGS_API gfxrecon_agsInitialize(int
...
 
+// When this API function is called, it means the app is trying to use AGS v5.x.x versions
+EXTERN_C AGSReturnCode AMD_AGS_API gfxrecon_agsInit(AGSContext**            context,
+                                                    const AGSConfiguration* config,
+                                                    AGSGPUInfo*             gpuInfo)
+{
+    if (gfxrecon::Initialize())
+    {
+        GetDispatchTable().agsInitNotSupported();
+    }
```
davidd-lunarg, line 126: Is `gfxrecon::Initialize()` and `AgsDispatchTable::agsInitNotSupported();` required, or can the error message be logged here directly and the new dispatch table entry `AgsDispatchTable::agsInitNotSupported();` be removed?

feimao-gfxr: Rodrigo asked the same question. Here is my answer: "I tested. If put the log code here, it won't do the log, actually doing nothing. To make it doing log here, I have to add logging setup solely for this module, reading environment parameters and calling log::init(settings) and all that. I think this way is cleaner."

framework/encode/custom_ags_wrappers.cpp
feimao-gfxr: Done

layer/ags_capture/dll_main.cpp
davidd-lunarg: I see. Does it keep the desired logging behavior if the dispatch table entry `agsInitNotSupported` is removed and this code is changed to this:
```
if (gfxrecon::Initialize())
{
    GFXRECON_LOG_ERROR_ONCE("AGS version mismatch for GFXR capture: GFXR doesn't support this version of "
                           "amd_ags_x64.dll. Please remove the capture DLL and rename the original DLL back.");
}
```
The reason I ask is because if the `agsInitNotSupported` table entry is not required, removing it cleans up the code in this file.

feimao-gfxr: If we change the code only to: 
```
if (gfxrecon::Initialize())
{
    GFXRECON_LOG_ERROR_ONCE("AGS version mismatch for GFXR capture: GFXR doesn't support this version of "
                           "amd_ags_x64.dll. Please remove the capture DLL and rename the original DLL back.");
} 
```, 
it won't log anything. We have to add logging setup code in this file to make it log. If we set up logging at this proxy API point, in this file, we can remove `agsInitNotSupported`. I looked at other proxy API dlls, there is no logging code. I thought the original GFXR designer meant to keep all the logging code at the encode level. So, now I am not sure which way is better. Both ways have pro and con.

Comment (davidd-lunarg) [reply layer/ags_capture/dll_main.cpp:126]: I see, thanks for the explanation--I was thinking that logging would be setup with `gfxrecon::Initialize`. I think this is OK as-is.

File: framework/encode/custom_ags_wrappers.cpp [hunk 3]
```diff
@@ -469,5 +469,17 @@ AMD_AGS_API AGSReturnCode agsDriverExtensionsDX12_SetMarker(AGSContext*
...
+AMD_AGS_API void agsInitNotSupported()
+{
+    static bool check_version = true;
+    if (check_version)
+    {
+        GFXRECON_LOG_ERROR_ONCE("AGS version mismatch for GFXR capture: GFXR doesn't support this version of "
+                           "amd_ags_x64.dll. Please remove the capture DLL and rename the original DLL back.");
+
+        check_version = false;
+    }
```
davidd-lunarg, line 481: Can the `check_version` variable and if-statement be removed now that `GFXRECON_LOG_ERROR_ONCE` is used? The macro for `GFXRECON_LOG_ERROR_ONCE` contains a static var and if-statement to log only once.

feimao-gfxr: Right! Fixed.

davidd-lunarg (approved):

bradgrantham-lunarg:
> CI gfxreconstruct build # 3001 failed.

I think this failure was likely caused because your PR branch was branched from `dev` before 32496a2170c178b143b48d103054130087080ade, and in the meantime we've also upgraded a CI machine that requires change 32496a2.  If you rebase to current `dev` and push, that LunarG CI run probably will succeed.

PR #1184, opened 2023-07-03
Add trace preload option

bartosz-muszarski-arm:
Adds --preload-measurement-range option allowing to preload a frame range specified with --measurement-frame-range from the trace file into a continuous, expandable buffer, in order to mitigate the impact of read file commands on performance measurements.

### DIFF
--- a/framework/decode/preload_file_processor.cpp
+++ b/framework/decode/preload_file_processor.cpp
@@ -115,12 +115,10 @@ bool PreloadFileProcessor::ProcessBlocks()
                         }
                         else
                         {
-                            success = ProcessFunctionCall(block_header, api_call_id);
-                            if (is_frame_delimiter)
+                            bool should_break = false;
+                            success           = ProcessFunctionCall(block_header, api_call_id, should_break);
+                            if (should_break)
                             {
-                                // Make sure to increment the frame number on the way out.
-                                ++current_frame_number_;
-                                ++block_index_;
                                 break;
                             }
                         }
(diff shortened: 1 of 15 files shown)


framework/decode/preload_file_processor.cpp
```diff
@@ -0,0 +1,369 @@
...
+
+    while (success)
+    {
+        success = ContinueDecoding();
+
+        if (success)
+        {
+            success = ReadBlockHeader(&block_header);
+
+            if (status_ != PreloadStatus::kRecord)
```
MarkY-LunarG, line 78: A comment here describing why you are not getting the block info on recording would be good.

framework/decode/preload_file_processor.cpp
```diff
@@ -0,0 +1,369 @@
...
+                    uint64_t           frame_number = 0;
+
+                    success = ReadBytes(&marker_type, sizeof(marker_type));
+
+                    if (success)
+                    {
+                        if (status_ == PreloadStatus::kRecord)
+                        {
+                            const auto is_frame_delimiter = IsFrameDelimiter(block_header.type, marker_type);
+                            preload_buffer_.Reserve(sizeof(block_header) + block_header.size);
```
MarkY-LunarG, line 215: I almost think these blocks should be in a helper function like:   `bool ReadPreloadBuffer(size_t header_size, size_t ignore_size)` or something like that.  Then you only have to pass in the things like `sizeof(marker_type)` here and appropriate items above .   And it will make this cleaner.

MarkY-LunarG (commented):
Just a few questions.

framework/decode/preload_file_processor.cpp
charles-lunarg: I concur, I will add that and refactor code to use it.

charles-lunarg (approved):
After much work, the PR is in good shape to be included.

The extending of FileProcessor revealed that it needed work cleaning up the interface & implementation in order to make future extensions more manageable. While there is possibly a less involved way to include the preload capabilities, the current changes in the PR are a best effort at cleaning things up while not rewriting the entire FileProcessor class.

PR #698, opened 2022-05-17
WIP: implement xdg-shell

okias:
Issues:
 - do not use wayland-scanner (manually imported xdg-shell-* files
 - crashes

I'm currently dropping it here to not forget about it :wink: and ofc in case someone would be interested in finishing/fixing it faster than me.

PR #1190, opened 2023-07-10
Support for XDG Shell on Wayland

marius-pelegrin-arm:
Hi !

Inspired by the PR #698, here is a working support for XDG Shell on Wayland. It is not a replacement of the classic Wayland Shell by XDG Shell but an addition : both are now supported, but XDG Shell is the default choice if both are available, as it is recommended in the Wayland official documentation.

This PR works and can be tested, it is however **not finished yet as choices need to be made** about the link model. Currently, in GFXReconstruct, the libwayland is only loaded at runtime if necessary. However, Wayland XDG Shell structs, functions and constants are not available in libwayland. They are instead expected to be directly included in the project using sources generated by `wayland-scanner`.
The issue is these structs, functions and constants depends on the one available in libwayland. So what happens is that if libwayland is not available at link time, the compilation fails.

I see several solutions to this, and there are probably several that I do not see, but here is, in my opinion, the best solution. We could, as I did temporarily in the PR, link against libwayland at link time instead of loading it at runtime. If we compile the XDG Shell sources, it means that cmake found Wayland anyway, so there's no more assumption on the platform or the user environment if we link to libwayland when it is found. The only "loss" is a performance cost - the libwayland will be loaded at startup even if Wayland is finally not used.

If this solution is chosen, the `WaylandLoader` class becomes useless (as everything is already linked, thus loaded when the program starts) and should be removed. I can do that work if this solution is chosen.

Of course, let me know if I haven't seen another issue that linking against libwayland could cause or if you have a better solution to this issue.


marius-pelegrin-arm:
Hello !

So the solution I propose in the end is given with these two commits.

The first commit is the one that I proposed here several months ago (with some changes for merge conflicts). Wayland is linked statically, but an issue found with this solution is that the generated binary will not run on a machine that does not have `libwayland.so` if it was compiled on a machine that has `libwayland.so`.

The second commit solves this issue and goes in the opposite direction. The main issue here is that `wayland-scanner` generates sources assuming that libwayland is linked statically: this is the main error.
So I added `wayland-protocols` as a submodule of GFXReconstruct, and created a python script `generate_wayland.py` that parses `xdg-shell.xml` to create sources (as would the `wayland-scanner` do) but that works with `libwayland.so` being loaded dynamically.
The script is generic enough to work with other wayland protocols if we ever need it, and the resulting "dynamic table" is stored and filled at runtime in the `WaylandLoader` class.

I think this solution is the best that can be made, because doing so has no cost on performances - everything is loaded once, at runtime, only if wayland is used -, do not break any portability - nothing is linked statically so the binary just checks that everything is available when asked at runtime and fails gracefully in all cases and is as future proof as the Vulkan generators already integrated in GFXReconstruct.

If you have any question or remark, do not hesitate !

lunarpapillo:
The LunarG CI failure # 3754 seemed to be due to an old version of `git` on the system that didn't correctly resolve SSL to get the submodule.  I've fixed the issue and restarted.

charles-lunarg (changes requested):
Everything in the code parts of the PR look good. The only change I request (but am happy to do for you if you'd rather) is moving the wayland-protocols submodule into a 'download on demand' dependency, ie the generate_wayland.py script would download the protocols repo when it is being called.

Currently, the strategy for dependencies is to vendor anything that isn't changing often and submodule the things that do change often. Vulkan-Headers updates every week or so, with content that gfxrecon uses. Whereas most of the dependencies are either not updated nearly as frequently, or are in a stable state which doesn't make sense to update all the time.

Wayland-protocols from my understanding is very much like Vulkan-Headers, (more like the gitlab vulkan repo actually) and updates frequently. However, gfxrecon-replay isn't going to be adding new protocols every other week, so a submodule isn't a great fit.

Thus, I would think that download on demand would allow anyone who wants to add a protocol to do so easily while not increasing the checked in repo size or excess traffic to wayland-protocols.

bradgrantham-lunarg:
@charles-lunarg when this lands, could you also check out #1245 to see if this PR resolves that issue?

marius-pelegrin-arm:
There are multiple ways to do this so I just used the `subprocess` python module to launch git commands as already done in some other scripts in the repository. If the repository exists, it just launch `git pull` inside of it. I still use the same path (`external/wayland-protocols`) to store the repository, and added it to the gitignore.
If you had something else in mind / something to add/modify, you can do it if you want =)

charles-lunarg (approved):
Changes look great now.

I found an example capture that resized and can confirm that resizing works as expected.

PR #1192, opened 2023-07-11
Fix Vulkan deferred operation handling errors

mizhen:
**The problem**
when using deferred operation object in command vkcreateraytracingpipelines, there are two cases with the driver. One is driver proceed all workload in current thread without deferring this command, another case is deferring this command to other threads. The second case is not covered by existing handling. It cause some
title playback failure because reading wrong pipeline data and capture/playback shader group handle, it also cause trim playback failure because ray tracing pipeline creation failure.

**The solution**
The pull request is the fix for these errors. It added needed handling for deferred command and fixed related errors on pipeline reading, capture/playback shader group handle handling and trim deferred operation object handling.

**Result**
All tests passed for the target title and other applications include trim trace and full trace capture/playback, these test results show the pull-request fixed the target title playback issue caused by these deferred operation handling errors.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -6401,9 +6401,14 @@ VkResult VulkanReplayConsumerBase::OverrideCreateRayTracingPipelinesKHR(
 
     if (deferred_operation_info)
     {
-        deferred_operation_info->join_state = VK_NOT_READY;
+        deferred_operation_info->pending_state = true;
         deferred_operation_info->record_modified_create_infos.clear();
         deferred_operation_info->record_modified_pgroups.clear();
+        deferred_operation_info->replayPipelines.resize(createInfoCount);
+        deferred_operation_info->capturePipelines.resize(createInfoCount);
+        memcpy(deferred_operation_info->capturePipelines.data(),
+               pPipelines->GetPointer(),
+               createInfoCount * sizeof(VkPipeline));
     }
 
     if (device_info->property_feature_info.feature_rayTracingPipelineShaderGroupHandleCaptureReplay)
@@ -6468,7 +6473,33 @@ VkResult VulkanReplayConsumerBase::OverrideCreateRayTracingPipelinesKHR(
                                                             createInfoCount,
                                                             modified_create_infos.data(),
                                                             in_pAllocator,
-                                                            out_pPipelines);
+                                                            deferred_operation_info->replayPipelines.data());
+        if ((result == VK_SUCCESS) || (result == VK_OPERATION_NOT_DEFERRED_KHR) ||
+            (result == VK_PIPELINE_COMPILE_REQUIRED_EXT))
+        {
+            // The above return values mean the command is not deferred and driver will finish all workload in current
+            // thread. Therefore the created pipelines can be read and copied to out_pPipelines which will be
+            // referenced later.
+            //
+            // Note:
+            //     Some pipelines might actually fail creation if the return value is VK_PIPELINE_COMPILE_REQUIRED_EXT.
+            //     These failed pipelines will generate VK_NULL_HANDLE.
+            //
+            //     If the return value is VK_OPERATION_DEFERRED_KHR, it means the command is deferred, and thus pipeline
+            //     creation is not finished. Subsequent handling will be done by
+            //     vkDeferredOperationJoinKHR/vkGetDeferredOperationResultKHR after pipeline creation is finished.
+            memcpy(
+                out_pPipelines, deferred_operation_info->replayPipelines.data(), createInfoCount * sizeof(VkPipeline));
+
+            if (deferred_operation_info)
+            {
+                // Eventhough vkCreateRayTracingPipelinesKHR was called with a valid deferred operation object, the
+                // driver may opt to not defer the command. In this case, set pending_state flag to false to skip
+                // vkDeferredOperationJoinKHR handling.
+                deferred_operation_info->pending_state = false;
+            }
+        }
+
         if (deferred_operation_info)
         {
             deferred_operation_info->record_modified_create_infos = std::move(modified_create_infos);
@@ -6498,8 +6529,9 @@ VkResult VulkanReplayConsumerBase::OverrideDeferredOperationJoinKHR(PFN_vkDeferr
                                                                     const DeviceInfo*              device_info,
                                                                     DeferredOperationKHRInfo* deferred_operation_info)
 {
-    if (deferred_operation_info->join_state == VK_SUCCESS)
+    if (deferred_operation_info->pending_state == false)
     {
+        // The deferred operation object has no deferred command or its deferred command has been finished.
         return VK_SUCCESS;
     }
 
@@ -6537,9 +6569,19 @@ VkResult VulkanReplayConsumerBase::OverrideDeferredOperationJoinKHR(PFN_vkDeferr
         j.get();
     }
 
-    deferred_operation_info->join_state = VK_SUCCESS;
+    AddHandles<PipelineInfo>(device_info->capture_id,
+                             deferred_operation_info->capturePipelines.data(),
+                             deferred_operation_info->capturePipelines.size(),
+                             deferred_operation_info->replayPipelines.data(),
+                             deferred_operation_info->replayPipelines.size(),
+                             &VulkanObjectInfoTable::AddPipelineInfo);
+
+    deferred_operation_info->pending_state = false;
     deferred_operation_info->record_modified_create_infos.clear();
     deferred_operation_info->record_modified_pgroups.clear();
+    deferred_operation_info->capturePipelines.clear();
+    deferred_operation_info->replayPipelines.clear();
+
     return VK_SUCCESS;
 }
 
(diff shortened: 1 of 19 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -6155,7 +6161,33 @@ VkResult VulkanReplayConsumerBase::OverrideCreateRayTracingPipelinesKHR(
                                                             createInfoCount,
                                                             modified_create_infos.data(),
                                                             in_pAllocator,
-                                                            out_pPipelines);
+                                                            deferred_operation_info->replayPipelines.data());
```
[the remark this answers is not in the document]
mizhen: Thanks for pointing that out. I'll follow up with a commit.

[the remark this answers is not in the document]
mizhen: Done.

framework/decode/vulkan_export_json_consumer_base.cpp
```diff
@@ -621,6 +622,30 @@ void VulkanExportJsonConsumerBase::Process_vkUpdateDescriptorSetWithTemplateKHR(
...
+    auto& args = jdata[NameArgs()];
+    HandleToJson(args["device"], device, json_options_);
+    HandleToJson(args["deferredOperation"], deferredOperation, json_options_);
+    HandleToJson(args["pipelineCache"], pipelineCache, json_options_);
+    FieldToJson(args["createInfoCount"], createInfoCount, json_options_);
+    FieldToJson(args["pCreateInfos"], pCreateInfos, json_options_);
+    FieldToJson(args["pAllocator"], pAllocator, json_options_);
+    HandleToJson(args["pPipelines"], pPipelines, json_options_);
+    WriteBlockEnd();
+}
```
andrew-lunarg, line 647: Please remove this, it is the same as the default generated version. We don't want custom code where generated code can work. You probably added this because the default implementation disappeared when you added `vkCreateRayTracingPipelinesKHR` to `vulkan_generators/blacklists.json`. Each generator can have its own blocklisk.blacklist JSON file: they don't have to share this one.

Either give the JSON consumer its own blocklist or remove this function from the default blocklist in the JSON consumer header and body generators with a line like `self.APICALL_BLACKLIST.remove('vkCreateRayTracingPipelinesKHR')` (untested), probably at the botton of `__init__()` or the top of `generate_feature()`.

framework/decode/vulkan_export_json_consumer_base.h
```diff
@@ -179,6 +180,17 @@ class VulkanExportJsonConsumerBase : public VulkanConsumer, public AnnotationHan
...
+        const ApiCallInfo&                                               call_info,
+        VkResult                                                         returnValue,
+        format::HandleId                                                 device,
+        format::HandleId                                                 deferredOperation,
+        format::HandleId                                                 pipelineCache,
+        uint32_t                                                         createInfoCount,
+        StructPointerDecoder<Decoded_VkRayTracingPipelineCreateInfoKHR>* pCreateInfos,
+        StructPointerDecoder<Decoded_VkAllocationCallbacks>*             pAllocator,
+        HandlePointerDecoder<VkPipeline>*                                pPipelines) override;
+
```
andrew-lunarg, line 193: Please remove this copy of the generated version. See other comment on body of same module.

andrew-lunarg (changes requested):

framework/encode/custom_vulkan_api_call_encoders.h
```diff
@@ -1,6 +1,7 @@
 /*
 ** Copyright (c) 2018-2020 Valve Corporation
 ** Copyright (c) 2018-2020 LunarG, Inc.
+** Copyright (c) 2023 Advanced Micro Devices, Inc. All rights reserved.
```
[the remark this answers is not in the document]
mizhen: Will remove the copyright. My change to this file was already covered in dev branch.

framework/decode/vulkan_export_json_consumer_base.cpp
mizhen: I've updated it accordingly, thanks.

framework/decode/vulkan_export_json_consumer_base.h
mizhen: I've updated it accordingly, thanks.

bradgrantham-lunarg:
> CI gfxreconstruct build # 3436 failed.

This failure is just a replay failure for an unknown struct in a `vulkaninfo` capture in our CI tests.  The failure occurred because the PR branch needs to be rebased on `dev` to bring in support for the upgraded Vulkan structs that were added in recent SDKs.

mizhen:
> > CI gfxreconstruct build # 3436 failed.
> 
> This failure is just a replay failure for an unknown struct in a `vulkaninfo` capture in our CI tests. The failure occurred because the PR branch needs to be rebased on `dev` to bring in support for the upgraded Vulkan structs that were added in recent SDKs.

I'll rebase, thank you.

mizhen:
> > > CI gfxreconstruct build # 3436 failed.
> > 
> > 
> > This failure is just a replay failure for an unknown struct in a `vulkaninfo` capture in our CI tests. The failure occurred because the PR branch needs to be rebased on `dev` to bring in support for the upgraded Vulkan structs that were added in recent SDKs.
> 
> I'll rebase, thank you.

Done. thanks.

lunarpapillo:
The last two failed runs #3442 and #3443 seem to be due to a crashed GPU.  I will be restarting your run when the machine is ready.

mizhen:
> The last two failed runs #3442 and #3443 seem to be due to a crashed GPU. I will be restarting your run when the machine is ready.

Thanks.

bradgrantham-lunarg:
While trying to test this, I made a simple case based on Sascha Willem's `raytracingbasic` example that creates a VkDeferredOperationKHR, uses it with vkCreateRayTracingPipelinesKHR, joins the operation until complete, and then destroys the operation.  Please see https://github.com/bradgrantham-lunarg/Vulkan/tree/brad-sascha-raytracingbasic-deferred-crtp.

Is this use of deferredoperation valid, in your opinion?  If not, would you let me know why?

With this branch, GFXR capture at c46523ac (dev at this moment) can capture and replay `raytracingbasic` without a problem, but your PR branch outputs this warning on capture:
```
[gfxrecon] WARNING - GetWrappedId() couldn't find Handle: 14829735431805717965's wrapper. It might have been destroyed
```
And then `gfxrecon-replay` from your PR branch fails with this assertion failure:
```
Assertion failed: (device_info != nullptr) && (pipeline_info != nullptr) && (pData != nullptr) && (pData->GetOutputPointer() != nullptr), file F:\gfxreconstruct\framework\decode\vulkan_replay_consumer_base.cpp, line 6598
```

Can you please address what might be happening there?  Thank you!

mizhen:
> While trying to test this, I made a simple case based on Sascha Willem's `raytracingbasic` example that creates a VkDeferredOperationKHR, uses it with vkCreateRayTracingPipelinesKHR, joins the operation until complete, and then destroys the operation. Please see https://github.com/bradgrantham-lunarg/Vulkan/tree/brad-sascha-raytracingbasic-deferred-crtp.
> 
> Is this use of deferredoperation valid, in your opinion? If not, would you let me know why?
> 
> With this branch, GFXR capture at [c46523a](https://github.com/LunarG/gfxreconstruct/commit/c46523ac3079e4dac5deaa7f21313c1917137f0c) (dev at this moment) can capture and replay `raytracingbasic` without a problem, but your PR branch outputs this warning on capture:
> 
> ```
> [gfxrecon] WARNING - GetWrappedId() couldn't find Handle: 14829735431805717965's wrapper. It might have been destroyed
> ```
> 
> And then `gfxrecon-replay` from your PR branch fails with this assertion failure:
> 
> ```
> Assertion failed: (device_info != nullptr) && (pipeline_info != nullptr) && (pData != nullptr) && (pData->GetOutputPointer() != nullptr), file F:\gfxreconstruct\framework\decode\vulkan_replay_consumer_base.cpp, line 6598
> ```
> 
> Can you please address what might be happening there? Thank you!
> 
> EDIT: I tested on an "NVIDIA GeForce RTX 2070 SUPER"

I'm looking into it, will keep you updated, thanks.

framework/encode/custom_vulkan_api_call_encoders.cpp
```diff
@@ -652,24 +653,40 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateRayTracingPipelinesKHR(VkDevice
...
-            ->EndGroupCreateApiCallCapture<VkDevice,
-                                           VkDeferredOperationKHR,
-                                           PipelineWrapper,
-                                           VkRayTracingPipelineCreateInfoKHR>(
-                result, device, deferredOperation, createInfoCount, pPipelines, pCreateInfos);
+    auto device_wrapper = GetWrapper<DeviceWrapper>(device);
+    if ((result != VK_OPERATION_DEFERRED_KHR) ||
+        (!device_wrapper->property_feature_info.feature_rayTracingPipelineShaderGroupHandleCaptureReplay))
+    {
+        // If the operation is not deferred by driver. or the system doesn't support
```
bradgrantham-lunarg, line 660: As discussed, this block ends up being written twice by the code in this branch.  I don't think that's what we want for this solution.  If there has to be a deferred operation read of the output variable, we probably need a MetaData command to write that value, and then replay should handle the case of postponing creation of the VkPipeline handles until that MetaData command is written.

I'm confused about what's happening here.  Do you have a test case where the application creates an RT pipeline and gets VK_OPERATION_DEFERRED_KHR then joins the deferred operation until success, then uses the pipeline handles but GFXR somehow writes the operations out of order so the pipeline handles are used in replay before their creation is complete?

bradgrantham-lunarg (changes requested):

framework/encode/custom_vulkan_api_call_encoders.cpp
bradgrantham-lunarg: > Do you have a test case where the application creates an RT pipeline and gets VK_OPERATION_DEFERRED_KHR then joins the deferred operation until success, then uses the pipeline handles but GFXR somehow writes the operations out of order so the pipeline handles are used in replay before their creation is complete?

If that's what's happening, then I think I need to understand how that could happen and we may need to introduce synchronization around the pipeline handles and anything using those pipeline handles, and not create this block twice.

bradgrantham-lunarg: Can you describe in steps how the problem occurs so I can understand how this branch is expected to  the issue?

mizhen: The modified app is valid.

Yes, I think the hardware doesn't support the feature rayTracingPipelineShaderGroupHandleCaptureReplay, it triggers the issue that the above code block generates one more trace block. I'll add a commit for the issue. It may playback the trace of raytracingbasic as current dev branch. However, for this hardware, it's not guarantee playback always work, the reason is the following:

GFXR current framework requires the feature for capture/replay raytracing titles. It needs the related GPU address/ShaderGroupHandle capture/replay extension to reproduce these capture time GPU address/handle value during replay time. Without the feature (rayTracingPipelineShaderGroupHandleCaptureReplay) support, it's not guarantee for replaying raytracing trace file because shadergroup handle value may have different value during replay time, especially for complicated apps or real game, simple app and full trace file replay sometimes may work because the used GPU addresses/handles way less than real title.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -6332,9 +6333,14 @@ VkResult VulkanReplayConsumerBase::OverrideCreateRayTracingPipelinesKHR(
 
     if (deferred_operation_info)
     {
-        deferred_operation_info->join_state = VK_NOT_READY;
+        deferred_operation_info->pending_state = true;
         deferred_operation_info->record_modified_create_infos.clear();
         deferred_operation_info->record_modified_pgroups.clear();
+        deferred_operation_info->replayPipelines.resize(createInfoCount);
+        deferred_operation_info->capturePipelines.resize(createInfoCount);
+        memcpy(deferred_operation_info->capturePipelines.data(),
```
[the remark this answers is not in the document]
mizhen: ok, will do.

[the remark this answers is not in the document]
mizhen: Done.

framework/encode/custom_vulkan_api_call_encoders.cpp
mizhen: Done.

framework/encode/custom_vulkan_api_call_encoders.cpp
```diff
@@ -652,24 +653,39 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateRayTracingPipelinesKHR(VkDevice
         omit_output_data = true;
     }
 
-    auto encoder = VulkanCaptureManager::Get()->BeginTrackedApiCallCapture(
-        format::ApiCallId::ApiCall_vkCreateRayTracingPipelinesKHR);
-    if (encoder)
+    auto device_wrapper = GetWrapper<DeviceWrapper>(device);
+    if (result != VK_OPERATION_DEFERRED_KHR)
```
locke-lunarg, line 657: Thank you, @mizhen , for fixing it. I didn't realize the `TrackRayTracingShaderGroupHandles` issue for the deferred case. This branch looks good to me. Could you please add a warning message for delaying writing the block of `vkCreateRayTracingPipelinesKHR` to the capture file? I think the user needs to know it.

locke-lunarg:
Thank you, @mizhen , for fixing it. I didn't realize the `TrackRayTracingShaderGroupHandles` issue for the deferred case. This branch looks good to me. Could you please add a warning message for delaying writing the block of `vkCreateRayTracingPipelinesKHR` to the capture file? I think the user needs to know it.

mizhen:
> `vkCreateRayTracingPipelinesKHR`

Done.

locke-lunarg:
@mizhen Thank you. Could you please `rebase dev` to fix conflicts? And then, it will be ready to merge.

mizhen:
> > @mizhen Thank you. Could you please `rebase dev` to fix conflicts? And then, it will be ready to merge.
> > OK, will do.

Done.

locke-lunarg:
CI gfxreconstruct build # 3629 passed.  It's good to merge.

PR #1193, opened 2023-07-12
Update headers to 1.3.257

locke-lunarg:


framework/generated/generated_vulkan_struct_to_json.cpp
```diff
@@ -305,9 +305,7 @@ void FieldToJson(nlohmann::ordered_json& jdata, const Decoded_StdVideoEncodeH264
         const Decoded_StdVideoEncodeH264SliceHeaderFlags& meta_struct = *data;
         FieldToJson(jdata["direct_spatial_mv_pred_flag"], decoded_value.direct_spatial_mv_pred_flag, options);
         FieldToJson(jdata["num_ref_idx_active_override_flag"], decoded_value.num_ref_idx_active_override_flag, options);
-        FieldToJson(jdata["no_output_of_prior_pics_flag"], decoded_value.no_output_of_prior_pics_flag, options);
-        FieldToJson(jdata["adaptive_ref_pic_marking_mode_flag"], decoded_value.adaptive_ref_pic_marking_mode_flag, options);
-        FieldToJson(jdata["no_prior_references_available_flag"], decoded_value.no_prior_references_available_flag, options);
+        FieldToJson(jdata["reserved"], decoded_value.reserved, options);
```
andrew-lunarg, line 308: The struct lost flags?

framework/generated/generated_vulkan_struct_to_json.cpp
```diff
@@ -431,15 +438,11 @@ void FieldToJson(nlohmann::ordered_json& jdata, const Decoded_StdVideoEncodeH264
...
-        FieldToJson(jdata["num_ref_idx_l1_active_minus1"], decoded_value.num_ref_idx_l1_active_minus1, options);
-        FieldToJson(jdata["cabac_init_idc"], decoded_value.cabac_init_idc, options);
-        FieldToJson(jdata["disable_deblocking_filter_idc"], decoded_value.disable_deblocking_filter_idc, options);
         FieldToJson(jdata["slice_alpha_c0_offset_div2"], decoded_value.slice_alpha_c0_offset_div2, options);
         FieldToJson(jdata["slice_beta_offset_div2"], decoded_value.slice_beta_offset_div2, options);
         FieldToJson(jdata["reserved1"], decoded_value.reserved1, options);
-        FieldToJson(jdata["reserved2"], decoded_value.reserved2, options);
+        FieldToJson(jdata["cabac_init_idc"], decoded_value.cabac_init_idc, options);
+        FieldToJson(jdata["disable_deblocking_filter_idc"], decoded_value.disable_deblocking_filter_idc, options);
         FieldToJson(jdata["pWeightTable"], meta_struct.pWeightTable, options);
```
andrew-lunarg, line 446: Fields are going away after they have been published?

andrew-lunarg (approved):
Looks like it was a pretty smooth update.

framework/generated/generated_vulkan_struct_to_json.cpp
bradgrantham-lunarg: That happens sometimes.  That would break existing programs as well as break existing traces that were already captured.  Perhaps in the future we could record the extension's version number with any structs related to the extension, and that would at least allow us to throw a warning.  This usually happens only very early in an extension's lifecycle, so I'm not inclined to worry about it much, but if it became a problem we could introduce versioned struct encoders and decoders.

framework/generated/generated_vulkan_struct_to_json.cpp
bradgrantham-lunarg: Ditto the above comment.

PR #1194, opened 2023-07-13
Terminate process once all trim ranges have been captured

panos-lunarg:
Once all frame ranges specified by
GFXRECON_CAPTURE_FRAMES/debug.gfxrecon.capture_frames) have been
recorded, GFXR will call exit() forcing the whole process to terminate.

New control options (GFXRECON_QUIT_AFTER_CAPTURE_FRAMES for desktop and
debug_gfxrecon.quit_after_capture_frames for Android) can be used to
enable or disable this.


andrew-lunarg:
I don't care myself but just mentioning this would cause an otherwise validation-passing replay to not pass validation since all Vulkan handles must be destroyed. I guess hard exiting this way might stop validation errors on exit being reported anyway even if the `--validate` flag was passed to the replay that was being recaptured for trim.

andrew-lunarg (approved):
LGTM

bradgrantham-lunarg:
It looks like this would also stop D3D12 captures after the last trim range.  @davidd-lunarg do you agree?  @panos-lunarg could you add your Vulkan usage note to the D3D12 usage `.md` as well?

framework/encode/capture_manager.cpp
```diff
@@ -610,6 +611,12 @@ void CaptureManager::CheckContinueCaptureForWriteMode()
                 capture_mode_ = kModeDisabled;
                 DestroyStateTracker();
                 compressor_ = nullptr;
+
+                if (quit_after_frame_ranges_)
+                {
+                    GFXRECON_LOG_INFO("All trim ranges have been captured. Quitting.");
+                    exit(0);
```
bradgrantham-lunarg, line 618: This skips `DestroyInstance` which calls `~CaptureManager()` which calls `util::PageGuardManager::Destroy();`.  I think exiting is relatively safe or we'd have real problems when applications crash and cause GFXR to terminate unexpectedly.  Still, what do you think about a more formal shutdown path here which closes out CaptureManager, calling the virtual dtor for the derived capture managers?

bradgrantham-lunarg: @panos-lunarg

panos-lunarg: `exit(0)` still exists but now I am installing an `atexit` function which should be called right after `exit()` and call the destructor

bradgrantham-lunarg (approved):

PR #1204, opened 2023-07-19
Enable trimming to queue submission range 

davidd-lunarg:
Enable trimming based on a range of queue submissions--`vkQueueSubmit` for
Vulkan, `ExecuteCommandLists` for DX12. Trimming is activated immediately
before the first submission and deactivated immediately after the last
submission in the range(s) specified.

Also refactor uint range parsing to be shared between various command line options.


framework/util/options.cpp
```diff
@@ -71,77 +73,81 @@ std::vector<FrameRange> GetFrameRanges(const std::string& args)
...
+                                     option_name);
                 invalid = true;
                 break;
             }
         }
 
         if (!invalid)
         {
-            FrameRange screenshot_range;
+            UintRange uint_range;
```
bradgrantham-lunarg, line 86: Could this be `boundary_range` or `trim_range` or something similar that is more descriptive?

bradgrantham-lunarg (approved):
Looks good with just the one note about naming `uint_range` variable

framework/util/options.cpp
davidd-lunarg: The intent is for this to be a generic function to parse uint ranges and the current name reflects that. Besides trim frame and queue submit ranges it is also used for screenshot ranges and could be used elsewhere range options are parsed (e.g., [`GetMeasurementFrameRange`](https://github.com/LunarG/gfxreconstruct/blob/016ac347a71c62e848715d47fa0a71629784a1ff/tools/tool_settings.h#L501))

bradgrantham-lunarg: Okay, yeah, that makes sense!

Issue #1199, opened 2023-07-14
Many Empty Generated UnwrapStructHandles() in .cpp File


andrew-lunarg:
There are UnwrapStructHandles() functions in `framework\generated\generated_vulkan_struct_handle_wrappers.cpp` which are empty except a useless branch, meaning we make a call to them which immediately returns. If the body is going to be empty, make it inline and generate it in the header so calls into it can be optimized away. Try to generate something like this in the header:  

```
inline void UnwrapStructHandles(VkBufferMemoryBarrier* value, HandleUnwrapMemory* unwrap_memory) {}
```

Examples of present situation:
```cpp
void UnwrapStructHandles(VkBufferMemoryBarrier* value, HandleUnwrapMemory* unwrap_memory)
{
    if (value != nullptr)
    {
    }
}

void UnwrapStructHandles(VkImageMemoryBarrier* value, HandleUnwrapMemory* unwrap_memory)
{
    if (value != nullptr)
    {
    }
}

void UnwrapStructHandles(VkMappedMemoryRange* value, HandleUnwrapMemory* unwrap_memory)
{
    if (value != nullptr)
    {
    }
}

void UnwrapStructHandles(VkSparseMemoryBind* value, HandleUnwrapMemory* unwrap_memory)
{
    if (value != nullptr)
    {
    }
}
...
```

David-J-Tinley:
Hi, I was wondering if I could be assigned this issue? It would be my first contribution, thank you!

David-J-Tinley:
@andrew-lunarg thank you! I will get started right away!

David-J-Tinley:
Hi @andrew-lunarg, I had a question. Can the functions that are being converted to inline functions in the header have a completely empty body? Ex: `inline void UnwrapStructHandles(VkPipelineLibraryCreateInfoKHR* value, HandleUnwrapMemory* unwrap_memory) {};` Or should I be putting some kind of function in the bodies? Thank you for your time!

andrew-lunarg:
@David-J-Tinley We would like to compile clean even if the unused parameters warning is enabled. I gave an example in the description which uses our `GFXRECON_UNREFERENCED_PARAMETER` macro to avoid compiler warnings. there are other mechanisms to avoid the warning that you could also try. It looks like the simplest is to simply avoid generating the names of parameters and just give their types as shown in this answer: https://stackoverflow.com/a/18786498. I have not tried that myself though IIRC.

The way I suggested:
```cpp
[...]

PR #1206, opened 2023-07-19
Updated empty generated UnwrapStructHandles() in .cpp file to inline functions in header

David-J-Tinley:
Removed the empty generated functions in `framework\generated\generated_vulkan_struct_handle_wrappers.cpp` and made them into inline functions in the header file at `framework\generated\generated_vulkan_struct_handle_wrappers.hpp ` This was for issue #1199


bradgrantham-lunarg:
Hi, @David-J-Tinley !  Thanks for your contribution!  Do you have changes to the Python generator scripts that you haven't committed?

David-J-Tinley:
Hi @bradgrantham-lunarg , sorry I completely missed the part in the issue #1199 messages where @andrew-lunarg said to rebuild the generated code with `python3 generate_vulkan.py`. Should I delete this pull request and resubmit another one after the rebuild? Thank you for your time!

andrew-lunarg:
@David-J-Tinley, if you want to keep at this, the task is to edit the python scripts for the C++ header and body that they automatically generate code with the same changes applied that you have made manually. 

You would also want to refresh your fork and then either rebase your branch on dev or just start a new branch since you would be editing different files anyway and your branch currently contains quite a lot of unrelated history.

Thinking some more about the problem it may not have been quite such a good first issue as I had imagined since the header generator would need to duplicate some of the work of the body generator to know whether it should emit a function prototype or an inline empty function in each case and there's a slightly higher level design decision to be made there about duplicating code versus factoring it out.

David-J-Tinley:
@andrew-lunarg oh ok, I think I misunderstood the overall issue but it is starting to make a little more sense. It sounds like it may be a little bit beyond what I could do unfortunately. Would you like me to remove my pull request? Thank you for taking the time to explain things to me though, I appreciate it!

andrew-lunarg:
@David-J-Tinley Sure no problem. Thanks for giving it a shot :).

PR #1207, opened 2023-07-20
discard-cached-psos renamed to use_cached_psos

igorbrsn:
**Problem:**
The replayer should always use 'discard-cached-psos'. Currently '--dcp' and '--discard-cached-psos' are options and could not be disabled. 

**Solution:**
Added option'--use-cached-psos' with default 'false'. 'discard_cached_psos' renamed to 'use_cached_psos' and would be set 'true' if used.

**Test:**
Tested with default, enabled and disabled. Warning issued when deprecated '--dcp' is used.

### DIFF
--- a/USAGE_desktop_D3D12.md
+++ b/USAGE_desktop_D3D12.md
@@ -183,7 +183,7 @@ Usage:
                         [--screenshot-dir <dir>] [--screenshot-prefix <file-prefix>]
                         [--sfa | --skip-failed-allocations] [--replace-shaders <dir>]
                         [--opcd | --omit-pipeline-cache-data] [--wsi <platform>]
-                        [--use-cached-psos <boolean>] [--surface-index <N>]
+                        [--use-cached-psos] [--surface-index <N>]
                         [--remove-unsupported] [--validate]
                         [--onhb | --omit-null-hardware-buffers]
                         [-m <mode> | --memory-translation <mode>]
@@ -285,9 +285,8 @@ Vulkan-only:
                                         and suballocations.
 
 D3D12-only:
-  --use-cached-psos <boolean>  Permit using cached PSOs when creating graphics or compute pipelines.
+  --use-cached-psos            Permit using cached PSOs when creating graphics or compute pipelines.
                                Disabling helps enable replay across changing driver installs.
-                               Default: disabled
   --debug-device-lost          Enables automatic injection of breadcrumbs into command buffers
                                and page fault reporting.
                                Used to debug Direct3D 12 device removed problems.
--- a/USAGE_desktop_D3D12.md
+++ b/USAGE_desktop_D3D12.md
@@ -285,9 +285,9 @@ Vulkan-only:
                                         and suballocations.
 
 D3D12-only:
-  --use-cached-psos <boolean>  Force CachedPSO to null when creating graphics or compute PSOs.
-                               Can help enable replay across changing driver installs.
-                               Default: enabled
+  --use-cached-psos <boolean>  Permit using cached PSOs when creating graphics or compute pipelines.
+                               Disabling helps enable replay across changing driver installs.
+                               Default: disabled
   --debug-device-lost          Enables automatic injection of breadcrumbs into command buffers
                                and page fault reporting.
                                Used to debug Direct3D 12 device removed problems.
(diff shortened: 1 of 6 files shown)


bradgrantham-lunarg:
In general I think this is fine, probably "dcp" should be the default.  However, "dcp" means "discard cached PSOs".  "enable-dcp" would mean "enable discard cached PSOs", the same thing as "dcp".   If this change makes "discard cached PSOs" the default and adds an option to *not* discard the PSOs, then I think the option would be called "use-cached-psos" or some such.

igorbrsn:
> > In general I think this is fine, probably "dcp" should be the default. However, "dcp" means "discard cached PSOs". "enable-dcp" would mean "enable discard cached PSOs", the same thing as "dcp". If this change makes "discard cached PSOs" the default and adds an option to _not_ discard the PSOs, then I think the option would be called "use-cached-psos" or some such.
> 
> Changed to --use-cached-psos

I am updating logic

tools/tool_settings.h
```diff
@@ -805,7 +806,14 @@ static gfxrecon::decode::DxReplayOptions GetDxReplayOptions(const gfxrecon::util
 
     if (arg_parser.IsOptionSet(kDiscardCachedPsosLongOption) || arg_parser.IsOptionSet(kDiscardCachedPsosShortOption))
     {
-        replay_options.discard_cached_psos = true;
+        GFXRECON_LOG_WARNING("The parameters --dcp and --discard-cached-psos have been depracated in favor for "
```
bradgrantham-lunarg, line 809: "deprecated"

USAGE_desktop_D3D12.md
```diff
@@ -183,7 +183,7 @@ Usage:
                         [--screenshot-dir <dir>] [--screenshot-prefix <file-prefix>]
                         [--sfa | --skip-failed-allocations] [--replace-shaders <dir>]
                         [--opcd | --omit-pipeline-cache-data] [--wsi <platform>]
-                        [--dcp | --discard-cached-psos] [--surface-index <N>]
+                        [--use-cached-psos <boolean>] [--surface-index <N>]
```
bradgrantham-lunarg, line 186: I think it shouldn't be a boolean if it's off if you specify nothing but turned on when you specify it.  Making this a boolean would make it inconsistent with other command-line switches.

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -2207,7 +2207,7 @@ HRESULT Dx12ReplayConsumerBase::OverrideLoadGraphicsPipeline(
                (desc != nullptr) && (desc->GetPointer() != nullptr) && (state != nullptr));
 
         auto desc2 = desc->GetPointer();
-        if (options_.discard_cached_psos)
+        if (options_.use_cached_psos == false)
```
bradgrantham-lunarg, line 2210: I prefer `!options_.use_cached_psos` to checking for explicit false.

bradgrantham-lunarg (changes requested):

USAGE_desktop_D3D12.md
```diff
@@ -285,16 +285,15 @@ Vulkan-only:
...
-                        (Same as --discard-cached-psos)
-  --debug-device-lost   Enables automatic injection of breadcrumbs into command buffers
-                        and page fault reporting.
-                        Used to debug Direct3D 12 device removed problems.
-  --fw <width,height>   Setup windowed and override resolution.
-                        (Same as --force-windowed)
-  --create-dummy-allocations Enables creation of dummy heaps and resources
-                             for replay validation.
+  --use-cached-psos            Permit using cached PSOs when creating graphics or compute pipelines.
+                               Disabling helps enable replay across changing driver installs.
```
davidd-lunarg, line 289: Can this text be changed to give some more detail about why a user might or might not want to enable the flag, for example:
```
  --use-cached-psos   Permit using cached PSOs when creating graphics or compute pipelines.
                      Using cached PSOs may reduce PSO creation time but may result in replay errors.
```

`PrintUsage()` in `replay_settings.h` also needs to be updated to match whatever text is used here.

davidd-lunarg (changes requested):
I've added a couple requests to fixup documentation for the new option. Also, please update the PR title and description to match the latest changes in the PR.

davidd-lunarg (approved):

PR #1208, opened 2023-07-20
Add --gpu option to the optimizer

feimao-gfxr:
**Problem**
A system may have multiple GPUs. We should be able to select a GPU for the optimization, because if the trace is captured with one of the GPUs, the optimization may be performed only with the same GPU. 

**Solution**
Add the option to the optimizer. 

**Result**
Tested. It is working.

### DIFF
--- a/tools/optimize/dx12_optimize_util.cpp
+++ b/tools/optimize/dx12_optimize_util.cpp
@@ -71,9 +71,10 @@ void CreateResourceValueTrackingConsumer(
     application = std::make_shared<gfxrecon::application::Application>(app_string, file_processor);
     application->InitializeDx12WsiContext();
 
-    // Use default replay options, except dcp.
+    // Use default replay options, except dcp and gpu index.
     decode::DxReplayOptions dx_replay_options;
     dx_replay_options.discard_cached_psos = true;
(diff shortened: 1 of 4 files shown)


davidd-lunarg (changes requested):

tools/optimize/dx12_optimize_util.cpp
```diff
@@ -71,9 +71,10 @@ void CreateResourceValueTrackingConsumer(
     application = std::make_shared<gfxrecon::application::Application>(app_string, file_processor);
     application->InitializeDx12WsiContext();
 
-    // Use default replay options, except dcp.
+    // Use default replay options, except dcp and gpu index.
     decode::DxReplayOptions dx_replay_options;
     dx_replay_options.discard_cached_psos = true;
+    dx_replay_options.override_gpu_index  = options.override_gpu_index;
```
davidd-lunarg, line 77: There are merge conflicts so this PR needs to be rebased onto the latest `dev` and the conflicts resolved. PR #1207 replaced the var `DxReplayOptions::discard_cached_psos` with `DxReplayOptions::use_cached_psos` so that may be the source of the conflict.

PR #1210, opened 2023-07-21
Use Dx12GpuVaMap to check valid GPUVAs in RV annotation

davidd-lunarg:


rurra-amd (approved):
Looks good to me but should also verify with @JerryAMD

davidd-lunarg:
Addressed by PR #1214

PR #1211, opened 2023-07-21
Add ags to convert tool

feimao-gfxr:
**Problem**
1. AGS support is already in GFXR. The convert output file should also include the AGS calls.
2. The index values in the convert output file doesn't match anything in the replay consumer. It makes it very hard to find the matching call in the convert file when debugging an issue in the replayer.

**Solution**
1. Add AGS ascii consumer to support AGS in the convert tool. 
2. Remove api_call_count_ from the convert consumer and use block index for index values in the convert output file. Block index values provides consistent matches between different decode consumers. 

**Result**
AGS calls are included in the convert output file. Indices are block index values now.

### DIFF
--- a/framework/decode/custom_ags_ascii_consumer.cpp
+++ b/framework/decode/custom_ags_ascii_consumer.cpp
@@ -59,51 +59,6 @@ static std::string GetAgsResultValueString(AGSReturnCode result)
     }
 }
 
-std::string InterfaceIdToString(const IID& iid)
-{
-    if (iid == IID_IDXGIDevice)
-        return "\"IID_IDXGIDevice\"";
-    if (iid == IID_IDXGIDevice1)
-        return "\"IID_IDXGIDevice1\"";
-    if (iid == IID_IDXGIDevice2)
-        return "\"IID_IDXGIDevice2\"";
-    if (iid == IID_IDXGIDevice3)
-        return "\"IID_IDXGIDevice3\"";
-    if (iid == IID_IDXGIDevice4)
-        return "\"IID_IDXGIDevice4\"";
-    if (iid == IID_ID3D12Device)
-        return "\"IID_ID3D12Device\"";
-    if (iid == IID_ID3D12Device1)
-        return "\"IID_ID3D12Device1\"";
-    if (iid == IID_ID3D12Device2)
-        return "\"IID_ID3D12Device2\"";
-    if (iid == IID_ID3D12Device3)
-        return "\"IID_ID3D12Device3\"";
-    if (iid == IID_ID3D12Device4)
-        return "\"IID_ID3D12Device4\"";
-    if (iid == IID_ID3D12Device5)
-        return "\"IID_ID3D12Device5\"";
-    if (iid == IID_ID3D12Device6)
-        return "\"IID_ID3D12Device6\"";
-    if (iid == IID_ID3D12Device7)
-        return "\"IID_ID3D12Device7\"";
-    if (iid == IID_ID3D12Device8)
-        return "\"IID_ID3D12Device8\"";
-    if (iid == IID_ID3D12Device9)
-        return "\"IID_ID3D12Device9\"";
-    if (iid == IID_ID3D12Device10)
-        return "\"IID_ID3D12Device10\"";
-    if (iid == IID_ID3D12Device11)
-        return "\"IID_ID3D12Device11\"";
-    if (iid == IID_ID3D12DebugDevice1)
-        return "\"IID_ID3D12DebugDevice1\"";
-    if (iid == IID_ID3D12DebugDevice)
-        return "\"IID_ID3D12DebugDevice\"";
-    if (iid == IID_ID3D12DebugDevice2)
-        return "\"IID_ID3D12DebugDevice2\"";
-    return "\"Invalid IID\"";
-}
-
 std::string FeatureLevelToString(const D3D_FEATURE_LEVEL& value)
 {
     switch (value)
--- a/framework/decode/custom_ags_ascii_consumer.cpp
+++ b/framework/decode/custom_ags_ascii_consumer.cpp
@@ -59,38 +59,6 @@ static std::string GetAgsResultValueString(AGSReturnCode result)
     }
 }
 
-std::string FeatureLevelToString(const D3D_FEATURE_LEVEL& value)
-{
-    switch (value)
-    {
-        case D3D_FEATURE_LEVEL_1_0_CORE:
-            return "D3D_FEATURE_LEVEL_1_0_CORE";
-        case D3D_FEATURE_LEVEL_9_1:
-            return "D3D_FEATURE_LEVEL_9_1";
-        case D3D_FEATURE_LEVEL_9_2:
-            return "D3D_FEATURE_LEVEL_9_2";
-        case D3D_FEATURE_LEVEL_9_3:
-            return "D3D_FEATURE_LEVEL_9_3";
-        case D3D_FEATURE_LEVEL_10_0:
-            return "D3D_FEATURE_LEVEL_10_0";
-        case D3D_FEATURE_LEVEL_10_1:
-            return "D3D_FEATURE_LEVEL_10_1";
-        case D3D_FEATURE_LEVEL_11_0:
-            return "D3D_FEATURE_LEVEL_11_0";
-        case D3D_FEATURE_LEVEL_11_1:
-            return "D3D_FEATURE_LEVEL_11_1";
-        case D3D_FEATURE_LEVEL_12_0:
-            return "D3D_FEATURE_LEVEL_12_0";
-        case D3D_FEATURE_LEVEL_12_1:
-            return "D3D_FEATURE_LEVEL_12_1";
-        case D3D_FEATURE_LEVEL_12_2:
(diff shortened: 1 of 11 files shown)


andrew-lunarg:
It looks like the PR needs a simple `clang-format.exe -i tools/convert/main.cpp` to get through our style check.

feimao-gfxr:
> It looks like the PR needs a simple `clang-format.exe -i tools/convert/main.cpp` to get through our style check.

Updated.

lunarpapillo:
CI build #3027 failed because the **Pixel 3** needs a factory reset.  I've restarted your run.

tools/convert/main.cpp
```diff
@@ -270,6 +275,18 @@ int main(int argc, const char** argv)
...
                                                                      : gfxrecon::util::kToString_Unformatted;
             dx12_ascii_consumer.Initialize(out_file_handle, dx12_json_flags);
+
+#ifdef GFXRECON_AGS_SUPPORT
+            gfxrecon::decode::AgsAsciiConsumer ags_ascii_consumer;
+            ags_ascii_consumer.Initialize(out_file_handle, dx12_json_flags);
+
+            gfxrecon::decode::AgsDecoder ags_decoder;
+
+            ags_decoder.AddConsumer(reinterpret_cast<gfxrecon::decode::AgsConsumerBase*>(&ags_ascii_consumer));
```
andrew-lunarg, line 285: This cast should not be needed, and definitely not a `reinterpret_cast` unless I'm missing something here. Casting a class to its public base class makes no sense. See: `class AgsAsciiConsumer : public AgsConsumerBase`.

This should be fine:
` ags_decoder.AddConsumer(&ags_ascii_consumer);`

andrew-lunarg (changes requested):
This seems fine to me with a few tweaks noted inline as separate comments.

I think ideally we'd prefer:
*  JSON output to use the nlohmann library in `external\nlohmann\include\nlohmann\json.hpp`, 
* to use `Json` and `json` in names of types and file instead of `Ascii` and `ascii`,
* and for the JSON output to be structured more like the Vulkan schema.

That ideal outcome would give us simpler code and consistency in output,
but this should still slot in next to the other JSON Consumers and could still be useful.

I agree about the change to block indexes, and good catch deleting the dead code counting calls but never using the count.

feimao-gfxr:
> This seems fine to me with a few tweaks noted inline as separate comments.
> 
> I think ideally we'd prefer:
> 
> * JSON output to use the nlohmann library in `external\nlohmann\include\nlohmann\json.hpp`,
> * to use `Json` and `json` in names of types and files instead of `Ascii` and `ascii`,
> * and for the JSON output to be structured more like the Vulkan schema.
> 

* JSON output to use the nlohmann library in `external\nlohmann\include\nlohmann\json.hpp`,
* to use `Json` and `json` in names of types and files instead of `Ascii` and `ascii`,
* and for the JSON output to be structured more like the Vulkan schema.

These are general efforts. Is there any work going on in parallel for the gfxrecon-convert project at LunarG? Will the future work on gfxrecon-convert be based on this PR? We don't want duplicated efforts.

andrew-lunarg:
There are ongoing efforts to work on Convert. The Vulkan support in Convert all uses the nlohmann library. The DX12 support is being moved to it. There is no internal plan to move AGS to it.

For the purpose of this PR, I think the important part of my previous review comment was:

> This seems fine to me with a few tweaks noted inline as separate comments.

I.e. we can get this in to dev now with a couple of tweaks and worry about bringing it more in-line with the JSON output for the other APIs later if there is desire to do so.

If you would rather wait and see what the changes to DX12 look like before merging this, that is also an option.

framework/decode/dx12_consumer_base.h
```diff
@@ -1,5 +1,6 @@
 /*
 ** Copyright (c) 2021 LunarG, Inc.
```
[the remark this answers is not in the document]
feimao-gfxr: Done

framework/decode/dx12_consumer_base.h
```diff
@@ -188,7 +189,6 @@ class Dx12ConsumerBase
     bool ei_workload_{ false };
     bool opt_fillmem_{ false };
 
-  private:
```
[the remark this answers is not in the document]
feimao-gfxr: Make it private. Another change is done accordingly, in dx12_ascii_consumer_base.h.

feimao-gfxr:
> There are ongoing efforts to work on Convert. The Vulkan support in Convert all uses the nlohmann library. The DX12 support is being moved to it. There is no internal plan to move AGS to it.
> 
> For the purpose of this PR, I think the important part of my previous review comment was:
> 
> > This seems fine to me with a few tweaks noted inline as separate comments.
> 
> I.e. we can get this in to dev now with a couple of tweaks and worry about bringing it more in-line with the JSON output for the other APIs later if there is desire to do so.
> 
> If you would rather wait and see what the changes to DX12 look like before merging this, that is also an option.

For moving to using nlohmann library and related file renaming, I think it would be better to do it together with (or after) the changes to DX12.

andrew-lunarg (approved):
Looks good to me. I think the `ToString` functions in util mentioned by David could be used though.

framework/decode/file_processor.h
```diff
@@ -158,7 +159,6 @@ class FileProcessor
     std::vector<uint8_t>                parameter_buffer_;
     std::vector<uint8_t>                compressed_parameter_buffer_;
     util::Compressor*                   compressor_;
-    uint64_t                            api_call_index_;
```
andrew-lunarg, line 161: Incidental changes like this which do not affect the PR are welcome to come in as separate PRs that we can approve fast. If possible it is good to at least have them in their own commits so the PR is atomic when it lands on our dev branch.

framework/decode/file_processor.h
```diff
@@ -1,6 +1,7 @@
 /*
 ** Copyright (c) 2018 Valve Corporation
 ** Copyright (c) 2018 LunarG, Inc.
+** Copyright (c) 2023 Advanced Micro Devices, Inc. All rights reserved.
 **
```
andrew-lunarg, line 5: Interesting question: since you deleted a line below, and didn't add a line, should you add a copyright notice here??? 🤔

framework/decode/file_processor.h
feimao-gfxr: This change has been reverted. Will be fixed with another PR.

framework/decode/file_processor.h
feimao-gfxr: The change has been reverted.

framework/decode/custom_ags_ascii_consumer.cpp
```diff
@@ -0,0 +1,745 @@
...
+    if (iid == IID_ID3D12DebugDevice1)
+        return "\"IID_ID3D12DebugDevice1\"";
+    if (iid == IID_ID3D12DebugDevice)
+        return "\"IID_ID3D12DebugDevice\"";
+    if (iid == IID_ID3D12DebugDevice2)
+        return "\"IID_ID3D12DebugDevice2\"";
+    return "\"Invalid IID\"";
+}
+
+std::string FeatureLevelToString(const D3D_FEATURE_LEVEL& value)
```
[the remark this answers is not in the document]
feimao-gfxr: Changed as suggested.

framework/decode/custom_ags_ascii_consumer.cpp
```diff
@@ -0,0 +1,701 @@
...
+                                        tab_count,
+                                        tab_size,
+                                        ToString(creationParams->iid, to_string_flags_, tab_count, tab_size));
+                          FieldToString(strStrm,
+                                        false,
+                                        "FeatureLevel",
+                                        to_string_flags_,
+                                        tab_count,
+                                        tab_size,
+                                        FeatureLevelToString(creationParams->FeatureLevel));
```
davidd-lunarg, line 258: Can this also use the `ToString` method from `generated_dx12_enum_to_string.h` and the function `std::string FeatureLevelToString(const D3D_FEATURE_LEVEL& value)` above be removed?

framework/decode/file_processor.h
andrew-lunarg: Great, I wasn't expecting that for this PR, but thank you!

andrew-lunarg (changes requested):
Taking my approval off the PR and requesting the duplicate DX12 enum to string functions are removed in favour of our existing `generated_dx12_enum_to_string.h`, unless there is something missing from them that these ones in the PR provide. In particular the changes requested by David in comments.

framework/decode/custom_ags_ascii_consumer.cpp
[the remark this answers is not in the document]
andrew-lunarg: It seems to still be here (`FeatureLevelToString`).

framework/decode/custom_ags_ascii_consumer.cpp
feimao-gfxr: Done. Sorry for the delay.

davidd-lunarg (approved):
LGTM although this PR branch is more than 100 commits behind `dev` so it might be a good idea to rebase manually before merging.

feimao-gfxr:
Closed, because it is replaced by PR #1360.

PR #1213, opened 2023-07-25
Add --flush-inside-measurement-range option

marius-pelegrin-arm:
Add option to flush and wait all current GPU work to finish at the end of every frame inside the measurement range.

This allows to take measurements on each frame individually without jobs interleaving between two frames.


MarkY-LunarG (approved):
Looks good to me.  I just added another commit to add to the documentation as well as support for the script.

PR #1216, opened 2023-07-31
Allow ags.lib/agsd.lib for win32 replayer/optimizer

feimao-gfxr:
**Problem**
We may need to use AGS functions, such as agsDriverExtensionsDX12_PushMarker, through statically linking to ags.lib or agsd.lib for replay project or optimize project. Currently the required libs are included for x64 builds of the replay project and the optimize project. The AGS libs are missing for win32 builds. 

**Solution**
Adjust related AGS related flags in CMakeLists files. 

**Result**
ags.lib/agsd.lib show up in the link input list for win32 builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -190,8 +190,8 @@ if(MSVC)
     
     if (${D3D12_SUPPORT})
         if (${GFXRECON_AGS_SUPPORT})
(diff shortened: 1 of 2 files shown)


feimao-gfxr:
This change is not currently required.

PR #1221, opened 2023-08-08
Print differences in supported extensions and features

panos-lunarg:
In case vkCreateInstance and vkCreateDevice fail at replay time because of an EXTENSION_NOT_PRESENT error, should now print which were the extensions that were not supported on the replay gpu and triggered the error.

Similarly features requested with vkCreateDevice during capture will be compared with what the replay device offers and print a warning for each feature not supported.

### DIFF
--- a/framework/generated/generated_vulkan_feature_util.cpp
+++ b/framework/generated/generated_vulkan_feature_util.cpp
@@ -34,7 +34,12 @@ GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 GFXRECON_BEGIN_NAMESPACE(feature_util)
 
-void RemoveUnsupportedFeatures(VkPhysicalDevice physicalDevice, PFN_vkGetPhysicalDeviceFeatures GetPhysicalDeviceFeatures, PFN_vkGetPhysicalDeviceFeatures2 GetPhysicalDeviceFeatures2, const void* pNext, const VkPhysicalDeviceFeatures* pEnabledFeatures)
+void CheckUnsupportedFeatures(VkPhysicalDevice physicalDevice,
+                             PFN_vkGetPhysicalDeviceFeatures  GetPhysicalDeviceFeatures,
+                             PFN_vkGetPhysicalDeviceFeatures2 GetPhysicalDeviceFeatures2,
+                             const void*                      pNext,
+                             const VkPhysicalDeviceFeatures*  pEnabledFeatures,
+                             bool                             remove_unsupported)
 {
     // If the pNext chain includes a VkPhysicalDeviceFeatures2 structure, then pEnabledFeatures must be NULL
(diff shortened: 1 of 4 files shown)


framework/generated/generated_vulkan_feature_util.cpp
```diff
@@ -63,22 +63,50 @@ void RemoveUnsupportedFeatures(VkPhysicalDevice physicalDevice, PFN_vkGetPhysica
...
                 if ((currentNext->storageBuffer16BitAccess == VK_TRUE) && (query.storageBuffer16BitAccess == VK_FALSE))
                 {
-                    GFXRECON_LOG_WARNING("Feature storageBuffer16BitAccess, which is not supported by the replay device, will not be enabled");
+                    if (remove)
+                    {
+                        GFXRECON_LOG_WARNING("Feature storageBuffer16BitAccess, which is not supported by the replay device, will not be enabled");
+                    }
+                    else
+                    {
+                        GFXRECON_LOG_WARNING("Feature storageBuffer16BitAccess requested at catpure is not supported by the replay device");
```
per-mathisen-arm, line 72: "catpure" - I love this typo ;-)

I would prepare part of the string in advance and save quite a few conditionals here, would make code easier to read.

panos-lunarg: Oops missed your comment. Will fix

framework/generated/generated_vulkan_feature_util.cpp
```diff
@@ -63,23 +79,43 @@ void RemoveUnsupportedFeatures(VkPhysicalDevice physicalDevice, PFN_vkGetPhysica
...
                 if ((currentNext->storageBuffer16BitAccess == VK_TRUE) && (query.storageBuffer16BitAccess == VK_FALSE))
                 {
-                    GFXRECON_LOG_WARNING("Feature storageBuffer16BitAccess, which is not supported by the replay device, will not be enabled");
-                    const_cast<VkPhysicalDevice16BitStorageFeatures*>(currentNext)->storageBuffer16BitAccess = VK_FALSE;
+                    GFXRECON_LOG_WARNING("Feature storageBuffer16BitAccess %s", warn_message);
+                    found_unsupported = true;
+
+                    if (remove_unsupported)
+                    {
+                        const_cast<VkPhysicalDevice16BitStorageFeatures*>(currentNext)->storageBuffer16BitAccess = VK_FALSE;
```
per-mathisen-arm, line 87: If you want to save even more lines of code, you could do "... = !remove_unsupported;" here instead of the if clause :-)

panos-lunarg:
I realized that `RemoveIgnorableExtensions` already implements the logic which prints the unsupported extensions both for `vkCreateInstance` and `vkCreateDevice`, so i removed that code. The unsupported features are not listed though

bradgrantham-lunarg (approved):

bradgrantham-lunarg (approved):

Issue #1122, opened 2023-05-10
Virtual swapchain causes screenshots mismatch for recaptured files.


locke-lunarg:
`WriteScreenshots` has a `DeviceWaitIdle`. `DeviceWaitIdle` could fix sync issues to generate a correct screenshot, even if it has any tasks that are uncompleted. But in some cases, `virtual swapchain` could cause `DeviceWaitIdle` helpless.

If virtual swapchain is enabled during re-capturing, the virtual swapchain code would be written to the file. And then during replay this recaptured file, it will run the captured virtual swapchain code, like
```
  "index": 335713,
  "function": {
    "name": "vkCmdBlitImage",
    …
      "srcImage": 73,
      "dstImage": 70,
    …
       }
```
`srcImage: 73` is the captured virtual swapchain image. `dstImage: 70` is a new virtual swapchain image. It copies a old virtual swapchain image to a new virtual swapchain image. The old virtual swapchain image could be an uncompleted image if it has sync issues, and then, output a screenshot (image: 70). Even though, `WriteScreenshots` has a `DeviceWaitIdle`, it's behind `vkCmdBlitImage`. It can't help `image: 70` that has already been copied from an uncompleted image. If virtual swapchain is disable during recapturing processing, it wouldn't have this issue. 

However, if the recapturing process is with screenshots, the `WriteScreenshots` code would be also written to the file. And the captured `WriteScreenshots` code is above the captured `virtual swapchain` code, so it would run `DeviceWaitIdle` before `vkCmdBlitImage`, so the new virtual swapchain image would be copied an completed old virtual swapchain image. The screenshot would be good.

`--sync` could fix this issue. But how could we fix this bug in the code?

locke-lunarg:
This issue should be closed. This issue comes from apps that have sync issues. It's not a gfxr issue. It's reasonable to use `--syn` to fix it. I don't think we could have a better idea than `--syn` for this case.

bradgrantham-lunarg:
I am not 100% convinced we don't have some issue with screenshot.  The app seems to work fine on its own, so I want to know for sure that the app has incorrect usage of synchronization under Vulkan and we're doing our synchronization correctly.  Mark volunteered to take a look.

MarkY-LunarG:
[...]

PR #1222, opened 2023-08-09
Modify virtual swapchain behavior

MarkY-LunarG:
Modify virtual swapchain to create a command buffer per queue and when we get a vkQueuePresent, perform the blits in the same queue. This resolves the issue where a QueuePresent is presented inline in a queue, but has no wait semaphore information.  Instead, the QueuePresent relied upon the previous submit (also in the same queue) to perform the synchronization.

Previously, the virtual swapchain had it's own Queue that it used which caused a race condition resulting in old images being used.

Also, instead of doing just a waitforidle, wait for all pending semaphores (which could include some that are pending but not submitted).
Also, don't wait for idle after the screenshot submit, wait for the fence.  This way we don't have to keep idling the GPU.

Fixes: #1122

### DIFF
--- a/framework/decode/vulkan_captured_swapchain.cpp
+++ b/framework/decode/vulkan_captured_swapchain.cpp
@@ -240,6 +240,37 @@ VkResult VulkanCapturedSwapchain::CreateRenderPass2(PFN_vkCreateRenderPass2     
     return func(device, create_info, allocator, render_pass);
 }
 
+void VulkanCapturedSwapchain::CmdPipelineBarrier(PFN_vkCmdPipelineBarrier     func,
+                                                 const CommandBufferInfo*     command_buffer_info,
+                                                 VkPipelineStageFlags         src_stage_mask,
+                                                 VkPipelineStageFlags         dst_stage_mask,
+                                                 VkDependencyFlags            dependency_flags,
+                                                 uint32_t                     memory_barrier_count,
+                                                 const VkMemoryBarrier*       memory_barriers,
+                                                 uint32_t                     buffer_memory_barrier_count,
+                                                 const VkBufferMemoryBarrier* buffer_memory_barriers,
+                                                 uint32_t                     image_memory_barrier_count,
+                                                 const VkImageMemoryBarrier*  image_memory_barriers)
+{
+    VkCommandBuffer command_buffer = VK_NULL_HANDLE;
+
+    if (command_buffer_info != nullptr)
+    {
+        command_buffer = command_buffer_info->handle;
+    }
+
+    func(command_buffer,
+         src_stage_mask,
+         dst_stage_mask,
+         dependency_flags,
+         memory_barrier_count,
+         memory_barriers,
+         buffer_memory_barrier_count,
+         buffer_memory_barriers,
+         image_memory_barrier_count,
+         image_memory_barriers);
+}
+
 void VulkanCapturedSwapchain::ProcessSetSwapchainImageStateCommand(
     const DeviceInfo*                                   device_info,
     SwapchainKHRInfo*                                   swapchain_info,
--- a/framework/decode/vulkan_captured_swapchain.cpp
+++ b/framework/decode/vulkan_captured_swapchain.cpp
@@ -240,37 +240,6 @@ VkResult VulkanCapturedSwapchain::CreateRenderPass2(PFN_vkCreateRenderPass2     
     return func(device, create_info, allocator, render_pass);
 }
 
-void VulkanCapturedSwapchain::CmdPipelineBarrier(PFN_vkCmdPipelineBarrier     func,
-                                                 const CommandBufferInfo*     command_buffer_info,
-                                                 VkPipelineStageFlags         src_stage_mask,
-                                                 VkPipelineStageFlags         dst_stage_mask,
-                                                 VkDependencyFlags            dependency_flags,
-                                                 uint32_t                     memory_barrier_count,
-                                                 const VkMemoryBarrier*       memory_barriers,
-                                                 uint32_t                     buffer_memory_barrier_count,
-                                                 const VkBufferMemoryBarrier* buffer_memory_barriers,
-                                                 uint32_t                     image_memory_barrier_count,
-                                                 const VkImageMemoryBarrier*  image_memory_barriers)
-{
-    VkCommandBuffer command_buffer = VK_NULL_HANDLE;
-
-    if (command_buffer_info != nullptr)
-    {
-        command_buffer = command_buffer_info->handle;
-    }
-
-    func(command_buffer,
-         src_stage_mask,
-         dst_stage_mask,
-         dependency_flags,
-         memory_barrier_count,
-         memory_barriers,
-         buffer_memory_barrier_count,
-         buffer_memory_barriers,
-         image_memory_barrier_count,
-         image_memory_barriers);
-}
-
 void VulkanCapturedSwapchain::ProcessSetSwapchainImageStateCommand(
     const DeviceInfo*                                   device_info,
     SwapchainKHRInfo*                                   swapchain_info,
@@ -279,7 +248,6 @@ void VulkanCapturedSwapchain::ProcessSetSwapchainImageStateCommand(
     const VulkanObjectInfoTable&                        object_info_table,
     SwapchainImageTracker&                              swapchain_image_tracker)
 {
-
     VkDevice       device    = device_info->handle;
     VkSwapchainKHR swapchain = swapchain_info->handle;
 
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -266,6 +266,7 @@ struct DeviceInfo : public VulkanObjectInfo<VkDevice>
     graphics::VulkanDevicePropertyFeatureInfo property_feature_info;
 
     std::unordered_map<uint32_t, VkDeviceQueueCreateFlags> queue_family_creation_flags;
+    std::vector<bool>                                      queue_family_index_enabled;
 
     std::vector<VkPhysicalDevice> replay_device_group;
 };
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -400,37 +400,16 @@ struct SwapchainKHRInfo : public VulkanObjectInfo<VkSwapchainKHR>
 
     // The following values are only used when loading the initial state for trimmed files.
     std::vector<uint32_t> queue_family_indices{ 0 };
-    uint32_t              virtual_queue_index{ 0 };
 
     // When replay is restricted to a specific surface, a dummy swapchain is created for the omitted surfaces, requiring
     // backing images.
+    uint32_t                  replay_image_count{ 0 };
     std::vector<ImageInfo>    image_infos;
     VkSwapchainCreateFlagsKHR image_flags{ 0 };
     VkFormat                  image_format{ VK_FORMAT_UNDEFINED };
     uint32_t                  image_array_layers{ 0 };
     VkImageUsageFlags         image_usage{ 0 };
     VkSharingMode             image_sharing_mode{ VK_SHARING_MODE_EXCLUSIVE };
-
-    // TODO: These values are used by the virtual swapchain.  They should be replaced with an opaque handle, similar to
-    // DeviceMemoryInfo::allocator_data, which is really a pointer to a struct that contains the virtual swapchain's
-    // internal info.  The memory for the struct referenced by the opaque handle would be managed by the virtual
-    // swapchain class, similar to the way that the VulkanRebindAllocator works.
-    struct VirtualImage
-    {
-        VkDeviceMemory                        memory{ VK_NULL_HANDLE };
-        VkImage                               image{ VK_NULL_HANDLE };
-        VulkanResourceAllocator::MemoryData   memory_allocator_data{ 0 };
-        VulkanResourceAllocator::ResourceData resource_allocator_data{ 0 };
-    };
-    uint32_t              replay_image_count{ 0 };
-    std::vector<uint32_t> capture_to_replay_index;
-    std::vector<VirtualImage>
-        virtual_swapchain_images; // Images created by replay, returned in place of the swapchain images.
-    std::vector<VkImage> replay_swapchain_images; // The real swapchain images.
-
-    std::vector<VkCommandPool>                blit_command_pools;
-    std::vector<std::vector<VkCommandBuffer>> blit_command_buffers;
-    std::vector<std::vector<VkSemaphore>>     blit_semaphores;
 };
 
 struct ValidationCacheEXTInfo : public VulkanObjectInfo<VkValidationCacheEXT>
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -266,6 +266,8 @@ struct DeviceInfo : public VulkanObjectInfo<VkDevice>
     graphics::VulkanDevicePropertyFeatureInfo property_feature_info;
 
     std::unordered_map<uint32_t, VkDeviceQueueCreateFlags> queue_family_creation_flags;
+    std::unordered_map<VkQueue, uint32_t>                  queue_family_indices;
+    std::unordered_map<VkQueue, uint32_t>                  queue_indices;
 
     std::vector<VkPhysicalDevice> replay_device_group;
 };
@@ -398,6 +400,7 @@ struct SwapchainKHRInfo : public VulkanObjectInfo<VkSwapchainKHR>
 
     // The following values are only used when loading the initial state for trimmed files.
     std::vector<uint32_t> queue_family_indices{ 0 };
+    uint32_t              virtual_queue_index{ 0 };
 
     // When replay is restricted to a specific surface, a dummy swapchain is created for the omitted surfaces, requiring
     // backing images.
@@ -419,13 +422,15 @@ struct SwapchainKHRInfo : public VulkanObjectInfo<VkSwapchainKHR>
         VulkanResourceAllocator::MemoryData   memory_allocator_data{ 0 };
         VulkanResourceAllocator::ResourceData resource_allocator_data{ 0 };
     };
-    uint32_t                     replay_image_count{ 0 };
-    std::vector<VirtualImage>    virtual_images; // Images created by replay, returned in place of the swapchain images.
-    std::vector<VkImage>         swapchain_images; // The real swapchain images.
-    VkQueue                      blit_queue{ VK_NULL_HANDLE };
-    VkCommandPool                blit_command_pool{ VK_NULL_HANDLE };
-    std::vector<VkCommandBuffer> blit_command_buffers;
-    std::vector<VkSemaphore>     blit_semaphores;
+    uint32_t              replay_image_count{ 0 };
+    std::vector<uint32_t> capture_to_replay_index;
+    std::vector<VirtualImage>
+        virtual_swapchain_images; // Images created by replay, returned in place of the swapchain images.
+    std::vector<VkImage> replay_swapchain_images; // The real swapchain images.
+
+    std::vector<VkCommandPool>                blit_command_pools;
+    std::vector<std::vector<VkCommandBuffer>> blit_command_buffers;
+    std::vector<std::vector<VkSemaphore>>     blit_semaphores;
 };
 
 struct ValidationCacheEXTInfo : public VulkanObjectInfo<VkValidationCacheEXT>
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -266,8 +266,6 @@ struct DeviceInfo : public VulkanObjectInfo<VkDevice>
     graphics::VulkanDevicePropertyFeatureInfo property_feature_info;
 
     std::unordered_map<uint32_t, VkDeviceQueueCreateFlags> queue_family_creation_flags;
-    std::unordered_map<VkQueue, uint32_t>                  queue_family_indices;
-    std::unordered_map<VkQueue, uint32_t>                  queue_indices;
 
     std::vector<VkPhysicalDevice> replay_device_group;
 };
@@ -275,6 +273,8 @@ struct DeviceInfo : public VulkanObjectInfo<VkDevice>
 struct QueueInfo : public VulkanObjectInfo<VkQueue>
 {
     std::unordered_map<uint32_t, size_t> array_counts;
+    uint32_t                             family_index;
+    uint32_t                             queue_index;
 };
 
 struct SemaphoreInfo : public VulkanObjectInfo<VkSemaphore>
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -41,8 +41,10 @@
 
 #include "generated/generated_vulkan_enum_to_string.h"
 
+#include <algorithm>
 #include <cstdint>
 #include <limits>
+#include <numeric>
 #include <unordered_set>
 #include <future>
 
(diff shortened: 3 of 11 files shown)


framework/decode/vulkan_captured_swapchain.cpp
```diff
@@ -240,37 +240,6 @@ VkResult VulkanCapturedSwapchain::CreateRenderPass2(PFN_vkCreateRenderPass2
...
-         src_stage_mask,
-         dst_stage_mask,
-         dependency_flags,
-         memory_barrier_count,
-         memory_barriers,
-         buffer_memory_barrier_count,
-         buffer_memory_barriers,
-         image_memory_barrier_count,
-         image_memory_barriers);
-}
```
MarkY-LunarG, line 272: This code was really useless and over complicated the swapchain path.  Just removed it here and added the direct function call in the consumer base above.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2592,8 +2599,9 @@ VulkanReplayConsumerBase::OverrideCreateDevice(VkResult            original_resu
             for (uint32_t q = 0; q < modified_create_info.queueCreateInfoCount; ++q)
             {
                 const VkDeviceQueueCreateInfo* queue_create_info = &modified_create_info.pQueueCreateInfos[q];
-                assert(device_info->queue_family_creation_flags.find(queue_create_info->queueFamilyIndex) ==
-                       device_info->queue_family_creation_flags.end());
+                if (queue_create_info)
```
MarkY-LunarG, line 2602: Not sure where this came from.  I'll remove it in an update.

MarkY-LunarG:
Moved to draft because I can't figure out why it's failing on internal CI systems yet.  I'll move it back when that runs clean.

MarkY-LunarG:
Right now I create 1 command buffer per Graphics queue on the device per swapchain image we're going to create.  This removes all race conditions I encountered.  If we don't have one per swapchain image, we can get queue presents coming in and messing up the command buffer before its been fully submitted. 

Some changes I could do but haven't (and maybe you think these are important:
 1) Only create pools, buffers, and semaphores for queues encountered during vkQueuePresent
 2) Use a synchronization priimitive (lock/mutex) to prevent command buffers while a queue is submiting

The first seems like a good idea, I think the second isn't worth it.

framework/decode/vulkan_object_info.h
```diff
@@ -268,6 +268,8 @@ struct DeviceInfo : public VulkanObjectInfo<VkDevice>
     graphics::VulkanDevicePropertyFeatureInfo property_feature_info;
 
     std::unordered_map<uint32_t, VkDeviceQueueCreateFlags> queue_family_creation_flags;
+    std::unordered_map<VkQueue, uint32_t>                  queue_family_indices;
+    std::unordered_map<VkQueue, uint32_t>                  queue_indices;
```
locke-lunarg, line 272: I think putting `queue_family_index` and `queue_index` in `QueueInfo` is better.

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -132,81 +135,153 @@ VkResult VulkanVirtualSwapchain::GetSwapchainImagesKHR(PFN_vkGetSwapchainImagesK
...
             {
-                VkCommandBufferAllocateInfo allocate_info = { VK_STRUCTURE_TYPE_COMMAND_BUFFER_ALLOCATE_INFO };
-                allocate_info.pNext                       = nullptr;
-                allocate_info.commandPool                 = swapchain_info->blit_command_pool;
-                allocate_info.level                       = VK_COMMAND_BUFFER_LEVEL_PRIMARY;
-                allocate_info.commandBufferCount          = 1;
-                VkCommandBuffer command_buffer            = VK_NULL_HANDLE;
-                result = device_table_->AllocateCommandBuffers(device, &allocate_info, &command_buffer);
-                if (result != VK_SUCCESS)
+                // Create one command pool per queue.
```
locke-lunarg, line 195: Maybe we could check if the family index of Queue has VK_QUEUE_GRAPHICS_BIT, and then to Create CommandPool, CommandBuffers and Semaphores

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -132,81 +135,153 @@ VkResult VulkanVirtualSwapchain::GetSwapchainImagesKHR(PFN_vkGetSwapchainImagesK
...
+                if (swapchain_info->blit_command_pools[queue_family_index] == VK_NULL_HANDLE)
```
locke-lunarg, line 196: Maybe we could check if the family index of Queue has VK_QUEUE_GRAPHICS_BIT, and then to Create CommandPool, CommandBuffers and Semaphores

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -416,196 +534,108 @@ VkResult VulkanVirtualSwapchain::QueuePresentKHR(PFN_vkQueuePresentKHR
...
-        const auto* swapchain_info      = swapchain_infos[i];
-        const auto& virtual_image       = swapchain_info->virtual_images[capture_image_index];
-        const auto& swapchain_image     = swapchain_info->swapchain_images[replay_image_index];
+        const auto& virtual_image       = swapchain_info->virtual_swapchain_images[capture_image_index];
+        const auto& replay_image        = swapchain_info->replay_swapchain_images[replay_image_index];
+
+        // Use a command buffer and semaphore from the same queue index
+        assert(swapchain_info->device_info->queue_family_indices.find(queue) !=
+               swapchain_info->device_info->queue_family_indices.end());
+        uint32_t queue_family_index = swapchain_info->device_info->queue_family_indices[queue];
```
locke-lunarg, line 562: If queue_family_index saves in QueueInfo, we don't need to find the Queue in DeviceInfo here.

locke-lunarg:
Virtual swapchain's Queue requires VK_QUEUE_TRANSFER_BIT. I don't know if it's possible that PresentQueue use a Queue that doesn't support VK_QUEUE_TRANSFER_BIT? or we can print a error when it happens.

framework/decode/vulkan_virtual_swapchain.cpp
MarkY-LunarG: Its not just used for Graphics, the queues used are also Transfer and compute.  At least for Wolfenstein.  I could initially just do the graphics ones, but I can create they dynamically on a QueuePresent of a new Queue.  Is that what you mean?

framework/decode/vulkan_object_info.h
MarkY-LunarG: The reason I didn't add it to the queue is this is defined during vkGetDeviceQueue and the logic to create the QueueInfo occurs after the override function is called.  So the only Info structure I had access to is the DeviceInfo struct.  Is there a way to have a `PostOverride` function?

```

void VulkanReplayConsumer::Process_vkGetDeviceQueue(
    format::HandleId                            device,
    uint32_t                                    queueFamilyIndex,
    uint32_t                                    queueIndex,
    HandlePointerDecoder<VkQueue>*              pQueue)
{
    auto in_device = GetObjectInfoTable().GetDeviceInfo(device);
    if (!pQueue->IsNull()) { pQueue->SetHandleLength(1); }
    QueueInfo handle_info;
    pQueue->SetConsumerData(0, &handle_info);

    OverrideGetDeviceQueue(GetDeviceTable(in_device->handle)->GetDeviceQueue, in_device, queueFamilyIndex, queueIndex, pQueue);

    AddHandle<QueueInfo>(device, pQueue->GetPointer(), pQueue->GetHandlePointer(), std::move(handle_info), &VulkanObjectInfoTable::AddQueueInfo);
}
```

framework/decode/vulkan_virtual_swapchain.cpp
MarkY-LunarG: See above.

framework/decode/vulkan_virtual_swapchain.cpp
locke-lunarg: No. I don't mean creating dynamically. I mean Queue support Graphics, not mean only support Graphics. For Example `Graphics` , `Graphics | Compute | Transfer` ,  `Graphics | Transfer` of Queues that we need. But `Compute | Transfer` , `Transfer`  of Queues that we don't need.

MarkY-LunarG: I could, but that doesn't mean another queue type doesn't come along and get used.  That's my concern.  But I can change it to do that for now.  @bradgrantham , what do you think?

framework/decode/vulkan_object_info.h
locke-lunarg: You can use 
`auto queue_info = reinterpret_cast<QueueInfo*>(pQueue->GetConsumerData(0));` 
to get `queue_info`  in `OverrideGetDeviceQueue`. 

You  can see
```
QueueInfo handle_info;
pQueue->SetConsumerData(0, &handle_info);
...
AddHandle<QueueInfo>(device, pQueue->GetPointer(), pQueue->GetHandlePointer(), std::move(handle_info), &VulkanObjectInfoTable::AddQueueInfo);
```
in `Process_vkGetDeviceQueue`.

`handle_info` is moved into table in  AddHandle.

framework/decode/vulkan_virtual_swapchain.cpp
locke-lunarg: It's ok to keep those queues if they might be used somewhere.

framework/decode/vulkan_object_info.h
MarkY-LunarG: That's what I was hoping for!  Thanks @locke-lunarg

framework/decode/vulkan_virtual_swapchain.cpp
MarkY-LunarG: Done.

framework/decode/vulkan_object_info.h
MarkY-LunarG: Done.

framework/decode/vulkan_virtual_swapchain.cpp
MarkY-LunarG: @bradgrantham-lunarg , the problem is that `VulkanVirtualSwapchain` only has 1 instance created per run and there could be multiple swapchains per multiple devices.  That's why all the data is stored in the `SwapchainInfo` which is created per swapchain per device.  I think the ideal would be if there was a `Swapchain` class and the `VirtualSwapchain` had been a derived child class and then data could have been stored in the class, but we don't have that design currently. 

Also, as I mentioned a transfer and a compute queue (and even a video queue) could have a `vkQueuePresent` called on it and we would run into the same situation.  For example, a transfer or compute workload might modify the swapchain image with data (let's say an ambient occlusion path). 

Of course, for video-only we probably would currently break. In that case, we'd need to submit an empty command buffer in the video-only queue that triggered a semaphore (let's say SemA), then the BLIT would be added to a graphics queue waiting on SemA to start and signalling SemB when done, then we would put wait for SemB to start the `Present` in the video only queue.

MarkY-LunarG: **NOTE** I will put a TODO comment about the video-only problem in the code, just so people are aware.

MarkY-LunarG: Ok, I now use `vkGetPhysicalDeviceSurfaceSupportKHR` to query which Queue Families can perform `VkQueuePresentKHR`.  If they do, then only for those do I create the command pool and buffers.

locke-lunarg: I think we should rename `blit_*` because we don't use `CmdBlitImage` any more, instead of `CmdCopyImage`.

locke-lunarg:
When I added `--validate` to enable validation layers to replay a capturing file of vkcube, it generated some VVL error messages. They didn't exist on branch dev.

framework/decode/vulkan_virtual_swapchain.h
```diff
@@ -118,15 +106,32 @@ class VulkanVirtualSwapchain : public VulkanSwapchain
...
+    // VirtualSwapchain-specific private information.
+    struct VirtualImage
+    {
+        VkDeviceMemory                        memory{ VK_NULL_HANDLE };
+        VkImage                               image{ VK_NULL_HANDLE };
+        VulkanResourceAllocator::MemoryData   memory_allocator_data{ 0 };
+        VulkanResourceAllocator::ResourceData resource_allocator_data{ 0 };
+    };
+
+    struct PrivateData
```
bradgrantham-lunarg, line 118: I think `PrivateData `should probably have a constructor which does all the work of creating these objects (and so may throw an exception or test failure some other way) and also have all the data it needs inside it to destroy itself (probably needs the device handle).  Then the `unordered_map `on `VkSwapchainKHR `should probably have `PrivateData` as the value and not a pointer, and the entry in the map should be emplaced with move semantics.  Then it correctly deletes itself when you call 

If that's not practical, `private_data_` should probably be an unordered_map with `unique_ptr<PrivateData>` values.

Also I suggest it should probably be named something more descriptive, like `SwapchainData`.

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -130,103 +163,260 @@ VkResult VulkanVirtualSwapchain::GetSwapchainImagesKHR(PFN_vkGetSwapchainImagesK
...
+            private_data->replay_swapchain_images.resize(*replay_image_count);
         }
     }
 
     // TODO: Adjust the swapchain image format if the specified format is not supported by the replay device.
 
-    auto result = func(device, swapchain, replay_image_count, replay_swapchain_images.data());
+    result = func(device, swapchain, replay_image_count, private_data->replay_swapchain_images.data());
+
+    if (static_cast<uint32_t>(private_data->capture_to_replay_index.size()) != capture_image_count)
```
bradgrantham-lunarg, line 186: I think in general the `static_cast` here should be on `capture_image_count` instead, up to `size_t`, because the result of `size` may be 64 bits and so the static cast on `size()` loses information.  It seems so unlikely though, that size() would return > 2^32 that maybe I don't care if you don't have time to make the change.

framework/decode/vulkan_virtual_swapchain.h
```diff
@@ -118,15 +106,32 @@ class VulkanVirtualSwapchain : public VulkanSwapchain
...
+    {
+        std::vector<VkCommandPool>                copy_command_pools;
```
bradgrantham-lunarg, line 120: Are these all indexed by queue index or queue family index?  Could they be part of a subobject?  Then you could even put them in a map and not worry about resizing all these vectors, etc.

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -458,56 +689,80 @@ VkResult VulkanVirtualSwapchain::QueuePresentKHR(PFN_vkQueuePresentKHR
...
+        auto command_buffer = private_data->copy_command_buffers[queue_family_index][capture_image_index];
+        auto copy_semaphore = private_data->copy_semaphores[queue_family_index][capture_image_index];
+
+        std::vector<VkSemaphore> wait_semaphores;
+        std::vector<VkSemaphore> signal_semaphores;
+
+        // Only wait on the first copy
+        if (i == 0)
+        {
+            if (present_info->waitSemaphoreCount > 0)
```
bradgrantham-lunarg, line 735: You can replace all this with a call to `std::fill` or `std::fill_n` with `back_inserter` (https://en.cppreference.com/w/cpp/iterator/back_inserter)

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2589,13 +2589,31 @@ VulkanReplayConsumerBase::OverrideCreateDevice(VkResult            original_resu
...
+            for (uint32_t q = 0; q < modified_create_info.queueCreateInfoCount; ++q)
+            {
+                const VkDeviceQueueCreateInfo* queue_create_info = &modified_create_info.pQueueCreateInfos[q];
+                if (queue_create_info->queueFamilyIndex > max_queue_family)
+                {
+                    max_queue_family = queue_create_info->queueFamilyIndex;
+                }
+            }
+            device_info->queue_family_index_enabled.resize(max_queue_family + 1);
+            for (uint32_t q = 0; q <= max_queue_family; ++q)
```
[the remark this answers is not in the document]
MarkY-LunarG: Looking at using your lambda example instead:

```
           auto max = [](uint32_t current_max, const VkDeviceQueueCreateInfo& dqci) {
                return std::max(current_max, dqci.queueFamilyIndex);
            };
            uint32_t max_queue_family =
                std::accumulate(modified_create_info.pQueueCreateInfos,
                                modified_create_info.pQueueCreateInfos + modified_create_info.queueCreateInfoCount,
                                0,
                                max);

 ```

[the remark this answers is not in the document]
Comment (bradgrantham-lunarg) [reply framework/decode/vulkan_replay_consumer_base.cpp:2604]: Well, *this* code where you set all the enabled vector to `false` is probably best for `std::fill`.  I assume you mean using `accumulate` on the loop above.

After looking into this some more, `resize` can take a value, so the most compact way to set this vector to `false` is probably just to `clear` it and then `resize(max_queue_family + 1, false)`.

File: framework/decode/vulkan_virtual_swapchain.h (hunk 7 above)
Comment (MarkY-LunarG) [reply framework/decode/vulkan_virtual_swapchain.h:118]: I'll convert to a unique_ptr vector and rename it to SwapchainData.

File: framework/decode/vulkan_virtual_swapchain.h (hunk 9 above)
Comment (MarkY-LunarG) [reply framework/decode/vulkan_virtual_swapchain.h:120]: I modified it to have a map in the `SwapchainData` associated per queue.  This will be ordered and should only be created for Queue families of interest.

File: framework/decode/vulkan_virtual_swapchain.cpp (hunk 10 above)
Comment (MarkY-LunarG) [reply framework/decode/vulkan_virtual_swapchain.cpp:735]: That does not work for C style arrays.  Instead I can do the following since it's a first assignment:

```
wait_semaphores.assign(present_info->pWaitSemaphores,
                       present_info->pWaitSemaphores + present_info->waitSemaphoreCount);
```

Comment (bradgrantham-lunarg) [reply framework/decode/vulkan_virtual_swapchain.cpp]: std::fill does work on C arrays but I agree if you have something that works well and is idiomatic, there's no need to change it.

Comment (bradgrantham-lunarg) [reply framework/decode/vulkan_virtual_swapchain.cpp]: I see - I think the original code I commented on was filling a `vector<bool>` but it's changed since then to be an assignment into a vector and for that case I think `assign` is great!

Comment (MarkY-LunarG) [reply framework/decode/vulkan_virtual_swapchain.cpp]: Done

File: framework/decode/vulkan_replay_consumer_base.cpp [hunk 12]
```diff
@@ -2589,13 +2589,31 @@ VulkanReplayConsumerBase::OverrideCreateDevice(VkResult            original_resu
             // Track state of physical device properties and features at device creation
             device_info->property_feature_info = property_feature_info;
 
+            // Keep track of what queue families this device is planning on using.  This information is
+            // very important if we end up using the VulkanVirtualSwapchain path.
+            uint32_t max_queue_family = 0;
+            for (uint32_t q = 0; q < modified_create_info.queueCreateInfoCount; ++q)
+            {
+                const VkDeviceQueueCreateInfo* queue_create_info = &modified_create_info.pQueueCreateInfos[q];
+                if (queue_create_info->queueFamilyIndex > max_queue_family)
```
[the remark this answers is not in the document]
MarkY-LunarG: Done

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -89,20 +101,40 @@ void VulkanVirtualSwapchain::DestroySwapchainKHR(PFN_vkDestroySwapchainKHR    fu
...
+            auto private_data = private_data_[swapchain];
+            if (private_data != nullptr)
+            {
+                for (const VirtualImage& image_info : private_data->virtual_swapchain_images)
+                {
+                    allocator->DestroyImageDirect(image_info.image, nullptr, image_info.resource_allocator_data);
+                    allocator->FreeMemoryDirect(image_info.memory, nullptr, image_info.memory_allocator_data);
+                }
+
+                for (uint32_t ii = 0; ii < static_cast<uint32_t>(private_data->copy_command_pools.size()); ++ii)
```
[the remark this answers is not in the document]
MarkY-LunarG: Done

framework/decode/vulkan_virtual_swapchain.h
MarkY-LunarG: Done

framework/decode/vulkan_virtual_swapchain.cpp
MarkY-LunarG: Done

framework/decode/vulkan_virtual_swapchain.h
MarkY-LunarG: Done

framework/decode/vulkan_replay_consumer_base.cpp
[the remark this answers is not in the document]
MarkY-LunarG: Done

framework/decode/vulkan_captured_swapchain.cpp
```diff
@@ -240,37 +240,6 @@ VkResult VulkanCapturedSwapchain::CreateRenderPass2(PFN_vkCreateRenderPass2
     return func(device, create_info, allocator, render_pass);
 }
 
-void VulkanCapturedSwapchain::CmdPipelineBarrier(PFN_vkCmdPipelineBarrier     func,
```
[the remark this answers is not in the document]
MarkY-LunarG: Done

locke-lunarg (approved):
Look good to me. Thank you, Mark.

MarkY-LunarG:
@bradgrantham-lunarg , does everything look good to you?  Locke approved this but I don't want to push it prematurely.

bradgrantham-lunarg (changes requested):
I requested three last changes, then I think this is ready to go.  Since I didn't mention these the last review, and repeatedly requesting changes probably gets annoying, I'd be happy to make them myself to this PR.  Would that work for you?

framework/decode/vulkan_virtual_swapchain.h
```diff
@@ -118,15 +120,43 @@ class VulkanVirtualSwapchain : public VulkanSwapchain
...
+        std::vector<VirtualImage> virtual_swapchain_images;
+        std::vector<VkImage>      replay_swapchain_images;
+    };
+
+    VkResult CreateVirtualSwapchainImage(const DeviceInfo*        device_info,
+                                         const VkImageCreateInfo& image_create_info,
+                                         VirtualImage&            image);
+
+    // Create an unordered map to associate the private information with a particular Vulkan swapchain
+    std::unordered_map<VkSwapchainKHR, std::unique_ptr<SwapchainData>> private_data_;
```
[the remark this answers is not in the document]
MarkY-LunarG: Renamed SwapchainResourceData and swapchain_resources_.

framework/decode/vulkan_virtual_swapchain.h
```diff
@@ -118,15 +120,43 @@ class VulkanVirtualSwapchain : public VulkanSwapchain
...
+        VkCommandPool                command_pool;
+        std::vector<VkCommandBuffer> command_buffers;
+        std::vector<VkSemaphore>     semaphores;
+    };
+
+    // Structure to track VulkanVirtualSwapchain private data specific to a particular VkSwapchainKHR handle.
+    struct SwapchainData
+    {
+        // Create a map that correlates copy command data with a queue family index.
+        std::map<size_t, CopyCmdData> copy_cmd_data;
```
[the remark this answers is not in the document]
MarkY-LunarG: `std::map` is faster. especially if the items are ordered, which they are.  That's why I used map.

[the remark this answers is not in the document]
bradgrantham-lunarg: unordered_map uses hashes and is O(1).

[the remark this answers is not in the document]
MarkY-LunarG: After discussion, Brad showed me that `std::map` is slower than `std::unordered_map` on average.  Switched.

PR #1226, opened 2023-08-14
Screenshot size during replay is now configurable

panos-lunarg:
Adds two new replay options that controls the screenshot dimensions:

`--screenshot-size <width>x<height>` explicitly sets the dimensions of the
generated screenshot

`--screenshot-scale <number>` Scales the swapchain images by that number.
Number can be decimal

### DIFF
--- a/framework/decode/screenshot_handler.cpp
+++ b/framework/decode/screenshot_handler.cpp
@@ -83,6 +83,8 @@ void ScreenshotHandler::WriteImage(const std::string&                      filen
                                    VkFormat                                format,
                                    uint32_t                                width,
                                    uint32_t                                height,
+                                   uint32_t                                copy_width,
+                                   uint32_t                                copy_height,
                                    VkImageLayout                           image_layout)
 {
     if ((device_table == nullptr) || (allocator == nullptr))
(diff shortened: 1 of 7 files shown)


framework/decode/screenshot_handler.cpp
```diff
@@ -92,6 +95,10 @@ void ScreenshotHandler::WriteImage(const std::string&                      filen
 
     VkResult result = VK_SUCCESS;
 
+    // If both copy_scale and copy_width are provided, use copy_scale.
```
bradgrantham-lunarg, line 98: I think the function that calls this function is a better location to do this conditional.  That is, receive only copy_width and copy_height in this function and calculate it in functions that call this function.

tools/tool_settings.h
```diff
@@ -464,6 +466,55 @@ static std::string GetScreenshotDir(const gfxrecon::util::ArgumentParser& arg_pa
...
+    {
+        std::size_t x = value.find("x");
+        if (x != std::string::npos)
+        {
+            try
+            {
+                width  = std::stoul(value.substr(0, x));
+                height = std::stoul(value.substr(x + 1));
+            }
+            catch (std::exception& e)
```
bradgrantham-lunarg, line 483: If any of these exceptions happened, that means we couldn't parse the screenshot size, right?  I think the user probably needs to be warned in each of these cases.

bradgrantham-lunarg (changes requested):

bradgrantham-lunarg:
Note that you may get a conflict with #1205, so please be cognizant of that PR if it goes in first.

bradgrantham-lunarg (approved):

PR #1233, opened 2023-08-22
Use system_properties.h api instead of getprop

### LINKED ISSUE #1231: Investigate use of the Android property API instead of calling `getprop`
If a continuous system query is performed, e.g. by Perfetto, and the capture trigger property is used, the process list will contain those calls to `getprop` command.

panos-lunarg:
Use the appropriate Android NDK Api to get property values instead of launching a new process with popen("getprop ...");

Fixes #1231

### DIFF
--- a/framework/util/platform.h
+++ b/framework/util/platform.h
@@ -53,6 +53,10 @@
 #include <sys/mman.h>
 #endif // WIN32
 
+#if defined(__ANDROID__)
+#include <sys/system_properties.h>
+#endif
+
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(util)
 GFXRECON_BEGIN_NAMESPACE(platform)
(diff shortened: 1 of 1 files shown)


bradgrantham-lunarg (approved):

framework/util/platform.h
```diff
@@ -288,28 +293,49 @@ inline void* GetProcAddress(LibraryHandle handle, const char* name)
     return dlsym(handle, name);
 }
 
+#if defined(__ANDROID__)
+static const prop_info* GetAndroidPropertyInfo(const char* name)
+{
+    // According to system_properties.h "Property lookup is expensive, so it can be useful to cache the result of this
+    // function.". For that reason we cache property infos.
```
dfriederich, line 300: I saw that comment too, how much overhead is it? Was not sure it is worth the added complexity.

framework/util/platform.h
```diff
@@ -288,28 +293,49 @@ inline void* GetProcAddress(LibraryHandle handle, const char* name)
...
+    static std::unordered_map<std::string, const prop_info*> property_name_info_cache;
+
+    const std::string prop_name = std::string(name);
+    const prop_info*  pi;
+
+    auto prop_name_info = property_name_info_cache.find(prop_name);
```
dfriederich, line 306: is this a problem in respect to multi-threading, or is this code never used concurrently?

framework/util/platform.h
```diff
@@ -288,28 +293,49 @@ inline void* GetProcAddress(LibraryHandle handle, const char* name)
...
+
 inline std::string GetEnv(const char* name)
 {
     std::string env_value;
 
 #if defined(__ANDROID__)
-    std::string command = "getprop ";
-    command += name;
+    // Note: GetAndroidPropertyInfo caches property info look ups. That means if a property is not set
+    // during start up, setting it or changing its value during run time will not be noticed and won't have any effect.
```
dfriederich, line 327: that's fine with our usage. But not sure other users wont run into this. For example we did have an issue where the property was not set in the initial run, and that caused issues of differently changing the output name. For that reason we now set the property before we start the app, so no problem for us. But this is a bit of a trap I could see others running into.

bradgrantham-lunarg: We discussed this at length internally and this comment was the result of the conversation.  Otherwise if we want to pick up at frame time the values of properties that were unset at startup we'd have the impact of `__system_property_find` every frame, at least until it wasn't `nullptr`.  I also don't know (as you asked above) how much performance impact there is here.  I gathered from Panos that he was just honoring the guidance in system_properties.h.

panos-lunarg: Yeah I didn't measure performance impact, I just followed the hint. I'll do some measurement to get a rough idea

panos-lunarg: Did some measurement and I didn't see any difference. I'll drop the cache for simplicity's sake

bradgrantham-lunarg (approved):

Issue #1189, opened 2023-07-08
[capture] Thread protect for create wrappers and remove wrappers


locke-lunarg:
`VulkanStateTableBase::mutex_` can't protect `InsertEntry` and `RemoveEntry`.  The thread protect need more wide. Because DestroyBuffer and RemoveWrapper are different functions. CreateBuffer and CreateWrappedHandle could happen between DestoryBuffer and RemoveWrapper.

> Process
> 1. GetDeviceTable(device)->DestroyBuffer(device, buffer, pAllocator)
> 2. CreateBuffer -> Another thread. Create a duplicated handle value.
> 3. CreateWrappedHandle -> The new wrapper can't be added to the map because the old one hadn't removed from the map. It prints a warning, like `Create a duplicated Handle: 2230487727664. This wrapper can't be written into VulkanStateHandleTable.`
> 4. DestroyWrappedHandle<BufferWrapper>(buffer)-> Now, the map doesn't have the new wrapper. 
>It could get a warning, like `GetWrappedId() couldn't find Handle: 2230487727664's wrapper. It might have been destroyed` in the following process.

The thread protect should cover between `DestroyBuffer` and `RemoveWrapper`, not only `RemoveEntry`.

Here is a sample code:
Using vkcube to run CreateBuffer and DestoryBuffer 10k times on different threads.
https://github.com/locke-lunarg/Vulkan-Tools/tree/thread

Thread protect on CreateBuffer and DestroyBuffer. We won't see the warning and crashes.
https://github.com/locke-lunarg/gfxreconstruct/tree/test-thread

locke-lunarg:
We could use `GFXRECON_FORCE_COMMAND_SERIALIZATION` to fix it before we fix this issue.

PR #1235, opened 2023-08-23
Fix multi-threading issue with using real Vulkan handles

mizhen:
**The problem**
   
With default GFXR setting, sometimes an app will destroy some Vulkan handles in one thread (A) and create same type of Vulkan handle in another thread (B). There is a gap of time in between when the real handle is destroyed, and when its wrappers were deleted from map in thread A. If during this time period, thread B was able to run, and creates same type of handles, and if any of the newly-created handles had the same value of those destroyed by thread A, capturing will crash because handle rapper handling error for the new created handle.
 
**The solution**

The fix fixed the issue by acquiring shared lock for all wrapped handle creation and acquiring exclusive lock for range from destroying real Vulkan handle to end of deleting related wrapper.

**Result**
Tested target title full trace and trim trace capture/playback with the build of the pull request, all tests passed, the commit fixed the issue.

### DIFF
--- a/android/framework/encode/CMakeLists.txt
+++ b/android/framework/encode/CMakeLists.txt
@@ -23,6 +23,8 @@ target_sources(gfxrecon_encode
                    ${GFXRECON_SOURCE_DIR}/framework/encode/vulkan_capture_manager.h
                    ${GFXRECON_SOURCE_DIR}/framework/encode/vulkan_capture_manager.cpp
                    ${GFXRECON_SOURCE_DIR}/framework/encode/vulkan_handle_wrappers.h
+                   ${GFXRECON_SOURCE_DIR}/framework/encode/vulkan_scoped_destroy_lock.h
+                   ${GFXRECON_SOURCE_DIR}/framework/encode/vulkan_scoped_destroy_lock.cpp
                    ${GFXRECON_SOURCE_DIR}/framework/encode/vulkan_handle_wrapper_util.h
                    ${GFXRECON_SOURCE_DIR}/framework/encode/vulkan_handle_wrapper_util.cpp
                    ${GFXRECON_SOURCE_DIR}/framework/encode/vulkan_state_info.h
--- a/framework/encode/CMakeLists.txt
+++ b/framework/encode/CMakeLists.txt
@@ -90,6 +90,8 @@ target_sources(gfxrecon_encode
                     ${CMAKE_CURRENT_LIST_DIR}/vulkan_capture_manager.h
                     ${CMAKE_CURRENT_LIST_DIR}/vulkan_capture_manager.cpp
                     ${CMAKE_CURRENT_LIST_DIR}/vulkan_handle_wrappers.h
+                    ${CMAKE_CURRENT_LIST_DIR}/vulkan_scoped_destroy_lock.h
+                    ${CMAKE_CURRENT_LIST_DIR}/vulkan_scoped_destroy_lock.cpp
                     ${CMAKE_CURRENT_LIST_DIR}/vulkan_handle_wrapper_util.h
                     ${CMAKE_CURRENT_LIST_DIR}/vulkan_handle_wrapper_util.cpp
                     ${CMAKE_CURRENT_LIST_DIR}/vulkan_state_info.h
--- a/framework/encode/vulkan_handle_wrapper_util.cpp
+++ b/framework/encode/vulkan_handle_wrapper_util.cpp
@@ -30,14 +30,9 @@ VulkanStateHandleTable state_handle_table_;
 
 std::shared_mutex mutex_for_create_destroy_handle_;
 
-void LockForDestroyHandle()
+std::unique_lock<std::shared_mutex> ScopedDestroyLock()
 {
-    mutex_for_create_destroy_handle_.lock();
-}
-
-void UnlockForDestroyHandle()
-{
-    mutex_for_create_destroy_handle_.unlock();
+    return std::move(std::unique_lock<std::shared_mutex>(mutex_for_create_destroy_handle_));
 }
 
 uint64_t GetWrappedId(uint64_t object, VkObjectType object_type)
--- a/framework/encode/vulkan_handle_wrapper_util.cpp
+++ b/framework/encode/vulkan_handle_wrapper_util.cpp
@@ -1,6 +1,5 @@
 /*
 ** Copyright (c) 2020 LunarG, Inc.
-** Copyright (c) 2023 Advanced Micro Devices, Inc. All rights reserved.
 **
 ** Permission is hereby granted, free of charge, to any person obtaining a
 ** copy of this software and associated documentation files (the "Software"),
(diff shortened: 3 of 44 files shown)


bradgrantham-lunarg:
Please keep PRs as separated as possible.  Could you take a minute to remove the copyright updates that aren't also in files adding this bug fix into a separate PR?

panos-lunarg:
Good catch noticing the race condition.

`CaptureManager` is already defining a `shared_mutex` with very similar functionality used to apply locking upon each api call entry and differentiate between shared and unique access.
My suggestion is to investigate if you declare there a new `shared_mutex` for the purpose of resource creation/destruction and similarly to the existing `AcquireSharedApiCallLock()` / `AcquireExclusiveApiCallLock()` define two functions to create and return unique/shared locks using the new mutex. This way you should not need to create two separate source files to isolate the new mutex.

Since you will need to use the new mutex in `framework/encode/vulkan_handle_wrapper_util.h` I am suspecting that you will face compilation errors due to cyclic dependencies of the headers. See if you can work around those and if not _then_ resolve into creating the new source files. If that's the case then use the `ScopedDestroyLock` in both places that use the `shared_mutex`

mizhen:
> Good catch noticing the race condition.
> 
> `CaptureManager` is already defining a `shared_mutex` with very similar functionality used to apply locking upon each api call entry and differentiate between shared and unique access. My suggestion is to investigate if you can declare there a new `shared_mutex` for the purpose of resource creation/destruction and similarly to the existing `AcquireSharedApiCallLock()` / `AcquireExclusiveApiCallLock()` define two functions to create and return unique/shared locks using the new mutex. This way you should not need to create two separate source files to isolate the new mutex.
> 
> Since you will need to use the new mutex in `framework/encode/vulkan_handle_wrapper_util.h` I am suspecting that you will face compilation errors due to cyclic dependencies of the headers. See if you can work around those and if not _then_ resolve into creating the new source files. If that's the case then use the `ScopedDestroyLock` in both places that use the `shared_mutex`

I'll try it out, will keep you updated. Thanks.

mizhen:
> > Good catch noticing the race condition.
> > `CaptureManager` is already defining a `shared_mutex` with very similar functionality used to apply locking upon each api call entry and differentiate between shared and unique access. My suggestion is to investigate if you can declare there a new `shared_mutex` for the purpose of resource creation/destruction and similarly to the existing `AcquireSharedApiCallLock()` / `AcquireExclusiveApiCallLock()` define two functions to create and return unique/shared locks using the new mutex. This way you should not need to create two separate source files to isolate the new mutex.
> > Since you will need to use the new mutex in `framework/encode/vulkan_handle_wrapper_util.h` I am suspecting that you will face compilation errors due to cyclic dependencies of the headers. See if you can work around those and if not _then_ resolve into creating the new source files. If that's the case then use the `ScopedDestroyLock` in both places that use the `shared_mutex`
> 
> I'll try it out, will keep you updated. Thanks.

Yes, I did meet the compilation errors due to cyclic dependencies with the change of moving class ScopedDestroyLock to capture_manager.h ( https://github.com/LunarG/gfxreconstruct/commit/1357ec061df94cf0d7b836c03b5d19dab829d8bb ). Unfortunately, I didn't find a workaround.

locke-lunarg:
Thank you @mizhen  to fix it. 
I have a issue about it. https://github.com/LunarG/gfxreconstruct/issues/1189
I wrote a test that create and destroy Buffer 10k times by diff threads.
https://github.com/locke-lunarg/Vulkan-Tools/tree/thread
This branch works for the test code.

mizhen:
> Thank you @mizhen for fixing it.
> 
> However, I think it's not enough if the Create protection is only in `CreateWrappedDispatchHandle`. We can look at an example in `VulkanCaptureManager::OverrideCreateBuffer`. After `CreateWrappedHandle`, the mutex is unlocked, and the buffer is available to be destroyed. But the BufferWrapper is still used after `CreateWrappedHandle`. Also, `CustomEncoderPostCall` could use the Wrapper. In this case, the Create protection has to be wider between `CreateWrappedDispatchHandle` and `CustomEncoderPostCall`. It's easier to lock it between `OverrideCreateBuffer` (or Vulkan function call) and `CustomEncoderPostCall`.
> 
> I have a small sample for the protection. [locke-lunarg@bf0a776](https://github.com/locke-lunarg/gfxreconstruct/commit/bf0a7761329c3580bd0d425c2d892d65a7c941a2) I posted an issue about it. #1189
> 
> Further, maybe we could give an individual mutex to every kind of handle, so `CreateBuffer` doesn't lock `CreateImage`. Or even further, one `CreateBuffer` should not lock another `CreateBuffer`, or `CreateBuffer` lock only its `DestroyBuffer`. But we could leave it in the future.

Thank you for your review and providing the sample.

#1189 is the issue show up in the target title, I think the key of the issue is because driver return previously deleted handle (driver reuse destroyed resource handle). The issue only happens when driver return previously deletes handle and before GFXR delete the wrapper of that handle. So the pull request makes deleting Vulkan object and its related wrapper become an atomic operation for wrapper creation. 

For the case in the comment which is related to VulkanCaptureManager::OverrideCreateBuffer, the behaviour of the target game and GFXR is as the following:
'''
[Target game]:
                              vkCreateBuffer (it's similiar for any other call handling)
                                                   
[GFXR capture the call]:             
                                  .......
                              CustomEncoderPreCall<format::ApiCallId::ApiCall_vkCreateBuffer>
                                     
                                  .......
                              VulkanCaptureManager::Get()->OverrideCreateBuffer
                                     
                                  .......
                              CustomEncoderPostCall<format::ApiCallId::ApiCall_vkCreateBuffer>
                                     
                                  .......
                              return result
                                     
[Target game]:                       
                              Target game get the returned buffer handle and use it (to access, delete or do something else)
'''

Therefore before the control return to target game, it cannot use the returned buffer handle or delete the buffer. In another word, if other thread of the game gets the control and delete a buffer, that will be some other buffer. For deleting other buffer during the creation handling of this buffer, it doesn't cause the multi-threading issue in the pull request.

During debugging/testing the pull request, also compared the performance with different lock scope, it shows different lock scope heavily affect the loading process in the running of target title, so the pull request tries to select a minimum scope.

locke-lunarg (approved):
Thank you. You are right. It's impossible to happen that the buffer gets destroyed during creating. The PR looks goo to me. Please rebase it and it's ready to merge. 

close: https://github.com/LunarG/gfxreconstruct/issues/1189

mizhen:
> Yes, you are right. It's impossible to happen that the buffer gets destroyed during creating. The PR looks good to me. Please rebase it, and it's ready to merge.
> 
> close: #1189

Ok, will do, thanks.

mizhen:
> > Yes, you are right. It's impossible to happen that the buffer gets destroyed during creating. The PR looks good to me. Please rebase it, and it's ready to merge.
> > close: #1189
> 
> Ok, will do, thanks.

Done.

lunarpapillo:
FYI, your LunarG CI build #3656 seems to have failed due to an unrelated GPU crash on one machine (passed on all others).  I've restarted the run.

PR #1239, opened 2023-08-29
gfxrecon-replay: Use infinite timeout for 3 Vulkan api calls

davidlunarg:
When replaying:
    vkAcquireProfilingLockKHR
    vkWaitForPresentKHR
    vkWaitSemaphores
use an infinite timeout if the expected result is VK_SUCCESS.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -41,8 +41,10 @@
 
 #include "generated/generated_vulkan_enum_to_string.h"
 
+#include <algorithm>
 #include <cstdint>
 #include <limits>
+#include <numeric>
 #include <unordered_set>
 #include <future>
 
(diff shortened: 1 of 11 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -6641,6 +6641,78 @@ VkResult VulkanReplayConsumerBase::OverrideGetPhysicalDeviceToolProperties(
...
+    VkResult                   result;
+
+    // If expected result is VK_SUCCESS, ensure that vkWaitSemaphores waits until semaphores
+    // are available by using a timeout of UINT64_MAX.
+    // If expected result is VK_TIMEOUT, try to get a timeout by using a timeout of 0.
+    // If expected result is anything else, use the passed in timeout value.
+    if (original_result == VK_SUCCESS)
+        timeout = std::numeric_limits<uint64_t>::max();
+    else if (original_result == VK_TIMEOUT)
+        timeout = 0;
```
MarkY-LunarG, line 6663: Does this guarantee timeout is returned?  Or should we also return TIMEOUT?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -6641,6 +6641,78 @@ VkResult VulkanReplayConsumerBase::OverrideGetPhysicalDeviceToolProperties(
...
+    assert((device_info != nullptr) && (pInfo != nullptr) && !pInfo->IsNull() && (pInfo->GetPointer() != nullptr));
+    VkDevice                   device    = device_info->handle;
+    const VkSemaphoreWaitInfo* wait_info = pInfo->GetPointer();
+    VkResult                   result;
+
+    // If expected result is VK_SUCCESS, ensure that vkWaitSemaphores waits until semaphores
+    // are available by using a timeout of UINT64_MAX.
+    // If expected result is VK_TIMEOUT, try to get a timeout by using a timeout of 0.
+    // If expected result is anything else, use the passed in timeout value.
+    if (original_result == VK_SUCCESS)
```
MarkY-LunarG, line 6660: NIT: All of these need braces.  I thought our coding standard required braces for `if` statements.  But I could be wrong.

framework/decode/vulkan_replay_consumer_base.cpp
davidlunarg: No, this doesn't guarantee a VK_TIMEOUT is returned. This is the case in which the capture returned VK_TIMEOUT, but for whatever reason, the replay returns a different value (maybe the timing in the replay is different so that a timeout doesn't occur, or the device is now in a bad state and the replay returns VK_ERROR_DEVICE_LOST). The calling code will print out a warning that the return value didn't match the capture. I'm not sure that we can replay correctly given that a timeout occurred during capture. This behavior is the same as in the replay of vkWaitForFence. I propose we leave the code as submitted in this PR.

MarkY-LunarG (approved):
LGTM

PR #1240, opened 2023-08-29
Modify virtual swapchain behavior

MarkY-LunarG:
Modify virtual swapchain to create a command buffer per queue and and per swapchain image.  Then when a vkQueuePresent occurs, perform the blits in the same queue as the presented queue to reduce race conditions.
This resolves the issue where a QueuePresent is presented inline in a queue, but has no wait semaphore information.  Instead, the QueuePresent relied upon the previous submit (also in the same queue) to perform the synchronization.

Previously, the virtual swapchain had it's own Queue that it used which caused a race condition resulting in old images being used.

Fixes: #1122

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,3 @@ __pycache__/
 .vscode/
 *.db
 *.pyc
-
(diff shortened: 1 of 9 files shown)


MarkY-LunarG:
This is the code original defined in PR #1222 , but that was prematurely checked in before being squashed.  Brad has reverted that change and this review should serve as the placeholder for that.

MarkY-LunarG:
And it's already one commit, just to be safe.

bradgrantham-lunarg (approved):

MarkY-LunarG:
Ok, now it's in properly!

PR #1241, opened 2023-08-29
cast a float result back to uint32_t

bradgrantham-lunarg:


panos-lunarg (approved):

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2147,12 +2147,12 @@ void VulkanReplayConsumerBase::WriteScreenshots(const Decoded_VkPresentInfoKHR*
                 // If both copy_scale and copy_width are provided, use copy_scale.
                 const uint32_t screenshot_width =
                     options_.screenshot_scale
-                        ? (options_.screenshot_scale * swapchain_info->width)
+                        ? static_cast<uint32_t>(options_.screenshot_scale * swapchain_info->width)
```
andrew-lunarg, line 2150:
```cpp
constexpr uint32_t truncate_to_uint32(const float f)
{
    GFXRECON_ASSERT_DEBUG(f >= 0.0f);
    return static_cast<uint32_t>(f);
}
```
somewhere in a `util` header to name what we are doing here and add the sanity precondition check on the param?

bradgrantham-lunarg: I read `static_cast<uint32>` on `float` as truncating.

This PR is just to fix warnings in /W3 build in Visual C++

It does seem like a good idea to check the range of the scale but that should probably be validated in input.   Want to add that in a PR?

PR #1243, opened 2023-08-30
Stricter compiler flags

panos-lunarg:


CMakeLists.txt
```diff
@@ -44,6 +44,8 @@ set(GFXRECONSTRUCT_PROJECT_VERSION_PATCH 0)
 
 set(GFXRECON_PROJECT_VERSION_SHA1 "unknown-build-source")
 
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pedantic -Wall -Wconversion -Wextra -Werror -Wno-unused-parameter -Wno-missing-field-initializers -Wno-sign-conversion")
```
andrew-lunarg, line 47: I think we want to be strict for our own internal builds but permissive for user builds so that package maintainers and others on diverse systems and compiler versions have an easy time but we make sure the code is in good shape.

PR #1244, opened 2023-09-01
gfxrecon_replay: Define VMA_ASSERT to avoid assert crashes

davidlunarg:
The debug compiled version of gfxrecon_replay would crash with an assert failure in VMA because the captured app did not destroy all images or buffers that it created before calling vkDestroyDevice. This fix changes the assert failure to a warning.

### DIFF
--- a/USAGE_desktop_Vulkan.md
+++ b/USAGE_desktop_Vulkan.md
@@ -509,6 +509,10 @@ During replay, swapchain indices for present can be different from captured indi
 
 Virtual Swapchain insulates higher layers in the Vulkan stack from these problems by creating a set of images, exactly matching the swapchain configuration at capture time, which it exposes for them to render into.  Before a present, it copies the virtual image to a target swapchain image for display. Since this issue can happen in many situations, virtual swapchain is the default setup. If the user wants to bypass the feature and use the captured indices to present directly on the swapchain of the replay implementation, they should add the `--use-captured-swapchain-indices` option when invoking `gfxrecon-replay`.
 
+### Debug mode VMA errors
+
+gfxrec-replay with the -m rebind option uses the Vulkan Memory Allocator library for memory allocations. If gfxrecon-replay is compiled debuggable, VMA_ASSERT errors in VMA can be trapped for debugging by setting GFXRECON_LOG_BREAK_ON_ERROR to true.
+
 ## Other Capture File Processing Tools
 
 ### Capture File Info
--- a/framework/decode/vulkan_rebind_allocator.cpp
+++ b/framework/decode/vulkan_rebind_allocator.cpp
@@ -29,19 +29,19 @@
 #ifdef NDEBUG
 #define VMA_ASSERT(expr_)
 #else
-#define VMA_ASSERT(expr_)                                                               \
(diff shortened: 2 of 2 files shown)


MarkY-LunarG (approved):
The changes look fine to me, but you've tested both scenarios with and without a string message in the assert and verified it worked?

davidlunarg:
>  you've tested both scenarios with and without a string message in the assert and verified it worked?

Yes, I have.

bradgrantham-lunarg (changes requested):
I think change to LOG_ERROR so that a user could enable GFXRECON_BREAK_ON_ERROR and kick into the debugger, and then add a note to USAGE*md that in debug mode a user may see a VMA assert and they can kick into the debugger with BREAK_ON_ERROR and then this is good to go.

davidlunarg:
I'm wondering if calling GFXRECON_LOG_ERROR is the right thing to do. I don't think the situation being identified is necessarily an error - the app has called vkDestroyDevice without first destroying all the images/buffers it created. That's not a fatal error, just a potential resource leak. And it is not a resource leak if the next thing the app does is exit, which would be the most common case. That's why I coded it using GRXRECON_LOG_WARNING.

MarkY-LunarG:
I would agree with @bradgrantham-lunarg  and say it is.  If an application triggers on an Assert, that indicates something is not right and a warning is not strong enough of a message.

bradgrantham-lunarg:
> I'm wondering if calling GFXRECON_LOG_ERROR is the right thing to do. I don't think the situation being identified is necessarily an error - the app has called vkDestroyDevice without first destroying all the images/buffers it created. That's not a fatal error, just a potential resource leak. And it is not a resource leak if the next thing the app does is exit, which would be the most common case. That's why I coded it using GRXRECON_LOG_WARNING.

Yes, that's interesting.

It *is* invalid usage to destroy a device without destroying all its child objects.

However, usually we LOG_ERROR on an unexpected internal situation, like couldn't open a file, or we can't find a Wrapper for an object.  We generally don't report on the application's behavior (with a few exceptions).  So we're just in a limbo where GFXR replay tries not to editorialize on the captured functions unless it would cause internal errors, and we've left that kind of error checking to VVL.

But VMA has the opposite principle - they by default assert and crash in this circumstance, probably assuming it's running under the real app and the user wants this behavior in DEBUG mode.  So at 30,000 feet view we're kind of not using VMA the way it should be used? :shrug:

If we could make LOG_WARNING kick into the debugger or otherwise allow a user debugging to break, that would be ideal. LOG_ERROR in the code and BREAK_ON_ERROR env variable seem expedient and a reasonable compromise but if you have an idea @davidlunarg I want to hear it.

P.S.  https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkDestroyDevice-device-00378, vkDeviceMemory is a child object as far as I read the spec - _An exception to this is when there is a parent/child relationship between objects. In this case, the application must not destroy a parent object before its children_, https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#fundamentals-objectmodel-lifetime

PR #1246, opened 2023-09-04
Added a check whether the ANDROID_SERIAL was set when mutiple devices are present

naomipappe:
This change is needed as when more than one device is connected, the script crashes if the selection is not explicitly specified. This change fixes it by adding a possibility to specify the device.


bradgrantham-lunarg:
@naomipappe Thanks for your PR submission!  Does `ANDROID_DEVICE` environment variable not work in your environment?

android/scripts/gfxrecon.py
```diff
@@ -187,22 +214,30 @@ def Replay(replay_args):
...
             cmd = ' '.join([adb_push, args.push_file, args.file])
+            if selection is not None:
+                cmd = InsertDeviceSelectionArgument(cmd, selection)
             print('Executing:', cmd)
             subprocess.check_call(shlex.split(cmd, posix='win' not in sys.platform))
 
         print('Executing:', adb_stop)
         subprocess.check_call(shlex.split(adb_stop, posix='win' not in sys.platform))
-
+
```
andrew-lunarg, line 229: Stray whitespace.

naomipappe: Removed the stray whitespace

PR #1248, opened 2023-09-07
Set CPU&GPU memory usage by command line option

JerryAMD:
**Problem:**
The GFXR trapped system error "bad allocations" on a VM system. But the system reported it had enough memory room for GFXR to replay.

**Solution:**
Add a replay option to set memory usage to avoid system memory issue.

### DIFF
--- a/framework/graphics/dx12_util.cpp
+++ b/framework/graphics/dx12_util.cpp
@@ -978,7 +978,7 @@ bool IsUma(ID3D12Device* device)
     return isUma;
 }
 
-uint64_t GetAvailableGpuAdapterMemory(IDXGIAdapter3* adapter, const bool is_uma)
+uint64_t GetAvailableGpuAdapterMemory(IDXGIAdapter3* adapter, double memory_usage, const bool is_uma)
 {
     uint64_t available_mem = 0;
 
@@ -992,9 +992,10 @@ uint64_t GetAvailableGpuAdapterMemory(IDXGIAdapter3* adapter, const bool is_uma)
         }
         if (SUCCEEDED(adapter->QueryVideoMemoryInfo(0, memory_segment, &video_memory_info)))
         {
-            if (video_memory_info.Budget > video_memory_info.CurrentUsage)
+            uint64_t total_memory = video_memory_info.Budget * memory_usage;
+            if (total_memory > video_memory_info.CurrentUsage)
             {
-                available_mem = video_memory_info.Budget - video_memory_info.CurrentUsage;
+                available_mem = total_memory - video_memory_info.CurrentUsage;
             }
             else
             {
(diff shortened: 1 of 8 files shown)


framework/graphics/dx12_util.cpp
```diff
@@ -1023,7 +1024,6 @@ uint64_t GetAvailableCpuMemory(double max_usage)
         GFXRECON_LOG_ERROR("Failed to get available virtual memory");
     }
 
-    // Only limit by available physical memory if max_usage <= 1.0.
     uint64_t avail_phys = std::numeric_limits<uint64_t>::max();
```
davidd-lunarg, line 1027: This comment `// Only limit by available physical memory if max_usage <= 1.0.` was removed but it appears to still be true.

framework/graphics/dx12_util.cpp
```diff
@@ -992,9 +992,10 @@ uint64_t GetAvailableGpuAdapterMemory(IDXGIAdapter3* adapter, const bool is_uma)
         }
         if (SUCCEEDED(adapter->QueryVideoMemoryInfo(0, memory_segment, &video_memory_info)))
         {
-            if (video_memory_info.Budget > video_memory_info.CurrentUsage)
+            uint64_t total_memory = video_memory_info.Budget * memory_usage;
```
davidd-lunarg, line 995: The `memory_usage` argument is not clamped so `total_memory` could be > `video_memory_info.Budget`. Is it allowed to for an application to use more than `video_memory_info.Budget`? If `memory_usage < 0`, what should `total_memory` be? This should also use a static_cast from double to uint64_t.

tools/tool_settings.h
```diff
@@ -880,6 +881,12 @@ static gfxrecon::decode::DxReplayOptions GetDxReplayOptions(const gfxrecon::util
         replay_options.override_object_names = true;
     }
 
+    const std::string& memory_usage = arg_parser.GetArgumentValue(kBatchingMemoryUsageArgument);
+    if (!memory_usage.empty())
+    {
+        replay_options.memory_usage = std::stoi(memory_usage);
```
davidd-lunarg, line 887: Should this value be clamped or ignored if it is out the range 0-100?

davidd-lunarg (changes requested):

framework/graphics/dx12_util.cpp
```diff
@@ -1030,27 +1032,33 @@ uint64_t GetAvailableCpuMemory(double max_usage)
...
 }
 
-bool IsMemoryAvailable(uint64_t required_memory, IDXGIAdapter3* adapter, double max_cpu_mem_usage, const bool is_uma)
+bool IsMemoryAvailable(uint64_t required_memory, IDXGIAdapter3* adapter, double max_mem_usage, const bool is_uma)
 {
     bool available = false;
 #ifdef _WIN64
     // For 32bit, only upload one buffer at one time, to save memory usage.
-    if (adapter != nullptr)
+    if (max_mem_usage > 0.0 && max_mem_usage <= 1.0)
```
davidd-lunarg, line 1044: If the user provides a value of 0, which is valid, this will log an error.

davidd-lunarg (changes requested):

davidd-lunarg (approved):

Issue #1169, opened 2023-06-28
Add GFXR version info to capture files and binaries


rurra-amd:
Currently it's not possible to determine what version of GFXR was used to create a capture file.

We could add a new metadata command that injects this for both Vulkan and DX12. Some of this info is present in `project_version.h` and comes from the root `CMakeLists.txt`. We would also like to embed the time at which the capture DLLs were built.

We would want to add this info to select binaries which we build, since their `File Properties` is pretty empty right now.

bradgrantham-lunarg:
Oh, I see, that's the version output from `convert`.  🤪

bradgrantham-lunarg:
We do store an Annotation in captures with the build information and other information in framework/encode/capture_manager.cpp.

It looks like I don't see that in current `convert` output.  @andrew-lunarg may have an insight.

andrew-lunarg:
We have a scheme implemented for storing this sort of information in the existing annotation block type rather than adding a new block type. Annotations go back to Dustin and this use of them was implemented in response to the closed issue #831.

As Brad mentioned the `"header": {}` object in the json output is the header for the json itself, and it is generated before any information has been read from the binary capture.

The annotations should be early in the file but I don't think we have set a hard rule on their ordering. 
They will not be output for captures made before #869 was merged.

The concept we are using is that each tool which touches a capture file (capture, optimise, compress, in-house tool, ...) should drop a dated* annotation at the top of the file so the file preserves a record of all operations performed on it. Since we use json, any structured data can be packed in there.

We have an issue for expanding the current scheme with more fields: https://github.com/LunarG/gfxreconstruct/issues/868 
We have a live issue for trimmed captures losing the tooling annotations: https://github.com/LunarG/gfxreconstruct/issues/1076

Info has support for the annotations as well as Convert. If they are going missing from convert for non-trimmed captures made after #869, that is a bug that I can look into. Info:
```
Annotation info:
        Total annotations: 1
        Operation annotations: 1
    
        {
            "tool": "capture",
            "gfxrecon-version": "0.9.15-dev",
            "vulkan-version": "1.3.231"
        }   
```
\* In universal time.

andrew-lunarg:
I have had to turn the operation output from Info back on in #1177.

PR #1250, opened 2023-09-11
Add GFXR version to binary file Properties

rurra-amd:
**Problem**
Inspection of file Properties of GFXR binaries comes back mostly empty. This can make it more difficult to track which version of a binary is distributed.

**Solution**
Add a resource file (`.rc`) which all binaries (capture + tools) can refer to. The `.rc` file gets its versioning data from pre-existing `project_version.h`. The build process will refer to the `.rc` file and populate the Details pane of binary file Properties.

This addresses https://github.com/LunarG/gfxreconstruct/issues/1169

**Testing**

Before:
image

After:
image

### DIFF
--- a/layer/CMakeLists.txt
+++ b/layer/CMakeLists.txt
@@ -52,7 +52,8 @@ target_sources(VkLayer_gfxreconstruct
                    ${CMAKE_CURRENT_LIST_DIR}/trace_layer.cpp
                    ${CMAKE_SOURCE_DIR}/framework/encode/custom_layer_func_table.h
                    ${CMAKE_SOURCE_DIR}/framework/generated/generated_layer_func_table.h
-                   $<$<BOOL:WIN32>:${CMAKE_CURRENT_LIST_DIR}/trace_layer.def>
+                   $<$<BOOL:WIN32>:${CMAKE_CURRENT_LIST_DIR}/trace_layer.def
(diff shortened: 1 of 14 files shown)


layer/CMakeLists.txt
```diff
@@ -52,7 +52,8 @@ target_sources(VkLayer_gfxreconstruct
                    ${CMAKE_CURRENT_LIST_DIR}/trace_layer.cpp
                    ${CMAKE_SOURCE_DIR}/framework/encode/custom_layer_func_table.h
                    ${CMAKE_SOURCE_DIR}/framework/generated/generated_layer_func_table.h
-                   $<$<BOOL:WIN32>:${CMAKE_CURRENT_LIST_DIR}/trace_layer.def>
+                   $<$<BOOL:WIN32>:${CMAKE_CURRENT_LIST_DIR}/trace_layer.def
+                   ${CMAKE_SOURCE_DIR}/version.rc>
```
davidd-lunarg, line 56: Does this work for other platforms? Should it be included on Windows only?
`$<$<BOOL:WIN32>:${CMAKE_SOURCE_DIR}/version.rc`

rurra-amd: This `.rc` file is a Windows thing so yeah I could add that bool check too.

rurra-amd: Done

davidd-lunarg (approved):

PR #1252, opened 2023-09-13
Enable info tool to display annotations

rurra-amd:
**Problem**
Annotations injected into capture file are not visible via `gfxrecon-info`. For Vulkan, this was because the annotation processor was never set. For D3D12, this was because it hasn't been implemented.

**Solution**
Unify how annotations are displayed for Vulkan and D3D12. Also be selective as to which ones to display for each API.

**Testing**
Running `gfxrecon-info` now displays this at the end:
image

### DIFF
--- a/framework/decode/annotation_handler.h
+++ b/framework/decode/annotation_handler.h
@@ -40,7 +40,7 @@ class AnnotationHandler
     virtual void ProcessAnnotation(uint64_t               block_index,
                                    format::AnnotationType type,
                                    const std::string&     label,
-                                   const std::string&     data){};
+                                   const std::string&     data) = 0;
 };
 
 GFXRECON_END_NAMESPACE(decode)
--- a/framework/decode/annotation_handler.h
+++ b/framework/decode/annotation_handler.h
@@ -40,7 +40,7 @@ class AnnotationHandler
     virtual void ProcessAnnotation(uint64_t               block_index,
                                    format::AnnotationType type,
                                    const std::string&     label,
-                                   const std::string&     data) = 0;
+                                   const std::string&     data) {};
 };
 
 GFXRECON_END_NAMESPACE(decode)
--- a/framework/decode/annotation_handler.h
+++ b/framework/decode/annotation_handler.h
@@ -40,7 +40,7 @@ class AnnotationHandler
     virtual void ProcessAnnotation(uint64_t               block_index,
                                    format::AnnotationType type,
                                    const std::string&     label,
-                                   const std::string&     data) {};
+                                   const std::string&     data){};
 };
 
 GFXRECON_END_NAMESPACE(decode)
--- a/framework/decode/dx12_stats_consumer.h
+++ b/framework/decode/dx12_stats_consumer.h
@@ -31,7 +31,7 @@
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
-class Dx12StatsConsumer : public Dx12Consumer, public gfxrecon::decode::AnnotationHandler
+class Dx12StatsConsumer : public Dx12Consumer
 {
   public:
     Dx12StatsConsumer() :
--- a/framework/decode/dx12_stats_consumer.h
+++ b/framework/decode/dx12_stats_consumer.h
@@ -393,23 +393,6 @@ class Dx12StatsConsumer : public Dx12Consumer, public gfxrecon::decode::Annotati
         }
     }
 
-    virtual void ProcessAnnotation(uint64_t                         block_index,
-                                   gfxrecon::format::AnnotationType type,
-                                   const std::string&               label,
-                                   const std::string&               data) override
-    {
-        ++annotation_count_;
-        if (type == gfxrecon::format::AnnotationType::kJson &&
-            label.compare(gfxrecon::format::kAnnotationLabelOperation) == 0)
-        {
-            operation_annotation_datas_.push_back(data);
-        }
-    }
-
-    uint64_t GetAnnotationCount() const { return annotation_count_; }
-
-    const std::vector<std::string>& GetOperationAnnotationDatas() const { return operation_annotation_datas_; }
-
   private:
     // Holds adapter descs that were obtained from the app calling GetDesc()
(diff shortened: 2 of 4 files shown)


tools/info/main.cpp
```diff
@@ -378,6 +442,8 @@ void GatherVulkanStats(const std::string& input_filename)
         vulkan_decoder.AddConsumer(&vulkan_stats_consumer);
         file_processor.AddDecoder(&vulkan_decoder);
 
+        file_processor.SetAnnotationProcessor(&vulkan_stats_consumer);
```
bradgrantham-lunarg, line 445: @andrew-lunarg , please merge #1177 ; that PR already contains this fix

andrew-lunarg: Done

tools/info/main.cpp
```diff
@@ -67,6 +69,12 @@ const char kOptions[] = "-h|--help,--version,--no-debug-popup,--exe-info-only,--
 
 const char kUnrecognizedFormatString[] = "<unrecognized-format>";
 
+struct AnnotationInfo
+{
+    std::string desc;
+    std::string data;
+};
+
```
andrew-lunarg, line 77: Could slot the dedicated AnnotationHandler in here. Doesnt need own h/cpp pair.
```cpp
class AnnotationRecorder : public gfxrecon::decode::AnnotationHandler {...}
```

andrew-lunarg:
I have a larger change to Info in mind that I welcome feedback on but I think this PR can be tweaked to come in before that and won’t make it any harder. https://github.com/LunarG/gfxreconstruct/issues/1256

tools/info/main.cpp
```diff
@@ -183,6 +191,68 @@ void GatherApiAgnosticStats(ApiAgnosticStats&                api_agnostic_stats,
...
+        std::vector<std::string> annotations;
+
+        if (annotation_count > 0)
+        {
+            if (operation_annotation_datas.size() > 0)
+            {
+                // Inspect annotations spotted in the capture file
+                for (const auto& operation : operation_annotation_datas)
+                {
+                    nlohmann::json json_obj = nlohmann::json::parse(operation);
```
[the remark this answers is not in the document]
rurra-amd: Done

rurra-amd:
> I think it makes more sense to make an AnnotationHandler subclass outside of the API Consumers rather than duplicating ProcessAnnotion() in the two consumers. It is my bad for putting it in the VulkanStatsConsumer in the first place: we had no Dx12 in mainline at that point and it wasn’t on my mind. Something like this (untested):
> 
> ```c++
> class AnnotationRecorder : public gfxrecon::decode::AnnotationHandler
> {
>   public:
> 
>     virtual void ProcessAnnotation(uint64_t                         block_index,
>                                    gfxrecon::format::AnnotationType type,
>                                    const std::string&               label,
>                                    const std::string&               data) override
>     {
>         ++annotation_count_;
>         if (type == gfxrecon::format::AnnotationType::kJson &&
>             label.compare(gfxrecon::format::kAnnotationLabelOperation) == 0)
>         {
>             operation_annotation_datas_.push_back(data);
>         }
>     }
> 
>     uint64_t GetAnnotationCount() const { return annotation_count_; }
> 
>     const std::vector<std::string>& GetOperationAnnotationDatas() const { return operation_annotation_datas_; }
> 
>   private:
>     std::vector<std::string> operation_annotation_datas_;
>     uint64_t                 annotation_count_{ 0 };
> };
> ```
> 
> We could use one of these `AnnotationRecorder` objects for both D3D and Vulkan passes.
> 
> Does this de-duplication effort make sense to you too? With that change and checking the JSON parse, in my mind we are good to go.

Yeah - duplicated code is never good.

framework/decode/dx12_stats_consumer.h
```diff
@@ -31,7 +31,7 @@
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
-class Dx12StatsConsumer : public Dx12Consumer
+class Dx12StatsConsumer : public Dx12Consumer, public gfxrecon::decode::AnnotationHandler
```
andrew-lunarg, line 34: We don't need this multiple inheritance anymore.

framework/decode/annotation_handler.h
```diff
@@ -40,7 +40,7 @@ class AnnotationHandler
     virtual void ProcessAnnotation(uint64_t               block_index,
                                    format::AnnotationType type,
                                    const std::string&     label,
-                                   const std::string&     data) = 0;
+                                   const std::string&     data){};
```
andrew-lunarg, line 43: I don't think we want a default NOP implementation. Any class deriving from `AnnotationHandler` should process annotations. I think the original version of the line was right.

andrew-lunarg (changes requested):
Looking very good. We are super-close. Just one more tweak I think!

framework/decode/dx12_stats_consumer.h
rurra-amd: Done

framework/decode/annotation_handler.h
rurra-amd: Done

andrew-lunarg:
Your branch had to be rebased on current dev and then the line 
`file_processor.SetAnnotationProcessor(&vulkan_stats_consumer);`
in Info's main.cpp deleted. I seem to have force-pushed that change to your repo at branch <https://github.com/rurra-amd/gfxreconstruct/tree/rurra-info-annotations>, rewriting the history there. My apologies: I didn't realise I could force push to someone else's fork!

Anyway, that should be the final tweak and I'll squash the change onto dev if the checks pass so no more work to do on your side.

andrew-lunarg (approved):

rurra-amd:
> Your branch had to be rebased on current dev and then the line `file_processor.SetAnnotationProcessor(&vulkan_stats_consumer);` in Info's main.cpp deleted. I seem to have force-pushed that change to your repo at branch https://github.com/rurra-amd/gfxreconstruct/tree/rurra-info-annotations, rewriting the history there. My apologies: I didn't realise I could force push to someone else's fork!
> 
> Anyway, that should be the final tweak and I'll squash the change onto dev if the checks pass so no more work to do on your side.

Thanks for the help Andrew!

PR #1255, opened 2023-09-14
Fix Android incremental builds

wpaul-samsung:
One issue is that the layer and the replay tool both rely on parts of the framework, and they both build the framework into the same output directory.  It also builds each ABI variant into the same output directory, so when doing a subsequent build, even if no inputs have changed, the outputs had been overwritten by other ABI builds, so it has to rebuild for each ABI.

### DIFF
--- a/android/layer/CMakeLists.txt
+++ b/android/layer/CMakeLists.txt
@@ -3,10 +3,10 @@ cmake_minimum_required(VERSION 3.4.1)
 get_filename_component(GFXRECON_SOURCE_DIR ../.. ABSOLUTE)
 
 include(../framework/cmake-config/PlatformConfig.cmake)
-add_subdirectory(../framework/util ${CMAKE_SOURCE_DIR}/../framework/util/build)
-add_subdirectory(../framework/graphics ${CMAKE_SOURCE_DIR}/../framework/graphics/build)
-add_subdirectory(../framework/format ${CMAKE_SOURCE_DIR}/../framework/format/build)
-add_subdirectory(../framework/encode ${CMAKE_SOURCE_DIR}/../framework/encode/build)
+add_subdirectory(../framework/util ${CMAKE_SOURCE_DIR}/../framework/util/build/layer/${ANDROID_ABI})
+add_subdirectory(../framework/graphics ${CMAKE_SOURCE_DIR}/../framework/graphics/build/layer/${ANDROID_ABI})
+add_subdirectory(../framework/format ${CMAKE_SOURCE_DIR}/../framework/format/build/layer/${ANDROID_ABI})
+add_subdirectory(../framework/encode ${CMAKE_SOURCE_DIR}/../framework/encode/build/layer/${ANDROID_ABI})
 
 add_library(VkLayer_gfxreconstruct SHARED "")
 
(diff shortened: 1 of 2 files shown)


wpaul-samsung:
Now that all objects are built into unique ABI directories, in theory each ABI could now be built in parallel rather than sequentially, which should reduce the clean build time by 4x.  I have been unable to figure out how to convince the Android Gradle plugin to execute the externalNativeBuild for each ABI in parallel though.

panos-lunarg:
Thanks! I'll take a look to see if I can make any improvements

panos-lunarg:
This patch indeed helps a lot with compilation times.
As you point out the next best improvement would be to figure out how to make the ABIs to compile in parallel. 
Unfortunately I couldn't find how to do that with the existing `externalNativeBuild` + `cmake` combination either.

On the other hand `ndk-build` _does_ build all the ABIs in parallel. I have uploaded [a branch on my repo](https://github.com/panos-lunarg/gfxreconstruct/tree/panos_ndk_build) with the all essential `Android.mk` files that builds the `libVkLayer_gfxreconstruct.so` and `libgfxrecon-replay.so`. 
I haven't figured out yet how to integrate this with `gradle` but I do see that `externalNativeBuild` apart from `cmake` has also a `ndkBuild` property/function so I believe this can be done.


On my desktop I get about a 2x faster compilation time with `ndk-build`. Of course this is just a rough estimate as I am comparing the time it takes the current `./gradlew` to generate the apk file with the the time it takes the `ndk-build` to build the 2 .so files for all ABIs. But I still believe is a good improvement

wpaul-samsung:
That's very encouraging.  I would think you hook it up to gradle simply by changing

```
    externalNativeBuild {
        cmake {
            path "CMakeLists.txt"
        }
    }
```

to 

```
    externalNativeBuild {
        ndkBuild{
            path "Android.mk"
        }
    }
```

wpaul-samsung:
@panos-lunarg Do I just do e.g. 'ndk-build -12' from the android directory?

panos-lunarg:
Yes, ndk-build -**j**12

But I force pushed on top of my branch more changes so this won't work any more.

I have resolved into calling `ndk-build` manually  through a gradle task. This way ABIs will be build in parallel. We also need `cmake` as it generates a .h file so it needs to run before `ndk-build`.

I still need to figure out how to include the `.so` files in the `.apk` as the ones generated by the task are not getting picked up.

So, with current patch, doing a `./gradlew assembleDebug` should compile all ABis in parallel but the final apk is not complete.

More detailed explanation:

There is no way to make `gradle` build ABI's in parallel. `ndk-build` understands the different ABIs and can build them in parallel but if invoked from `gradle` through `externalNativeBuild.ndkBuild` will again compile them one after the other so there is no profit from doing so.

As far as I can tell `cmake` does not understand ABIs and that it needs to compile multiple times so I resolved into having `gradle` call `ndk-build` through a task. Unfortunately ditching `cmake` completely is not that simple as we depend on it to generate `project_version.h` using the `cmake` configure_file` function.

So I have added a gradle task that invokes `ndk-build` as a final step when doing a debug or release build.

I still need to figure out how to properly pack the `.so` files in the `.apk` and do a few more clean ups in the gradle scipt (only does `assembleDebug` and also fix `gradlew clean` atm)

wpaul-samsung:
> Edit: Also this change is starting to get more complex than I was expecting. I am not sure if we really want all these changes given that your patch already gives a considerable speed up

I'll leave that up to you.  It would be nice to have faster clean builds as it would reduce PR validation times. etc.  But my bigger concern is local incremental builds working properly. which my change does do.  If you think it's going to take too much time to reduce clean build times. we can live without it.

PR #1259, opened 2023-09-19
Display call index and thread on failing calls

bartosz-muszarski-arm:
Adds call index and thread id to the error message displayed when the returned value does not match the original one

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -1124,7 +1124,10 @@ const VkAllocationCallbacks* VulkanReplayConsumerBase::GetAllocationCallbacks(
     return nullptr;
 }
 
-void VulkanReplayConsumerBase::CheckResult(const char* func_name, VkResult original, VkResult replay)
+void VulkanReplayConsumerBase::CheckResult(const char*                func_name,
+                                           VkResult                   original,
+                                           VkResult                   replay,
+                                           const decode::ApiCallInfo& call_info)
 {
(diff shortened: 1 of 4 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -1133,11 +1136,14 @@ void VulkanReplayConsumerBase::CheckResult(const char* func_name, VkResult origi
...
             // errors are not treated as fatal, but will be reported as warnings below, allowing the replay to attempt
             // to continue for the case where an application may have queried for formats that it did not use.
-            GFXRECON_LOG_FATAL("API call %s returned error value %s that does not match the result from the "
-                               "capture file: %s.  Replay cannot continue.",
-                               func_name,
-                               util::ToString<VkResult>(replay).c_str(),
-                               util::ToString<VkResult>(original).c_str());
+            GFXRECON_LOG_FATAL(
+                "API call at index: %d thread: %d %s returned error value %s that does not match the result from the "
+                "capture file: %s.  Replay cannot continue.",
```
andrew-lunarg, line 1141: Would you mind fixing the double space after the period here (to ` %s. Replay cannot`) since you are touching the line anyway?

andrew-lunarg (changes requested):
Looks like good code and a useful addition to me. Just that stray space that was written by someone else would be nice to have nuked.

andrew-lunarg (approved):

andrew-lunarg:
LGTM, Thanks!

PR #1261, opened 2023-09-19
gfxrecon-replay - handle wrap around of xcb events sequence numbers

davidlunarg:
The sequence number in the cookie returned by many xcb calls (i.e. xcb_map_window) is 32 bits. The sequence number in the event returned by xcb_wait_for_event is 16 bits. When comparing the two sequence numbers to see if the xcb call has completed, need to mask off the top 16 bits from the cookie sequence number.

### DIFF
--- a/framework/application/xcb_window.cpp
+++ b/framework/application/xcb_window.cpp
@@ -473,9 +473,12 @@ void XcbWindow::InitializeAtoms()
     bypass_compositor_atom_ = GetAtomReply(connection, kBypassCompositorName, bypass_compositor_atom_cookie);
 }
 
-void XcbWindow::CheckEventStatus(uint32_t sequence, uint32_t type)
+void XcbWindow::CheckEventStatus(uint16_t sequence, uint32_t type)
 {
-    if ((sequence >= pending_event_.sequence) && (type == pending_event_.type))
+    // Note that sequence is 16 bits and pending_event_.sequence is 32 bits, so we only compare the lower 16 bits.
+    // We're checking for equality, so checking for the arrival of certain events needs to be done in the same order
+    // as the xcb calls that cause those events to be generated.
+    if ((sequence == static_cast<uint16_t>(pending_event_.sequence & 0xffff)) && (type == pending_event_.type))
     {
         pending_event_.complete = true;
     }
--- a/framework/application/xcb_window.h
+++ b/framework/application/xcb_window.h
@@ -44,20 +44,20 @@ class XcbWindow : public decode::Window
 
     xcb_atom_t GetDeleteWindowAtom() const { return delete_window_atom_; }
 
-    void MapNotifyReceived(uint32_t sequence, bool visible)
+    void MapNotifyReceived(uint16_t seq16, bool visible)
     {
-        CheckEventStatus(sequence, XCB_MAP_NOTIFY);
+        CheckEventStatus(seq16, XCB_MAP_NOTIFY);
         visible_ = visible;
     }
 
-    void ResizeNotifyReceived(uint32_t sequence, uint32_t width, uint32_t height)
+    void ResizeNotifyReceived(uint16_t seq16, uint32_t width, uint32_t height)
     {
-        CheckEventStatus(sequence, XCB_CONFIGURE_NOTIFY);
+        CheckEventStatus(seq16, XCB_CONFIGURE_NOTIFY);
         width_  = width;
         height_ = height;
     }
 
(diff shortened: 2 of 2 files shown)


davidlunarg:
Still working on this change, but thought I would a PR to see how it does on CI.

davidlunarg:
This is no longer a draft PR. I would like to have it reviewed so I can merge it.

MarkY-LunarG (approved):
Looks good to me.  Good comment as well.

andrew-lunarg (approved):
Looks good to me.

Alternative 1: We don't really need to change the function signatures, just the comparison (but then we'd need to mask both side of the comparison for safety).

Alternative 2: We could make `EventInfo.sequence` `uint16_t`, which would do the throwing away of bits upon assignment to it and thereby avoid the need to mask.

Took me a moment digging in to xcb docs to grok the comment. Also, arg, why is sequence 32 bit _here_ in xcb and 16 bit _there_ in it???

davidlunarg:
> Looks good to me.
> 
> Alternative 1: We don't really need to change the function signatures, just the comparison (but then we'd need to mask both side of the comparison for safety).
> 
> Alternative 2: We could make `EventInfo.sequence` `uint16_t`, which would do the throwing away of bits upon assignment to it and thereby avoid the need to mask.
> 
> Took me a moment digging in to xcb docs to grok the comment. Also, arg, why is sequence 32 bit _here_ in xcb and 16 bit _there_ in it???

I changed the function signatures to accurately reflect what is being passed - the sequence number is 16 bits as it arrived from xcb, so we won't mislead someone reading the code that it is 32 bits.

Why does xcb use 32 bits in the sequence number in the cookie that is returned by xcb calls, but 16 bits in the notification event? I started out thinking that 16 bits is used in the notification event in an attempt to save memory, since lots of events are generated. But the events are problably short-lived and the memory re-used, so the amount of extra memory used would be minimal. I dunno.

andrew-lunarg:
> I changed the function signatures to accurately reflect what is being passed - the sequence number is 16 bits as it arrived from xcb, so we won't mislead someone reading the code that it is 32 bits.
> 
> Why does xcb use 32 bits in the sequence number in the cookie that is returned by xcb calls, but 16 bits in the notification event? I started out thinking that 16 bits is used in the notification event in an attempt to save memory, since lots of events are generated. But the events are problably short-lived and the memory re-used, so the amount of extra memory used would be minimal. I dunno.

Makes sense!

I asked a friend and this is what they said:

> Why is the sequence number sometimes 32 bit and sometimes 16 bit in the XCB API?

The sequence number in the XCB API can be either 32-bit or 16-bit depending on the context. The sequence number is used to uniquely identify each request or event sent between the X client and the X server.

In the XCB API, the sequence number is represented by the xcb_void_cookie_t type. This type is defined as a 32-bit value when the sequence number needs to be 32 bits, and as a 16-bit value when the sequence number can be represented by 16 bits.

The reason for having both 32-bit and 16-bit sequence numbers is to optimize memory usage and network bandwidth. In situations where the number of requests or events is expected to be small, a 16-bit sequence number is sufficient and reduces the overhead of larger data types. On the other hand, in scenarios where a large number of requests or events are expected, a 32-bit sequence number is used to accommodate the larger range of values.

Here are some additional details regarding the sequence numbers in the XCB API:

* In the XCB protocol specification, the sequence number is defined as a CARD16 (16-bit) value by default. This is the most common case and is used when the number of requests or events is expected to be within the range of a 16-bit value.
* In situations where the number of requests or events exceeds the range of a 16-bit value, the XCB protocol allows for the use of a 32-bit sequence number. This is achieved by extending the sequence number field to a CARD32 (32-bit) value, which provides a much larger range for sequence numbers.
* The decision to use a 32-bit or 16-bit sequence number is typically made by the X server implementation. It depends on factors such as the expected number of simultaneous clients, the expected rate of requests and events, and the available resources.
* When using the XCB API, developers don't need to manually handle the sequence numbers. The XCB library takes care of managing the sequence numbers internally and ensures that each request or event is assigned a unique sequence number.

In summary, the sequence number in the XCB API can be either 32-bit or 16-bit depending on the situation. The choice of using a 32-bit or 16-bit sequence number is determined by factors such as the expected number of requests or events and the available resources. The XCB library handles the management of sequence numbers internally, so developers don't need to worry about it in most cases.

https://www.phind.com/search?cache=h3ji83h983v3rgh0agfkmnbp

bradgrantham-lunarg (approved):

PR #1264, opened 2023-09-20
Added a build type to the version string.

naomipappe:
Currently, when querying a version of the `gfxreconstruct` tool, it is unclear what build type does the binary have. 

This change improves on that by adding a build type to the version string.


andrew-lunarg:
Good observation! Is this PR solving an issue that has affected you in your use of GFXR / causes you an ongoing problem in your work or is it just something that struck you as wrong when you saw the version string output?

Would you mind pasting into the description what this outputs in the places `GFXRECON_PROJECT_VERSION_STRING` is used?

The only mild objection to adding build config information to the version string would be the concern that scripts in production in the wild might already be parsing this string and might not handle the change. Hopefully, teams using GFXR have a managed upgrade process that would catch that though.

naomipappe:
Hello Andrew!

The PR is providing a way for clients to be sure that they are using the release builds for performance measurements. That way, they can easily eliminate the possibility of using a debug build, when unsatisfactory performance is encountered. 

As for the output example, it would look like this:
```
gfxrecon-replay version info:
  GFXReconstruct Version 1.0.0-dev (branch:SHA1) Debug
  Vulkan Header Version 1.3.261
```
and analogously for release.

charles-lunarg (commented):
The content of the PR is good - being able to 'know' what build configuration was used helps figure out what gfxrecon build was used.

However, as implemented this doesn't work for multi-config systems like visual studio. That is because CMAKE_BUILD_TYPE is never set as the build system can different types without needing to reconfigure CMake.

I have a solution in mind that I will push up to remedy this. Only thing I need to verify is that make/ninja/android functions with the solution.

charles-lunarg (approved):
Looks like I've fixed all the small issues I encountered. Going ahead and merging.

PR #1265, opened 2023-09-20
Adding the capture options to annotation for display with info tool

naomipappe:
Several options are recommended when dealing with visual corruption on traces. It would be good to know, which non-default options were used when capturing a trace. 

This change captures this information into the operation annotation, if non-default parameters were used, and can be viewed with `gfxrecon-info` tool.

### DIFF
--- a/tools/info/main.cpp
+++ b/tools/info/main.cpp
@@ -220,19 +220,20 @@ void GatherApiAgnosticStats(ApiAgnosticStats&                api_agnostic_stats,
 
 std::string GetJsonValue(const nlohmann::json& json_obj, const std::string& key)
 {
-    std::string out = "";
-
-    if (json_obj.contains(key))
+    std::string out                    = "";
+    auto        search_result_iterator = json_obj.find(key);
+    if (search_result_iterator != json_obj.end())
     {
-        try
+        const nlohmann::json& value = *search_result_iterator;
+        if (value.is_object())
         {
-            out = json_obj.at(key).get<std::string>();
-        }
-        catch (const nlohmann::json::type_error& te)
-        {
-            out += "\n\t" + json_obj.at(key).dump(kDefaultIndent);
+            out += "\n\t" + value.dump(kDefaultIndent);
             out.pop_back();
             out += "\t}";
+        }
+        else
+        {
+            out = value;
         }
     }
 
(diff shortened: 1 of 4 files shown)


naomipappe:
This change relies on https://github.com/LunarG/gfxreconstruct/pull/1155 to be merged first.

lunarpapillo:
The LunarG CI failure appears to be due to an unrelated GPU crash.  I've restored the machine and will restart the test.

tools/info/main.cpp
```diff
@@ -222,7 +224,16 @@ std::string GetJsonValue(const nlohmann::json& json_obj, const std::string& key)
...
+        try
+        {
+            out = json_obj.at(key).get<std::string>();
+        }
+        catch (const nlohmann::json::type_error& te)
+        {
+            out += "\n\t" + json_obj.at(key).dump(kDefaultIndent);
+            out.pop_back();
+            out += "\t}";
+        }
```
charles-lunarg, line 236: The use of try-catch here feels awkward, as it should be possible to know if its a string or not ahead of time.

Should be possible to call `is_string()` before printing, and using the fallback if it isn't a string.

https://json.nlohmann.me/api/basic_json/is_string/

I say this because other uses of try-catch in the codebase are in main() functions, or are handling potentially invalid input (like stof), whereas here there is a way to "know" what to do without relying on an exception to do so.

charles-lunarg (commented):
Other than a weird usage of try-catch, the functionality of this PR is good.

tools/info/main.cpp
naomipappe: Oh yes, somehow I missed this method while I was reading the documentation. Your suggestion makes much more sense. Furthermore, thanks to your comment I revisited this piece of code, and made further adjustments.

charles-lunarg (approved):
Muuuch nicer to read and understand.

bradgrantham-lunarg:
I think it's probably best to remove any functionality from this PR that depends on #1155 and submit those dependencies as a separate PR - it sounds like @charles-lunarg has approved this and this can go in quickly.

naomipappe:
The dependency that it had on #1155 was the display of the fence delay option that PR added. I have removed that when I updated the PR recently.

Issue #1256, opened 2023-09-14
GFXRecon-Info Should Take One Pass Over File, Not 3 and Growing


andrew-lunarg:
Info currently takes up to three passes over the file to generate stats.
Our current capture files are all single-API other than the D3D12+AGS combo\*, limiting that to two passes, but we foresee use-cases for mixed API capture files.
This code from Info's main shows the structure.
```cpp
        if (gfxrecon::decode::DetectAPIs(input_filename, detected_d3d12, detected_vulkan)) // One pass over file. Partial but being partial makes it a heuristic.
        {
            if (detected_d3d12)
            {
                GatherD3D12Stats(input_filename); // Second pass over file.
            }
            if (detected_vulkan)
            {
                GatherVulkanStats(input_filename); // Third pass over file.
            }
            // Nth pass for API-N ???
        }
```
If we keep adding new APIs, following this structure will mean adding passes over the capture file.
If we do introduce more mixed-API captures, following this structure will mean extending the detection pass over the entire capture file, rather than terminating it early as now, since a new API could appear at any point.

# The Alternative

Instead of this we should do one pass over the file with all the stats consumers engaged.
* The stats consumers can each record the presence of their respective API as a side effect of their operation. No separate DetectionConsumers need be used.
* Annotations can be recorded outside of all consumers during this single pass by a single object shared by all API consumers:
  `class AnnotationRecorder : public gfxrecon::decode::AnnotationHandler {...}`

After the file is processed
* API-agnostic printing code can dump the same stats as would currently be duplicated for Vulkan and D3D12 mixed capture, but only dump them once. 
* API-specific printing code can dump the same stats as now.
* API-specific printing code can pull out annotations of interest to dump or we can have a shared api-agnostic section for annotations too.

## Benefits
* Makes Info streaming-capable again.
* Reduces IO required, and thus performance when IO is bottleneck (slow drives and network drives over narrow wires).
* Naturally separates API agnostic info from API-specific in future multi-API capture files.

# See also
* https://github.com/LunarG/gfxreconstruct/issues/1025
* Running Info already isn't cheap, justifying another look at it: https://github.com/LunarG/gfxreconstruct/issues/725

# Footnotes
\* Which we treat as a D3D12 capture.

per-mathisen-arm:
[...]

PR #1267, opened 2023-09-20
Initial support for D3D11

dgraves:
Adds initial support for the Direct3D11 API. Capture and replay are stable, but trimming is not currently supported.

The existing D3D code generators were updated to generate D3D11 code along with the existing D3D12 and DXGI related code, which results in most of the D3D11 code being appended to the same files containing the D3D12 and DXGI code. I thought it would be better to let the project developers determine how to best reorganize and rename the existing files, rather than attempt to do it as part of this submission.


dgraves:
When adding D3D11 support to gfxrecon-info, the multiple passes that had been performed for gathering info were merged into a single pass, which may address an item described in the recently created #1256. 

The gfxrecon-info output was slightly modified to move the swapchain info from a position in the middle for the D3D12 info to a position just after the D3D12 info, with the label renamed from "D3D12 swapchain info" to "DXGI swapchain info". For a file containing both D3D11 and D3D12 data, this would result in the D3D11 info being printed, followed by the D3D12 info, and then the DXGI info. If this change to the output is an issue, the swapchain info can be moved back to its original position with its original label and can then be duplicated with the D3D11 info.

Riesi:
@dgraves 
Why did you force push remove the commit implementing d3d11 support?
And also why did you close the PR?
d3d11 support seems like a good addition to this project.

PR #1268, opened 2023-09-21
Trying ndk-build

panos-lunarg:


bradgrantham-lunarg (approved):

juan-lunarg:
> This is done because with ndk-build is easier to get ABIs to compile in parallel

Looking at Jenkins CI all the Android devices support `arm64-v8a` why create an APK with more than necessary ABIs at all except for the final release? I don't understand the motivation here.

Do we have a customer asking us to routinely test `armeabi-v7a`?

PR #1275, opened 2023-09-26
Add support for staging buffer/image for the Rebind Allocator.

laurentiu-erhan-arm:
Some traces are having corruptions due to missing support for non host visible memory 
Case not handled: Staging Image copy with offset


framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -899,6 +939,187 @@ void VulkanRebindAllocator::ReportBindImage2Incompatibility(uint32_t
...
+                }
+            }
+
+            if (original_image)
+            {
+                VkBufferImageCopy region{};
+                region.bufferOffset      = 0;
+                region.bufferRowLength   = 0;
+                region.bufferImageHeight = 0;
+                region.imageSubresource  = { VK_IMAGE_ASPECT_COLOR_BIT, 0, 0, 1 };
```
panos-lunarg, line 1065: What happens in case of multiple mip maps / array layers?

laurentiu-erhan-arm: In the current implementation the other subresources would be ignored (except the first one). This is maybe a TODO for later.

panos-lunarg: In that case I believe a `GFXRECON_LOG_WARNING` would be necessary, otherwise you might still get corruption (although less likely) but  you won't get any hints about what might be causing it

laurentiu-erhan-arm: Thank you for the suggestion, i have added the warning messages

bradgrantham-lunarg (approved):

PR #1276, opened 2023-09-26
Add options to save and load pipeline caches

marius-pelegrin-arm:
Add three options `--save-pipeline-cache`, `--load-pipeline-cache` and `--add-new-pipeline-caches` to manually handle pipeline cache creation/loading instead of using the ones saved in the capture file.

This is useful when pipeline creation is slow on the target system. You can generate the cache once then run the trace multiple times, or generate it on a faster device type with the same GPU and driver.

When using `--save-pipeline-cache` the contents of the pipeline caches are written to the specified file just before their destruction, and just before the destruction of the last bound pipeline for caches created by `--create-new-pipeline-cache`, or at the end of the program if no destruction occurs before.

When using `--load-pipeline-cache` the contents of the pipeline caches are loaded just before their creation, and just before the creation of the pipeline for caches created by
`--create-new-pipeline-cache`.

If set, allows gfxreconstruct to create new vkPipelineCache objects when it encounters a pipeline created without cache. This option can be used in coordination with `--save-pipeline-cache` and `--load-pipeline-cache`.


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7341,6 +7487,14 @@ VkResult VulkanReplayConsumerBase::OverrideCreateRayTracingPipelinesKHR(
     VkDeferredOperationKHR                   in_deferredOperation =
         (deferred_operation_info != nullptr) ? deferred_operation_info->handle : VK_NULL_HANDLE;
     VkPipelineCache in_pipelineCache = (pipeline_cache_info != nullptr) ? pipeline_cache_info->handle : VK_NULL_HANDLE;
+    VkPipelineCache overridePipelineCache = in_pipelineCache;
+
+    // If there is no pipeline cache and we want to create a new one
+
+    if (in_pipelineCache == VK_NULL_HANDLE && options_.add_new_pipeline_caches)
+    {
+        overridePipelineCache = CreateNewPipelineCache(device_info, *pPipelines->GetPointer());
```
charles-lunarg, line 7646: `--add-new-pipeline-cache` only adds a cache to pipelines which do not use one. The cache can be saved with `--save-pipeline-cache` but it wont be used by `--load-pipeline-cache` because when `--add-new-pipeline-cache` is set, a new cache is created while when it is not, no cache is provided since the captured command has a null handle. At least, that is my understanding of the code - I could be wrong about this.

I believe `--add-new-pipeline-cache` should switch behavior based on whether --save-pipeline-cache is specified where a single cache is created and used for all 'null cache' pipelines, then when --load-pipeline-cache the code looks through the loaded file for a cache entry which corresponds to the 'null cache' pipelines and uses that. This way, the caching can still occur.

charles-lunarg (commented):
Seems to be a useful ability to change out the pipeline cache without recapturing a whole new trace - pipeline caches are in a way transient anyhow. 
I do have a question about logging - I found it difficult to know if/when things where happening without inserting "saving pipeline cache" type debug messages. If I found it useful while validating the feature, I would expect others to find it useful in the log. I did not include the messages I wrote while investigating the PR, (rather just rebased on dev).

fabian-lunarg:
_This is useful when pipeline creation is slow on the target system. You can generate the cache once then run the trace multiple times, or generate it on a faster device type with the same GPU and driver._

since this seem primarily aimed at decreasing time of pipeline-creation, it's maybe worth pointing out:
couple months ago we introduced a flag `--pipeline-creation-jobs X` | `--pcj X` 
which will compile pipelines in parallel using a threadpool, decreasing compilation run-time.

the two features 'could' be used in combination and are mostly orthogonal, 
but in its current state the functionality added in this PR would be circumvented when `--pcj X` is used.

reason is use of an async-override code-path, e.g. `AsyncCreateGraphicsPipelines` instead of `OverrideCreateGraphicsPipelines` (which contains the added pipeline-cache logic).

this is something to keep in mind and should either be fixed in this PR or in an immediate follow-up issue/PR.
what do you think @marius-pelegrin-arm ?

per-mathisen-arm:
Fixing later sounds good to me. Might be good to warn if a user tries to use both, though.

fabian-lunarg:
ok, then we could add into `VulkanReplayConsumerBase::VulkanReplayConsumerBase`:

```
if (UseAsyncOperations())
{
   ...
    if (options_.add_new_pipeline_caches || !options_.save_pipeline_cache_filename.empty() ||
        !options_.load_pipeline_cache_filename.empty())
    {
        GFXRECON_LOG_WARNING("Requested both asynchronous pipeline-creation (--pipeline-creation-jobs) and "
                             "explicit pipeline-caches (--save-pipeline-cache | --load-pipeline-cache). This is "
                             "currently not supported and will "
                             "prevent usage of pipeline-caches.");
    }
}
```

marius-pelegrin-arm:
Sounds good to me, I will make another PR for when both options are used together

fabian-lunarg (approved):
no further concerns. after we merge here, we will create a follow-up issue to look into interop with --pcj option. until then a warning will be issued when using both features.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -222,11 +222,42 @@ VulkanReplayConsumerBase::VulkanReplayConsumerBase(std::shared_ptr<application::
             num_threads += (int32_t)std::thread::hardware_concurrency();
         }
         background_queue_.set_num_threads(std::clamp<uint32_t>(num_threads, 0, std::thread::hardware_concurrency()));
+
+        if (options_.add_new_pipeline_caches || !options_.save_pipeline_cache_filename.empty() ||
+            !options_.load_pipeline_cache_filename.empty())
+        {
+            GFXRECON_LOG_WARNING("Requested both asynchronous pipeline-creation (--pipeline-creation-jobs) and "
+                                 "explicit pipeline-caches (--save-pipeline-cache | --load-pipeline-cache). This is "
+                                 "currently not supported and will "
```
MarkY-LunarG, line 231: Nit: This line and the next could probably b merged in the same line.

fabian-lunarg: yeah, that hand-writing looks like autocorrect :) 
done

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5998,110 +6029,146 @@ VkResult VulkanReplayConsumerBase::OverrideCreatePipelineCache(
...
+            // this behaviour so that a cache can be created incrementally by loading the partial cache from the
+            // previous run and feeding it to create the next cache with more data until everything is built.
+            override_create_info.initialDataSize = 0;
+            override_create_info.pInitialData    = nullptr;
+        }
+    }
+
+    // If pipeline cache must not be loaded
+
+    else if (options_.omit_pipeline_cache_data)
```
MarkY-LunarG, line 6059: It seems to me that if the omit is present, we shouldn't do any pipeline cache loading.  Is this really the way we want to behave?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5998,110 +6029,146 @@ VkResult VulkanReplayConsumerBase::OverrideCreatePipelineCache(
...
+
+    else if ((override_create_info.pInitialData != nullptr) && (override_create_info.initialDataSize != 0))
     {
-        auto& create_info = *pCreateInfo->GetPointer();
-        if ((create_info.pInitialData != nullptr) && (create_info.initialDataSize != 0))
-        {
-            // keep track if external synchronization is required
-            auto handle_info = reinterpret_cast<VulkanPipelineCacheInfo*>(pPipelineCache->GetConsumerData(0));
-            handle_info->requires_external_synchronization =
-                create_info.flags & VK_PIPELINE_CACHE_CREATE_EXTERNALLY_SYNCHRONIZED_BIT;
```
MarkY-LunarG, line 6032: Was this intentionally removed?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -10399,5 +10608,143 @@ void VulkanReplayConsumerBase::SetCurrentBlockIndex(uint64_t block_index)
...
+    {
+        GFXRECON_LOG_ERROR("Could not open pipeline cache file '%s' for loading. Error: '%s'",
+                           options_.save_pipeline_cache_filename.c_str(),
+                           strerror(error));
+        return;
+    }
+
+    format::HandleId idRead;
+    uint64_t         cacheSizeRead;
+
```
MarkY-LunarG, line 10624: Should we at least add a header so we know we're reading a pipeline cache file?

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: agree with you, omit means omit. looking into it

fabian-lunarg: to me it looks it matches the previous behavior in this case. I understand the intention as "omit any data" - a cache is created in any case, but might be empty. we set member `omitted_pipeline_cache_data_` to true in case we actually omitted anything

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: done

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: hmm, not sure. the feature will only read files explicitly passed on cmd-line. even when loading 'valid' pipeline-cache-files written by gfxr, they might still be in-compatible for other reasons (different driver/vendor). while I also think this file could contain 'some' more useful metadata, like a header identifying it as pipeline-cache, meta-data about driver/vendor producing the cache etc. - I don't see an immediate danger or risk of leaving the plain cache-file structure as-is and improve on it in a follow-up, when we gathered some experience using it. what do you think?

framework/decode/vulkan_replay_consumer_base.cpp
MarkY-LunarG: But what if the cache isn't empty.  We're saying if the user selects load cache and omit cache, the load overrides the omit.  Seems wrong to me.  Maybe someone else can comment.

fabian-lunarg: ok, took me a bot longer. went back to that part and rewrote to fully respect `options_.omit_pipeline_cache_data`. no data will be loaded if this is set and "load cache" option cannot override.

framework/decode/vulkan_replay_consumer_base.cpp
MarkY-LunarG: Let's push it to later for now.  Create a new issue.

MarkY-LunarG (approved):

PR #1279, opened 2023-09-26
Remove redundant calls to VulkanCaptureManager::Get()

### LINKED ISSUE #1198: Redundant Calls to VulkanCaptureManager::Get()
Reviewing a header update I noticed there can be four or five calls to `VulkanCaptureManager::Get()` in a single generated wrapper function in the capture layer. E.g. for [CmdSetDepthBias2EXT](https://github.com/LunarG/gfxreconstruct/pull/1193/files#diff-c1244ea3fcd8bc72af59da2cd327146aeec97a98d032827e12c898ab47b7c17cR18564). The function `VulkanCaptureManager::Get()` is inline and simply returns a static variable, but every time the compiler can't prove to itself that that static variable hasn't been modified indirectly through some alias, it will go and get its value from the static, thus keeping that containing page in the TLB and that containing cacheline cached, taking slots from other data.

Perhaps we should have code like the following at the top of every generated capture wrapper function and then use the local variable throughout the function.
```
VulkanCaptureManager* manager = VulkanCaptureManager::Get();
GFXRECON_ASSERT(manager != nullptr);
```

charles-lunarg:
The VulkanCaptureManager::Get() function is static and returns a pointer to the current capture manager instance. While the compiler should inline these calls, it may fail due to possible aliasing as the instance pointer is not const (but should never change after setup).
It is better to call get once and explicitly put it in a local variable.

Fixes #1198


andrew-lunarg:
I know it is a lot of faff but would you have time to do a capture with and without this change to verify there isn't a nonintuitive performance regression since it touches every function we capture? Nothing exhaustive, just a few frames of a reasonably representative "interactive graphical application". I'm guessing there will be no measurable change but you never know unless...

bradgrantham-lunarg:
Update - this PR crashes in CreateInstance at time of writing - in Debug it asserts and in Release it dereferences a nullptr.  @charles-lunarg knows about the issue and will address as time permits.

andrew-lunarg:
On reflection, happy to put this in if the crash is fixed. A minimal check that perf isn't unintuitively negatively affected would be nice but I wouldn't block the PR waiting on that.

charles-lunarg:
Finally got around to fixing the crash. Problem was that `VulkanCaptureManager::CreateInstance()` needs to be called before `VulkanCaptureManager::Get()` will work. The current change moves the call to `CreateInstance()` to be inside of `vkCreateInstance`, which was the most straight forwards way to make that work since special casing `vkCreateInstance` would make the codegen much more convoluted.

bradgrantham-lunarg:
I'm not sure moving CreateInstance is necessary and it caused a lot of churn in the code.  I think I'd prefer you to return to the previous version of the code but then just special-case CreateInstance so it doesn't create a `manager` variable.

charles-lunarg:
I did the change that moved CreateInstance because I felt it would cause a lot of changes in the python script. But, after trying it out, it isn't as many changes as I feared.

bradgrantham-lunarg (approved):

PR #1280, opened 2023-09-27
Write replay measurements into a JSON file - Finalised

andrew-lunarg:
PR #1175 was merged with dev, mixing commits into the history and creating a merge commit, which we do not do. I would have had to rewrite history to do the required rebase and push to the original branch which I don't want to do so I made a fresh PR for the final rebase on dev. There were also a couple of small changes, one of which needed my knowledge of upcoming changes to the codebase so I went ahead and made them myself.


marius-pelegrin-arm:
Arf, so sorry, and I did the same mistake with https://github.com/LunarG/gfxreconstruct/pull/1276
Is there a way for me to fix this for this other PR ?

andrew-lunarg:
> Is there a way for me to fix this for this other PR ?
If you know the commit ID before the merge you can reset your branch's head to that. E.g. if you made a tag before merging as a backup.

I did something like this to get a fresh new branch with your changes as a single commit after the head of dev:

```bash
git diff dev yourbranch > patch.diff
git checkout dev
git checkout -b fresh_branch
git apply patch.diff
```

PR #1281, opened 2023-09-28
Removed std:remove For Measurements File

andrew-lunarg:
Deleted one of the two `std::remove()`s and documented and checked the other.


andrew-lunarg:
Draft as it is untested. Will ad-hoc test if approach looks good versus just deleting the remove entirely.

bradgrantham-lunarg (approved):
^pertial^partial and then good to go in

PR #1285, opened 2023-09-29
use new SDK tagging scheme

bradgrantham-lunarg:
Update the SDK tag Android build action to catch `vulkan-sdk-*`

### DIFF
--- a/.github/workflows/sdk_android_build.yml
+++ b/.github/workflows/sdk_android_build.yml
@@ -18,7 +18,7 @@ name: SDK Android Build
 on:
   push:
     tags:
-      - sdk-*
+      - vulkan-sdk-*
 
 jobs:
   android:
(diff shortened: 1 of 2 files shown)


juan-lunarg (approved):

juan-lunarg (changes requested):
You are going to need to update your CMake code as well:

image

https://github.com/search?q=repo%3ALunarG%2Fgfxreconstruct%20sdk-&type=code

bradgrantham-lunarg:
> You are going to need to update your CMake code as well:

Excellent catch.  Thanks!  We probably don't need to detect "^sdk" *and* "^vulkan-sdk" but it may be helpful for someone - getting the right regex incantation to catch both may not be that hard but testing it will be annoying.  For now I may just update to "vulkan-sdk" and then after a "vulkan-sdk" tag exists  and I can test it I'll go back and figure out detecting both forms.

juan-lunarg (approved):

juan-lunarg (approved):

andrew-lunarg (approved):
Looks okay to me, caveat I have never touched these areas myself.

PR #1286, opened 2023-09-29
Update to Vulkan-Headers v1.3.265

andrew-lunarg:
Updates Vulkan-Headers to v1.3.265 and fixes what breaks as a result of that.

# Stages of Upgrade to v1.3.265

## v1.3.262: Looks fine for mechanical generation.
  * `VK_QCOM_image_processing2`: Looks fine for mechanical generation.
  * `VK_QCOM_filter_cubic_weights`: Looks fine for mechanical generation.
  * `VK_QCOM_filter_cubic_clamp`: Looks fine for mechanical generation.
  * `VK_QCOM_ycbcr_degamma`: Looks fine for mechanical generation.
  
## v1.3.263: Looks fine for mechanical generation.
  * `VK_NV_descriptor_pool_overallocation`: Looks fine for mechanical generation.
  * There is some random enum churn that looks okay.
  * Video encode/decode churn. 
    *  `StdVideoEncodeH265LongTermRefPics`: Looks fine.
    
## v1.3.264
  * `VK_MSFT_layered_driver`: Looks fine for mechanical generation.
  * `VK_EXT_frame_boundary`: Has a complex-looking struct:
    * [`VkFrameBoundaryEXT`](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkFrameBoundaryEXT.html).
      * Void ptr `pTag` gets represented in the `Decoded_VkFrameBoundaryEXT` as:
         `uint64_t pTag{ 0 };`
         It should be:
         `PointerDecoder<uint8_t> pTag;`
         It gets encoded like this:
         `encoder->EncodeSizeTValue(value.tagSize);`
         `encoder->EncodeVoidPtr(value.pTag);`
         Should be like this:
         `encoder->EncodeSizeTValue(value.tagSize);`
         `encoder->EncodeVoidArray(value.pTag, value.tagSize);`
         It gets decoded as:
         `bytes_read += ValueDecoder::DecodeAddress((buffer + bytes_read), (buffer_size - bytes_read), &(wrapper->pTag));
    value->pTag = nullptr;`
         It should be:
         `bytes_read += wrapper->pTag.DecodeVoid((buffer + bytes_read), (buffer_size - bytes_read));
    value->pTag = wrapper->pTag.GetPointer();`
         The `FieldToJson()` call used by Convert looks okay but it will dispatch to the wrong version as `pTag` has the wrong type.
    
         This is all because `vk.xml` is wrong. It should specify the length like this:
         `<member len="tagSize">const <type>void</type>* <name>pTag</name></member>`
         It is actually like this with no size specified:
         `<member optional="true">const <type>void</type>* <name>pTag</name></member>`
         * **Conclusion**: Nothing for us to do in our repo. The `vk.xml` file needs fixing (maybe it already is fixed in a tag below).
         * **EDIT: Conclusion 2**: A workaround is required until the xml is fixed in a later release. 
         
      * It could have maybe:
        * [ ] Convert: dump `tagSize` bytes of the data pointed-to by `pTag` to a binary blob.

## v1.3.265: No change to code gen

Changes are in `*.hpp` and other areas that don't affect us.

# Deferred to Next PR

##  v1.3.266 < Currently working on

* `vkLatencySleepNV` is failing in call encoders generation because the argument `VkLatencySleepInfoNV* pSleepInfo` is not const even though it is an input argument and we use the const attribute to detect input singular pointers.
* Android types need to be recreated in platform_types.h for other platforms.

# Triggered CI Builds
* Vulkan-Headers v1.3.262 ->  queue ID 52720, build # 3306 ✅
* Vulkan-Headers v1.3.263 ->  queue ID 52792, build # 3308
* Vulkan-Headers v1.3.264 ->  queue ID 52792, build # 3309


framework/generated/vulkan_generators/gencode.py
```diff
@@ -873,11 +873,28 @@ def gen_target(args):
...
+        # Workaround 1.2.264 VkFrameBoundaryEXT.pTag lacking len field:
+        # <https://github.com/KhronosGroup/Vulkan-Docs/pull/2240>
+        if type.get('name') == 'VkFrameBoundaryEXT' and type.get('category') == 'struct':
+            for member in tree.iter('member'):
+                name = member.find('name')
+                if(name != None and name.text == 'pTag'):
+                    member.set('len', 'tagSize')
+                    ## @todo Early-out the top-level find as all patches are complete (throw?)
+    end_timer(args.time, '* Time to patch ElementTree =')
+
```
andrew-lunarg, line 897: I'd like eye on this please. This changes the in-memory representation of vk.xml to patch an issue in Vulkan Headers that came in at 1.3.264 and has not been fixed yet. It lets our automated codegen handle VkFrameBoundaryEXT correctly for encode and decode.

bradgrantham-lunarg (approved):
Looks good to me.  Have asked @mikes-lunarg to pass his eyes over the change to `gencode.py` just to see if anything jumps out.

framework/generated/vulkan_generators/gencode.py
mikes-lunarg: LGTM. You _could_ wrap it all up in a single XPath query, but I'm not sure if that is any more clear:
```
    if ptag_member := tree.find('types/type[@name="VkFrameBoundaryEXT"]/member[name="pTag"]'):
        ptag_member.set('len', 'tagSize')
```

mikes-lunarg (approved):

locke-lunarg (approved):

andrew-lunarg:
Thanks @mikes-lunarg ! Done.

PR #1287, opened 2023-10-02
Added handling missing VK_EXT_swapchain_colorspace

naomipappe:
- Check whether the color space used by swapchain is supported by the replay device. Inform the user through log if it is not.
- Introduce an option, that when enabled, will fallback to supported color space for swapchain.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -64,24 +64,22 @@ const std::unordered_set<std::string> kSurfaceExtensions = {
     VK_KHR_WIN32_SURFACE_EXTENSION_NAME,   VK_KHR_XCB_SURFACE_EXTENSION_NAME, VK_KHR_XLIB_SURFACE_EXTENSION_NAME
 };
 
-const char                                kSwapchainColorspaceExtensionName[] = "VK_EXT_swapchain_colorspace";
-const std::unordered_set<VkColorSpaceKHR> kColorspaceSwapchainExtension       = { VK_COLOR_SPACE_ADOBERGB_LINEAR_EXT,
-                                                                                  VK_COLOR_SPACE_ADOBERGB_NONLINEAR_EXT,
-                                                                                  VK_COLOR_SPACE_BT2020_LINEAR_EXT,
-                                                                                  VK_COLOR_SPACE_BT709_LINEAR_EXT,
-                                                                                  VK_COLOR_SPACE_BT709_NONLINEAR_EXT,
-                                                                                  VK_COLOR_SPACE_DCI_P3_LINEAR_EXT,
-                                                                                  VK_COLOR_SPACE_DCI_P3_NONLINEAR_EXT,
-                                                                                  VK_COLOR_SPACE_DISPLAY_P3_NONLINEAR_EXT,
-                                                                                  VK_COLOR_SPACE_DOLBYVISION_EXT,
-                                                                                  VK_COLOR_SPACE_EXTENDED_SRGB_LINEAR_EXT,
-                                                                                  VK_COLOR_SPACE_EXTENDED_SRGB_NONLINEAR_EXT,
-                                                                                  VK_COLOR_SPACE_HDR10_HLG_EXT,
-                                                                                  VK_COLOR_SPACE_HDR10_ST2084_EXT,
-                                                                                  VK_COLOR_SPACE_PASS_THROUGH_EXT };
+const std::unordered_set<VkColorSpaceKHR> kColorspaceSwapchainExtension = { VK_COLOR_SPACE_ADOBERGB_LINEAR_EXT,
+                                                                            VK_COLOR_SPACE_ADOBERGB_NONLINEAR_EXT,
+                                                                            VK_COLOR_SPACE_BT2020_LINEAR_EXT,
+                                                                            VK_COLOR_SPACE_BT709_LINEAR_EXT,
+                                                                            VK_COLOR_SPACE_BT709_NONLINEAR_EXT,
+                                                                            VK_COLOR_SPACE_DCI_P3_LINEAR_EXT,
+                                                                            VK_COLOR_SPACE_DCI_P3_NONLINEAR_EXT,
+                                                                            VK_COLOR_SPACE_DISPLAY_P3_NONLINEAR_EXT,
+                                                                            VK_COLOR_SPACE_DOLBYVISION_EXT,
+                                                                            VK_COLOR_SPACE_EXTENDED_SRGB_LINEAR_EXT,
+                                                                            VK_COLOR_SPACE_EXTENDED_SRGB_NONLINEAR_EXT,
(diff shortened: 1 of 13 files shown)


framework/decode/vulkan_replay_options.h
```diff
@@ -47,6 +47,7 @@ struct VulkanReplayOptions : public ReplayOptions
     bool                         skip_failed_allocations{ false };
     bool                         omit_pipeline_cache_data{ false };
     bool                         remove_unsupported_features{ false };
+    bool                         colorspace_fallback{ false };
```
bradgrantham-lunarg, line 50: Our options are frequently "verb noun", so I suggest "use_colorspace_fallback" or "use_fallback_colorspace" for this variable and the CL option, etc

naomipappe: I renamed the CL option and the variable as suggested.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -64,6 +64,23 @@ const std::unordered_set<std::string> kSurfaceExtensions = {
     VK_KHR_WIN32_SURFACE_EXTENSION_NAME,   VK_KHR_XCB_SURFACE_EXTENSION_NAME, VK_KHR_XLIB_SURFACE_EXTENSION_NAME
 };
 
+const std::unordered_set<VkColorSpaceKHR> kColorspaceSwapchainExtension = { VK_COLOR_SPACE_ADOBERGB_LINEAR_EXT,
```
charles-lunarg, line 67: This is susceptible to code rot. Future extensions may add more enums that must be added but are very easy to forget to update this code.

I was going to suggest inverting this by only checking to see if the colorspace *isn't* VK_COLOR_SPACE_SRGB_NONLINEAR_KHR however, because there are multiple extensions that provide colorspaces, the code has to know which extension the colorspace comes from in order to check whether that extension is enabled.

The only reasonable alternative for me is generating a utility that maps a colorspace enum to its originating extension, which is effectively what is present here.

Generating something like the following:
```c++
std::unordered_map<VkColorSpaceKHR, const char*> kColorSpaceExtensionMap = { 
{VK_COLOR_SPACE_SRGB_NONLINEAR_KHR, nullptr}, // nullptr since its not from an extension
{VK_COLOR_SPACE_ADOBERGB_NONLINEAR_EXT, VK_EXT_SWAPCHAIN_COLOR_SPACE_EXTENSION_NAME},
{VK_COLOR_SPACE_BT2020_LINEAR_EXT, VK_EXT_SWAPCHAIN_COLOR_SPACE_EXTENSION_NAME},
...
```

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2283,6 +2300,35 @@ VulkanReplayConsumerBase::OverrideCreateInstance(VkResult original_result,
...
+                        auto iter = std::find_if(
+                            filtered_extensions.begin(), filtered_extensions.end(), [](const char* extension) {
+                                return util::platform::StringCompare(VK_AMD_DISPLAY_NATIVE_HDR_EXTENSION_NAME,
+                                                                     extension) == 0;
+                            });
+                        if (iter != filtered_extensions.end())
+                        {
+                            filtered_extensions.erase(iter);
+                        }
+                    }
```
charles-lunarg, line 2330: This is a decent amount of duplicate code. I would strongly recommend moving the logic into a helper function, just like `feature_util::RemoveUnsupportedExtensions()`. Could be called `feature_util::RemoveSupportedExtension(filtered_extensions, "VK_EXT_<extension_name>)` or something similar.

bradgrantham-lunarg:
Hi, @naomipappe , do you have any thoughts about the feedback from @charles-lunarg ?  Thanks!

naomipappe:
Yes, I fully agree with the comments about reducing the code duplication, and I am currently looking into how code generation is done in gfxreconstruct to implement the automatically generated map to avoid code rot.

charles-lunarg (approved):
Thank you for going through the somewhat laborious effort of making this generated rather than hardcoded. Makes the implementation a bit cleaner too I think, even if it involves more code.
I don't know if/when more swapchain formats will be added, but gfxr will handle it with ease now.

framework/generated/generated_vulkan_constant_maps.h
```diff
@@ -0,0 +1,75 @@
...
+{VK_COLOR_SPACE_ADOBERGB_NONLINEAR_EXT, "VK_EXT_swapchain_colorspace"},
+{VK_COLOR_SPACE_PASS_THROUGH_EXT, "VK_EXT_swapchain_colorspace"},
+{VK_COLOR_SPACE_EXTENDED_SRGB_NONLINEAR_EXT, "VK_EXT_swapchain_colorspace"},
+{VK_COLOR_SPACE_DCI_P3_LINEAR_EXT, "VK_EXT_swapchain_colorspace"},
+{VK_COLOR_SPACE_DISPLAY_NATIVE_AMD, "VK_AMD_display_native_hdr"},
+};
+
+static const std::vector<const char *> kColorSpaceExtensionNames = {
+"VK_AMD_display_native_hdr",
+"VK_EXT_swapchain_colorspace",
```
charles-lunarg, line 69: It appears that the Ubuntu-Release CI job is failing due to different output from the generator.

https://github.com/LunarG/gfxreconstruct/actions/runs/7100394848/job/19340944141?pr=1287

I have python 3.11 on my machine and what is checked into the repo matches what is generated. Ubuntu-Release CI job runs on ubuntu-latest, which uses python 3.10. What Python version are you using?

framework/generated/vulkan_generators/vulkan_constant_maps_generator.py
```diff
@@ -0,0 +1,117 @@
...
+        for key, value in colorspace_extension_map.items():
+            write(f'{{{key}, {value}}},', file=self.outFile)
+        write('};', file=self.outFile)
+        self.newline()
+
+        write(
+            'static const std::vector<const char *> kColorSpaceExtensionNames = {',
+            file=self.outFile
+        )
+        for extension_name in colorspace_extensions_names:
```
charles-lunarg, line 106:
```suggestion
        for extension_name in sorted(colorspace_extensions_names):
```
This should be enough to fix the lack of ordering in the generator output. Turns out set() is unordered, while dict() maintains insertion order in python 3.7+. But since we only need the order during printing, we can sort it before output like so.

charles-lunarg (changes requested):

framework/generated/vulkan_generators/vulkan_constant_maps_generator.py
charles-lunarg: Went ahead and pushed the fix. Its a one liner after all.

charles-lunarg (approved):

framework/generated/generated_vulkan_constant_maps.h
naomipappe: I used Python 3.8.10 for this

PR #1288, opened 2023-10-02
Re-attempt adding Android Frame boundary extension implementation

MarkY-LunarG:
This updates PR #981 to the latest `dev` tree source.

Re-attempting to bring it into GFXR since Google has requested support for it.


MarkY-LunarG:
Previous comments from the other PR from @mikes-lunarg (just to keep them all in one place):

> Capture the vkFrameBoundaryANDROID function and treat it as a frame delimiter.
>
> It is very simple extension. The most complicated parts are:
>
> There was an existing bug in the capture layer where it wouldn't always return an pointer for layer-implemented functions. This is handled in the first commit.
> VK_ANDROID_frame_boundary is not in the main spec so this change also adds a way to patch the vk.xml tree so the code generators can know about the extension.
> 
> Questions:
> 
> Is this useful? I thinks so, HWUI is already using it. Ideally this private vendor extension would be replaced with a EXT or KHR extension and hosted in the main spec.
> Should there be an option to enable it? Probably yes. It makes sense to make undocumented/experimental functionality opt-in.
> Should it be exposed on Android only? I'm not sure. Exposing it on other platforms doesn't hurt, but also isn't very useful.
> Should it be exposed in the layer manifest file? Probably no. This is a private extension so I don't see much value in telling the desktop loader we support it.
>
>Future work:
>
>It would be nice to hook this up to the virtual swapchain (and screenshot!) on replay. Replaying a blank screen isn't very exciting...

bradgrantham-lunarg (approved):

MarkY-LunarG:
Rebased on Andrew's changes, and later re-generated files to make sure nothing new was needed.  Just making sure it passes through CI now.

PR #1289, opened 2023-10-02
Fix resource state D3D12 err msg when trimming

davidd-lunarg:
When saving and restoring resource data in Dx12ResourceDataUtil, the resource state is transitioned to D3D12_RESOURCE_STATE_COMMON. After the resource is copied to or from, the resource state is promoted to D3D12_RESOURCE_STATE_COPY_SOURCE or D3D12_RESOURCE_STATE_COPY_DEST. Use the promoted state as the before state of the transition after the copy to avoid D3D12 debug layer errors about resource state mismatch.

### DIFF
--- a/framework/graphics/dx12_resource_data_util.cpp
+++ b/framework/graphics/dx12_resource_data_util.cpp
@@ -688,6 +688,7 @@ Dx12ResourceDataUtil::ExecuteCopyCommandList(ID3D12Resource*                    
 
                 // Prepare resource state.
                 AddTransitionBarrier(command_list_, target_resource, i, before_states[i], copy_state);
+                dx12::ResourceStateInfo promoted_state = copy_state;
 
                 // Copy data.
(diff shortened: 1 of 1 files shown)


framework/graphics/dx12_resource_data_util.cpp
```diff
@@ -688,6 +688,7 @@ Dx12ResourceDataUtil::ExecuteCopyCommandList(ID3D12Resource*
 
                 // Prepare resource state.
                 AddTransitionBarrier(command_list_, target_resource, i, before_states[i], copy_state);
+                dx12::ResourceStateInfo promoted_state = copy_state;
```
rurra-amd, line 691: Would it be better to rename:

`copy_state` to `pre_copy_state` 
`promoted_state` to `post_copy_state` 

?

rurra-amd (changes requested):

davidd-lunarg:
After additional testing, I found that this change introduces some rendering artifacts for the first trimmed frame in some rare cases. Closing and won't merge.

PR #1295, opened 2023-10-05
Added Frame Range Option to Convert Tool

laurentiu-erhan-arm:
Frame range option allows to only retrieve relevant data and reduce waiting time, especially useful for  bigger traces

Change-Id: I3b7298a20c41d2c67a59069ddf31961f37a3b12d

### DIFF
--- a/tools/convert/main.cpp
+++ b/tools/convert/main.cpp
@@ -47,7 +47,7 @@ using Dx12JsonConsumer =
 #endif
 const char kOptions[] = "-h|--help,--version,--no-debug-popup,--file-per-frame,--include-binaries,--expand-flags";
 
-const char kArguments[] = "--output,--format,--log-level";
+const char kArguments[] = "--output,--format,--log-level,--frame-range";
 
 static void PrintUsage(const char* exe_name)
 {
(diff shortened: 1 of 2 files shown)


tools/convert/main.cpp
```diff
@@ -72,6 +72,9 @@ static void PrintUsage(const char* exe_name)
     GFXRECON_WRITE_CONSOLE(
         "  --file-per-frame\tCreates a new file for every frame processed. Frame number is added as a suffix");
     GFXRECON_WRITE_CONSOLE("                  \tto the output file name.");
+    GFXRECON_WRITE_CONSOLE(
+        "  --frame-range\t\tFrame range to be converted. In order to retrieve Trim Trace State, frame 0 has "
+        "to be in frame range.");
```
andrew-lunarg, line 77: It would be good to tell the user the format for specifying the frame range here.

laurentiu-erhan-arm: I have updated the help message to:
Frame ranges to be converted. In order to retrieve Trim Trace State, frame 0 has to be in frame range.
Frame ranges should be specified in ascending order and cannot overlap. Example: 0-5,15,20,25-50,51.

MarkY-LunarG:
Rebased the changes and did a cleanup change on top afterwards to remove some extra if's and branching.

MarkY-LunarG (approved):

PR #1297, opened 2023-10-05
Faster compilation

panos-lunarg:
Function InitializeOutputStructPNext in
generated_vulkan_replay_consumer.cpp is taking a
significant time to compile for some build configuration. Converting it to a non template function drops compilation times drastically

### DIFF
--- a/framework/generated/generated_vulkan_replay_consumer.cpp
+++ b/framework/generated/generated_vulkan_replay_consumer.cpp
@@ -37,8 +37,7 @@
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
-template <typename T>
-void InitializeOutputStructPNext(StructPointerDecoder<T> *decoder);
+void InitializeOutputStructPNext(size_t len, const void* input, void* output, size_t struct_type_size);
 
 void VulkanReplayConsumer::Process_vkCreateInstance(
     const ApiCallInfo&                          call_info,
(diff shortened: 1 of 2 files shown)


framework/generated/generated_vulkan_replay_consumer.cpp
```diff
@@ -9203,18 +9202,15 @@ void VulkanReplayConsumer::Process_vkCmdDrawMeshTasksIndirectCountEXT(
...
-    for( size_t i = 0 ; i < len; ++i )
+    if (!input || !output) return;
+
+    for (size_t i = 0 ; i < len; ++i)
     {
-        const auto* in_pnext = reinterpret_cast<const VkBaseInStructure*>(input[i].pNext);
-        if( in_pnext == nullptr ) continue;
-        auto* output_struct = reinterpret_cast<VkBaseOutStructure*>(&output[i]);
+        const auto* in_pnext = reinterpret_cast<const VkBaseInStructure*>(static_cast<const uint8_t*>(input) + i * struct_type_size)->pNext;
+        if (in_pnext == nullptr) continue;
```
andrew-lunarg, line 9212: If this a problem, it was a problem in the original code too, but are we leaving an uninitialized hole in the output array if we hit a null ptr on input here? Should we zero the pnext of the output at least?

andrew-lunarg (approved):
I love it. C++ templates are a form of code generation but we are already doing code generation here so why generate code that generates other code? This is better.

panos-lunarg:
Dropping this in favor of #1318

PR #1301, opened 2023-10-06
Add vkCmdInsertDebugUtilsLabelEXT as a frame delimiter

marius-pelegrin-arm:
Currently only `vkCmdDebugMarkerInsertEXT` (under conditions on the name of the marker) is considered a frame delimiter (aside from queue present).

However, `VK_EXT_debug_marker` is deprecated and promoted to the `VK_EXT_debug_utils` extension that support an almost identical command `vkCmdInsertDebugUtilsLabelEXT`.

This commit only add this command as a frame delimiter under the same conditions on the name of the label (which replaces the marker).


MarkY-LunarG:
Force pushed a rebase on latest dev to resolve issues.

MarkY-LunarG (approved):

PR #1302, opened 2023-10-06
Added -j/--parallel option to build script

laurentiu-erhan-arm:
CMake 3.12 and above support parallel build level. An option to support this has been added to the build script

Change-Id: Ibc36cb9fbc7a77b826a030bce934cc3bdca48dde


etang-cw:
The project requires CMake 3.16, so I don't think the 3.12 check is necessary:
https://github.com/LunarG/gfxreconstruct/blob/b40adaaa3fc67540c83619a36e7a60a39f856ec5/CMakeLists.txt#L28

Could you also support automatically using `os.cpu_count()` cores without having to externally calculate that?  Maybe with `--parallel 0` or something?  I know it's easy to tell people to use `$(nproc)`, but once you try to extend that to more platforms you end up with "well on Linux you do `$(nproc)`, but on macOS you should use `$(sysctl -n hw.ncpu)`, and on Windows..." and it ends up being much easier to have the program calculate it for you.

And finally, since you're adding the option, might as well make the [ci script](https://github.com/LunarG/gfxreconstruct/blob/dev/.github/workflows/ci_build.yml) use it to speed up builds

laurentiu-erhan-arm:
Thank you for the suggestions. 

I have removed the parts of the build script that checked for CMake versions lower than CMake 3.16.3.
Handled/Assigning passing value 0 (for --parallel) as using all logical CPUs. 
I have added the --parallel 0 option to the ci and release build. (This may not be the best option depending if the machine will run multiple builds in parallel, so I am open for a suggestions on a more conservative value)

etang-cw:
> This may not be the best option depending if the machine will run multiple builds in parallel, so I am open for a suggestions on a more conservative value

None of the GH Actions documentation I've seen mentions trying to reduce the number of jobs you run to leave cores for other jobs, so I'd expect they give your VM the number of cores they expect you to use.

etang-cw (approved):

andrew-lunarg (approved):
I'm used to setting this via the env var `CMAKE_BUILD_PARALLEL_LEVEL` and documented the use of that in one of our usage or build or contrib docs but this seems more discoverable and cleaner.

I don't think this will cause the generation of `add_definitions(/MP)` for Windows MSVC builds because the environment variable doesn't so it isn't the final word on parallel builds.

PR #1304, opened 2023-10-09
Remove oldswapchain images when swapchaincreate is captured

davidlunarg:
This is an experimental change. A generated file was directly modified so as to see how CI tests behave when images are removed from oldswapchain.


MarkY-LunarG:
Do we need to worry about the swapping usages in use?  Do we need a wait for device idle?  I can't see on my phone, so maybe one is there already.

PR #1306, opened 2023-10-10
Fix in ProcessInitImageCommand

panos-lunarg:
When capturing trimmed under certain conditions texture data were dumped to the capture file using a staging buffer and at during replay they were loaded back to the texture without a staging buffer. This occasionally led to corrupted textures. This patch should address this issue

### DIFF
--- a/framework/encode/vulkan_state_writer.cpp
+++ b/framework/encode/vulkan_state_writer.cpp
@@ -1755,7 +1755,7 @@ void VulkanStateWriter::WriteImageMemoryState(const VulkanStateTable& state_tabl
             VkMemoryPropertyFlags memory_properties = 0;
             if (memory_wrapper != nullptr)
             {
-                GetMemoryProperties(device_wrapper, memory_wrapper, state_table);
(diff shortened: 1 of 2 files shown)


framework/encode/vulkan_state_writer.cpp
```diff
@@ -1755,7 +1755,7 @@ void VulkanStateWriter::WriteImageMemoryState(const VulkanStateTable& state_tabl
             VkMemoryPropertyFlags memory_properties = 0;
             if (memory_wrapper != nullptr)
             {
-                GetMemoryProperties(device_wrapper, memory_wrapper, state_table);
+                memory_properties = GetMemoryProperties(device_wrapper, memory_wrapper, state_table);
```
panos-lunarg, line 1758: I caught this while reading the code. It is not directly related to the issue this patch intents to fix.

panos-lunarg: Line was changed by 009b820f924dcb269365e71bebc9e100fb4358fb (@locke-lunarg )

bradgrantham-lunarg (approved):

PR #1307, opened 2023-10-10
Upgrade to Agility SDK 1.610.5

rurra-amd:
**Problem**
As cutting edge apps adopt features of newer Agility SDKs, GFXR needs to keep up in order to reliably capture/replay those apps.

**Solution**
Upgrade from 1.606.4 to 1.610.5

**Testing**
Was able to capture/replay a few apps using the upgraded SDK.


framework/encode/custom_dx12_api_call_encoders.cpp
```diff
@@ -99,5 +100,35 @@ void Encode_ID3D12Resource_WriteToSubresource(ID3D12Resource_Wrapper* wrapper,
...
+{
+    auto encoder = D3D12CaptureManager::Get()->BeginMethodCallCapture(
+        format::ApiCallId::ApiCall_ID3D12Device12_GetResourceAllocationInfo3, wrapper->GetCaptureId());
+    if (encoder)
+    {
+        encoder->EncodeUInt32Value(visibleMask);
+        encoder->EncodeUInt32Value(numResourceDescs);
+        EncodeStructArray(encoder, pResourceDescs, numResourceDescs);
+        encoder->EncodeUInt32Array(pNumCastableFormats, numResourceDescs);
+        for (UINT i = 0; i < numResourceDescs; i++)
```
davidd-lunarg, line 120: I am not able to find any documentation for `GetResourceAllocationInfo3` beyond the declaration in `d3d12.h`, but based on the arguments it looks like `ppCastableFormats` is an array of size `numResources` and each `ith` entry of `ppCastableFormats` is an array of size `pNumCastableFormats[i]`. Both `pNumCastableFormats` and `ppCastableFormats` are marked as optional arguments so they may be null.

This 2D array encoding appears to be a unique case for DX12. There is an existing method `ParameterEncoder::EncodeArray2D` that might be useful for handling the encoding of `ppCastableFormats`. See also `ParameterEncoder::EncodeUInt32Array2D`.

Also, because `GetResourceAllocationInfo3` is added to `blacklists.json` there needs to be a custom decoder in `Dx12DecoderBase` and handling for it in `Dx12DecoderBase::DecodeMethodCall`.

davidd-lunarg (changes requested):

framework/encode/custom_dx12_api_call_encoders.cpp
rurra-amd: Ok updated code. I added the missing decode paths and tweaked the encode. Spent a little time trying to adopt the 2D array encoders, but haven't had luck there getting past the compiler. So just encoding the enums values one by one in this case.

davidd-lunarg: This [commit](https://github.com/davidd-lunarg/gfxreconstruct/commit/8a4232047a40b84f07a1846de22410a7f3d44df9) shows an example of using `EncodeArray2D`, although I don't have a test case to verify it with.

rurra-amd: Done - thanks for the help tightening this up

PR #1310, opened 2023-10-13
C++ Indenter for Python Codegen

andrew-lunarg:
I made this function for Dx12 Convert\* because `inspect.cleandoc` was infuriatingly not doing what I needed. I've pulled it out and applied it on the Vulkan side of JSON generation to try to keep the DX12 branch manageable. If this lands in dev, that other branch will rebase on these changes and simplify for future reviewers.

For anyone reviewing this it might be helpful to note the changes in the generated files are whitespace-only.

\* https://github.com/andrew-lunarg/gfxreconstruct/tree/andy-enhance-better-d3d12-convert

### DIFF
--- a/framework/generated/base_generators/reformat_code.py
+++ b/framework/generated/base_generators/reformat_code.py
@@ -1,0 +1,67 @@
+#!/usr/bin/python3 -i
+#
+# Copyright (c) 2023 Valve Corporation
+# Copyright (c) 2023 LunarG, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to
+# deal in the Software without restriction, including without limitation the
+# rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+# sell copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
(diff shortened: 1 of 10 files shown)


framework/generated/base_generators/reformat_code.py
```diff
@@ -0,0 +1,67 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+# IN THE SOFTWARE.
+
+## @file Helpers used to clean up generated C++ code, allowing generators to be
+## a little rougher in their indentation and then fix the result in post.
+## For example they could use tripple quote multiline string literals to
```
andrew-lunarg, line 26: triple

framework/generated/generated_vulkan_struct_to_json.cpp
```diff
@@ -14012,3510 +14969,3513 @@ void FieldToJson(nlohmann::ordered_json& jdata, const Decoded_VkDrawMeshTasksInd
...
+                const auto* pnext = reinterpret_cast<const Decoded_VkPhysicalDeviceShaderObjectPropertiesEXT*>(data->GetMetaStructPointer());
+                FieldToJson(jdata, pnext, options);
+                break;
             }
-            
+
             case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_TILE_PROPERTIES_FEATURES_QCOM:
             {
-               const auto* pnext = reinterpret_cast<const Decoded_VkPhysicalDeviceTilePropertiesFeaturesQCOM*>(data->GetMetaStructPointer());
-               FieldTo
```
bradgrantham-lunarg, line 18473: This is an interesting non-white-space change.  It seems like the right thing, and I wonder what was happening before but I'm encouraged that this PR fixes it.

bradgrantham-lunarg (approved):

framework/generated/generated_vulkan_struct_to_json.cpp
andrew-lunarg, line 17284: I see a whitespace change @bradgrantham-lunarg . What do you see?

bradgrantham-lunarg: Yeah, I think so.

(For anyone else reading, `git diff --ignore-blank-lines --ignore-all-space --ignore-cr-at-eol` does not output any files that have only whitespace changes.  With that from the command line, it is easier to find that the file `framework/generated/generated_vulkan_struct_to_json.cpp` at line 18468 added a pair of braces around the default case.  That is fine and I think is actually better than what was there before.)

PR #1311, opened 2023-10-13
Updated Vulkan-Headers to v1.3.268

andrew-lunarg:
This is a NOP for us. There are no changes to codegen.

This release is mostly valid usage json changes and minor diffs to the OO .hpp wrappers.

Syncing with other Vulkan SDK projects for next release. Otherwise, the only point of doing this would be to make the next header update simpler / clearer, narrowing the range of commits to look at should there be any issues in that later update.


andrew-lunarg:
> CI gfxreconstruct build # 3385 failed.

Didn't really fail. GPU died on one machine.

andrew-lunarg:
I would like us to get the following PR in ahead of this one: https://github.com/LunarG/gfxreconstruct/pull/1309

PR #1312, opened 2023-10-13
Fix trim trace replay memory requirement mismatch error

mizhen:
**The problem**
   
For current trim handling, the create info of image/buffer is different between capture and replay because VK_IMAGE_USAGE_TRANSFER_SRC_BIT or VK_BUFFER_USAGE_TRANSFER_SRC_BIT is added to usage of related create info during capturing trim trace. For some hardware/driver, this makes the memory requirement different for same resource between capture and replay even on same platform, it triggers the following binding memory failure due to memory requirement mismatch for some title. 

**The solution**

If the trace file is trim trace file, add VK_BUFFER_USAGE_TRANSFER_SRC_BIT or VK_IMAGE_USAGE_TRANSFER_SRC_BIT to related buffer or image create info during playback to make it same with the value during capture time.

**Result**
Tested target title with the build of the pull request, it fix the issue.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -1,6 +1,7 @@
 /*
 ** Copyright (c) 2018-2020 Valve Corporation
 ** Copyright (c) 2018-2023 LunarG, Inc.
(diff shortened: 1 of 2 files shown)


panos-lunarg (approved):
Changes look good to me

PR #1315, opened 2023-10-18
Better handle for RT applications from optimize tool

panos-lunarg:
The optimize tool was not taking under consideration ray tracing
features. As a result when applied on a trimmed capture containing such
calls all relative resources were ommited from the optimized tool
leading to corruption during replay.

One key element missing is the relation between top (tlas) and bottom
(blas) level acceleration structures. This relation is described when
the tlas is being build. However when capturing trimmed the tlas build
commands are likely not to be within the capture range and as a result
they are not dumped into the capture file. This makes it impossible for
the optimize tool to deduce from the used descriptors which blases it
should retain in the optimized file.

To solve this a new metacommand is introduced which describes the tlas
to blas dependency and is writted to the capture file during capture.
For each tlas that is built the new metacommand is written to the
capture file and it's essentially a list of all the blas ids the tlas
uses.

### DIFF
--- a/framework/decode/referenced_resource_table.cpp
+++ b/framework/decode/referenced_resource_table.cpp
@@ -35,7 +35,7 @@ void ReferencedResourceTable::AddResource(format::HandleId resource_id)
     }
 }
 
-void ReferencedResourceTable::AddResource(format::HandleId parent_id, format::HandleId resource_id)
+void ReferencedResourceTable::AddResource(format::HandleId parent_id, format::HandleId resource_id, bool add_children)
 {
     if ((parent_id != format::kNullHandleId) && (resource_id != format::kNullHandleId))
     {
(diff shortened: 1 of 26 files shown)


framework/format/format.h
```diff
@@ -618,6 +619,14 @@ struct Dx12RuntimeInfoCommandHeader
     Dx12RuntimeInfo  runtime_info;
 };
 
+struct TlasToBlasDependencyHeader
```
bradgrantham-lunarg, line 622: Would it be possible to make this a more generic dependency command, like ParentToChildDependencyHeader?  Can we determine dependency from the parent and child object types?  Or does this need to implicitly be specifically for AccelerationStructure handleIds?

framework/decode/referenced_resource_table.cpp
```diff
@@ -104,7 +115,7 @@ void ReferencedResourceTable::AddResourceToContainer(format::HandleId container_
     }
 }
 
-void ReferencedResourceTable::AddResourceToUser(format::HandleId user_id, format::HandleId resource_id)
+void ReferencedResourceTable::AddResourceToUser(format::HandleId user_id, format::HandleId resource_id, bool print)
```
bradgrantham-lunarg, line 118: Did you mean to remove `print` here and in other locations?

framework/encode/vulkan_state_tracker.cpp
```diff
@@ -1424,5 +1469,129 @@ void VulkanStateTracker::DestroyState(SwapchainKHRWrapper* wrapper)
...
+        return;
+    }
+
+    for (uint32_t c = 0; c < command_buffer_count; ++c)
+    {
+        CommandBufferWrapper* cmd_buf_wrapper = GetWrapper<CommandBufferWrapper>(command_buffers[c]);
+
+        for (auto& tlas_build_info : cmd_buf_wrapper->tlas_build_info_map)
+        {
+            // Find to which device memory this address belongs to
```
bradgrantham-lunarg, line 1510: You can delete the last `to` because you already said `to which`

bradgrantham-lunarg (changes requested):
I need a moment to better understand the descent of the structs passed to Build but here are some initial thoughts.  Also note my single comment that was about generalizing the metadata command to record dependencies.

framework/format/format.h
panos-lunarg: Although currently we don't need for other use cases, making it more generic sounds like a good idea.

framework/decode/referenced_resource_table.cpp
panos-lunarg: Right, missed that

framework/encode/vulkan_state_tracker.cpp
panos-lunarg: fixed

bradgrantham-lunarg (approved):

PR #1316, opened 2023-10-19
Fix rename playback hang

igorbrsn:
**The problem**
When run gfxrecon-replay-rename.py with some gfxr traces, the playback may stuck (hang). The issue seems to be related to stdout and stderr PIPE for subprocess.Popen().
Manually rename `gfxrecon-replay.exe` to the application name, playback does not have the hang issue.
If gfxrecon replay displays many messages in the console, the rename Python script could fail. Some applications output many `No ID3D12StateObject was set on the command list before DispatchRays was called. Unable to map the values in the shader table. Replay may fail.` errors to console 'stdout' and replay hangs. 

**The solution**
According to documentation using poll with `stdout = PIPE` can lead to deadlock. `subprocess.run(command, check=True)` can be used to solve this problem. 

**Result**
Playback of traces no longer hangs.

### DIFF
--- a/scripts/gfxrecon-replay-renamed.py
+++ b/scripts/gfxrecon-replay-renamed.py
@@ -53,14 +53,10 @@ def rename_replayer(encoded_app_executable):
     return replayer_name_renamed
 
 def run_command(command):
-    process = subprocess.Popen(command, 
-                                shell=True,
-                                stdout=subprocess.PIPE,
-                                stderr=subprocess.PIPE,
-                                universal_newlines=True)
-    while process.poll() is None:
-        line = process.stdout.readline()
-        print(line)
+    try:
+        subprocess.run(command, check=True)
+    except subprocess.CalledProcessError as e:
+        print ("subprocess returncode: ", e.returncode)
 
 # Run the replayer
 def run_replayer(replay_tool_path, args):


bradgrantham-lunarg:
This failed our internal CI because the base commit for this PR is quite old and so can't replay our current vulkaninfo test capture.  Rebasing the PR's branch to `dev` should solve the problem.

scripts/gfxrecon-replay-renamed.py
```diff
@@ -58,6 +58,7 @@ def run_command(command):
                                 stdout=subprocess.PIPE,
                                 stderr=subprocess.PIPE,
                                 universal_newlines=True)
+    stdout, stderr = process.communicate()
```
bradgrantham-lunarg, line 61: This gets any available output from `process` and then discards it because no code uses the resulting variables "stdout" and "stderr".  (https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate - `Read data from stdout and stderr, until end-of-file is reached. Wait for process to terminate`)  I think the loop afterward is supposed to drain output from the subprocess but after this call to `communicate` will have no results.  This PR should probably print the `stdout` and `stderr` returned and then should probably delete the `poll` loop afterwards.

bradgrantham-lunarg (changes requested):

bradgrantham-lunarg:
Because `gfxrecon-optimize-renamed.py` is based on this code, it needs these changes, too.

bradgrantham-lunarg:
> > Because `gfxrecon-optimize-renamed.py` is based on this code, it needs these changes, too.
> `gfxrecon-optimize-renamed.py` script does not use Popen() and PIPE, so no change is needed.

Oh, right!  My mistake.

bradgrantham-lunarg (approved):

PR #1317, opened 2023-10-19
macOS support

etang-cw:
Adds support for macOS / MoltenVK


andrew-lunarg:
We have a PR in-flight which should achieve the same effect as one of the bonus feature commits in your PR: https://github.com/LunarG/gfxreconstruct/pull/1297/

Would you mind taking a look at that and seeing if we need to pull in anything from your commit? Yours for reference: https://github.com/LunarG/gfxreconstruct/pull/1317/commits/5334fc6d5d6f3ef520f700961b069b7631b7ea46

andrew-lunarg:
We have another PR contribution which adds a parallel build option like the one you add in a commit on this PR. https://github.com/LunarG/gfxreconstruct/pull/1302

There are some differences such as them checking the Cmake version is high enough. Could you also take a look at that and comment what might be missing that your commit https://github.com/LunarG/gfxreconstruct/pull/1317/commits/71f63dc45cc3dea45f6b626b7c58baaec5b805d2 adds?

etang-cw:
Good to see I'm not the only one who noticed those

No preferences on #1297, they both accomplish the same thing

For #1302, I commented my needs on the PR

I'll leave the commits here for now in case anyone wants to use them for reference, and remove them once those PRs get merged (I'll need to add whatever ends up being the ncpu-parallel syntax to the macOS build anyways)

panos-lunarg:
Yes I think Evan's 5334fc6d5d6f3ef520f700961b069b7631b7ea46 is better than #1297

etang-cw:
Do you want me to pull it out into a separate PR?

panos-lunarg:
Sure, why not. Although I don't have the last call on what gets merged, I think this is useful anyway and independent of the general purpose of this PR

bradgrantham-lunarg:
Yes, I think we'd prefer to have unrelated functionality in separate PRs if it's not too much trouble.

etang-cw:
Also, not sure how I didn't notice earlier, but 46a425ecd50bd16b6897d309a267535ccac0c6cc is actually a reversion of f7b95927db84c0bc3b6d6ad0f34f74276aa6c412, which is what broke macOS support.  Do you know what 32-bit platform f7b95927db84c0bc3b6d6ad0f34f74276aa6c412 was fixing?  The Android CI seems to be having no trouble here...

external/stb/stb_image_write.h
```diff
@@ -773,7 +773,9 @@ static int stbi_write_hdr_core(stbi__write_context *s, int x, int y, int comp, f
 #ifdef __STDC_LIB_EXT1__
       len = sprintf_s(buffer, sizeof(buffer), "EXPOSURE=          1.0000000000000\n\n-Y %d +X %d\n", y, x);
 #else
-      len = sprintf(buffer, "EXPOSURE=          1.0000000000000\n\n-Y %d +X %d\n", y, x);
+      len = snprintf(buffer, sizeof(buffer), "EXPOSURE=          1.0000000000000\n\n-Y %d +X %d\n", y, x);
```
bradgrantham-lunarg, line 776: We've tried to keep from changing imported sources (VMA and STB image write) - can this be done some other way?

framework/util/page_guard_manager.cpp
```diff
@@ -102,6 +102,12 @@ static LONG WINAPI PageGuardExceptionHandler(PEXCEPTION_POINTERS exception_point
 #include <signal.h>
 #include <unistd.h>
 
+#ifdef __APPLE__
```
bradgrantham-lunarg, line 105: Can these be made constants a la `const` or `constexpr` or such?

external/stb/stb_image_write.h
etang-cw: I can
- Disable `-Wdeprecated-declarations`, as Apple blocks sprintf by flagging it deprecated
- Mark the stb include directory as a system include directory in cmake, which will disable warnings, but may also disable dependency calculations [according to cmake docs](https://cmake.org/cmake/help/latest/command/target_include_directories.html)

Preferences?

bradgrantham-lunarg: I think maybe mark it as a system include directory and see what the result is there.  Thanks!

bradgrantham-lunarg:
I built the tree but of course my first test was to try to replay a vkcube capture from Linux, which failed because the original capture didn't use PORTABILITY ext or CreateInstance flag bit.  What do you think of forcing the portability extension and the portability bit on in `VulkanReplayConsumer` if it's running on Mac?  (May be `VulkanReplayConsumerBase`, I forget where the actual extension and feature handling is done)

external/stb/stb_image_write.h
etang-cw: Sounds good

etang-cw:
> I built the tree but of course my first test was to try to replay a vkcube capture from Linux, which failed because the original capture didn't use PORTABILITY ext or CreateInstance flag bit. What do you think of forcing the portability extension and the portability bit on in `VulkanReplayConsumer` if it's running on Mac? (May be `VulkanReplayConsumerBase`, I forget where the actual extension and feature handling is done)

I've been loading MoltenVK directly so I totally forgot that was a thing
Yeah, we should do that

bradgrantham-lunarg:
> CI gfxreconstruct build # 3457 failed.

This is an internal CI failure because we don't (as yet) install `packaging` Python module and you committed a change to import and use that at d51ade2c82bed85f9ffd179de57455ea27da0866.  Otherwise I suspect our internal CI would have succeeded as it did before.  I suggest breaking out the switch to `packaging` from `distutils`to a separate PR in order to merge this since we'd have to request an internal CI change and wait on that resolution otherwise.

bradgrantham-lunarg:
Very nice!  (golf clap)

`tools/replay/gfxrecon-replay --remove-unsupported -m rebind --wsi metal ~/Downloads/vkcube_20220718T185641.gfxr`

Screen Shot 2023-11-06 at 12 56 18 PM

That's from a capture of vkcube on Linux.

etang-cw:
> I suggest breaking out the switch to packaging from distutils to a separate PR in order to merge this since we'd have to request an internal CI change and wait on that resolution otherwise.

It's available at #1331

bradgrantham-lunarg:
> > I suggest breaking out the switch to packaging from distutils to a separate PR in order to merge this since we'd have to request an internal CI change and wait on that resolution otherwise.
> 
> It's available at #1331

Thanks!  And, just to be clear, unless I really misunderstand git, your push of 01fbefa to `macos` still has d51ade2 in its lineage so it still imports and uses `packaging` - did you mean for that to push a version of the branch without that?

etang-cw:
> did you mean for that to push a version of the branch without that?

GitHub Actions' macOS runner has Python 3.12, so it's kind of required to pass that
So either way we'll have at least one CI failing

etang-cw:
I guess I could modify the CI script to use `python3.11`, should I do that?

bradgrantham-lunarg:
> I guess I could modify the CI script to use `python3.11`, should I do that? Edit: Doing that for now

I see that my old incomplete MacOS port (based on @kvark's earlier MacOS port) now fails also.  I suppose if the script can be forced to an earlier version on MacOS runners, we can try to get this merged and I can also start the conversation about getting packaging installed on our CI systems so we can use it and then unlock the version in the script.

framework/util/page_guard_manager.cpp
etang-cw: Now `constexpr`

etang-cw:
Okay.  I removed the parallelize build commit so this no longer conflicts with #1302

(Though whichever gets merged last will still want to be modified to do a parallel macOS build in the CI)

bradgrantham-lunarg:
https://github.com/LunarG/gfxreconstruct/pull/1332 removes the need for `packaging` and `distutils`

.github/workflows/scripts/build-dependencies-macos.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+set -e
+
```
richard-lunarg, line 4: Just a note that the VulkanSDK now sets the macOS deployment target minimum to 11.0. 10.13 is getting rather old.

external/VulkanMemoryAllocator/include/vk_mem_alloc.h
```diff
@@ -5686,8 +5686,7 @@ class VmaJsonWriter
     // Writes a number value.
     void WriteNumber(uint32_t n);
     void WriteNumber(uint64_t n);
-    void WriteSize(uint32_t n);
```
richard-lunarg, line 5689: Are you using a custom version of VMA for this?

etang-cw: No, it's a revert of f7b95927db84c0bc3b6d6ad0f34f74276aa6c412
It doesn't seem like it broke any CIs but you may want to test some 32-bit platforms with it?

.github/workflows/scripts/build-dependencies-macos.sh
etang-cw: Okay, I updated the script

bradgrantham-lunarg:
@etang-cw Thank you very much for this PR and sticking with us through the feedback and requested edits!

PR #1318, opened 2023-10-20
Improve clang arm64 compile time

etang-cw:
Alternative implementation of #1297 (see discussion in #1317)

clang's arm64 register allocator really doesn't like that giant switch statement, and making it codegen the switch statement tons of times as a part of a template function isn't helping.


panos-lunarg (approved):
lgtm

andrew-lunarg (approved):
It looks good to me. Helps Android build times. Helps MacOS build times (and clang builds in general). Probably reduces code size and thus improves performance. Clear win. 

Thank you for spotting this problem and supplying a clean targeted PR to  fix it.

PR #1321, opened 2023-10-26
Cleanup Vulkan enum ToString codegen

### LINKED ISSUE #1158: Remove Unused JSON Formatting Params From generated_vulkan_enum_to_string.h/cpp
We have a bunch of functions hanging aound from the old method of generating JSON which look like this:

```cpp
template <> std::string ToString<VkImageType>(const VkImageType& value, ToStringFlags, uint32_t, uint32_t)
{
    switch (value) {
    case VK_IMAGE_TYPE_1D: return "VK_IMAGE_TYPE_1D";
    case VK_IMAGE_TYPE_2D: return "VK_IMAGE_TYPE_2D";
    case VK_IMAGE_TYPE_3D: return "VK_IMAGE_TYPE_3D";
    default: break;
    }
    return "Unhandled VkImageType";
}
```

They are still useful for log messages but should be cleaned-up like the following and then used (possibly also by the current JSON output which duplicates them):

```cpp
std::string ToString(const VkImageType value)
{
    std::string val;
    switch (value) {
    case VK_IMAGE_TYPE_1D: val = "VK_IMAGE_TYPE_1D";
    case VK_IMAGE_TYPE_2D: val = "VK_IMAGE_TYPE_2D";
    case VK_IMAGE_TYPE_3D: val = "VK_IMAGE_TYPE_3D";
    default: val = "Unhandled VkImageType: " + to_string(value);
    }
    return val;
}
```

See:
framework/generated/generated_vulkan_enum_to_string.h
framework/generated/generated_vulkan_enum_to_string.cpp

charles-lunarg:
Removes redundant parameters from the ToString functions. Make the Vulkan JSON Enum ToString use the regular enum ToString.

Fixes #1158


andrew-lunarg:
Witth this PR landed and #1362 too, much of the remaining ToString code going back to the ToAscii tool may turn dead and be reapable. Check out and maybe other files:

    framework/util/custom_common_to_string.cpp
    framework/util/custom_common_to_string.h
    framework/util/to_string.cpp
    framework/util/to_string.h

PR #1322, opened 2023-10-26
ci: Use lukka/get-cmake 3.16.3 in actions runs & Update Github CI to actions/checkout@v4

### LINKED ISSUE #1283: Add lukka/get-cmake@latest to release_build.yml and sdk_android_build.yml
Just take a moment to make sure this doesn't break anything.

charles-lunarg:
The version of CMake used by NDK is not able to be specified by lukka/get-cmake. Thus the sdk_android_build.yml can't be updated using it.

Fixes #1283

### DIFF
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -33,7 +33,7 @@ jobs:
           }
     steps:
     - name: Clone repository
-      uses: actions/checkout@v1
+      uses: actions/checkout@v4
     - name: Install build dependencies
       run: |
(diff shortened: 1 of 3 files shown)


charles-lunarg:
Well, we can't use 3.16.3 for the windows actions runs because we use the latest visual studio version (2022) on those systems which isn't supported by 3.16. I have made the linux runs use 3.16 while letting the windows runs use latest. This is consistent with other repos due to using ubuntu 20 as our current requirements minimum.

PR #1326, opened 2023-10-30
Fix `OverrideAllocateMemory` for trimmed traces

marius-pelegrin-arm:
When capturing a replay of a capture and `vkAllocateMemory` is called with `VkMemoryOpaqueCaptureAddressAllocateInfo` already in the `pNext` chain, the precedent behavior was to "add" another `VkMemoryOpaqueCaptureAddressAllocateInfo` in front of it.

I think the best behavior is to check for the presence of this struct in the `pNext` chain and, if present, not to touch anything. That is what this commit does.


per-mathisen-arm:
Note that this happens for example when doing trim on a trace with captured addresses, and the validation layers will not like this.

panos-lunarg (approved):
Good catch. Changes look good to me

PR #1328, opened 2023-11-03
Removed Empty Default Cases

andrew-lunarg:
It's good practice to avoid a default case that does nothing for a switch case statement that switches on an enum so that the compiler will complain if one of the enumerants is accidentally missed. By extension it is good practice to always avoid such empty defaults so as not to get in the habit of writing them and to save on lines of code which do nothing.


bradgrantham-lunarg (approved):

andrew-lunarg:
> CI gfxreconstruct build # 3451 failed.

Not a true fail: there was a github outage at the moment I submitted the PR.

bradgrantham-lunarg:
> > CI gfxreconstruct build # 3451 failed.
> 
> Not a true fail: there was a github outage at the moment I submitted the PR.

Have re-run job as #3452

andrew-lunarg:
I saw #3452 and #3453 passed in the CI dashboard even though they hadn't shown up here before I squashed to dev.

PR #1331, opened 2023-11-06
Python 3.12 compatibility

etang-cw:
[Python 3.12 removes distutils](https://peps.python.org/pep-0632/)

### DIFF
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -25,7 +25,7 @@ GFXReconstruct build script
 '''
 
 import argparse
-import distutils.version
+import packaging.version
 import os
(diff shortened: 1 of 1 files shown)


etang-cw:
As noted in #1302, the cmake version check isn't needed anymore, since we require cmake 3.16 in the CMakeLists
The only other usage we have is the `VERSION = distutils.version.StrictVersion('0.0.0')`, whose only use is to get converted back into a string
Would it make sense to drop the dependency altogether?

etang-cw:
No longer needed after #1332

PR #1332, opened 2023-11-07
build: remove requirement on distutils

lunarpapillo:
distutils is deprecated in Python 3.12.  We were only using it for version number checking, and only for "."-separated lists of integers; we don't require the extended version syntax supported by PEP 440.

Simple version checking can be done easily via list comparison of integers, which supports the same "less than" semantics as version comparison, without requiring any external dependency.

build.py:
- use lists of integers to reflect versions being compared

### DIFF
--- a/scripts/build.py
+++ b/scripts/build.py
@@ -25,7 +25,6 @@ GFXReconstruct build script
 '''
 
 import argparse
-import distutils.version
 import os
 import platform
 import re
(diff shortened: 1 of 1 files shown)


bradgrantham-lunarg (approved):

bradgrantham-lunarg:
Waiting on internal CI run

Issue #759, opened 2022-08-11
handle VK_ERROR_OUT_OF_DATE_KHR in Virtual Swapchain when possible


bradgrantham-lunarg:
Since Virtual Swapchain decouples rendering from the swapchain, VSC should be able to easily recreate the replay swapchain, the command buffer used to render to the replay swapchain, and any other resources used to transfer the virtual swapchain image to the "real" swapchain image.

Do this task *after* Virtual Swapchain has been merged.

bradgrantham-lunarg:
#1254 may be fixed by this

per-mathisen-arm:
I did implement this in Lavatube already, but I never found any reliable way to test it. If you know a way, please let me know and I'll see if it works well enough to port it over. (Actually, it would be nice to have a Vulkan testing layer that could deliberately inject error conditions like these...)

PR #1348, opened 2023-11-13
Add replay option with user specified window location

mizhen:
**The problem**
   
Allow user to specified window orgin/starting location and replay GFXR trace in windowed mode without resizing the window resolution (it should be the same size as captured).

**The solution**

Add replay option with user specified window location. The feature is for both D3D12 and Vulkan on Windows platform. It allows user to specify window starting location and force playback in windowed mode with original resolution (instead of full screen).

In addition to the new option, the pull request also includes a bug fix to the following issue:
    --fw option failed to force windowed for full screen resolution. For example, for 4k trace file on 4k screen, the force-windowed option always fails if set --fw 3840,2160, it will always be full-screen mode.
	
**Result**
Tested DX12/Vulkan samples and target title trace file with the build of the pull request, all passed.

### DIFF
--- a/framework/decode/vulkan_virtual_swapchain.cpp
+++ b/framework/decode/vulkan_virtual_swapchain.cpp
@@ -615,8 +615,9 @@ VkResult VulkanVirtualSwapchain::AcquireNextImageKHR(VkResult                  o
     }
 
     result = func(device, swapchain, timeout, semaphore, fence, image_index);
-    if (result != VK_SUCCESS)
+    if ((result != VK_SUCCESS) && (result != VK_SUBOPTIMAL_KHR))
     {
+        // TODO: Add some handling of optimization with VK_SUBOPTIMAL_KHR.
         GFXRECON_LOG_ERROR("Virtual swapchain failed AcquireNextImageKHR 0x%08x for swapchain (ID = %" PRIu64 ")",
                            result,
                            swapchain_info->capture_id);
@@ -640,8 +641,9 @@ VkResult VulkanVirtualSwapchain::AcquireNextImage2KHR(VkResult                  
     }
 
     VkResult result = func(device, acquire_info, image_index);
-    if (result != VK_SUCCESS)
+    if ((result != VK_SUCCESS) && (result != VK_SUBOPTIMAL_KHR))
     {
+        // TODO: Add some handling of optimization with VK_SUBOPTIMAL_KHR.
         GFXRECON_LOG_ERROR("Virtual swapchain failed AcquireNextImage2KHR 0x%08x for swapchain (ID = %" PRIu64 ")",
                            result,
                            swapchain_info->capture_id);
--- a/USAGE_desktop_D3D12.md
+++ b/USAGE_desktop_D3D12.md
@@ -208,6 +208,7 @@ Usage:
                         [--onhb | --omit-null-hardware-buffers]
                         [-m <mode> | --memory-translation <mode>]
                         [--fw <width,height> | --force-windowed <width,height>]
+                        [--fwo <x,y> | --force-windowed-origin <x,y>]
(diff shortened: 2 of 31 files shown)


perim:
How will you deal with multiple windows?  For the patrace project we record the window positions of each created window so that they can be recreated in the same position on replay. This is especially useful for UI apps that are composed of multiple windows side by side of each other. Could that be a solution here as well?

mizhen:
> How will you deal with multiple windows? For the patrace project we record the window positions of each created window so that they can be recreated in the same position on replay. This is especially useful for UI apps that are composed of multiple windows side by side of each other. Could that be a solution here as well?

In current code (with or without the pull request), for supporting the multiple windows which are part of UI apps, current code need some changes such as recording the order, location, size and related style parameters of windows into some new trace block to replace current ResizeWindowCommand or ResizeWindowCommand2 for the window creation and resize during playback time, plus some other related changes. However, I think multiple windows of UI apps handling can be a new feature to implement in future.


As for this pull-request replaces existing fixed window origin 0,0 with user specified location and other related full screen / windowed mode changes, it doesn't append any handling with multiple windows of which the location has different values.

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -608,8 +608,9 @@ VkResult VulkanVirtualSwapchain::AcquireNextImageKHR(VkResult                  o
     }
 
     result = func(device, swapchain, timeout, semaphore, fence, image_index);
-    if (result != VK_SUCCESS)
+    if ((result != VK_SUCCESS) && (result != VK_SUBOPTIMAL_KHR))
```
bradgrantham-lunarg, line 611: I don't think we want to start failing for this result code.  `SUBOPTIMAL` isn't a failure.  If you'd like, you could put in a TODO comment referencing https://github.com/LunarG/gfxreconstruct/issues/759 but take out these tests of `result`.

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_virtual_swapchain.cpp
mizhen: Yes, VK_SUBOPTIMAL_KHR isn't a failure.

If the test is only (result != VK_SUCCESS),  VK_SUBOPTIMAL_KHR value will match it and proceed the following error output:
        GFXRECON_LOG_ERROR("Virtual swapchain failed AcquireNextImageKHR 0x%08x for swapchain (ID = %" PRIu64 ")",
                           result,
                           swapchain_info->capture_id);
                           
That's why the test (result != VK_SUBOPTIMAL_KHR) is added.

bradgrantham-lunarg: Oh!  I read this change as the opposite of what you've actually added.  Yes, this is correct.

locke-lunarg:
Could we combine `force_windowed` and `force_windowed_origin` into one option? or do we prefer to keep them separating?

mizhen:
> LGTM. But, could we combine `force_windowed` and `force_windowed_origin`? or do we prefer to separate?

Currently the function of the two options is different, one is --force-windowed <width,height>, another is --force-windowed-origin <xloc,yloc>. We can combine the two into --force-windowed <width,height,x,y> in the future when virtual swapchain supports both dx12 and vulkan and can scale to specific window size.

locke-lunarg:
I think it's ok to combine them in the future. Please rebase it, and pass the test, and then it will be good to merge.

mizhen:
Will fix the macOS build issues.

mizhen:
> > I think it's ok to combine them in the future. Please rebase it, and pass the test, and then it will be good to merge.
> 
> OK, will do that.

Done.

locke-lunarg (approved):

Issue #1203, opened 2023-07-18
after gfxrecon-convert,the json file can not  fully convert information about vkUpdateDescriptorSetWithTemplateKHR.pData is incomplete


I-dog:
after gfxrecon-convert,the json file can not  fully convert information about vkUpdateDescriptorSetWithTemplateKHR.pData is incomplete

I-dog:
In the JSON file, there are many pDescriptorUpdateEntries in vkCreateDescriptorUpdateTemplateKHR. However, in vkUpdateDescriptorSetWithTemplateKHR, pData only records one buffer information.

####
{
  "index": 22315,
  "function": {
    "name": "vkCreateDescriptorUpdateTemplateKHR",
    "thread": 2,
    "return": "VK_SUCCESS",
    "args": {
      "device": 5,
      "pCreateInfo": {
        "sType": "VK_STRUCTURE_TYPE_DESCRIPTOR_UPDATE_TEMPLATE_CREATE_INFO",
        "flags": "0x00000000",
        "descriptorUpdateEntryCount": 5,
        "pDescriptorUpdateEntries": [
          {
            "dstBinding": 0,
            "dstArrayElement": 0,
            "descriptorCount": 1,
            "descriptorType": "VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC",
            "offset": 0,
            "stride": 24
          },
          {
            "dstBinding": 1,
            "dstArrayElement": 0,
            "descriptorCount": 1,
            "descriptorType": "VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC",
            "offset": 24,
            "stride": 24
          },
          {
            "dstBinding": 2,
            "dstArrayElement": 0,
            "descriptorCount": 1,
            "descriptorType": "VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC",
            "offset": 48,
            "stride": 24
          },
          {
            "dstBinding": 3,
            "dstArrayElement": 0,
            "descriptorCount": 1,
            "descriptorType": "VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC",
            "offset": 72,
            "stride": 24
          },
          {
            "dstBinding": 4,
            "dstArrayElement": 0,
            "descriptorCount": 1,
            "descriptorType": "VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC",
[...]

PR #1350, opened 2023-11-15
Fixes for handling vkUpdateDescriptorSetWithTemplate

naomipappe:
- Fixes a crash when using `vkUpdateDescriptorSetWithTemplate` with acceleration structures, which was caused by allocating new template memory for the acceleration structure information to fit, but forgetting to correct the cached pointers to previously allocated template memory
- Fixes an incomplete information from convert tool for `vkUpdateDescriptorSetWithTemplate`  (#1203)


MarkY-LunarG:
Submitted some fixes to this to support future descriptor templates.  The original code was good, just lacked that future ability.

MarkY-LunarG:
No worries!  I just was trying to clean things up before I pushed.  :-)

PR #1352, opened 2023-11-15
Fix VK_EXT_debug_marker extension not available

marius-pelegrin-arm:
Fix miscellaneous typos, in particular one in trace_layer.cpp that made VK_EXT_debug_marker extension not available if the driver does not support it (even if supported by the layer).


lunarpapillo:
CI failure above was due to a crashed GPU.  I'm resetting and restarting the job.

PR #1353, opened 2023-11-15
Improve decode block error logging

MarkY-LunarG:
When a decode block error occurs, indicate which frame and block are responsible.  This should help at least diagnose if a block is at the begining or end of a trace.

### DIFF
--- a/framework/decode/file_processor.cpp
+++ b/framework/decode/file_processor.cpp
@@ -399,7 +399,10 @@ bool FileProcessor::ProcessBlocks()
                 else
                 {
                     // Unrecognized block type.
-                    GFXRECON_LOG_WARNING("Skipping unrecognized file block with type %u", block_header.type);
+                    GFXRECON_LOG_WARNING("Skipping unrecognized file block with type %u (frame %u block %" PRIu64 ")",
+                                         block_header.type,
+                                         current_frame_number_,
+                                         block_index_);
                     GFXRECON_CHECK_CONVERSION_DATA_LOSS(size_t, block_header.size);
                     success = SkipBytes(static_cast<size_t>(block_header.size));
                 }
@@ -413,7 +416,9 @@ bool FileProcessor::ProcessBlocks()
                     // end of file warning when the file is at EOF without an error. For this case (the normal EOF case)
                     // we print nothing at EOF, or print an error message and set the error code directly when not at
                     // EOF.
-                    GFXRECON_LOG_ERROR("Failed to read block header");
+                    GFXRECON_LOG_ERROR("Failed to read block header (frame %u block %" PRIu64 ")",
+                                       current_frame_number_,
+                                       block_index_);
                     error_state_ = kErrorReadingBlockHeader;
                 }
             }
@@ -507,7 +512,7 @@ void FileProcessor::HandleBlockReadError(Error error_code, const char* error_mes
     }
     else
     {
-        GFXRECON_LOG_ERROR("%s", error_message);
+        GFXRECON_LOG_ERROR("%s (frame %u block %" PRIu64 ")", error_message, current_frame_number_, block_index_);
         error_state_ = error_code;
     }
 }


bradgrantham-lunarg:
This is a great addition!

What do you think about adding the "frame %lu, block %lu" to the end of the error messages?  I'm not sure if Bob's scripts or any scripts in the wild expect to find these existing error messages by searching for "[gfxrecon] WARNING - Failed" or "[gfxrecon] WARNING - Missing".  I also think you could omit the square brackets in that case; parsing out those frame numbers and block numbers should be pretty straightforward.

MarkY-LunarG:
> This is a great addition!
> 
> What do you think about adding the "frame %lu, block %lu" to the end of the error messages? I'm not sure if Bob's scripts or any scripts in the wild expect to find these existing error messages by searching for "[gfxrecon] WARNING - Failed" or "[gfxrecon] WARNING - Missing". I also think you could omit the square brackets in that case; parsing out those frame numbers and block numbers should be pretty straightforward.

I talked to Bob and he said his scripts will handle square brackets or parentheses or whatever.  In fact, he would only look at this manually if an issue arose, he has no plans to automate a script on this.  Although he says he could simply using regular expressions if he needed to.   Therefore, I moved it to the end as you suggested and put them in parentheses instead.  Additionally, I lower-cased "frame" and "block" based on his suggestion.

MarkY-LunarG:
Adding @lunarpapillo just so he can sit in and see his name being thrown around so Cavalierly.

bradgrantham-lunarg (approved):

framework/decode/file_processor.cpp
```diff
@@ -413,7 +416,8 @@ bool FileProcessor::ProcessBlocks()
                     // end of file warning when the file is at EOF without an error. For this case (the normal EOF case)
                     // we print nothing at EOF, or print an error message and set the error code directly when not at
                     // EOF.
-                    GFXRECON_LOG_ERROR("Failed to read block header");
+                    GFXRECON_LOG_ERROR(
+                        "Failed to read block header (frame %lu block %lu)", current_frame_number_, block_index_);
```
andrew-lunarg, line 420: Dangerous 64/32 bit format codes? There is an explicit `uint32_t` and an explicit `uint64_t`, both given the loosy-goosey format code `%lu` (`unsigned long`) which potentially means different things on different compilers/platforms/32/64 bit build versions. At least on some build on some platform we care about, one of `uint32_t` or `uint64_t` is not going to be a `%lu`.

Here is a belt and braces version:

```cpp
GFXRECON_LOG_ERROR("Failed to read block header (frame %u block %llu)", static_cast<unsigned>(current_frame_number_), static_cast<unsigned long long>(block_index_));
```
I don't think you need the casts on 32 bit or 64 bit platforms we care about but they won't hurt.

framework/decode/file_processor.cpp
```diff
@@ -507,7 +511,7 @@ void FileProcessor::HandleBlockReadError(Error error_code, const char* error_mes
     }
     else
     {
-        GFXRECON_LOG_ERROR("%s", error_message);
+        GFXRECON_LOG_ERROR("%s (frame %lu block %lu)", error_message, current_frame_number_, block_index_);
```
andrew-lunarg, line 514: Format codes as per my other comment.

andrew-lunarg: Another reason for stepping up to C++ 20 with its modern text formatting.

framework/decode/file_processor.cpp
```diff
@@ -399,7 +399,10 @@ bool FileProcessor::ProcessBlocks()
                 else
                 {
                     // Unrecognized block type.
-                    GFXRECON_LOG_WARNING("Skipping unrecognized file block with type %u", block_header.type);
+                    GFXRECON_LOG_WARNING("Skipping unrecognized file block with type %u (frame %lu block %lu)",
+                                         block_header.type,
+                                         current_frame_number_,
```
andrew-lunarg, line 404: The `block_header.type` looks fine but the other two as per my other comment.

andrew-lunarg (changes requested):
Late here but I don't think I'm confused. See the comments.

framework/decode/file_processor.cpp
MarkY-LunarG: Frame count is uint32, so I switched to `%u` which should be universal.  Block index was uint64_t so I switched to `%" PRIu64"` which also is universal without having to do any casts.

framework/decode/file_processor.cpp
MarkY-LunarG: Resolved same was as above.

framework/decode/file_processor.cpp
andrew-lunarg: Oh cool, I didn't actually know about those format specifiers. They do exactly what we want (https://www.phind.com/search?cache=llnuuji1691dcyfheuxw3lvs, https://en.cppreference.com/w/cpp/header/cinttypes).

As long as nobody changes the types of these members later, the casts will never be needed.

andrew-lunarg (approved):

andrew-lunarg:
Oh cool, I didn't actually know about those format specifiers. They do exactly what we want (https://www.phind.com/search?cache=llnuuji1691dcyfheuxw3lvs, https://en.cppreference.com/w/cpp/header/cinttypes).

As long as nobody changes the types of these members later, the casts I suggested will never be needed.

PR #1354, opened 2023-11-15
Support for `VK_EXT_frame_boundary` and usage in offscreen swapchain

marius-pelegrin-arm:
This pull request is in fact two PRs with one depending on the other, built in two commits.

## Support for VK_EXT_frame_boundary

Add support for the frame boundary extension `VK_EXT_frame_boundary`. This includes:
- `VkFrameBoundaryEXT` structures found in the pNext chains of `vkSubmitInfo` and `vkSubmitInfo2` are now interpreted as frame delimiters at capture time if `flags` member has the `VK_FRAME_BOUNDARY_FRAME_END_BIT_EXT` bit set.
- The same structures in the same calls support screenshots at replay time by interpreting `pImages` as the rendered frames.
- `VK_EXT_frame_boundary` is now listed as a supported layer device extension (programs that use the layer have access to the extension even if the driver does not support it)

Extended support for this extension that is NOT implemented by this commit could include:
- Interpretation of `VkFrameBoundaryEXT` structure when found in `VkPresentInfoKHR` and `VkBindSparseInfo` (eg. by considering the *real* render frames to be the ones specified by the structures instead of the one presented)
- Support for `VkFrameBoundaryEXT` even when the `VK_FRAME_BOUNDARY_FRAME_END_BIT_EXT` bit is not set (eg. by saving the resources associated in a map/vector to be taken into account when the real frame end is encountered)

This extension should finally normalize what a frame is, what resources are linked to a specific frames etc. Thus it is important for GFXReconstruct to support it for the tracing of applications that do not display graphics to a monitor like offscreen renderers.

## Add `--offscreen-swapchain-frame-boundary` option

This option inserts a `VkFrameBoundaryEXT` from `VK_EXT_frame_boundary` into a command buffer submission called where `vkQueuePresentKHR` should have been called by an offscreen swapchain.

By doing this, it is now possible to capture a trace replayed with the option `--swapchain offscreen` AND to still have frame numbering and the possibility to take screenshots.

This is very important for automated pipelines that run on a server without WSI context and still need to take screenshots, or do
fast forwarding...

### DIFF
--- a/USAGE_android.md
+++ b/USAGE_android.md
@@ -867,6 +867,15 @@ optional arguments:
                         Swap the swapchain color space if unsupported by replay device.
                         Check if color space is not supported by replay device and swap
                         to VK_COLOR_SPACE_SRGB_NONLINEAR_KHR. (forwarded to replay tool).
+  --offscreen-swapchain-frame-boundary
+                        Should only be used with offscreen swapchain. Activates
+                        the extension VK_EXT_frame_boundary (always supported
+                        if trimming, checks for driver support otherwise) and
+                        inserts command buffer submission with
+                        VkFrameBoundaryEXT where vkQueuePresentKHR was called
+                        in the original capture. This allows preserving frames
+                        when capturing a replay that uses. offscreen swapchain.
+                        (forwarded to replay tool)
 ```
 
 The command will force-stop an active replay process before starting the replay
--- a/USAGE_desktop_Vulkan.md
+++ b/USAGE_desktop_Vulkan.md
@@ -540,6 +540,14 @@ Optional arguments:
               Swap the swapchain color space if unsupported by replay device.
               Check if color space is not supported by replay device and
               fallback to VK_COLOR_SPACE_SRGB_NONLINEAR_KHR.
+  --offscreen-swapchain-frame-boundary
+              Should only be used with offscreen swapchain.
+              Activates the extension VK_EXT_frame_boundary (always supported if
+              trimming, checks for driver support otherwise) and inserts command
+              buffer submission with VkFrameBoundaryEXT where vkQueuePresentKHR
+              was called in the original capture.
+              This allows preserving frames when capturing a replay that uses.
+              offscreen swapchain.
 ```
 
 ### Key Controls
--- a/android/scripts/gfxrecon.py
+++ b/android/scripts/gfxrecon.py
@@ -85,6 +85,7 @@ def CreateReplayParser():
     parser.add_argument('--validate', action='store_true', default=False, help='Enables the Khronos Vulkan validation layer (forwarded to replay tool)')
     parser.add_argument('--onhb', '--omit-null-hardware-buffers', action='store_true', default=False, help='Omit Vulkan calls that would pass a NULL AHardwareBuffer* (forwarded to replay tool)')
     parser.add_argument('--use-colorspace-fallback', action='store_true', default=False, help='Swap the swapchain color space if unsupported by replay device. Check if color space is not supported by replay device and swap to VK_COLOR_SPACE_SRGB_NONLINEAR_KHR. (forwarded to replay tool).')
+    parser.add_argument('--offscreen-swapchain-frame-boundary', action='store_true', default=False, help='Should only be used with offscreen swapchain. Activates the extension VK_EXT_frame_boundary (always supported if trimming, checks for driver support otherwise) and inserts command buffer submission with VkFrameBoundaryEXT where vkQueuePresentKHR was called in the original capture. This allows preserving frames when capturing a replay that uses. offscreen swapchain. (forwarded to replay tool)')
     parser.add_argument('--mfr', '--measurement-frame-range', metavar='START-END', help='Custom framerange to measure FPS for. This range will include the start frame but not the end frame. The measurement frame range defaults to all frames except the loading frame but can be configured for any range. If the end frame is past the last frame in the trace it will be clamped to the frame after the last (so in that case the results would include the last frame). (forwarded to replay tool)')
     parser.add_argument('--measurement-file', metavar='DEVICE_FILE', help='Write measurements to a file at the specified path. Default is: \'/sdcard/gfxrecon-measurements.json\' on android and \'./gfxrecon-measurements.json\' on desktop. (forwarded to replay tool)')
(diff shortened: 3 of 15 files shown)


USAGE_android.md
```diff
@@ -863,6 +863,15 @@ optional arguments:
...
                         to VK_COLOR_SPACE_SRGB_NONLINEAR_KHR. (forwarded to replay tool).
+  --offscreen-swapchain-frame-boundary
+                        Should only be used with offscreen swapchain. Activate
+                        the extension VK_EXT_frame_boundary (always supported
+                        if trimming, check for driver support otherwise) and
+                        insert command buffer submission with
+                        VkFrameBoundaryEXT where vkQueuePresentKHR was called
+                        in the original capture. This allows to preserve frames
+                        when capturing a replay that uses. offscreen swapchain.
+                        (forwarded to replay tool)
```
charles-lunarg, line 874:
```suggestion
                        Should only be used with offscreen swapchain. Activates
                        the extension VK_EXT_frame_boundary (always supported
                        if trimming, checks for driver support otherwise) and
                        inserts command buffer submission with
                        VkFrameBoundaryEXT where vkQueuePresentKHR was called
                        in the original capture. This allows preserving frames
                        when capturing a replay that uses offscreen swapchain.
                        (forwarded to replay tool)
```

USAGE_desktop_Vulkan.md
```diff
@@ -536,6 +536,14 @@ Optional arguments:
...
               Check if color space is not supported by replay device and
               fallback to VK_COLOR_SPACE_SRGB_NONLINEAR_KHR.
+  --offscreen-swapchain-frame-boundary
+              Should only be used with offscreen swapchain.
+              Activate the extension VK_EXT_frame_boundary (always supported if
+              trimming, check for driver support otherwise) and insert command
+              buffer submission with VkFrameBoundaryEXT where vkQueuePresentKHR
+              was called in the original capture.
+              This allows to preserve frames when capturing a replay that uses.
+              offscreen swapchain.
```
charles-lunarg, line 546:
```suggestion
              Should only be used with offscreen swapchain.
              Activates the extension VK_EXT_frame_boundary (always supported if
              trimming, checks for driver support otherwise) and inserts command
              buffer submission with VkFrameBoundaryEXT where vkQueuePresentKHR
              was called in the original capture.
              This allows preserving frames when capturing a replay that uses
              offscreen swapchain.
```

android/scripts/gfxrecon.py
```diff
@@ -85,6 +85,7 @@ def CreateReplayParser():
     parser.add_argument('--validate', action='store_true', default=False, help='Enables the Khronos Vulkan validation layer (forwarded to replay tool)')
     parser.add_argument('--onhb', '--omit-null-hardware-buffers', action='store_true', default=False, help='Omit Vulkan calls that would pass a NULL AHardwareBuffer* (forwarded to replay tool)')
     parser.add_argument('--use-colorspace-fallback', action='store_true', default=False, help='Swap the swapchain color space if unsupported by replay device. Check if color space is not supported by replay device and swap to VK_COLOR_SPACE_SRGB_NONLINEAR_KHR. (forwarded to replay tool).')
+    parser.add_argument('--offscreen-swapchain-frame-boundary', action='store_true', default=False, help='Should only be used with offscreen swapchain. Activate the extension VK_EXT_frame_boundary (always supported if trimming, check for driver support otherwise) and insert command buffer submission with VkFrameBoundaryEXT where vkQueuePresentKHR was called in the original capture. This allows to preserve frames when capturing a replay that uses. offscreen swapchain. (forwarded to replay tool)')
```
charles-lunarg, line 88:
```suggestion
    parser.add_argument('--offscreen-swapchain-frame-boundary', action='store_true', default=False, help='Should only be used with offscreen swapchain. Activates the extension VK_EXT_frame_boundary (always supported if trimming, checks for driver support otherwise) and inserts command buffer submission with VkFrameBoundaryEXT where vkQueuePresentKHR was called in the original capture. This allows preserving frames when capturing a replay that uses offscreen swapchain. (forwarded to replay tool)')
```

framework/decode/vulkan_offscreen_swapchain.cpp
```diff
@@ -222,7 +276,36 @@ VkResult VulkanOffscreenSwapchain::QueuePresentKHR(VkResult
                                                    const QueueInfo*                      queue_info,
                                                    const VkPresentInfoKHR*               present_info)
 {
-    if (present_info->waitSemaphoreCount > 0)
+    if (insert_frame_boundary_ && command_buffer_ != VK_NULL_HANDLE)
```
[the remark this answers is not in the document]
marius-pelegrin-arm: Yes, but the command buffer creation can fail. If it is the case (and that we're in release mode, because there are asserts) then it's better to just not randomly crash in a segfault here and have the insertion ignored...

framework/decode/vulkan_offscreen_swapchain.cpp
```diff
@@ -103,38 +103,49 @@ VkResult VulkanOffscreenSwapchain::CreateSwapchainKHR(VkResult
...
-        commandBufferBeginInfo.flags            = 0;
-        commandBufferBeginInfo.pInheritanceInfo = nullptr;
-
-        result = device_table_->BeginCommandBuffer(command_buffer_, &commandBufferBeginInfo);
-        GFXRECON_ASSERT(result == VK_SUCCESS);
-
-        result = device_table_->EndCommandBuffer(command_buffer_);
-        GFXRECON_ASSERT(result == VK_SUCCESS);
-
+        for (uint32_t family_index = 0; family_index < device_info->queue_family_index_enabled.size(); family_index++)
```
MarkY-LunarG, line 109: Just add a comment above this explaining why your doing this so people can understand what's going on.  Just something similar to the first sentence of your commit is fine.

MarkY-LunarG (approved):

framework/decode/vulkan_offscreen_swapchain.cpp
marius-pelegrin-arm: Done =)

charles-lunarg (approved):
I added a fix for the one issue I saw (command buffer/pool's queue family must match the queue it is being submitted on, easiest solution is to create a command buffer/pool per queue family to be able to submit on any queue).

So this looks ready to go in, the local testing I've done shown it works as expected.

PR #1356, opened 2023-11-15
Testing: Dump DX12 resources for drawcall

locke-lunarg:
Test sample: D3D12HelloTexture. It's for renderpass test. I added renderpass, constant buffer, more triangles, and test code about copy resource and separated renderpass.
https://github.com/locke-lunarg/DirectX-Graphics-Samples/tree/test-dump

For not renderpass case, now, I tested  HelloWorld project in DirectX-Graphics-Samples. I'm doing more tests.

Add option: `--dump-resources drawcall-n` . n is the index of drawcall.
To output readable data. Set `TEST_READABLE = true` in  `dx12_dump_resources.cpp`.
Float for vertex, index, constant buffer.
Image for shader resource, render target.

The basic concept:
`Not renderpass`: Add copy resource command in before and after drawcall.
E.g.:
Original
```
cmd->Reset();
...
cmd->Draw();
...
cmd->Draw();  -> Target
...
cmd->End();
```
New
```
cmd->Reset();
...
cmd->Draw();
...
cmd->CopyResource(); -> Add
cmd->draw();  -> Target
cmd->CopyResource(); -> Add
...
cmd->End();
```
`Renderpass`: Add renderpasses in before and after drawcall.
E.g.:
Original
```
cmd->Reset();
...
cmd->BeginRenderPass();
...
cmd->Draw();
...
cmd->Draw();  -> Target
...
cmd->EndRenderPass();
...
cmd->End();
```
New
```
cmd->Reset();
...
cmd->BeginRenderPass();
...
cmd->Draw();
...
cmd->EndRenderPass();    -> Add
cmd->CopyResource();     -> Add
cmd->BeginRenderPass();  -> Add
cmd->Draw();  -> Target
cmd->EndRenderPass();    -> Add
cmd->CopyResource();     -> Add
cmd->BeginRenderPass();  -> Add
...
cmd->EndRenderPass();
...
cmd->End();
```

According to my test, it is ok to insert the renderpass in anyplace in the original renderpass. And then, the EndingAcccess of the renderpass and the BeginningAccess of following renderpasses have to be PRESERVE to keep the the rendering process for this following renderpass, or it will clear the render target.

### DIFF
--- a/framework/decode/custom_dx12_replay_commands.h
+++ b/framework/decode/custom_dx12_replay_commands.h
@@ -45,6 +45,66 @@ struct CustomReplayPostCall
 };
 
 template <>
+struct CustomReplayPreCall<format::ApiCallId::ApiCall_ID3D12GraphicsCommandList_ResourceBarrier>
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase* replay, Args... args)
+    {
+        replay->PreCall_ID3D12GraphicsCommandList_ResourceBarrier(args...);
+    }
+};
+
+template <>
+struct CustomReplayPreCall<format::ApiCallId::ApiCall_ID3D12Device_CreateConstantBufferView>
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase* replay, Args... args)
+    {
+        replay->PreCall_ID3D12Device_CreateConstantBufferView(args...);
+    }
+};
+
+template <>
+struct CustomReplayPreCall<format::ApiCallId::ApiCall_ID3D12Device_CreateShaderResourceView>
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase* replay, Args... args)
+    {
+        replay->PreCall_ID3D12Device_CreateShaderResourceView(args...);
+    }
+};
+
+template <>
+struct CustomReplayPostCall<format::ApiCallId::ApiCall_ID3D12Device_CreateRenderTargetView>
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase* replay, Args... args)
+    {
+        replay->PostCall_ID3D12Device_CreateRenderTargetView(args...);
+    }
+};
+
+template <>
+struct CustomReplayPostCall<format::ApiCallId::ApiCall_ID3D12Device_CreateDepthStencilView>
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase* replay, Args... args)
+    {
+        replay->PostCall_ID3D12Device_CreateDepthStencilView(args...);
+    }
+};
+
+template <>
+struct CustomReplayPostCall<format::ApiCallId::ApiCall_ID3D12GraphicsCommandList_OMSetRenderTargets>
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase* replay, Args... args)
+    {
+        replay->PostCall_ID3D12GraphicsCommandList_OMSetRenderTargets(args...);
+    }
+};
+
+template <>
 struct CustomReplayPreCall<format::ApiCallId::ApiCall_ID3D12GraphicsCommandList_DrawInstanced>
 {
     template <typename... Args>
--- a/framework/decode/custom_dx12_replay_commands.h
+++ b/framework/decode/custom_dx12_replay_commands.h
@@ -145,6 +145,66 @@ struct CustomReplayPostCall<format::ApiCallId::ApiCall_ID3D12GraphicsCommandList
 };
 
 template <>
+struct CustomReplayPreCall<format::ApiCallId::ApiCall_ID3D12GraphicsCommandList_DrawIndexedInstanced>
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase* replay, Args... args)
+    {
+        replay->PreCall_ID3D12GraphicsCommandList_DrawIndexedInstanced(args...);
+    }
+};
+
+template <>
+struct CustomReplayPostCall<format::ApiCallId::ApiCall_ID3D12GraphicsCommandList_DrawIndexedInstanced>
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase* replay, Args... args)
+    {
+        replay->PostCall_ID3D12GraphicsCommandList_DrawIndexedInstanced(args...);
+    }
+};
+
+template <>
+struct CustomReplayPreCall<format::ApiCallId::ApiCall_ID3D12GraphicsCommandList_Dispatch>
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase* replay, Args... args)
+    {
+        replay->PreCall_ID3D12GraphicsCommandList_Dispatch(args...);
+    }
+};
+
+template <>
+struct CustomReplayPostCall<format::ApiCallId::ApiCall_ID3D12GraphicsCommandList_Dispatch>
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase* replay, Args... args)
+    {
+        replay->PostCall_ID3D12GraphicsCommandList_Dispatch(args...);
+    }
+};
+
+template <>
+struct CustomReplayPreCall<format::ApiCallId::ApiCall_ID3D12GraphicsCommandList_ExecuteIndirect>
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase* replay, Args... args)
+    {
+        replay->PreCall_ID3D12GraphicsCommandList_ExecuteIndirect(args...);
+    }
+};
+
+template <>
+struct CustomReplayPostCall<format::ApiCallId::ApiCall_ID3D12GraphicsCommandList_ExecuteIndirect>
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase* replay, Args... args)
+    {
+        replay->PostCall_ID3D12GraphicsCommandList_ExecuteIndirect(args...);
+    }
+};
+
+template <>
 struct CustomReplayPostCall<format::ApiCallId::ApiCall_ID3D12CommandQueue_ExecuteCommandLists>
 {
     template <typename... Args>
--- a/framework/decode/custom_dx12_replay_commands.h
+++ b/framework/decode/custom_dx12_replay_commands.h
@@ -105,6 +105,26 @@ struct CustomReplayPostCall<format::ApiCallId::ApiCall_ID3D12GraphicsCommandList
 };
 
 template <>
+struct CustomReplayPreCall<format::ApiCallId::ApiCall_ID3D12GraphicsCommandList4_BeginRenderPass>
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase* replay, Args... args)
+    {
+        replay->PreCall_ID3D12GraphicsCommandList4_BeginRenderPass(args...);
+    }
+};
+
+template <>
+struct CustomReplayPostCall<format::ApiCallId::ApiCall_ID3D12GraphicsCommandList4_BeginRenderPass>
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase* replay, Args... args)
+    {
+        replay->PostCall_ID3D12GraphicsCommandList4_BeginRenderPass(args...);
+    }
+};
+
+template <>
 struct CustomReplayPreCall<format::ApiCallId::ApiCall_ID3D12GraphicsCommandList_DrawInstanced>
 {
     template <typename... Args>
--- a/framework/decode/custom_dx12_replay_commands.h
+++ b/framework/decode/custom_dx12_replay_commands.h
@@ -65,12 +65,22 @@ struct CustomReplayPreCall<format::ApiCallId::ApiCall_ID3D12Device_CreateConstan
 };
 
 template <>
-struct CustomReplayPreCall<format::ApiCallId::ApiCall_ID3D12Device_CreateShaderResourceView>
+struct CustomReplayPostCall<format::ApiCallId::ApiCall_ID3D12Device_CreateConstantBufferView>
 {
     template <typename... Args>
     static void Dispatch(Dx12ReplayConsumerBase* replay, Args... args)
     {
-        replay->PreCall_ID3D12Device_CreateShaderResourceView(args...);
+        replay->PostCall_ID3D12Device_CreateConstantBufferView(args...);
+    }
+};
+
+template <>
+struct CustomReplayPostCall<format::ApiCallId::ApiCall_ID3D12Device_CreateShaderResourceView>
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase* replay, Args... args)
+    {
+        replay->PostCall_ID3D12Device_CreateShaderResourceView(args...);
     }
 };
 
--- a/framework/decode/custom_dx12_replay_commands.h
+++ b/framework/decode/custom_dx12_replay_commands.h
@@ -1,0 +1,80 @@
+/*
+** Copyright (c) 2023 LunarG, Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and associated documentation files (the "Software"),
+** to deal in the Software without restriction, including without limitation
+** the rights to use, copy, modify, merge, publish, distribute, sublicense,
+** and/or sell copies of the Software, and to permit persons to whom the
+** Software is furnished to do so, subject to the following conditions:
+**
+** The above copyright notice and this permission notice shall be included in
+** all copies or substantial portions of the Software.
+**
+** THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+** IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+** FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+** AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+** FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+** DEALINGS IN THE SOFTWARE.
+*/
+
+#ifndef GFXRECON_ENCODE_CUSTOM_DX12_REPLAY_COMMANDS_H
+#define GFXRECON_ENCODE_CUSTOM_DX12_REPLAY_COMMANDS_H
+
+#include "format/api_call_id.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(decode)
+
+template <format::ApiCallId Id>
+struct CustomReplayPreCall
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase*, Args...)
+    {}
+};
+
+template <format::ApiCallId Id>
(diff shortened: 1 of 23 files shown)


framework/graphics/dx12_dump_resources.h
```diff
@@ -0,0 +1,162 @@
...
+    D3D12_RENDER_PASS_ENDING_ACCESS              record_depth_ending_access{};
+    D3D12_RENDER_PASS_ENDING_ACCESS              record_stencil_ending_access{};
+    D3D12_RENDER_PASS_FLAGS                      record_render_pass_flags{ D3D12_RENDER_PASS_FLAG_NONE };
+
+    ~TrackDumpResources() {}
+};
+
+// TODO: This class copys a lot of code to write json from VulkanExportJsonConsumerBase.
+//       We might need a class for writing json.
+class Dx12DumpResources
```
locke-lunarg, line 98: `Dx12DumpResources` output json and resources. A lot of json code are copied from `VulkanExportJsonConsumerBase`. We should have a common class for them.

framework/decode/dx12_browse_consumer.h
```diff
@@ -0,0 +1,272 @@
...
+        begin_renderpass_code_index = 0;
+        drawcall_code_index         = 0;
+        execute_code_index          = 0;
+        captured_vertex_buffer_view_gvas.clear();
+        captured_index_buffer_view_gva = decode::kNullGpuAddress;
+        descriptor_heap_ids.clear();
+    }
+};
+
+class Dx12BrowseConsumer : public Dx12Consumer
```
locke-lunarg, line 63: `Dx12BrowseConsumer` browses the file before replay to find the target drawcall code index and some key code indices, and track resource infos. That could help replay easier. But it doesn't track render target infos because the useful info is replay cpuDescriptor. It's only available in replay.

framework/decode/custom_dx12_replay_commands.h
```diff
@@ -0,0 +1,160 @@
...
+    {}
+};
+
+template <format::ApiCallId Id>
+struct CustomReplayPostCall
+{
+    template <typename... Args>
+    static void Dispatch(Dx12ReplayConsumerBase*, Args...)
+    {}
+};
```
locke-lunarg, line 45: Add `CustomReplayPreCall` and `CustomReplayPostCall` in every function call to track resources and add copy resource cmd.

framework/decode/dx12_object_info.h
```diff
@@ -303,7 +310,8 @@ struct D3D12ResourceInfo : DxObjectExtraInfo
     std::map<uint64_t, uint64_t>                       mapped_gpu_addresses;
     std::map<uint64_t, graphics::Dx12ShaderIdentifier> mapped_shader_ids;
 
-    D3D12_RESOURCE_DESC1 desc = {};
+    D3D12_RESOURCE_DESC1  desc = {};
+    D3D12_RESOURCE_STATES current_state{ D3D12_RESOURCE_STATE_COMMON };
```
locke-lunarg, line 314: It needs to track the current state of resources for change state for copy resources.

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -1204,6 +1214,17 @@ HRESULT Dx12ReplayConsumerBase::OverrideCreateCommittedResource(
...
 
+    D3D12_RESOURCE_STATES modified_initial_resource_state = InitialResourceState;
+    if ((options_.dump_resources_type != DumpResourcesType::kNone) &&
+        (modified_initial_resource_state & D3D12_RESOURCE_STATE_ALL_SHADER_RESOURCE))
+    {
+        // shader resources can't change state after SetGraphicsRootDescriptorTable.
+        // But even if it changes state and copy resources before SetGraphicsRootDescriptorTable,
+        // it can't solve it, because it also has to copy resources after drawcall.
+        // In this case, adding a COPY_SOURCE state here, so it doesn't need to change state for copy resources.
+        modified_initial_resource_state |= D3D12_RESOURCE_STATE_COPY_SOURCE;
```
locke-lunarg, line 1225: shader resources can't change state after SetGraphicsRootDescriptorTable. In this case, adding a COPY_SOURCE state for shader resource, so it doesn't need to change state for copy resources.

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -3771,5 +3801,577 @@ std::wstring Dx12ReplayConsumerBase::ConstructObjectName(format::HandleId captur
...
+
+        if (options_.dump_resources_type != DumpResourcesType::kNone)
+        {
+            // This should be in the override functionsince it modifies the parameters. But here is ok.
+            barriers[i].Transition->decoded_value->StateBefore = resource_extra_info->current_state;
+            if (barriers[i].Transition->decoded_value->StateAfter & D3D12_RESOURCE_STATE_ALL_SHADER_RESOURCE)
+            {
+                // shader resources can't change state after SetGraphicsRootDescriptorTable.
+                // Adding a COPY_SOURCE state here, so it doesn't need to change state for copy resources.
+                barriers[i].Transition->decoded_value->StateAfter |= D3D12_RESOURCE_STATE_COPY_SOURCE;
```
locke-lunarg, line 3825: shader resources can't change state after SetGraphicsRootDescriptorTable. In this case, adding a COPY_SOURCE state for shader resource, so it doesn't need to change state for copy resources.

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -3771,5 +3801,577 @@ std::wstring Dx12ReplayConsumerBase::ConstructObjectName(format::HandleId captur
...
+        auto                              queue  = reinterpret_cast<ID3D12CommandQueue*>(object_info->object);
+        auto                              device = graphics::dx12::GetDeviceComPtrFromChild<ID3D12Device>(queue);
+        graphics::dx12::ID3D12FenceComPtr fence;
+        auto                              hr = device->CreateFence(0, D3D12_FENCE_FLAG_NONE, IID_PPV_ARGS(&fence));
+        GFXRECON_ASSERT(SUCCEEDED(hr));
+
+        hr = graphics::dx12::WaitForQueue(queue, fence, UINT64_MAX);
+        GFXRECON_ASSERT(SUCCEEDED(hr));
+
+        dump_resources_->WriteResources(track_dump_resources_);
```
locke-lunarg, line 4100: After ExecuteCommandLists and WaitForQueue, output the resources.

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -3771,5 +3801,577 @@ std::wstring Dx12ReplayConsumerBase::ConstructObjectName(format::HandleId captur
...
+                                                                              UINT               StartInstanceLocation)
+{
+    if (options_.dump_resources_type != DumpResourcesType::kNone &&
+        track_dump_resources_.target.drawcall_code_index == call_info.index)
+    {
+        auto commandlist = reinterpret_cast<ID3D12GraphicsCommandList*>(object_info->object);
+        if (track_dump_resources_.target.begin_renderpass_code_index == 0)
+        {
+            // The program doesn't use renderpass.
+            AddCopyResourceCommandsForAfterDrawcall(commandlist);
```
locke-lunarg, line 4041: After adding copy resource, it still has to run the original ExecuteCommandLists, not a new ExecuteCommandLists because some FIllMemoryCommand runs before the original ExecuteCommandLists. If it runs a new ExecuteCommandLists, it might miss FIllMemoryCommand.

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -3771,5 +3801,577 @@ std::wstring Dx12ReplayConsumerBase::ConstructObjectName(format::HandleId captur
...
+        for (uint32_t i = 0; i < NumRenderTargets; ++i)
+        {
+            rt_descs[i].EndingAccess->decoded_value->Type = D3D12_RENDER_PASS_ENDING_ACCESS_TYPE_PRESERVE;
+        }
+
+        auto ds_desc = pDepthStencil->GetMetaStructPointer();
+        if (ds_desc)
+        {
+            ds_desc->DepthEndingAccess->decoded_value->Type   = D3D12_RENDER_PASS_ENDING_ACCESS_TYPE_PRESERVE;
+            ds_desc->StencilEndingAccess->decoded_value->Type = D3D12_RENDER_PASS_ENDING_ACCESS_TYPE_PRESERVE;
```
locke-lunarg, line 3936: The EndingAcccess of the renderpass and the BeginningAccess of following renderpasses have to be PRESERVE to keep the the render process for this following renderpass, or it will clear the render target.

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -3771,5 +3801,577 @@ std::wstring Dx12ReplayConsumerBase::ConstructObjectName(format::HandleId captur
...
+            D3D12_RENDER_PASS_DEPTH_STENCIL_DESC* p_ds_desc = nullptr;
+            D3D12_RENDER_PASS_DEPTH_STENCIL_DESC  ds_desc{};
+            if (track_dump_resources_.depth_stencil_heap_id != 0)
+            {
+                p_ds_desc                           = &ds_desc;
+                ds_desc.cpuDescriptor               = track_dump_resources_.replay_depth_stencil_handle;
+                ds_desc.DepthBeginningAccess.Type   = D3D12_RENDER_PASS_BEGINNING_ACCESS_TYPE_PRESERVE;
+                ds_desc.StencilBeginningAccess.Type = D3D12_RENDER_PASS_BEGINNING_ACCESS_TYPE_PRESERVE;
+                ds_desc.DepthEndingAccess.Type      = D3D12_RENDER_PASS_ENDING_ACCESS_TYPE_PRESERVE;
+                ds_desc.StencilEndingAccess.Type    = D3D12_RENDER_PASS_ENDING_ACCESS_TYPE_PRESERVE;
```
locke-lunarg, line 4019: Here is to insert renderpass in before drawcall. The BeginningAccess an EndingAcccess of the renderpass have to be PRESERVE to keep the the render process for this following renderpass, or it will clear the render target.

andrew-lunarg:
Is it clean with the debug layer?
https://learn.microsoft.com/en-us/windows/win32/direct3d12/understanding-the-d3d12-debug-layer

RE: https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist-copyresource
> The debug layer issues an error if the source subresource is not in the [D3D12_RESOURCE_STATE_COPY_SOURCE](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_resource_states) state.
> The debug layer issues an error if the destination subresource is not in the [D3D12_RESOURCE_STATE_COPY_DEST ](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_resource_states)state.

andrew-lunarg:
The driver for a tiled GPU would have to do a lot of work behind the scenes to make this work. I'm not sure if that is relevant with D3D12 though. Presumably the Adrenos in Qualcomm ARM64 Windows 11 / D3D12 are either never in tiling mode on Windows or not in tiling mode in this situation but might be worth testing on the Windows ARM Devkit (Volterra). I don't know if any of the Chinese IHVs licensing Imagination tiling GPUs have D3D12 drivers.

Something to bear in mind if issue come in one day on odd hardware.

andrew-lunarg:
Can you hold off on the JSON changes? I have a branch coming in that touches the same file in addressing D3D12 JSON: https://github.com/andrew-lunarg/gfxreconstruct/tree/andy-enhance-better-d3d12-convert

rurra-amd (commented):
Any thoughts on doing this as part of `gfxrecon-extract` ?

framework/graphics/dx12_dump_resources.cpp
```diff
@@ -0,0 +1,365 @@
...
+#include "graphics/dx12_util.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(graphics)
+
+// TEST_READABLE is only for test because data type could be various.
+// But here uses fixed type.
+// float for vertex, index, constant buffer.
+// image for shader resource, render target.
+const bool TEST_READABLE = false;
```
locke-lunarg, line 38: To output readable data. Set TEST_READABLE = true.
Float for vertex, index, constant buffer.
Image for shader resource, render target.

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -3771,5 +3801,577 @@ std::wstring Dx12ReplayConsumerBase::ConstructObjectName(format::HandleId captur
...
+                                                                             UINT               StartInstanceLocation)
+{
+    if (options_.dump_resources_type != DumpResourcesType::kNone &&
+        track_dump_resources_.target.drawcall_code_index == call_info.index)
+    {
+        auto commandlist = reinterpret_cast<ID3D12GraphicsCommandList*>(object_info->object);
+        if (track_dump_resources_.target.begin_renderpass_code_index == 0)
+        {
+            // The program doesn't use renderpass.
+            AddCopyResourceCommandsForBeforeDrawcall(commandlist);
```
locke-lunarg, line 3989: Add commands for before drawcall.

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -3771,5 +3801,577 @@ std::wstring Dx12ReplayConsumerBase::ConstructObjectName(format::HandleId captur
...
+            D3D12_RENDER_PASS_DEPTH_STENCIL_DESC* p_ds_desc = nullptr;
+            D3D12_RENDER_PASS_DEPTH_STENCIL_DESC  ds_desc{};
+            if (track_dump_resources_.depth_stencil_heap_id != 0)
+            {
+                p_ds_desc                           = &ds_desc;
+                ds_desc.cpuDescriptor               = track_dump_resources_.replay_depth_stencil_handle;
+                ds_desc.DepthBeginningAccess.Type   = D3D12_RENDER_PASS_BEGINNING_ACCESS_TYPE_PRESERVE;
+                ds_desc.StencilBeginningAccess.Type = D3D12_RENDER_PASS_BEGINNING_ACCESS_TYPE_PRESERVE;
+                ds_desc.DepthEndingAccess           = track_dump_resources_.record_depth_ending_access;
+                ds_desc.StencilEndingAccess         = track_dump_resources_.record_stencil_ending_access;
```
locke-lunarg, line 4071: Here is to insert renderpass in after drawcall. The BeginningAccess have to be PRESERVE to keep the the render process from the previous renderpass, or it will clear the render target. The EndingAccess have to be the same to the original renderpass.

davidd-lunarg:
@rurra-amd 

> Any thoughts on doing this as part of `gfxrecon-extract` ?

Full replay is required so that the GPU resource data is in the required state. `gfxrecon-extract` does not replay any GPU commands, it only has a limited consumer that processes the data for certain captured API calls.

locke-lunarg:
When the commadlist is created with D3D12_COMMAND_LIST_TYPE_BUNDLE, it could not use CopyResources. Fixing...

framework/decode/dx12_browse_consumer.h
```diff
@@ -0,0 +1,353 @@
...
+        }
+    }
+
+    virtual bool IsComplete(uint64_t block_index) override
+    {
+        return (block_index > kMaxDX12BlockLimit) || (target_command_list_ != format::kNullHandleId);
+    }
+
+  private:
+    static int const    kMaxDX12BlockLimit = 1000;
```
davidd-lunarg, line 304: This will limit the `Dx12BrowseConsumer` to processing only the first 1000 GFXR blocks so it may not allow for dumping resources for draw calls that don't happen at the front of the file. There should probably not be a block limit when computing `IsComplete` for this consumer.

locke-lunarg: Thank you. Fixed.

locke-lunarg:
Now, for `CopyResources`, it copys the whole ID3D12Resource. The required data could be a piece of the ID3D12Resource, not the whole ID3D12Resource. We need to handle resource views' offset, byte stride, count, ...

locke-lunarg:
Update:
1. more supported
    1. Support DrawIndexedInstanced, Dispatch, ExecuteIndirect
    2. Initialize Tracking in CreateCommandList.
    3. Add resource_id in .json

2. Change dx12 dump resources option
    The arguments becomes three indices, submit index, command index,
    drawcall index. The command index is based on its in
    ExecuteCommandLists. The target is the drawcall index of the command
    index of the submit index.

3. Add offset and size for CopyResousrce
    Add offset and size for buffer copy for vertex, index, const buffer,
    and shader resource. But textures copy for shader resource, render target,
    depth stencil don't work.
    
    Add descriptor handles check to dump the set descriptor handles,
    instead of all heaps. See MatchDescriptorCPUGPUHandle.

4. Add test flags for dx12 dump
   1. To output readable data. Set "TEST_READABLE = true" in dx12_dump_resources.cpp.
       This feature is only for test because data type could be various.
       But here uses fixed type.
       float for vertex, index, constant buffer.
       image for shader resource, render target.

   2. To skip dump shader resources, Set "TEST_SHADER_RES = false" in dx12_replay_consumer_base.cpp.
       Now, dump shader resources could cost too much memory to cause crash.

locke-lunarg:
A issue is about output shader resource's images. It happens the size is wrong, so it couldn't output the image. I guess the reason is the format is DXGI_FORMAT_BC1_UNORM. `WriteBmpImage` doesn't support it.

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -3771,5 +3801,789 @@ std::wstring Dx12ReplayConsumerBase::ConstructObjectName(format::HandleId captur
...
+    if (call_info.index == track_dump_resources_.target.set_render_targets_code_index)
+    {
+        auto rt_handles = pRenderTargetDescriptors->GetMetaStructPointer();
+        track_dump_resources_.render_target_heap_ids.resize(NumRenderTargetDescriptors);
+        track_dump_resources_.replay_render_target_handles.resize(NumRenderTargetDescriptors);
+        for (uint32_t i = 0; i < NumRenderTargetDescriptors; ++i)
+        {
+            track_dump_resources_.render_target_heap_ids[i]       = rt_handles[i].heap_id;
+            track_dump_resources_.replay_render_target_handles[i] = *rt_handles[i].decoded_value;
+        }
```
davidd-lunarg, line 3966: If `RTsSingleHandleToDescriptorRange == true` and `NumRenderTargetDescriptors > 1` then `pRenderTargetDescriptors` will only point the the first descriptor and will only have one valid meta struct. The other descriptors are contiguous in the descriptor heap. Here is a rough draft of a possible solution:

```
auto descriptor_increment = 0;
if (RTsSingleHandleToDescriptorRange)
{
    // Determine descriptor increment for contiguous render target descriptors.
    auto cmd_list = reinterpret_cast<ID3D12GraphicsCommandList*>(object_info->object);
    auto device   = graphics::dx12::GetDeviceComPtrFromChild<ID3D12Device>(cmd_list);
    if (device != nullptr)
    {
        descriptor_increment = device->GetDescriptorHandleIncrementSize(D3D12_DESCRIPTOR_HEAP_TYPE_RTV);
    }
    else
    {
        // ERROR
    }
}

for (uint32_t i = 0; i < NumRenderTargetDescriptors; ++i)
{
    if (RTsSingleHandleToDescriptorRange)
    {
        // All render target descriptors are contiguous in the same descriptor heap.
        track_dump_resources_.render_target_heap_ids[i] = rt_handles[0].heap_id;
        track_dump_resources_.replay_render_target_handles[i] =
            D3D12_CPU_DESCRIPTOR_HANDLE{ (*rt_handles[0].decoded_value).ptr + descriptor_increment };
    }
    else
    {
        track_dump_resources_.render_target_heap_ids[i]       = rt_handles[i].heap_id;
        track_dump_resources_.replay_render_target_handles[i] = *rt_handles[i].decoded_value;
    }
}
```

Some references:
- Notes on `RTsSingleHandleToDescriptorRange` here: https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist-omsetrendertargets
- Encode function which only encodes one descriptor if `RTsSingleHandleToDescriptorRange == true`: https://github.com/LunarG/gfxreconstruct/blob/278ceb7a7be02f7eeb928ff43af80f0674fc0088/framework/generated/generated_dx12_api_call_encoders.cpp#L5829

locke-lunarg: Updated

PR #1360, opened 2023-11-17
Add AGS to convert tool, rebased.

feimao-gfxr:
**Problem**
1. AGS support is already in GFXR. The convert output file should also include the AGS calls.
2. The index values in the convert output file doesn't match anything in the replay consumer. It makes it very hard to find the matching call in the convert file when debugging an issue in the replayer.

**Solution**
1. Add AGS ascii consumer to support AGS in the convert tool. 
2. Remove api_call_count_ from the convert consumer and use block index for index values in the convert output file. Block index values provides consistent matches between different decode consumers. 

**Result**
AGS calls are included in the convert output file. Indices are block index values now.


feimao-gfxr:
This PR is to replay PR #1211.

Issue #1269, opened 2023-09-21
Issue when copying pNext chains in VulkanStateTracker


marius-pelegrin-arm:
## Issue

The issue happens when you trim a capture that contains certain kind of Vulkan calls and then try to replay the trimmed capture.

### Description

When saving a `pNext` structure in a wrapper in the `VulkanStateTracker`, the copy of the structure does not copy resources accessed through a pointer.

For example, in `VulkanStateTracker::TrackImageMemoryBinding`, if a `VkBindImageMemoryDeviceGroupInfo` structure exists in the `pNext` chain of `bind_info_pnext`, the pointer `VkBindImageMemoryDeviceGroupInfo::pDeviceIndices` will be copied as-is in `wrapper->bind_pnext`.

If the underlying resource is freed by the application before the call to `VulkanStateWrite::WriteState`, this results in undefined behavior, ranging from data corruption to crash when replaying the capture.

### Reproducing this issue

Take an application that regularly calls a vulkan command that takes a `pNext` chain structure as an argument and where you can easily insert a structure containing a pointer (other than `pNext`) owned by the application in the `pNext` chain. For example, take an application that calls `vkBindImageMemory2` at each frame with a `VkBindImageMemoryDeviceGroupInfo` containing a device index in the `pNext` chain of `pBindInfos`.

Capture this application using GFXRECON_CAPTURE_FRAMES starting after a big number of calls (let's say frame 100).

Replay the capture obtained. You will most probably obtain a crash.

charles-lunarg:
Closing as #1363 is now merged

PR #1363, opened 2023-11-20
Fix `pNext` chains unproperly copied in `VulkanStateTracker` (Issue #1269)

marius-pelegrin-arm:
This commit fixes the issue #1269

The first idea was to modify the `UnwrapStruct*Handles` functions to allow more flexibility on what is unwrapped (Vulkan handles, application owned pointers or both).

But I couldn't find a way to properly modify these functions without either altering their behavior, their performance, or having to dig into deeper fundamental changes.

So instead, this commit simply add two new generated files, `generated_vulkan_struct_trackers` header and body.

The idea is to implement a function `TrackStruct` and its overload over all Vulkan structs, that deep-copy the structs by also copying application owned dynamically allocated memory.
This way, the copy returned by `TrackStruct` can be kept indefinitely without undefined behavior, and not just in the time of a call.

All the other file changes are to integrate this new call.


charles-lunarg:
Haven't reviewed the code at all, but at first glance this is the same concept as the 'safe structs' in Vulkan-ValidationLayers. I previously considered moving the safe structs codegen from validation layers over to the Vulkan-Utility-Libraries, but dismissed it as just needless moving of code. But seeing that the same need appears in gfxreconstruct, that makes it clear that its useful for more than just validation layers. Definitely not a priority, but good to know that many repositories could take advantages of a "safe struct" utility library.

charles-lunarg:
My initial code review says that this change is good, and running it through vulkaninfo didn't find any problems. That said, I haven't tried using the repro case and would like to give that a whirl before any final approvals.

MarkY-LunarG:
Verified that this did fix the issue with the repro case I have.

MarkY-LunarG (approved):

PR #1370, opened 2023-11-24
Marked Consumer Process Funcs as Override

andrew-lunarg:
Should make sure that generation of derived class functions declarations can never drift from base class versions leading to base class stubs being silently called.


andrew-lunarg:
The force-push was to remove changes to the ASCII DX12 consumer since that is going away and avoid conflicts with my main convert PR. This can now go in before or after that. I believe it is ready now.

PR #1374, opened 2023-11-27
Integrate cpp code to latest

MarkY-LunarG:
This change brings in the Samsung "toCpp" code generation to GFXReconstruct.  The build for this tool is not enabled by default, and there is a README.md in the tool folder which states that it is considered to be in an Alpha state with some obvious holes.  However, it does convert multiple game captures on my Linux system into an executable that can be run and modified.

Unfortunately, the way we brought the code in created a Blob of the original Samsung code, so it will be an large review process.


framework/decode/vulkan_cpp_consumer_base.cpp
```diff
@@ -0,0 +1,2967 @@
...
+        fprintf(file, "// could be turned into a run-time check dependent on dlsym returning a valid pointer for\n");
+        fprintf(file, "// AHardwareBuffer_lockPlanes.\n");
+        fprintf(file, "#if __ANDROID_API__ >= 29\n");
+        fprintf(file, "\t\t\tif (desc.usage & AHARDWAREBUFFER_USAGE_CPU_WRITE_MASK)\n");
+        fprintf(file, "\t\t\t{\n");
+        fprintf(file, "\t\t\t\tAHardwareBuffer_Planes ahb_planes;\n");
+        fprintf(file, "\t\t\t\tahwbuf_res =\n");
+        fprintf(file,
+                "\t\t\t\t\tAHardwareBuffer_lockPlanes(buffer, AHARDWAREBUFFER_USAGE_CPU_WRITE_OFTEN, -1, "
+                "nullptr, &ahb_planes);\n");
```
MarkY-LunarG, line 2771: This code is not working.  I just encountered it being used and am working on fixing the issues with it.  So please currently ignore all "AHardwareBuffer" items in this file for now.

MarkY-LunarG: Fixes have been applied.

framework/generated/generated_vulkan_consumer.h
```diff
@@ -1803,6 +1804,9 @@ class VulkanConsumer : public VulkanConsumerBase
         format::HandleId                            commandBuffer,
         StructPointerDecoder<Decoded_VkVideoDecodeInfoKHR>* pDecodeInfo) {}
 
+
+
+
```
locke-lunarg, line 1809: Is it possible to remove those newlines?

locke-lunarg (commented):
I wonder if `external/Tocpp-Android-Template/` is a public github repo. If it's, should we add the repo in external, like Vulkan-Headers, instead of copying it? Or we prefer to copy it?

PR #1375, opened 2023-11-29
Fix a crash on replay for trimmed trace

naomipappe:
When the descriptor sets are updated using `vkUpdateDescriptorSetWithTemplate` in the captured application, and the update in question involves updating acceleration structures, the trimmed trace would crash on replay.

This is caused by the fact that state recreation uses `vkUpdateDescriptorSets` and `VkWriteDescriptorSet`, which requires the information for the update of acceleration structures be chained in the `pNext` field of relevant `VkWriteDescriptorSet` with `VkWriteDescriptorSetAccelerationStructureKHR` structure filled appropriately. As the `vkUpdateDescriptorSetWithTemplate` does not share this structure, the `pNext` ends up empty on state recreation, and thus causes a crash.

This change fixes that, by filling out the required structure and copying it.

This change also depends on #1363 to properly copy the structure data.

### DIFF
--- a/framework/encode/vulkan_state_tracker.cpp
+++ b/framework/encode/vulkan_state_tracker.cpp
@@ -1120,12 +1120,23 @@ void VulkanStateTracker::TrackUpdateDescriptorSetWithTemplate(VkDescriptorSet   
 
                 for (uint32_t i = 0; i < current_writes; ++i)
                 {
-                    auto accel_struct = reinterpret_cast<const VkAccelerationStructureKHR*>(src_address);
-                    dst_view_ids[i]   = GetWrappedId<AccelerationStructureKHRWrapper>(*accel_struct);
-                    dst_info[i]       = *accel_struct;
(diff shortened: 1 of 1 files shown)


bradgrantham-lunarg:
I think if you make your branch for this PR to be the #1363 branch plus these changes, then CI won't fail because of missing functions, and after #1363 is merged the commits in this PR that are also in #1363's branch will not be duplicated when this PR is merged.

lunarpapillo:
The LunarG CI failure appears to be due to an unrelated GPU crash. I've restored the machine and will restart the test.

MarkY-LunarG:
I wrote a test to verify this fix in [my fork of Sascha Willem's examples](https://github.com/MarkY-LunarG/Vulkan/tree/marky_move_raytrace_reflect_to_desc_templates).  I'm also going to create a separate issue to not use the pnext chains like we are (as you mentioned) but to eventually fix this correctly.

MarkY-LunarG:
Rebased on dev (now that #1363 was integrated it through this into chaos.

PR #1378, opened 2023-12-01
Extra Error Handling in Convert Main

andrew-lunarg:
Checking of file operation return values and exiting if file per frame and stdout options clash.


tools/convert/main.cpp
```diff
@@ -230,74 +255,88 @@ int main(int argc, const char** argv)
...
-                        json_consumer.StartFile(out_file_handle);
+                        out_stream.Reset(out_file_handle);
+                        json_writer.StartStream(&out_stream);
                     }
                     else
                     {
-                        GFXRECON_LOG_ERROR("Failed to create file: '%s'.", json_filename.c_str());
+                        GFXRECON_LOG_ERROR("Failed to open file: '%s'.", json_filename.c_str());
                         ret_code = 1;
                     }
```
andrew-lunarg, line 333: Everything will probably crash and burn while outputting the next frame to a closed file so in addition, maybe one of 
* Exit here.
* Plumb in some sort of NullOutStream instead of the file stream which drops the frame.
* Reset the FileOutStream to a null file handle and make sure output ops are robust to a null handle.

andrew-lunarg:
I rebased it on latest dev so it can be merged and think it is ready.

charles-lunarg (approved):
Went ahead and removed the use of `goto` since it wasn't consistent with the other tools.

PR #1382, opened 2023-12-03
Convert DX12 Binary Blobs in Functions

andrew-lunarg:
Provide a mechanism for registering function and method arguments
 to be dumped to separate binary files with a single line in the
 Python generator of the DX12 JSON Consumer.

Example usage provided for one argument:

    ID3D12Device.CreateRootSignature(...,pBlobWithRootSignature)

Follow-up to PR https://github.com/LunarG/gfxreconstruct/pull/1362

### DIFF
--- a/framework/generated/dx12_generators/dx12_json_consumer_body_generator.py
+++ b/framework/generated/dx12_generators/dx12_json_consumer_body_generator.py
@@ -57,6 +57,16 @@ class Dx12JsonConsumerBodyGenerator(Dx12JsonConsumerHeaderGenerator, Dx12JsonCom
 
     JSON_OVERRIDES = {}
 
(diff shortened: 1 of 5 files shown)


rurra-amd:
Could this also be used to dump shader bytecode to disk as PSOs are being created?

bradgrantham-lunarg:
@rurra-amd Yes, I think the purpose of this PR was to demonstrate how BLOBs would be identified in the DX12 generator so that `convert` would write BLOBs to the BLOB subdirectory.  Vulkan already does that for e.g. the SPIR-V code parameter passed to `vkCreateShaderModule`.  There's only the one example in this PR, so I think the way forward probably would be someone work on this for a first pass of the obvious BLOBs in DX12, merge that, and then iterate.

PR #1389, opened 2023-12-07
exit with 0 if capture wasn't optimizable

bradgrantham-lunarg:


charles-lunarg (approved):
Only nit is that I prefer giving member variables in classes default values (like false) rather than using an initializer list in a constructor, but that's not something to fix in this PR - just stating my preference.

Gave it a spin on my local machine and didn't see any cause for concern.

charles-lunarg:
If there was any PR to squash into a single commit, its this one due to the need to fix formatting issues.

bradgrantham-lunarg:
> Only nit is that I prefer giving member variables in classes default values (like false) rather than using an initializer list in a constructor, but that's not something to fix in this PR - just stating my preference.
> 
> Gave it a spin on my local machine and didn't see any cause for concern.

I agree but I decided to match the existing style, which I think is important.

bradgrantham-lunarg:
> If there was any PR to squash into a single commit, its this one due to the need to fix formatting issues.

omg clang-format

PR #1390, opened 2023-12-07
Fix CommandHandleType::NumHandleTypes not last of enum

marius-pelegrin-arm:
This create random crashes and undefined behaviors when trying to iterate over handle types or accessing resources...


charles-lunarg (approved):
Yikes! This is a scary bug to be present for as long as it has been.

Something that irks me is that this issue can happen again in a subsequent header update. It occurs to me to try to automate generating these handles, or have some sort of static check that NumHandleTypes is the last enum, but well those add a lot of work for a small gain. 

Maybe a simple `NumHandleTypes // THIS MUST BE THE LAST ENUM` comment would suffice?

PR #1391, opened 2023-12-12
Fix forgotten handles in VulkanStateWriter

marius-pelegrin-arm:
Some handle types are not correctly handled at fast forwarding time, in particular some switch cases are missing in `VulkanStateWriter::CheckCommandHandle`. This causes strange artifacts, for example when fast-forwarding Sascha Willem's `pushdescriptors` sample.


lunarpapillo:
LunarG CI failure seems to be due to GPU crash.  Restarting.

PR #1397, opened 2024-01-02
exit 65 (EX_DATAERR) if file can't be optimized, for now

bradgrantham-lunarg:


tools/optimize/main.cpp
```diff
@@ -133,7 +133,7 @@ void GetUnreferencedResources(const std::string&                              in
         {
             GFXRECON_WRITE_CONSOLE("File did not contain trim state setup - no optimization was performed");
             gfxrecon::util::Log::Release();
-            exit(0);
+            exit(65);
```
panos-lunarg, line 136: 65 sounds oddly specific but why not?

panos-lunarg (approved):

PR #1403, opened 2024-01-11
Check return value of GetRayTracingCaptureReplayShaderGroupHandlesKHR [WIP]

charles-lunarg:
Would appreciate being given direction on whether the error handling is appropriate - the error code will be propagated upwards which may cause the application to fail (due to exiting on any error). It may also cause replay to fail because the error code for this command won't be VK_SUCCESS.

### DIFF
--- a/framework/encode/vulkan_capture_manager.cpp
+++ b/framework/encode/vulkan_capture_manager.cpp
@@ -1320,15 +1320,19 @@ VulkanCaptureManager::OverrideCreateRayTracingPipelinesKHR(VkDevice             
                                      pCreateInfos[i].groupCount;
                 std::vector<uint8_t> data(data_size);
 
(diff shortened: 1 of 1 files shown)


bradgrantham-lunarg:
I think we generally fail if the call succeeded in capture and fails in replay.  I think calling `CheckResult` is appropriate.

bradgrantham-lunarg (approved):

bradgrantham-lunarg:
This PR failed internal CI because it caused recapture of the raytracing captures to fail.  But they were actually failing silently, so a change to CI to represent this PR is appropriate.

bradgrantham-lunarg:
Once the clang-format issue reported by the GitHub Actions CI build is resolved ("CI Build / Ubuntu GCC Release") then please merge this PR.

charles-lunarg:
I'm going to re-run the tests just to be sure it wasn't a fluke (and that the current CI works with this change)

PR #1410, opened 2024-01-27
resolve an overridden function with different signature

bradgrantham-lunarg:
With full warnings enabled on macOS I saw an error with `OutputStream::Reset()` being shadowed by `FileOutputStream::Reset(FILE*)`.  Resolve this shadow by changing the name of Reset for `MemoryOutputStream` to `Clear` (to match C++ container method names), remove the base class `Reset`, and make `FileOutputStream::Reset` not be an override.


yaman-lunarg (approved):
Modifications to classes and subsequent usage are consistent.

PR #1418, opened 2024-02-02
Fix for acceleration structure tracking in the state tracker

panos-lunarg:
When building aTLASes both primitiveCount and deviceAddress can be zero. These cases should be ignored by the state tracker


charles-lunarg (approved):
This approval is mostly to convey that this patch fixes the issue I was seeing with local traces.

Issue #1343, opened 2023-11-09
Metadata Blocks Lack a Block Index in JSON


andrew-lunarg:
There may be some work to plumb the index through (todo: look into it and update this description).

## Metadata Block Example

```json
{
  "meta": {
    "name": "FillMemoryCommand",
    "args": {
      "memory_id": 1173905080320,
      "offset": 0,
      "size": 36,
      "data": "[Binary data]"
    }
  }
}
```

## DX12 Method Call Block Example

```json
{
  "index": 62, // <----------------------- Nice, we want that for metadata blocks too.
  "method": {
    "name": "Map",
    "thread": 1,
    "object": {
      "type": "ID3D12Resource",
      "handle": 28
    },
    "return": 0,
    "args": {
      "Subresource": 0,
      "pReadRange": null,
      "ppData": 1173905080320
    }
  }
}
```

PR #1423, opened 2024-02-15
gfxrecon-convert: Emit index field when writing metadata JSON blocks

nickdriscoll-lunarg:
This PR is to fix issue https://github.com/LunarG/gfxreconstruct/issues/1343 where gfxrecon-convert would not emit indices for metadata blocks.

### DIFF
--- a/framework/decode/dx12_consumer_base.h
+++ b/framework/decode/dx12_consumer_base.h
@@ -83,7 +83,7 @@ class Dx12ConsumerBase : public MetadataConsumerBase, public MarkerConsumerBase
                                                            UINT                                     SrcDepthPitch)
     {}
 
-    void SetCurrentBlockIndex(uint64_t block_index) { current_block_index_ = block_index; }
+    virtual void SetCurrentBlockIndex(uint64_t block_index) override { block_index_ = block_index; }
 
     bool ContainsDxrWorkload() const { return dxr_workload_; }
 
--- a/framework/decode/dx12_consumer_base.h
+++ b/framework/decode/dx12_consumer_base.h
@@ -92,14 +92,11 @@ class Dx12ConsumerBase : public MetadataConsumerBase, public MarkerConsumerBase
     bool ContainsOptFillMem() const { return opt_fillmem_; }
 
   protected:
-    auto GetCurrentBlockIndex() { return current_block_index_; }
+    auto GetCurrentBlockIndex() { return block_index_; }
 
     bool dxr_workload_{ false };
     bool ei_workload_{ false };
     bool opt_fillmem_{ false };
-
-  private:
-    uint64_t current_block_index_{ 0 };
 };
 
 GFXRECON_END_NAMESPACE(decode)
--- a/framework/decode/file_processor.cpp
+++ b/framework/decode/file_processor.cpp
@@ -1795,7 +1795,7 @@ bool FileProcessor::ProcessMetaData(const format::BlockHeader& block_header, for
                         {
                             if (decoder->SupportsMetaDataId(meta_data_id))
                             {
-                                        decoder->DispatchSetTlasToBlasDependencyCommand(header.parent_id, blases);
+                                decoder->DispatchSetTlasToBlasDependencyCommand(header.parent_id, blases);
                             }
                         }
                     }
--- a/framework/decode/file_processor.cpp
+++ b/framework/decode/file_processor.cpp
@@ -797,7 +797,6 @@ bool FileProcessor::ProcessMetaData(const format::BlockHeader& block_header, for
             {
                 if (decoder->SupportsMetaDataId(meta_data_id))
                 {
-                    decoder->SetCurrentBlockIndex(block_index_);
                     decoder->DispatchResizeWindowCommand(
                         command.thread_id, command.surface_id, command.width, command.height);
                 }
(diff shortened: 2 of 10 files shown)


bradgrantham-lunarg:
Hey, Panos, could you take a look at this PR?  Thank you!

framework/decode/vulkan_decoder_base.h
```diff
@@ -229,6 +231,8 @@ class VulkanDecoderBase : public ApiDecoder
         HandlePointerDecoder<VkPipeline>                                pPipelines;
     };
     std::unordered_map<format::HandleId, DeferredOperationFunctionCallData> record_deferred_operation_function_call;
+
+    uint64_t block_index_;
```
panos-lunarg, line 235: It is not necessary for the `VulkanDecoderBase` to store the current block index

framework/decode/vulkan_decoder_base.cpp
```diff
@@ -515,5 +536,10 @@ void VulkanDecoderBase::DispatchSetTlasToBlasDependencyCommand(format::HandleId
     }
 }
 
+void VulkanDecoderBase::SetCurrentBlockIndex(uint64_t block_index)
+{
+    block_index_ = block_index;
```
panos-lunarg, line 541: In order to have consistency between the dx12 and the Vulkan base consume classes this should be implemented the same way it is implemented in the `Dx12DecoderBase`:
```
void Dx12DecoderBase::SetCurrentBlockIndex(uint64_t block_index)
{
    for (auto consumer : consumers_)
    {
        consumer->SetCurrentBlockIndex(block_index);
    }
}
```

framework/decode/metadata_consumer_base.h
```diff
@@ -103,6 +106,11 @@ class MetadataConsumerBase
     virtual void ProcessInitSubresourceCommand(const format::InitSubresourceCommandHeader& command_header,
                                                const uint8_t*                              data)
     {}
+
+    inline void SetCurrentBlockIndex(uint64_t block_index)
```
panos-lunarg, line 110: Declaring this in the common (between dx12 and Vulkan) consumer class is a good idea.
Since `Dx12ConsumerBase` inherits from `MetadataConsumerBase` needs to be also updated:
- `Dx12ConsumerBase::SetCurrentBlockIndex` needs to be marked `vitual` + `override`
- `Dx12ConsumerBase::current_block_index_` needs to be removed as it will inherit the newly added `block_index_` from this class

framework/decode/file_processor.cpp
```diff
@@ -706,6 +706,7 @@ bool FileProcessor::ProcessMetaData(const format::BlockHeader& block_header, for
                 {
                     if (decoder->SupportsMetaDataId(meta_data_id))
                     {
+                        decoder->SetCurrentBlockIndex(block_index_);
```
panos-lunarg, line 709: These calls to `decoder->SetCurrentBlockIndex(block_index_);` are not necessary since the file processor is already calling the all registered decoders' `SetCurrentBlockIndex` from `ProcessBlocks` (line 254)

panos-lunarg (commented):
The changes are in the right direction and should achieve the desired result.
I think it needs to be implemented slightly differently though in order to be consistent with the existing functionality in the dx12 decoders and consumers

nickdriscoll-lunarg:
Panos,
I have made the suggested edits and it looks like its building everywhere, though I'm not sure why the final check won't turn green. I did accept the CLA.
Let me know if this still needs work.

Thank you.

framework/decode/dx12_consumer_base.h
```diff
@@ -83,7 +83,7 @@ class Dx12ConsumerBase : public MetadataConsumerBase, public MarkerConsumerBase
                                                            UINT                                     SrcDepthPitch)
     {}
 
-    void SetCurrentBlockIndex(uint64_t block_index) { current_block_index_ = block_index; }
+    virtual void SetCurrentBlockIndex(uint64_t block_index) override { block_index_ = block_index; }
```
panos-lunarg, line 86: A few lines below (unfortunately github does not allow me to put a comment there) you need to update `GetCurrentBlockIndex` (line 95) to return `block_index_` instead of `current_block_index_`.
Also delete lines 100 to 102 since variable `current_block_index_` is no longer used

framework/decode/file_processor.cpp
```diff
@@ -797,6 +797,7 @@ bool FileProcessor::ProcessMetaData(const format::BlockHeader& block_header, for
             {
                 if (decoder->SupportsMetaDataId(meta_data_id))
                 {
+                    decoder->SetCurrentBlockIndex(block_index_);
```
panos-lunarg, line 800: Please delete the remaining calls to `SetCurrentBlockIndex`. `FileProcessor::ProcessBlocks()` takes care of calling that function (line 254)

framework/decode/metadata_consumer_base.h
```diff
@@ -34,6 +34,9 @@ GFXRECON_BEGIN_NAMESPACE(decode)
 /// implement to handle metacommands.
 class MetadataConsumerBase
 {
+  protected:
+    uint64_t block_index_;
```
panos-lunarg, line 38: nitpicking: I think we tend to declare private variable members at the bottom of the class declarations (after private function declarations)

panos-lunarg (approved):
lgtm

PR #1424, opened 2024-02-16
Fix swapchain image wrappers lifetime

ziga-lunarg:
If an application creates multiple swapchains, the images retrieved with vkGetSwapchainImages() can be the same for different swapchains. The parents of swapchain images were not tracked and when one swapchain was destroyed the swapchain image wrappers were as well, even though another swapchain might still be using them


locke-lunarg (approved):
LGTM!

PR #1425, opened 2024-02-17
fixes #1422: gfxrecon-replay to create --screenshot-dir if it doesn't exist

### LINKED ISSUE #1422: gfxrecon-replay --screenshot-dir does not create a directory if the specified one doesn't exist
Other tools like compress & extract do, which is inconsistent.

yaman-lunarg:
resolution for 1422

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -2056,6 +2056,25 @@ void VulkanReplayConsumerBase::InitializeScreenshotHandler()
 
     if (!options_.screenshot_dir.empty())
     {
+        if (util::filepath::Exists(options_.screenshot_dir))
+        {
+            if (!util::filepath::IsDirectory(options_.screenshot_dir))
+            {
+                GFXRECON_WRITE_CONSOLE("Error while creating directory %s: Already exists as file",
+                                       options_.screenshot_dir.c_str());
+                exit(-1);
+            }
+        }
+        else
+        {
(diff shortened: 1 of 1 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2056,6 +2056,10 @@ void VulkanReplayConsumerBase::InitializeScreenshotHandler()
 
     if (!options_.screenshot_dir.empty())
     {
+        if(!util::filepath::IsDirectory(options_.screenshot_dir))
```
bradgrantham-lunarg, line 2059: What will happen if the thing named already exists and is a file?  I think it may be useful to look at the example in `tools/extract/main.cpp`.

MarkY-LunarG (approved):

framework/decode/vulkan_replay_consumer_base.cpp
bradgrantham-lunarg: yes!

yaman-lunarg: May I squash & merge now?

MarkY-LunarG: Yes, squash then "Rebase and Merge".  I think you might have to squash it locally yourself first.  IDK if GitHub does that but I always manually do it.

yaman-lunarg: My final commit was already squashed locally.

PR #1426, opened 2024-02-19
Refactor methods in Dx12ResourceDataUtil

davidd-lunarg:
Refactor common code for reading or writing resource data into static
functions. The existing functions work the same. The new functions can
be used to allow more control for reading and writing resources.


locke-lunarg (approved):
LGTM!

PR #1428, opened 2024-02-22
Propagate virtual swapchain image compression

marius-pelegrin-arm:
When a `VkImageCompressionControlEXT` instance is found in the `pNext` chain of a `VkSwapchainCreateInfoKHR` instance at swapchain creation for a virtual swapchain, the same compression control should be used to create the images of virtual swapchain (the one not created by the WSI).

This is what this commit does.

### DIFF
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -35,7 +35,6 @@
 #include "vulkan/vulkan.h"
 
 #include <memory>
-#include <optional>
 #include <string>
 #include <unordered_map>
 #include <unordered_set>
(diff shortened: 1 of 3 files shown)


MarkY-LunarG (approved):

framework/decode/vulkan_object_info.h
```diff
@@ -404,6 +405,9 @@ struct SwapchainKHRInfo : public VulkanObjectInfo<VkSwapchainKHR>
     // The following values are only used when loading the initial state for trimmed files.
     std::vector<uint32_t> queue_family_indices{ 0 };
 
+    // For virtual swapchain, we want the compression on virtual images to be the same as on the true swapchain
+    std::optional<VkImageCompressionControlEXT> compression_control;
```
MarkY-LunarG, line 409: Brad has concerns that this isn't forward looking enough and this should be a shared ptr.  I'll look at this and see.  My concern is we can't have a vector of shared pointers because we would have ones for each type.  however, we would only be allocating the size of a pointer for each Swapchain versus an entire structure.

MarkY-LunarG: Brad did catch something.  There is a pointer to an array in this structure which needs to be re-allocated and freed.  This is a blocker, but I will look at it.

MarkY-LunarG: Reverted to using a shared_ptr and creating a copy of the flags that are kept around until swapchain destruction.

Issue #1228, opened 2023-08-16
Make GFXRECON_PAGE_GUARD_ALIGN_BUFFER_SIZES true by default


bradgrantham-lunarg:
If the option `GFXRECON_PAGE_GUARD_ALIGN_BUFFER_SIZES` is not set, is it really possible to get corruption?  Or has PageGuard tracking   improved so that we don't have missing updates for overlapping allocations?  I don't know enough about the implementation to say.  If so, shouldn't we make `GFXRECON_PAGE_GUARD_ALIGN_BUFFER_SIZES` the default?

panos-lunarg:
> is it really possible to get corruption?

Yes.
Imagine you have two buffers, let's say one is an mvp matrix that you update each frame from the host.
The other one is a single u32 value that the get updated by the gpu and the hosts reads it every frame.
Lets say the application binds those two buffer next to each other in the same device memory. One is 64 bytes and the other one is 4 bytes. Since `GFXRECON_PAGE_GUARD_ALIGN_BUFFER_SIZES` is false, depending on the driver, the buffers can actually be bound right next to each other, without any space between them. First you update the mvp and then you read the u32. That read will actually memcpy a whole page from the mapped memory overwriting the mvp.

Setting `GFXRECON_PAGE_GUARD_ALIGN_BUFFER_SIZES` to true will "fool" the application that each buffer needs a whole page of memory so the two buffers will have enough space between them.

> Or has PageGuard tracking improved so that we don't have missing updates for overlapping allocations?

Page guard sees mappings and is not aware what's inside those mappings. So I don't think it's going to be some straightforward change to make page guard aware of that.

I guess the logic behind having `GFXRECON_PAGE_GUARD_ALIGN_BUFFER_SIZES` false by default is so that the application's commands are recorded with as little modification as possible.

I understand the default false is a source of problems with the CI which leads to a lot of wasted time. I agree with setting this to true by default.

panos-lunarg:
No I think `GFXRECON_PAGE_GUARD_ALIGN_BUFFER_SIZES` is the right name. When true the adjusting is done when a `vkGet(Device)[Buffer|Image]MemoryRequirements(2)` is called and it adjusts the `size` and `alignment` reported by these functions by rounding them up to the next multiple of the system's page size.

For example:
```
void PostProcess_vkGetBufferMemoryRequirements(VkDevice              device,
                                                   VkBuffer              buffer,
[...]

PR #1431, opened 2024-02-29
Make GFXRECON_PAGE_GUARD_ALIGN_BUFFER_SIZES true by default

nickdriscoll-lunarg:
Implementing the suggestion from issue #1228


panos-lunarg:
It would be a good idea to update `USAGE_android.md` and `USAGE_desktop_Vulkan.md` so that they state the new default value

PR #1433, opened 2024-03-01
Fix measurement options for android

bartosz-muszarski-arm:
Flush/quit on measurement range options should not have an argument to parse.
Using the following commandline for android replay:
`gfxrecon.py replay --quit-after-measurement-range trace.gfxr`
Will currently generate faulty argument list:
`--quit-after-measurement-range True trace.gfxr`
that won't be parsed properly in the replayer. Correct argument list should be:
`--quit-after-measurement-range trace.gfxr`
This PR removes the extra bool parameter from quit-after-measurement-range flush-measurement-range and flush-inside-measurement-range options.

### DIFF
--- a/android/scripts/gfxrecon.py
+++ b/android/scripts/gfxrecon.py
@@ -181,15 +181,12 @@ def MakeExtrasString(args):
 
     if args.quit_after_measurement_range:
         arg_list.append('--quit-after-measurement-range')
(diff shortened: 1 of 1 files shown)


lunarpapillo:
The LunarG CI failure was due to a crashed Wayland server.  I'm restarting this run.

charles-lunarg (approved):
Agreed, looks like a copy-paste type error that snuck past reviewers.

PR #1435, opened 2024-03-01
Remove duplicate line in .clang-format

charles-lunarg:
This cases newer versions of clang-format to fail to format the file.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -6,7 +6,6 @@ AlignEscapedNewlines: true
 AlignOperands: true
 AlignTrailingComments: true
 AllowAllParametersOfDeclarationOnNextLine: true
-AllowShortFunctionsOnASingleLine: true
 AllowShortCaseLabelsOnASingleLine: false
 AllowShortFunctionsOnASingleLine: Inline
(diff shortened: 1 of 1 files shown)


bradgrantham-lunarg:
`AllowShortFunctionsOnASingleLine: true` is an alias for `All`.  So a repeated directive but not a duplicate.  LGTM at any rate

charles-lunarg:
> `AllowShortFunctionsOnASingleLine: true` is an alias for `All`. So a repeated directive but not a duplicate. LGTM at any rate

Wait if true means all does that mean this change would alter the format of files? I don’t want to check this in and break subsequent pr’s due to wrong formatting.

bradgrantham-lunarg:
I don't think so; you're deleting the first instance of the directive which is (I think) replaced by the later directive.

```
$  clang-format -style='{AllowShortFunctionsOnASingleLine: true, AllowShortFunctionsOnASingleLine: Inline}' -dump-config | grep Short | grep ShortFunctions
AllowShortFunctionsOnASingleLine: Inline
```

PR #1439, opened 2024-03-05
support AGS 6.2.0

igorbrsn:
**The problem**
New AGS 6.2 version support is not supported.

**The solution**
6.2 changes are:

added `shaderClock`.
warning in capture if major version is not supported.
replay warning message if version is not 6.2.0.
backward compatibility supported - we still can replay 6.0.1 captures. We also can capture and replay 6.0.1 version.   
ags.lib and agsd.lib updated to 6.2.0

**Result**
Based on my testing I was able to capture and replay both Full and Trimmed version of 6.2 and 6.0 AGS


lunarpapillo:
CI failure seems to be due to a busy machine.  Re-running.

davidd-lunarg (changes requested):

davidd-lunarg (approved):

PR #1440, opened 2024-03-07
Option for offscreen frame boundary conversion

marius-pelegrin-arm:
Add `--use-ext-frame-boundary` option to convert all offscreen frame boundaries to `VK_EXT_frame_boundary` frame boundaries.

Extensions used to specify an offscreen frame boundary can be unsupported by the replay device. As there are many of such extensions, it seems important to be able to "convert" these frame boundaries to a more "standard" one.
The `VK_EXT_frame_boundary` seems to be the best choice as it is specialy designed for this purpose, platform independent, and supports all functionnalities supported by other extensions:
- Specifying resources attached to the frame (images and buffers)
- Depending on the execution of other commands
- Specifying frame number, label and description

This commit depends on:
- #1301 - Not mandatory, but this commit "extend" it
- #1448 - This commit relies on top of a bug corrected by it, so this should be better to merge it first then rebase this commit on top of it

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -2143,6 +2143,59 @@ void VulkanReplayConsumerBase::WriteScreenshots(const Decoded_VkPresentInfoKHR* 
     }
 }
 
+void VulkanReplayConsumerBase::FillFrameBoundaryExtFromCommandBufferInfo(const CommandBufferInfo* command_buffer_info,
+                                                                         VkFrameBoundaryEXT*      frame_boundary,
+                                                                         std::vector<VkImage>&    frame_boundary_images)
+{
+    assert(command_buffer_info->is_frame_boundary);
+
+    frame_boundary_images.clear();
+
+    for (size_t i = 0; i < command_buffer_info->frame_buffer_ids.size(); ++i)
+    {
+        auto framebuffer_info = object_info_table_.GetFramebufferInfo(command_buffer_info->frame_buffer_ids[i]);
+
+        for (size_t j = 0; j < framebuffer_info->attachment_image_view_ids.size(); ++j)
+        {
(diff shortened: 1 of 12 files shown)


marius-pelegrin-arm:
I just saw that I forgot that PR #1301 wasn't merged. So... This commit "relies" on this PR. There's no real dependency (this commit still has a point and can perfectly be written without PR #1301) but this commit applies modification to the override of `vkCmdInsertDebugUtilsLabelEXT` that is created by PR #1301.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7074,6 +7225,103 @@ VkResult VulkanReplayConsumerBase::OverrideCreateFramebuffer(
...
+        frameBoundaryExt.pBuffers    = nullptr;
+        frameBoundaryExt.tagName     = frameBoundaryExt.frameID;
+        frameBoundaryExt.tagSize     = 0;
+        frameBoundaryExt.pTag        = nullptr;
+
+        submitInfo.pNext = &frameBoundaryExt;
+
+        VkQueue queue;
+        device_table->GetDeviceQueue(device, queueFamily, 0, &queue);
+        device_table->QueueSubmit(queue, 1, &submitInfo, VK_NULL_HANDLE);
```
MarkY-LunarG, line 7315: One of the big concerns I have here is something we ran into before with the virtual swapchain code.  We were running into a race condition when code was being submitted into a different Queue than the triggering semaphore completion work.  This resulted in us performing the virtual swapchain update with incorrect content.  Sometimes "future" content so we saw frame stuttering.

In this case, you may be receiving the semaphore from work that executed a queue that is not index 0.  If this is the case, you're adding additional code to queue 0 potentially well after the work was done.  This may result in the application already sending pending work onto that queue and the resultant "end of frame" may actually be after this content which correctly belongs to the next frame.

PR #1443, opened 2024-03-11
Clarify Vulkan-specific items

MarkY-LunarG:
Since GFXR started as a Vulkan project, many of the items have no clear identifier in their name.  But now that GFXR supports DX and may support other APIs in the future, we should clarify Vulkan-specific items.

### DIFF
--- a/framework/encode/custom_vulkan_api_call_encoders.cpp
+++ b/framework/encode/custom_vulkan_api_call_encoders.cpp
@@ -164,7 +164,7 @@ static void EncodeDescriptorUpdateTemplateInfo(VulkanCaptureManager*     manager
             {
                 size_t              offset = entry_info.offset + (entry_info.stride * i);
                 const VkBufferView* entry  = reinterpret_cast<const VkBufferView*>(bytes + offset);
-                encoder->EncodeHandleValue<VulkanBufferViewWrapper>(*entry);
+                encoder->EncodeHandleValue<vulkan_wrappers::BufferViewWrapper>(*entry);
             }
         }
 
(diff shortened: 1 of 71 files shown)


framework/encode/parameter_encoder.h
```diff
@@ -78,7 +78,7 @@ class ParameterEncoder
     void EncodeFunctionPtr(T value)                                                                                   { EncodeValue(reinterpret_cast<format::AddressEncodeType>(value)); }
 
     template<typename Wrapper>
-    void EncodeHandleValue(typename Wrapper::HandleType value)                                                        { EncodeHandleIdValue(GetWrappedId<Wrapper>(value)); }
+    void EncodeHandleValue(typename Wrapper::HandleType value)                                                        { EncodeHandleIdValue(GetVulkanWrappedId<Wrapper>(value)); }
```
davidd-lunarg, line 81: Add `Vulkan` to the parameter encoder name as well? e.g., `EncodeHandleValue` -> `EncodeVulkanHandleValue`

framework/encode/custom_vulkan_api_call_encoders.cpp
```diff
@@ -164,7 +164,7 @@ static void EncodeDescriptorUpdateTemplateInfo(VulkanCaptureManager*     manager
             {
                 size_t              offset = entry_info.offset + (entry_info.stride * i);
                 const VkBufferView* entry  = reinterpret_cast<const VkBufferView*>(bytes + offset);
-                encoder->EncodeHandleValue<BufferViewWrapper>(*entry);
+                encoder->EncodeHandleValue<vulkan_wrappers::BufferViewWrapper>(*entry);
```
davidd-lunarg, line 167: I wonder if there is another solution for delineating these Vulkan wrappers that doesn't require the `vulkan_wrappers::` prefix on every object. Some potential ideas:
1. Add a Vulkan-specific namespace (e.g., `gfxrecon::encode::vulkan` or `gfxrecon::encode::vk`) for all Vulkan-specific encode files. A Vulkan-specific namespace could also be added for `gfxrecon::decode`, e.g., `gfxrecon::decode::vulkan`
1. Adopt the DX12 wrapper naming pattern `[DX12 object name]_Wrapper`, so `vulkan_wrappers::BufferViewWrapper` becomes `VkBufferView_Wrapper`

MarkY-LunarG: Your idea of the full name might work, but doing additional namespaces was much more involved than I wanted to go down.

bradgrantham-lunarg (approved):
I'll merge when CI succeeds

PR #1446, opened 2024-03-12
Tracking correct CopyAS dest resource size

JerryAMD:
There are two major copy modes for CopyAS
For CLONE mode, the dest AS size should be same as source AS size. For other copy mode, GFXR is not able to get dest size yet, so put 1 for now.

### DIFF
--- a/framework/encode/dx12_state_tracker.cpp
+++ b/framework/encode/dx12_state_tracker.cpp
@@ -1018,8 +1018,11 @@ Dx12StateTracker::CommitAccelerationStructureCopyInfo(DxAccelerationStructureCop
     // Mark that the source acceleration structure was copied.
     source_build_info.was_copy_source = true;
 
-    // TODO: Set dest_size from post build info.
-    dest_build_info.dest_size = 1;
+    if (D3D12_RAYTRACING_ACCELERATION_STRUCTURE_COPY_MODE_CLONE != accel_struct_copy.mode)
+    {
+        // TODO: Set dest_size from post build info.
+        dest_build_info.dest_size = 1;
+    }
 
     inputs_data_resource = dest_build_info.input_data_resource;
 


davidd-lunarg:
`dest_size` is set from prebuild info `ResultDataMaxSizeInByte`:
https://github.com/LunarG/gfxreconstruct/blob/e28151fd413c5e5b710819196c54b6588a350a4d/framework/encode/dx12_state_tracker.cpp#L597
And the docs for `D3D12_RAYTRACING_ACCELERATION_STRUCTURE_COPY_MODE_CLONE` say `The size of the destination is identical to the size of the source.`

Is `ResultDataMaxSizeInByte` guaranteed to be the actual size of the build acceleration structure, or is it an upper limit and the size won't be known unless post build info is queried?

JerryAMD:
> `dest_size` is set from prebuild info `ResultDataMaxSizeInByte`:
> 
> https://github.com/LunarG/gfxreconstruct/blob/e28151fd413c5e5b710819196c54b6588a350a4d/framework/encode/dx12_state_tracker.cpp#L597
> 
> 
> And the docs for `D3D12_RAYTRACING_ACCELERATION_STRUCTURE_COPY_MODE_CLONE` say `The size of the destination is identical to the size of the source.`
> Is `ResultDataMaxSizeInByte` guaranteed to be the actual size of the build acceleration structure, or is it an upper limit and the size won't be known unless post build info is queried?

yes
https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#copyraytracingaccelerationstructure-structures
"The size of the destination is identical to the size of the source."

davidd-lunarg:
The docs for [D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_CURRENT_SIZE_DESC](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_raytracing_acceleration_structure_postbuild_info_current_size_desc#members) say:

If the acceleration structure hasn’t had a compaction operation performed on it, this size is the same one reported by GetRaytracingAccelerationStructurePrebuildInfo.

So I agree, copy size should be the same `ResultDataMaxSizeInByte` when cloning an uncompacted AS.

Thanks for the fix!

davidd-lunarg (approved):

PR #1447, opened 2024-03-13
Add missed "is_fullscreen" value in DxgiSwapchainInfo

feimao-gfxr:
**Problem**
Some trim traces won't go through the replayer or optimizer successfully. If an application is in full-screen state, it should be set back to windowed state at the end of the replay. Currently, for some trim traces, the debug layer also complains with "DXGI ERROR: IDXGISwapChain::Release: Swapchain Released while fullscreen. Switch it to the windowed state first. [ MISCELLANEOUS ERROR https://github.amd.com/Developer-Solutions/gfxreconstruct/pull/66: ]."

**Solution** 
Fixed this bug in Dx12ReplayConsumerBase::SetSwapchainInfo() function. 

**Result**
The error can be avoided.


davidd-lunarg (approved):
Thank you for the fix!

PR #1448, opened 2024-03-18
Fix VK_EXT_frame_boundary extension activation

marius-pelegrin-arm:
In `OverrideCreateInstance`, the extension name was added to `filtered_extensions` AFTER the vector size was saved in the create info structure.

This is obviously an error (it still worked because this option is mainly used when capturing the replay, and that the capture layer ALWAYS support the extension even if not activated).

The bug was introduced by #1354


bradgrantham-lunarg (approved):

marius-pelegrin-arm:
Finally a bigger commit containing this fix and other fixes has been made: #1478 
So I close this one.

PR #1457, opened 2024-03-22
Rerun codegen to get missing sTypes

charles-lunarg:
Commit 4577d24 which added support for VK_NV_low_latency and commit cfbdab5 which adds codegen for walking decoded pNext chains were merged without rebasing & rerunning codegen first.

Not sure how to prevent this problem in the future as 1. codegen changes aren't super frequent and 2. it would require rerunning CI ontop of dev before ANY change is merged.


bradgrantham-lunarg:
I just pulled down `dev` and ran the generators and there's no changes.

bradgrantham-lunarg (approved):

mikes-lunarg:
> Not sure how to prevent this problem in the future as 1. codegen changes aren't super frequent and 2. it would require rerunning CI ontop of dev before ANY change is merged.

ANGLE handles this by generating a json file that contains the hashes of every other generated file. This index file is also marked as a binary file so git won't even attempt to merge it, guaranteeing that simultaneous codegen changes will always result in a merge conflict.

PR #1461, opened 2024-03-25
Tracking SO for EI

JerryAMD:
If EI only use max command count, GFXR lost tracking its SO.


framework/decode/dx12_resource_value_mapper.cpp
```diff
@@ -449,9 +449,16 @@ void Dx12ResourceValueMapper::PostProcessExecuteIndirect(DxObjectInfo* command_l
               state_object_extra_info,
               { command_signature_extra_info, argument_buffer_object_info, argument_buffer_offset } });
     }
-    else
+    else if (max_command_count != 0)
```
davidd-lunarg, line 452: It appears based on the [spec for EI](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist-executeindirect#parameters) that if `max_command_count == 0` then EI won't execute any commands, so this conditional could be moved up:
```
if (max_command_count != 0)
{
    if ((count_buffer_object_info != nullptr) && (count_buffer_object_info->object != nullptr))
    {
        // existing code to handle count buffer use...
    }
    else
    {
        // updated code to handle no count buffer...
    }
}

File: framework/decode/dx12_resource_value_mapper.cpp [hunk 2]
```diff
@@ -449,9 +449,16 @@ void Dx12ResourceValueMapper::PostProcessExecuteIndirect(DxObjectInfo* command_l
...
     {
         // Add resource value offsets to resource_value_infos based on the command signature's arguments.
+        for (auto& info : command_signature_extra_info->resource_value_infos)
+        {
+            if (info.type == ResourceValueType::kIndirectArgumentDispatchRays)
+            {
+                const_cast<D3D12StateObjectInfo*>(info.state_object) = state_object_extra_info;
```
davidd-lunarg, line 459: This modifies the `ResourceValueInfo` objects associated with the command signature. However the command signature could be used in other command lists. Instead I think `GetExecuteIndirectResourceValues` should be modified to take in a state object and use that to set the state object for the resource values it finds.

I think there may also be an issue when count buffer is used and this change may address it as well. The state object set on the count buffer `ResourceValueInfo` does not appear to get passed onto the commands in `Dx12ResourceValueMapper::MapValue`:
https://github.com/LunarG/gfxreconstruct/blob/426e2c62d6eb9d24a5cf659526696c2bb9a44a5a/framework/decode/dx12_resource_value_mapper.cpp#L1262-L1282

davidd-lunarg (changes requested):

framework/decode/dx12_resource_value_mapper.cpp
JerryAMD: The issue is when using Max_Command_Count & IndirectArgumentDispatchRays, the resource value info did not associated to current CommandList's StateObject. For other case, so far so good.

JerryAMD: This modifies the ResourceValueInfo objects associated with the command signature. But this only update if type is ResourceValueType::kIndirectArgumentDispatchRays, and update with current active SO in the command list. And call GetExecuteIndirectResourceValues right after the update and copy all SO over to dst_resource_value_info_map. If the command signature is using by other command list, gfxr will replace the SO by active SO in that command list. No conflict.

davidd-lunarg (approved):

Issue #1454, opened 2024-03-20
VulkanVirtualSwapchain to passthrough the swapchain under certain conditions


yaman-lunarg:
From @bradgrantham-lunarg 
We have the class `VulkanVirtualSwapchain` which makes images to stand in for the images from the Swapchain.  This isn't necessary if the number of swapchain images in replay is the same as in capture and also if the present mode is FIFO.  It would be helpful if `VulkanVirtualSwapchain` could detect that circumstance and then just passthrough the swapchain functions for that particular `VkSwapchain`.

bradgrantham-lunarg:
The spec seems to say that FIFO means the order of presentation of images to the display is equivalent to the order of images QueuePresent'd, but that the Acquire image _ordering_ is not guaranteed.  We didn't seem to have a problem previously capturing and replaying FIFO with the same number of replay images as capture images so it seemed reasonable to assume that the Acquire image queue and the presentation image queue were essentially equivalent.  (We had issues with other present modes and when #images was different in replay from capture, but not when #images was the same and FIFO mode.)  Our experience had been that the ordering is maintained but that could have been dependent on the implementation.  Additional feedback on private channels indicates that this feature would fail on some implementations.

So don't implement this feature as the spec does not seem to guarantee that it will work.

PR #1462, opened 2024-03-25
Use original swapchain images under certain conditions, instead of virtual swapchain

yaman-lunarg:
#1454: Replay to use original swapchain under certain conditions, instead of virtual swapchain.


per-mathisen-arm:
Unfortunately, this is not safe to do. FIFO guarantees only that you will see the images in the order you submit them, _not_ that images will be returned to you in the same order, or in any deterministic order at all.

bradgrantham-lunarg:
> Unfortunately, this is not safe to do. FIFO guarantees only that you will see the images in the order you submit them, _not_ that images will be returned to you in the same order, or in any deterministic order at all.

I think we do get the images and then submit them in order as submitted in replay.  In trimming we cycle through them to get to the "correct" image index for the frame.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2056,6 +2056,25 @@ void VulkanReplayConsumerBase::InitializeScreenshotHandler()
 
     if (!options_.screenshot_dir.empty())
     {
+        if (util::filepath::Exists(options_.screenshot_dir))
```
bradgrantham-lunarg, line 2059: You've already committed this change to `dev`.  I think it's harmless (squash merge will probably just omit it) but it's confusing that it's listed as part of this PR. Is it possible you had not updated `dev` on your system before you made this branch?

bradgrantham-lunarg: Please rebase this branch on current `dev` so this change isn't listed as part of this PR.

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -363,7 +365,17 @@ VkResult VulkanVirtualSwapchain::CreateSwapchainResourceData(const DeviceInfo*
             }
         }
     }
-
+    // Intercept: Issue #1454
```
bradgrantham-lunarg, line 368: Is this the only function that operates on `virtual_swapchain_images`?  I imagine you'd want special handling in any function that creates/updates/destroys that member.  And probably you want to not create them if you don't need them.

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_replay_consumer_base.cpp
yaman-lunarg: Quite possible.  Just clicked update-branch below.

framework/decode/vulkan_virtual_swapchain.cpp
yaman-lunarg: Also `VulkanVirtualSwapchain::QueuePresentKHR()` skips the copy if the swapchain has been determined to have the bypass feature.

bradgrantham-lunarg: I think the short answer is that once this mode is identified, every subsequent member of this class should behave like VulkanCapturedSwapchain.  It may be that isn't possible, because some calls to VulkanCapturedSwapchain members are made before the bypass can be identified.  An example is `ProcessSetSwapchainImageStatePreAcquire` may need to be implemented in `VulkanVirtualSwapchain`

PR #1464, opened 2024-03-25
1454 Bypass virtual swapchain under certain conditions

### LINKED ISSUE #1454: VulkanVirtualSwapchain to passthrough the swapchain under certain conditions

yaman-lunarg:
Fixes #1454: Replay to use original swapchain under certain conditions, instead of virtual swapchain.


bradgrantham-lunarg:
Note this was previously #1462, feedback from that PR should be addressed in this PR.

PR #1469, opened 2024-03-27
Add application exe name for the Android applications

yalcinmelihyasin:


framework/util/file_path.cpp
```diff
@@ -435,6 +435,30 @@ void GetApplicationInfo(FileInfo& file_info)
...
+    std::ifstream cmdlineFile("/proc/self/cmdline");
+
+    if (cmdlineFile.is_open())
+    {
+        std::string line;
+        std::getline(cmdlineFile, line);
+
+        size_t pos = std::string::npos;
+        // Command line arguments separated by null bytes
+        if ((pos = line.find('\0')) != std::string::npos)
```
bradgrantham-lunarg, line 448: A string is null-terminated by definition so this will always find the end of the string; can you tell me more about what you intended here?

framework/util/file_path.cpp
```diff
@@ -435,6 +435,30 @@ void GetApplicationInfo(FileInfo& file_info)
...
+        {
+            strcpy(file_info.AppName, line.substr(0, pos).c_str());
```
bradgrantham-lunarg, line 450: I'd prefer strncpy here

bradgrantham-lunarg: Sorry, my mistake, @davidd-lunarg points out that we have a platform `StringCopy` function, please check that out.

framework/util/file_path.cpp
yalcinmelihyasin: Oh! I got confused with `/proc/self/cmdline`. While reading it, arguments are separated with null characters. So I kinda tried to get the first argument by reading before the null character but probably when reading with getline it automatically copies the first argument only.

Let me fix it. This will also remove fix the other issue.

framework/util/file_path.cpp
```diff
@@ -435,6 +435,28 @@ void GetApplicationInfo(FileInfo& file_info)
     {
         GFXRECON_LOG_WARNING("Failed to retrieve the application executable name");
     }
+#elif defined(__ANDROID__) || defined(__linux__)
+    std::ifstream cmdlineFile("/proc/self/cmdline");
+
+    if (cmdlineFile.is_open())
+    {
+        std::string line;
+        std::getline(cmdlineFile, line);
```
bradgrantham-lunarg, line 444: Can you do one better here and use `istream::getline`?  I think that would allow you to skip the copy.  Something like `        cmdlineFile.getline(file_info.AppName, sizeof(file_info.AppName), '\0');`?

framework/util/file_path.cpp
```diff
@@ -435,6 +435,28 @@ void GetApplicationInfo(FileInfo& file_info)
...
+
+        if (!line.empty())
+        {
+            strcpy(file_info.AppName, line.c_str());
+            GetFileInfo(file_info, file_info.AppName);
+        }
+        else
+        {
+            GFXRECON_LOG_WARNING("Failed to open '/proc/self/cmdline' to get the application executable name");
```
bradgrantham-lunarg, line 453: Please update this warning message to match the condition check (or maybe it goes away if you don't check `line.empty`)

framework/util/file_path.cpp
yalcinmelihyasin: Ptal

framework/util/file_path.cpp
yalcinmelihyasin: Done!

framework/util/file_path.cpp
yalcinmelihyasin: Done!

framework/util/file_path.cpp
yalcinmelihyasin: Please review the current state. if you think that we should still change it, I can.

framework/util/file_path.cpp
```diff
@@ -435,6 +435,28 @@ void GetApplicationInfo(FileInfo& file_info)
...
+    std::ifstream cmdlineFile("/proc/self/cmdline");
+
+    if (cmdlineFile.is_open())
+    {
+        std::string line;
+        std::getline(cmdlineFile, line);
+
+        if (!line.empty())
+        {
+            util::platform::StringCopy(file_info.AppName, sizeof(file_info.AppName), line.c_str(), line.size() * sizeof(line.c_str()[0]));
```
bradgrantham-lunarg, line 448: I would like to get rid of this copy as I noted above, but for later `For [std::string](https://en.cppreference.com/w/cpp/string/basic_string), the elements are bytes` so the multiplication by sizeof is not necessary.

bradgrantham-lunarg (changes requested):

bradgrantham-lunarg:
@yalcinmelihyasin  Thanks for this PR!  It will be great to have the executable name for Linux as well as Android.

framework/util/file_path.cpp
yalcinmelihyasin: Done

bradgrantham-lunarg (approved):

PR #1477, opened 2024-04-01
Add --parallel option to macOS CI builds

etang-cw:
Speed up mac CI builds


bradgrantham-lunarg:
I canceled our internal CI build since the only file changed is the GH workflow file and that's not used by our internal CI.

PR #1478, opened 2024-04-02
Fix OverrideCreateInstance and OverrideCreateDevice

marius-pelegrin-arm:
These two functions have become very hard to read and modify, and full of small edge-case bugs because of this. With this commit I propose to clean a little these two functions, by:
- Removing some unnecessary nested if that should belong in a validation layer, and eventually replacing them with an assert
- Uniformizing variable names between these two functions as they perform a lot of very similar operations
- Adding VK_EXT_frame_boundary as a DEVICE extension, and not an instance extension, which it has never been (and also doing so before setting the modified_extensions in the create info, not after)
- Adding fake support for VK_EXT_frame_boundary
- Removing unsupported device features should only happen when `--remove-unsupported` option is set

### DIFF
--- a/framework/decode/vulkan_feature_util.cpp
+++ b/framework/decode/vulkan_feature_util.cpp
@@ -146,6 +146,21 @@ bool IsSupportedExtension(const std::vector<VkExtensionProperties>& properties, 
     return false;
 }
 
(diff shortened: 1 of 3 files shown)


charles-lunarg (approved):
I have rebased on dev to allow a more recent CI run to be done.

The changes look great, cleans up a lot of accumulated cruft.

PR #1479, opened 2024-04-02
Remove the amount block limitation to DX12 API detect

JerryAMD:


bradgrantham-lunarg:
I don't think this should be removed - it was put in there on purpose so that `gfxrecon-info` ran in a short amount of time with a false result for DX12 if run on a Vulkan capture.

framework/decode/dx12_detection_consumer.h
```diff
@@ -50,11 +50,10 @@ class Dx12DetectionConsumer : public Dx12Consumer
 
     virtual bool IsComplete(uint64_t block_index) override
     {
-        return (block_index > kMaxDX12BlockLimit) || WasD3D12APIDetected();
+        return WasD3D12APIDetected();
```
rurra-amd, line 53: Instead of removing it, you could make `kMaxDX12BlockLimit` remain as a default, but also provide an option thru command line to set a higher limit to cover cases where `kMaxDX12BlockLimit` is too small.

rurra-amd (changes requested):

framework/decode/dx12_detection_consumer.h
rurra-amd: Should explain in the PR desc why this is desired.

As far as I know this is a pretty rare case. If this causes the API to not be detected then we can tell user to expand the limit.

PR #1482, opened 2024-04-05
Track AHB buffers with userfaultfd

panos-lunarg:
Although original page guard memory tracking method (mprotect) is tracking imported AHB buffers, the initial userfaultfd commit did not add support for them.

With this commit imported AHB buffer are now tracked with the userfaultfd method too.

However there is a caveat:
Userfaultfd requires that both base address and length of registered memory regions to be paged aligned. Since in the case of imported buffers we cannot back the regions with a shadow memory (we cannot intercept their allocation and return a shadow pointer instead of the original one), when registering the buffer to uffd we need to page align both base address and length. This can effectively extend the tracking to memory that is not inside the actual buffer and can lead to undefined behavior. A warning message is printed if such a case is detected.

### DIFF
--- a/framework/util/platform.h
+++ b/framework/util/platform.h
@@ -558,6 +558,12 @@ inline size_t GetAlignedSize(size_t size, size_t align_to)
     return size;
 }
 
+inline bool IsAddressAligned(const void* address, size_t alignment)
+{
+    uintptr_t uint_addr = reinterpret_cast<uintptr_t>(address);
+    return (uint_addr % alignment) == 0;
+}
+
 inline LibraryHandle OpenLibrary(const std::vector<std::string>& name_list)
 {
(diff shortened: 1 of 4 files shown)


framework/util/platform.h
```diff
@@ -558,6 +558,12 @@ inline size_t GetAlignedSize(size_t size, size_t align_to)
     return size;
 }
 
+inline bool IsAddressAligned(const void* address, size_t alignment)
+{
+    uintptr_t uint_addr = reinterpret_cast<uintptr_t>(address);
+    return !(uint_addr % alignment);
```
bradgrantham-lunarg, line 564: I suggest comparing explicitly against 0 here for clarity.

bradgrantham-lunarg (approved):
I just had the one request

PR #1486, opened 2024-04-07
Dont call vkWaitForFences with fenceCount 0

ziga-lunarg:
If a shadow fence is already signaled then it is erased from modified_fences, and if modified_fence_count is 0 the call to vkWaitForFences should be skipped because that is not valid according to VUID-vkWaitForFences-fenceCount-arrayLength


davidlunarg (approved):
Looks good to me.

bradgrantham-lunarg (approved):

PR #1487, opened 2024-04-08
Fix the extra frame count when flags is DXGI_PRESENT_TEST in Present(…)

shabei:
Fix the extra frame count when flags is DXGI_PRESENT_TEST in Present(...flags)

Problem:
The capture and replay counting the frame by using the Present(…, flags ) 
as the separator, it doesn’t skip the frame count when flags is DXGI_PRESENT_TEST.

Solution:
If flags is DXGI_PRESENT_TEST in Present(...flags), skip the frame count in both of 
capture and replay.

### DIFF
--- a/framework/encode/d3d12_capture_manager.cpp
+++ b/framework/encode/d3d12_capture_manager.cpp
@@ -598,7 +598,10 @@ void D3D12CaptureManager::PrePresent(IDXGISwapChain_Wrapper* swapchain_wrapper)
 
 void D3D12CaptureManager::PostPresent(IDXGISwapChain_Wrapper* swapchain_wrapper, UINT flags)
 {
-    EndFrame(flags);
+    if (flags != DXGI_PRESENT_TEST)
+    {
+        EndFrame();
+    }
 }
 
 void D3D12CaptureManager::PreProcess_IDXGISwapChain_Present(IDXGISwapChain_Wrapper* wrapper,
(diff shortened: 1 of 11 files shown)


davidd-lunarg (changes requested):
The GitHub check failed the formatting test so clang format probably needs to be applied.
https://github.com/LunarG/gfxreconstruct/actions/runs/8606384083/job/23672704699?pr=1487

tools/replay/desktop_main.cpp
```diff
@@ -263,6 +263,12 @@ int main(int argc, const char** argv)
                 {
 #if defined(D3D12_SUPPORT)
                     dx12_replay_consumer.PostReplay();
+                    if (file_processor.UsesFrameMarkers() == false && 
+                        dx12_replay_consumer.GetDXGITestPresentCount() > 0)
+                    {
+                        GFXRECON_LOG_WARNING_ONCE("This capture contains %u test frames.",
```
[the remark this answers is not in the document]
shabei: The message will be logged only one time at the end of the replay.

framework/encode/d3d12_capture_manager.cpp
```diff
@@ -596,9 +596,12 @@ void D3D12CaptureManager::PrePresent(IDXGISwapChain_Wrapper* swapchain_wrapper)
     }
 }
 
-void D3D12CaptureManager::PostPresent(IDXGISwapChain_Wrapper* swapchain_wrapper)
+void D3D12CaptureManager::PostPresent(IDXGISwapChain_Wrapper* swapchain_wrapper, UINT flags)
 {
-    EndFrame();
+    if (flags != DXGI_PRESENT_TEST)
```
davidd-lunarg, line 601: Should this be `if ((flags & DXGI_PRESENT_TEST) == 0)` to handle the case where `flags` has other bits set with `DXGI_PRESENT_TEST`?

shabei: correct

PR #1489, opened 2024-04-09
Fix uffd uninitialized variable

panos-lunarg:


panos-lunarg:
Fix is included in #1503

PR #1491, opened 2024-04-09
Alpha channel is discarded in screenshots before writing to disk #1455

### LINKED ISSUE #1455: Discard alpha in screenshot
One motivation for dumping alpha was that it provided more information to a developer who was using alpha (if the alpha values made it into the attachment).  However, RGB-only is most useful / least surprising because several developers have fought with screenshots that don't display as expected or compare correctly only because of the alpha channel. Maybe there are contexts like VR or a composited desktop where the alpha is useful, but those are probably much less common than our standard single-application desktop/mobile capture/replay workflow.

We are likely to merge the ability to dump intermediate buffers after Draw/Dispatch/TraceRays shortly.  A developer can use that to dump the actual attachments and therefore get alpha before QueuePresent and at more appropriate times for their pipeline.

For this task, remove alpha from the screenshot dumps for BMP and PNG, both Vulkan and DX12.

yaman-lunarg:
Fixes #1455 

Added `util::imagewriter::WriteBmpImageNoAlpha()` and `util::imagewriter::WritePngImageNoAlpha()`.
Kept previous versions in case the output of alpha channel becomes a command-line option.

### DIFF
--- a/framework/util/image_writer.cpp
+++ b/framework/util/image_writer.cpp
@@ -149,7 +149,7 @@ bool WriteBmpImageNoAlpha(
     // BMP image data requires row to be a multiple of 4 bytes
     // Round-up row size to next multiple of 4, if it isn't already
     const uint32_t rowSizeNoAlpha = ((width * kImageBppNoAlpha) + 3u) & (~0x03);
-    uint32_t bmp_image_size = height * rowSizeNoAlpha;
+    uint32_t       bmp_image_size = height * rowSizeNoAlpha;
     if (bmp_image_size <= data_size)
     {
         FILE*   file   = nullptr;
@@ -216,7 +216,7 @@ bool WritePngImage(
     bool success = false;
 
 #ifdef GFXRECON_ENABLE_PNG_SCREENSHOT
-    uint32_t row_pitch = pitch == 0 ? width * kImageBpp : pitch;
+    uint32_t row_pitch               = pitch == 0 ? width * kImageBpp : pitch;
     stbi_write_png_compression_level = 4;
     if (1 == stbi_write_png(filename.c_str(), width, height, kImageBpp, data, row_pitch))
     {
--- a/framework/util/image_writer.cpp
+++ b/framework/util/image_writer.cpp
@@ -146,9 +146,10 @@ bool WriteBmpImageNoAlpha(
         row_pitch = pitch;
     }
 
-    // BMP image data doesn't require row padding, so compute size with height * width * kImageBppNoAlpha, not row_pitch
-    // * height.
-    uint32_t bmp_image_size = height * width * kImageBppNoAlpha;
+    // BMP image data requires row to be a multiple of 4 bytes
+    // Round-up row size to next multiple of 4, if it isn't already
+    const uint32_t rowSizeNoAlpha = ((width * kImageBppNoAlpha) + 3u) & (~0x03);
+    uint32_t bmp_image_size = height * rowSizeNoAlpha;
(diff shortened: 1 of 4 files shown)


framework/util/image_writer.cpp
```diff
@@ -132,13 +135,88 @@ bool WriteBmpImage(
...
+
+            util::platform::FileWrite(&file_header, sizeof(file_header), 1, file);
+            util::platform::FileWrite(&info_header, sizeof(info_header), 1, file);
+
+            // Y needs to be inverted when writing the bitmap data.
+            auto height_1 = height - 1;
+            auto bytes    = reinterpret_cast<const uint8_t*>(data);
+
+            // Row data without alpha
+            uint8_t* rowBytes = new uint8_t[rowSizeNoAlpha]{ 0 };
```
bradgrantham-lunarg, line 189: I think if you use an auto `vector` here, you don't have to invoke `new`, and then you don't have to `delete` it - the storage will go away automatically.  So maybe something like `std::vector<uint8_t> rowBytes[rowSizeNoAlpha];` and then use "rowBytes.data()" in `util::platform::FileWrite()`.

framework/util/image_writer.cpp
```diff
@@ -132,13 +135,88 @@ bool WriteBmpImage(
...
+    uint32_t row_pitch = width * kImageBpp;
+
+    if (pitch != 0)
+    {
+        row_pitch = pitch;
+    }
+
+    // BMP image data requires row to be a multiple of 4 bytes
+    // Round-up row size to next multiple of 4, if it isn't already
+    const uint32_t rowSizeNoAlpha = ((width * kImageBppNoAlpha) + 3u) & (~0x03);
```
bradgrantham-lunarg, line 151: To round up to an alignment, please use `util::Platform::GetAlignedSize()`, e.g. `rowSizeNoAlpha = util::Platform::GetAlignedSize(width, kImageBppNoAlpha)`.

framework/util/image_writer.cpp
```diff
@@ -149,6 +227,38 @@ bool WritePngImage(
...
+    const std::string& filename, uint32_t width, uint32_t height, uint64_t data_size, const void* data, uint32_t pitch)
+{
+    bool success = false;
+
+#ifdef GFXRECON_ENABLE_PNG_SCREENSHOT
+    uint32_t row_pitch               = pitch == 0 ? width * kImageBpp : pitch;
+    stbi_write_png_compression_level = 4;
+
+    const uint32_t row_pitch_no_alpha = (row_pitch * kImageBppNoAlpha) / kImageBpp;
+    uint8_t*       dataNoAlpha        = new uint8_t[height * kImageBppNoAlpha * row_pitch_no_alpha];
```
bradgrantham-lunarg, line 240: Please use the `vector` idiom here too.

bradgrantham-lunarg (changes requested):

yaman-lunarg:
Would it be a good idea to give the user the option to keep alpha values?

bradgrantham-lunarg (approved):

bradgrantham-lunarg:
> Would it be a good idea to give the user the option to keep alpha values?

So far we've had no indication that anyone has required the alpha values for anything.  We may discover that they do, but we are also getting close to merging a PR to allow saving out inputs and outputs to any draw call.  Saving the color output attachments of the last draw call before QueuePresent would be (I think) a much better solution and gives the real framebuffer contents before presentation.

So I think please squash-merge this and if anyone out in the ecosystem squawks then we can add back in a flag to optionally turn on alpha.

PR #1495, opened 2024-04-12
Uniform behavior for measurements in `FpsInfo`

marius-pelegrin-arm:
Having uniform/standardized log formats/behaviors allow for automation tools to be easier to implement and maintain.

As the measurements done by GFXReconstruct are performance free and were already done on the whole replay if no frame-range was specified, I propose to consider `gfxrecon-replay` as always acquiring measurements, either on the whole replay or on the specified frame-range.

Thus there is no need for special message/behavior depending if a frame-range has been manually specified. That's what this commit implements:
- A measurement file is always created if the replay has been run successfully
- The same log message at the end of the replay is displayed, no matter if a range has been specified or not and no matter how much time the (potentially non-existant) load has taken

In addition, as `CLOCK_MONOTONIC` has an arbitrary starting point, `CLOCK_BOOTTIME` start/end time have been added to the measurement file for time to be coherent between multiple runs of `gfxrecon-replay`

### DIFF
--- a/framework/graphics/fps_info.cpp
+++ b/framework/graphics/fps_info.cpp
@@ -26,37 +26,13 @@
 #include "util/date_time.h"
 #include "util/logging.h"
 #include "util/platform.h"
+#include "nlohmann/json.hpp"
 #include "util/json_util.h"
 
-#include "nlohmann/json.hpp"
 #include <cinttypes>
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(graphics)
-
-static double GetElapsedSeconds(uint64_t start_time, uint64_t end_time)
-{
-    return util::datetime::ConvertTimestampToSeconds(
-        util::datetime::DiffTimestamps(static_cast<int64_t>(start_time), static_cast<int64_t>(end_time)));
-}
-
-static void
-WriteFpsToConsole(const char* prefix, uint64_t start_frame, uint64_t end_frame, int64_t start_time, int64_t end_time)
-{
-    assert(end_frame >= start_frame && end_time >= start_time);
-
-    double   diff_time_sec = GetElapsedSeconds(static_cast<uint64_t>(start_time), static_cast<uint64_t>(end_time));
-    uint64_t total_frames  = (end_frame - start_frame) + 1;
-    double   fps           = (diff_time_sec > 0.0) ? (static_cast<double>(total_frames) / diff_time_sec) : 0.0;
-    GFXRECON_WRITE_CONSOLE("%s %f fps, %f seconds, %" PRIu64 " frame%s, framerange %" PRIu64 "-%" PRIu64,
-                           prefix,
-                           fps,
-                           diff_time_sec,
-                           total_frames,
-                           total_frames > 1 ? "s" : "",
-                           start_frame,
-                           end_frame);
-}
 
 FpsInfo::FpsInfo(uint64_t               measurement_start_frame,
                  uint64_t               measurement_end_frame,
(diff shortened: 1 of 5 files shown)


fabian-lunarg:
_A measurement file is always created if the replay has been run successfully_

why is it necessary to set as default? 
when timing is not the scope for a particular use-case, this will still add clutter in form of those measurement-files?
what when the filesystem is read-only etc.?
this feels definitely like an opt-in feature for me.

marius-pelegrin-arm:
Hello !

So I just updated the commit with a new behavior that should be better.

What is already in the dev branch:
- There is an implicit default measurement range that is the entire trace
- There is a default measurement file path (platform specific)

What my commit still does:
- Measurements are just timestamp queries, which given the cpu-load of GFXReconstruct, can be considered "performance free", so they are still queried on the entire frame-range, no matter if the frame range is specified or not
- The performance logging is still "uniformized" so that it always display the exact same message, instead of having a message different depending on if the frame range is specified or not

What my commit doesn't do anymore (and what it does instead):
- A measurement file isn't created in all cases, but only if the measurement frame range has been specified

framework/graphics/fps_info.cpp
```diff
@@ -68,17 +44,24 @@ FpsInfo::FpsInfo(uint64_t               measurement_start_frame,
...
+    measurement_file_name_(measurement_file_name), frame_start_time_(0), frame_durations_(),
     quit_after_frame_(quit_after_frame), quit_frame_(quit_frame)
 {
     if (has_measurement_range_)
     {
-        GFXRECON_ASSERT(!measurement_file_name_.empty());
+        if (util::filepath::IsFile(measurement_file_name_))
+        {
+            // To avoid thinking an ancient file is the result of this run
+            std::remove(measurement_file_name_.c_str());
```
MarkY-LunarG, line 63: I would use GFXRECON_LOG_WARNING_ONCE to note that it is physically replacing the old file.

framework/graphics/fps_info.cpp
```diff
@@ -189,54 +127,92 @@ bool FpsInfo::ShouldWaitIdleAfterFrame(uint64_t frame)
...
-        uint64_t total_frames  = measurement_end_frame_ - measurement_start_frame_;
-        double   fps           = static_cast<double>(total_frames) / diff_time_sec;
-        GFXRECON_WRITE_CONSOLE("Measurement range FPS: %f fps, %f seconds, %lu frame%s, 1 loop, framerange [%lu-%lu)",
-                               fps,
-                               diff_time_sec,
-                               total_frames,
-                               total_frames > 1 ? "s" : "",
-                               measurement_start_frame_,
-                               measurement_end_frame_);
     }
```
MarkY-LunarG, line 216: It used to output information if there was no measurement range, but now it doesn't.  Should we restore that so you get at least the total elapsed time and FPS?  Or is that not even a possible path with this new logic?

marius-pelegrin-arm: No, it's the opposite, `measurement_start_frame_` and `measurement_end_frame_` are ALWAYS defined, and by default they are set as the first and last frame. This message will always be displayed and will have the same format no matter if a measurement range was specified or not.
What defines if a measurement range was specified is the boolean `has_measurement_range_`, and it will be used to determine if a measurement file is created.

framework/graphics/fps_info.cpp
```diff
@@ -118,64 +105,16 @@ void FpsInfo::EndFrame(uint64_t frame)
...
-                {
-                    GFXRECON_LOG_ERROR(
-                        "Failed to open measurements file '%s' (Error %i).", measurement_file_name_.c_str(), result);
-                    GFXRECON_LOG_ERROR("%s", std::strerror(result));
-                }
-            }
+            measurement_end_boot_time_    = util::datetime::GetBootTime();
+            measurement_end_process_time_ = util::datetime::GetProcessTime();
+            measurement_end_time_         = frame_end_time;
+            ended_measurement_            = true;
```
MarkY-LunarG, line 117: Nit: When I ran I see this:

```
================== Start timer (Frame: 500) ==================
Load time:  0.000000 seconds (frame 1)
Total time: 106.769700 seconds
Measured FPS: 99.404368 fps, 5.029960 seconds, 500 frames, 1 loop, framerange [500-1000)
```

It would be nice if you added a "End timer" comment here as well because it just ends and I wasn't quite sure when the frame range covered while I was replaying.

PR #1496, opened 2024-04-15
Fix exiting on vkDestroySwapchainKHR on macOS

etang-cw:
vkDestroySwapchainKHR closes its window, and we took that to mean the user had closed the window and wanted to exit

Also adds a printout on user close so similar issues are easier to debug in the future

### DIFF
--- a/framework/application/metal_window.mm
+++ b/framework/application/metal_window.mm
@@ -40,6 +40,7 @@ typedef void(^GFXReconKeyCallback)(gfxrecon::application::Application*);
 
 @implementation GFXReconWindowDelegate
 - (void)windowWillClose:(NSNotification*)notification {
+    GFXRECON_LOG_DEBUG_ONCE("User closed window");
     [NSApp terminate:self];
 }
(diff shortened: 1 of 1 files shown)


framework/application/metal_window.mm
```diff
@@ -40,6 +40,7 @@ - (void)windowWillClose:(NSNotification*)notification;
 
 @implementation GFXReconWindowDelegate
 - (void)windowWillClose:(NSNotification*)notification {
+    GFXRECON_LOG_FATAL_ONCE("User closed window");
```
bradgrantham-lunarg, line 43: Can this be dropped?  It won't be a surprise for replay to end if the user pressed "X" on a MacOS window.

bradgrantham-lunarg (approved):
Just the one question about the LOG_FATAL_ONCE

framework/application/metal_window.mm
etang-cw: I switched it to a debug log

PR #1497, opened 2024-04-15
Warn if screenshot indexing change is detected

davidd-lunarg:
GFXR's DX12 frame counting logic changed. In some existing captures, the frame that a screenshot index corresponds to may have changed so log a warning.


shabei (approved):
Looks good.

Issue #1494, opened 2024-04-11
Handle use of AHB on Android without QueuePresent with FrameBoundary as frame


bradgrantham-lunarg:
In some titles on Android (Sascha's raytracing demo was reported, , an AHardwareBuffer is used as the framebuffer (color attachment) and then is noted as the framebuffer in vkFrameBoundaryANDROID using a semaphore from a QueueSubmit.  Presumably the Android compositor then takes that AHB and blits it to the screen or composites it, etc (whatever). 

We don't handle this case in replay so we can't show anything on the screen nor take screenshots.  We have https://github.com/panos-lunarg/gfxreconstruct/tree/panos_hack_is_frame_boundary to attempt to save screenshots but this is just a hack.

For this issue, treat this mode as a first-class presentation mode; make sure capture honors this mode for trim range frame numbers (probably already does because of vkFrameBoundaryANDROID) and then augment our frame processing and display in replay on Android to show the completed image and correctly save off screenshots.  Stretch goal would be to get an AHB from compositor/HWUI and use that so that display is handled the same way it was in capture.

marius-pelegrin-arm:
Hello !

I created a PR (#1498) for supporting screenshots from `vkFrameBoundaryANDROID`. If accepted (or at least of the override is created) I will also modify #1440 to convert `vkFrameBoundaryANDROID` to a `VkFrameBoundaryEXT` when the option is set.

Presenting the image given in `vkFrameBoundaryANDROID` remains to be done.

PR #1498, opened 2024-04-17
Make vkFrameBoundaryANDROID trigger screenshots

marius-pelegrin-arm:
This commit is part of what is requested in #1494


bradgrantham-lunarg:
@panos-lunarg @locke-lunarg could you take a look at this PR, please?  It would be helpful to merge for other sponsors so we should prioritize it.  Thank you!

locke-lunarg:
These code is duplicated with `VulkanReplayConsumerBase::WriteScreenshots`. It will be better if it has a function for these two cases.

panos-lunarg:
Some thoughts on `vkFrameBoundaryANDROID` and virtual swapchains:
`VulkanReplayConsumerBase::WriteScreenshots` takes the image information from the `SwapchainKHRInfo` which should take care of virtual swapchains. 
On the other hand `vkFrameBoundaryANDROID` takes as a direct argument a handle to the `VkImage` that's going to be presented/swapped. I assume that the image handles are acquired through `vkGetSwapchainImagesKHR` which takes care of virtual swapchains. So it should be fine.
@locke-lunarg : does this sound correct to you?

locke-lunarg:
I'm not sure if the image on `vkFrameBoundaryANDROID` is from `vkGetSwapchainImagesKHR`. My option is quit simple that it could have a function, like  `VulkanReplayConsumerBase::WriteScreenshots(ImageInfo* image_info)`.

And then
```
VulkanReplayConsumerBase::WriteScreenshots(const Decoded_VkPresentInfoKHR* meta_info)
{
   ...
          WriteScreenshots(image_info);
   ...
}
```

```
VulkanReplayConsumerBase::OverrideFrameBoundaryANDROID(....)
{
   ...
          WriteScreenshots(image_info);
   ...
}
```

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7229,6 +7229,60 @@ VkResult VulkanReplayConsumerBase::OverrideCreateFramebuffer(
...
+                    : (options_.screenshot_height ? options_.screenshot_height : image_info->extent.height);
+
+            auto instance_table = GetInstanceTable(device_info->parent);
+            GFXRECON_ASSERT(instance_table != nullptr);
+
+            VkPhysicalDeviceMemoryProperties memory_properties;
+            instance_table->GetPhysicalDeviceMemoryProperties(device_info->parent, &memory_properties);
+
+            screenshot_handler_->WriteImage(filename_prefix,
+                                            device,
```
panos-lunarg, line 7267: `devivce_info` instead of `device`

panos-lunarg:
> I'm not sure if the image on `vkFrameBoundaryANDROID` is from `vkGetSwapchainImagesKHR`. My option is quit simple that it could have a function, like `VulkanReplayConsumerBase::WriteScreenshots(ImageInfo* image_info)`.

Yes, I agree with this suggestion.
Basically my question was about whether this change is compatible with virtual swapchain. As far as I can tell I think it is.

locke-lunarg:
> Yes, I agree with this suggestion. Basically my question was about whether this change is compatible with virtual swapchain. As far as I can tell I think it is.

Good point. For virtual swapchain, GetSwapchainImagesKHR return virtual swapchain images, not real swapchain images, so `vkFrameBoundaryANDROID` would use virtual swapchain images to output image, instead of real swapchain images. But virtual swapchain images and real swapchain images are the same, so it's not a problem.

framework/decode/vulkan_replay_consumer_base.cpp
marius-pelegrin-arm: Not sure to understand... WriteImage takes a VkDevice as an argument, not a DeviceInfo

marius-pelegrin-arm:
> I'm not sure if the image on `vkFrameBoundaryANDROID` is from `vkGetSwapchainImagesKHR`. My option is quit simple that it could have a function, like `VulkanReplayConsumerBase::WriteScreenshots(ImageInfo* image_info)`.
> 
> And then
> 
> ```
> VulkanReplayConsumerBase::WriteScreenshots(const Decoded_VkPresentInfoKHR* meta_info)
> {
>    ...
>           WriteScreenshots(image_info);
>    ...
> }
> ```
> 
> ```
> VulkanReplayConsumerBase::OverrideFrameBoundaryANDROID(....)
> {
>    ...
>           WriteScreenshots(image_info);
>    ...
> }
> ```
> 
> to avoid the duplicated code.

Sure the duplication is ugly, but the only code "really" duplicated here is the computation of the screenshot size. I cannot create the proposed function simply because there is no `image_info` accessible from the swapchain, only a vector of `VkImage`.
All the code that could be factorized was factorized in `WriteImage`

framework/decode/vulkan_replay_consumer_base.cpp
panos-lunarg: Right, I saw a compilation error when I cherry picked this patch in dev. In this branch this is correct but from what I see in current dev, `WriteImage` has changed.

locke-lunarg (approved):
Thank you. LGTM

locke-lunarg:
Sorry. That's a wrong click.

locke-lunarg:
@marius-pelegrin-arm  Please solve the conflict, and then it's good to merge.

bradgrantham-lunarg:
> Hi, @locke-lunarg ! I resolved a conflict but that wasn't sufficient; there are compile errors. Will you check out Marius' branch and see if the fix is obvious?

Never mind, I think I see the issue.

PR #1501, opened 2024-04-18
Cmake now generates compile_commands.json file

panos-lunarg:


bradgrantham-lunarg (changes requested):
I think this shouldn't be on by default.  Can we add it in as OFF or set some kind of GFXRECON_EXPORT_COMPILE_COMMANDS that then sets CMAKE_EXPORT_COMPILE_COMMANDS?  Also, this doesn't work for Windows, only for Makefile and Ninja generators, so people expecting a list of commands on Windows who don't know the details of CMAKE_EXPORT_COMPILE_COMMANDS may be disappointed (like I was).

PR #1502, opened 2024-04-18
Do not track raw pnext-data in DescriptorInfo

fabian-lunarg:
this was basically duplicating storage of VkAccelerationStructureKHR-handles and only used to track descriptors containing VkAccelerationStructureKHR. we still need a pNext-pointer/struct later, when assembling VkWriteDescriptorSet, but can create that on-the-fly using stack-memory.

### DIFF
--- a/framework/encode/vulkan_state_tracker.cpp
+++ b/framework/encode/vulkan_state_tracker.cpp
@@ -663,30 +663,6 @@ void VulkanStateTracker::TrackUpdateDescriptorSets(uint32_t                    w
             {
                 auto& binding = wrapper->bindings[current_binding];
 
-                binding.write_pnext = nullptr;
-                binding.write_pnext_memory.Reset();
-                if (write->pNext != nullptr)
-                {
-                    binding.write_pnext = TrackStruct(write->pNext, &binding.write_pnext_memory);
-                    auto* pnext         = reinterpret_cast<const VkBaseInStructure*>(binding.write_pnext);
-                    switch (pnext->sType)
-                    {
-                        case VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET_ACCELERATION_STRUCTURE_KHR:
-                        {
-                            auto* acc_struct = reinterpret_cast<VkWriteDescriptorSetAccelerationStructureKHR*>(
-                                const_cast<void*>(binding.write_pnext));
-                            binding.record_write_set_accel_structs.clear();
-                            std::copy(acc_struct->pAccelerationStructures,
-                                      acc_struct->pAccelerationStructures + acc_struct->accelerationStructureCount,
-                                      std::back_inserter(binding.record_write_set_accel_structs));
-                            acc_struct->pAccelerationStructures = binding.record_write_set_accel_structs.data();
-                            break;
-                        }
-                        default:
-                            break;
-                    }
-                }
-
                 // Update current and write counts for binding's descriptor count. If current count is
                 // greater than the count for the descriptor range defined by dstArrayElement through binding count,
                 // consecutive bindings are being updated.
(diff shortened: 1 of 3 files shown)


bradgrantham-lunarg:
Without understanding this change in too much depth yet, this changes code that was commited in 0d04134e578e4177038e30589a319f3266ce875e, "Trimming for deferred operation" in https://github.com/LunarG/gfxreconstruct/pull/776.  I don't think we have a host deferred operation test.  Have you tried a test with trimming for any of the samples mentioned in the PR https://github.com/LunarG/gfxreconstruct/pull/653 that was one of the source branches for #776?

fabian-lunarg:
I compiled and captured mentioned samples from https://github.com/nvpro-samples/vk_raytracing_tutorial_KHR :

- vk_ray_tracing_advanced_compilation_KHR_app
- vk_ray_tracing_animation_KHR_app
- vk_ray_tracing_instances_KHR_app

I can confirm that capture(w. trimming)/replay of those samples (and some others I tested) worked for the baseline and this PR, no notable changes whatsoever.
this was expected as this PR 'should' not temper with anything related to "host deferred operation".
there are some issues with said samples, when replaying with `gfxrecon-replay --validate`, but those issues seem unrelated to the changes applied here. I'll look around if we already know/track those issues.

framework/encode/vulkan_state_tracker.cpp
```diff
@@ -663,30 +663,6 @@ void VulkanStateTracker::TrackUpdateDescriptorSets(uint32_t                    w
...
-                    {
-                        case VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET_ACCELERATION_STRUCTURE_KHR:
-                        {
-                            auto* acc_struct = reinterpret_cast<VkWriteDescriptorSetAccelerationStructureKHR*>(
-                                const_cast<void*>(binding.write_pnext));
-                            binding.record_write_set_accel_structs.clear();
-                            std::copy(acc_struct->pAccelerationStructures,
-                                      acc_struct->pAccelerationStructures + acc_struct->accelerationStructureCount,
-                                      std::back_inserter(binding.record_write_set_accel_structs));
-                            acc_struct->pAccelerationStructures = binding.record_write_set_accel_structs.data();
```
locke-lunarg, line 682: `binding.record_write_set_accel_structs` and `binding->acceleration_structures` are duplicated. It's good to remove `binding.record_write_set_accel_structs` and use `binding->acceleration_structures`. 

But my concern is to remove `binding.write_pnext = TrackStruct(write->pNext, &binding.write_pnext_memory);` and `write.pNext = binding->write_pnext;`. The original code is to copy and write all pnexts. But this PR become to only write `VkWriteDescriptorSetAccelerationStructureKHR`. The others are ignored. I'm not sure if it's safety. Should we keep that copying and writing all pnexts? or only `VkWriteDescriptorSetAccelerationStructureKHR`?

fabian-lunarg: @locke-lunarg

that's a good point, my reasoning to remove all raw pNext-data was this:

if not `nullptr`, we can expect `pNext` to either hold:
- a `VkWriteDescriptorSetAccelerationStructureKHR` (and we're already saving that data in `binding->acceleration_structures` - so all good)
- a `VkWriteDescriptorSetAccelerationStructureNV` (we do not support it and it's unclear/debatable if we should)
- a `VkWriteDescriptorSetInlineUniformBlock` -> this is addressed in a follow-up [PR](https://github.com/LunarG/gfxreconstruct/pull/1504), adding support for it.

I hence argue we got this covered, both relevant cases can do without the raw pNext data,
the follow-up PR should clear things up and cover the missing inline-uniform-block case.

framework/encode/vulkan_state_writer.cpp
```diff
@@ -2333,6 +2333,11 @@ void VulkanStateWriter::WriteDescriptorUpdateCommand(format::HandleId
 
     const VkCopyDescriptorSet* copy = nullptr;
 
+    // scratch-space for a potential pNext-struct
+    constexpr size_t max_num_bytes_p_next_data = std::max(sizeof(VkWriteDescriptorSetAccelerationStructureKHR),
+                                                          sizeof(VkWriteDescriptorSetInlineUniformBlockEXT));
+    std::array<uint8_t, max_num_bytes_p_next_data> p_next_data{};
+
```
fabian-lunarg, line 2340: @locke-lunarg this is already hinting at the [follow-up](https://github.com/LunarG/gfxreconstruct/pull/1504)

we'll do the same thing for `VkWriteDescriptorSetInlineUniformBlockEXT`

locke-lunarg (approved):
LGTM!

PR #1504, opened 2024-04-22
Add support for VK_EXT_inline_uniform_block 

fabian-lunarg:
overall a low-hanging fruit, adding into existing infrastructure:
- tracking of a binary blob `inline_uniform_block` in DescriptorInfo
- add-hoc (re-)creation of associated pNext-structure for `VkWriteDescriptorSet`
- adapt places skipping over the feature or labeling it unsupported
- gfxrecon-convert: remove a warning about the feature being unsupported

tested successfully with a capture/replay of Sascha's corresponding sample:
https://github.com/SaschaWillems/Vulkan/blob/master/examples/inlineuniformblocks/inlineuniformblocks.cpp

### DIFF
--- a/framework/encode/vulkan_state_writer.cpp
+++ b/framework/encode/vulkan_state_writer.cpp
@@ -2344,9 +2344,13 @@ void VulkanStateWriter::WriteDescriptorUpdateCommand(format::HandleId           
     const VkCopyDescriptorSet* copy = nullptr;
 
     // scratch-space for a potential pNext-struct
-    constexpr size_t max_num_bytes_p_next_data = std::max(sizeof(VkWriteDescriptorSetAccelerationStructureKHR),
-                                                          sizeof(VkWriteDescriptorSetInlineUniformBlockEXT));
+    constexpr size_t max_num_bytes_p_next_data =
+        std::max(sizeof(VkWriteDescriptorSetAccelerationStructureKHR), sizeof(VkWriteDescriptorSetInlineUniformBlock));
     std::array<uint8_t, max_num_bytes_p_next_data> p_next_data{};
+
+    write->pBufferInfo      = nullptr;
+    write->pImageInfo       = nullptr;
+    write->pTexelBufferView = nullptr;
 
     switch (write->descriptorType)
     {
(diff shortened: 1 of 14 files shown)


fabian-lunarg:
I created two captures, baseline vs. this PR, of Sascha's **inlineuniformblocks** sample:

screenshot_frame_500_baseline
screenshot_frame_500_PR1504

the material is set using inline-uniform-blocks, so the spheres in baseline-replay are all black.
this is also correctly reported during capturing:
`[gfxrecon] WARNING - Descriptor type inline uniform block is not currently supported`

this PR adds support for capturing the inline-uniforms and removes the warnings, the replay contains the correct colors.

fabian-lunarg:
- error: ‘VK_FORMAT_R16G16_SFIXED5_NV’ was not declared in this scope
- error: ‘VK_PIPELINE_CREATE_2_ENABLE_LEGACY_DITHERING_BIT_EXT’ was not declared in this scope

weird, looks the header changed, removing deprecated things!?

locke-lunarg:
@fabian-lunarg You need to generate new code.

locke-lunarg:
@fabian-lunarg Your commit changed `external\Vulkan-Headers`. That might not be correct.

bradgrantham-lunarg:
> You changed `external\Vulkan-Headers` to cause those errors. That's probably a merged mistake. We usually don't change `external\Vulkan-Headers` until SDK update.

That's probably me.  Let me take a look.

fabian-lunarg:
I think it might have been a missing `git submodule init / update`involved. seems to be fine now

framework/encode/vulkan_state_writer.cpp
```diff
@@ -2344,47 +2344,49 @@ void VulkanStateWriter::WriteDescriptorUpdateCommand(format::HandleId
...
+        case VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK:
+            if (binding->inline_uniform_block != nullptr)
+            {
+                auto& p_next    = *reinterpret_cast<VkWriteDescriptorSetInlineUniformBlock*>(p_next_data.data());
+                p_next.sType    = VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET_INLINE_UNIFORM_BLOCK;
+                p_next.pNext    = nullptr;
+                p_next.pData    = binding->inline_uniform_block.get();
+                p_next.dataSize = binding->count;
+                write->pNext    = &p_next;
+            }
```
locke-lunarg, line 2383: I wonder if it's possible the binding has both `inline_uniform_block` and `acceleration_structures`. This code is for that the binding has only `inline_uniform_block` or only `acceleration_structures`.

locke-lunarg: It looks that it allows only one type. It's not a problem.

locke-lunarg (approved):
LGTM!

bradgrantham-lunarg (approved):

PR #1510, opened 2024-04-24
Use static liblz4-flavour on Apple-platforms

fabian-lunarg:
we normally link dynamically against liblz4,
but that causes trouble on macOS (and iOS) w.r.t. code signatures.
this PR makes minor changes in our toplevel-CMakeLists and FindLZ4.cmake, in order to use a static liblz4 instead

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -229,6 +229,12 @@ if(WIN32)
 find_package(Detours)
 endif() # WIN32
 
+if(APPLE)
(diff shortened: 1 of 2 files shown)


fabian-lunarg:
seeing some CI issues on github's side:
https://www.githubstatus.com/incidents/7txcg0j03kkg

bradgrantham-lunarg (approved):

PR #1514, opened 2024-04-25
Move base_generator.py to base_generators

davidd-lunarg:
base_generator.py is used by both Vulkan and DX12 generation code. Move it from vulkan_generators folder to base_generators. Also make gencode.py names unique between the two APIs.


MarkY-LunarG:
I'd like a discussion on this with @bradgrantham too. I was planning towards going in the other direction because our current base folder has API specific info in it and shouldn't.  This is causing me painful debug issues in my new project.

PR #1525, opened 2024-04-30
Add missing code-path for VK_EXT_inline_uniform_block

fabian-lunarg:
support for `VK_EXT_inline_uniform_block`was added recently. 
this addresses missing things for the combination with `vkUpdateDescriptorSetWithTemplate`.

note:
there are already only few samples for `VK_EXT_inline_uniform_block`, same for `vkUpdateDescriptorSetWithTemplate`.
so I forked Sascha's samples and added a modified version of **raytracingreflections** to cover this scenario, similar to a fork done by @MarkY-LunarG 

used this to verify the changes work:
https://github.com/fabian-lunarg/Vulkan/blob/custom_samples/examples/raytracingreflectionsdesctemplates_inline_uniform/raytracingreflectionsdesctemplates_inline_uniform.cpp

- descriptors are updated per frame via `vkUpdateDescriptorSetWithTemplate` (update after bind is enabled everywhere)
- uniform-buffer was replaced with an inline-uniform-block

with this PR recording&replay of this sample (and hopefully any application combining `VK_EXT_inline_uniform_block` and `vkUpdateDescriptorSetWithTemplate`) works now, both with and without trimming

### DIFF
--- a/framework/decode/vulkan_cpp_consumer_base.cpp
+++ b/framework/decode/vulkan_cpp_consumer_base.cpp
@@ -2158,11 +2158,13 @@ void VulkanCppConsumerBase::GenerateDescriptorUpdateTemplateData(DescriptorUpdat
     std::vector<std::string>    buffer_desc_info_variables;
     std::vector<std::string>    texel_desc_info_variables;
     std::vector<std::string>    accel_desc_info_variables;
+    std::vector<std::string>    inline_uniform_block_info_variables;
 
-    uint32_t image_info_count        = static_cast<uint32_t>(decoder->GetImageInfoCount());
-    uint32_t buffer_info_count       = static_cast<uint32_t>(decoder->GetBufferInfoCount());
-    uint32_t texel_buffer_view_count = static_cast<uint32_t>(decoder->GetTexelBufferViewCount());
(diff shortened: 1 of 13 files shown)


framework/decode/vulkan_cpp_consumer_base.cpp
```diff
@@ -2285,6 +2303,18 @@ void VulkanCppConsumerBase::GenerateDescriptorUpdateTemplateData(DescriptorUpdat
                 variables.push_back(std::move(offset));
                 break;
             }
+            case VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK:
+            {
+                // TODO: needs testing, unsure when/how we'll use this
```
locke-lunarg, line 2308: Could you add a issue for this TODO? The others look good to me.

locke-lunarg (approved):

framework/decode/vulkan_cpp_consumer_base.cpp
fabian-lunarg: done -> https://github.com/LunarG/gfxreconstruct/issues/1549

fabian-lunarg:
thank you @locke-lunarg!

PR #1534, opened 2024-05-02
Android doc cleanup

mikes-lunarg:
Most of this was motivated by a recent exercise of taking captures to reproduce driver bugs on a stock user Android build:

- Rework the Android Writable Locations to encourage app-specific external storage path which requires no additional permissions
- Modernize the layer loading section to use `gpu_debug_layers_app` along with the replay tool APK
- Remove the global layer instructions, in my experience these cause way more problems than they solve.
- Remove the section on rooting, it should not be required for developers debugging their own apps.

### DIFF
--- a/USAGE_android.md
+++ b/USAGE_android.md
@@ -19,20 +19,16 @@ to one of these other documents:
  * [GFXReconstruct for Desktop Vulkan](./USAGE_desktop_Vulkan.md)
  * [GFXReconstruct for Desktop D3D12](./USAGE_desktop_D3D12.md)
 
-
 ## Index
 
-1. [Behavior on Android](#behavior-on-android)
-    1. [Android Writable Locations](#android-writable-locations)
-2. [Capturing API Calls](#capturing-api-calls)
+1. [Capturing API Calls](#capturing-api-calls)
     1. [Before Use](#before-use)
-    2. [To Root or Not To Root](#to-root-or-not-to-root)
-    2. [Enabling the Layer with ADB](#enabling-the-layer-with-adb)
+    2. [Enabling the Capture Layer](#enabling-the-capture-layer)
     3. [Capture Options](#capture-options)
     4. [Capture Files](#capture-files)
     6. [Capture Limitations](#capture-limitations)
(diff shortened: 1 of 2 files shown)


mark-lunarg:
:+1:

USAGE_android.md
```diff
@@ -19,101 +19,70 @@ to one of these other documents:
...
-Android device.
-The final "external storage" result varies based on Android version but some
-locations that can be tried are:
-
- - `/sdcard/Download`
- - `/storage/emulated/0/Download`
- - `/sdcard/Android/data/${Application Full Name}`
- - `/sdcard`
- - `/mnt/shell/emulated/0`
-
```
MarkY-LunarG, line 66: Isn't this still needed for release versions of an application?  It still might be your own application showing different behavior in a release build with no standard mechanism.

USAGE_android.md
```diff
@@ -234,176 +204,38 @@ It may also be set as a post attach command in the project configuration:
...
 
-Together, the last 3 commands look like the following.
+Together, the last 4 commands look like the following:
 
 ```bash
 adb shell settings put global enable_gpu_debug_layers 1
-adb shell settings put global gpu_debug_app ${Application Full Name}
+adb shell settings put global gpu_debug_app ${Package Name}
 adb shell settings put global gpu_debug_layers VK_LAYER_LUNARG_gfxreconstruct
+adb shell settings put global gpu_debug_layers_app com.lunarg.gfxreconstruct.replay
```
MarkY-LunarG, line 229: What versions of Android is this available in?

USAGE_android.md
mikes-lunarg: I kept the writable location section but moved it under the Capturing API Calls > Before Use section below. I also tweaked it a bit to emphasize the `/sdcard/Android/data/${Package Name}` path since it seemed to be the most reliable on both the user and userdbg builds I was testing.

USAGE_android.md
mikes-lunarg: As far as I can tell, this has been available since Android 10

bradgrantham-lunarg (approved):

PR #1539, opened 2024-05-08
Fix handle research in code generation

marius-pelegrin-arm:
The method `BaseGenerator.check_struct_pnext_handles` in `base_generator.py` should return `True` if the specified struct contains handles, handle pointers, or if a struct that contains handle or handle pointers can be referenced through it (either by the `pNext` chain or with a member that would be a struct or struct pointer).

Until now, to determine if a struct can reference such another struct, the method was retrieving `valid_extension_structs` which query structs that contains a field `structextends` with the checked struct. The issue is that if struct `A` and struct `B` both extend struct `C`, then struct `A` can be in the `pNext` chain of struct `B` (and otherwise) which was not detected.

This commit proposes to solve this issue by including in the `valid_extensions_structs` all the structs extending the parent struct of the checked struct.

### DIFF
--- a/framework/generated/generated_vulkan_api_call_encoders.cpp
+++ b/framework/generated/generated_vulkan_api_call_encoders.cpp
@@ -12718,10 +12718,13 @@ VKAPI_ATTR void VKAPI_CALL CmdSetRenderingAttachmentLocationsKHR(
     {
         encoder->EncodeVulkanHandleValue<vulkan_wrappers::CommandBufferWrapper>(commandBuffer);
         EncodeStructPtr(encoder, pLocationInfo);
-        manager->EndCommandApiCallCapture(commandBuffer);
+        manager->EndCommandApiCallCapture(commandBuffer, TrackCmdSetRenderingAttachmentLocationsKHRHandles, pLocationInfo);
     }
 
-    vulkan_wrappers::GetDeviceTable(commandBuffer)->CmdSetRenderingAttachmentLocationsKHR(commandBuffer, pLocationInfo);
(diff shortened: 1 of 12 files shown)


MarkY-LunarG:
@marius-pelegrin-arm, the codegen has changed drastically since you've made this change.  Can you verify if the current code suffers from the same issue?

marius-pelegrin-arm:
The current code suffers from the same issue, that's certain, but I couldn't really rebase my change because I am not sure to understand the current code. You have at the same time some "global" dicts (`KhronosBaseGenerator.global_structs_with_handles`) and "local" dicts generated from `BaseGenerator.check_struct_member_handles` but the name is ambiguous as to what "with handles" means.

More precisely, the new "global" variable seems to not include AT ALL what happens in the pNext chains... while the `check_struct_member_handles` was handling pNext chain but just "badly" which is what the commit was trying to fix.

TLDR: The problem is still there but I am not sure what the code is trying to do so it's hard to find the proper way to fix it

MarkY-LunarG:
@marius-pelegrin-arm , do you have a way of us testing this? 

Regarding the naming:  `check_struct_member_handles` should check to see that any structure member (or if it is another structure or pointer to possible structures like a `pNext`) may contain handles.  This is important to know in GFXR because during encode, we have to properly encode the correct handle information in place of those handles, and then during decode we need to replace the stored values with the correct ones at replay. 

The work has been tweaked yet again mostly due to findings while porting OpenXR.  But if you pointed us to an example (perhaps in private email) we could look at it and resolve the issue.

PR #1540, opened 2024-05-10
Insert dummy annotations in place of optimized out resources

panos-lunarg:
FileOptimizer now should insert empty annotation meta commands in place of the resources that are removed as not referenced in a trimmed capture range. This should keep the block index when replaying an optimized trimmed capture in alignment with the block index calculated at capture time.

### DIFF
--- a/tools/optimize/file_optimizer.cpp
+++ b/tools/optimize/file_optimizer.cpp
@@ -22,11 +22,13 @@
 
 #include "file_optimizer.h"
 
+#include "format/format.h"
 #include "format/format_util.h"
 #include "util/logging.h"
 #include "util/platform.h"
 
 #include <cassert>
+#include <string>
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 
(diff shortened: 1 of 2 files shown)


tools/optimize/file_optimizer.cpp
```diff
@@ -102,6 +102,22 @@ bool FileOptimizer::FilterInitBufferMetaData(const format::BlockHeader& block_he
         // If the buffer is in the unused list, omit its initialization data from the file.
         if (unreferenced_ids_.find(header.buffer_id) != unreferenced_ids_.end())
         {
+            // In its place insert a dummy annotation meta command. This should keep the block index when
+            // replaying an optimized trimmed capture in in alignment with the block index calculated
+            // at capture time
+            format::AnnotationHeader annotation;
+            annotation.block_header.size = format::GetAnnotationBlockBaseSize();
+            annotation.block_header.type = format::BlockType::kAnnotation;
+            annotation.annotation_type   = format::kUnknown;
```
bradgrantham-lunarg, line 111: Could this annotation contain text instead or even JSON to the effect that it replaced an omitted block?  I think it would be helpful if the annotation even said what was omitted.

bradgrantham-lunarg (changes requested):

tools/optimize/file_optimizer.cpp
panos-lunarg: sure

bradgrantham-lunarg (approved):

PR #1547, opened 2024-05-14
Support D3D12_ELEMENTS_LAYOUT_ARRAY_OF_POINTERS

davidd-lunarg:
Add support for BuildRaytracingAccelerationStructure when D3D12_ELEMENTS_LAYOUT_ARRAY_OF_POINTERS is used in the build desc.

### DIFF
--- a/framework/encode/dx12_state_tracker.cpp
+++ b/framework/encode/dx12_state_tracker.cpp
@@ -192,6 +192,19 @@ void Dx12StateTracker::TrackExecuteCommandLists(ID3D12CommandQueue_Wrapper* queu
                                                 UINT                        num_lists,
                                                 ID3D12CommandList* const*   lists)
 {
+    GFXRECON_ASSERT(queue_wrapper != nullptr);
+    GFXRECON_ASSERT(queue_wrapper->GetObjectInfo() != nullptr);
+    auto queue_info = queue_wrapper->GetObjectInfo();
+    auto queue      = queue_wrapper->GetWrappedObjectAs<ID3D12CommandQueue>();
+    auto device     = graphics::dx12::GetDeviceComPtrFromChild<ID3D12Device>(queue);
+
+    graphics::dx12::ID3D12CommandAllocatorComPtr     post_build_as_copy_cmd_allocator = nullptr;
+    graphics::dx12::ID3D12GraphicsCommandList4ComPtr post_build_as_copy_cmd_list      = nullptr;
+
+    HRESULT  result                                 = S_OK;
+    uint64_t highest_as_build_id                    = 0;
+    bool     executing_acceleration_structure_build = false;
+
     for (UINT i = 0; i < num_lists; ++i)
     {
         auto list_wrapper = reinterpret_cast<ID3D12CommandList_Wrapper*>(lists[i]);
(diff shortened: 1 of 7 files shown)


framework/encode/dx12_state_tracker.cpp
```diff
@@ -618,17 +741,39 @@ void Dx12StateTracker::TrackBuildRaytracingAccelerationStructure(
             build_info.inputs.pGeometryDescs  = nullptr;
             build_info.inputs.ppGeometryDescs = nullptr;
         }
+        else if (desc->Inputs.Type == D3D12_RAYTRACING_ACCELERATION_STRUCTURE_TYPE_TOP_LEVEL)
+        {
+            // TEMP: Force all TLAS builds to take this code path for testing.
+            // if (desc->Inputs.DescsLayout == D3D12_ELEMENTS_LAYOUT_ARRAY_OF_POINTERS)
```
davidd-lunarg, line 747: This conditional is temporarily commented out for testing. Build TLAS with D3D12_ELEMENTS_LAYOUT_ARRAY_OF_POINTERS isn't used heavily, so this forces all Build TLAS calls to take the new code path. Once tests pass and before merging to `dev` this conditional will be restored.

framework/encode/dx12_state_writer.cpp
```diff
@@ -1373,26 +1373,44 @@ void Dx12StateWriter::WriteEnableDRED()
...
+    uint64_t max_inputs_buffer_size = 0;
     state_table.VisitWrappers([&](ID3D12Resource_Wrapper* resource_wrapper) {
         GFXRECON_ASSERT(resource_wrapper != nullptr);
         GFXRECON_ASSERT(resource_wrapper->GetObjectInfo() != nullptr);
 
         const auto resource_info = resource_wrapper->GetObjectInfo();
 
+        if (device_wrapper == nullptr)
+        {
+            device_wrapper = resource_info->device_wrapper;
```
rurra-amd, line 1388: This happened?

Should we log a `GFXRECON_LOG_WARNING`?

davidd-lunarg: The logic here is to set `device_wrapper` once in the loop so that we have a `ID3D12Device_Wrapper` for use in constructing a `Dx12ResourceDataUtil` after the loop. I don't think `resource_info->device_wrapper` should ever be null. Line 1401 below ensures a device wrapper was found with an assert.

framework/encode/dx12_state_tracker.cpp
```diff
@@ -618,17 +741,42 @@ void Dx12StateTracker::TrackBuildRaytracingAccelerationStructure(
             build_info.inputs.pGeometryDescs  = nullptr;
             build_info.inputs.ppGeometryDescs = nullptr;
         }
+        else if (desc->Inputs.Type == D3D12_RAYTRACING_ACCELERATION_STRUCTURE_TYPE_TOP_LEVEL)
+        {
+            // This code path adds support for top level AS builds where `DescsLayout ==
+            // D3D12_ELEMENTS_LAYOUT_ARRAY_OF_POINTERS`. Any top level AS build--regardless of DescLayout value--could
+            // also use this code path, but this path is newer and not as thoroughly tested so use the original code
+            // path where possible.
```
davidd-lunarg, line 749: @rurra-amd Added this comment as suggested.

PR #1555, opened 2024-05-15
openxr: Move xrFrameEnd to manually generated encode

MarkY-LunarG:
Move the xrFrameEnd function to be manually defined so that we can trigger a capture manager EndFrame() call after all is done.


jzulauf-lunarg (approved):
LGTM

PR #1556, opened 2024-05-15
Add DX12 code generation out-of-date check to the GitHub CI build

bradgrantham-lunarg:


.github/workflows/ci_build.yml
```diff
@@ -92,8 +94,30 @@ jobs:
             cc: "cl", cxx: "cl"
           }
     steps:
-    - name: Clone repository
-      uses: actions/checkout@v1
+    - name: Require Python 3.10
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.10'
+        #         cache: 'pip' # caching pip dependencies
```
mikes-lunarg, line 101: did the cache option not work?

mikes-lunarg (approved):
LGTM. I played around with it on my fork so I could see what a failure looks like.

.github/workflows/ci_build.yml
bradgrantham-lunarg: With the cache line in there I got this error:
```
Run actions/setup-python@v5
Installed versions
Error: No file in D:\a\gfxreconstruct\gfxreconstruct matched to [**/requirements.txt or **/pyproject.toml], make sure you have checked out the target repository
```
I'll take out the `cache` line.

mikes-lunarg: Ah ok. From the docs I it looks like they expect a requirements file so they can hash and use it as a cache key: https://github.com/actions/setup-python?tab=readme-ov-file#caching-packages-dependencies

Then you would call pip with `pip install -r requirements.txt`. That would let you version-lock your python dependencies and invalidate the cache when they get updated.

mikes-lunarg: Looks like the pip install only takes 10-20 seconds of a 10 minute job, so caching is probably overkill anyway!

bradgrantham-lunarg: Just running the generators required SO MANY dependencies (and I discovered each one after waiting 15 minutes for the CI build to run) that I just didn't care enough to enable caching.  Thank you for doing the footwork to figure out how I would reintroduce the cache!

PR #1557, opened 2024-05-16
Do second round API detect for optimizer if needed

JerryAMD:
**Problem**
For some cases, 1000 blocks is not enough for API detection.

**Solution**
If the first round detection with 1000 blocks doesn't return any detection result, we will do a second round API detection with no-block-limit parameter. The detection will be done with no block limit, till one of targeted API calls is found. A message will show in the console if the second round is needed.

**Result**
It looks for most cases, 1000 blocks should be enough for the API detection. The second round detection should always give us the result and it will be performed only the first round detection (short) doesn't provide any result.


davidd-lunarg (approved):

davidd-lunarg:
Looks good, but it needs clang format before it can be merged.

PR #1559, opened 2024-05-20
Add pipeline creation attempt annotation

### LINKED ISSUE #1197: Add a MetaDataHeader for option GFXRECON_ALLOW_PIPELINE_COMPILE_REQUIRED
`GFXRECON_ALLOW_PIPELINE_COMPILE_REQUIRED` will ignore writting Create*Pipeline to file when it's with `VK_PIPELINE_CREATE_FAIL_ON_PIPELINE_COMPILE_REQUIRED_BIT`. But we want to leave a record in the file to convert it into .json file or print it during replay, like "application requested creategraphicspipelines but that wasn't passed on to Vulkan implementation".

beau-lunarg:
fixes #1197


fabian-lunarg (approved):
code looks straight-forward and clean. also there 'might' be something to follow-up wrt trimming. we'll need to think of a testcase for this, but I'd say let's move on and merge

PR #1560, opened 2024-05-20
Fix for 32/64-bit portability when encoding inline-uniform-blocks

fabian-lunarg:
- Do not use `ParameterEncoder::EncodeUInt8Array`,  avoid usage of non-portable `sizeof(size_t)`.
- Add '`ParameterEncoder::EncodeRawBytes`' to write raw-byte-arrays without meta-data (and avoid looping over the bytes with 'EncodeUInt8Value')


fabian-lunarg:
follow-up fix for https://github.com/LunarG/gfxreconstruct/pull/1525 @davidlunarg

davidd-lunarg (approved):

fabian-lunarg:
thank you @davidd-lunarg!

PR #1561, opened 2024-05-21
openxr: Manually handle xrEnumerateSwapchainImages

MarkY-LunarG:
Because of the way the xrEnumerateSwapchainImages uses the XrSwapchainImagesBaseHeader struct in a direct array, we can't dynamically allocate the array elements, but we have to allocate it all up front and then read the correct data.  Because of this, make a "Peak" function which looks ahead and determines the appropriate structure type and then perform the allocate and read of the correct structures.


MarkY-LunarG:
From a suggestion by @bradgrantham-lunarg, I looked at `VkAccelerationStructureGeometryKHR` and our issue is a little more complex than that one.  With the `VkAccelerationStructureGeometryKHR` being a `union`, the size is correct for any of the child types (since a union is as big as its largest type).  So when you do an allocate for the array:

```
wrapper->pGeometries =
   DecodeAllocator::Allocate<StructPointerDecoder<Decoded_VkAccelerationStructureGeometryKHR>>();
```

It works because the size will work for any of the children.  However, this does not work for the `BaseHeader` because the base structure is actually the minimum size of any of the child structures.  So we can't just call Allocate on it. 

We might be able to do a union, but then the type wouldn't match up and it would have to be patched everywhere (i.e. `XrSwapchainImageBaseHeader` would have to be replaced with `OurInternalSwapchainUnion` everywhere the former was used in the encode or decode path).  That would work, but I'm not sure it's any cleaner than what I've done in  this PR.

MarkY-LunarG:
Actually, after more thought, the custom `union` idea won't work because the array is tightly packed.  So if the user passed in an array of child A (smaller type) and we tried to force a union (aligned with a bigger type) things would just fail.  So we can't use the union in this case.

PR #1563, opened 2024-05-22
Set size of Dx12 swapchain to associated windows

JerryAMD:
**Problem**
IDXGIFactory::CreateSwapChainForHwnd allow swapchain's width and height as zero. When replay the trace, GFXR creates 0x0 window to replay. https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_2/nf-dxgi1_2-idxgifactory2-createswapchainforhwnd#remarks

**Solution**
Inject a metadata block right before CreateSwapChainXXX() call and record the proper swapchain size. Replace the proper swapchain size by the metablock during replay.


framework/generated/dx12_generators/dx12_wrapper_body_generator.py
```diff
@@ -758,6 +758,11 @@ def write_class_member_def(self, class_info, indent=''):
                 expr += unwrapped_args
             expr += ');\n'
 
+            method_names = ["CreateSwapChainForHwnd", "CreateSwapChainForCoreWindow"]
```
davidd-lunarg, line 761: Instead of modifying the generators, please add these functions to `capture_overrides.json` and call `UpdateSwapChainSize` from the new override methods in `D3D12CaptureManager`.

`CreateSwapChainForComposition` should also be included.

davidd-lunarg (approved):
Looks good. There are conflicts with recent changes in `dev` so it will need to be rebased manually before it can be merged. Thanks!

PR #1568, opened 2024-05-30
Fix crash when window cannot be opened with same resolution

marius-pelegrin-arm:
This commit partially fix issue #759 

Fix a crash on `VK_ERROR_OUT_OF_DATE_KHR` when the window of the replayer cannot be resized to the size of the swapchain at capture time and that the replayer uses virtual swapchain.

The idea is to store the actual size of the window that could be opened at replay time in the virtual swapchain so that the virtual swapchain has the same size as at capture time, and the actual swapchain is the size of the window. When the image is copied from the virtual image to the actual image, only a sub-part of the image is copied.


MarkY-LunarG:
Rebasing this on the latest tree.  I'll be overwritting your tree.

PR #1577, opened 2024-06-11
Add SpirV-reflection to detect usage of buffer-references

fabian-lunarg:
- external: replace duplicated spirv.h / spirv-reflect with submodules
- Adds a SpirV-parser to detect dereferencing of buffer-references (refs that are actually used), 
- trace back references to their definition and **extract descriptor-set/binding/push-constant-block**, 
- from there, using spirv_reflect: **extract buffer-offsets and potential array-strides**.
- finally: **issue a warning** if the spirv contained any buffer-references -> currently not supported 

there is a standalone version of the parser available here:
https://github.com/fabian-lunarg/SPIRV-Parsing-Example (forked from @spencer-lunarg 's proof-of-concept)

 I used that to confirm correctness and measure performance on:
 - Sascha's shader-codebase (2 shaders contain buffer-references)
 - my hobby-engine shaders https://github.com/crocdialer/vierkant/tree/master/shaders (9 shaders contain buffer-references)

note:
this is currently only handling `vkCreateShaderModule`, leaving other entries dealing with spirv for a follow-up:
- `vkCreateShaderEXT`
- `VkShaderModuleCreateInfo` inlined directly into `VkPipelineShaderStageCreateInfo::pNext`


framework/encode/vulkan_capture_manager.cpp
```diff
@@ -2822,5 +2823,35 @@ void VulkanCaptureManager::PreProcess_vkBindImageMemory2(VkDevice
...
+                                                            VkDevice                        device,
+                                                            const VkShaderModuleCreateInfo* pCreateInfo,
+                                                            const VkAllocationCallbacks*    pAllocator,
+                                                            VkShaderModule*                 pShaderModule)
+{
+    // spirv-parsing for buffer-references
+    gfxrecon::util::SpirVParsingUtil spirVParsingUtil;
+    printf("PostProcess_vkCreateShaderModule\n");
+
+    // TODO: adding this block compiles/links fine, but bricks the layer somehow!?
```
fabian-lunarg, line 2836: this is where I'm currently stuck. 
commenting in this block will not issue any error/warning, but silently suppress the layer from loading/working.

fabian-lunarg: ok, avoiding usage of operator-new in SpirVParsingUtil solved the silent bail-out.

framework/util/spirv_parsing_util.cpp
```diff
@@ -0,0 +1,476 @@
...
+        while (object_insn)
+        {
+            switch (object_insn->opcode())
+            {
+                case spv::OpConvertUToPtr:
+                case spv::OpCopyLogical:
+                case spv::OpLoad:
+                    object_insn = FindDef(object_insn->operand(0));
+                    break;
+                case spv::OpAccessChain:
```
fabian-lunarg, line 211: this extraction of access-chain -> indices 'seems' to work properly. does this look sane to you @spencer-lunarg ?

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -2822,5 +2823,34 @@ void VulkanCaptureManager::PreProcess_vkBindImageMemory2(VkDevice
     }
 }
 
+void VulkanCaptureManager::PostProcess_vkCreateShaderModule(VkResult                        result,
```
spencer-lunarg, line 2826: heads up, there are 3 spots you can get SPIR-V

1. here
2. `vkCreateShaderEXT`
3. with both `VK_KHR_maintenance5` and `graphicsPipelineLibrary` you can inline things and just pass a `VkShaderModuleCreateInfo` directly into`VkPipelineShaderStageCreateInfo::pNext`

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -2822,5 +2823,34 @@ void VulkanCaptureManager::PreProcess_vkBindImageMemory2(VkDevice
...
+        gfxrecon::util::SpirVParsingUtil spirv_util;
+
+        if (spirv_util.Parse(pCreateInfo->pCode, pCreateInfo->codeSize))
+        {
+            auto buffer_reference_infos = spirv_util.GetBufferReferenceInfos();
+
+            if (!buffer_reference_infos.empty())
+            {
+                GFXRECON_LOG_WARNING_ONCE(
+                    "Shader is using the 'GL_EXT_buffer_reference2' feature. "
```
spencer-lunarg, line 2848: for VVL, we say things like

`Shader is using the SPV_KHR_physical_storage_buffer (aka GL_EXT_buffer_reference)`

basically use the SPIR-V term incase they aren't using GLSL to create the shader

framework/encode/vulkan_capture_manager.cpp
fabian-lunarg: cool, only had vkCreateShaderEXT on the radar (think those are subject to follow-up PRs)

framework/util/spirv_parsing_util.cpp
```diff
@@ -0,0 +1,476 @@
...
+        }
+        else if (opcode == spv::OpDecorate)
+        {
+            decorations_instructions_.push_back(&insn);
+        }
+
+        // There is always a load that does the dereferencing
+        if (opcode != spv::OpLoad)
+        {
+            continue;
```
spencer-lunarg, line 398: just noting that the `Parse` function currently is heavily aimed toward buffer device addressing, which I think is totally fine, but something to caution if someone wanted to add something different to this as well

framework/util/spirv_parsing_util.cpp
spencer-lunarg: something to be aware of in general is there could be 2 `OpAccessChains` chained together

https://github.com/KhronosGroup/SPIRV-Guide/blob/main/chapters/access_chains.md#example---chained-access-chains

framework/util/spirv_parsing_util.cpp
```diff
@@ -0,0 +1,476 @@
...
+                    object_insn = FindDef(object_insn->operand(0));
+                    break;
+                case spv::OpAccessChain:
+                {
+                    std::vector<uint32_t> indices;
+                    for (uint32_t i = 1; i < object_insn->num_operands(); ++i)
+                    {
+                        if (auto ins = FindDef(object_insn->operand(i)))
+                        {
+                            if (ins->opcode() == spv::OpConstant)
```
spencer-lunarg, line 218: so this is fine for 99% of cases... but for VVL we ended up creating this helper for many reason

```
// While simple, function name provides a more human readable description why Word(3) is used.
//
// The current various uses for constant values (OpAccessChain, OpTypeArray, LocalSize, etc) all have spec langauge making sure they
// are scalar ints. It is also not valid for any of these use cases to have a negative value. While it is valid SPIR-V to use 64-bit
// int, found writting test there is no way to create something valid that also calls this function. So until a use-case is found,
// we can safely assume returning a uint32_t is ok.
uint32_t Instruction::GetConstantValue() const {
    // This should be a OpConstant (not a OpSpecConstant), if this asserts then 2 things are happening
    // 1. This function is being used where we don't actually know it is a constant and is a bug in the validation layers
    // 2. The CreateFoldSpecConstantOpAndCompositePass didn't fully fold everything and is a bug in spirv-opt
    assert(Opcode() == spv::OpConstant);
    return Word(3);
}
```

.... for now, I would just go `if not a `OpConstant` exit cleanly and mark as a "edge case"

framework/util/spirv_parsing_util.cpp
fabian-lunarg: yes, good point. maybe just name it `ParseBufferReferences` since that all it currently does

spencer-lunarg: ya, mainly to give the future person a fair idea it is not just a simple "add my small code in it"

fabian-lunarg:
awesome, thank you @spencer-lunarg

framework/util/spirv_parsing_util.cpp
spencer-lunarg: thinking about this more, the main 2 things to worry about is

1. `OpSpecConstant` (Which I am not sure if fully legal here without trying to create it manually in SPIR-V, so might be not something to worry about)
2. `OpConstant` **can** be 64-bit, but have found almost never in the wild

again, I think you have this wrapped around safely enough... just dropping a lot of un-fun SPIR-V knowledge

fabian-lunarg: here I was also frequently seeing an `OpLoad` instead, which I just ignored because it still yielded useful info. 
for a statement like: 
```
layout(buffer_reference, scalar) readonly buffer VertexBufferPtr { packed_vertex_t v[]; };
layout(binding = 0, set = 0) readonly buffer Vertices { VertexBufferPtr vertex_buffers[]; };
...
Vertex v = unpack(vertex_buffers[draws[indices.mesh_draw_index].mesh_index].v[gl_VertexIndex]);
```
we get:
`buffer-reference: (Vertices) -> vertex_buffers (set: 0, binding: 0, buffer-offset: 0, array-stride: 8)`
-> a buffer with tightly packed references, all we want to know

spencer-lunarg: this just means the access is dynamic

so if I went `data[3]` its will be `OpConstant 3` but if I go `data[index]` it will be an `OpLoad` to the `index` variable as its dynamic and you don't know which index, but you know still where it started from which is what leads you to the descriptor set/binding

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -2822,5 +2823,34 @@ void VulkanCaptureManager::PreProcess_vkBindImageMemory2(VkDevice
...
+        // spirv-parsing for buffer-references
+        gfxrecon::util::SpirVParsingUtil spirv_util;
+
+        if (spirv_util.ParseBufferReferences(pCreateInfo->pCode, pCreateInfo->codeSize))
+        {
+            auto buffer_reference_infos = spirv_util.GetBufferReferenceInfos();
+
+            if (!buffer_reference_infos.empty())
+            {
+                GFXRECON_LOG_WARNING_ONCE(
```
davidd-lunarg, line 2847: If the buffer device capture replay feature is enabled will SPIRV buffer references still be a problem for GFXR, or can the warning be skipped?

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -2822,5 +2823,34 @@ void VulkanCaptureManager::PreProcess_vkBindImageMemory2(VkDevice
...
+                    "Shader is using the 'SPV_KHR_physical_storage_buffer' feature. "
```
davidd-lunarg, line 2848: It might be good to also add the warning to replay. From a user's perspective, replay may be where they see something go wrong so having the warning logged in the replay's output could be helpful.

framework/encode/vulkan_capture_manager.cpp
fabian-lunarg: as I understand the feature, it would give you the same addresses for consecutive runs of the same application, on the same device. 
_"indicates that the implementation supports saving and reusing buffer and device addresses"._

we still need to do the detection in spirv, later extract the addresses from buffers/copies, map+track the pointed-to buffers, etc.
I cannot see how "device capture replay" could help or solve that on its own, esp. when replaying on other driver/devices.
so I think we should issue this warning also with capture-replay being available (and maybe re-visit the other warning as I'm unclear what exactly we'd gain from having the feature available)

framework/encode/vulkan_capture_manager.cpp
fabian-lunarg: Added parsing + warning to vulkan_replay_consumer, issuing this warning if a replay hits a buffer-ref:

_[gfxrecon] WARNING - Shader is using the 'SPV_KHR_physical_storage_buffer' feature. Resource tracking for buffers accessed via references is currently unsupported, so replay may fail._

framework/encode/vulkan_capture_manager.cpp
davidd-lunarg: My understanding was that BDAs need to be identified in shaders so those BDAs can be translated from capture to replay time values. If capture and replay is done on the same device/driver using the BDA capture replay feature then capture-time and replay-time BDA values should be the same and no translating is needed. Is there other mapping or tracking that needs to be done for GFXR to support this use case?

I agree translation is needed for future support of capture and replay on different devices/drivers--and also may be needed to support devices/drivers that don't support the BDA capture replay feature.

fabian-lunarg: _Is there other mapping or tracking that needs to be done for GFXR to support this use case?_
in my test-cases so far, using ubuntu/x86_64/nvidia, I can confirm this is like you point out. samples using buffer-references seem to 'magically' capture/replay fine locally. trimming would be a case I 'think' we would currently hit a wall. we do not track buffer-references as pipeline-dependencies, so would potentially trim them away.

fabian-lunarg:
we need to resolve an upstream-issue with spirv-reflect:
https://github.com/KhronosGroup/SPIRV-Reflect/issues/271

PR #1578, opened 2024-06-12
Remove empty command buffer from frame boundary submission

marius-pelegrin-arm:
To submit VkFrameBoundaryEXT, a queue submission is necessary but not having a command buffer inside the queue submission. Empty command buffers are badly handled by some drivers and the management of this buffer uselessly requires some resources and can cause bugs when the replayer terminates unexpectedly.

I propose to simply remove this empty command buffer entirely.

### DIFF
--- a/framework/decode/vulkan_offscreen_swapchain.cpp
+++ b/framework/decode/vulkan_offscreen_swapchain.cpp
@@ -47,8 +47,7 @@ VkResult VulkanOffscreenSwapchain::CreateSurface(VkResult                       
 
     // For multi-surface captures, when replay is restricted to a specific surface, only create a surface for
     // the specified index.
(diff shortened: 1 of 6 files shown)


charles-lunarg (approved):

charles-lunarg:
I've updated the branch to be ontop of dev which allows for an up-to-date CI run to occur. Assuming it passes, this PR is ready to be merged.

marius-pelegrin-arm:
Hi ! I just saw that the option is broken in your repository because when the `VulkanSwapchainOptions` structure has been created, the offscreen swapchain frame boundary option has been forgotten.

I added a commit to fix that.

charles-lunarg (approved):
Thanks for catching and fixing that!

PR #1585, opened 2024-06-14
Add support for `VK_EXT_device_fault` for better crash reporting.

naomipappe:
This commit adds some additional error reporting on failed `CheckResult` call, in case the replay device supports the `VK_EXT_device_fault`.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -4165,9 +4165,9 @@ VkResult VulkanReplayConsumerBase::OverrideAllocateMemory(
 
             VkMemoryAllocateInfo                     modified_allocate_info = (*replay_allocate_info);
             VkMemoryOpaqueCaptureAddressAllocateInfo address_info           = {
-                VK_STRUCTURE_TYPE_MEMORY_OPAQUE_CAPTURE_ADDRESS_ALLOCATE_INFO,
-                modified_allocate_info.pNext,
-                opaque_address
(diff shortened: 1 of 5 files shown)


naomipappe:
So I did this PR when I though there is a need of this kind of reporting in the tool, but as I have recently learned of the Crash Diagnostic Layer, which also displays this information and more, is configurable and plug-in, I guess this PR does not make sense. Would it be a good idea to build Crash Diagnostic Layer alongside GFXR and include it if enabled via command line argument?

MarkY-LunarG:
> So I did this PR when I though there is a need of this kind of reporting in the tool, but as I have recently learned of the Crash Diagnostic Layer, which also displays this information and more, is configurable and plug-in, I guess this PR does not make sense. Would it be a good idea to build Crash Diagnostic Layer alongside GFXR and include it if enabled via command line argument?

Hi @naomipappe , that's exactly what I was going to suggest.  Especially because we might not want to enable it all the time in GFXR because it does have an impact on performance and possibly behavior (depending on the implementation in the driver).  Feel free to close this.  If not, just ping me again and I can look at the updated change.

MarkY-LunarG:
It's  been over a month.  I will close this for now.  Please re-open if there's more to this that the Crash Diagnostic Layer can't provide.

naomipappe:
Thanks @MarkY-LunarG, I was preoccupied with this and forgot to close it. If I have any follow-up I will open a new pull-request instead of reviving this discussion.

PR #1586, opened 2024-06-17
Add MacOS-related extensions to allow-list for missing extensions

fabian-lunarg:
- `VK_EXT_METAL_SURFACE_EXTENSION_NAME`
- `VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME`

will be used during a gfxr-capture on MacOS, but should be ignored, if missing, during a replay.

### DIFF
--- a/framework/decode/vulkan_feature_util.cpp
+++ b/framework/decode/vulkan_feature_util.cpp
@@ -44,12 +44,9 @@ GFXRECON_BEGIN_NAMESPACE(feature_util)
 // enable the extension with no support present on the replay device (usually because the layer is
 // no longer there)
 std::set<std::string> kIgnorableExtensions = {
-    VK_EXT_TOOLING_INFO_EXTENSION_NAME,
-    VK_EXT_DEBUG_MARKER_EXTENSION_NAME,
-    "VK_ANDROID_frame_boundary",
-    "VK_EXT_frame_boundary",
-    VK_EXT_METAL_SURFACE_EXTENSION_NAME,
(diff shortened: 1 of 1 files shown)


MarkY-LunarG (commented):
Did you run clang-format?  I'm surprised it would accept the table as it is laid out in your changes.

fabian-lunarg:
> Did you run clang-format? I'm surprised it would accept the table as it is laid out in your changes.

yes strange, but that's coming from CLion's builtin clang-format.

bradgrantham-lunarg:
Why would macOS need this but the other WSI extensions wouldn't?  Should we either put all the WSI extensions in here or just leave the macOS ones out and let people provide `--remove-unsupported` like they have for running a capture from one platform on another WSI in the past?

fabian-lunarg:
> Why would macOS need this but the other WSI extensions wouldn't? Should we either put all the WSI extensions in here or just leave the macOS ones out and let people provide `--remove-unsupported` like they have for running a capture from one platform on another WSI in the past?

yes, I agree. we should have here either all WSI-extension (and making usage of `--remove-unsupported` obsolete for WSI-stuff) or none at all. we also got an option `--wsi auto` (it does not seem to work though) that 'could' be used to express "I don't care about WSI etc." and would select the 'correct' WSI backend for the platform and also act like `--remove-unsupported` wrt. WSI-extensions.

guess best is to close here with 'won't do' and phrase the question differently:
should we implement a different (default)-behavior, do we e.g. want to be more tolerant with mismatching WSI/extensions without explicit parameters?

PR #1588, opened 2024-06-17
openxr: Add openxr replay support to Session

jzulauf-lunarg:
Narrow path implementation. Only entry-points used by a single test capture.  Android build not updated.


jzulauf-lunarg:
@MarkY-LunarG  -- builds on Win and Linux not sure what is needed to build on Android ...

PR #1589, opened 2024-06-17
use NDK 26

bradgrantham-lunarg:


bradgrantham-lunarg:
Somewhere between NDK 21 and 24 we introduced a performance regression in WriteState and WriteResourceMemoryState in particular.  A developer demonstrated a 10x slowdown in some experiments in trimmed capture from commits before 2d372404 to commits including and after 2d372404.  Preliminary tests on indicate NDK 26 restores trimmed capture performance to levels before 2d372404.  Suspects include this line from the r25b release notes:
> [Issue 1740](https://github.com/android/ndk/issues/1740): Fixed the legacy toolchain when using CMake's Release build configuration. Since r23b it has not be receiving any optimization flag. It will now receive -O3. If you're building with AGP and haven't overridden AGP's default CMake modes, this change does not affect you, as AGP uses RelWithDebInfo by default.

ziga-lunarg (approved):

MarkY-LunarG (approved):

mikes-lunarg (approved):

PR #1590, opened 2024-06-18
Fix handle mapping/unwrapping for acceleration structure building

marius-pelegrin-arm:
The mapping/unwrapping of handles contained in `VkAccelerationStructureBuildGeometryInfoKHR` was not done correctly for multiple reasons:
- The field `VkAccelerationStructureBuildGeometryInfoKHR::ppGeometries` was not correctly handled because of the double pointer
- The type `VkAccelerationStructureGeometryDataKHR` was ignored because it is a union, even though one of the fields of the union can contains handles to map/unwrap in the pNext chain
- VkMicromapEXT was not entirely handled

I tried to fix this in the most clean way to be able to map/unwrap correctly opacity micromaps

### DIFF
--- a/framework/decode/custom_vulkan_struct_handle_mappers.cpp
+++ b/framework/decode/custom_vulkan_struct_handle_mappers.cpp
@@ -113,5 +113,60 @@ void MapStructHandles(Decoded_VkWriteDescriptorSet* wrapper, const VulkanObjectI
     }
 }
 
+void MapStructHandles(Decoded_VkAccelerationStructureGeometryKHR* wrapper,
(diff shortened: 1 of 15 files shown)


fabian-lunarg (approved):

charles-lunarg (approved):
I concur with Fabian - the acceleration structures need custom logic to handle encode/decode.

While it is theoretically possible to code-gen it, having done such logic before it is not fun nor very robust - future API's with similarly complex patterns will invariably need the codegen to be tweaked, so manually writing it is a better option.

PR #1597, opened 2024-06-21
Handle soft signal interrupts during file processing

per-mathisen-arm:
These signals are sometimes randomly encountered especially on networked filesystems and should be handled by resuming the file operation, rather than aborting or causing errors later during file processing.

The Windows code was not touched because I do not really know how these functions work on that platform.

### DIFF
--- a/framework/decode/dx12_dump_resources.cpp
+++ b/framework/decode/dx12_dump_resources.cpp
@@ -1731,7 +1731,7 @@ void DefaultDx12DumpResourcesDelegate::WriteBlockEnd()
     // Dominates profiling (2/2):
     const std::string block =
         json_data_.dump(json_options_.format == util::JsonFormat::JSONL ? -1 : util::kJsonIndentWidth);
-    util::platform::FileWriteNoLock(block.data(), sizeof(std::string::value_type), block.length(), json_file_handle_);
+    util::platform::FileWriteNoLock(block.data(), block.length() * sizeof(std::string::value_type), json_file_handle_);
     util::platform::FileFlush(json_file_handle_); /// @todo Implement a FileFlushNoLock() for all platforms.
 }
 
@@ -1743,9 +1743,9 @@ bool DefaultDx12DumpResourcesDelegate::WriteBinaryFile(const std::string&       
     FILE* file_output = nullptr;
     if (util::platform::FileOpen(&file_output, filename.c_str(), "wb") == 0)
     {
-        util::platform::FileWrite(data.data() + offset, size, 1, file_output);
+        bool success = util::platform::FileWrite(data.data() + offset, size, file_output);
         util::platform::FileClose(file_output);
-        return true;
+        return success;
     }
     return false;
 }
--- a/framework/decode/file_processor.cpp
+++ b/framework/decode/file_processor.cpp
@@ -467,9 +467,12 @@ bool FileProcessor::ReadCompressedParameterBuffer(size_t  compressed_buffer_size
 
 bool FileProcessor::ReadBytes(void* buffer, size_t buffer_size)
 {
-    size_t bytes_read = util::platform::FileRead(buffer, 1, buffer_size, file_descriptor_);
-    bytes_read_ += bytes_read;
-    return (bytes_read == buffer_size);
+    if (util::platform::FileRead(buffer, buffer_size, file_descriptor_))
+    {
+        bytes_read_ += buffer_size;
+        return true;
+    }
+    return false;
 }
 
 bool FileProcessor::SkipBytes(size_t skip_size)
--- a/framework/decode/json_writer.cpp
+++ b/framework/decode/json_writer.cpp
@@ -206,9 +206,9 @@ bool JsonWriter::WriteBinaryFile(const std::string& filename, uint64_t data_size
     FILE* file_output = nullptr;
     if (util::platform::FileOpen(&file_output, filename.c_str(), "wb") == 0)
     {
-        util::platform::FileWrite(data, static_cast<size_t>(data_size), 1, file_output);
+        bool success = util::platform::FileWrite(data, static_cast<size_t>(data_size), file_output);
         util::platform::FileClose(file_output);
-        return true;
+        return success;
     }
     return false;
 }
--- a/framework/util/json_util.cpp
+++ b/framework/util/json_util.cpp
@@ -352,15 +352,10 @@ static bool WriteBinaryFile(const std::string& filename, uint64_t data_size, con
     bool  written_all = false;
     if (util::platform::FileOpen(&file_output, filename.c_str(), "wb") == 0)
     {
-        const uint64_t written = util::platform::FileWrite(data, 1, static_cast<size_t>(data_size), file_output);
-        if (written >= data_size)
+        written_all = util::platform::FileWrite(data, static_cast<size_t>(data_size), file_output);
+        if (!written_all)
         {
-            written_all = true;
-        }
-        else
-        {
-            GFXRECON_LOG_ERROR(
-                "Only wrote %" PRIu64 " bytes of %" PRIu64 " data to file %s.", written, data_size, filename.c_str());
+            GFXRECON_LOG_ERROR("Failed to write %" PRIu64 " bytes to file %s.", data_size, filename.c_str());
         }
         util::platform::FileClose(file_output);
     }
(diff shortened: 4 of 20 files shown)


framework/util/platform.h
```diff
@@ -614,7 +628,14 @@ inline bool StringContains(const char* text, const char* substring)
...
 inline int32_t FilePuts(const char* char_string, FILE* stream)
 {
-    return fputs(char_string, stream);
+    int result = 0;
+    int err    = 0;
+    do
+    {
+        result = fputs(char_string, stream);
+        err    = ferror(stream);
+    } while (result == EOF && (err == EWOULDBLOCK || err == EINTR || err == EAGAIN));
```
MarkY-LunarG, line 637: Should this be result != EOF?

per-mathisen-arm: I have little experience with fputs, but it seems correct. The man page seems to indicate that any condition where it does not write the full desired character will make it return EOF, which would be our case here. So EOF does not mean literal "end of file" (not sure what that would mean for a write operation anyways).

However, there is a more subtle error in my patch, as it does not handle incomplete reads or writes correctly. fputs does not allow you to handle this at all, and I checked Bionic, where it does seem that it can do partial writes (at least theoretically), but it will not tell you how many bytes it successfully output - so no way to cleanly recover. It would seem safer to implement FilePuts in terms of fwrite, which does tell you this and allows you to handle it.

Now that I'm having a second look - can we simplify the interface here? I don't see why we are exposing element count * element size to the caller here, as I don't think the calling code really makes an educated choice in their usage. Solving partial reads and writes is easier if we can manage these in the lowest level code. I also don't understand why we're returning the written/read size, as none of the calling code actually do anything useful with it, and we could just as well just be returning a bool.

MarkY-LunarG: The modifications you are suggesting sound good to me.  Let me know when you have them.  That is weird, BTW, about how that function returns EOF, but thanks for verifying.

per-mathisen-arm:
This PR now contains two commits. The first implements proper support for partial reads and writes, but making the code quite convoluted and ugly. The next commit cleans this up by removing the leak of low-level complexity from the lowest level code into higher level code where this complexity is merely ignored anyways.

MarkY-LunarG:
@per-mathisen-arm , did you mean to close this?

per-mathisen-arm:
No, closing wasn't intended. Fixed. Thanks :-)

framework/decode/json_writer.cpp
```diff
@@ -204,9 +204,9 @@ std::string JsonWriter::GenerateFilename(const std::string_view filename)
 bool JsonWriter::WriteBinaryFile(const std::string& filename, uint64_t data_size, const uint8_t* data)
 {
     FILE* file_output = nullptr;
-    if (util::platform::FileOpen(&file_output, filename.c_str(), "wb") == 0)
+    if (util::platform::FileOpen(&file_output, filename.c_str(), "wb") == 0 &&
+        util::platform::FileWrite(data, static_cast<size_t>(data_size), file_output))
```
MarkY-LunarG, line 208: Aren't there scenarios where this leave open the file if the write fails for some error other than the ones you expect?

framework/decode/dx12_dump_resources.cpp
```diff
@@ -1595,7 +1595,7 @@ void Dx12DumpResources::WriteBlockEnd()
     // Dominates profiling (2/2):
     const std::string block =
         json_data_.dump(json_options_.format == util::JsonFormat::JSONL ? -1 : util::kJsonIndentWidth);
-    util::platform::FileWriteNoLock(block.data(), sizeof(std::string::value_type), block.length(), json_file_handle_);
+    util::platform::FileWriteNoLock(block.data(), block.length(), json_file_handle_);
```
MarkY-LunarG, line 1598: Should this  be `block.length() * sizeof(std::string::value_type)` for wide characters?

framework/util/json_util.cpp
```diff
@@ -352,15 +352,9 @@ static bool WriteBinaryFile(const std::string& filename, uint64_t data_size, con
     bool  written_all = false;
     if (util::platform::FileOpen(&file_output, filename.c_str(), "wb") == 0)
     {
-        const uint64_t written = util::platform::FileWrite(data, 1, static_cast<size_t>(data_size), file_output);
-        if (written >= data_size)
+        if (!util::platform::FileWrite(data, static_cast<size_t>(data_size), file_output))
```
MarkY-LunarG, line 355: `written_all` is no longer set to true.

MarkY-LunarG (commented):
A few questions and at least one bug I saw.

framework/util/json_util.cpp
MarkY-LunarG: Latest changes seem to fix this.

framework/decode/dx12_dump_resources.cpp
MarkY-LunarG: Latest changes seem to fix this.

framework/decode/json_writer.cpp
MarkY-LunarG: Latest changes seem to fix this.

per-mathisen-arm:
Any clue why the CI is failing here?

framework/decode/file_processor.cpp
```diff
@@ -485,9 +485,12 @@ bool FileProcessor::ReadCompressedParameterBuffer(size_t  compressed_buffer_size
 
 bool FileProcessor::ReadBytes(void* buffer, size_t buffer_size)
 {
-    size_t bytes_read = util::platform::FileRead(buffer, 1, buffer_size, file_descriptor_);
-    bytes_read_ += bytes_read;
-    return (bytes_read == buffer_size);
+    if (util::platform::FileRead(buffer, buffer_size, file_descriptor_))
```
bradgrantham-lunarg, line 488: This changes the behavior of this function - if FileRead read short, previously `bytes_read` would be updated even though it reported `false`.  With this change, `bytes_read` won't be updated.  So if a caller was okay with FileProcessor::ReadBytes making a short read for some reason and expecting `bytes_read` to be updated, that won't work with this change.

bradgrantham-lunarg:
> Any clue why the CI is failing here?

Our D3D12HelloTriangle capture file fails to replay.  We can send you a copy.  What's the best way to get it to you?

framework/decode/file_processor.cpp
per-mathisen-arm: FileRead will now no longer ever read short, however. When called, it either fails or does a full read. This is guaranteed by the full item call to fread (ie calling it with bytes item size and 1 item count). Same for FileWrite. This greatly simplifies calling code.

framework/util/platform.h
```diff
@@ -612,24 +626,38 @@ inline bool StringContains(const char* text, const char* substring)
...
+inline bool FileRead(void* buffer, size_t bytes, FILE* stream)
 {
-    return fread(buffer, element_size, element_count, stream);
+    size_t read_count = 0;
+    int    err        = 0;
+    do
+    {
+        read_count = fread(buffer, bytes, 1, stream);
+        err        = ferror(stream);
+    } while (!feof(stream) && read_count < 1 && (err == EWOULDBLOCK || err == EINTR || err == EAGAIN));
```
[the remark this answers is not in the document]
per-mathisen-arm: Oh, that is an interesting if very subtle corner case. I can change all the return conditions to include "|| bytes == 0".

per-mathisen-arm:
> Our D3D12HelloTriangle capture file fails to replay. We can send you a copy. What's the best way to get it to you?

I think we managed to reproduce and fix the issue here already.

per-mathisen-arm:
(BTW, not sure what I'm doing that keeps automatically closing this ticket.)

per-mathisen-arm:
Anything more that needs to be done here to merge?

charles-lunarg:
Uhhhh I did NOT mean to close the PR - that happened automatically when I force pushed after rebasing on dev so that latest CI would run and we can get this merged - I didn't see anything unresolved or keeping it from being merged.

charles-lunarg:
What I believe has happened is 2 things. Firstly, because the branch being merged was named 'dev', it seems that github closed the PR whenever per-mathisen-arm/dev was pushed to match LunarG/dev. Second, in my attempt to push to your fork with a rebased dev in order to run CI, I accidentally pushed LunarG/dev. Since github closed the PR immediately, I do not have the permission to 'fix' my mistake by pushing with the actual commits. 
Instead, I have created a PR to per-mathisen-arm which, once merged, should reopen this PR.

charles-lunarg (approved):

PR #1598, opened 2024-06-21
openxr: Add replay support through EndFrame

jzulauf-lunarg:
Correctly follows state machine and call ordering through first attempted present.


MarkY-LunarG (approved):
A lot of this will be replaced by my changes, but if it helps you for now.  I think you  should put it in.

PR #1600, opened 2024-06-21
Add missing shadow fences from VkSwapchainPresentFenceInfoEXT

### LINKED ISSUE #1599: VulkanSwapchain impls need to handle VkSwapchainPresentFenceInfoEXT in pNext chain in QueuePresent methods
Some content adds a fence signal to Present using the VkSwapchainPresentFenceInfoEXT extension struct - subsequent GetFenceStatus on that fence never returns SUCCESS and WaitForFences would also never return SUCCESS.

Concrete implementations of VulkanSwapchain need to search for VkSwapchainPresentFenceInfoEXT, and if found need to either
* pass off in the case of the current surface maching `surface_index`, or
* probably(?) immediately signal the fence if the surface is not being drawn (and therefore content is discarded but waits should  not block

ziga-lunarg:
Fences can be signaled with VkSwapchainPresentFenceInfoEXT through vkQueuePresent, and if we are replaying with --surface-index the surface used by vkQueuePresent might not be used. Therefore we have to shadow signal the fence.

vkGetFenceStatus must also check for shadow fences

Closes #1599

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -3235,6 +3235,11 @@ VkResult VulkanReplayConsumerBase::OverrideGetFenceStatus(PFN_vkGetFenceStatus f
         return result;
     }
 
+    if (shadow_fences_.find(fence) != shadow_fences_.end())
+    {
+        return result;
+    }
+
     // If you find this loop to be infinite consider adding a limit in the same way
     // it is done for GetEventStatus and GetQueryPoolResults.
     do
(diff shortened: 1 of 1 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -6244,6 +6249,21 @@ VulkanReplayConsumerBase::OverrideQueuePresentKHR(PFN_vkQueuePresentKHR
                 }
             }
         }
+        const auto* present_fence_info =
```
bradgrantham-lunarg, line 6252: This will only find VkSwapchainPresentFenceInfoEXT if it's the first extension struct in the chain.  Could you descend the chain just to be sure?

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_replay_consumer_base.cpp
ziga-lunarg: Not sure if I'm missing something, but `GetPNextMetaStruct` goes through all pNext

bradgrantham-lunarg: Nope, you were not missing anything and today I learned something new.

bradgrantham-lunarg (approved):

MarkY-LunarG (approved):
LGTM

mikes-lunarg (approved):

PR #1601, opened 2024-06-25
Change DX12 dump resource json output header so it is consistant with Vulkan

davidlunarg:
Also changed the name of the output json file to < capturefilename >_dr.json for both DX12 and Vulkan. It previously was < capturefilename >_rd.json.

### DIFF
--- a/framework/decode/dx12_dump_resources.cpp
+++ b/framework/decode/dx12_dump_resources.cpp
@@ -75,9 +75,9 @@ void Dx12DumpResources::StartDump(ID3D12Device* device, const std::string& captu
     {
         json_filename_ = json_filename_.substr(0, ext_pos);
     }
-    json_options_.data_sub_dir = json_filename_;
-    json_options_.root_dir     = util::filepath::GetBasedir(json_filename_);
     json_filename_ += "_dr." + util::get_json_format(json_options_.format);
+    json_options_.data_sub_dir = util::filepath::GetFilenameStem(json_filename_);
+    json_options_.root_dir     = util::filepath::GetBasedir(json_filename_);
 
     util::platform::FileOpen(&json_file_handle_, json_filename_.c_str(), "w");
 
@@ -1431,7 +1431,16 @@ void Dx12DumpResources::WriteResource(nlohmann::ordered_json&   jdata,
         return;
     }
 
-    std::string file_name = prefix_file_name + "res_id_" + std::to_string(resource_data->source_resource_id);
+    std::string file_name = prefix_file_name;
+    if (prefix_file_name.empty() || prefix_file_name.back() == '_')
+    {
+        file_name += "res_id_";
+    }
+    else
+    {
+        file_name += "_res_id_";
+    }
+    file_name += std::to_string(resource_data->source_resource_id);
 
     util::FieldToJson(jdata["res_id"], resource_data->source_resource_id, json_options_);
     std::string json_path = suffix.empty() ? "file" : suffix + "File";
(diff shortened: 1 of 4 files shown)


framework/decode/dx12_dump_resources.cpp
```diff
@@ -60,31 +68,33 @@ void Dx12DumpResources::StartDump(ID3D12Device* device, const std::string& filen
...
         json_filename_ = json_filename_.substr(0, ext_pos);
     }
-    json_filename_ += "_resources_submit_" +
-                      std::to_string(track_dump_resources_.target.dump_resources_target.submit_index) + "_command_" +
-                      std::to_string(track_dump_resources_.target.dump_resources_target.command_index) + "_drawcall_" +
-                      std::to_string(track_dump_resources_.target.dump_resources_target.drawcall_index) + "." +
-                      util::get_json_format(json_options_.format);
-
-    json_options_.data_sub_dir = util::filepath::GetFilenameStem(json_filename_);
+    json_options_.data_sub_dir = json_filename_;
```
davidd-lunarg, line 78: This should be `json_options_.data_sub_dir = util::filepath::GetFilenameStem(json_filename_);`

framework/decode/dx12_dump_resources.cpp
```diff
@@ -60,31 +68,33 @@ void Dx12DumpResources::StartDump(ID3D12Device* device, const std::string& filen
...
     json_options_.root_dir     = util::filepath::GetBasedir(json_filename_);
+    json_filename_ += "_dr." + util::get_json_format(json_options_.format);
```
davidd-lunarg, line 80: The line
`json_filename_ += "_dr." + util::get_json_format(json_options_.format);`
Should go above the line
`json_options_.data_sub_dir = util::filepath::GetFilenameStem(json_filename_);`

Otherwise capture filename parts may get cut off. For example:
`CapturedApp.exe_20240625T123147_dr_shader_resourceres_id_28_sub_0_before.bin`
becomes
`CapturedApp_shader_resourceres_id_28_sub_0_before.bin`
when the capture file is named `CapturedApp.exe_20240625T123147.gfxr`

framework/decode/dx12_dump_resources.cpp
```diff
@@ -1421,10 +1431,10 @@ void Dx12DumpResources::WriteResource(nlohmann::ordered_json&   jdata,
         return;
     }
 
-    std::string file_name = prefix_file_name + "_res_id_" + std::to_string(resource_data->source_resource_id);
+    std::string file_name = prefix_file_name + "res_id_" + std::to_string(resource_data->source_resource_id);
```
davidd-lunarg, line 1434: I think the underscore should not be removed. Filenames should be:
`...shader_resource_res_id_...`
instead of
`...shader_resourceres_id_...`

Same for the other places `_res_id_` was changed to `res_id_`

davidd-lunarg (changes requested):

framework/decode/dx12_dump_resources.cpp
davidlunarg: But sometimes prefix_file_name is null, resulting in two underscores before res_id in the filename. Should I check for a null prefix_file_name?

davidd-lunarg: That value shouldn't be null. I have a bug fix that builds on this PR that should resolve that so we shouldn't get the double underscore in the name.

davidlunarg: Not sure it this change is the cause, but I'm getting file names with consecutive underscores, i.e. D3D12Multithreading.exe_trim_trigger_20240423T120337_dr__res_id_20_sub_0_.bin

davidd-lunarg: It was an existing problem. It also resulted in some missing files in the json and/or missing dumped .bin files.

davidd-lunarg (approved):

PR #1602, opened 2024-06-25
Fix bug in which no .bin files are created when doing a dx12 dump resources

davidlunarg:
Files would not be created when the full path was specified for the capture file on the command line.


davidlunarg:
Abandoning this PR. It has two commits, one of them is a duplicate of a previous commit.

PR #1604, opened 2024-06-26
uffd: Do pthread_cond_timedwait instead

panos-lunarg:
It is possible that the signal send by the uffd logic to block the applications threads cannot be delivered. For this reason do pthread_cond_timedwait on the condition variable instead of pthread_cond_wait which causes a deadlock in these cases.

### DIFF
--- a/framework/util/page_guard_manager_uffd.cpp
+++ b/framework/util/page_guard_manager_uffd.cpp
@@ -31,6 +31,7 @@
 #include <cinttypes>
 #include <sys/syscall.h>
 #include <sys/ioctl.h>
+#include <time.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <poll.h>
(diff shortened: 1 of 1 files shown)


framework/util/page_guard_manager_uffd.cpp
```diff
@@ -727,13 +737,27 @@ uint32_t PageGuardManager::UffdBlockFaultingThreads()
     {
         while (blocked_threads < threads_to_block)
         {
-            ret = pthread_cond_wait(&wait_for_threads_cond, &wait_for_threads_mutex);
+            struct timespec timeout;
+            if (clock_gettime(CLOCK_REALTIME, &timeout))
```
bradgrantham-lunarg, line 741: Does this add another system call in this thread which may add a place for the kernel scheduler to pre-empt this thread?

panos-lunarg: Yes but I think it's ok. This is not the same as the page guard watcher thread which we want to be "fast".

bradgrantham-lunarg: 👍

bradgrantham-lunarg (approved):

PR #1605, opened 2024-06-26
dx dump: fix some res files not being written

davidd-lunarg:


davidlunarg:
I like the more descriptive file names. And the new items in the json output file. Looks good.

PR #1607, opened 2024-06-27
Asynchronous Pipeline Creation

fabian-lunarg:
this PR adds a threadpool and uses that to compile graphics-pipelines asynchronously.

- Add Threadpool to utils
- Add a generic `vulkan_struct_deep_copy` to create required deep-copies of input structures
- Add a type-trait `has_handle_future` to add the option to retrieve/sync async-handles via std::future,
  -  currently that's only `VkPipeline` (but should also include `VkShaderEXT`)
- Add `AddHandleArrayAsync` routine to handle-mapping API
- Add tracking and deferred deletion of handles that are used as dependencies during asynchronous tasks
-  currently only supports `vkCreateGraphicsPipelines`
- default: do not use any asynchronous operation or threadpool unless explicitly requested:
  - Add CLI-option `--pipeline-creation-jobs` or `--pcj` to specify number of background-threads
  - negative numbers yield (MAX_THREADS - arg), e.g. `gfxrecon-replay --pcj -1` uses 15 cores on a 16-core-CPU

missing things (in a follow-up):
- extend support to `vkCreateComputePipelines`, `vkCreateRaytracingPipelines`, `vkCreateShadersEXT`

### DIFF
--- a/framework/decode/vulkan_handle_mapping_util.h
+++ b/framework/decode/vulkan_handle_mapping_util.h
@@ -51,13 +51,21 @@ static typename T::HandleType MapHandle(format::HandleId             id,
         if (info != nullptr)
         {
             handle = info->handle;
+
+            if constexpr (has_handle_future_v<T>)
+            {
+                if (info->handle == VK_NULL_HANDLE && info->future.valid())
+                {
+                    const auto& [result, async_handles] = info->future.get();
+                    handle                              = async_handles[info->future_handle_index];
+                }
+            }
         }
         else
         {
             GFXRECON_LOG_WARNING("Failed to map handle for object id %" PRIu64, id);
         }
     }
-
     return handle;
 }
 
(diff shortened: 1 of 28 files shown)


framework/decode/vulkan_handle_mapping_util.h
```diff
@@ -51,13 +51,21 @@ static typename T::HandleType MapHandle(format::HandleId             id,
...
             handle = info->handle;
+
+            if constexpr (has_handle_future_v<T>)
+            {
+                if (info->handle == VK_NULL_HANDLE && info->future.valid())
+                {
+                    const auto& [result, async_handles] = info->future.get();
+                    handle                              = async_handles[info->future_handle_index];
+                }
+            }
```
fabian-lunarg, line 62: @MarkY-LunarG - I wanted to get your opinion on those handle-mapping additions (I know you got a PR refactoring those a bit, move to another file etc.)

nickdriscoll-lunarg (approved):

PR #1616, opened 2024-07-08
Uffd restart interrupted system calls

mikes-lunarg:
I've been looking at some userfaultfd crashes in the Android testing spreadsheet, specifically "row 20" since the crash is easy to reproduce and gives you enough time to attach a debugger.

I thought maybe `UffdBlockFaultingThreads()` wasn't blocking enough threads since it seems like there is a potential race condition during the first access by a thread, before it gets added to the `uffd_fault_causing_threads` list. However, preemptively adding more threads to `uffd_fault_causing_threads` actually made things worse, crashing almost immediately after the extra thread IDs were added. Then I tried disabling `UffdBlockFaultingThreads` and the crashes went away, but there were lots of graphical glitches in another title (makes sense, that synchronization was in there for a reason).

I was able to take an strace and saw this for one of the crashing threads during the RT signal handler:
```
9272  futex(0x74b322a930, FUTEX_WAIT_BITSET_PRIVATE, 4294967294, NULL, FUTEX_BITSET_MATCH_ANY <unfinished ...>
9200  tgkill(9093, 9272, SIGRT_9 <unfinished ...>
9272  <... futex resumed> )             = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
9272  --- SIGRT_9 {si_signo=SIGRT_9, si_code=SI_TKILL, si_pid=9093, si_uid=10289} ---
  <more syscalls inside the signal handler>
9272  rt_sigreturn({mask=[QUIT USR1 PIPE RTMIN]} <unfinished ...>
9272  <... rt_sigreturn resumed> )      = -1 EINTR (Interrupted system call)
9272  --- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_MAPERR, si_addr=0x10} ---
```
I think this is what is happening:

- Thread 9272 makes a futex system call
- Thread 9200 touches a faulted page, triggering an RT signal to Thread 9272 from UffdBlockFaultingThreads because it has also touched a faulted page before
- Thread 9272 runs the RT signal handler, interrupting the futex call
- Thread 9272 resumes the futex call and returns EINTR because it was interrupted
- Something bad happens, maybe the app didn't expect EINTR? Maybe that somehow results in it dereferencing a null pointer? This is the part I am least sure about
- Seg fault

There's a nice hint in the strace log about `SA_RESTART`, so I tried it and it worked! No more crash, no more corruption.

And here is the updated strace:
```
10631 futex(0x74b322b770, FUTEX_WAIT_BITSET_PRIVATE, 4294967294, NULL, FUTEX_BITSET_MATCH_ANY <unfinished ...>
10584 tgkill(10493, 10631, SIGRT_9 <unfinished ...>
10631 <... futex resumed> )             = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
10631 --- SIGRT_9 {si_signo=SIGRT_9, si_code=SI_TKILL, si_pid=10493, si_uid=10289} ---
  <more syscalls inside the signal handler>
10631 rt_sigreturn({mask=[QUIT USR1 PIPE RTMIN]}) = 501221603184
10631 futex(0x74b322b770, FUTEX_WAIT_BITSET_PRIVATE, 4294967294, NULL, FUTEX_BITSET_MATCH_ANY <unfinished ...>
10584 tgkill(10493, 10631, SIGRT_9 <unfinished ...>
10631 <... futex resumed> )             = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
10631 --- SIGRT_9 {si_signo=SIGRT_9, si_code=SI_TKILL, si_pid=10493, si_uid=10289} ---
  <more syscalls inside the signal handler>
10631 rt_sigreturn({mask=[QUIT USR1 PIPE RTMIN]}) = 501221603184
10631 futex(0x74b322b770, FUTEX_WAIT_BITSET_PRIVATE, 4294967294, NULL, FUTEX_BITSET_MATCH_ANY <unfinished ...>
10584 tgkill(10493, 10631, SIGRT_9 <unfinished ...>
10631 <... futex resumed> )             = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
10631 --- SIGRT_9 {si_signo=SIGRT_9, si_code=SI_TKILL, si_pid=10493, si_uid=10289} ---
  <more syscalls inside the signal handler>
10631 rt_sigreturn({mask=[QUIT USR1 PIPE RTMIN]} <unfinished ...>
10631 <... rt_sigreturn resumed> )      = 501221603184
10631 futex(0x74b322b770, FUTEX_WAIT_BITSET_PRIVATE, 4294967294, NULL, FUTEX_BITSET_MATCH_ANY <unfinished ...>
```

This shows the same futex call getting restarted each time it exits the RT handler.

I made this a draft since I would like to do more testing. So far it seems like there isn't really a downside to enabling the `SA_RESTART` flag, especially when interrupting application threads running code we do not control.


framework/util/page_guard_manager_uffd.cpp
```diff
@@ -215,7 +215,8 @@ bool PageGuardManager::UffdSetSignalHandler()
     // Install signal handler for the RT signal
     {
         struct sigaction sa = {};
-        sa.sa_flags         = SA_SIGINFO;
+        // Minimize side effects of the RT signal handler by restarting interrupted system calls when possible
+        sa.sa_flags = SA_RESTART;
         sigemptyset(&sa.sa_mask);
         sa.sa_handler = UffdStaticSignalHandler;
```
panos-lunarg, line 221: I was about to comment that since `SA_SIGINFO` was removed the signal handler should be registered in `sa.sa_sigaction` instead of `sa.sa_handler` but it seems it's the other way around. So it was wrong all the time but now should be correct.

panos-lunarg (approved):

bradgrantham-lunarg (approved):

PR #1621, opened 2024-07-11
uffd: Remove threads that do not exist anymore

panos-lunarg:


mikes-lunarg (approved):
This all looks good. It does make me wonder if the insert operation on `uffd_fault_causing_threads` is properly guarded. An insert on another thread could potentially invalidate the iterator used here.

panos-lunarg:
Right, good catch. Requires some locking

framework/util/page_guard_manager_uffd.cpp
```diff
@@ -710,27 +712,31 @@ uint32_t PageGuardManager::UffdBlockFaultingThreads()
...
 
     threads_to_block = 0;
-    for (const auto& entry : uffd_fault_causing_threads)
+    thread_set_mutex_.lock();
+    for (auto tid = uffd_fault_causing_threads_.begin(); tid != uffd_fault_causing_threads_.end();)
     {
         // Don't send the signal to this thread.
-        if (this_tid == entry)
+        if (this_tid == *tid)
         {
```
ziga-lunarg, line 720: You need a `++tid;` here

ziga-lunarg (changes requested):

framework/util/page_guard_manager_uffd.cpp
panos-lunarg: oops. right

ziga-lunarg (approved):

PR #1634, opened 2024-07-25
Enable marking of api calls injected by the replayer

naomipappe:
For different reasons, it is useful to be able to mark the calls injected during the replay, as "injected". 

In this change, it is done by calling functions that can mark the beginning of injected calls section and the ending of such section. This functions are loaded from a layer, and therefore this change is versatile in regards to what do you do with the calls that are marked as injected.

This change enables this, by adding a new replay option, that can be used to specify the names of the layers (that should be present on `VK_LAYER_PATH`) as a comma separated list. This layers are then loaded when the instance is created, and are setup when the relevant device is created. The callbacks are stored inside a utility class, that is easy to use as a singleton in any file where you need your calls marked. To mark the calls, you should provide `VulkanObjectInfo<T>`, or a handle id of a physical device directly.

### DIFF
--- a/framework/util/marking_layers.h
+++ b/framework/util/marking_layers.h
@@ -62,7 +62,7 @@ class MarkingLayersUtil
         {
             if (!scope_closed_)
             {
-                GFXRECON_LOG_FATAL("Marking was not closed, previous begin call at %s, %s", file_.c_str(), line_);
+                GFXRECON_LOG_FATAL("Marking was not closed, previous begin call at %s, %d", file_.c_str(), line_);
                 GFXRECON_ASSERT(false);
             }
             file_         = file;
--- a/framework/util/marking_layers.h
+++ b/framework/util/marking_layers.h
@@ -1,0 +1,166 @@
+/*
+** Copyright (c) 2024 Valve Corporation
+** Copyright (c) 2024 LunarG, Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and associated documentation files (the "Software"),
+** to deal in the Software without restriction, including without limitation
+** the rights to use, copy, modify, merge, publish, distribute, sublicense,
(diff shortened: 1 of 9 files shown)


bradgrantham-lunarg:
Marked P1 because we want it soon but this ties into future development including incorporating capture into replay.  Additionally I'm unclear why the interface needs to be so flexible - a pair of functions to put before and after one or more Vulkan calls to mark the calls synthetic seems like it would be sufficient.

naomipappe:
Glad to hear your thoughts! By flexible interface you mean that the marking of the calls is essentially tied to some physical device?

framework/util/marking_layers.h
```diff
@@ -0,0 +1,167 @@
...
+    void AddCallbacks(format::HandleId physical_device_id, const VkPhysicalDeviceToolProperties& tool_properties)
+    {
+        // Do not add callbacks for the layers that were not enabled by the respective argument
+        if (layer_names_.count(tool_properties.layer) == 0)
+        {
+            return;
+        }
+        GFXRECON_LOG_DEBUG(
+            "Adding a %s layer callbacks from physical device %" PRIu64, tool_properties.layer, physical_device_id);
+        callbacks_[physical_device_id].emplace_back(*reinterpret_cast<MarkInjectedCallbacks*>(tool_properties.pNext));
```
bradgrantham-lunarg, line 115: Where is it documented in the spec that a tools layer's VkPhysicalDeviceToolProperties::pNext is a callback?

per-mathisen-arm:
As discussed a while ago, the method of layer communication proposed here is just a quick hack, mostly for discussion of what a longer term solution should look like. Probably that would involve defining a new Vulkan extension, either through Khronos or just a private "fake" extension.

fabian-lunarg:
superseded by https://github.com/LunarG/gfxreconstruct/pull/3191

PR #1640, opened 2024-07-29
Fix noop functions address comparison

marius-pelegrin-arm:
It sometimes is necessary to check if a vulkan call in a table is noop or not.
Current prototypes of noop functions include the `static` keyword, making address comparison impossible because the address changes for each translation unit.

These functions are all one-liners with a simple call to a logging message. I propose to replace the `static` keyword with `inline`. The C++ standard specifies that inline functions have the same address across translation units.


lunarpapillo:
The **LunarG** CI failure was apparently due to a hardware GPU failure.  That machine has been taken out of rotation.  I've not restarted the run because the **Windows MSVC Debug** failure means another code change will be required anway.

fabian-lunarg:
the msvc-error here was `error C1128: number of sections exceeded object file format limit: compile with /bigobj`,
indicating that this is something we would need to address project-wide or at least for affected `gfxrecon_decode`  target.

this post https://stackoverflow.com/questions/15110580/penalty-of-the-msvs-compiler-flag-bigobj 
indicates that in 2024 there should be no good reason against using `/bigobj` flag. 
what do you think? @bradgrantham-lunarg @charles-lunarg

charles-lunarg:
API Dump uses Bigobj, and so does Validation Layers (if I remember correctly).

I do not think there is any reason to not use bigobj - this is some CMake code which should enable it for gfxrecon_decode.
We could set it project wide, although I am loathe to do that since there are many targets in the repo (also because setting stuff globally in cmake is not great).
```cmake
if (MSVC)
    target_compile_options(gfxrecon_decode PRIVATE /bigobj)
endif()
```

charles-lunarg:
I found that we already *do* set /bigobj on specific files. I wonder if the issue is that the /bigobj flag isn't set for the entire target instead - ie each .cpp may fall under the limit, but when the library is created by combining the files, it goes over the limit.
I'm working on a commit to fix it (waiting for CI to pass before pushing it here).

charles-lunarg (approved):

PR #1643, opened 2024-07-30
Refactor encode::vulkan_trackers::TrackStruct() routine

fabian-lunarg:
- Built on top of graphics::vulkan_struct_deep_copy routines
- Avoids duplication, reduces code-footprint, less coupling, less memory-allocations


nickdriscoll-lunarg (approved):
These changes look good to me.

PR #1649, opened 2024-08-03
Fix wrong setting name in android usage doc

ziga-lunarg:


mikes-lunarg (approved):
Nice catch, I hope that typo didn’t cause too much trouble

PR #1650, opened 2024-08-05
Check SPIRV/Buffer-References inlined in pNext

fabian-lunarg:
- `VkPipelineShaderStageCreateInfo` 'can' provide a `VkShaderModuleCreateInfo` in pNext
- Add helper-routines to check for that case
- Add a nullptr-check required in certain cases, in presence of asynchronous pipeline-creation (`--pcj X` option)
- Refactoring and doc of `vulkan_struct_extract_handle_ids`

### DIFF
--- a/framework/graphics/vulkan_check_buffer_references.cpp
+++ b/framework/graphics/vulkan_check_buffer_references.cpp
@@ -1,0 +1,70 @@
+/*
+** Copyright (c) 2024 LunarG, Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and associated documentation files (the "Software"),
+** to deal in the Software without restriction, including without limitation
+** the rights to use, copy, modify, merge, publish, distribute, sublicense,
+** and/or sell copies of the Software, and to permit persons to whom the
+** Software is furnished to do so, subject to the following conditions:
+**
+** The above copyright notice and this permission notice shall be included in
+** all copies or substantial portions of the Software.
+**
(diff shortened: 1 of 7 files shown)


framework/graphics/vulkan_check_buffer_references.cpp
```diff
@@ -0,0 +1,72 @@
...
+** DEALINGS IN THE SOFTWARE.
+*/
+
+#include "vulkan_check_buffer_references.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(graphics)
+
+template <>
+void vulkan_check_buffer_references(const VkGraphicsPipelineCreateInfo* create_infos, uint32_t create_info_count)
```
nickdriscoll-lunarg, line 29: Where is the warning issued in these functions? I see it descend the pNext chains, but it doesn't appear to do anything else to me.

fabian-lunarg: in the inner loop it will pass the spirv-blobs to `graphics::vulkan_check_buffer_references(module_create_info->pCode, module_create_info->codeSize);`

fabian-lunarg: https://github.com/LunarG/gfxreconstruct/blob/df9f0ed0fbcf76c5ff6c48aab44ad86c61a57e65/framework/graphics/vulkan_check_buffer_references.h#L51

nickdriscoll-lunarg: My bad. Things look good to me then.

fabian-lunarg:
thanks @NickDriscoll !

PR #1652, opened 2024-08-05
Remove assert from DestroyCommandPool

bartosz-muszarski-arm:
VkDestroyCommandPool specs state that pool can be a null handle, yet its override in replay asserts the pool info struct pointer.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -7932,9 +7932,10 @@ void VulkanReplayConsumerBase::OverrideDestroyCommandPool(
     CommandPoolInfo*                                     pool_info,
     StructPointerDecoder<Decoded_VkAllocationCallbacks>* pAllocator)
 {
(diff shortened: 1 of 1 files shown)


charles-lunarg (approved):
Yep, the vulkan API is quite permissive in that respect.

Seems the change comes from a BIG PR which is an easy place for such assumptions to not be challenged.
https://github.com/LunarG/gfxreconstruct/pull/1608/

The exact commit which adds the null handle checks.
https://github.com/LunarG/gfxreconstruct/pull/1608/commits/24abbfb2ad0bb25f8c96fb2e76eb9e8d17729163

bradgrantham-lunarg (approved):

lunarpapillo:
The CI build failure was due to a misconfigured machine.  I've restarted the run.

lunarpapillo:
And the most recent failure occurred because of a machine reboot, potentially due to update, but possibly a machine crash.  I've requeued the test run and will be watching.

PR #1655, opened 2024-08-07
Rework async-dependency-tracking to be deterministic

fabian-lunarg:
- resolve vkDestroyFoo for tracked async-handles:  sync/block instead of deferred-delete
-> Avoids shuffling of certain vkDestroyFoo calls when using `--pcj X` option


nickdriscoll-lunarg (approved):
Looks good. Nothing jumps out at me as problematic.

PR #1656, opened 2024-08-07
Enable capture of D3D12 apps to work again

rurra-amd:
**Problem**
STL locks were producing a null access violations during initialization of our capture libs. This happens when the capture libs were built with certain versions of Visual Studio. What this means for us is that we crash while attempting to capture with those libs.

**Solution**
Define `_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR`. Documentation is limited but here are some details: https://github.com/microsoft/STL/wiki/Changelog#vs-2022-1710

**Testing**
Can capture apps again when building libs with latest VS. Need more testing.


dgraves:
This change is needed to avoid an incompatibility between GFXReconstruct DLLs built with VS2022 version 17.10 (and later) and some versions of msvcp140.dll. Some apps ship with the incompatible version of msvcp140.dll in the same folder as their executable, which can be deleted or renamed as a workaround, but it would be convenient if this change could be merged.

davidd-lunarg (approved):

PR #1657, opened 2024-08-07
build: Apply /bigobj for encode & decode targets

charles-lunarg:
Rather than adding the /bigobj flag to individual files, apply it to the entire static library on MSVC

### DIFF
--- a/framework/decode/CMakeLists.txt
+++ b/framework/decode/CMakeLists.txt
@@ -262,16 +262,8 @@ target_sources(gfxrecon_decode
 endif()
 
 if (MSVC)
-    # These files may fail to compile with VS2017 and older, requiring the default section limit of 2^16 to be increased.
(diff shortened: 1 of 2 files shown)


nickdriscoll-lunarg (approved):
👍

bradgrantham-lunarg:
Is there a performance impact?

charles-lunarg:
I do not know nor have a good way to validate the performance impact quickly. 

The only tangible drawback to /bigobj appears to be

> Linkers that shipped prior to Visual C++ 2005 cannot read .obj files that were produced with /bigobj.

They also mentioned a 2% increase in binary size, which isn't too surprising due to using uint32 instead of uint16's. But I do not see any mention of performance overhead.

I suppose it doesn't matter as we are hitting this limit regardless - and even if we decrease the object count right now as more things are added to vulkan, we will be back here soon enough.

PR #1658, opened 2024-08-08
Rework vulkan optimizer

bartosz-muszarski-arm:
Current implementation does not allow to implement different kind of optimizations in a modular and scalable way.
This commit established base infrastructure allowing implementation of modules that can process vulkan calls and apply modifications to them.
Gfxrecon-optimize for vulkan uses 2 pass approach:
* first pass to generate optimization data 
* second pass to generate modified trace
Modifications are performed in predefined order for each call separately.
Each modifier is given a pointer to same modifiable parameter buffer, a method to insert new calls before current call and a flag to request removal of current call from trace.

### DIFF
--- a/framework/decode/vulkan_example_modifier.h
+++ b/framework/decode/vulkan_example_modifier.h
@@ -1,22 +1,190 @@
+/*
+** Copyright (c) 2024 LunarG, Inc.
+** Copyright (c) 2024 Arm Limited and/or its affiliates <open-source-office@arm.com>
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and associated documentation files (the "Software"),
+** to deal in the Software without restriction, including without limitation
+** the rights to use, copy, modify, merge, publish, distribute, sublicense,
+** and/or sell copies of the Software, and to permit persons to whom the
+** Software is furnished to do so, subject to the following conditions:
+**
+** The above copyright notice and this permission notice shall be included in
+** all copies or substantial portions of the Software.
+**
+** THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+** IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+** FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+** AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+** FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+** DEALINGS IN THE SOFTWARE.
+*/
+
 #ifndef GFXRECON_DECODE_VULKAN_EXAMPLE_MODIFIER_H
 #define GFXRECON_DECODE_VULKAN_EXAMPLE_MODIFIER_H
 
-#include "decode/referenced_resource_table.h"
-#include "generated/generated_vulkan_consumer.h"
 #include "util/defines.h"
 #include "util/vulkan_modifier_base.h"
-
+#include "encode/struct_pointer_encoder.h"
 #include "vulkan/vulkan.h"
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
+// Example usage of VulkanModifier to add, modify and remove calls from trace.
+// This is just for demonstration purposes.
+// This modifier will use vkSetDebugUtilsObjectNameEXT call to name all compacted copies of
+// other acceleration structures. Performs several operaions:
+// 1. Adds vkSetDebugUtilsObjectNameEXT calls after relevant vkCreateAccelerationStructureKHR
+// 2. Removes existing vkSetDebugUtilsObjectNameEXT calls for objects marked in step 1.
+// 3. Modifies VkCreateInstance call to include VK_EXT_debug_utils in enabled extension list
+// To test this example you need any trace that performs an acceleration structure copy with
+// VK_COPY_ACCELERATION_STRUCTURE_MODE_COMPACT_KHR mode, for example:
+// https://github.com/ARM-software/tracetooltests/blob/main/src/vulkan_as_3.cpp
 class VulkanExampleModifier : public util::VulkanModifierBase
 {
   public:
     VulkanExampleModifier(){};
 
-    virtual bool CanOptimize() override { return true; }
+    virtual bool CanOptimize() override { return compacted_copies_.empty(); }
+
+    // Example call modification
+    // Add debug utils to list of extensions
+    void Process_vkCreateInstance(const ApiCallInfo&                                   call_info,
+                                  VkResult                                             returnValue,
+                                  StructPointerDecoder<Decoded_VkInstanceCreateInfo>*  pCreateInfo,
+                                  StructPointerDecoder<Decoded_VkAllocationCallbacks>* pAllocator,
+                                  HandlePointerDecoder<VkInstance>*                    pInstance)
+    {
+        // Just modification pass, nothing to analyze
+        if (parameter_buffer_)
+        {
+            auto create_info     = pCreateInfo->GetMetaStructPointer()->decoded_value;
+            auto extension_count = create_info->enabledExtensionCount;
+            if (extension_count)
+            {
+                bool debug_utils_enabled = false;
+                for (uint32_t i = 0; i < extension_count; i++)
+                {
+                    std::string extension_name = create_info->ppEnabledExtensionNames[i];
+                    if (extension_name == VK_EXT_DEBUG_UTILS_EXTENSION_NAME)
+                    {
+                        debug_utils_enabled = true;
+                        break;
+                    }
+                }
+                if (!debug_utils_enabled)
+                {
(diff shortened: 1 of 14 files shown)


tools/optimize/vulkan_file_optimizer.cpp
```diff
@@ -0,0 +1,198 @@
...
+    {
+        WriteFunctionCall(new_call->call_id, new_call->thread_id, &(new_call->parameter_buffer));
+    }
+
+    return success;
+}
+
+// TODO: This is the same code used by CaptureManager to write function call data. It could be moved to a format
+// utility.
+void VulkanFileOptimizer::WriteFunctionCall(format::ApiCallId               call_id,
```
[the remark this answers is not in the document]
bartosz-muszarski-arm: Sure, I think we should have a separate PR focused on refactoring. This duplication in particular is a recurring problem as there are several abstractions already that need to write calls to a file. Could use a more generic "FileWriter"

tools/optimize/vulkan_file_optimizer.cpp
```diff
@@ -0,0 +1,198 @@
...
+    std::vector<std::unique_ptr<util::CallModifierBase::NewCallData>> new_post_calls;
+
+    for (auto& modifier : optimization_data_->modifiers)
+    {
+        modifier->SetParameterBuffer(&buffer);
+        decoder.AddConsumer(modifier.get());
+        decode::DecodeAllocator::Begin();
+        decoder.DecodeFunctionCall(call_id, call_info, buffer.GetData(), buffer.GetDataSize());
+        decode::DecodeAllocator::End();
+        decoder.RemoveConsumer(modifier.get());
```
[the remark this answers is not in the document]
bartosz-muszarski-arm: The problem here was multiple modifiers might want to change the call parameters, so the values of buffer.GetData() and GetDataSize() might change after every modifier pass. So we can't really have a decoder with multiple consumers as the architecture suggests, but it can still work if the consumers are added and removed in a loop like that.
The call gets decoded, then is processed by consumer, the consumer modifies the parameter buffer by encoding modified data into it, then the next consumer receives modified buffer that needs to be decoded again.
Alternatively we could define a new consumer class that allows for modification of the calls parameters once decoded and avoid this decode-encode loop.

fabian-lunarg:
I'm currently investigating the failed CI-run from last week. it looks many tests related to `dump_resources` feature fail, due to different json out. this might require adjusting our ground-truth, let me come back for this

framework/decode/vulkan_example_modifier.h
```diff
@@ -0,0 +1,193 @@
...
+// This is just for demonstration purposes.
+// This modifier will use vkSetDebugUtilsObjectNameEXT call to name all compacted copies of
+// other acceleration structures. Performs several operaions:
+// 1. Adds vkSetDebugUtilsObjectNameEXT calls after relevant vkCreateAccelerationStructureKHR
+// 2. Removes existing vkSetDebugUtilsObjectNameEXT calls for objects marked in step 1.
+// 3. Modifies VkCreateInstance call to include VK_EXT_debug_utils in enabled extension list
+// To test this example you need any trace that performs an acceleration structure copy with
+// VK_COPY_ACCELERATION_STRUCTURE_MODE_COMPACT_KHR mode, for example:
+// https://github.com/ARM-software/tracetooltests/blob/main/src/vulkan_as_3.cpp
+class VulkanExampleModifier : public util::VulkanModifierBase
```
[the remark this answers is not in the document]
bartosz-muszarski-arm: It's meant for demonstration only, to showcase how the optimization interface was intended to be used, thus the name. Has to real use other than that, can be removed completely.

tools/optimize/vulkan_file_optimizer.cpp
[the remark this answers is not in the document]
fabian-lunarg: ok, I can see why you duplicated + adapted this `VulkanStateWriter::WriteFunctionCall`. it should be a shared utility but is somewhat entangled with other class internals. I'll create a separate issue to refactor.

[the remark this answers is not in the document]
fabian-lunarg: ok, nevermind. I added a commit consolidating that part and avoid duplication.

tools/optimize/main.cpp
```diff
@@ -113,103 +116,87 @@ static void PrintUsage(const char* exe_name)
...
         {
-            GFXRECON_WRITE_CONSOLE("File did not contain trim state setup - no optimization was performed");
-            gfxrecon::util::Log::Release();
-            exit(65);
+            throw std::runtime_error("Failed to scan input file for optimizations");
         }
-        else if ((file_processor.GetCurrentFrameNumber() > 0) &&
-                 (file_processor.GetErrorState() == gfxrecon::decode::FileProcessor::kErrorNone))
+
+        // TODO: resref_consumer should derive from VulkanModifierBase and be handled like other
```
fabian-lunarg, line 156: we should add a clear condition to the TODO (or remove it). when will it be removed? like a specific date or best a referenced issue/PR-#

framework/decode/vulkan_example_modifier.h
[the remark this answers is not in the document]
fabian-lunarg: but then, I see another issue. usage of `VulkanExampleModifier` has been hard-coded into 'tools/optimize/main.cpp'

tools/optimize/main.cpp
```diff
@@ -113,103 +116,87 @@ static void PrintUsage(const char* exe_name)
...
 {
-    GFXRECON_ASSERT(unreferenced_ids != nullptr);
+    auto result = std::make_unique<gfxrecon::VulkanFileOptimizer::VulkanOptimizationData>();
 
     gfxrecon::decode::FileProcessor file_processor;
     if (file_processor.Initialize(input_filename))
     {
         gfxrecon::decode::VulkanDecoder                    decoder;
         gfxrecon::decode::VulkanReferencedResourceConsumer resref_consumer;
+        auto example_modifier = std::make_unique<gfxrecon::decode::VulkanExampleModifier>();
```
fabian-lunarg, line 143: usage of `VulkanExampleModifier` is hard-coded here and should not be, since it's a production codebase/tool.

when we remove that addition I do not see a clear way how modifiers can be added.
maybe I'm missing something, but what is the concept here, like e.g. will there be cmd-line parameters to `gfxrecon-optimize` to add/select modifiers?

tools/optimize/main.cpp
bartosz-muszarski-arm: Will remove it, this no longer applies

tools/optimize/vulkan_file_optimizer.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+
+bool VulkanFileOptimizer::ProcessFunctionCall(const format::BlockHeader& block_header, format::ApiCallId call_id)
+{
+    // Regular funciton call processing/compression handling - duplicated in multiple places, TODO: extract, reuse
+    size_t              parameter_buffer_size = static_cast<size_t>(block_header.size) - sizeof(call_id);
+    uint64_t            uncompressed_size     = 0;
+    decode::ApiCallInfo call_info{ GetCurrentBlockIndex() };
+    bool                success = ReadBytes(&call_info.thread_id, sizeof(call_info.thread_id));
```
fabian-lunarg, line 35: `success` is not checked after reading here

tools/optimize/vulkan_file_optimizer.cpp
```diff
@@ -0,0 +1,142 @@
...
+*/
+
+#include "tools/optimize/vulkan_file_optimizer.h"
+#include "framework/format/format_util.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+
+bool VulkanFileOptimizer::ProcessFunctionCall(const format::BlockHeader& block_header, format::ApiCallId call_id)
+{
+    // Regular funciton call processing/compression handling - duplicated in multiple places, TODO: extract, reuse
```
fabian-lunarg, line 31: typo in comment and unclear TODO. either remove or add more information?

tools/optimize/vulkan_file_optimizer.h
```diff
@@ -0,0 +1,60 @@
...
+class VulkanFileOptimizer : public FileOptimizer
+{
+  public:
+    struct VulkanOptimizationData
+    {
+        std::unordered_set<gfxrecon::format::HandleId>         unreferenced_ids;
+        std::vector<std::unique_ptr<util::VulkanModifierBase>> modifiers;
+    };
+
+    explicit VulkanFileOptimizer(VulkanOptimizationData* optimization_data) :
```
fabian-lunarg, line 43: `VulkanOptimizationData*` is passed as raw-pointer, which is then stored as member.
better pass by value and move into place?

bartosz-muszarski-arm: You mean transfer the ownership of optimization_data inside VulkanFileOptimizer? I don't mind, but also don't see the benefit. The purpose here was to avoid copies that may become costly as the optimization data grows. Caller maintains ownership, consumer gets the input as raw ptr.

fabian-lunarg: `std::unique_ptr<gfxrecon::VulkanFileOptimizer::VulkanOptimizationData>
GetVulkanOptimizationData(...)` 

returns a unique-pointer.
so yes, I think it would be better to transfer ownership of that instead of using a raw-pointer.
there would be no copy and the benefit would be to avoid a potentially dangling pointer in `optimization_data_`

bartosz-muszarski-arm: Returning unique ptr here does not generate a copy

MarkY-LunarG:
Closing for now

PR #1660, opened 2024-08-12
State files

panos-lunarg:


USAGE_android.md
```diff
@@ -306,6 +306,9 @@ option values.
...
+| Use asset file                                 | debug.gfxrecon.capture_android_dump_assets                         | BOOL    | Setting this triggers a dump of all assets into the asset file. Since android options cannot be set by the layer, dumping is done whenever this option switches between from `false` to `true` or from `true` to `false`. Default is: `false`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     |
```
csteinberg-samsung, line 310: should this be "Dump asset file" instead of "Use asset file"?

csteinberg-samsung: I also wonder if this should only trigger on the false -> true setting?

USAGE_desktop_Vulkan.md
```diff
@@ -261,7 +261,9 @@ option values.
...
+| Use asset file                                 | GFXRECON_CAPTURE_USE_ASSET_FILE                         | BOOL    | When set to `true` assets (images, buffers and descriptors) will be stored separately into an asset file instead of the capture file.                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
```
csteinberg-samsung, line 265: wouldn't desktop also benefit from a dump asset file trigger like android has?

framework/encode/custom_vulkan_api_call_encoders.cpp
```diff
@@ -403,6 +403,12 @@ VKAPI_ATTR uint64_t VKAPI_CALL GetBlockIndexGFXR()
     return manager->GetBlockIndex();
 }
 
+VKAPI_ATTR void VKAPI_CALL WriteAssetsGFXR()
```
csteinberg-samsung, line 406: I believe you mentioned during our sync meeting that there was a private API that our layer could trigger to force an asset dump outside of a capture trigger, is this that function?

w-pearson: Based on 87ce6507840794d59ded9fac887ae8e399327096, I'm pretty sure it is.

USAGE_desktop_Vulkan.md
w-pearson: I'm pretty sure that doesn't make sense, for the same reason `capture_android_trigger` doesn't exist on desktop - you can't modify environment variables at runtime (but android properties can be modified at runtime). Though a hotkey on desktop would make more sense.

USAGE_android.md
panos-lunarg: > should this be "Dump asset file" instead of "Use asset file"?

Right, copy paste mistake

> I also wonder if this should only trigger on the false -> true setting?

Initially that was my idea too but then I realized that this is problematic. If an asset dump is triggered by the android property, once the assets are dumped the property must be transitioned to false, otherwise it will keep dumping every frame. The option needs to be cleared by the application but, at least as far as I can tell, android apps do not have permission to do that.

framework/encode/custom_vulkan_api_call_encoders.cpp
panos-lunarg: Yes, `WriteAssetsGFXR` is exposed just like `GetBlockIndexGFXR` ( https://github.com/LunarG/gfxreconstruct/commit/87ce6507840794d59ded9fac887ae8e399327096#diff-4cd2c76addccf755b343693d32cb024f7b90d9d4d2fd2ecc7d997921ee1c533bL35 ) that you use in the perfetto layer and in the same fashion should it should be discoverable through the vulkan loader (I haven't tested it but crossing fingers that it should work)

panos-lunarg: Realized that the naming I used was very inconsistent. Changed it to 
`{ "DumpAssetsGFXR", reinterpret_cast<PFN_vkVoidFunction>(encode::DumpAssetsGFXR) }`
Function should now be discoverable through the Vulkan loader by `vkGetInstanceProcAddr([instance], "DumpAssetsGFXR");`

PR #1661, opened 2024-08-12
Avoid async w. externally synched pipeline-caches

fabian-lunarg:
- Keep track of externally synchronized pipeline-caches
- Avoid async operations when using externally synched pipeline-cache

### DIFF
--- a/framework/generated/generated_vulkan_replay_consumer.cpp
+++ b/framework/generated/generated_vulkan_replay_consumer.cpp
@@ -988,8 +988,11 @@ void VulkanReplayConsumer::Process_vkCreateGraphicsPipelines(
     if (UseAsyncOperations())
     {
         auto task = AsyncCreateGraphicsPipelines(call_info, returnValue, in_device, in_pipelineCache, createInfoCount, pCreateInfos, pAllocator, pPipelines);
-        AddHandlesAsync<PipelineInfo>(device, pPipelines->GetPointer(), pPipelines->GetLength(), std::move(handle_info), &VulkanObjectInfoTable::AddPipelineInfo, std::move(task));
-        return;
+        if(task)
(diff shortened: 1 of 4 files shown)


framework/generated/generated_vulkan_replay_consumer.cpp
```diff
@@ -988,8 +988,11 @@ void VulkanReplayConsumer::Process_vkCreateGraphicsPipelines(
     if (UseAsyncOperations())
     {
         auto task = AsyncCreateGraphicsPipelines(call_info, returnValue, in_device, in_pipelineCache, createInfoCount, pCreateInfos, pAllocator, pPipelines);
```
charles-lunarg, line 990: Wouldn't it make more sense to check for the extern sync'd pipeline cache flag here? That way the code doesn't need to call the AsyncCreate<xxx>Pipelines functions in the first place? 

Mainly thinking about how confusing control flow is when you have to call into a function in order to know if you need to skip the function.

fabian-lunarg: _Wouldn't it make more sense to check for the extern sync'd pipeline cache flag here?_

honestly, biggest reason for this control-flow was our code-gen. 
having the statement here would require generating it via python-codegen here:
but because it's only relevant for 2-3 cases, 
I chose to put in the `AsyncCreate***Pipelines` overrides, which are not autogenerated.

fabian-lunarg: _when you have to call into a function in order to know if you need to skip the function_
what I liked about returning the default lambda:
you can regard as an extra `bool` return type, indicating success.

charles-lunarg (approved):

Issue #1575, opened 2024-06-10
Provide some way of replicating environment variables in capture file


bradgrantham-lunarg:
Depends on https://github.com/LunarG/gfxreconstruct/issues/1574

Maybe repurpose `gfxrecon-replay-renamed.py` to handle general replay intelligence (for desktop) and set helpful environment variables as well as rename the replayer.

nickdriscoll-lunarg:
Closed by #1663

PR #1663, opened 2024-08-12
Save/Load process environment variables in capture file

### LINKED ISSUE #1574: Save process environment variables in capture file
Add some kind of a block to capture that contains the process environment variables.  E.g. "MESA_SHADER_CACHE_DIR" is an important variable for allowing Mesa to use the same cache for created graphics pipeline (compiled shaders) from run to run.  If the environment was stored in the capture file, someone could take the output of `gfxrecon-info` or `gfxrecon-convert` and pull out environment variables of interest.  This could be automated by a replayer script, maybe repurpose `gfxrecon-replay-renamed.py` to apply some heuristics and goop to set helpful environment variables, etc.

nickdriscoll-lunarg:
Fixes #1574 and #1575 

This change introduces a new metadata block for saving the captured process's environment variables, which is written out when the capture file is created.
Also adds an argument to `gfxrecon-replay-renamed.py` for preserving specified environment variable values during replay.

### DIFF
--- a/framework/format/format.h
+++ b/framework/format/format.h
@@ -153,6 +153,7 @@ enum class MetaDataType : uint16_t
     kReserved29                             = 29,
     kReserved30                             = 30,
     kReserved31                             = 31,
+    kSetEnvironmentVariablesCommand         = 32,
 };
 
 // MetaDataId is stored in the capture file and its type must be uint32_t to avoid breaking capture file compatibility.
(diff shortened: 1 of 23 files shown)


framework/format/format.h
```diff
@@ -648,6 +649,17 @@ struct ParentToChildDependencyHeader
     uint32_t                    child_count;
 };
 
+static const char kEnvironmentStringDelimeter = (char)-1;
```
fabian-lunarg, line 652: delimeter could be a `constexpr`

fabian-lunarg (approved):
nice, looks good to go

framework/format/format.h
nickdriscoll-lunarg: Done

PR #1665, opened 2024-08-13
Load gfxrecon layer when replaying with a release build on Android

panos-lunarg:
This should allow recapturing when replaying with a release build on Android


bradgrantham-lunarg:
If I provide this option on Windows or Linux, will it recapture there, too?

panos-lunarg:
> If I provide this option on Windows or Linux, will it recapture there, too?

Right, this was problematic. I found a different way by modifying `build.gradle` files

PR #1666, opened 2024-08-13
Fix regression: handle vkDestroy*** calls being passed null-handles

fabian-lunarg:
running an (inactive) capture from our NAS revealed a recent regression in vkDestroy*** calls.

related to work on:
https://github.com/LunarG/Projects/issues/749

goal is to re-add the capture referenced above to our internal CI asap, it's covering relevant edge-cases.


nickdriscoll-lunarg (approved):
Seems fine to me.

fabian-lunarg:
thank you @nickdriscoll-lunarg!

PR #1670, opened 2024-08-19
Capture one specific App

yuacui01:
- When there are multi-app was opened at the same time, if user only want to capture one specific app, then you can do it by setting the progress name.


MarkY-LunarG:
@yuacui01 can you sign the CDA as well before we can continue? Thank you.

yuacui01:
Hi @MarkY-LunarG , I've already agreed to the CLA for LunarG/gfxreconstruct. Is it OK to continue? If there are still problems, please told me. Thanks!

MarkY-LunarG:
@yuacui01 , it's still showing as not signed.  Also it says that `parkerinsh` is not a signer so their commit won't be accepted (see 3rd message from the top of this thread).  Is that you from a separate account?

yuacui01:
Hi @MarkY-LunarG , I can see this information. Does it mean I've already signed?
Untitled

MarkY-LunarG:
@yuacui01 , I think it does mean you signed, but it says one of the commits is from @parkerinsh .  That person didn't sign.  I was wondering if that was you with a different account?   I believe that's the restriction. It won't even let commits in from other engineers that haven't signed.

yuacui01:
Hi @MarkY-LunarG , @parkerinsh is my colleague. But he forget the password and is unable to reset it. So I have to re-submit this change: https://github.com/LunarG/gfxreconstruct/pull/2206
Is it OK?

PR #1680, opened 2024-08-22
dx12 option: GFXRECON_CAPTURE_DRAW_CALLS 

locke-lunarg:
- dx12 option: GFXRECON_CAPTURE_DRAW_CALLS
Specify one index or a range indices drawacalls(include dispatch) based on a ExecuteCommandList index and a CommandList index to capture. The index is 1-based. The args are one submit index, one command index, one or a range indices of draw calls, one or a range indices of bundle drawcalls(option), like "1,1,1" or "1,1,1-2" or "1,1,1,1-2". The forth arg is an option for bundle case. If the the 3rd arg is a bundle commandlist, but it doesn't set the 4th arg, it will set 1 as default. Default is: Empty string (all drawcalls are captured).
Dump resources doesn't the 4th arg. Although three args are enough for dump resources, it will add the 4th arg on it.

- Not unwrapped objects for override functions
Override functions need wrapped objects and wrappers.

- Split commandlist into three for target drawcalls
It does something similar to dump resources, but a bit different. Trimming drawcalls is not like dump resources that could browse the captured file to find target commandlist and drawcall before replay.
For trimming drawcalls, it doesn't know the target commandlist until running the target ExecuteCommandLists. In this case, it assumes every commandlist could be the target commandlist. Every commandlist has three splitted commandlists and add command to the splitted commandlists.
It also can't know which drawcall is the target in advance, commandlist has a drawcall_count to count it. Splitted commandlists will add some redundant commands since it couldn't know the target drawcall in advance, But it shouldn't affect the result and performance.

Based on https://github.com/LunarG/gfxreconstruct/pull/1727


framework/encode/capture_manager.cpp
```diff
@@ -392,6 +393,11 @@ bool CommonCaptureManager::Initialize(format::ApiFamilyId                   api_
             GFXRECON_ASSERT((trim_boundary_ == CaptureSettings::TrimBoundary::kFrames) ||
                             (trim_boundary_ == CaptureSettings::TrimBoundary::kQueueSubmits));
 
+            if (trim_boundary_ == CaptureSettings::TrimBoundary::kQueueSubmits)
+            {
+                GFXRECON_LOG_WARNING("Trimming queue submits will be changed to 0-based.");
```
davidd-lunarg, line 398: Let's use this message:
"Capture has enabled the GFXRECON_CAPTURE_QUEUE_SUBMITS option. This option currently uses 1-based indexing to identify the queue submit range. In the future it will be switched to 0-based indexing."

framework/encode/d3d12_capture_manager.cpp
```diff
@@ -380,19 +383,32 @@ void D3D12CaptureManager::InitializeID3D12ResourceInfo(ID3D12Device_Wrapper*
...
+
+    graphics::dx12::ID3D12DeviceComPtr device;
+    DecrementCallScope();
+    wrapper->GetDevice(IID_PPV_ARGS(&device));
+    IncrementCallScope();
+    GFXRECON_ASSERT(device);
+
+    ID3D12Device_Wrapper* device_wrapper = nullptr;
+    device->QueryInterface(IID_IUnknown_Wrapper, reinterpret_cast<void**>(&device_wrapper));
+    GFXRECON_ASSERT(device_wrapper);
```
davidd-lunarg, line 404: Is it necessary to use `DecrementCallScope` here? `if (call_scope > 1)` then `ID3D12DeviceChild_Wrapper::GetDevice` returns `GetWrappedObjectAs<ID3D12DeviceChild>()->GetDevice(riid, ppvDevice);` which looks like it should be the native DX12 ID3D12Device, not a wrapped object. Then `ID3D12Device_Wrapper::GetExistingWrapper` could be used to get the device wrapper.

For example:
```
// Get the swapchain's native ID3D12Device
graphics::dx12::ID3D12DeviceComPtr device;
wrapper->GetDevice(IID_PPV_ARGS(&device)); 

// Get the ID3D12Device_Wrapper
ID3D12Device_Wrapper* device_wrapper = ID3D12Device_Wrapper::GetExistingWrapper(device);
```

framework/util/options.cpp
```diff
@@ -134,7 +135,7 @@ std::vector<UintRange> GetUintRanges(const char* args, const char* option_name)
                 next_allowed = ranges.back().last + 1;
             }
 
-            if (uint_range.first >= next_allowed)
+            if (!check_overlap_range || uint_range.first >= next_allowed)
```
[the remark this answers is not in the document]
locke-lunarg: `check_overlap_range` and `allow_zero` are for different purpose, so they don't affect each other.
 
For trimming frames and queues, they need to check overlap ranges. Their args could be 100-200,150-250. These ranges are overlap, so log a warning and skip. But this check is inapplicable to trim draw calls.

`allow_zero` is for 0-based. The original code will log a warning and skip if the value is 0.

framework/encode/api_capture_manager.h
```diff
@@ -200,6 +200,8 @@ class ApiCaptureManager
     util::Keyboard&                   GetKeyboard() { return common_manager_->GetKeyboard(); }
     const std::string&                GetScreenshotPrefix() const { return common_manager_->GetScreenshotPrefix(); }
     util::ScreenshotFormat            GetScreenshotFormat() { return common_manager_->GetScreenshotFormat(); }
+    auto                              GetTrimBundary() const { return common_manager_->GetTrimBundary(); }
```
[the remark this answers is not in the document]
locke-lunarg: fixed

USAGE_desktop_D3D12.md
```diff
@@ -110,6 +110,7 @@ Quit after capturing frame ranges | GFXRECON_QUIT_AFTER_CAPTURE_FRAMES | BOOL |
...
 Capture Specific GPU Queue Submits | GFXRECON_CAPTURE_QUEUE_SUBMITS | STRING | Specify one or more comma-separated GPU queue submit call ranges to capture.  Queue submit calls are `vkQueueSubmit` for Vulkan and `ID3D12CommandQueue::ExecuteCommandLists` for DX12. Queue submit ranges work as described above in `GFXRECON_CAPTURE_FRAMES` but on GPU queue submit calls instead of frames.  Default is: Empty string (all queue submits are captured).
+Capture Specific Drawcalls | GFXRECON_CAPTURE_DRAWCALLS | STRING | Specify one index or a range indices drawacalls(include dispatch) based on a ExecuteCommandList index and a CommandList index to capture. The index is 0-based. The args are one submit index, one command index, one or a range indices of draw calls, one or a range indices of bundle drawcalls(option), like "0,0,0" or "0,0,0-2" or "0,0,0-2,0". The forth arg is an option for bundle case. If the the 3rd arg is a bundle commandlist, but it doesn't set the 4th arg, it will set 0 as default. Default is: Empty string (all drawcalls are captured).
```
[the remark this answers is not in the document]
locke-lunarg: fixed

framework/encode/capture_manager.cpp
locke-lunarg: fixed

framework/encode/d3d12_capture_manager.cpp
locke-lunarg: fixed

locke-lunarg:
Some discussion about trim swapchain buffer in https://github.com/LunarG/gfxreconstruct/pull/1727

framework/encode/d3d12_capture_manager.cpp
```diff
@@ -1770,21 +1851,156 @@ void D3D12CaptureManager::PreProcess_ID3D12CommandQueue_ExecuteCommandLists(ID3D
...
 
 void D3D12CaptureManager::PostProcess_ID3D12CommandQueue_ExecuteCommandLists(ID3D12CommandQueue_Wrapper* wrapper,
                                                                              UINT                        num_lists,
                                                                              ID3D12CommandList* const*   lists)
 {
-    PostQueueSubmit();
+    GFXRECON_ASSERT(num_lists > 0 && lists);
+
+    ID3D12CommandList_Wrapper* cmd_wrapper = nullptr;
+    lists[0]->QueryInterface(IID_IUnknown_Wrapper, reinterpret_cast<void**>(&cmd_wrapper));
```
davidd-lunarg, line 1932: I think this might return a pointer to an `IUnknown*` (not a `ID3D12CommandList_Wrapper*`) because  `call scope >= 1` and `IID_IUnknown_Wrapper` is used.

`ID3D12CommandList* const* lists` should already be wrapper pointers, they just need to be reinterpret cast:
`ID3D12CommandList_Wrapper* cmd_wrapper = reinterpret_cast<ID3D12CommandList_Wrapper>(list[0]);`

This also applies to other places in this PR. Generally `QueryInterface` should not be used to get a wrapper object.

davidd-lunarg (changes requested):

framework/encode/d3d12_capture_manager.cpp
```diff
@@ -1770,21 +1851,156 @@ void D3D12CaptureManager::PreProcess_ID3D12CommandQueue_ExecuteCommandLists(ID3D
...
+            ID3D12CommandList_Wrapper*      target_wrapper = nullptr;
+            target_cmdlist->QueryInterface(IID_IUnknown_Wrapper, reinterpret_cast<void**>(&target_wrapper));
+            GFXRECON_ASSERT(target_wrapper);
+
+            auto target_info = target_wrapper->GetObjectInfo();
+            GFXRECON_ASSERT(target_info);
+
+            if (!target_info->is_split_commandlist)
+            {
+                DecrementCallScope();
```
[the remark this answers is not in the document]
locke-lunarg: In order to write `wrapper->ExecuteCommandLists(cmdlists.size(), cmdlists.data());` that is between `ActivateTrimmingDrawCalls` and `DeactivateTrimmingDrawCalls`.  Since call_scope_ has been 1, it needs to `DecrementCallScope`.

In old version, I ran unwrapped queue and `Encode_ID3D12CommandQueue_ExecuteCommandLists` to write this `ExecuteCommandLists`. This version run wrapped queue to write ExecuteCommandLists.

framework/encode/d3d12_capture_manager.cpp
```diff
@@ -1770,21 +1851,156 @@ void D3D12CaptureManager::PreProcess_ID3D12CommandQueue_ExecuteCommandLists(ID3D
...
+                // target of splitted
+                common_manager_->ActivateTrimmingDrawCalls(format::ApiFamilyId::ApiFamily_D3D12);
+
+                std::vector<ID3D12CommandList*> cmdlists;
+
+                auto target_draw_call_cmd = target_info->split_command_sets[graphics::dx12::kDrawCallArrayIndex].list;
+                GFXRECON_ASSERT(target_draw_call_cmd);
+                cmdlists.emplace_back(target_draw_call_cmd);
+
+                wrapper->ExecuteCommandLists(cmdlists.size(), cmdlists.data());
```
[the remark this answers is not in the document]
locke-lunarg: 1910 is also changed to `wrapper`. Since I have to change the native queue to wrapper in `PreProcess_ID3D12CommandQueue_ExecuteCommandLists` to track resource states. For consistency, I also changed here and 1910.  But it's ok to use native. I just needs to add `Encode_ID3D12CommandQueue_ExecuteCommandLists` to write it. Do you prefer using native queue?

framework/encode/d3d12_capture_manager.cpp
locke-lunarg: I don't know. When I used `reinterpret_cast` to cast to wrapper, I sometimes got nullptr. I didn't study deeply why it happened. But `QueryInterface` never happened nullptr issue.  I feel `QueryInterface` is safer. I could try `reinterpret_cast` again to see if nullptr happens again.

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -365,26 +365,115 @@ void Dx12ReplayConsumerBase::ApplyBatchedResourceInitInfo(
...
+            auto object_info           = GetObjectInfo(resource_info.second->resource_id);
+            auto extra_info            = GetExtraInfo<D3D12ResourceInfo>(object_info);
+            auto swapchain_info        = GetObjectInfo(extra_info->swap_chain_id);
+            auto swapchain_extra_info  = GetExtraInfo<DxgiSwapchainInfo>(swapchain_info);
+            auto swapchain             = reinterpret_cast<IDXGISwapChain3*>(swapchain_info->object);
+            swapchain_infos[swapchain] = swapchain_extra_info;
+
+            for (auto &state : resource_info.second->before_states)
+            {
+                if (state.states = D3D12_RESOURCE_STATE_PRESENT)
```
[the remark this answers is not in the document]
locke-lunarg: fixed

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -365,26 +365,115 @@ void Dx12ReplayConsumerBase::ApplyBatchedResourceInitInfo(
...
+        std::unordered_map<ID3D12Resource*, ResourceInitInfo*> swapchain_resource_infos;
+        std::unordered_map<ID3D12Resource*, ResourceInitInfo*> others_resource_infos;
+
+        for (auto& resource_info : resource_infos)
         {
+            auto object_info = GetObjectInfo(resource_info.second.resource_id);
+            if (object_info->extra_info != nullptr)
+            {
+                auto extra_info = GetExtraInfo<D3D12ResourceInfo>(object_info);
+                if (extra_info->swap_chain_id)
```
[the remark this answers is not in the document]
locke-lunarg: fixed

framework/encode/d3d12_capture_manager.cpp
locke-lunarg: fixed

framework/encode/d3d12_capture_manager.cpp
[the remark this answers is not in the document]
davidd-lunarg: I see. I think it would be good to avoid the need to call `Decrement/IncrementScope` so if switching to use native handle isn't too messy that would be my preference.

framework/encode/d3d12_capture_manager.cpp
davidd-lunarg: > When I used reinterpret_cast to cast to wrapper, I sometimes got nullptr.
Does this still happen? I wouldn't expect any nullptrs there. For example `Dx12StateTracker::TrackExecuteCommandLists` does the following and doesn't null-check the wrapper.
```
        auto list_wrapper = reinterpret_cast<ID3D12CommandList_Wrapper*>(lists[i]);
        auto list_info    = list_wrapper->GetObjectInfo();
```

locke-lunarg: I changed to `reinterpret_cast` and didn't see nullptr happen again. I think I made mistakes to use `reinterpret_cast` before. Maybe I tried casting a native object to a wrapper.

framework/encode/d3d12_capture_manager.cpp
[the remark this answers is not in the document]
locke-lunarg: I just added a commit that moved the extra `ExecuteCommandLists` from `PreProcess_ID3D12CommandQueue_ExecuteCommandLists` to `OverrideID3D12CommandQueue_ExecuteCommandLists`, so all three extra `ExecuteCommandLists` are in `OverrideID3D12CommandQueue_ExecuteCommandLists` to make the code simpler.

In this case, `OverrideID3D12CommandQueue_ExecuteCommandLists` has to call `Decrement/IncrementScope`  since the first extra `ExecuteCommandLists` needs to be tracked.  I could limit `Decrement/IncrementScope` only for the first `ExecuteCommandLists`. The second and the third use native queue if that is you preference.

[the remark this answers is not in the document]
locke-lunarg: I update that commit that only the first `ExecuteCommandLists` use the wrapped queue. `Decrement/IncrementScope` is only limited on the first `ExecuteCommandLists`. The rest of two use the native queue.

framework/encode/capture_manager.cpp
```diff
@@ -778,6 +815,24 @@ void CommonCaptureManager::CheckStartCaptureForTrackMode(format::ApiFamilyId api
     }
 }
 
+void CommonCaptureManager::ActivateTrimmingDrawCalls(format::ApiFamilyId api_family)
```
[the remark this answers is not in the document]
locke-lunarg: Will do.

framework/generated/dx12_generators/dx12_wrapper_body_generator.py
```diff
@@ -772,6 +783,49 @@ def write_class_member_def(self, class_info, indent=''):
...
+                expr += indent + '{\n'
+                expr += indent1 + 'manager->DecrementCallScope();\n'
+                expr += indent1 + 'auto trim_draw_calls_command_sets = manager->GetCommandListsForTrimDrawCalls(this, format::ApiCall_{}_{});\n'.format(class_name, method_name)
+                expr += indent1 + 'for(auto& command_set : trim_draw_calls_command_sets)\n'
+                expr += indent1 + '{\n'
+
+                if class_name != 'ID3D12GraphicsCommandList':
+                    expr += indent2 + 'graphics::dx12::{}ComPtr command_list = nullptr;\n'.format(class_name)
+                    expr += indent2 + 'command_set.list->QueryInterface(IID_PPV_ARGS(&command_list));\n'
+                    expr += indent2 + 'auto* wrapper = reinterpret_cast<{}_Wrapper*>(command_list.GetInterfacePtr());\n'.format(class_name)
```
davidd-lunarg, line 802: I think these lines could be changed to avoid the QueryInterface call. Since the call scope is decremented to 0 here, QueryInterface may have some side-effects for capture/tracking and is probably cleaner to avoid using it.

```
  expr += indent2 + 'auto* base_wrapper = reinterpret_cast<ID3D12GraphicsCommandList_Wrapper*>(command_set.list.GetInterfacePtr());\n'
  expr += indent2 + 'auto* wrapper = dynamic_cast<{}_Wrapper*>(base_wrapper);\n'.format(class_name)
  expr += indent2 + 'GFXRECON_ASSERT(wrapper != nullptr);\n'
```

Which results in this code in `generated_dx12_wrappers.cpp`:
```
  auto* base_wrapper = reinterpret_cast<ID3D12GraphicsCommandList_Wrapper*>(command_set.list.GetInterfacePtr());
  auto* wrapper = dynamic_cast<ID3D12GraphicsCommandList1_Wrapper*>(base_wrapper);
  GFXRECON_ASSERT(wrapper != nullptr);
```

davidd-lunarg: One change to the code above. Instead of `dynamic_cast` I think `static_cast` can be used since we know all `*CommandList_Wrappers` are created as the most derived class `ID3D12GraphicsCommandList9_Wrapper` in `void WrapID3D12CommandList(REFIID riid, void** object, DxWrapperResources* resources)` and we can drop the assert. So instead, the generated code would be:

```
  auto* base_wrapper = reinterpret_cast<ID3D12GraphicsCommandList_Wrapper*>(command_set.list.GetInterfacePtr());
  auto* wrapper = static_cast<ID3D12GraphicsCommandList1_Wrapper*>(base_wrapper);
```

locke-lunarg: How about just `auto* wrapper = reinterpret_cast<ID3D12GraphicsCommandList1_Wrapper*>(command_set.list.GetInterfacePtr());`  to reduce one cast. Is it safe?

davidd-lunarg:
>PR https://github.com/LunarG/gfxreconstruct/pull/1711 makes some changes to the way mutex locking is handled in ActivateTrimming(). When PR https://github.com/LunarG/gfxreconstruct/pull/1711 is merged, those changes may also need to be added here to ActivateTrimmingDrawCalls().

https://github.com/LunarG/gfxreconstruct/pull/1711 was merged

framework/generated/dx12_generators/dx12_wrapper_body_generator.py
locke-lunarg: fixed

framework/encode/d3d12_capture_manager.cpp
```diff
@@ -2952,5 +3123,595 @@ void D3D12CaptureManager::UpdateSwapChainSize(uint32_t width, uint32_t height, I
...
+
+        queue->ExecuteCommandLists(cmdlists.size(),
+                                   UnwrapObjects<ID3D12CommandList>(cmdlists.data(), cmdlists.size(), unwrap_memory));
+        return true;
+    }
+    return false;
+}
+
+bool D3D12CaptureManager::HasSplitCommandLists(UINT num_lists, ID3D12CommandList* const* lists)
+{
```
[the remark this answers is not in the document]
locke-lunarg: fixed

davidd-lunarg (approved):

PR #1689, opened 2024-08-27
Fix allocator null for trim capture

jirongjiao:
**Problem**
During trim capture of some D3D12 apps, some command allocators may be released during state tracking stage, resulting trace replay crash at CreateCommandList() due to null command allocator.

**Solution**
Create temp command allocator for CreateCommandList when it's allocator has been released during state tracking.


lunarpapillo:
The **LunarG CI** failure seemed to be due to a crashed X server.  I'm retriggering your run.

Issue #1526, opened 2024-04-30
ray-tracing captures are not tracked correctly


bradgrantham-lunarg:
When trimming, in state tracking (before the frame range) we write out the buffer contents for an AS directly rather than writing commands to recreate the AS in the state setup block.  This is not guaranteed to work and I'm surprised that we sometimes see it succeed.

fabian-lunarg:
I closed above issue as duplicate, but still it contains very valuable insights:

_It appears that gfx-reconstruct is relying on the opaque capture address obtained from vkGetDeviceMemoryOpaqueCaptureAddress to infer the corresponding vkDeviceMemory for a given address retrieved from vkGetAccelerationStructureDeviceAddress.
...
Opaque addresses should not be compared or offset against one another. A buffer's opaque capture address can be completely different from its BufferDeviceAddress_

we need to keep in mind those address-spaces can/will be distinct.
thank you @dorian-apanel-intel for pointing that out.

fabian-lunarg:
the last PR related to 'portable raytracing' just went in. I'll leave the issue open for a bit, in case something was missed. @bradgrantham-lunarg

dorian-apanel-intel:
@fabian-lunarg Is capturing a capture supported?
I captured 8000 frames of a RT heavy game, but when I tried to subcapture three frames from it, I see the gfx-reconstruct layer tries to do a `vkMapMemory` on memory allocated with non host visible flags (`VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT`) and fails

```
VkLayer_api_dump.dll!vkMapMemory(VkDevice_T * device, VkDeviceMemory_T * memory, unsigned __int64 offset, unsigned __int64 size, unsigned int flags, void * * ppData) Line 2606	C++
VkLayer_gfxreconstruct.dll!gfxrecon::encode::VulkanStateTracker::TrackTlasToBlasDependencies(unsigned int command_buffer_count, VkCommandBuffer_T * const * command_buffers) Line 2445	C++
VkLayer_gfxreconstruct.dll!gfxrecon::encode::VulkanCaptureManager::PreProcess_vkQueueSubmit(std::shared_lock<std::shared_mutex> & current_lock, VkQueue_T * queue, unsigned int submitCount, const VkSubmitInfo * pSubmits, VkFence_T * fence) Line 2402	C++
[...]

PR #1690, opened 2024-08-28
Raytracing: Shader-group-handle mapping

fabian-lunarg:
- extract+provide required raytracing device-properties for capture- and replay-time (handle-sizes & alignments)
- add required vanilla hash-utils (hash_combine, hash_range -> matching boost's utils)
- create an efficient wrapper-type (`shader_group_handle_t`) that can be compared and hashed (-> can be used as key in std::unordered_map)
- provide a LUT inside `decode::PipelineInfo` to map from capture- to replay-time handles.

this matches @bradgrantham-lunarg's [proof-of-concept](https://github.com/LunarG/gfxreconstruct/compare/dev...bradgrantham-lunarg:gfxreconstruct:brad-hack-shader-group-handles), fleshing out some details, but please note:
no re-mapping is actually added/performed (that part of the proof-of-concept was still too hacky). 
so this PR is only a necessary stepping-stone and should not change GFXR's behavior in any way.

### DIFF
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -32,7 +32,6 @@
 #include "graphics/vulkan_device_util.h"
 #include "graphics/vulkan_shader_group_handle.h"
 #include "util/defines.h"
-#include "util/hash.h"
 
 #include "vulkan/vulkan.h"
 
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -42,6 +42,7 @@
 #include <unordered_set>
 #include <vector>
 #include <future>
+#include <optional>
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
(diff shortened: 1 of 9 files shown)


framework/decode/vulkan_object_info.h
```diff
@@ -461,6 +473,9 @@ struct PipelineInfo : public VulkanObjectInfoAsync<VkPipeline>
 
     // Is VK_DYNAMIC_STATE_VERTEX_INPUT_BINDING_STRIDE_EXT enabled
     bool dynamic_vertex_binding_stride{ false };
+
+    // map capture- to replay-time shader-group-handles
+    std::unordered_map<graphics::shader_group_handle_t, graphics::shader_group_handle_t> shader_group_handle_map;
```
fabian-lunarg, line 478: @NickDriscoll  this LUT is pretty much the outcome of this PR. 
we'll be able to map from capture- to replay-time handles (but have to figure out how/when/where we want to do it ... in a follow-up)

fabian-lunarg:
the PR belongs to this issue: https://github.com/LunarG/gfxreconstruct/issues/1526

framework/decode/vulkan_object_info.h
nickdriscoll-lunarg: Makes sense to me.

nickdriscoll-lunarg (approved):
I don't see any issues with these changes.

fabian-lunarg:
thank you @nickdriscoll-lunarg!

PR #1695, opened 2024-08-29
Track external android format

ziga-lunarg:
The external format from android must be tracked, because in GetImageResourceSizesOptimal we create temporary images from the same format to calculate the resource size

If format is VK_FORMAT_UNDEFINED and the external android format is not provided it will crash on android

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -4921,6 +4921,13 @@ VulkanReplayConsumerBase::OverrideCreateImage(PFN_vkCreateImage                 
         {
             image_info->queue_family_index = 0;
         }
+
+        auto external_format_android = graphics::GetPNextStruct<VkExternalFormatANDROID, VkImageCreateInfo>(
+            replay_create_info, VK_STRUCTURE_TYPE_EXTERNAL_FORMAT_ANDROID);
+        if (external_format_android)
+        {
+            image_info->external_format = external_format_android->externalFormat;
+        }
     }
 
     return result;
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -358,6 +358,8 @@ struct ImageInfo : public VulkanObjectInfo<VkImage>
     VkImageUsageFlags     usage{ 0 };
     VkImageType           type{};
     VkFormat              format{};
+    // VkExternalFormatANDROID
+    uint64_t              external_format{ 0 };
     VkExtent3D            extent{ 0, 0, 0 };
     VkImageTiling         tiling{};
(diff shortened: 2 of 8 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4921,6 +4921,17 @@ VulkanReplayConsumerBase::OverrideCreateImage(PFN_vkCreateImage
         {
             image_info->queue_family_index = 0;
         }
+
```
bradgrantham-lunarg, line 4924: Can you use `GetPNextStruct` from `graphics/vulkan_util.h` to find this struct?

bradgrantham-lunarg (approved):
Looks good to me except possibly using the convenience function if it fits.

What would a minimal test program for this case look like?

framework/decode/vulkan_replay_consumer_base.cpp
ziga-lunarg: Done

ziga-lunarg:
> Looks good to me except possibly using the convenience function if it fits.
> 
> What would a minimal test program for this case look like?

A minimal test program would have to create an AHardwareBuffer and then create an image with an external android format, and allocate device memory with the AHardwareBuffer and bind it to the image. And I think this only happens in a trimmed capture. This should reproduce it, but not 100% sure.

mikes-lunarg (approved):

bradgrantham-lunarg:
> > Looks good to me except possibly using the convenience function if it fits.
> > What would a minimal test program for this case look like?
> 
> A minimal test program would have to create an AHardwareBuffer and then create an image with an external android format, and allocate device memory with the AHardwareBuffer and bind it to the image. And I think this only happens in a trimmed capture. This should reproduce it, but not 100% sure.

Do we have a capture we could put in CI that replicates this or could you quickly write a program to replicate this?

ziga-lunarg:
> > > Looks good to me except possibly using the convenience function if it fits.
> > > What would a minimal test program for this case look like?
> > 
> > 
> > A minimal test program would have to create an AHardwareBuffer and then create an image with an external android format, and allocate device memory with the AHardwareBuffer and bind it to the image. And I think this only happens in a trimmed capture. This should reproduce it, but not 100% sure.
> 
> Do we have a capture we could put in CI that replicates this or could you quickly write a program to replicate this?

I do have a capture, but I've realized this fix is not enough. There is another problem later in the capture, that I thought was unrelated, but it seems that it is.

bradgrantham-lunarg:
Can this be merged or is it part of a later PR?

ziga-lunarg:
> Can this be merged or is it part of a later PR?

I might close this and open a different PR, because this only fixes a small part of the issue.

PR #1697, opened 2024-08-30
Fixes to dump resources arg parsing

davidlunarg:


tools/replay/parse_dump_resources_cli.cpp
```diff
@@ -437,19 +445,31 @@ bool parse_dump_resources_arg(gfxrecon::decode::VulkanReplayOptions& vulkan_repl
...
 
+                vulkan_replay_options.TraceRays_Indices.push_back(std::vector<uint64_t>());
                 if (TraceRays)
                 {
-                    vulkan_replay_options.TraceRays_Indices.push_back(std::vector<uint64_t>());
                     vulkan_replay_options.TraceRays_Indices[i].push_back(TraceRays);
                 }
 
-                vulkan_replay_options.QueueSubmit_Indices.insert(QueueSubmit);
+                if (vulkan_replay_options.QueueSubmit_Indices.find(QueueSubmit) !=
```
bradgrantham-lunarg, line 460: Could you use the idiom here `auto [itr, inserted] = vulkan_replay_options.QueueSubmit_Indices.insert(QueueSubmit);` and then `if(!inserted) {parse_error = true...}`?  `std::unordered_set::insert` will fail if the element is already in there.  If you'd like I'll make the commit but I don't have a test case here.

bradgrantham-lunarg (approved):
LGTM with the one suggestion about `insert`.  I'll leave validating the checking to @panos-lunarg

tools/replay/parse_dump_resources_cli.cpp
```diff
@@ -78,51 +78,42 @@ static bool CheckIndicesForErrors(const gfxrecon::decode::VulkanReplayOptions& v
...
 
     if (vulkan_replay_options.RenderPass_Indices.size())
     {
-        bool is_3d = false;
+        bool is_complete = false;
         for (const auto& indices0 : vulkan_replay_options.RenderPass_Indices)
         {
             if (!indices0.size())
             {
+                is_complete = true;
```
panos-lunarg, line 99: Will this allow a `RenderPass` entry that is not 3d?
RenderPass needs to be a 3d array like this:
```
"RenderPass": [
        [[]], [ [ 401, 419 ] ]
    ],
```

tools/replay/parse_dump_resources_cli.cpp
```diff
@@ -78,51 +78,42 @@ static bool CheckIndicesForErrors(const gfxrecon::decode::VulkanReplayOptions& v
     // Indices should be provided sorted
     if (vulkan_replay_options.Draw_Indices.size())
     {
-        bool is_2d = false;
         for (const auto& indices : vulkan_replay_options.Draw_Indices)
         {
-            is_2d |= (indices.size() > 0);
```
panos-lunarg, line 84: DrawIndices must be 2d array

tools/replay/parse_dump_resources_cli.cpp
```diff
@@ -191,20 +163,26 @@ static bool CheckIndicesForErrors(const gfxrecon::decode::VulkanReplayOptions& v
...
 
-    if (vulkan_replay_options.Draw_Indices.size() && !vulkan_replay_options.RenderPass_Indices.size())
+    for (int i = 0; i < vulkan_replay_options.Draw_Indices.size(); i++)
     {
-        GFXRECON_LOG_ERROR("ERROR - incomplete --dump-resources parameters");
-        GFXRECON_LOG_ERROR("DrawCall indices must always be acompanied by render pass indices");
-        return true;
-    }
+        // Make sure each Draw has corresponding RenderPass indices
+        if (vulkan_replay_options.Draw_Indices[i].size() != vulkan_replay_options.RenderPass_Indices[i].size())
```
panos-lunarg, line 169: This check is not correct. There can be several draw calls between a BeginRenderpass and EndRenderpass.
I.e.
```
    "Draw": [
        [
            473,
            474,
            475,
            476,
            477,
            478,
            479,
            480,
            481,
            482,
            483,
            484,
            485,
            486
        ]
    ],
    "RenderPass": [
        [
            [
                466,
                487
            ]
        ]
    ],
```

tools/replay/parse_dump_resources_cli.cpp
davidlunarg: Change made in 15d6ab7561eccb3b1c1c40c0f9c8b412bc5ba741

tools/replay/parse_dump_resources_cli.cpp
davidlunarg: Fixed in 2d8a134f8a9f952fd31a128e78d2b0b455f1a261

tools/replay/parse_dump_resources_cli.cpp
davidlunarg: Fix in 2d8a134f8a9f952fd31a128e78d2b0b455f1a261

tools/replay/parse_dump_resources_cli.cpp
davidlunarg: Fix in 2d8a134f8a9f952fd31a128e78d2b0b455f1a261

locke-lunarg (approved):

tools/replay/parse_dump_resources_cli.cpp
```diff
@@ -78,51 +79,36 @@ static bool CheckIndicesForErrors(const gfxrecon::decode::VulkanReplayOptions& v
...
             {
-                GFXRECON_LOG_ERROR("ERROR - incomplete --dump-resources parameters");
-                GFXRECON_LOG_ERROR("DrawCall indices are not sorted")
+                GFXRECON_LOG_ERROR("ERROR - incorrect --dump-resources parameters");
+                GFXRECON_LOG_ERROR("Draw indices are not sorted")
                 return true;
             }
         }
-
-        if (!is_2d)
```
panos-lunarg, line 94: The dimensions check is completely gone now

tools/replay/parse_dump_resources_cli.cpp
```diff
@@ -78,51 +79,36 @@ static bool CheckIndicesForErrors(const gfxrecon::decode::VulkanReplayOptions& v
...
-            GFXRECON_LOG_ERROR("ERROR - incomplete --dump-resources parameters");
-            GFXRECON_LOG_ERROR("DrawCall indices must be a 2 dimensional array")
-            return true;
-        }
     }
 
     if (vulkan_replay_options.RenderPass_Indices.size())
     {
-        bool is_3d = false;
+        bool is_complete = true;
```
panos-lunarg, line 95: I don't think this is correct. If `vulkan_replay_options.RenderPass_Indices` is empty, `is_complete` will stay true and there won't be a warning

tools/replay/parse_dump_resources_cli.cpp
```diff
@@ -191,20 +158,48 @@ static bool CheckIndicesForErrors(const gfxrecon::decode::VulkanReplayOptions& v
...
+    if (vulkan_replay_options.Draw_Indices.size() != vulkan_replay_options.RenderPass_Indices.size())
     {
         GFXRECON_LOG_ERROR("ERROR - incomplete --dump-resources parameters");
-        GFXRECON_LOG_ERROR("DrawCall indices must always be acompanied by render pass indices");
+        GFXRECON_LOG_ERROR("Draw indices must always be accompanied by RenderPass indices");
         return true;
     }
+    for (int i = 0; i < vulkan_replay_options.Draw_Indices.size(); i++)
+    {
+        if (vulkan_replay_options.Draw_Indices[i].size() != vulkan_replay_options.RenderPass_Indices[i].size())
```
panos-lunarg, line 170: This check is wrong. It will print a warning for valid input.
Perhaps you want to do
```
if (!vulkan_replay_options.Draw_Indices[i].empty() && vulkan_replay_options.RenderPass_Indices[i].empty())
```

tools/replay/parse_dump_resources_cli.cpp
davidlunarg: But we check for indices1.size() > 0, which is checking to make sure it is 2d.

tools/replay/parse_dump_resources_cli.cpp
davidlunarg: RenderPass indices can be empty when doing a tracerays or dispatch dump, so we shouldn't give a warning then. Or am I wrong on RenderPass indices being empty on tracerays/dispatch?

tools/replay/parse_dump_resources_cli.cpp
panos-lunarg: Not exactly.
When specifying a draw call for dumping you need to specify the indices of BeginRenderPass and EndRenderPass that include that draw call. If you want to dump multiple draw calls from the same render pass you don't have to specify the indices for Begin/End again. Like in this example

```
    "Draw": [
        [
            473,
            474,
            475,
            476,
            477,
            478,
            479,
            480,
            481,
            482,
            483,
            484,
            485,
            486
        ]
    ],
    "RenderPass": [
        [
            [
                466,
                487
            ]
        ]
    ],
```
In the example above it will be
`vulkan_replay_options.Draw_Indices[0].size() == 14` and `vulkan_replay_options.RenderPass_Indices[0].size() == 1`

tools/replay/parse_dump_resources_cli.cpp
panos-lunarg: Right sorry. If `vulkan_replay_options.RenderPass_Indices` is empty the above if case will be false

tools/replay/parse_dump_resources_cli.cpp
panos-lunarg: > But we check for indices1.size() > 0

Can't see it. On which line?

davidlunarg: Ahh, I was looking at a few lines down where we check RenderPassIndices at line 100.

But Draw_Indices will not be 2d if all dump resources are on TraceRays and Dispatch calls. So I removed the check.

panos-lunarg (approved):

PR #1700, opened 2024-09-02
Refactor pNext-util, enforce usage

fabian-lunarg:
- pNext-chains: replace a commonly repeated loop-pattern with an existing utility-function
- resolve sType with existing GetSType-util
- rename, doc and move to dedicated header
- applied in locations matching grep-pattern '== VK_STRUCTURE_TYPE'


charles-lunarg (approved):
Love to see the code become more self consistent! And with a good cleanup of the template code underneath too :D

PR #1701, opened 2024-09-02
openxr: Modify xrPollEvent check frequency

MarkY-LunarG:
During replay, when xrPollEvent should be returning a valid value, instead of checking every 1 nanosecond for a new event, start at a nanosecond and double every time we check.  This should reduce the amount of xrPollEvent spam we see occur on replay.

### DIFF
--- a/framework/decode/openxr_replay_consumer_base.cpp
+++ b/framework/decode/openxr_replay_consumer_base.cpp
@@ -652,11 +652,13 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&           
     XrResult replay_result;
 
     // WIP: Put this constant somewhere interesting
-    const uint32_t kRetryLimit = 10000;
-    uint32_t       retry_count = 0;
+    const uint32_t kRetryLimit      = 10000;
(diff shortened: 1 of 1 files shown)


framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -685,10 +686,19 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
...
-                else
+                else if (replay_result == XR_EVENT_UNAVAILABLE)
                 {
                     // Yield and retry
-                    std::this_thread::sleep_for(std::chrono::nanoseconds(1));
+                    std::this_thread::sleep_for(std::chrono::nanoseconds(sleep_time));
+
+                    // Next time, sleep for double what we initially slept for.  This way if we're just
+                    // spinning, we spin less and less each time.
+                    sleep_time *= 2;
```
jzulauf-lunarg, line 696: Should add a max sleep of 500ms or so, and reduce the retry limit s.t. the maximum wait is < something like 5-10 seconds total.

jzulauf-lunarg (changes requested):
One comment on max time and max retry.

PR #1707, opened 2024-09-03
upgrade versions of actions/*-artifact

bradgrantham-lunarg:


bradgrantham-lunarg:
I downloaded and compared the listings in the .zip artifacts for this CI build (https://github.com/LunarG/gfxreconstruct/actions/runs/10692368513?pr=1707) versus the artifacts for the most recent successful build (https://github.com/LunarG/gfxreconstruct/actions/runs/10708528600?pr=1710).

The list of files is identical except sorted differently and the new artifact .ZIPs don't call out subdirectories as their own entries.

So I conclude that switching from `v1` to `v4` is successful.

mikes-lunarg (approved):

fabian-lunarg (approved):
straight-forward!

PR #1708, opened 2024-09-04
Fix capture deadlock when `vkQueuePresentKHR` unlocks a CPU wait

marius-pelegrin-arm:
This bug can be observed when capturing the Khronos Vulkan-Sample "timeline_semaphore"

In a situation where you have two threads:
- one waiting on a synchronization primitive (`VkSemaphore`, `VkFence`....)
- another on which a call to `vkQueuePresentKHR` is necessary to signal that primitive

Then if the "CPU waiting call" (`vkWaitForFences`, `vkWaitSemaphores`...) is reached by thread 1 before `vkQueuePresentKHR` is reached by thread 2, the `api_call_mutex_` of the `CaptureManager` is locked by the CPU waiting call which makes it impossible to lock by the call to `vkQueuePresentKHR` (which is always an exclusive lock) and we end up in a deadlock situation.

I don't see why the lock inside `vkQueuePresentKHR` should be exclusive, so I propose to solve this issue by simply using the same system of shared lock as for all the other calls.

### DIFF
--- a/framework/generated/generated_vulkan_api_call_encoders.cpp
+++ b/framework/generated/generated_vulkan_api_call_encoders.cpp
@@ -8601,7 +8601,17 @@ VKAPI_ATTR VkResult VKAPI_CALL QueuePresentKHR(
 {
     VulkanCaptureManager* manager = VulkanCaptureManager::Get();
     GFXRECON_ASSERT(manager != nullptr);
-    auto api_call_lock = VulkanCaptureManager::AcquireExclusiveApiCallLock();
+    auto force_command_serialization = manager->GetForceCommandSerialization();
+    std::shared_lock<CommonCaptureManager::ApiCallMutexT> shared_api_call_lock;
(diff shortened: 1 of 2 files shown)


dgraves:
An additional change would be needed here to release the shared lock and acquire the exclusive lock when trimming is activated by a present call. Recently, I was looking at the exclusive locks in present due to a noticeable perf impact they have on some DX12 applications.

An earlier version of the code used a shared lock in present and then acquired the exclusive lock when activating trimming. The exclusive locks were moved to the present calls by 6bf1a6a89a80ef7e490f0b9ca4259d8386e1da21 and 500bb7c1269745d02ae0cee0c5ba721d55e9b078 to avoid a deadlock issue, which looks to have been caused by two threads acquiring the shared lock and then attempting to acquire the exclusive lock without releasing the shared lock. 

I have a local change to acquire the shared lock on present and then release it before acquiring the exclusive lock when activating trimming, which has been going through testing and should be ready for submission soon. This change wouldn't necessarily replace what you have here but could be added on top of it.

PR #1710, opened 2024-09-04
Use reflection to get DXIL library shader exports

davidd-lunarg:


davidd-lunarg:
> CI gfxreconstruct build # 4762 failed.

This is failing because it is finding the wrong `dxcompiler.dll`. GFXR may need to redistribute that file from the SDK, I will look into it.

framework/decode/dx12_resource_value_mapper.cpp
```diff
@@ -141,6 +148,30 @@ void CopyMappedResourceValuesFromSrcToDst(std::map<uint64_t, T>&
     }
 }
 
+std::string DemangleDxilExportName(const std::string& mangled_name)
+{
+    size_t demangled_name_start = mangled_name.find_first_of("?");
+    size_t demangled_name_end   = mangled_name.find_first_of("@");
+
+    std::string demangled_name;
```
rurra-amd, line 156: Since we check for empty string at the bottom, would it be safer here to explicitly init `demangled_name` to empty?

framework/decode/dx12_resource_value_mapper.cpp
```diff
@@ -1699,10 +1731,89 @@ void Dx12ResourceValueMapper::GetStateObjectLrsAssociationInfo(
             auto num_exports           = dxil_lib_desc_decoder->GetPointer()->NumExports;
             if (num_exports == 0)
             {
-                // TODO: Parse the names of all shaders exported from the DXIL library.
+#if defined(GFXRECON_DXC_SUPPORT)
+                // If D3D12_DXIL_LIBRARY_DESC::NumExports == 0, everything in the DXIL library is exported. Use
+                // reflection to get the list of exported shader names.
+                HRESULT                         hr;
```
rurra-amd, line 1737: There's a lot of machinery here - do we know how it might affect CPU-side perf to do this work?

Wondering if any of these things can be a 1-time init and reusable thing? For instance all `DxcCreateInstance` calls

JerryAMD (approved):

framework/decode/dx12_resource_value_mapper.cpp
davidd-lunarg: Done

framework/decode/dx12_resource_value_mapper.cpp
davidd-lunarg: I did not do any specific performance measuring. This code should only run during unoptimized DXR or EI replay where performance is already heavily affected.

The `dxc_utils` object could be reused but I don't know whether that is a heavy-weight operation. The other classes are based on the incoming DXIL library so the code would need to compare the DXIL library's bytes to detect if previous objects or results could be reused.

framework/decode/dx12_resource_value_mapper.cpp
```diff
@@ -1699,10 +1731,89 @@ void Dx12ResourceValueMapper::GetStateObjectLrsAssociationInfo(
...
+                hr                                        = DxcCreateInstance(CLSID_DxcUtils, IID_PPV_ARGS(&dxc_utils));
+                if (SUCCEEDED(hr))
+                {
+                    graphics::dx12::IDxcContainerReflectionComPtr dxc_container_reflection = nullptr;
+                    DxcCreateInstance(CLSID_DxcContainerReflection, IID_PPV_ARGS(&dxc_container_reflection));
+                    if (SUCCEEDED(hr))
+                    {
+                        // Create a DXC blob from the DXIL library's bytes.
+                        graphics::dx12::IDxcBlobEncodingComPtr dxc_blob_encoding;
+                        hr = dxc_utils->CreateBlob(dxil_lib_desc_decoder->GetPointer()->DXILLibrary.pShaderBytecode,
```
dgraves, line 1748: I think it should be possible to use CreateBlobFromPinned here to avoid a copy to a new allocation,

davidd-lunarg: Changed `CreateBlob` to `CreateBlobFromPinned`.

PR #1711, opened 2024-09-04
Adjust trimming state snapshot locking

dgraves:
The exclusive lock that was intended to be acquired when writing the trimming state snapshot was only acquired when trimming was activated by a present call, where the exclusive lock was always acquired, which also had a negative impact on capture performance. This change makes the following adjustments:
- Acquire the shared lock at present, instead of the exclusive lock.
- Release the shared lock and acquire the exclusive lock when activating trimming.
- Acquire the exclusive lock when trimming is activated by API calls other than present.

### DIFF
--- a/framework/encode/capture_manager.cpp
+++ b/framework/encode/capture_manager.cpp
@@ -682,15 +682,16 @@ bool CommonCaptureManager::RuntimeTriggerDisabled()
     return result;
 }
 
-void CommonCaptureManager::CheckContinueCaptureForWriteMode(format::ApiFamilyId api_family,
-                                                            uint32_t            current_boundary_count)
+void CommonCaptureManager::CheckContinueCaptureForWriteMode(format::ApiFamilyId              api_family,
+                                                            uint32_t                         current_boundary_count,
+                                                            std::shared_lock<ApiCallMutexT>& current_lock)
 {
     if (!trim_ranges_.empty())
     {
         if (current_boundary_count == (trim_ranges_[trim_current_range_].last + 1))
         {
             // Stop recording and close file.
-            DeactivateTrimming();
+            DeactivateTrimming(current_lock);
             GFXRECON_LOG_INFO("Finished recording graphics API capture");
 
             // Advance to next range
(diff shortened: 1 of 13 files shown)


davidd-lunarg (approved):

framework/encode/capture_manager.cpp
```diff
@@ -1048,26 +1051,56 @@ bool CommonCaptureManager::CreateCaptureFile(format::ApiFamilyId api_family, con
...
 }
 
-void CommonCaptureManager::ActivateTrimming()
+void CommonCaptureManager::ActivateTrimming(std::shared_lock<ApiCallMutexT>& current_lock)
 {
-    capture_mode_ |= kModeWrite;
+    auto has_shared_lock = current_lock.owns_lock();
+    if (has_shared_lock)
+    {
+        current_lock.unlock();
```
panos-lunarg, line 1059: By the time `unlock()` returns and until `AcquireExclusiveApiCallLock()` gets the `unique_lock` there's no locking at all. Can this cause problems?

dgraves: Not in this case. Other threads can acquire the shared lock up until the current thread acquires the unique lock, regardless of whether the current thread does or does not hold the shared lock. The current thread just needs to acquire the unique lock so that it can prevent other threads from modifying API state while it transitions from tracking mode to writing mode and writes the state snapshot, to avoid a potential race condition around the mode transition.

panos-lunarg (approved):

davidd-lunarg (changes requested):
~~When testing this on my machine the captured application becomes unresponsive (possible deadlock?) when trimming is activated via frame range or trigger key. Could you verify whether you see the same behavior. I tested on Windows 10 with the D3D12HelloTriangle and D3D12Multithreading sample applications.~~

This may be due to some local system changes. I will retest and get back.

davidd-lunarg:
> When testing this on my machine the captured application becomes unresponsive (possible deadlock?) when trimming is activated via frame range or trigger key. Could you verify whether you see the same behavior. I tested on Windows 10 with the D3D12HelloTriangle and D3D12Multithreading sample applications.

I've retested locally. This issue happens when `GFXRECON_FORCE_COMMAND_SERIALIZATION` is enabled.

dgraves:
> I've retested locally. This issue happens when `GFXRECON_FORCE_COMMAND_SERIALIZATION` is enabled.

This is most likely due to each API call acquiring the unique lock instead of the shared lock when `GFXRECON_FORCE_COMMAND_SERIALIZATION` is enabled. In this case the `ActivateTrimming` function would be attempting to acquire the unique lock a second time, which is undefined behavior for std::shared_mutex. I'll make an update to support the `GFXRECON_FORCE_COMMAND_SERIALIZATION` case.

davidd-lunarg (approved):

PR #1533, opened 2024-05-02
Vulkan portable raytracing

bartosz-muszarski-arm:

PR #1712, opened 2024-09-05
Track Buffer-Device-Addresses

fabian-lunarg:
- Add `VulkanBufferTracker` class -> map device-addresses (ranges) to BufferInfo*, O(log n)
- Add `OverrideGetBufferDeviceAddress`, track capture/replay buffer-addresses during replay
- Add `OverrideCmdTraceRaysKHR`, sprinkle asserts using buffer-tracker queries
- Augment `OverrideBindBufferMemory` with tracking logic

Building up necessary foundation as demonstrated by @bartosz-muszarski-arm's  [portable-raytracing branch](https://github.com/LunarG/gfxreconstruct/pull/1533)

### DIFF
--- a/framework/decode/vulkan_buffer_tracker.cpp
+++ b/framework/decode/vulkan_buffer_tracker.cpp
@@ -38,11 +38,6 @@ void decode::VulkanBufferTracker::TrackBuffer(const decode::BufferInfo* buffer_i
         {
             _capture_addresses[buffer_info->capture_address] = buffer_info->capture_id;
         }
-
-        if (buffer_info->replay_address != 0)
-        {
-            _replay_addresses[buffer_info->replay_address] = buffer_info->capture_id;
-        }
     }
 }
 
@@ -51,14 +46,7 @@ void VulkanBufferTracker::RemoveBuffer(const BufferInfo* buffer_info)
     if (buffer_info != nullptr)
     {
         _capture_addresses.erase(buffer_info->capture_address);
-        _replay_addresses.erase(buffer_info->replay_address);
     }
-}
-
-const decode::BufferInfo*
-decode::VulkanBufferTracker::GetBufferByReplayDeviceAddress(VkDeviceAddress replay_address) const
-{
-    return GetBufferInfo(replay_address, _replay_addresses);
 }
 
 const decode::BufferInfo*
--- a/framework/decode/vulkan_buffer_tracker.cpp
+++ b/framework/decode/vulkan_buffer_tracker.cpp
@@ -73,7 +73,7 @@ const BufferInfo* VulkanBufferTracker::GetBufferInfo(VkDeviceAddress            
     if (!address_map.empty())
     {
         // find first address equal or greater
-        auto address_it = address_map.upper_bound(device_address);
+        auto address_it = address_map.lower_bound(device_address);
 
         if (address_it == address_map.end() || address_it->first > device_address)
         {
--- a/framework/decode/vulkan_buffer_tracker.cpp
+++ b/framework/decode/vulkan_buffer_tracker.cpp
@@ -30,7 +30,7 @@ VulkanBufferTracker::VulkanBufferTracker(const VulkanObjectInfoTable& object_inf
     _object_info_table(object_info_table)
 {}
 
-void decode::VulkanBufferTracker::TrackBuffer(decode::BufferInfo* buffer_info)
+void decode::VulkanBufferTracker::TrackBuffer(const decode::BufferInfo* buffer_info)
 {
     if (buffer_info != nullptr)
     {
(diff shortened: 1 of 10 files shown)


fabian-lunarg:
belongs to [portable raytracing issue](https://github.com/LunarG/gfxreconstruct/issues/1526)

framework/decode/vulkan_buffer_tracker.cpp
```diff
@@ -0,0 +1,105 @@
...
+        {
+            // not found
+            if (address_it == address_map.begin())
+            {
+                return nullptr;
+            }
+
+            // decrement iterator, now pointing to the first VkDeviceAddress that is lower than device_address
+            address_it--;
+        }
```
charles-lunarg, line 88: Wouldn't upper_bound work here as well as it gets the "same value or next lowest"? or does this need to get the next lowest if its the same value?

The logic here is difficult to follow, so I want to be doubly sure I understand it before approving.

fabian-lunarg: yes, you're right both `lower_bound` and `upper_bound` would work here without modifying anything else. 

but there is a minor difference: 
lower_bound -> key >= device_address
upper_bound -> key > device_address

`lower_bound` 'can' yield a key **equal** to device_address, allowing to skip the following if-statement,
while upper_bound will always be greater than device_address, which will be followed by the iterator-decrement.

so, with both lower_bound and upper_bound we are in the exact same situation after the if-statement:
-> `// found_address is lower or equal to device_address`

fabian-lunarg: side-comment:
lower_bound/upper_bound make slightly more sense for a `std::multimap`,
there defining a sub-range with equal keys.
so in cases like this, for regular `std::map`, they appear 'almost' identical in practice

per-mathisen-arm:
I suspect that in the end we won't actually need the replayer address lookup functionality, just for the capture address.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7782,22 +7808,39 @@ VkDeviceAddress VulkanReplayConsumerBase::OverrideGetBufferDeviceAddress(
...
     }
+    VkDevice                         device       = device_info->handle;
+    const VkBufferDeviceAddressInfo* address_info = pInfo->GetPointer();
+
+    // retrieve replay-time device-address
+    VkDeviceAddress replay_device_address = func(device, address_info);
 
     if (!device_info->allocator->SupportsOpaqueDeviceAddresses())
     {
+        // TODO: make this warning obsolete by re-mapping addresses (and a whole lot more ...)
```
MarkY-LunarG, line 7819: Should we make  an issue for this?  I'd like to have issues tracking TODOs that aren't going to be fixed in the near term (say 1-2 days) just so we don't loose track of this if you get redirected to something else in the near term.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -8890,6 +8990,17 @@ void VulkanReplayConsumerBase::UpdateDescriptorSetInfoWithTemplate(DescriptorSet
     }
 }
 
+VulkanBufferTracker& VulkanReplayConsumerBase::GetBufferTracker(VkDevice device)
+{
+    auto it = _buffer_trackers.find(device);
+    if (it == _buffer_trackers.end())
+    {
+        auto [new_it, success] = _buffer_trackers.insert({ device, VulkanBufferTracker(object_info_table_) });
```
MarkY-LunarG, line 8998: What is the ", success" part doing?  It's not used later so if it's receiving something, it's wasted.

MarkY-LunarG (commented):
A few questions.

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: we do already track the progress for the overall initiative [here](https://github.com/LunarG/gfxreconstruct/issues/1526), but should make that clear in the TODO, under which condition/when we'll remove it again. will do!

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: yes, true, we don't really need it. guess was just drinking the structured bind cool-aid to assign variable-names (vs. first/second from std::pair) to the return pair (and normally no compiler-warnings for unused elements). will change

fabian-lunarg: the success-value would be used to check if we actually inserted something (or if an element was already present for that key). but that's obsolete here as you point out

framework/decode/vulkan_replay_consumer_base.cpp
MarkY-LunarG: No need.  If it's tracked that's all.

framework/decode/vulkan_replay_consumer_base.cpp
MarkY-LunarG: @bradgrantham-lunarg mentioned this isn't a big deal.  So don't worry about it.

MarkY-LunarG (approved):

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: slightly improved wording for the TODO and added the issue-number, so they hopefully won't stay there for long :)

fabian-lunarg:
many thanks @charles-lunarg & @MarkY-LunarG!

PR #1736, opened 2024-09-12
openxr: Relocate the prelock call to be prelock

jzulauf-lunarg:
also, move custom handle expansion into locked section


MarkY-LunarG:
Looks like you're fixing a bad change of mine.  Thanks

PR #1739, opened 2024-09-13
Expose 2 symbols from android_main

panos-lunarg:
MainGetCurrentBlockIndex and MainGetLoadingTrimmedState whould de discoveralbe through dlopen + dlsym from the android replay native .so:

- MainGetCurrentBlockIndex(): Returns the current block index as calculated from the file processor

- MainGetLoadingTrimmedState(): Returns wheter file processor is between a FrameStateBegin/FrameStateEnd markers pair

### DIFF
--- a/framework/decode/mark_injected_commands.cpp
+++ b/framework/decode/mark_injected_commands.cpp
@@ -1,0 +1,75 @@
+/*
+** Copyright (c) 2024 Valve Corporation
+** Copyright (c) 2024 LunarG, Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and associated documentation files (the "Software"),
+** to deal in the Software without restriction, including without limitation
+** the rights to use, copy, modify, merge, publish, distribute, sublicense,
+** and/or sell copies of the Software, and to permit persons to whom the
+** Software is furnished to do so, subject to the following conditions:
+**
+** The above copyright notice and this permission notice shall be included in
+** all copies or substantial portions of the Software.
+**
+** THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+** IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+** FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+** AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+** FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+** DEALINGS IN THE SOFTWARE.
+*/
+
+#include "util/defines.h"
+#include "util/logging.h"
+#include "mark_injected_commands.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(decode)
+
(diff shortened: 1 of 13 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7440,13 +7493,11 @@ VkResult VulkanReplayConsumerBase::OverrideCreateAccelerationStructureKHR(
                 capture_id);
         }
 
-        result = device_table->CreateAccelerationStructureKHR(
-            device, &modified_create_info, GetAllocationCallbacks(pAllocator), replay_accel_struct);
+        result = func(device, &modified_create_info, GetAllocationCallbacks(pAllocator), replay_accel_struct);
```
bradgrantham-lunarg, line 7496: This doesn't seem related to the PR purpose - is this and the other changes to the RT calls where you pass in `func` a bugfix unrelated to the block number and marking generated commands?

panos-lunarg: I was looking for injected commands by searching for places where `device_table` is referenced to fetch the api call function pointer. While doing so I discovered these cases where an api function pointer is fetched from `device_table` unnecessarily while the pointer is already available.
 
Strictly speaking this is not 100% related to PR's purposes but, given the opportunity, I clean up those places. I can revert if you insist.

framework/util/mark_injected_commands.cpp
```diff
@@ -0,0 +1,75 @@
...
+    BeginInjectCommands_fp = begin_fp;
+    EndInjectCommands_fp   = end_fp;
+}
+
+void BeginInjectedCommands()
+{
+#ifdef GFXRECON_DEBUG_BUILD
+    if (injecting_api_calls_g)
+    {
+        GFXRECON_LOG_WARNING("Nested BeginInjectedCommands/EndInjectedCommands");
```
davidd-lunarg, line 51: Nice error check. My understanding of it is that if we hit this message it means that GFXR code is incorrectly using the Begin/End calls. If that is the case this could be upgraded to a `GFXRECON_ASSERT` since it indicates a code problem, not a runtime problem.

framework/util/mark_injected_commands.cpp
```diff
@@ -0,0 +1,75 @@
...
+GFXRECON_BEGIN_NAMESPACE(util)
+
+static void BeginEndInjectedCommandsNoop() {}
+
+#ifdef GFXRECON_DEBUG_BUILD
+static bool injecting_api_calls_g = false;
+#endif
+
+static PFN_BeginInjectedCommands BeginInjectCommands_fp = BeginEndInjectedCommandsNoop;
+static PFN_EndInjectedCommands   EndInjectCommands_fp   = BeginEndInjectedCommandsNoop;
```
davidd-lunarg, line 38: Will modules outside of `decode` need this utility class? If not then moving them to the `decode` namespace and directory could help with code organization, and I think it would prevent the static variables and `extern "C"` method from existing in non-decode libraries that use the `util` library.

davidd-lunarg (approved):

framework/util/mark_injected_commands.cpp
panos-lunarg: good point

framework/decode/mark_injected_commands.cpp
dfriederich, line 1: I wonder if we should have some header/place where all exported api's are declared with some docu.
Right now the exported functions are defined in different cpp files, and if I don't know what to look out for I would not find any connection. (well maybe there is no connection :-) ).

panos-lunarg:
@dfriederich All your comments have good points. I hoped I addressed all of them

framework/decode/vulkan_replay_consumer_base.cpp
bradgrantham-lunarg: No, sounds good to me.  In the future it is slightly helpful to separate clean ups into separate PRs so I can go through those more quickly.

bradgrantham-lunarg (approved):

PR #1741, opened 2024-09-13
Fix warnings

dfriederich:
Minor changes to reduce warnings seen when using gfxreconstruct, no (intentional) functional changes.

Using struct as in the definition for forward declaration of the ReplayDeviceInfo tag (clang -Wmismatched-tags)

Explicitly define assignment operators for clang
 -Wdeprecated-copy-with-user-provided-copy warning.
  clang output:
> definition of implicit copy assignment operator for 'ShaderDescriptorInfo'
> is deprecated because it has a user-provided copy constructor


dfriederich:
minor change, mostly to avoid getting warnings when using gfxreconstruct in a project as dependency with enabled warnings.

dfriederich:
@bradgrantham-lunarg  thanks for the format fix :-)

PR #1746, opened 2024-09-18
Remove block_index from thread data

panos-lunarg:


bradgrantham-lunarg (approved):

bradgrantham-lunarg:
The current use case is to do something in a Vulkan layer upchain of GFXR, marking before calling downchain and then marking after calling downchain with the the block number.  In order to exactly correspond to the command just called, having the block number in the thread seems to make the most sense.  Would this PR break that expectation?

PR #1747, opened 2024-09-18
vulkan dump resources: arg parsing fix

panos-lunarg:
Add a check to detect render pass indices that do not have corresponding draw call indices

### DIFF
--- a/tools/replay/parse_dump_resources_cli.cpp
+++ b/tools/replay/parse_dump_resources_cli.cpp
@@ -74,7 +74,7 @@ static bool AreIndicesSorted(const std::vector<uint64_t>& indices)
     return true;
 }
 
-static bool CheckIndicesForErrors(const gfxrecon::decode::VulkanReplayOptions& vulkan_replay_options)
+static bool CheckIndicesForErrors(gfxrecon::decode::VulkanReplayOptions& vulkan_replay_options)
 {
     // Indices should be provided sorted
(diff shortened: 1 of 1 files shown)


davidlunarg (approved):
Looks good to me.

bradgrantham-lunarg (approved):

tools/replay/parse_dump_resources_cli.cpp
```diff
@@ -175,14 +175,40 @@ static bool CheckIndicesForErrors(const gfxrecon::decode::VulkanReplayOptions& v
...
+    for (; draw_it != vulkan_replay_options.Draw_Indices.end() &&
+           rp_it != vulkan_replay_options.RenderPass_Indices.end();)
+    {
+        if (draw_it->empty() || rp_it->empty())
         {
-            GFXRECON_LOG_ERROR("ERROR - incomplete --dump-resources parameters");
-            GFXRECON_LOG_ERROR("Each Draw item must have a corresponding RenderPass item");
-            return true;
+            draw_it = vulkan_replay_options.Draw_Indices.erase(draw_it);
+            rp_it   = vulkan_replay_options.RenderPass_Indices.erase(rp_it);
```
MarkY-LunarG, line 197: It seems weird to me that we're stripping items out in a "CheckIndicesForErrors" call.  If this is performing cleanup, then perhaps a different name for a function like "SanitizeDumpResourcesIndices"?

PR #1752, opened 2024-09-20
Update headers to 1.3.295

nickdriscoll-lunarg:
Update Vulkan headers to 1.3.295.
`VK_KHR_pipeline_binary` was the big new addition, which touched some seemingly-unused paths in the generators which were fixed.


framework/generated/generated_vulkan_enum_to_json.h
```diff
@@ -508,7 +510,6 @@ void FieldToJson(nlohmann::ordered_json& jdata, const VkVideoDecodeUsageFlagBits
 void FieldToJson(nlohmann::ordered_json& jdata, const VkVideoEncodeCapabilityFlagBitsKHR& value, const util::JsonOptions& options = util::JsonOptions());
 void FieldToJson(nlohmann::ordered_json& jdata, const VkVideoEncodeContentFlagBitsKHR& value, const util::JsonOptions& options = util::JsonOptions());
 void FieldToJson(nlohmann::ordered_json& jdata, const VkVideoEncodeFeedbackFlagBitsKHR& value, const util::JsonOptions& options = util::JsonOptions());
-void FieldToJson(nlohmann::ordered_json& jdata, const VkVideoEncodeFlagBitsKHR& value, const util::JsonOptions& options = util::JsonOptions());
```
charles-lunarg, line 511: Why does VkVideoEncodeFlagBitsKHR get removed? Is it a duplicate entry?

framework/generated/generated_vulkan_enum_to_string.cpp
```diff
@@ -4657,16 +4675,6 @@ template <> std::string ToString<VkVideoEncodeFeedbackFlagBitsKHR>(VkFlags vkFla
...
-template <> std::string ToString<VkVideoEncodeFlagBitsKHR>(const VkVideoEncodeFlagBitsKHR& value, ToStringFlags, uint32_t, uint32_t)
-{
-    return "Unhandled VkVideoEncodeFlagBitsKHR";
-}
-
-template <> std::string ToString<VkVideoEncodeFlagBitsKHR>(VkFlags vkFlags, ToStringFlags, uint32_t, uint32_t)
-{
-    return BitmaskToString<VkVideoEncodeFlagBitsKHR>(vkFlags);
-}
-
```
charles-lunarg, line 4669: Same here, VkVideoEncodeFlagBitsKHR seems to be removed.

framework/generated/generated_vulkan_enum_to_string.h
```diff
@@ -317,8 +317,6 @@ template <> std::string ToString<VkVideoEncodeContentFlagBitsKHR>(const VkVideoE
 template <> std::string ToString<VkVideoEncodeContentFlagBitsKHR>(VkFlags vkFlags, ToStringFlags toStringFlags, uint32_t tabCount, uint32_t tabSize);
 template <> std::string ToString<VkVideoEncodeFeedbackFlagBitsKHR>(const VkVideoEncodeFeedbackFlagBitsKHR& value, ToStringFlags toStringFlags, uint32_t tabCount, uint32_t tabSize);
 template <> std::string ToString<VkVideoEncodeFeedbackFlagBitsKHR>(VkFlags vkFlags, ToStringFlags toStringFlags, uint32_t tabCount, uint32_t tabSize);
-template <> std::string ToString<VkVideoEncodeFlagBitsKHR>(const VkVideoEncodeFlagBitsKHR& value, ToStringFlags toStringFlags, uint32_t tabCount, uint32_t tabSize);
-template <> std::string ToString<VkVideoEncodeFlagBitsKHR>(VkFlags vkFlags, ToStringFlags toStringFlags, uint32_t tabCount, uint32_t tabSize);
```
charles-lunarg, line 321: Also here.

framework/generated/generated_vulkan_enum_to_json.h
nickdriscoll-lunarg: For some reason, the `requires` field of `VkVideoEncodeFlagsKHR` was removed from `vk.xml`.
```
-        <type requires="VkVideoEncodeFlagBitsKHR"                   category="bitmask">typedef <type>VkFlags</type> <name>VkVideoEncodeFlagsKHR</name>;</type>
+        <type                                                       category="bitmask">typedef <type>VkFlags</type> <name>VkVideoEncodeFlagsKHR</name>;</type>
```

Perhaps that's causing the generator to miss it somehow? Trying to figure it out.

charles-lunarg: https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6781

I think the requires was removed because nobody used the bitflag. So I believe my concern is misplaced. Yes the codegen removed it but that is expected. When the bitflag becomes used the codegen will re-add it.

framework/generated/vulkan_generators/base_generator.py
```diff
@@ -716,6 +716,9 @@ def make_value_info(self, params):
             else:
                 array_length = self.get_array_len(param)
 
+            # Get array length for HandlesInfo structs e.g. VkPipelineBinaryHandlesInfoKHR
+            
+
```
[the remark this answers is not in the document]
nickdriscoll-lunarg: Removed.

charles-lunarg (approved):
I didn't do a terribly deep review, but all of the things that caught my attention on the review I did do have been fixed or were found to be non-issues.

framework/generated/vulkan_generators/vulkan_replay_consumer_body_generator.py
```diff
@@ -893,14 +893,14 @@ def make_body_expressions(self, return_type, name, values, is_override):
...
-                                                'SetStructHandleLengths<Decoded_{}>({paramname}->GetMetaStructPointer(), {paramname}->GetLength());'
+                                                'SetStructArrayHandleLengths<Decoded_{}>({paramname}->GetMetaStructPointer(), {paramname}->GetLength());'
                                                 .format(
                                                     value.base_type,
                                                     paramname=value.name
                                                 )
                                             )
                                         postexpr.append(
-                                            'AddStructHandles<Decoded_{basetype}>({}, {name}->GetMetaStructPointer(), {}, &GetObjectInfoTable());'
+                                            'AddStructArrayHandles<Decoded_{basetype}>({}, {name}->GetMetaStructPointer(), {name}->GetLength(), {}, {name}->GetLength(), &GetObjectInfoTable());'
```
davidd-lunarg, line 903: The control flow for this block is:    
```
if value.is_pointer or value.is_array:
    if is_input:
    else:
        if value.is_array:
        else:
            if is_extenal_object:
            elif self.is_handle(value.base_type):
            else:
                if self.is_array_len(value.name, values):
                elif self.is_struct(value.base_type):
                    if value.base_type in self.structs_with_handles:
```

So this is generating the code to consume a pointer to a single (non-array) struct, and that struct contains handles. In that case I wonder if it these should be:
`SetStructHandleLengths({paramname}->GetMetaStructPointer());`
and
`AddStructHandles({}, {name}->GetMetaStructPointer(), {}, &GetObjectInfoTable());`

This avoids the call to `SetStructArrayHandleLengths` which in turn calls `SetStructHandleLengths`.  That being said, I believe this code works correctly as-is so I consider this an optional change, especially if it has already been tested.

davidd-lunarg (approved):

PR #1753, opened 2024-09-20
vulkan: Split OverrideCreateInstance

jzulauf-lunarg:
into reusable pieces. The before and after dispatch processing will be needed by OpenXr calls that create vulkan instances.

To trigger CI on the change, and subject to review not necessarily, for commit.


MarkY-LunarG (approved):
Changes look good to me.  Not sure if you want approval from someone else, though.

bradgrantham-lunarg (approved):

Issue #1734, opened 2024-09-11
[capture] trimmed capture of VkSurfaceFullScreenExclusiveWin32InfoEXT crashes


bradgrantham-lunarg:
**Describe the capture bug:**
Capture of vkcube modified to add VkSurfaceFullScreenExclusiveWin32InfoEXT and VkSurfaceFullScreenExclusiveInfoEXT causes crash in GFXR

**Build Environment:**
Windows 10
dev, ed4b4b66da027efc8a4a1897f0ebc8d58be633bc

**Capture Settings:**
VK_INSTANCE_LAYERS=VK_LAYER_LUNARG_gfxreconstruct VK_LAYER_PATH=$GFXR_ROOT/build/layer/Debug GFXRECON_CAPTURE_TRIGGER=F5  vkcube.exe

**To Reproduce:**
Apply the attached [patch.txt](https://github.com/user-attachments/files/16967622/patch.txt) to vkcube

**Screenshots:**
NA

**System environment:**
RTX 2070 Super, 551.76
However, crash is in GFXR and appears unrelated to the driver or GPU

**Additional information (optional):**
Capture succeeds in f6e175620b7fbf7554d60abc06c72a30223e809d

In particular capture succeeds if this line is commented, but that is just for investigation, not a fix:
https://github.com/LunarG/gfxreconstruct/blob/544fb679efa651542c3a1938a8221c2361967083/framework/generated/generated_vulkan_struct_deep_copy.cpp#L15211

dgraves:
For some additional details about the crash, an access violation is raised when a generated struct copy function specifies an opaque handle as the source parameter for a memcpy call. 

This commit adds logic to the struct copy code generator to detect these types of opaque objects and avoid attempting to copy them: fb10c2bbc4b88969885389aa0aa58e17bea73049

There are some additional details in the commit message. The commit could be used as a starting point for a fix. There is also an additional consideration, as noted in the commit message, needed for VkPushDescriptorSetWithTemplateInfoKHR::pData, which would require a custom copy routine based on the layout of the descriptor update template.

PR #1756, opened 2024-09-24
Fix Capture Crash Caused by Error in Handling VkSurfaceFullScreenExclusiveWin32InfoEXT

mizhen:
**The problem**
   
When generating deep copy source code for VkSurfaceFullScreenExclusiveWin32InfoEXT, the member "hmonitor" should be
treated as a non-pointer member because it is an opaque pointer defined by the Windows platform. For a valid hmonitor, it is guaranteed that it is not a null pointer, but it may not access the pointed memory. When capturing a specific title, the capturing crashed due to accessing memory through the pointer, causing an access violation.

**The solution**

The commit changed the related source generator to use the member "hmonitor" of VkSurfaceFullScreenExclusiveWin32InfoEXT as a non-pointer member. It does not generate any source code to copy its pointed data.

**Result**

Tested the target title with the pull request build and found that it resolved the access violation.

### DIFF
--- a/framework/generated/generated_vulkan_struct_deep_copy.cpp
+++ b/framework/generated/generated_vulkan_struct_deep_copy.cpp
@@ -15208,7 +15208,6 @@ size_t vulkan_struct_deep_copy(const VkSurfaceFullScreenExclusiveWin32InfoEXT* s
         }
         auto handle_pointer_member = create_pointer_helper_function(base_struct, i, offset, out_data);
         handle_pnext(base_struct, i, offset, out_data);
-        handle_pointer_member(base_struct.hmonitor, 1);
     }
     return offset;
 }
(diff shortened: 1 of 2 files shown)


fabian-lunarg:
hello @mizhen,

many thanks, your observation is absolutely correct, we must not try to de-reference those opaque `void*` handles.
we' re already tracking the [issue](https://github.com/LunarG/gfxreconstruct/issues/1734) and are reviewing a [fix-PR](https://github.com/LunarG/gfxreconstruct/pull/1738) for it.
there are other cases affected too, e.g. `VkWin32SurfaceCreateInfoKHR::handle`, `VkViSurfaceCreateInfoNN::window`, `VkIOSSurfaceCreateInfoMVK::pView`.
so we exclude all opaque+external handle-members and avoid treating them as pointers.

fabian-lunarg:
closing this PR in favor of another [fix-PR](https://github.com/LunarG/gfxreconstruct/pull/1738),
applying a slightly more general fix and covering additional cases.

PR #1757, opened 2024-09-24
Fix Capture Crash Caused by Error in Parsing Shader Code

mizhen:
**The problem**
   
Certain optimizations may remove SPIR-V instructions such as OpName (actually some title may remove all such instructions with shader optimization tool), which can lead to a nullptr name in the spirv_reflect SpvReflectDescriptorBinding struct. This occurs because spirv_reflect sets the name by keeping track of these SPIR-V instructions. The error causes an access violation issue when parsing the shader code.

**The solution**

The commit addresses the issue by setting the root_name to empty in handling this scenario when parsing the shader code.

**Result**


fabian-lunarg:
cool, thank you @mizhen!

fabian-lunarg (approved):

davidd-lunarg (approved):

PR #1759, opened 2024-09-25
Add acceleration structure overrides

fabian-lunarg:
- Add code-overrides for vulkan's acceleration-structure API
- Extend address-tracking to cover acceleration-structures
- Start with device-address replacement for binding-tables and acceleration-structure buffers
- Addition to capture device-address state-tracking
- Add tracking for physical-device raytracing-properties 
- Leave TODOs 

belongs to [portable raytracing initiative](https://github.com/LunarG/gfxreconstruct/issues/1526)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7556,22 +7556,63 @@ void VulkanReplayConsumerBase::OverrideCmdBuildAccelerationStructuresKHR(
...
+                case VK_GEOMETRY_TYPE_INSTANCES_KHR:
+                {
+                    const auto& instances = geometry->geometry.instances;
+                    // TODO: inside buffer:
+                    // replace all VkAccelerationStructureInstanceKHR::accelerationStructureReference
+                    address_remap(instances.data.deviceAddress);
+                    break;
+                }
+                default:
+                    break;
```
MarkY-LunarG, line 7606: Please add an error and/or assert here so we can catch missing cases in the future.

framework/encode/vulkan_state_writer.cpp
```diff
@@ -1202,6 +1193,25 @@ void VulkanStateWriter::WriteTlasToBlasDependenciesMetadata(const VulkanStateTab
...
+{
+    state_table.VisitWrappers([&](const vulkan_wrappers::PhysicalDeviceWrapper* wrapper) {
+        assert(wrapper != nullptr);
+
+        if (wrapper->ray_tracing_pipeline_properties != std::nullopt)
+        {
+            parameter_stream_.Clear();
+            encoder_.EncodeHandleIdValue(wrapper->handle_id);
+            VkPhysicalDeviceProperties2 properties2 = {};
+            properties2.sType                       = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2;
```
MarkY-LunarG, line 1206: Need to check we're Vulkan 1.1 or greater, or enable and/or call KHR if it is available.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7945,8 +7946,11 @@ VkDeviceAddress VulkanReplayConsumerBase::OverrideGetBufferDeviceAddress(
     VkDevice                         device       = device_info->handle;
     const VkBufferDeviceAddressInfo* address_info = pInfo->GetPointer();
 
+    // TODO: we get back a noop here due to always encoding KHR flavor
```
MarkY-LunarG: This sounds like something that should have an issue tracking if you didn't fix it by the end.

MarkY-LunarG (commented):
Looks good in general.  I'm a little concerned that you're using core and extension entrypoints without checking to make sure they were enabled or that the version of Vulkan supports it (in several locations I pointed out).  Please look those over.

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -798,52 +799,56 @@ VkResult VulkanCaptureManager::OverrideCreateBuffer(VkDevice
...
-                         ((pCreateInfo->usage & address_usage_flags) != address_usage_flags)))
-    {
-        // If the buffer has shader device address usage, but the device address capture replay flag was not set, it
-        // needs to be set here.  We create copy from an override to prevent the modified pCreateInfo from being
-        // written to the capture file.
-        modified_create_info.flags |= address_create_flags;
-        modified_create_info.usage |= address_usage_flags;
+            if (address_usage_flags2_KHR != nullptr)
+            {
+                *address_usage_flags2_KHR |= VK_BUFFER_USAGE_2_SHADER_DEVICE_ADDRESS_BIT_KHR;
```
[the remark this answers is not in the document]
fabian-lunarg: I don't think so. the original application used the pNext-chain to pass in [https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkBufferUsageFlags2CreateInfoKHR.html](VkBufferUsageFlags2CreateInfoKHR). this effectively overrides the old usage-flag, so we have to do the same in replay.  this option was added with `VK_KHR_maintenance5` but since we know the original application used it here, an additional check is not required for a faithful replay?

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: done

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: I confirmed the confusion around vkGetBufferDeviceAddressKHR (w. wrong KHR suffix) to be a bug related to trimming, which was/is present in dev.
created[ this issue](https://github.com/LunarG/gfxreconstruct/issues/1778) to track it and left a TODO with the issue-#

MarkY-LunarG: Nice.  Thanks!

framework/encode/vulkan_capture_manager.cpp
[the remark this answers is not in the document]
MarkY-LunarG: Ah, good point!

MarkY-LunarG (approved):

PR #1768, opened 2024-09-26
Vulkan dump resources: Add support for R32G32B32A32_SFLOAT images

panos-lunarg:


framework/util/image_writer.cpp
```diff
@@ -566,6 +566,49 @@ static const uint8_t* ConvertIntoTemporaryBuffer(uint32_t    width,
...
+            for (uint32_t y = 0; y < height; ++y)
+            {
+                for (uint32_t x = 0; x < 4 * width; x += 4)
+                {
+                    float float_r = float_vals[x];
+                    float float_g = float_vals[x + 1];
+                    float float_b = float_vals[x + 2];
+                    float float_a = float_vals[x + 3];
+
+                    uint8_t r = static_cast<uint8_t>(float_r * 255.0f);
```
bradgrantham-lunarg, line 582: If the user wants the range [0.0f,1.0f] to be evenly distributed among uint8_t values, I think `255.0f` should be `std::nextafter(256.0f, 0.0f)` (which will equate to 255.99...f).  Also, `SFLOAT` isn't normalized so you probably need to clamp the source values here to [0.0f, 1.0f].  Maybe some day in the future we can go back and alter the other case statements to match but not in this PR.

bradgrantham-lunarg (changes requested):

framework/util/image_writer.cpp
bradgrantham-lunarg: To prevent this feedback from blocking this PR when it applies to all other cases in image_writer.cpp and not just this new addition, I've moved this feedback to https://github.com/LunarG/gfxreconstruct/issues/1790 and will remove my feedback on this PR.  Note that values will not be distributed uniformly from [0,1] to [0,255] and values outside the range [0,1] will wrap.

bradgrantham-lunarg (approved):

panos-lunarg:
Submitted https://github.com/LunarG/gfxreconstruct/pull/1799 instead

PR #1774, opened 2024-09-27
Two bugfixes to enable OpenXR-SDK-Source HelloXr support --api Vulkan

jzulauf-lunarg:
1) Ensure decoded base header children have defined state

    The lower level decoders are always setting values, so force a good
    state at initialization (zeros)

2) Make view local viewing 1.0 compatible

    Convert xrLocateSpaces calls (which are 1.1 only) to xrLocateSpace -- which is available on XR 1.0


MarkY-LunarG (approved):
LGTM

PR #1780, opened 2024-10-02
Add log message on app gain/loss of focus and when unpausing application

marius-pelegrin-arm:
It makes more sense to have messages for both when the application is paused and unpaused than only when paused.
In addition, in the context of an automated process, it is useful to be able to differentiate, only with the logs, an application that was paused because of a manual input to an application paused because of a lose of focus.

This commit adds these log messages

### DIFF
--- a/framework/application/application.cpp
+++ b/framework/application/application.cpp
@@ -181,16 +181,20 @@ void Application::Run()
 
 void Application::SetPaused(bool paused)
 {
-    paused_ = paused;
-
-    if (paused_ && (file_processor_ != nullptr))
+    if (file_processor_ != nullptr && paused_ != paused)
     {
         uint32_t current_frame = file_processor_->GetCurrentFrameNumber();
-        if (current_frame > 0)
+        if (paused)
         {
-            GFXRECON_LOG_INFO("Paused at frame %u", file_processor_->GetCurrentFrameNumber());
+            GFXRECON_LOG_INFO("Paused at frame %u", current_frame);
+        }
+        else
+        {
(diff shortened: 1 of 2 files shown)


bradgrantham-lunarg:
This doesn't appear to work like I'd expect.  The pause and resume messages are out of order.  Here's the result from our job 4940, before my commit to your branch:
```
10-02 07:51:32.388  4998  5031 I gfxrecon: Unpaused from frame 0
10-02 07:51:32.446  4998  5031 I gfxrecon: Paused at frame 0
```

framework/application/application.cpp
```diff
@@ -181,16 +181,20 @@ void Application::Run()
...
 void Application::SetPaused(bool paused)
 {
-    paused_ = paused;
-
-    if (paused_ && (file_processor_ != nullptr))
+    if (file_processor_ != nullptr && paused_ != paused)
     {
         uint32_t current_frame = file_processor_->GetCurrentFrameNumber();
-        if (current_frame > 0)
+        if (paused_)
```
bradgrantham-lunarg, line 187: Is it possible this should be `paused`?  `paused_` (with underscore) is the _previous_ value of the pause, so this would invert the messages, I think.

bradgrantham-lunarg (changes requested):

framework/application/application.cpp
marius-pelegrin-arm: Yes, this is a mistake on my side, I will fix that =)

MarkY-LunarG (approved):
Issues have been fixed.  I've verified.

PR #1698, opened 2024-08-30
Initial support for D3D11

dgraves:
This is the code that was originally submitted as #1267, adding support for D3D11, rebased on the latest dev branch with some additional bug fixes and enhancements.

PR #1785, opened 2024-10-04
Adjust output array sizes for D3D12 replay

dgraves:
This PR adds D3D12 replay support for adjusting the size of the memory allocated for variable length output array parameters to match the actual size reported at replay, which may be different than the size reported at capture.  It is based on the implementation of an existing Vulkan replay feature providing the same functionality.

A simple way to test this change is to find an application that calls IDXGIOutput::GetDisplayModeList and make a capture of it running through a remote desktop session. Then replay the capture on the local device. Fewer values will be reported for the remote desktop session than the local session. With this change, replay will report something similar to:
> [gfxrecon] INFO - Replay adjusted the IDXGIOutput::GetDisplayModeList array count: capture count = 1, replay count = 54

Prior to this change replay would report:
> [gfxrecon] WARNING - IDXGIOutput_GetDisplayModeList returned DXGI_ERROR_MORE_DATA, which does not match the value returned at capture S_OK.

The commits submitted with this PR were taken from the branch associated with #1698. Two of the earlier commits adjust array processing for some D3D11 specific cases, to handle the SAL result_bytebuffer annotation and process objects created by functions that do not return an HRESULT value, which modify the same Python code that was later modified for the array size adjustment. They were included to avoid conflicts arising from their removal but could be removed if required.

### DIFF
--- a/framework/decode/dx12_object_info.h
+++ b/framework/decode/dx12_object_info.h
@@ -69,6 +69,23 @@ enum class DxObjectInfoType : uint32_t
 };
 
 //
+// Enumerations for storing variable length array replay sizes.
+//
+enum class VariableLengthArrayIndices : uint32_t
+{
+    kDxgiObjectArrayGetPrivateData = 0,
+    kDxgiOutputArrayGetDisplayModeList,
+    kDxgiOutput1ArrayGetDisplayModeList1,
+    kD3D12ObjectArrayGetPrivateData,
+    kD3D12Device5ArrayEnumerateMetaCommands,
+    kD3D12Device5ArrayEnumerateMetaCommandParameters,
+    kD3D12InfoQueueArrayGetMessage,
+    kD3D12InfoQueueArrayGetStorageFilter,
+    kD3D12InfoQueueArrayGetRetrievalFilter,
+    kD3D12ShaderCacheSessionArrayFindValue
+};
+
+//
 // Structures for storing DirectX object info.
 //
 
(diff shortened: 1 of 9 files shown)


framework/decode/dx12_object_info.h
```diff
@@ -197,11 +214,12 @@ struct DxObjectExtraInfo
...
-    format::HandleId                   capture_id{ format::kNullHandleId };
-    uint64_t                           ref_count{ 1 };
-    uint64_t                           extra_ref{ 0 };
-    std::unique_ptr<DxObjectExtraInfo> extra_info;
+    IUnknown*                                              object{ nullptr };
+    format::HandleId                                       capture_id{ format::kNullHandleId };
+    uint64_t                                               ref_count{ 1 };
+    uint64_t                                               extra_ref{ 0 };
+    std::unique_ptr<DxObjectExtraInfo>                     extra_info;
+    std::unordered_map<VariableLengthArrayIndices, size_t> array_counts;
```
bradgrantham-lunarg, line 222: Will there ever be more than one entry in this map?

dgraves: Yes, currently ID3D12Device5 could have up to 2 entries and ID3D12InfoQueue can have up to 3 entries. D3D11 support would also include up to 8 entries for ID3D11DeviceContext, 2 for ID3D11Device, and 2 for ID3D11ClassInstance.

bradgrantham-lunarg (approved):

PR #1787, opened 2024-10-07
openxr: Unknown events in capture but not replay

MarkY-LunarG:
Some unknown events were triggered in some OpenXR captures on the Meta Quest headset, but are not showing up during my replay run. So, if the event is not known by GFXR and it's not handled, just continue onward as if nothing happened, but keep track of the event in case it shows up at some point in the future.

### DIFF
--- a/framework/decode/openxr_replay_consumer_base.cpp
+++ b/framework/decode/openxr_replay_consumer_base.cpp
@@ -734,8 +734,8 @@ void OpenXrReplayConsumerBase::UpdateState_xrCreateSession(
     HandlePointerDecoder<XrSession>*                   session,
     XrResult                                           replay_result)
 {
-    Decoded_XrSessionCreateInfo* decoded_info   = createInfo->GetMetaStructPointer();
-    SessionData&                 session_data   = AddSessionData(*session->GetPointer());
+    Decoded_XrSessionCreateInfo* decoded_info = createInfo->GetMetaStructPointer();
+    SessionData&                 session_data = AddSessionData(*session->GetPointer());
     session_data.AddGraphicsBinding(MakeGraphicsBinding(decoded_info));
 }
 
@@ -932,19 +932,45 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&           
                 }
                 else if (replay_result == XR_EVENT_UNAVAILABLE)
                 {
-                    // Yield and retry
-                    std::this_thread::sleep_for(std::chrono::nanoseconds(sleep_time));
+                    // If the event that is expected is not known to  us, and we didn't encounter it , then this is a
+                    // bad situtation and needs to be reported.
+                    if (IsHandledEventType(capture_event->type))
+                    {
+                        // Yield and retry
+                        std::this_thread::sleep_for(std::chrono::nanoseconds(sleep_time));
 
-                    if (sleep_time < kMaxSleepLimitNs)
-                    {
-                        // Next time, sleep for double what we initially slept for.  This way if we're just
-                        // spinning, we spin less and less each time.
-                        sleep_time *= 2;
+                        if (sleep_time < kMaxSleepLimitNs)
+                        {
+                            // Next time, sleep for double what we initially slept for.  This way if we're just
+                            // spinning, we spin less and less each time.
+                            sleep_time *= 2;
+                        }
+                        // Clamp it to the max (should stay here from this point forward)
+                        if (sleep_time > kMaxSleepLimitNs)
+                        {
+                            sleep_time = kMaxSleepLimitNs;
+                        }
                     }
-                    // Clamp it to the max (should stay here from this point forward)
-                    if (sleep_time > kMaxSleepLimitNs)
+                    else
                     {
-                        sleep_time = kMaxSleepLimitNs;
+                        // If this is an event we don't know anything about (it is unknown to GFXR), then
+                        // add it to the list of events we check against in the future.
+                        GFXRECON_LOG_WARNING("Unhandled Event %s (0x%x %d) did not occur, proceeding as normal",
+                                             GetEventTypeString(capture_event->type),
+                                             capture_event->type,
+                                             capture_event->type);
+                        unhandled_events_from_capture_.push_back(*capture_event);
+
+                        if (ManageEventDequeSize(unhandled_events_from_capture_))
+                        {
+                            GFXRECON_LOG_WARNING("Unhandled capture events list is size adjusted down to %d",
+                                                 unhandled_events_from_capture_.size());
+                        }
+
+                        // Return the expected replay code just because we don't know how to handle this and
+                        // we never received it (just so we don't assert out).
+                        replay_result = returnValue;
+                        break;
                     }
                 }
                 else
@@ -958,38 +984,13 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&           
         // We timed out and never got the event we expect, so handle it appropriately.
         if (capture_event->type != out_eventData->type)
         {
-            // If the event was one we were supposed to handle, then this is a bad situtation and
-            // needs to be reported.
-            if (IsHandledEventType(capture_event->type))
-            {
-                GFXRECON_LOG_ERROR("Event %s (0x%x %d) never occurred!",
-                                   GetEventTypeString(capture_event->type),
-                                   capture_event->type,
-                                   capture_event->type);
+            GFXRECON_LOG_ERROR("Event %s (0x%x %d) never occurred!",
+                               GetEventTypeString(capture_event->type),
+                               capture_event->type,
+                               capture_event->type);
 
-                // Runtime never gave us the event we were looking for
-                replay_result = XR_ERROR_RUNTIME_FAILURE;
-            }
-            else
-            {
-                // If this is an event we don't know anything about (it is unknown to GFXR), then
-                // add it to the list of events we check against in the future.
-                GFXRECON_LOG_WARNING("Unhandled Event %s (0x%x %d) never occurred, pretending everything is still fine",
-                                     GetEventTypeString(capture_event->type),
-                                     capture_event->type,
-                                     capture_event->type);
-                unhandled_events_from_capture_.push_back(*capture_event);
-
-                if (ManageEventDequeSize(unhandled_events_from_capture_))
-                {
-                    GFXRECON_LOG_WARNING("Unhandled capture events list is size adjusted down to %d",
-                                         unhandled_events_from_capture_.size());
-                }
-
-                // Return the expected replay code just because we don't know how to handle this and
-                // we never received it (just so we don't assert out).
-                replay_result = returnValue;
-            }
+            // Runtime never gave us the event we were looking for
+            replay_result = XR_ERROR_RUNTIME_FAILURE;
         }
     }
     else
@@ -1107,11 +1108,11 @@ void OpenXrReplayConsumerBase::UpdateState_xrAcquireSwapchainImage(
     PointerDecoder<uint32_t>*                                  index,
     XrResult                                                   replay_result)
 {
-    uint32_t    capture_index = *index->GetPointer();
-    uint32_t    out_index     = *index->GetOutputPointer();
+    uint32_t capture_index = *index->GetPointer();
+    uint32_t out_index     = *index->GetOutputPointer();
 
     SwapchainData& swapchain_data = GetSwapchainData(swapchain);
-    replay_result        = swapchain_data.AcquireSwapchainImage(capture_index, out_index);
+    replay_result                 = swapchain_data.AcquireSwapchainImage(capture_index, out_index);
 }
 
 void OpenXrReplayConsumerBase::Process_xrReleaseSwapchainImage(
--- a/framework/decode/openxr_replay_consumer_base.cpp
+++ b/framework/decode/openxr_replay_consumer_base.cpp
@@ -764,7 +764,7 @@ struct EventStrings
     char            name[128];
 };
 static EventStrings events_to_string[] = {
-    { XR_TYPE_UNKNOWN, "Unknown Event Type" },
+    { XR_STRUCTURE_TYPE_MAX_ENUM, "Unknown Event Type" },
     { XR_TYPE_EVENT_DATA_BUFFER, "XR_TYPE_EVENT_DATA_BUFFER" },
     { XR_TYPE_EVENT_DATA_INSTANCE_LOSS_PENDING, "XR_TYPE_EVENT_DATA_INSTANCE_LOSS_PENDING" },
     { XR_TYPE_EVENT_DATA_SESSION_STATE_CHANGED, "XR_TYPE_EVENT_DATA_SESSION_STATE_CHANGED" },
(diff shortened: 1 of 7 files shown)


framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -889,11 +920,42 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
...
+                                   GetEventTypeString(capture_event->type),
+                                   capture_event->type,
+                                   capture_event->type);
+
+                // Runtime never gave us the event we were looking for
+                replay_result = XR_ERROR_RUNTIME_FAILURE;
+            }
+            else
+            {
+                GFXRECON_LOG_WARNING("Unhandled Event %s (0x%x %d) never occurred, pretending everything is still fine",
```
jzulauf-lunarg, line 935: If this isn't a handled event type, should we even go through the timeout loop?  I guess since all events are the same size, we don't have to have been complied with a header that knows about them to capture and match on replay...

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -780,6 +780,19 @@ static EventStrings events_to_string[] = {
...
+    for (size_t ii = 0; ii < len; ++ii)
+    {
+        if (type == events_to_string[ii].type)
+        {
+            return true;
+        }
+    }
+    return false;
+}
+
```
jzulauf-lunarg, line 795: On the "default return" wondering if we should labled it uniquely from XR_TYPE_UNKNOWN, as (ironically) TYPE_UNKNOWN is a known enum.  That an event has XR_TYPE_UNKNOWN (0) is a distinct reporting state from type not found in list of known type enum values. "unknown extension event" ?

jzulauf-lunarg (changes requested):
The loop termination issue is critical.  The 'erase' performance should have a think-about.  The IsHandled  and GetEvent... functions... question of style.

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -821,14 +834,28 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
         }
     }
 
+    // We ignored previous events that were not indicated in the capture yet, so see if this one is in the list already
+    for (size_t ii = 0; ii < received_events_.size(); ++ii)
```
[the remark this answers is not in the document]
MarkY-LunarG: Good catch

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -821,14 +834,28 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
...
+
     XrEventDataBuffer* out_eventData =
         eventData->IsNull() ? nullptr : eventData->AllocateOutputData(1, { XR_TYPE_EVENT_DATA_BUFFER, nullptr });
     InitializeOutputStructNext(eventData);
 
     XrResult replay_result;
 
     // WIP: Put this constant somewhere interesting
-    const uint32_t kRetryLimit      = 10;
+    const uint32_t kRetryLimit      = 16;
```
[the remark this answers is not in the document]
MarkY-LunarG: I like 22-24.  That's about 4ms-16ms waiting total.  I think the event loop should be tight and I don't want to wait for 1/3 of a second.  I'll do 24.

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -821,14 +834,28 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
...
+    // We ignored previous events that were not indicated in the capture yet, so see if this one is in the list already
+    for (size_t ii = 0; ii < received_events_.size(); ++ii)
+    {
+        if (skipped_unhandled_events_[ii].type == capture_event->type)
+        {
+            GFXRECON_LOG_WARNING("Previously recorded but skipped event %s (0x%x, %u)",
+                                 GetEventTypeString(capture_event->type),
+                                 capture_event->type,
+                                 capture_event->type);
+            skipped_unhandled_events_.erase(skipped_unhandled_events_.begin() + ii);
```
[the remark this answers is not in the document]
MarkY-LunarG: Looks like a deque might be better.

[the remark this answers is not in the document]
MarkY-LunarG: Deque does not end up performing the addition copy on erase.

framework/decode/openxr_replay_consumer_base.cpp
MarkY-LunarG: Hmm.... IDK.  I could see waiting a few loops.  Because what if it does come in.  But maybe we should only loop like 3 additional times?   How important do you think it is?

framework/decode/openxr_replay_consumer_base.cpp
MarkY-LunarG: The reason I didn't is in case we end up intentionally ignoring known events for some reason.  But I could consider it at some point in the future.

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -802,16 +832,32 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
...
     {
-        if (received_events_[ii].type == capture_event->type)
+        if (unhandled_events_from_replay_[ii].type == capture_event->type)
         {
             GFXRECON_LOG_WARNING("Previously received event %s (0x%x, %u)",
                                  GetEventTypeString(capture_event->type),
                                  capture_event->type,
                                  capture_event->type);
-            received_events_.erase(received_events_.begin() + ii);
+            unhandled_events_from_replay_.erase(unhandled_events_from_replay_.begin() + ii);
```
jzulauf-lunarg, line 845: INFO instead of WARNING?

jzulauf-lunarg: Also, why are we erasing everything in `unhandled...` before this one?

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -802,16 +832,32 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
...
+    // replay, see if this incoming event is one of those.  If so, remove it from the list and return.
+    for (size_t ii = 0; ii < unhandled_events_from_capture_.size(); ++ii)
+    {
+        if (unhandled_events_from_capture_[ii].type == capture_event->type)
+        {
+            GFXRECON_LOG_WARNING("Previously recorded but skipped event %s (0x%x, %u)",
+                                 GetEventTypeString(capture_event->type),
+                                 capture_event->type,
+                                 capture_event->type);
+            unhandled_events_from_capture_.erase(unhandled_events_from_capture_.begin() + ii);
```
jzulauf-lunarg, line 860: same question -- why everything up to the currently found event?

jzulauf-lunarg: Shouldn't the unhandled_event_from_capture be compared with outEventData in the while loop before?

This is just weeding out the current capture event if the previous instance of that event was unhandled.  If we are just going to keep the last event of a given type, that's fine, but then we shouldn't return, and go look for the new one, having simply erased the old

If we're going to attempt to handle all event in order of capture (up to the LIMIT above), then, this check should be in the while loop, and a match should erase the unhandled_event_from_capture, and retry (maybe resetting the retry count and interval).

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -823,8 +869,8 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
     XrResult replay_result;
 
     // WIP: Put this constant somewhere interesting
-    const uint32_t kRetryLimit      = 10;
-    const int64_t  kMaxSleepLimitNs = 500000000; // 500ms
+    const uint32_t kRetryLimit      = 24;
+    const int64_t  kMaxSleepLimitNs = 50000000; // 50ms
```
jzulauf-lunarg, line 873: with the correct kRetryLimit the kMaxSleepLimitNs is redundant 1 << 24 is 16ms (for a total wait of 32ms)   (yes I know the two variables were originally my idea, but i was being too lazing to compute the total wait of 1 << (retry)  - 1 :)

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -836,43 +882,72 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
...
                 }
                 else if (replay_result == XR_EVENT_UNAVAILABLE)
                 {
-                    // Yield and retry
-                    std::this_thread::sleep_for(std::chrono::nanoseconds(sleep_time));
-
-                    if (sleep_time < kMaxSleepLimitNs)
+                    // If the event that is expected is not known to  us, and we didn't encounter it , then this is a
+                    // bad situtation and needs to be reported.
+                    if (IsHandledEventType(capture_event->type))
```
jzulauf-lunarg, line 914: Why don't we retry for unknown events, especially if we think them, likely and valid?

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -836,43 +882,72 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
             replay_result  = GetInstanceTable(in_instance)->PollEvent(in_instance, out_eventData);
             retry_count++;
 
+            GFXRECON_LOG_INFO("Looking for event %s (0x%x, %u)",
+                              GetEventTypeString(capture_event->type),
+                              capture_event->type,
+                              capture_event->type);
+
+            // If the event returned is not the one we expect, we may still want to keep track of
+            // it or bail.
```
jzulauf-lunarg, line 891: Shouldn't we check for SUCCESS before we inspect the out_eventData?

jzulauf-lunarg (commented):
Much clearer. Still some concerns on how events are being handled.

framework/decode/openxr_replay_consumer_base.cpp
MarkY-LunarG: I don't follow.  It's the beginning + index (or ii) which should be the element.  Why would it be erasing everything before this one?

framework/decode/openxr_replay_consumer_base.cpp
MarkY-LunarG: >  same question -- why everything up to the currently found event?

Same point, it is an erase on a singular iterator.

> Shouldn't the unhandled_event_from_capture be compared with outEventData in the while loop before?

Why, if we already received the captured the event from a previous  `xrPollEvent` call?  We don't need to spin wait for something we already got later.

> This is just weeding out the current capture event if the previous instance of that event was unhandled. If we are just going to keep the last event of a given type, that's fine, but then we shouldn't return, and go look for the new one, having simply erased the old

I don't follow.  Why not handle the event outside of the loop if we already received one.  That means while we were looking for something else, we received the event that the captured event wanted to encounter now. 

For example, let's say we got 5 unknown and unhandled events during capture and then a `XR_TYPE_EVENT_DATA_SESSION_STATE_CHANGED` event.  During  replay, if we get the `XR_TYPE_EVENT_DATA_SESSION_STATE_CHANGED`  and then the 5 unknown events, your logic would only store one unknown event and then spin on all the rest because it would only save one.

framework/decode/openxr_replay_consumer_base.cpp
MarkY-LunarG: Originally, I did, but I thought you implied in a previous review that I should not be.  I'll change it back.

framework/decode/openxr_replay_consumer_base.cpp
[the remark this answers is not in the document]
MarkY-LunarG: I removed having both the time and the limit and switched to just the retry limit of 24 for now.

framework/decode/openxr_replay_consumer_base.cpp
MarkY-LunarG: As I mentioned above, I just abandoned the sleep limit time because the retry limit is good enough.

framework/decode/openxr_replay_consumer_base.cpp
MarkY-LunarG: Probably a good point.

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -821,14 +834,28 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
         }
     }
 
+    // We ignored previous events that were not indicated in the capture yet, so see if this one is in the list already
```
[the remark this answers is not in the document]
MarkY-LunarG: Renamed variables and updated comments.

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -859,59 +904,84 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
...
+            {
+                GFXRECON_LOG_ERROR("Event %s (0x%x %d) never occurred!",
+                                   GetEventTypeString(capture_event->type),
+                                   capture_event->type,
+                                   capture_event->type);
+
+                // Runtime never gave us the event we were looking for
+                replay_result = XR_ERROR_RUNTIME_FAILURE;
+            }
+            else if (options_.ignore_unhandled_unknown_capture_events)
```
jzulauf-lunarg, line 964: is this condition backwards? if ... ignore... then we are storing it?

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -859,59 +904,84 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
...
+            {
+                // If this is an event we don't know anything about (it is unknown to GFXR), then
+                // add it to the list of events we check against in the future.
+                GFXRECON_LOG_WARNING(
+                    "Expected Capture Event %s (0x%x %d) did not occur, but since it is unknown, it will be skipped",
+                    GetEventTypeString(capture_event->type),
```
jzulauf-lunarg, line 970: Since this is on the else of "IsHandled..." isn't this guaranteed to be "Unknown Event"

jzulauf-lunarg (commented):
Still confused as to what the code is trying to do as well as some of the details.  Time to do an interactive review in chat.

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -780,6 +780,19 @@ static EventStrings events_to_string[] = {
     { XR_TYPE_EVENT_DATA_USER_PRESENCE_CHANGED_EXT, "XR_TYPE_EVENT_DATA_USER_PRESENCE_CHANGED_EXT" },
 };
 
+bool IsHandledEventType(XrStructureType type)
+{
+    size_t len = sizeof(events_to_string) / sizeof(EventStrings);
+    for (size_t ii = 0; ii < len; ++ii)
```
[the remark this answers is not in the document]
MarkY-LunarG: I could.

framework/decode/openxr_replay_consumer_base.cpp
jzulauf-lunarg: Discussed in video chat.

MarkY-LunarG:
Abandoning for #1860.

PR #1789, opened 2024-10-07
Fixup AGS library linking and install

davidd-lunarg:
Link against the static AGS library. Update code, documentation, and log messages.


framework/encode/ags_dispatch_table.h
```diff
@@ -47,7 +53,8 @@ struct AgsDispatchTable
     AGS_DRIVEREXTENSIONSDX12_PUSHMARKER    agsDriverExtensionsDX12_PushMarker{ nullptr };
     AGS_DRIVEREXTENSIONSDX12_POPMARKER     agsDriverExtensionsDX12_PopMarker{ nullptr };
     AGS_DRIVEREXTENSIONSDX12_SETMARKER     agsDriverExtensionsDX12_SetMarker{ nullptr };
-    AGS_INIT_NOT_SUPPORTED                 agsInitNotSupported{ nullptr }; // In case, old version dll is used by the application
+    AGS_INIT_NOT_SUPPORTED agsInitNotSupported{ nullptr }; // In case, old version dll is used by the application
```
JerryAMD, line 56: Is the line be formatted?

davidd-lunarg: I believe clang formatted it like this in order to fit the comment on the same line.

JerryAMD (approved):

CMakeLists.txt
```diff
@@ -199,18 +199,23 @@ if(MSVC)
...
+                if (AGS_FOUND)
                     add_definitions(-DGFXRECON_AGS_SUPPORT)
 
                     # The value for option GFXRECON_AGS_SUPPORT gets cached so use a non-cached variable
                     # to determine the final result.
                     set(GFXRECON_AGS_SUPPORT_FINAL ON)
+                else()
+                    message(WARNING "GFXRECON_AGS_SUPPORT was requested but the required dependencies were not found. "
+                            "AGS support will be disabled.")
                 endif() # TARGET AGS::AGS
```
chrislb-amd, line 211: Comment has now changed name, so should be # AGS_FOUND

chrislb-amd:
I wonder if instead of caching binaries for AGS, we should submodule https://github.com/GPUOpen-LibrariesAndSDKs/AGS_SDK directly. Are the any aversions to submodule use?

bradgrantham-lunarg:
> I wonder if instead of caching binaries for AGS, we should submodule https://github.com/GPUOpen-LibrariesAndSDKs/AGS_SDK directly. Are there any aversions to submodule use?

No aversions to submodules, but we've mostly done that to import registry XMLs and headers.

We've mostly avoided having to build binaries for other projects from the source as part of our build - you can see we've included the binaries for the compression libraries in the repo, with varying degrees of success.  That's always been slightly awkward - it seems like an added burden for one platform to need to download the binaries for the other platforms, but that does only occur generally once.

Recently we've added SPIRV-Reflect, which we compile as part of our build.  I think that may have been done as an expedient because we know it's very cross-platform and a straightforward build.  @fabian-lunarg ?

How often does AGS change?  If the binaries would change very infrequently, I would argue having the built artifacts makes some sense.  If it changes frequently, then making building it directly from a submodule is a good idea.

fabian-lunarg:
no aversion against submodules, but it's still not an easy or general choice.
for SPIRV-Reflect (and SPIRV-Headers) we chose the submodule-approach over binaries in `external/precompiled`. the reasons were:

- in general submodules(or FetchContent) give more confidence about which exact version we have, compared with precompiled binaries
- it's one(1) C-file (compiled as static lib) vs. many library-permutations for all platforms. that's much easier to maintain
- it might occasionally require updates (e.g. we ran into seqfaults and could fix those upstream)
- we already were using submodules for Vulkan-Headers, otherwise `FetchContent` might have been an alternative

AGS seems to have a slightly different profile here regarding its build-process, frequency of required updates, number of platforms to cover in `external/precompiled` etc.

davidd-lunarg:
> We've mostly avoided having to build binaries for other projects from the source as part of our build

The AGS GitHub repo includes the binaries at https://github.com/GPUOpen-LibrariesAndSDKs/AGS_SDK/tree/master/ags_lib/lib so GFXR wouldn't need to build them. As far as I am aware, source code is not available for the AGS libraries.

> Are there any aversions to submodule use?

The size of the AGS repo is ~180MB. The files added in this PR are ~36MB, so maybe that influences the decision.

1. Is there a way for a submodule to only download certain files from a repo?
2. Can a submodule be not-default? Non-Windows platforms don't need the AGS libs so it would be good if it was ignored when `git submodule update --init` is run.

CMake has FetchContent which might allow us to accomplish 1 and/or 2 but it does add a new method for managing dependencies in GFXR.

---

It isn't clear to me that submodule provides any major benefit, so my thought is to leave the dependency as is. I could look into FetchContent if that is desired.

davidd-lunarg:
I've updated the code to use CMake's FetchContent module. Now the AGS GitHub is cloned during the configuration step of CMake. It will only be cloned if AGS is enabled and supported, so other platforms won't need to download the files. In my testing it only seemed to clone once for a clean CMake build and subsequent cmake configuration runs finished more quickly.

fabian-lunarg (approved):
nice, good solution with `FetchContent`

chrislb-amd (approved):

PR #1792, opened 2024-10-08
MacOS: Fix compiler-warning about missing switch-cases

fabian-lunarg:
- macOS 14.7, clang-1500.3.9.4

fixes warnings:

```
warning: enumeration value 'VK_INDIRECT_EXECUTION_SET_INFO_TYPE_MAX_ENUM_EXT' not handled in switch [-Wswitch]
warning: 11 enumeration values not handled in switch: 'VK_INDIRECT_COMMANDS_TOKEN_TYPE_DRAW_INDEXED_EXT', 'VK_INDIRECT_COMMANDS_TOKEN_TYPE_DRAW_EXT', 'VK_INDIRECT_COMMANDS_TOKEN_TYPE_DRAW_INDEXED_COUNT_EXT'... [-Wswitch]
```

those warnings were not treated as errors but should have? fyi @bradgrantham @MarkY-LunarG

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -317,10 +317,6 @@ if(BUILD_WERROR)
         target_compile_options(platform_specific INTERFACE /WX)
     else()
         target_compile_options(platform_specific INTERFACE -Werror)
-        if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-            # TODO: Actually fix these warnings
-            target_compile_options(platform_specific INTERFACE -Wno-error=switch)
-        endif()
     endif()
 endif()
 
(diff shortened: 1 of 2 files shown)


fabian-lunarg:
found the reason for compilation to pass despite warnings.
only for clang, we allow this specific type of warning but have a **TODO** to "actually fix them":

https://github.com/LunarG/gfxreconstruct/blob/6a2a225c5f907d094d9a3e7cbfba7af9fd616811/CMakeLists.txt#L322

I think we already did that work and can in fact remove this exception from cmake, happy day :)

bradgrantham-lunarg (approved):

PR #1800, opened 2024-10-09
Fix timing mismatch issue for vkGetQueryPoolResults

mizhen:
**The problem**

Some timing mismatch issue was observed when playing back a particular title's trace. vkGetQueryPoolResults returned VK_NOT_READY at playback time, while it returned VK_SUCCESS at the capture time.

**The solution**

The commit adds changes to handle the timing difference between capture time and playback time.

**Result**

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -3315,12 +3315,13 @@ VkResult VulkanReplayConsumerBase::OverrideGetQueryPoolResults(PFN_vkGetQueryPoo
     VkQueryPool query_pool = query_pool_info->handle;
     size_t      retries    = 0;
 
+    // Some timing mismatch issue was observed when replaying a specific title's trace. At playback time,
+    // vkGetQueryPoolResults returned VK_NOT_READY, while at capture time it returned VK_SUCCESS. Therefore, we handle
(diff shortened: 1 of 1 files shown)


mizhen:
> This looks like you'd lose the timeout implemented in [75bc5b3](https://github.com/LunarG/gfxreconstruct/commit/75bc5b32da36cb172d57ef28f55d14458279c6f1). In order to keep that functionality, should this be converted to a time-based timeout?

Hi @bradgrantham, in the case mentioned in [75bc5b3], an infinite loop occurred in several D3D11 games running with DXVK. It seems that the query always returns VK_NOT_READY, indicating that an error has occurred prior to this call. It's possible that every following query for the same pool would also return VK_NOT_READY. Therefore, the time-based timeout value needs to be very short because it affects the frame rate.

Another approach is similar to the current vkGetFenceStatus handling. It uses command options --skip-get-fence-status or --skip-get-fence-ranges to skip the query. So, if there's no issue with vkGetFenceStatus, there's no timeout. If there's an issue with vkGetFenceStatus, the user can set the option to skip the call. We can implement the same handling, such as using --skip-get-querypool-results or --skip-get-querypool-results-ranges to skip the query for the case that an infinite loop occurred in a specific title. What do you think of this approach?

mizhen:
The issue was addressed in [PR#](https://github.com/LunarG/gfxreconstruct/pull/1843)

Issue #1722, opened 2024-09-05
[openxr] Handle extension filtering and detection


MarkY-LunarG:
The Vulkan GFXR path handles extensions in several ways:

 * Unknown extensions may be filtered out
 * Required extensions are tracked and may cause failures on replay
 
OpenXR should follow the same behavior and pay attention to the same environment variables when necessary.

MarkY-LunarG:
Filter out every Graphics API other than Vulkan  when  enabled.  So intercept "EnumerateExtensions" and filter out things we don't want.  Can we do  this?

PR #1801, opened 2024-10-09
openxr: Disable replay w/ unknown extensions

MarkY-LunarG:
Move some replay options into the base class for sharing among all the paths.  Update the OpenXR replay options to be read during replay execution.
Remove unknown extensions if the replay option is enabled.

### DIFF
--- a/framework/decode/dx_replay_options.h
+++ b/framework/decode/dx_replay_options.h
@@ -56,13 +56,7 @@ struct DxReplayOptions : public ReplayOptions
     bool                 override_object_names{ false };
     bool                 enable_dump_resources{ false };
     DumpResourcesTarget  dump_resources_target{};
-
-    util::ScreenshotFormat       screenshot_format{ util::ScreenshotFormat::kBmp };
-    std::vector<ScreenshotRange> screenshot_ranges;
-    std::string                  screenshot_dir;
-    std::string                  screenshot_file_prefix{ kDefaultScreenshotFilePrefix };
-    std::string                  replace_dir;
-    int32_t                      memory_usage{ kDefaultBatchingMemoryUsage };
+    int32_t              memory_usage{ kDefaultBatchingMemoryUsage };
 };
 
 GFXRECON_END_NAMESPACE(decode)
--- a/framework/decode/openxr_feature_util.cpp
+++ b/framework/decode/openxr_feature_util.cpp
@@ -1,0 +1,125 @@
+/*
+** Copyright (c) 2020-2024 LunarG, Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and associated documentation files (the "Software"),
+** to deal in the Software without restriction, including without limitation
+** the rights to use, copy, modify, merge, publish, distribute, sublicense,
+** and/or sell copies of the Software, and to permit persons to whom the
+** Software is furnished to do so, subject to the following conditions:
+**
+** The above copyright notice and this permission notice shall be included in
+** all copies or substantial portions of the Software.
+**
+** THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+** IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+** FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+** AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+** FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+** DEALINGS IN THE SOFTWARE.
+*/
+
+#ifdef ENABLE_OPENXR_SUPPORT
+
+#include "decode/openxr_feature_util.h"
+
+#include "util/logging.h"
+#include "util/platform.h"
+
+#include <cassert>
+#include <set>
+#include <string>
+#include <algorithm>
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(decode)
+GFXRECON_BEGIN_NAMESPACE(feature_util)
+
+XrResult GetApiLayers(PFN_xrEnumerateApiLayerProperties enum_layer_pros, std::vector<XrApiLayerProperties>* layers)
+{
+    assert(layers != nullptr);
+
+    XrResult result = XR_ERROR_INITIALIZATION_FAILED;
+
+    if (enum_layer_pros != nullptr)
+    {
+        uint32_t layer_count = 0;
+        result               = enum_layer_pros(layer_count, nullptr, nullptr);
+
+        if ((result == XR_SUCCESS) && (layer_count > 0))
+        {
+            layers->assign(layer_count, { XR_TYPE_API_LAYER_PROPERTIES, nullptr });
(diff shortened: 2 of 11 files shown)


MarkY-LunarG:
Should resolve #1722

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -554,20 +569,26 @@ void OpenXrReplayConsumerBase::Process_xrCreateApiLayerInstance(
...
+        }
+        last_next_struct = const_cast<XrBaseInStructure*>(next_struct);
+        next_struct      = next_struct->next;
     }
-    init_android.next = reinterpret_cast<const void*>(next_struct);
-
-    XrInstanceCreateInfo new_create_info = *create_info;
-    new_create_info.next                 = &init_android;
-    create_info                          = &new_create_info;
+    init_android.next         = reinterpret_cast<const void*>(next_struct);
```
jzulauf-lunarg, line 584: this is just setting init_android.next to nullptr, but we already updated it by using it as the first last_next_struct

framework/decode/replay_options.h
```diff
@@ -41,25 +41,32 @@ struct ScreenshotRange
...
-    uint32_t    windowed_height{ 0 };
-    bool        force_windowed_origin{ false };
-    int32_t     window_topleft_x{ 0 };
-    int32_t     window_topleft_y{ 0 };
-    int32_t     override_gpu_index{ -1 };
-    std::string capture_filename;
-    bool        enable_print_block_info{ false };
-    int64_t     block_index_from{ -1 };
-    int64_t     block_index_to{ -1 };
+    bool                         enable_validation_layer{ false };
```
jzulauf-lunarg, line 44: If were going to change anything that isn't OXR specific, that should be in a separate commit before the OXR specific one, s.t. we can put it on dev.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5696,7 +5696,7 @@ VkResult VulkanReplayConsumerBase::OverrideCreateShaderModule(
            (pShaderModule != nullptr) && !pShaderModule->IsNull());
 
     auto original_info = pCreateInfo->GetPointer();
-    if (original_result < 0 || options_.replace_dir.empty())
+    if (original_result < 0 || options_.replace_shader_dir.empty())
```
jzulauf-lunarg, line 5699: see other comment re: non OXR changes

tools/replay/android_main.cpp
```diff
@@ -156,16 +156,18 @@ void android_main(struct android_app* app)
                                                      replay_options.preload_measurement_range,
                                                      measurement_file_name);
 
-                vulkan_replay_consumer.SetFatalErrorHandler([](const char* message) { throw std::runtime_error(message); });
+                vulkan_replay_consumer.SetFatalErrorHandler(
```
jzulauf-lunarg, line 159: Accidental change?

tools/tool_settings.h
```diff
@@ -927,8 +928,24 @@ static void GetReplayOptions(gfxrecon::decode::ReplayOptions&      options,
...
+
+    if (arg_parser.IsOptionSet(kRemoveUnsupportedOption))
+    {
+        options.remove_unsupported_features = true;
+    }
+
     IsForceWindowed(options, arg_parser);
     SetWindowOrigin(options, arg_parser);
+
+    options.screenshot_ranges      = GetScreenshotRanges(arg_parser);
```
jzulauf-lunarg, line 945: OXR currently has no screenshot capabilities. (don't know how we'd add without implementing a compositor)

jzulauf-lunarg (changes requested):
Critical: the `while (next_struct != nullptr)` loop has a couple logical errors

Stylistic: at least 3 places where an Algorithms Library approach would be cleaner/more modern

Maintainability: OXR and non-OXR changes mix in a single commit.  Not sure why the vulkan and dx options changes are needed.  If they are, should be in a separate commit

framework/decode/openxr_feature_util.cpp
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+
+    return result;
+}
+
+bool IsSupportedApiLayer(const std::vector<XrApiLayerProperties>& properties, const char* layer)
+{
+    assert(layer != nullptr);
+
```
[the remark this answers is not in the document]
MarkY-LunarG: Sure, although this was a direct copy of the Vulkan code but this is cleaner.

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -554,20 +569,26 @@ void OpenXrReplayConsumerBase::Process_xrCreateApiLayerInstance(
...
+    const XrBaseInStructure* next_struct      = reinterpret_cast<const XrBaseInStructure*>(create_info->next);
+    while (next_struct != nullptr)
     {
-        next_struct = next_struct->next;
+        // Skip any old Android create info structs
+        if (next_struct->type == XR_TYPE_INSTANCE_CREATE_INFO_ANDROID_KHR)
+        {
+            last_next_struct->next = next_struct->next;
+        }
+        last_next_struct = const_cast<XrBaseInStructure*>(next_struct);
```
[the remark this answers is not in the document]
MarkY-LunarG: Good catch.

framework/decode/openxr_replay_consumer_base.cpp
MarkY-LunarG: Not if there is no Android create info.  But I'll see what I can do to clean it up.

framework/decode/openxr_feature_util.cpp
```diff
@@ -0,0 +1,147 @@
...
+    return false;
+}
+
+void RemoveUnsupportedExtensions(const std::vector<XrExtensionProperties>& properties,
+                                 std::vector<const char*>*                 extensions)
+{
+    assert(extensions != nullptr);
+
+    auto extensionIter = extensions->begin();
+    while (extensionIter != extensions->end())
```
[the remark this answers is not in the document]
MarkY-LunarG: This does not look like it would save us any code.  It would be roughly the same as what's here and this code was pulled from the Vulkan-side but re-purposed for OpenXR.  Plus we would have to re-define the IsSupportedExtension behavior above to work better with a non-iterator approach, and that function is being used beyond this internal case.  I think it's best to leave it as is for now.

framework/decode/dx_replay_options.h
```diff
@@ -56,13 +56,7 @@ struct DxReplayOptions : public ReplayOptions
     bool                 override_object_names{ false };
     bool                 enable_dump_resources{ false };
     DumpResourcesTarget  dump_resources_target{};
-
-    util::ScreenshotFormat       screenshot_format{ util::ScreenshotFormat::kBmp };
```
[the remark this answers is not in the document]
MarkY-LunarG: I'm trying to simplify the options so that items are shared.  We can just move this into dev earlier if we want with the Vulkan-specific changes.

framework/decode/replay_options.h
MarkY-LunarG: Now in a separate commit.

framework/decode/vulkan_replay_consumer_base.cpp
MarkY-LunarG: Now in a separate commit.

tools/tool_settings.h
MarkY-LunarG: Yes, but DX and Vulkan both  shared this code.  But it is now outside of the OpenXR-specific changes.

framework/decode/openxr_replay_consumer_base.cpp
MarkY-LunarG: It is now cleaned up.

tools/replay/android_main.cpp
MarkY-LunarG: Not there any more.

jzulauf-lunarg (commented):
LGTM

jzulauf-lunarg (approved):

PR #1804, opened 2024-10-10
Speed up GitHub CI build by limiting ABI and building release and debug separately

bradgrantham-lunarg:


fabian-lunarg (approved):
nice, this is both faster and more in line with the other configs

PR #1805, opened 2024-10-10
fix collision of kViewRelativeLocation

bradgrantham-lunarg:


jzulauf-lunarg (approved):
Seems fine if it works on dev.

PR #1820, opened 2024-10-20
Fix importing AHB memory with rebind

ziga-lunarg:
VMA does not support external memory. If memory is allocated with AHardwareBuffer we must manually allocate it in rebind

### DIFF
--- a/framework/decode/vulkan_rebind_allocator.cpp
+++ b/framework/decode/vulkan_rebind_allocator.cpp
@@ -58,6 +58,7 @@
 #include "format/format_util.h"
 #include "util/logging.h"
 #include "util/platform.h"
+#include "graphics/vulkan_struct_get_pnext.h"
 
 #include "generated/generated_vulkan_enum_to_string.h"
 
@@ -491,6 +492,13 @@ VkResult VulkanRebindAllocator::AllocateMemory(const VkMemoryAllocateInfo*  allo
         (*allocator_data) = reinterpret_cast<uintptr_t>(memory_alloc_info);
 
         result = VK_SUCCESS;
+
+        // Track android hardware buffer
+        if (auto import_ahb_info =
+                graphics::vulkan_struct_get_pnext<VkImportAndroidHardwareBufferInfoANDROID>(allocate_info))
+        {
+            memory_alloc_info->ahb = import_ahb_info->buffer;
+        }
     }
 
     return result;
@@ -506,6 +514,11 @@ void VulkanRebindAllocator::FreeMemory(VkDeviceMemory               memory,
     {
         // Clear references from resources to the allocation info and cleanup allocation info memory.
         auto memory_alloc_info = reinterpret_cast<MemoryAllocInfo*>(allocator_data);
+
+        if (memory_alloc_info->ahb)
+        {
+            functions_.free_memory(device_, memory_alloc_info->ahb_memory, nullptr);
+        }
 
         for (const auto& entry : memory_alloc_info->original_buffers)
         {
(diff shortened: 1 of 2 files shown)


framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -483,6 +483,21 @@ VkResult VulkanRebindAllocator::AllocateMemory(const VkMemoryAllocateInfo*  allo
         (*allocator_data) = reinterpret_cast<uintptr_t>(memory_alloc_info);
 
         result = VK_SUCCESS;
+
+        // Track android hardware buffer
+        const VkBaseOutStructure* pNext = reinterpret_cast<const VkBaseOutStructure*>(allocate_info->pNext);
```
bradgrantham-lunarg, line 488: Can you use `vulkan_struct_get_pnext` for this?

bradgrantham-lunarg:
@mikes-lunarg can you make a minute to look at this?  Thank you!

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -710,64 +718,94 @@ VkResult VulkanRebindAllocator::BindImageMemory(VkImage
...
+            VkMemoryDedicatedAllocateInfo dedicatedAllocateInfo{};
+            dedicatedAllocateInfo.sType = VK_STRUCTURE_TYPE_MEMORY_DEDICATED_ALLOCATE_INFO;
+            dedicatedAllocateInfo.pNext = &importAHBInfo;
+            dedicatedAllocateInfo.image = image;
+
+            VkMemoryAllocateInfo allocate_info{};
+            allocate_info.sType           = VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO;
+            allocate_info.pNext           = &dedicatedAllocateInfo;
+            allocate_info.allocationSize  = memory_alloc_info->allocation_size;
+            allocate_info.memoryTypeIndex = memory_alloc_info->original_index;
```
mikes-lunarg, line 739: This seems to make an assumption that the memory type index won't change between capture and replay. Do we need a TODO or an issue to make this smarter? The rebind allocator is supposed to be the "most portable"

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -710,64 +718,94 @@ VkResult VulkanRebindAllocator::BindImageMemory(VkImage
...
-        create_info.pUserData      = nullptr;
+        if (memory_alloc_info->ahb)
+        {
+            VkImportAndroidHardwareBufferInfoANDROID importAHBInfo;
+            importAHBInfo.sType  = VK_STRUCTURE_TYPE_IMPORT_ANDROID_HARDWARE_BUFFER_INFO_ANDROID;
+            importAHBInfo.pNext  = nullptr;
+            importAHBInfo.buffer = memory_alloc_info->ahb;
+
+            VkMemoryDedicatedAllocateInfo dedicatedAllocateInfo{};
+            dedicatedAllocateInfo.sType = VK_STRUCTURE_TYPE_MEMORY_DEDICATED_ALLOCATE_INFO;
```
mikes-lunarg, line 731: Is dedicated allocation always required for AHB import? Or are you just giving the driver extra info since you know it is dedicated?

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -710,64 +718,94 @@ VkResult VulkanRebindAllocator::BindImageMemory(VkImage
...
+            allocate_info.pNext           = &dedicatedAllocateInfo;
+            allocate_info.allocationSize  = memory_alloc_info->allocation_size;
+            allocate_info.memoryTypeIndex = memory_alloc_info->original_index;
+            result                        = functions_.allocate_memory(device_, &allocate_info, nullptr, &memory);
 
-        VmaAllocationInfo allocation_info;
-        result = vmaAllocateMemoryForImage(allocator_, image, &create_info, &allocation, &allocation_info);
+            if (result >= 0)
+            {
+                assert(allocate_info.memoryTypeIndex < replay_memory_properties_.memoryTypeCount);
```
mikes-lunarg, line 744: I would expect the memory index bounds to be checked before allocating the memory instead of after.

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -794,67 +832,99 @@ VkResult VulkanRebindAllocator::BindImageMemory2(uint32_t                     bi
...
 
-                VmaAllocationInfo allocation_info;
-                result = vmaAllocateMemoryForImage(allocator_, image, &create_info, &allocation, &allocation_info);
+                    if (result >= 0)
+                    {
+                        assert(allocate_info.memoryTypeIndex < replay_memory_properties_.memoryTypeCount);
 
-                if (result >= 0)
+                        result = functions_.bind_image_memory(device_, image, memory, memory_offset);
+                    }
```
mikes-lunarg, line 863: This block looks duplicated from BindImageMemory, does it make sense to put it in a helper?

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -794,67 +832,99 @@ VkResult VulkanRebindAllocator::BindImageMemory2(uint32_t                     bi
...
-                    requirements);
-                create_info.requiredFlags  = 0;
-                create_info.preferredFlags = 0;
-                create_info.memoryTypeBits = 0;
-                create_info.pool           = VK_NULL_HANDLE;
-                create_info.pUserData      = nullptr;
+                auto memory_alloc_info = reinterpret_cast<MemoryAllocInfo*>(allocator_memory_data);
+                if (memory_alloc_info->ahb)
+                {
+                    VkDeviceMemory memory        = bind_infos[i].memory;
```
mikes-lunarg, line 838: Why initialize `memory` from `bind_infos`? It will just get overwritten.

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -710,64 +718,94 @@ VkResult VulkanRebindAllocator::BindImageMemory(VkImage
...
+            dedicatedAllocateInfo.sType = VK_STRUCTURE_TYPE_MEMORY_DEDICATED_ALLOCATE_INFO;
+            dedicatedAllocateInfo.pNext = &importAHBInfo;
+            dedicatedAllocateInfo.image = image;
+
+            VkMemoryAllocateInfo allocate_info{};
+            allocate_info.sType           = VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO;
+            allocate_info.pNext           = &dedicatedAllocateInfo;
+            allocate_info.allocationSize  = memory_alloc_info->allocation_size;
+            allocate_info.memoryTypeIndex = memory_alloc_info->original_index;
+            result                        = functions_.allocate_memory(device_, &allocate_info, nullptr, &memory);
```
mikes-lunarg, line 740: Would prefer not to reuse the input value `memory` to hold the new memory handle. Also it seems like we are going to leak this handle and never free it.

mikes-lunarg (changes requested):

bradgrantham-lunarg (approved):

PR #1827, opened 2024-10-23
Separate dx12 codegen

MarkY-LunarG:
Separate out the DX12 codegen from the Vulkan codegen.  The DX12  codegen was heavily tied in with the Vulkan codegen which leads to difficulty when trying to add in other Khronos API codegen (since we have to be careful of breaking other APIs).


bradgrantham-lunarg:
Removed myself from reviewers in favor of @davidd-lunarg

MarkY-LunarG:
Reverted `yapf` change.

davidd-lunarg (approved):

MarkY-LunarG:
Squashing for merge, once it passes I will merge.

PR #1834, opened 2024-10-24
Have info tool expose SubSysId during GPU enumeration

rurra-amd:
**Problem**
If there are multiple of the same GPUs on a system, using info tool to identify the right GPU isn't straight forward

**Solution**
Expose `SubSysId` beside each enumerated adapter. For some reason DXGI is showing the same adapter twice, so with this change we can identify their uniqueness.

**Testing**
Before (2-GPU system):
```
GPU index       GPU name
0               AMD Radeon RX 7900 XTX
1               AMD Radeon RX 7900 XTX
2               AMD Radeon RX 7900 XTX
3               Microsoft Basic Render Driver
```

After (2-GPU system):
```
GPU index       GPU name                        SubSysId
0               AMD Radeon RX 7900 XTX          123456
1               AMD Radeon RX 7900 XTX          ABCDEF
2               AMD Radeon RX 7900 XTX          123456
3               Microsoft Basic Render Driver   0
```

### DIFF
--- a/tools/info/main.cpp
+++ b/tools/info/main.cpp
@@ -694,7 +694,7 @@ static bool CheckOptionEnumGpuIndices(const char* exe_name, const gfxrecon::util
             gfxrecon::graphics::dx12::ActiveAdapterMap adapters{};
             gfxrecon::graphics::dx12::TrackAdapters(result, reinterpret_cast<void**>(&factory1), adapters);
 
-            GFXRECON_WRITE_CONSOLE("GPU index\tGPU name");
+            GFXRECON_WRITE_CONSOLE("GPU index\tGPU name\tSubSys ID");
             for (size_t index = 0; index < adapters.size(); ++index)
             {
                 for (auto adapter : adapters)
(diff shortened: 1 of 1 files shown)


davidd-lunarg:
> After (2-GPU system):
> 
> ```
> GPU index       GPU name                        SubSysId
> 0               AMD Radeon RX 7900 XTX          123456
> 1               AMD Radeon RX 7900 XTX          ABCDEF
> 2               AMD Radeon RX 7900 XTX          123456
> 3               Microsoft Basic Render Driver   0
> ```

So if I am reading this correctly, it looks like this lists 2 entries for the same GPU--entries 0 and 2 in this example.
1. Does selecting one or the other result in different behavior?
    a. If not, I am wondering if GFXR should only display 1 of the duplicate GPUs to avoid user confusion.
3. Is it possible to correspond the SybSysId to the physical card?

rurra-amd:
> > After (2-GPU system):
> > ```
> > GPU index       GPU name                        SubSysId
> > 0               AMD Radeon RX 7900 XTX          123456
> > 1               AMD Radeon RX 7900 XTX          ABCDEF
> > 2               AMD Radeon RX 7900 XTX          123456
> > 3               Microsoft Basic Render Driver   0
> > ```
> 
> So if I am reading this correctly, it looks like this lists 2 entries for the same GPU--entries 0 and 2 in this example.
> 
> 1. Does selecting one or the other result in different behavior?
>    a. If not, I am wondering if GFXR should only display 1 of the duplicate GPUs to avoid user confusion. Or at least mark it as a duplicate, e.g.: ` 2               AMD Radeon RX 7900 XTX          123456 (duplicate)`
> 2. Is it possible to correspond the SybSysId to the physical card?

1. In this case, replay worked the same when specifying `--gpu 0` or `--gpu 2`. The same GPU ended up performing the work.
2. From our testing the `SubSysId` does indicate uniqueness in adapters.  Replaying with `--gpu 1` made the other GPU light up. We get the `0-2` list from GPU enumeration, and the only difference we saw in the adapter `desc` was a different adapter LUIDs. The rest was the same.

davidd-lunarg (approved):

PR #1837, opened 2024-10-25
dump resources: use scaled extent when creating a scaled image

davidlunarg:


bradgrantham-lunarg (approved):

framework/graphics/vulkan_resources_util.cpp
```diff
@@ -2195,7 +2195,7 @@ VkResult VulkanResourcesUtil::BlitImage(VkImage               image,
     create_info.flags                 = 0;
     create_info.imageType             = type;
     create_info.format                = dst_format;
-    create_info.extent                = extent;
+    create_info.extent                = scaled_extent;
```
panos-lunarg, line 2198: Right. We need `scaled_extent` here only when `scale` is greater than `1.0f`. If `scale` is less than 1 we need the original non scaled dimensions. If we create the result image with the scaled down dimensions it can have less mipmap levels than the original one. In this case we will dump less levels (i.e. less images) than what is expected.

For example we want to dump an image 4x4 with 3 mip levels (4x4, 2x2 and 1x1) with a scale factor of 0.5. Result image will be 2x2 so that means 2 mip levels. So we create the result image with 4x4 and blit using the scaled down dimensions. This will generate 2 images of 1x1 but 3 images in total, just as many as the original levels.

So this line should be: `create_info.extent = (scale > 1.0f) ? scaled_extend : extend`

bradgrantham-lunarg (approved):

panos-lunarg (approved):

PR #1644, opened 2024-07-30
Asset file for trimmed captures

panos-lunarg:

PR #1838, opened 2024-10-27
CombineAndWriteToFile() now takes a file output as an argument

panos-lunarg:


panos-lunarg:
This should address an issue raised during code review of https://github.com/LunarG/gfxreconstruct/pull/1644

PR #1839, opened 2024-10-27
Add test app framework and some initial tests

beau-lunarg:


charles-lunarg (commented):
Overall looks to be what I expect.

As the vk-bootstrap author, I want to suggest so many minor things because they were written for the 'library' which can't assume how the code will be used - whereas here we can make that assumption and remove it. But for the time being that is just fine, it won't hurt us to have.

This looks to be a fine base to create bespoke test cases with. None of my comments are fixes, just comments about the code.

test/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+###############################################################################
+# Copyright (c) 2018-2020 LunarG, Inc.
```
ziga-lunarg, line 2: Update copyright year

test/test_apps/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+###############################################################################
+# Copyright (c) 2018-2020 LunarG, Inc.
```
ziga-lunarg, line 2: Update copyright year

test/test_apps/triangle/triangle.cpp
```diff
@@ -201,61 +214,61 @@ void create_graphics_pipeline(gfxrecon::test::Init& init, RenderData& data) {
...
-        auto result = init.disp.createFramebuffer(&framebuffer_info, nullptr, &data.framebuffers[i]);
+        auto result = init.disp.createFramebuffer(&framebuffer_info, nullptr, &this->framebuffers[i]);
         VERIFY_VK_RESULT("failed to create framebuffer", result);
     }
 }
 
-void create_command_buffers(gfxrecon::test::Init& init, RenderData& data) {
-    data.command_buffers.resize(data.framebuffers.size());
+void Triangle::create_command_buffers() {
+    this->command_buffers.resize(this->framebuffers.size());
```
MarkY-LunarG: Do you really need to use `this->` everywhere here?

test/test_apps/common/test_app_base.cpp
```diff
@@ -2351,22 +2351,27 @@ std::exception sdl_exception() {
...
-    gfxrecon::test::InstanceBuilder instance_builder;
-    init.instance = instance_builder.use_default_debug_messenger().request_validation_layers().build();
+void device_initialization_phase_2(InstanceBuilder const& instance_builder, Init& init)
+{
+    init.instance = instance_builder.build();
 
     init.inst_disp = init.instance.make_table();
 
     init.surface = gfxrecon::test::create_surface_sdl(init.instance, init.window);
+}
```
MarkY-LunarG: What's the benefit of splitting this up?

beau-lunarg: Between phase 1 and 2, there is an opportunity for test apps to customize instance creation via the InstanceBuilder.
Between phases 2 and 3, apps can customize physical device selection via the PhysicalDeviceSelector.
Between phase 3 and 4, apps can customize device creation via the DeviceBuilder.
Betwen phase 4 and 5, apps can customize swapchain creation via the SwapchainBuilder.

This is threaded together in TestAppBase::run.  However, I left open the possiblity of test apps not using TestAppBase and left the device_initialization method from VkBootstrap as an alternative.  That method does not allow for the customizations above.  I used these "phase" functions to make sure the two implementations were similar.  In the future I could see making them part of the public interface if people need customization but don't want to use TestAppBase.  Or we could force TestAppBase and collapse the functionality into TestAppBase::run.

external/Vulkan-Utility-Libraries/vk_enum_string_helper.h
```diff
@@ -0,0 +1,9617 @@
+// *** THIS FILE IS GENERATED - DO NOT EDIT ***
```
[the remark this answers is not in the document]
beau-lunarg: SDL3 was originally a submodule.  In the meeting last week it was mentioned that we were moving to FetchContent where possible, so I switched it.

test/test_apps/triangle/triangle.cpp
beau-lunarg: Fixed

test/test_apps/CMakeLists.txt
beau-lunarg: Fixed

test/CMakeLists.txt
beau-lunarg: Fixed

test/test_apps/triangle/app.cpp
```diff
@@ -0,0 +1,498 @@
...
+
+const int MAX_FRAMES_IN_FLIGHT = 2;
+
+class App : public gfxrecon::test::TestAppBase
+{
+  public:
+    App() = default;
+
+  private:
+    VkQueue graphics_queue;
```
antonio-lunarg, line 45: I guess we can follow [these guidelines](https://google.github.io/styleguide/cppguide.html#Variable_Names) adding an underscore at the end to member variable names, so it's more clear when they are used in methods, therefore `this->` could be dropped with no issues.

test/test_apps/common/test_app_base.h
```diff
@@ -0,0 +1,1094 @@
...
+    ~TestAppBase()                             = default;
+    TestAppBase(const TestAppBase&)            = delete;
+    TestAppBase& operator=(const TestAppBase&) = delete;
+    TestAppBase(TestAppBase&&)                 = delete;
+    TestAppBase& operator=(TestAppBase&&)      = delete;
+
+    void recreate_swapchain(bool wait_for_idle);
+
+    virtual void setup();
+    virtual bool frame(const int frame_num) = 0;
```
antonio-lunarg, line 1080: No need for `const` to primitive parameters, in general, as those are copied when functions are called.

test/test_apps/triangle/app.cpp
```diff
@@ -0,0 +1,498 @@
...
+
+#include <SDL3/SDL_main.h>
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+
+GFXRECON_BEGIN_NAMESPACE(test_app)
+
+GFXRECON_BEGIN_NAMESPACE(triangle)
+
+const int MAX_FRAMES_IN_FLIGHT = 2;
```
antonio-lunarg, line 37: nit: I would use `size_t` here as it is used for sizing the `command_pools` array and `current_frame` is `size_t` as well.

test/test_apps/common/test_app_base.h
bradgrantham-lunarg: There's a tiny bit of value here in that the implementation of this function will give an error if it changes frame_num.  So although a prototype is (among other things) a contract with the caller, in this case it's also a bit of a helpful restriction on the implementor.  I dunno if that was the intention.

bradgrantham-lunarg:
@charles-lunarg @MarkY-LunarG @antonio-lunarg @ziga-lunarg - if you believe your comments have been addresses, please click the resolve button or comment on them as something you still would like to see fixed.

test/test_apps/triangle/app.cpp
antonio-lunarg: I would not mind keeping `this->`, to be clear, but I believe sticking to the practice currently used is the way to go.

cmake/FindSDL3.cmake
```diff
@@ -0,0 +1,7 @@
+include(FetchContent)
```
antonio-lunarg, line 1: May I suggest to add a status message here?
```
message(STATUS "Fetching SDL2 files from ${SDL_GIT}")
```

test/test_apps/common/test_app_base.cpp
MarkY-LunarG: Don't change it for me.  Maybe as you add more features in each phase, rename them from "phase_x" to "device_initialization_<descrip>" to be clearer.

cmake/FindSDL3.cmake
antonio-lunarg: Sorry, I put SDL2, but the actual version is defined by the GIT_TAG, so I guess it would be nice to use that in the message :)

test/test_apps/common/test_app_base.h
antonio-lunarg: Just noticed the `bool` return type is used by the `test_app_base` to determine whether to exit.
I believe it would be nice to add this info to the function documentation, e.g.:
```c
/// @return `true` to continue running, `false` to exit
```

test/test_apps/common/test_app_base.h
```diff
@@ -0,0 +1,1094 @@
...
+#include <cstring>
+
+#include <optional>
+#include <exception>
+#include <vector>
+#include <string>
+
+#include <vulkan/vulkan_core.h>
+
+#include <SDL3/SDL.h>
```
antonio-lunarg, line 37: Do we need SDL3 for all test apps?

test/test_apps/common/test_app_base.h
```diff
@@ -0,0 +1,1094 @@
...
+const char* to_string(QueueError err);
+const char* to_string(DeviceError err);
+const char* to_string(SwapchainError err);
+
+std::exception to_exception(InstanceError err);
+std::exception to_exception(PhysicalDeviceError err);
+std::exception to_exception(QueueError err);
+std::exception to_exception(DeviceError err);
+std::exception to_exception(SwapchainError err);
+
```
antonio-lunarg, line 184: All these should return `std::runtime_error`, otherwise [slicing](https://en.wikipedia.org/wiki/Object_slicing) will happen

test/test_apps/common/test_app_base.h
beau-lunarg: There could be test apps that don't require it, but since this is supposed to be a quick foundation for all test apps, there is going to be functionality that is included in the base that isn't needed for all of them.  For window and surface creation, SDL is quick and easy, so TestAppBase uses it to bootstrap.

test/test_apps/common/test_app_base.cpp
```diff
@@ -0,0 +1,2981 @@
...
+        CASE_TO_STRING(SurfaceSupportError, failed_get_surface_capabilities)
+        CASE_TO_STRING(SurfaceSupportError, failed_enumerate_surface_formats)
+        CASE_TO_STRING(SurfaceSupportError, failed_enumerate_present_modes)
+        CASE_TO_STRING(SurfaceSupportError, no_suitable_desired_format)
+        default:
+            return "";
+    }
+}
+
+std::exception to_exception(SurfaceSupportError error)
```
antonio-lunarg, line 2181: Do not forget these two -> `std::runtime_error`

antonio-lunarg (approved):

bradgrantham-lunarg (approved):

test/test_apps/shader-objects/shaders/green.fspv
spencer-lunarg, line 1: super random nit. I think in long run you are better off going `green.frag.spv`

It is what other things do, keeps all your spirv as `.spv` files (for ease of searching and what not) and better to start now before you get too much that changing it requires a lot more work

PR #1841, opened 2024-10-28
Combine common code from vulkan capture manager and vulkan state writer

ziga-lunarg:
This is needed for #1812

### DIFF
--- a/framework/encode/vulkan_capture_common.cpp
+++ b/framework/encode/vulkan_capture_common.cpp
@@ -1,6 +1,5 @@
 /*
- ** Copyright (c) 2019-2020 LunarG, Inc.
- ** Copyright (c) 2023 Advanced Micro Devices, Inc. All rights reserved.
+ ** Copyright (c) 2024 LunarG, Inc.
  **
  ** Permission is hereby granted, free of charge, to any person obtaining a
  ** copy of this software and associated documentation files (the "Software"),
@@ -99,8 +98,6 @@ void CommonWriteCreateHardwareBufferCmd(format::ThreadId                        
         {
             vulkan_state_writer->OutputStreamWrite(plane_info.data(), planes_size);
         }
-
-        vulkan_state_writer->IncrementBlocksWritten();
     }
 #else
     GFXRECON_UNREFERENCED_PARAMETER(thread_id);
--- a/framework/encode/vulkan_capture_common.cpp
+++ b/framework/encode/vulkan_capture_common.cpp
@@ -1,0 +1,263 @@
+/*
+ ** Copyright (c) 2019-2020 LunarG, Inc.
+ ** Copyright (c) 2023 Advanced Micro Devices, Inc. All rights reserved.
+ **
+ ** Permission is hereby granted, free of charge, to any person obtaining a
+ ** copy of this software and associated documentation files (the "Software"),
+ ** to deal in the Software without restriction, including without limitation
+ ** the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ ** and/or sell copies of the Software, and to permit persons to whom the
+ ** Software is furnished to do so, subject to the following conditions:
+ **
+ ** The above copyright notice and this permission notice shall be included in
+ ** all copies or substantial portions of the Software.
+ **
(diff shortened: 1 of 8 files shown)


framework/encode/vulkan_capture_common.cpp
```diff
@@ -0,0 +1,263 @@
...
+    else if (vulkan_state_writer)
+    {
+        vulkan_state_writer->OutputStreamWrite(&create_buffer_cmd, sizeof(create_buffer_cmd));
+
+        if (planes_size > 0)
+        {
+            vulkan_state_writer->OutputStreamWrite(plane_info.data(), planes_size);
+        }
+
+        vulkan_state_writer->IncrementBlocksWritten();
```
panos-lunarg, line 103: I think you can increment the counter from where the state writer calls `CommonProcessHardwareBuffer`.
This way the common code will not have to know that the state writer keeps a counter and `IncrementBlocksWritten` will not be necessary.

framework/encode/vulkan_capture_common.cpp
```diff
@@ -0,0 +1,263 @@
...
+    GFXRECON_UNREFERENCED_PARAMETER(hardware_buffer);
+    GFXRECON_UNREFERENCED_PARAMETER(plane_info);
+    GFXRECON_UNREFERENCED_PARAMETER(vulkan_capture_manager);
+    GFXRECON_UNREFERENCED_PARAMETER(vulkan_state_writer);
+
+    GFXRECON_LOG_ERROR("Skipping create AHardwareBuffer command write for unsupported platform");
+#endif
+}
+
+static void CommonWriteFillMemoryCmd(format::HandleId      memory_id,
```
panos-lunarg, line 117: I think it would be better if instead of defining this function to pass the pointer to Writer's/Manager's `WriteFillMemoryCmd` as an additional argument to `CommonProcessHardwareBuffer`. But this would require writing a declaration for that function pointer somewhere that is common to all managers and is api agnostic since this function is referred to from a lot of places (dx12 capture manager,  vulkan capture manager and state writer, and the page guard manager).

I think this gets outside of the scope of this PR so I'll just leave this here as a suggestion for a different PR.

framework/encode/vulkan_capture_common.cpp
```diff
@@ -0,0 +1,263 @@
+/*
+ ** Copyright (c) 2019-2020 LunarG, Inc.
```
panos-lunarg, line 2: Update year in copyright header

framework/encode/vulkan_capture_common.h
```diff
@@ -0,0 +1,54 @@
+/*
+ ** Copyright (c) 2019-2021 LunarG, Inc.
+ ** Copyright (c) 2023 Advanced Micro Devices, Inc. All rights reserved.
```
panos-lunarg, line 3: Update year in copyright header

panos-lunarg (approved):

framework/encode/vulkan_capture_common.cpp
ziga-lunarg: Good catch

framework/encode/vulkan_capture_common.cpp
```diff
@@ -0,0 +1,260 @@
...
+GFXRECON_BEGIN_NAMESPACE(encode)
+
+void CommonWriteCreateHardwareBufferCmd(format::ThreadId                                    thread_id,
+                                        format::HandleId                                    memory_id,
+                                        AHardwareBuffer*                                    hardware_buffer,
+                                        const std::vector<format::HardwareBufferPlaneInfo>& plane_info,
+                                        VulkanCaptureManager*                               vulkan_capture_manager,
+                                        VulkanStateWriter*                                  vulkan_state_writer)
+{
+#if defined(VK_USE_PLATFORM_ANDROID_KHR)
```
antonio-lunarg, line 39: Just wondering: as this file is mostly android-only code, can it be conditionally included in the source list at cmake configuration-time (when configuring for Android)?

ziga-lunarg: It could be, but I think in the future more function should be added here, including code that is not android specific. If you look at the `vulkan_state_writer.cpp` there are many TODOs saying it is common code and should be combined

bradgrantham-lunarg (approved):

PR #1844, opened 2024-10-29
ShaderGroupHandles: tracking/trimming

fabian-lunarg:
- various changes to state-writer: always provide shader-group-handles
- track bound pipelines in replay, retrieve map of old/new handles in CmdTraceRays
- Add a generic murmur32-implementation to util/hash.h
- Add a 'linear hashmap' type (cpu-flavor), populate with handles
- Add unit-test for linear_hashmap

next in follow-up:
- TODO: linear hashmap gpu-flavor
- TODO: run compute-shader over SBT, replace handles (in-place / shadow-buffer)

this PR belongs to https://github.com/LunarG/gfxreconstruct/issues/1526

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -8159,8 +8159,10 @@ VkResult VulkanReplayConsumerBase::OverrideGetAndroidHardwareBufferPropertiesAND
 
 void VulkanReplayConsumerBase::ClearCommandBufferInfo(VulkanCommandBufferInfo* command_buffer_info)
 {
+    GFXRECON_ASSERT(command_buffer_info != nullptr)
     command_buffer_info->is_frame_boundary = false;
     command_buffer_info->frame_buffer_ids.clear();
+    command_buffer_info->bound_pipeline_id = format::kNullHandleId;
 }
 
 VkResult VulkanReplayConsumerBase::OverrideBeginCommandBuffer(
@@ -8495,7 +8497,7 @@ void VulkanReplayConsumerBase::OverrideCmdTraceRaysKHR(
                 address_remap(in_pCallableShaderBindingTable);
             }
 
-            // TODO: run Repl8cer! (create linear hashmap (x), run compute-shader (), profit ())
+            // TODO: run sbt-handle replacer. (create linear hashmap (x), [create shadow-buffer], run compute-shader)
             util::linear_hashmap<graphics::shader_group_handle_t, graphics::shader_group_handle_t> hashmap;
             for (const auto& [lhs, rhs] : shader_group_handles)
             {
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -7747,6 +7747,7 @@ VkResult VulkanReplayConsumerBase::OverrideCreateRayTracingPipelinesKHR(
                                                          &pPipelines->GetPointer()[createInfoCount]);
     }
 
+    // NOTE: this is basically never true and does not look like it's going to change soon
     if (device_info->property_feature_info.feature_rayTracingPipelineShaderGroupHandleCaptureReplay)
     {
         // Modify pipeline create infos with capture replay flag and data.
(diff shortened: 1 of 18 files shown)


framework/generated/generated_vulkan_replay_consumer.cpp
```diff
@@ -49,25 +49,25 @@ void VulkanReplayConsumer::Process_vkCreateInstance(
     HandlePointerDecoder<VkInstance>*           pInstance)
 {
     if (!pInstance->IsNull()) { pInstance->SetHandleLength(1); }
-    VulkanInstanceInfo handle_info;
+    InstanceInfo handle_info;
```
MarkY-LunarG: I assume you fixed this later since it shouldn't compile until you do.  You should likely squash the fix for this with this change or else the tree won't be bisect-able.

framework/util/hash.h
```diff
@@ -98,6 +98,117 @@ std::size_t hash_range(It first, It last)
     return seed;
 }
 
+//! stripped-out finalizer from murmur3_32
+inline uint32_t murmur3_fmix32(uint32_t h)
```
MarkY-LunarG: Do we need to add copyright notice info in our header if we're pulling source from elsewhere?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -8481,25 +8478,42 @@ void VulkanReplayConsumerBase::OverrideCmdTraceRaysKHR(
...
-             in_pMissShaderBindingTable,
-             in_pHitShaderBindingTable,
-             in_pCallableShaderBindingTable,
-             width,
-             height,
-             depth);
+        else
+        {
+            // tmp: guard execution, this would crash otherwise
+            func(commandBuffer,
```
MarkY-LunarG, line 8510: Isn't this call still needed in the above `if`?

MarkY-LunarG (commented):
A couple of questions.
My big concern is the func call skipped. 
Other than those concerns,  because this will break the bi-sectable nature of the tree, I think when you commit it should be squashed to avoid this.

framework/util/hash.h
fabian-lunarg: I'm not entirely sure, the algorithm itself should be public-domain. 
the code itself was written by me, based on an older implementation I already had and then c++-templatized.
so there was not really a source I could point to, other then the original author's repo
-> https://github.com/aappleby/smhasher/blob/master/src/MurmurHash3.cpp

the license-header says:
_// MurmurHash3 was written by Austin Appleby, and is placed in the public
// domain. The author hereby disclaims copyright to this source code._

should we adopt that?

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: yes, absolutely. we will always want to call this function in the end.

the branching above should just set the stage for it (optionally adapting addresses, fixing binding-tables, etc.).

currently the condition omits to call `vkCmdTraceRays` in case we know it will crash (because the sbt is not valid and we did not yet fix it) .
this was mainly for me to find out how far we get after that point, will remove the `else` clause

framework/generated/generated_vulkan_replay_consumer.cpp
fabian-lunarg: yes, this was fixed by a rebase in latest commit. we normally stash our PR-commits when merging, so that should remove this intermediate state.

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: `vkCmdTraceRays` not skipped anymore

framework/util/hash.h
```diff
@@ -19,6 +19,10 @@
 ** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
 ** FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 ** DEALINGS IN THE SOFTWARE.
+**
+//-----------------------------------------------------------------------------
+// MurmurHash3 was written by Austin Appleby, and is placed in the public
+// domain. The author hereby disclaims copyright to this source code.
```
fabian-lunarg, line 25: is this acceptable to just extend here? @MarkY-LunarG

MarkY-LunarG (approved):
Nothing really jumps out to me.  I'm glad you have the hashmap tests, but I'm not sure how much they'll help.  But I'd rather have more tests than not enough.  So we can always strip them later if we feel they're not of much help.

framework/util/hash.h
fabian-lunarg: thanks @MarkY-LunarG!

PR #1850, opened 2024-10-30
Add VulkanDirectAllocator

### LINKED ISSUE #1824: Can VulkanResourceAllocator “Direct” methods be simplified?
Can VulkanResourceAllocator “Direct” methods be simplified to just calling vkAllocateMemory at the call sites that currently call the Direct methods?  If not, can they be factored out into the base allocator and simplify our implementations?  If that doesn't makes sense, that's okay and add some comments to that effect.

antonio-lunarg:
Here's my attempt at simplifying the `VulkanResourceAllocator` interface by extracting `*Direct` functions into a separate class `VulkanDirectAllocator`.

This is one option: `VulkanResourceAllocator` subclasses, just need to implement `GetDirectAllocator()`.

Currently, `VulkanDirectAllocator` is just a copy of `VulkanDefaultAllocator`. If we want to keep it decoupled by the other allocators, it can evolve separately.

Another option might be having `VulkanDirectAllocator` accept a pointer to the concrete allocator, and make all direct functions forward their arguments to the concrete allocator. E.g.:

```c++
class VulkanDirectAllocator {
    VulkanResourceAllocator& alloc_;

    VkResult CreateBuffer(const VkBufferCreateInfo*    create_info,
                          const VkAllocationCallbacks* allocation_callbacks,
                          VkBuffer*                    buffer,
                          ResourceData*                allocator_data)
    {
        return alloc_.CrateBuffer(create_info, allocation_callbacks, format::kNullHandleId, buffer, allocator_data);
    }
}
```

Closes #1824

### DIFF
--- a/framework/decode/vulkan_resource_allocator.h
+++ b/framework/decode/vulkan_resource_allocator.h
@@ -235,9 +235,13 @@ class VulkanResourceAllocator
                                                        const ResourceData*                    allocator_resource_datas,
                                                        const MemoryData* allocator_memory_datas) = 0;
 
-    // Direct allocation methods that perform memory allocation and resource creation without performing memory
-    // translation.  These methods allow the replay tool to allocate staging resources through the resource allocator so
-    // that the allocator is aware of all allocations performed at replay.
+    // Direct allocation methods allocate memory and create resources without performing memory translation, while using
+    // the replay memory type of the current allocator implementation.
+    //
+    // Replay allocates staging resources, which were not created during capture, for operations such as screenshot
+    // creation and resource initialization during the initial state setup for trimmed files. Allocating staging
+    // resources through the resource allocator allows the resource allocator to be aware of all memory allocations
+    // performed for replay.
     virtual VkResult CreateBufferDirect(const VkBufferCreateInfo*    create_info,
                                         const VkAllocationCallbacks* allocation_callbacks,
                                         VkBuffer*                    buffer,


framework/decode/screenshot_handler.cpp
```diff
@@ -366,7 +367,7 @@ void ScreenshotHandler::WriteImage(const std::string&                      filen
                 if (result == VK_SUCCESS)
                 {
                     void* data = nullptr;
-                    result     = allocator->MapResourceMemoryDirect(
+                    result     = allocator->GetDirectAllocator().MapResourceMemory(
```
bradgrantham-lunarg, line 370: This line is returning `VK_ERROR_MEMORY_MAP_FAILED` in this PR but returns `VK_SUCCESS` in `dev` branch.  This is why at least vkcube_95_105 test case is failing in Jenkins CI.

antonio-lunarg:
To recap: I got no luck with the initial approach. I noticed I could not use the default allocator as a direct allocator, cause I saw in some cases (`-m rebind`) direct functions were called for resources created using the rebind allocator, and this resulted in a crash.

For this reason, I opted for the "forwarding" approach, where the direct allocator calls functions of the current allocator, and I discovered there's not "perfect" forwarding, so I had to keep a bunch of direct functions in the resource allocator interface.

antonio-lunarg:
I just pushed a new version which adds two new functions to the resource allocator for specifying which device memory properties to use when binding memory. In some ways I think it clarifies a bit the intent, especially for binding staging buffers created during replay (which expects replay device memory properties).

Said that, there not much I can do to move `MapResourceMemoryDirect` and `UnmapResourceMemoryDirect` as each allocator implements those "directly", no forwarding here. So, if we conclude that this is just overcomplicating the memory allocator interface in general, I would go for adding some documentation on the `*Direct` functions in `VulkanResourceAllocator`.

bradgrantham-lunarg (approved):

PR #1851, opened 2024-10-30
Move vulkan codegen generators

MarkY-LunarG:
These changes moves the current Vulkan and base generators into a better layout that can share code in a more understandable way.

Instead of:
```
   base_generators/
     base* files
   dx12_generators/
     dx12*  files
   vulkan_generators/
     vulkan* files
```

After this change:
```
   dx12_generators/
     dx12*  files
   khronos_generators/
      khronos_base_* files     <-  base* files
      vulkan_generators/
           vulkan* files
```

Also removed some duplication in the files and created utility functions to remove some additional Vulkan-specific items in the lower levels.

This is a big step towards building the appropriate structure for sharing files with OpenXR.


framework/generated/khronos_generators/khronos_base_generator.py
```diff
@@ -1075,4 +1075,20 @@ def getBaseInputStructureName(self):
 
     def getBaseOutputStructureName(self):
         """Intended to be overridden."""
-        return ''
\ No newline at end of file
+        return ''
+
+    def getStructTypeVarName(self):
+        """Intended to be overridden."""
```
jzulauf-lunarg, line 1101: Same comment as above to get  some sort of assert on not overriding.

framework/generated/khronos_generators/base_generators/base_struct_handle_mappers_body_generator.py
```diff
@@ -29,60 +29,33 @@ class BaseStructHandleMappersBodyGenerator():
...
-            )
-            write('', file=self.outFile)
-            write('        switch (base->sType)', file=self.outFile)
-            write('        {', file=self.outFile)
-            write('        default:', file=self.outFile)
-            write(
-                '            // TODO: Report or raise fatal error for unrecongized sType?',
-                file=self.outFile
-            )
-        else:
```
jzulauf-lunarg: Easiest to review this section with whitespace disabled, even though it's python, the removal of the `if else` for API type you can see what *hasn't* changed.

jzulauf-lunarg (approved):
These look good. Probably can squash the "Rename Base..."

framework/generated/khronos_generators/base_generators/base_decoder_body_generator.py
```diff
@@ -120,7 +120,6 @@ def make_cmd_body(self, return_type, name, values):
             else:
                 arg_names.append(value.name)
 
-        # Vulkan return is very simple.
```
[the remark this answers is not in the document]
MarkY-LunarG: It was here originally because of the shared DX12 code.  That's why it was removed.

framework/generated/khronos_generators/khronos_base_generator.py
```diff
@@ -1068,3 +1068,11 @@ def make_value_info(self, params):
                     break
 
         return values
+
+    def getBaseInputStructureName(self):
+        """Intended to be overridden."""
```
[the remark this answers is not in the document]
MarkY-LunarG: Good idea.  I'll do this to all of these.

framework/generated/khronos_generators/khronos_base_generator.py
```diff
@@ -858,6 +859,11 @@ def genType(self, typeinfo, name, alias):
                 self.genUnion(typeinfo, name, alias)
         elif (category == 'handle'):
             self.handle_names.add(name)
+            if (
+                type_elem is not None and type_elem.find('type') is not None
+                and '_DEFINE_HANDLE' == type_elem.find('type').text[2:]
```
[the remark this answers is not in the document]
MarkY-LunarG: It didn't come anywhere.  We used to hard-code the dispatchable handles in Vulkan.  This code automatically determines what handles are dispatchable based on how they are defined in the registry file.  That's why it's a separate commit.  The title of the commit states this as:

```
codgen: Automatically determine dispatchable handles
```

and only adds  this detection logic.

MarkY-LunarG:
@locke-lunarg  or @davidd-lunarg , can one of you review  this because it's codegen?

locke-lunarg (approved):

framework/generated/khronos_generators/khronos_base_decoder_body_generator.py
```diff
@@ -213,7 +192,7 @@ def make_decode_invocation(self, value, preamble, main_body, epilogue):
                     is_class and value.pointer_count > 1
                 ):
                     if type_name in self.base_header_structs.keys():
-                        base_type_name = self.gen_child_var_name(value.base_type)
+                        base_type_name = self.makeSimpleVarName(value.base_type)
```
locke-lunarg, line 195: I wonder what we should unify the name rule to pothole_case(make_simple_var_name)? or is it better to use dromedaryCase?

locke-lunarg:
I think we should rename `base_generator.py` to `vulkan_base_generator.py` since only vulkan uses it, and `BaseGenerator` inherites from `KhronosBaseGenerator`. `BaseGenerator` name confused me to make me to feel that  `KhronosBaseGenerator` inherited from `BaseGenerator`.

framework/generated/khronos_generators/khronos_base_decoder_body_generator.py
jzulauf-lunarg: @bradgrantham @MarkY-LunarG -- is there a standard all new code (including new functions for this refactor) should adhere to.  Right now it's a mix.

bradgrantham-lunarg: I don't have a preference but we state PEP-0008 in CONTRIBUTING.md.  PEP-8 appears to prefer pothole/snake case for functions and variables.  Am I reading it right?  https://peps.python.org/pep-0008/#function-and-variable-names

MarkY-LunarG: I was following the Khronos standard of naming which uses lower-camel-case for function names and lower-snake-case for variables.  I honestly missed the whole PEP-0008 statement.  I'm ok with reverting it back.

MarkY-LunarG: For example, `beginFile`, `endFile`, etc.

jzulauf-lunarg: There are some functions with a specified naming from from the reg.py calls.  We can PEP-0008 everything else in the new code.

Having function and variable naming match is somewhat odd until one considers how lambda heavy idiomatic/pythonic coding is.

I'll review that new def's match the standard.

jzulauf-lunarg:
Comments resolved on new branch -- https://github.com/LunarG/gfxreconstruct/pull/1898

Issue #1761, opened 2024-09-25
Vulkan dump resources: don't exit() process


davidd-lunarg:
`VulkanReplayDumpResourcesBase::QueueSubmit` calls exit in some failure cases. Try to find a way to avoid the exit call and continue (e.g., return/log an error, throw an exception, etc).

bradgrantham-lunarg:
Fixed in #1852

PR #1852, opened 2024-10-31
dump resources: remove calls to exit(1)

davidlunarg:


framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -1879,7 +1879,6 @@ VkResult VulkanReplayDumpResourcesBase::QueueSubmit(const std::vector<VkSubmitIn
                     GFXRECON_LOG_ERROR("Dumping draw calls failed (%s). Terminating.",
                                        util::ToString<VkResult>(res).c_str())
                     Release();
-                    exit(1);
```
bradgrantham-lunarg, line 1882: Removing these means that the message is no longer true.  But the message probably means that execution can't continue.  Doesn't this need to terminate somehow, either by `abort` or `throw` or some such?

bradgrantham-lunarg: GFXRECON_LOG_FATAL?

bradgrantham-lunarg:
Fixes #1761

framework/decode/vulkan_replay_dump_resources.cpp
antonio-lunarg: Looking at `framework/util/logging.h`, it seems there is no real difference between `GFXRECON_LOG_ERROR` and `GFXRECON_LOG_FATAL`. Do we want `FATAL` to terminate every time?

bradgrantham-lunarg (approved):

bradgrantham-lunarg (approved):

roberto-c (approved):

PR #1855, opened 2024-11-01
screenshot: Print message on memory map failure

### LINKED ISSUE #1853: Make sure all screenshot paths that can fail print a message
Discovered a path in `ScreenshotHandler::WriteImage` that failed silently if `MapMemory` failed.  Every path that may fail to write a screenshot should cause a GFXRECON_LOG_FATAL or at least ERROR because failure to write a screenshot would be unexpected and probably break the next step in a pipeline of commands, e.g. CI.

antonio-lunarg:
Report the memory map failure by logging the error, as the screenshot is not generated.

Closes #1853


bradgrantham-lunarg:
Looks good.  1853 says "all screenshot paths that can fail"; are there other paths through WriteImage that could cause an image to not be written that need a `LOG_ERROR`?

bradgrantham-lunarg (approved):

PR #1860, opened 2024-11-01
Marky update loop event handling

MarkY-LunarG:


framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -826,15 +848,15 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
...
 
     // We received events that haven't been handled yet already, so see if this one is in the list already
-    for (size_t ii = 0; ii < received_events_.size(); ++ii)
+    for (size_t ii = 0; ii < previously_received_unhandled_events_.size(); ++ii)
     {
-        if (received_events_[ii].type == capture_event->type)
+        if (previously_received_unhandled_events_[ii].type == capture_event->type)
         {
-            GFXRECON_LOG_WARNING("Previously received event %s (0x%x, %u)",
+            GFXRECON_LOG_WARNING("Previously received unhandled event %s (0x%x, %u)",
```
jzulauf-lunarg, line 855: INFO  "Using previously received event of type ..." ?

MarkY-LunarG:
Accidentally included a submodule update in last run.  Fixed.

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -843,17 +865,40 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
         eventData->IsNull() ? nullptr : eventData->AllocateOutputData(1, { XR_TYPE_EVENT_DATA_BUFFER, nullptr });
     InitializeOutputStructNext(eventData);
 
-    XrResult replay_result;
+    // If we have not previously started tracking the event behavior for this event type, start tracking it.
+    if (event_behavior_tracking_.find(capture_event->type) == event_behavior_tracking_.end())
+    {
+        event_behavior_tracking_[capture_event->type] = EventBehaviorTracking();
```
jzulauf-lunarg, line 871: `std::unordered_map::operator[]` automatically inserts a default initialized entry, when key is not found.  So this block is unneeded.

What would be useful instead is a reference `EventBehaviorTracking &behavior_tracking = event_behavior_tracking_[capture_event->type]` given that '[]' is best O(1) is worst case O(n).  and that same dereference is done many times below.

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -843,17 +865,40 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
...
+        int64_t  sleep_time  = 1;
+
+        if (event_behavior_tracking_[capture_event->type].allow_timeout_reduction)
+        {
+            // Adjust the time we wait so that we don't wait as long by dropping the retry count
+            // by the number of times we've missed the event previously.
+            uint32_t current_reduction = event_behavior_tracking_[capture_event->type].current_timeout_reduction;
+            if (current_reduction >= retry_limit)
+            {
+                retry_limit = 0;
```
jzulauf-lunarg, line 890: Needs to be one, or we won't even look for the event once.  We're not going to drop anything, as we save anything `other` that we get in `previously_received_unhandled_events_`  (unless we think there's going to be a storm of missing events)

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -863,55 +908,62 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
             {
                 if (replay_result == XR_SUCCESS)
                 {
-                    // Add it to the list of received events
-                    received_events_.push_back(*out_eventData);
+                    // If this was a valid event, but not the one we're interested in, record
+                    // it to a list of received events for us to use later.
+                    previously_received_unhandled_events_.push_back(*out_eventData);
                     GFXRECON_LOG_WARNING("Recording event for later %s (0x%x, %u)",
```
jzulauf-lunarg, line 914: INFO

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -863,55 +908,62 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
...
                                          capture_event->type,
                                          capture_event->type);
 
                     // If we grow too lare on the event vector, it's probably because we have
-                    // received a bunch of events we can not handle.  So remove the first 100
+                    // received a bunch of events we can not handle.  So remove the first 300
                     // events to make room for more without bloating the list size.
                     // TODO: Perhaps do this more elegantly?
-                    if (received_events_.size() > 1000)
+                    if (previously_received_unhandled_events_.size() > 1000)
```
jzulauf-lunarg, line 923: 1000 and 300 should be const statics of the ReplayConsumer

jzulauf-lunarg: and maybe like 5000, and 500?

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -863,55 +908,62 @@ void OpenXrReplayConsumerBase::Process_xrPollEvent(const ApiCallInfo&
...
-                    }
+                    // Next time, sleep for double what we initially slept for.  This way if we're just
+                    // spinning, we spin less and less each time.
+                    sleep_time *= 2;
                 }
                 else
                 {
+                    // An error of some kind occurred
                     GFXRECON_LOG_ERROR("xrPollEvent encountered an error of type 0x%x", replay_result);
                     break;
```
jzulauf-lunarg, line 946: Do we want to break, or just keep retrying?

framework/decode/openxr_replay_consumer_base.h
```diff
@@ -484,7 +484,7 @@ class OpenXrReplayConsumerBase : public OpenXrConsumer
 
     VulkanReplayConsumerBase* vulkan_replay_consumer_ = nullptr;
 
-    std::vector<XrEventDataBuffer> received_events_;
+    std::deque<XrEventDataBuffer> previously_received_unhandled_events_;
```
jzulauf-lunarg, line 487: `XrEvenDataBuffer` are large, and maybe shuffled.  Do we want to store a unique_ptr to them instead (or use std::list to ensure no shuffling occurs)?

jzulauf-lunarg (commented):
This is much nicer, a couple nits, a functional question, and a storage question or two.

framework/decode/openxr_replay_consumer_base.h
```diff
@@ -526,6 +526,28 @@ class OpenXrReplayConsumerBase : public OpenXrConsumer
     {
         return GetReplayData<SwapchainData>(swapchain_id);
     }
+
+    // Track event-specific behavior in a structure (which will be stored in a map).
+    struct EventBehaviorTracking
+    {
+        // Do we allow the event to waited on less and less each time we fail to
+        // properly receive the event?
+        bool allow_timeout_reduction{ true };
```
jzulauf-lunarg, line 535: for better packing int, bool, bool

jzulauf-lunarg (commented):
One more little nit.

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -734,8 +736,8 @@ void OpenXrReplayConsumerBase::UpdateState_xrCreateSession(
     HandlePointerDecoder<XrSession>*                   session,
     XrResult                                           replay_result)
 {
-    Decoded_XrSessionCreateInfo* decoded_info   = createInfo->GetMetaStructPointer();
-    SessionData&                 session_data   = AddSessionData(*session->GetPointer());
+    Decoded_XrSessionCreateInfo* decoded_info = createInfo->GetMetaStructPointer();
+    SessionData&                 session_data = AddSessionData(*session->GetPointer());
```
jzulauf-lunarg, line 740: Intentional white space change?

framework/decode/openxr_replay_consumer_base.cpp
MarkY-LunarG: I think for now we should just break on any error.  I think the logic to handle error scenarios other than the expected errors (unavailable, success) result in something else entirely than simple logic we have (because the runtime may have exited) and I don't want to start running down that rabbit trail right now.

framework/decode/openxr_replay_consumer_base.cpp
MarkY-LunarG: I didn't know that about the initialize.  Thanks.

framework/decode/openxr_replay_consumer_base.h
MarkY-LunarG: Using a list now based on your suggestion.

framework/decode/openxr_replay_consumer_base.cpp
MarkY-LunarG: I switched to constants.  And I set them to 2000 and 500.  I don't like it to get too big because they are large structures.

framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -801,7 +823,7 @@ static EventStrings events_to_string[] = {
 const char* GetEventTypeString(XrStructureType type)
 {
     size_t len = sizeof(events_to_string) / sizeof(EventStrings);
-    for (size_t ii = 0; ii < len; ++ii)
+    for (size_t ii = 1; ii < len; ++ii)
```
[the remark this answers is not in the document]
jzulauf-lunarg: Sorry, I think I fouled this up.  For "Get.*String" we do need entry 0.  I thought this was for the `IsEventHandled` which seems to have disappeared since the last time I looked at this more carefully.

jzulauf-lunarg (approved):
LGTM

PR #1861, opened 2024-11-01
Add missing swapchain image destroy

ziga-lunarg:


bradgrantham-lunarg (approved):

framework/encode/vulkan_state_tracker.cpp
```diff
@@ -1935,6 +1935,7 @@ void VulkanStateTracker::DestroyState(vulkan_wrappers::SwapchainKHRWrapper* wrap
     std::unique_lock<std::mutex> lock(state_table_mutex_);
     for (auto entry : wrapper->child_images)
     {
+        DestroyState(entry);
```
mikes-lunarg, line 1938: Does this need similar logic to https://github.com/LunarG/gfxreconstruct/blob/dev/framework/encode/vulkan_handle_wrapper_util.h#L650-L658 to handle child images that belong to multiple swapchains?

ziga-lunarg: Good catch

mikes-lunarg: Maybe not, the most recent push is now crashing on replay

mikes-lunarg (approved):
This change fixes a very real regression so I think it is good to merge. I want to understand why we can't treat the swapchain image lifetimes consistently between the state table and the handle table so I am going to keep looking at this, but it shouldn't hold up this PR.

panos-lunarg (approved):

PR #1862, opened 2024-11-01
Dump root parameters, insted of descriptor heap

locke-lunarg:
Ex: 
new
```
    "graphics_root_parameters": [
      {
        "root_parameter_index": 2,
        "root_signature_type": "D3D12_ROOT_PARAMETER_TYPE_DESCRIPTOR_TABLE",
        "cmd_bind_type": "D3D12_ROOT_PARAMETER_TYPE_DESCRIPTOR_TABLE",
        "tables": [
          {
            "range_type": "D3D12_DESCRIPTOR_RANGE_TYPE_CBV",
            "num_descriptors": 14,
            "descs": [
              {
                "buffer_location": 13749356032,
                "heap_id": 65,
                "heap_index": 719156,
                "res_id": 6230,
                "dimension": "D3D12_RESOURCE_DIMENSION_BUFFER",
                "subs": [
                  {
                    "index": 0,
                    "offset": 165376,
                    "size": 256,
                    "before_file": "shadowtombraider-no-dxr_dr_cbv_res_id_6230_sub_0_before.bin",
                    "after_file": "shadowtombraider-no-dxr_dr_cbv_res_id_6230_sub_0_after.bin"
                  }
```
old
```
    "descriptor_heap": {
      "heap_id_80": {
        "constant_buffer_views": [
          {
            "res_id": 6368,
            "sub": [
              {
                "index": 0,
                "offset": 508416,
                "size": 256,
                "before_file": "shadowtombraider-no-dxr_draw_3803_0_0_0_dr_cbv_res_id_6368_sub_0_before.bin",
                "after_file": "shadowtombraider-no-dxr_draw_3803_0_0_0_dr_cbv_res_id_6368_sub_0_after.bin"
```

1. Dump `SetXXXRootDescriptorTable`, `SetXXXRootConstantBufferView`, `SetXXXRootShaderResourceView` and `SetXXXRootUnorderedAccessView` from `ID3D12GraphicsCommandList` , instead of descriptor heaps. Root parameters are more straightforward since they are used by shaders.

2. Find RootParameters info in `dx12_browse_consumer.h` and `Dx12DumpResources::CreateRootSignature`. 

3.  Dump some infos about root parameters.  Although `root_signature_type` and `cmd_bind_type` are the same thing, but they come from different places. I saw some titles that they were different, so I chose to write both.  That's a mistake and should trigger validation log.  Added a function `Dx12DumpResourcesDelegate::WriteNote` to write some help info into .json, like error log.

4. Added two flags in `dx12_dump_resources.cpp`: `TEST_WRITE_NOT_FOUND_VIEWS` and `TEST_WRITE_NULL_RESOURCE_VIEWS`.

The `NumDescriptors` in root parameters could be large, but the index might not be found in `cbv_srv_uav_infos` of `D3D12DescriptorHeapInfo`. Set `TEST_WRITE_NOT_FOUND_VIEWS`: true to write the heap info even if the index isn't found. Set false to skip it.

`BufferLocation` in `CreateConstantBufferView` or `pResource` in `CreateShaderResourceView` or `CreateUnorderedAccessView` could be 0. Set `TEST_WRITE_NULL_RESOURCE_VIEWS`: true to write the heap info even if the value is 0. Set false to skip it.

5. `CreateConstantBufferView`, `CreateShaderResourceView` and `CreateUnorderedAccessView` could set the same `DestDescriptor`. The newer view setting could override the older to get the `DestDescriptor`. In this case, changed these struct infos into one struct info.

6. If the target draw call is dispatch, dump compute root parameters. If it's not, dump graphics. But if it's `ExecuteIndirect`, dump both graphics and compute root parameters, since we don't know what it's inside.

7. `CaptureGPUAddrMatchDescriptorHeap` find the heap index for the address of root descriptor tables. Use the index to find the view in the map in the heap.

Not using `MatchDescriptorCPUGPUHandle` and "for loop" to find the view any longer. `MatchDescriptorCPUGPUHandle` has a bug to find a wrong view.

`RelayCPUAddrMatchDescriptorHeap` find the heap index for rtvs and dsv.

8. Skip dump vertices, index, rtv and dsv if the target draw call is dispatch.

### DIFF
--- a/framework/decode/dx12_dump_resources.cpp
+++ b/framework/decode/dx12_dump_resources.cpp
@@ -24,6 +24,7 @@
 #include "decode/dx12_dump_resources.h"
 // TODO: It should change the file name of "vulkan"
 #include "generated/generated_vulkan_struct_to_json.h"
+#include "generated/generated_dx12_enum_to_string.h"
 #include "decode/dx12_object_mapping_util.h"
 #include "util/platform.h"
 #include "util/logging.h"
@@ -45,6 +46,12 @@ GFXRECON_BEGIN_NAMESPACE(decode)
 // According to the resource's desc.Dimension, float is for buffer, image is for the others.
 constexpr bool TEST_READABLE   = false;
 constexpr bool TEST_SHADER_RES = true;
+
+// root parameter's descriptor range type's index mightn't have a view in heap.
+constexpr bool TEST_WRITE_NOT_FOUND_VIEWS = true;
+
+// resource id or BufferLocation could be 0 in view of heap.
+constexpr bool TEST_WRITE_NULL_RESOURCE_VIEWS = true;
 
 static const char* Dx12ResourceTypeToString(Dx12DumpResourceType type)
 {
@@ -70,6 +77,10 @@ static const char* Dx12ResourceTypeToString(Dx12DumpResourceType type)
             return "eiArgs";
         case Dx12DumpResourceType::kExecuteIndirectCount:
             return "eiCount";
+        case Dx12DumpResourceType::kGraphicsRootParameters:
+            return "graphics_root_parameters";
+        case Dx12DumpResourceType::kComputeRootParameters:
+            return "compute_root_parameters";
         case Dx12DumpResourceType::kUnknown:
         default:
             return "";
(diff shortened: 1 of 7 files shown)


framework/decode/dx12_dump_resources.cpp
```diff
@@ -540,20 +577,418 @@ void Dx12DumpResources::BeginRenderPass(
...
+bool CaptureGPUAddrMatchDescriptorHeap(const D3D12_GPU_DESCRIPTOR_HANDLE capture_gpu_addr,
+                                       const D3D12DescriptorHeapInfo&    heap_info,
+                                       uint32_t&                         out_heap_index)
 {
-    for (auto gpu_addr : captured_gpu_addrs)
+    auto increment            = (*heap_info.capture_increments)[heap_info.descriptor_type];
+    auto capture_gpu_addr_end = heap_info.capture_gpu_addr_begin + heap_info.descriptor_count * increment;
+
+    bool is_match = true ? (heap_info.capture_gpu_addr_begin <= capture_gpu_addr.ptr &&
+                            capture_gpu_addr.ptr <= capture_gpu_addr_end)
```
davidd-lunarg, line 588: Should this be `capture_gpu_addr.ptr < capture_gpu_addr_end` instead of `capture_gpu_addr.ptr <= capture_gpu_addr_end`

framework/decode/dx12_dump_resources.cpp
```diff
@@ -540,20 +577,418 @@ void Dx12DumpResources::BeginRenderPass(
...
+
+bool ReplayCPUAddrMatchDescriptorHeap(const D3D12_CPU_DESCRIPTOR_HANDLE replay_cpu_addr,
+                                      const D3D12DescriptorHeapInfo&    heap_info,
+                                      uint32_t&                         out_heap_index)
+{
+    auto increment           = (*heap_info.replay_increments)[heap_info.descriptor_type];
+    auto replay_cpu_addr_end = heap_info.replay_cpu_addr_begin + heap_info.descriptor_count * increment;
+
+    bool is_match = true ? (heap_info.replay_cpu_addr_begin <= replay_cpu_addr.ptr &&
+                            replay_cpu_addr.ptr <= replay_cpu_addr_end)
```
davidd-lunarg, line 612: Should this be `replay_cpu_addr.ptr < replay_cpu_addr_end` instead of `replay_cpu_addr.ptr <= replay_cpu_addr_end`

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -4532,32 +4549,34 @@ void Dx12ReplayConsumerBase::PostCall_ID3D12Device_CreateShaderResourceView(
...
+        srv_info.desc     = *(pDesc->GetMetaStructPointer()->decoded_value);
+        srv_info.is_desc_null = false;
 
         if (pResource != format::kNullHandleId)
         {
             auto desc        = reinterpret_cast<ID3D12Resource*>(GetObjectInfo(pResource)->object)->GetDesc();
             auto mip_count   = desc.MipLevels;
             auto array_count = desc.DepthOrArraySize;
-            switch (info.view.ViewDimension)
+            switch (srv_info.desc.ViewDimension)
```
davidd-lunarg, line 4572: Is `subresource_indices` for all the `decode::DH*Info` classes used only for dump resources? If so, what do you think of moving all the code to find `subresource_indices` to the `Dx12DumpResources` class in order to keep non-dump resources code paths from doing additional work?

For example, could this switch statement be moved to a new static function such as `Dx12DumpResources::GetDescriptorSubresourceIndices(DHShaderResourceViewInfo& srv_info);` and call it here: `Dx12DumpResources::GetDescriptorSubresourceIndices(srv_info.desc);`? And same thing for the other descriptor types that have `subresource_indicies`

framework/decode/dx12_object_info.h
```diff
@@ -286,49 +284,66 @@ struct D3D12DeviceInfo : DxObjectExtraInfo
     bool is_uma{ false };
 };
 
-struct ConstantBufferInfo
+// Constant Buffer View, Shader Resource View and Unordered Access View could overrride each other.
+// So they should be in the one container.
+struct DHConstantBufferViewInfo
```
davidd-lunarg, line 289: 👍

framework/decode/dx12_dump_resources.cpp
locke-lunarg: fixed

framework/decode/dx12_dump_resources.cpp
locke-lunarg: fixed

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -4533,272 +4530,32 @@ void Dx12ReplayConsumerBase::PostCall_ID3D12Device_CreateShaderResourceView(
...
 
-        if (pResource != format::kNullHandleId)
+        if (options_.enable_dump_resources)
         {
-            auto desc        = reinterpret_cast<ID3D12Resource*>(GetObjectInfo(pResource)->object)->GetDesc();
-            auto mip_count   = desc.MipLevels;
-            auto array_count = desc.DepthOrArraySize;
-            switch (info.view.ViewDimension)
+            GFXRECON_ASSERT(dump_resources_);
+            if (srv_info.resource_id == format::kNullHandleId)
```
davidd-lunarg, line 4551: Should this be `if (srv_info.resource_id != format::kNullHandleId)`?

locke-lunarg: Thank you for catch. Fixed.

locke-lunarg:
I added a small change. The given parameter in `GetDescriptorSubresourceIndices` changed `D3D12_RESOURCE_DESC` to `DxObjectInfo`.

locke-lunarg:
Added one commit to move writing info code to functions to make it simpler. It doesn't change output data.

davidd-lunarg (approved):

PR #1863, opened 2024-11-02
Do not keep asset file open

panos-lunarg:
Closing the file can allow external users to keep track of when GFXR is done updating the file and therefore is safe to access it.

### DIFF
--- a/framework/encode/api_capture_manager.h
+++ b/framework/encode/api_capture_manager.h
@@ -43,15 +43,16 @@ class ApiCaptureManager
     static auto AcquireExclusiveApiCallLock() { return std::move(CommonCaptureManager::AcquireExclusiveApiCallLock()); }
 
     // Virtual interface
-    virtual void CreateStateTracker()                                  = 0;
-    virtual void DestroyStateTracker()                                 = 0;
-    virtual void WriteTrackedState(util::FileOutputStream* file_stream,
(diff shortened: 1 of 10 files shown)


framework/encode/api_capture_manager.h
```diff
@@ -47,11 +47,11 @@ class ApiCaptureManager
     virtual void DestroyStateTracker()                                 = 0;
     virtual void WriteTrackedState(util::FileOutputStream* file_stream,
                                    format::ThreadId        thread_id,
-                                   util::FileOutputStream* asset_file_stream,
-                                   const std::string&      asset_file_name) = 0;
+                                   util::FileOutputStream* asset_file_stream = nullptr,
```
antonio-lunarg, line 50: I would advice against default values in virtual functions. An alternative which I like more would be to introduce a new function. E.g.:

```cpp
virtual void WriteTrackedState(util::FileOutputStream* file_stream, format::ThreadId thread_id) = 0;
virtual void WriteTrackedStateWithAsset(util::FileOutputStream* file_stream,
                                   format::ThreadId        thread_id,
                                   util::FileOutputStream* asset_file_stream,
                                   const std::string&      asset_file_name) = 0;
```

panos-lunarg: Actually this is a very good idea. Thanks, Antonio

PR #1871, opened 2024-11-06
tocpp: Cast values to platfrom full types

antonio-lunarg:
Fix invalid conversion errors by casting argument values to their target
platform full types. This fixes compilation issues for code like this:
```cpp
VkWaylandSurfaceCreateInfoKHR waylandSurfaceCreateInfoKHR_17 {
  VkStructureType(1000006000),
  NULL,
  VkWaylandSurfaceCreateFlagsKHR(0),
  0,
  0x2000000,
};
```
which compiled with g++ 14.2.1 20240912 yields this:
```
src/frame_0000_0000.cpp:132:25:
  error: invalid conversion from ‘int’ to ‘wl_surface*’ [-fpermissive]
```


antonio-lunarg:
As an alternative, we could cast the value to `void *` instead of the platform full types.

locke-lunarg (approved):

PR #1872, opened 2024-11-07
Relax VulkanStateWriter's IsMemoryReadable() criteria

panos-lunarg:


panos-lunarg:
This change slashed the WriteState() time by 50% on the s22. Unfortunately it has the complete opposite effect on every other platform I tested it. Leaving this in limbo in case there's a way to predict whether this can benefit or not

PR #1877, opened 2024-11-19
Add sparse image and buffer support

bradgrantham-lunarg:
The current trim handling for buffers and images does not support sparse buffers or sparse images. It only allows one image or buffer binding to a single memory range of a single memory object. As a result, it cannot manage sparse resources, which has led to missing asset issues in trim trace for some titles that utilize sparse resources. This commit introduces support for sparse buffers and sparse images (using opaque memory binding only), effectively resolving the missing asset issue.

Originally by Ming Zheng <Ming.Zheng@amd.com>


framework/encode/vulkan_capture_manager.h
```diff
@@ -816,6 +920,50 @@ class VulkanCaptureManager : public ApiCaptureManager
...
 
+    void PostProcess_vkCreateBuffer(VkResult                     result,
+                                    VkDevice                     device,
+                                    const VkBufferCreateInfo*    pCreateInfo,
+                                    const VkAllocationCallbacks* pAllocator,
+                                    VkBuffer*                    pBuffer)
+    {
+        if (IsCaptureModeTrack() && (result == VK_SUCCESS) && (pCreateInfo != nullptr))
+        {
+            assert(state_tracker_ != nullptr);
```
antonio-lunarg, line 931: `assert` -> `GFXRECON_ASSERT`

framework/encode/vulkan_capture_manager.h
```diff
@@ -816,6 +920,50 @@ class VulkanCaptureManager : public ApiCaptureManager
         }
     }
 
+    void PostProcess_vkCreateBuffer(VkResult                     result,
+                                    VkDevice                     device,
+                                    const VkBufferCreateInfo*    pCreateInfo,
```
antonio-lunarg, line 925: Do we want snake case here?

framework/encode/vulkan_capture_manager.h
```diff
@@ -816,6 +920,50 @@ class VulkanCaptureManager : public ApiCaptureManager
...
+
+    void PostProcess_vkCreateImage(VkResult                     result,
+                                   VkDevice                     device,
+                                   const VkImageCreateInfo*     pCreateInfo,
+                                   const VkAllocationCallbacks* pAllocator,
+                                   VkImage*                     pImage)
+    {
+        if (IsCaptureModeTrack() && (result == VK_SUCCESS) && (pCreateInfo != nullptr))
+        {
+            assert(state_tracker_ != nullptr);
```
antonio-lunarg, line 953: ditto

framework/encode/vulkan_capture_manager.h
```diff
@@ -1380,6 +1528,7 @@ class VulkanCaptureManager : public ApiCaptureManager
     std::unique_ptr<VulkanStateTracker>             state_tracker_;
     HardwareBufferMap                               hardware_buffers_;
     std::mutex                                      deferred_operation_mutex;
+    std::mutex                                      sparse_resource_mutex;
```
antonio-lunarg, line 1531: How about moving the comment in `PostProcess_vkQueueBindSparse` here?

```
// In default mode, the capture manager uses a shared mutex to capture every API function. As a result,
// multiple threads may access the sparse resource maps concurrently. Therefore, we use a dedicated mutex
// for write access to these maps.
```

framework/encode/vulkan_state_writer.cpp
```diff
@@ -1725,31 +1840,112 @@ void VulkanStateWriter::WriteImageMemoryState(const VulkanStateTable& state_tabl
...
-                info.sType                 = VK_STRUCTURE_TYPE_BIND_IMAGE_MEMORY_INFO;
-                info.pNext                 = wrapper->bind_pnext;
-                info.image                 = wrapper->handle;
-                info.memory                = memory_wrapper->handle;
-                info.memoryOffset          = wrapper->bind_offset;
-                EncodeStructArray(&encoder_, &info, 1);
-                encoder_.EncodeEnumValue(VK_SUCCESS);
+                const vulkan_wrappers::DeviceWrapper* device_wrapper = wrapper->bind_device;
+                const VulkanDeviceTable*              device_table   = &device_wrapper->layer_table;
+                assert((device_wrapper != nullptr) && (device_table != nullptr));
```
antonio-lunarg, line 1876: ditto

framework/encode/vulkan_capture_manager.h
nickdriscoll-lunarg: I believe the answer is "yes" as it matches the other `PostProcess_*` functions.

framework/encode/vulkan_state_writer.cpp
```diff
@@ -1361,7 +1361,7 @@ void VulkanStateWriter::ProcessBufferMemory(const vulkan_wrappers::DeviceWrapper
         const uint8_t*                              bytes          = nullptr;
         std::vector<uint8_t>                        data;
 
-        assert((buffer_wrapper != nullptr) && (memory_wrapper != nullptr));
+        assert(buffer_wrapper != nullptr);
```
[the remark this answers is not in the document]
nickdriscoll-lunarg: Done

framework/encode/vulkan_state_writer.cpp
nickdriscoll-lunarg: Done

framework/encode/vulkan_capture_manager.h
nickdriscoll-lunarg: I agree with this. The comment describes a piece of data (the mutex) and not the following code, so it should go with the definition of that data.

locke-lunarg:
https://github.com/LunarG/gfxreconstruct/pull/2201 merged.

PR #1883, opened 2024-11-25
Vulkan dump resources: Option to dump alpha separately

panos-lunarg:
New option --dump-resources-dump-separate-alpha will force the alpha channel of dumped images to be dumped into a separate file

### DIFF
--- a/framework/util/image_writer.cpp
+++ b/framework/util/image_writer.cpp
@@ -83,8 +83,8 @@ const uint32_t kImageBpp    = 4;  // Expecting 4 bytes per pixel for 32-bit BGRA
 const uint16_t kBmpBitCountNoAlpha = 24; // Expecting 24-bit BGR bitmap data.
 const uint32_t kImageBppNoAlpha    = 3;  // Expecting 3 bytes per pixel for 32-bit BGRA bitmap data; alpha removed.
 
-static std::unique_ptr<uint8_t[]> temporary_buffer;
-static size_t                     temporary_buffer_size = 0;
+static size_t               temporary_buffer_size = 0;
+static std::vector<uint8_t> temporary_buffer;
 
 #define CheckFwriteRetVal(_val_, _file_)                                                              \
     {                                                                                                 \
@@ -96,15 +96,6 @@ static size_t                     temporary_buffer_size = 0;
             return false;                                                                             \
         }                                                                                             \
     }
-
-static void ResizeTemporaryBuffer(size_t new_size)
-{
-    if (new_size > temporary_buffer_size)
-    {
-        temporary_buffer_size = new_size;
-        temporary_buffer      = std::make_unique<uint8_t[]>(new_size);
-    }
-}
 
 static const uint8_t* ConvertIntoTemporaryBuffer(uint32_t    width,
                                                  uint32_t    height,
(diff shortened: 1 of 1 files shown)


USAGE_android.md
```diff
@@ -919,6 +919,8 @@ optional arguments:
               Enables dumping of all image sub resources (mip map levels and array layers)
   --dump-resources-dump-raw-images
               When enabled all image resources will be dumped verbatim as raw bin files.
+  --dump-resources-dump-separate-alpha
```
dfriederich, line 922: nit, I don't mind much using this option, but for other gfxr users, I wonder if it should not be the default, even if we do need the ability not do dump it separately.

dfriederich: An option might make more sense if gfxr would emit the alpha for all png's, say also for screenshots if this option is present. And not emit them otherwise. But as I said, I don't mind using the option much, just with its current state to only apply to the dump resources api's, it's not the default I would have picked :-).

framework/decode/vulkan_replay_dump_resources_json.cpp
```diff
@@ -183,14 +187,33 @@ void VulkanReplayDumpResourcesJson::InsertImageInfo(nlohmann::ordered_json& json
...
 
-    if (filename_before != nullptr)
+    if (separate_alpha && vkuFormatHasAlpha(image_format))
     {
-        json_entry["beforeFile"] = *filename_before;
-        json_entry["afterFile"]  = filename;
+        if (filename_before != nullptr)
+        {
+            json_entry["beforeFile"]       = *filename_before;
+            json_entry["beforeFile-alpha"] = util::filepath::InsertFilenamePostfix(*filename_before, "_alpha");
```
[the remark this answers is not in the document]
panos-lunarg: sure

USAGE_android.md
bradgrantham-lunarg: Other clients of the dump resources API don't have a problem with combined alpha.  I think default combined is fine.

framework/util/image_writer.cpp
```diff
@@ -96,6 +97,15 @@ static size_t                     temporary_buffer_size = 0;
         }                                                                                             \
     }
 
+static void ResizeTemoraryBuffer(size_t new_size)
```
bradgrantham-lunarg: Could the temporary be `vector<uint8_t>` and then you can use `resize`?

bradgrantham-lunarg: That is to say, you could call `resize` instead of calling this function.

framework/util/image_writer.cpp
```diff
@@ -331,6 +334,73 @@ static const uint8_t* ConvertIntoTemporaryBuffer(uint32_t    width,
     return reinterpret_cast<const uint8_t*>(temporary_buffer.get());
 }
 
+static uint8_t* ExtractAlphaChannel(uint32_t width, uint32_t height, const void* data, uint32_t data_pitch, bool is_png)
```
bradgrantham-lunarg, line 337: Could this be `expand_to_rgb` and invert the sense of the comparison?  I'd like if possible for parameters to generally express the lowest meaningful thing; if we were to add another luminance-capable image then a future reader would need to guess `is_png` means "can have one-channel encoding".

bradgrantham-lunarg (approved):

framework/util/image_writer.cpp
panos-lunarg: sure

framework/util/image_writer.cpp
panos-lunarg: sure

dfriederich:
Works for me :-).
Thanks for updating the attribute names.

PR #1884, opened 2024-11-25
Fix a bug in pipeline-destruction

fabian-lunarg:
- apply changes suggested by clang-tidy, most important `declaration shadows a local variable`
- this fixes a bug in pipeline-destruction (we did not destroy the pipeline)

note:
in other cases we treat warnings as errors, but apparently tolerate variable-shadowing.
I argue we should be more strict about this. it can create nasty and subtle bugs, as seen here.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -1679,9 +1679,9 @@ void VulkanReplayConsumerBase::GetMatchingDeviceGroup(VulkanInstanceInfo*       
 {
     for (auto handle_id : capture_device_group)
     {
-        auto physical_device_info = GetObjectInfoTable().GetVkPhysicalDeviceInfo(handle_id);
-        GetMatchingDevice(instance_info, physical_device_info);
-        replay_device_group.emplace_back(physical_device_info->handle);
+        auto pd_info = GetObjectInfoTable().GetVkPhysicalDeviceInfo(handle_id);
+        GetMatchingDevice(instance_info, pd_info);
+        replay_device_group.emplace_back(pd_info->handle);
     }
 
     // Modified physical device group to fit requirements(VVL).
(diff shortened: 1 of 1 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -10040,7 +10033,7 @@ void VulkanReplayConsumerBase::OverrideDestroyPipeline(
 
     if (pipeline_info != nullptr)
     {
-        VkPipeline in_pipeline =
+        in_pipeline =
```
fabian-lunarg, line 10036: this is the main issue here. instead of assigning to `in_pipeline`, a local variable with same name was created.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5824,8 +5823,7 @@ VkResult VulkanReplayConsumerBase::OverrideCreateShaderModule(
 
             if (options_.dumping_resources)
             {
-                VulkanShaderModuleInfo* shader_info =
-                    reinterpret_cast<VulkanShaderModuleInfo*>(pShaderModule->GetConsumerData(0));
+                auto* shader_info = reinterpret_cast<VulkanShaderModuleInfo*>(pShaderModule->GetConsumerData(0));
```
antonio-lunarg, line 5826: `auto` should be enough

antonio-lunarg (approved):

framework/decode/vulkan_replay_consumer_base.cpp

antonio-lunarg: even though.. https://releases.llvm.org/18.1.6/tools/clang/tools/extra/docs/clang-tidy/checks/readability/qualified-auto.html

fabian-lunarg: yeah, we currently have a rather fragmented state wrt. that. maybe we should come back to this with a more targeted refactoring-PR. there are other examples reported by clang-tidy, like e.g. redundant usage of virtual/override keywords, empty default-ctors, etc.

antonio-lunarg: I suppose we might prioritize some checks that are more valuable than others in finding potential bugs.

fabian-lunarg: so far that's mostly `-Wshadow` for me. many remarks from clang-tidy are in fact more about "tidiness" or even just cosmetic.

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: the symptom can be seen using `--validate`, we currently leak `VkPipeline` objects, spilling validation-warnings when the capture terminates: ```vkDestroyDevice():  OBJ ERROR : For VkDevice 0x5cf6ebc66d00[], VkPipeline 0x421b27000000016d[] has not been destroyed.``` - this is fixed here

bradgrantham-lunarg (approved):

PR #1889, opened 2024-11-26
dx12 dump: Check duplicated notes

locke-lunarg:
Since of "before" and "after", the infos are written twice. But the infos paths are the same so the second time would override the first time. It doesn't duplicate. But notes are array. The second time won't
override the first time. It causes duplicated data. In this case, it
needs to check if the note is duplicated before writing.


davidlunarg (approved):
Looks good

PR #1895, opened 2024-11-28
New preload

zehao-jing-arm:
Move all file processor and decoder into the preload stage. The data will be saved to a new data structure called 'packet' When replay, the vulkan replay consumer will process those packets directly.

The new preload only works for vulkan API


zehao-jing-arm:
Please have a look at this patch :)

framework/application/application.cpp
```diff
@@ -155,9 +158,25 @@ void Application::Run()
                 auto preload_frames_count = fps_info_->ShouldPreloadFrames(frame_number);
                 if (preload_frames_count > 0U)
                 {
-                    auto* preload_processor = dynamic_cast<decode::PreloadFileProcessor*>(file_processor_);
-                    GFXRECON_ASSERT(preload_processor)
-                    preload_processor->PreloadNextFrames(preload_frames_count);
+                    if (detected_vulkan_)
```
bradgrantham-lunarg, line 161: Detection and processing of API is handled in other parts of the system by having vectors of classes that may or may not take responsibility of blocks.  E.g. there are different Decoder and the FileProcessor walks a list of Decoders to decode and then consume blocks.  This hardcoding is counter to that design.  PreloadAllocator should fit into that design in some way that does not detect or hardcode API types in Application class.

MarkY-LunarG:
Closing this as it is for now.

PR #1897, opened 2024-12-02
Track vkCreateDebugReportCallbackEXT

panos-lunarg:


wpaul-samsung:
@panos-lunarg I finally had some time to check this out.  It actually works nicely and fulfilled the ask.  Unfortunately, I ran into another case where our layer simply does not get called for vkQueueSubmits coming from Android's hwui (skia).  It seems to somehow bypass Vulkan layers.  I see some hwui rendering that calls vkQueueSubmit directly in the driver without going through any layers.  I believe this is the culprit: https://android.googlesource.com/platform/frameworks/base/+/84666a2a3a77/libs/hwui/renderthread/VulkanManager.cpp#65

I can't follow that code too well, but I believe my eyes when I see the driver saying it got a queue submit when our layer did not.  This means that we can rely on our layer to keep track of the submission ids to exactly match those from the driver.  Because of that, we won't need this change.  It does work for many cases, but without it working every time, we can't rely on it.  Sorry for the wasted effort.  :(

panos-lunarg:
>This means that we can rely on our layer to keep track of the submission ids

I guess you mean "we can **not** rely".
In that case I'll close this PR

PR #1902, opened 2024-12-03
Fix `VK_EXT_frame_boundary` fake support

marius-pelegrin-arm:
Fix `VK_EXT_frame_boundary` fake support at device creation at capture and replay time when `VkPhysicalDeviceFrameBoundaryFeaturesEXT` is present by removing it before the actual call to the driver (and adding it again at capture time to not have a visible difference from user's perspective)

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -32,7 +32,7 @@
 #include "decode/vulkan_offscreen_swapchain.h"
 #include "decode/vulkan_address_replacer.h"
 #include "decode/vulkan_enum_util.h"
-#include "decode/vulkan_feature_util.h"
+#include "graphics/vulkan_feature_util.h"
 #include "decode/vulkan_object_cleanup_util.h"
 #include "format/format.h"
 #include "format/format_util.h"
@@ -2702,7 +2702,7 @@ void VulkanReplayConsumerBase::ModifyCreateInstanceInfo(
 
     // Sanity checks depending on extension availability
     std::vector<VkExtensionProperties> available_extensions;
-    if (feature_util::GetInstanceExtensions(instance_extension_proc, &available_extensions) == VK_SUCCESS)
+    if (graphics::feature_util::GetInstanceExtensions(instance_extension_proc, &available_extensions) == VK_SUCCESS)
     {
         // Always enable portability enumeration if available
         modified_create_info.flags &= ~VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR;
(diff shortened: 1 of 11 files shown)


MarkY-LunarG:
I'm confused how we would get a capture scenario where the app uses frame boundary and it's not supported.  Is this a capture of a previous replay that always leaves it on?

per-mathisen-arm:
This extension may be implemented only by a layer and meant to be replayed on another device later that does support it. It may also be that the extension is only meant for tools parsing, as appears the intent behind the Android frame boundary extension, which was originally only implemented by tools and the Android UI apps.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2707,40 +2709,52 @@ void VulkanReplayConsumerBase::ModifyCreateDeviceInfo(
...
             });
             modified_extensions.erase(iter);
+
+            VkBaseOutStructure* current = reinterpret_cast<VkBaseOutStructure*>(&modified_create_info);
+
+            while (current->pNext != nullptr)
+            {
+                if (current->pNext->sType == VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FRAME_BOUNDARY_FEATURES_EXT)
+                {
+                    current->pNext = current->pNext->pNext;
```
MarkY-LunarG, line 2742: I think you should at least put a warning message out indicating that this packet was stripped.

marius-pelegrin-arm: Done

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -674,6 +679,38 @@ VkResult VulkanCaptureManager::OverrideCreateDevice(VkPhysicalDevice
...
+
+        if (frame_boundary_features_parent->pNext == nullptr)
+        {
+            frame_boundary_features_parent = nullptr;
+        }
+        else
+        {
+            frame_boundary_features =
+                reinterpret_cast<VkPhysicalDeviceFrameBoundaryFeaturesEXT*>(frame_boundary_features_parent->pNext);
+            frame_boundary_features_parent->pNext = frame_boundary_features_parent->pNext->pNext;
```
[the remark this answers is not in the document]
marius-pelegrin-arm: I added a warning message.

In theory these structures don't need to be removed at all when the extension is not queried - at least that's what the standard specifies, these structures should be ignored.

In practice drivers are not always standard compliant. This option was implemented because we encountered this issue with some device. The structure being present in other calls was fine, but was problematic at device creation.

lunarpapillo:
The CI failure above seems to be due to a **Pixel 6** issue.  We're working on it and will restart this test run as soon as it's been resolved.

MarkY-LunarG:
I'm trying to understand something. 

If this is capture time, it makes more sense for this to be an external layer.  I don't think we want GFXR supporting something that isn't on the underlying system.  That opens a whole new can of worms in my opinion.

As far as replay goes?
1) Why not just pass in `--remove-unsupported` to replay?  This works in general
2) If you don't disable the extension, and you fake it, GFXR is still going to pass in the structures to the native implementation (which could cause implementation failures and/or crashes):
    * `VkPhysicalDeviceFrameBoundaryFeaturesEXT` to `VkPhysicalDeviceFeatures2`, `VkDeviceCreateInfo`
    * `VkFrameBoundaryEXT` into `VkSubmitInfo`, `VkSubmitInfo2`, `VkPresentInfoKHR`, `VkBindSparseInfo`

@bradgrantham-lunarg, what are your thoughts?

per-mathisen-arm:
I think general use of --remove-unsupported is very harmful as it removes essential sanity checks and often turn immediate and clear error messages into obscure crashes or rendering artifacts instead, and in the worst case leads to getting bad data (eg wrong performance measurements). It should not be relied upon and its usage should not be encouraged. Ideally it should be renamed to --ignore-sanity-checks as that is what it actually does. Instead, it would be better to give users a tool to detect and remove unused features and extensions from traces.

Yes, if we fake extension support, then we need to absorb any calls and any structures that the application makes to this extension, including anything added to any pNext chain. There is already precedent for doing this kind of extension faking between tracing layers and the app for VK_EXT_tooling_info. I think this kind of thing is very useful and a powerful tool.

MarkY-LunarG:
After some investigation and discussion, we've decided this doesn't fit with our desire for GFXR.  We don't want to be simulating extensions inside of the capture layer as that can change the behavior of an application during capture time.

We believe this makes most sense to implement in a downstream layer that simulates the extension leaving GFXR unchanged.  Plus you can verify before enabling GFXR that the behavior you desire still occurs.

MarkY-LunarG:
I was incorrect, the original support was added with #1478.  This change fixes that.  In general in the future we don't want simulation behavior in the capture side, but this is (as Marius mentioned) just fixing existing code.

MarkY-LunarG (approved):

PR #1903, opened 2024-12-03
note that SIGSEGV can also happen on desktop Vulkan

bradgrantham-lunarg:


panos-lunarg:
Pushed a [PR in your remote](https://github.com/bradgrantham-lunarg/gfxreconstruct/pull/3) with the equivalent command for GDB. I have a pinned tab in the browser always open in the [stack overflow answer](https://stackoverflow.com/questions/23362984/gdb-nostop-sigsegv-on-a-specific-thread) for the command so I think this is a good chance to document this as well

panos-lunarg (approved):
lgtm

antonio-lunarg (approved):

PR #1891, opened 2024-11-27
Add option for stricter compiler warnings

fabian-lunarg:
- Add a cmake-option: raise compiler-warnings level for GCC and clang
- apply stricter settings for targets `gfxrecon_format`, `gfxrecon_util`, `gfxrecon_graphics`

first step towards https://github.com/LunarG/gfxreconstruct/issues/1890

PR #1907, opened 2024-12-04
make one frame number instance consistent

bradgrantham-lunarg:
From question on #1891 

Mostly just waiting to see that this compiles correctly on all platforms


bradgrantham-lunarg:
And then go remove the static casts that are added in #1891

PR #1913, opened 2024-12-12
External memory FD

antonio-lunarg:
Initial support for VK_KHR_external_memory_fd.

Capture: once a device memory imported from external FD is bound to a buffer, an InitBuffer command is generated with the current content of the memory.

Replay: the import memory FD structure is removed from the pNext chain of `VkMemoryAllocateInfo`, and the initial content of the buffer is initialized through the InitBuffer command.

It supports images as well!

### DIFF
--- a/framework/decode/vulkan_default_allocator.cpp
+++ b/framework/decode/vulkan_default_allocator.cpp
@@ -661,7 +661,7 @@ void VulkanDefaultAllocator::ReportBindIncompatibility(const VkMemoryRequirement
             auto     memory_alloc_info = reinterpret_cast<MemoryAllocInfo*>(allocator_memory_data);
             uint32_t memory_type_index = memory_alloc_info->memory_type_index;
 
-            if ((requirements[i].memoryTypeBits & (1 << memory_type_index)) != 0)
+            if ((requirements[i].memoryTypeBits & (1 << memory_type_index)) == 0)
             {
                 GFXRECON_LOG_FATAL(
                     "Resource memory bind failed: resource is not compatible with the specified memory type.");
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -4359,7 +4359,8 @@ VkResult VulkanReplayConsumerBase::OverrideAllocateMemory(
         assert(allocator != nullptr);
 
         VulkanResourceAllocator::MemoryData allocator_data;
-        auto                                replay_allocate_info = pAllocateInfo->GetPointer();
+
+        auto *modified_allocate_info = const_cast<VkMemoryAllocateInfo*>(pAllocateInfo->GetPointer());
         auto                                replay_memory        = pMemory->GetHandlePointer();
         auto                                capture_id           = (*pMemory->GetPointer());
 
@@ -4372,7 +4373,11 @@ VkResult VulkanReplayConsumerBase::OverrideAllocateMemory(
         bool                address_override_found = false;
         bool                uses_import_memory     = false;
         uint64_t            opaque_address         = 0;
-        VkBaseOutStructure* current_struct = reinterpret_cast<const VkBaseOutStructure*>(replay_allocate_info)->pNext;
+
+        // FD is not available at replay time
+        graphics::vulkan_struct_remove_pnext<VkImportMemoryFdInfoKHR>(modified_allocate_info);
+
+        VkBaseOutStructure* current_struct = reinterpret_cast<const VkBaseOutStructure*>(modified_allocate_info)->pNext;
 
         size_t                                            host_pointer_size = 0;
         std::unique_ptr<void, std::function<void(void*)>> external_memory_guard(
@@ -4403,9 +4408,9 @@ VkResult VulkanReplayConsumerBase::OverrideAllocateMemory(
             {
                 auto import_info = reinterpret_cast<VkImportMemoryHostPointerInfoEXT*>(current_struct);
 
-                GFXRECON_CHECK_CONVERSION_DATA_LOSS(size_t, replay_allocate_info->allocationSize);
+                GFXRECON_CHECK_CONVERSION_DATA_LOSS(size_t, modified_allocate_info->allocationSize);
 
-                size_t allocation_size = static_cast<size_t>(replay_allocate_info->allocationSize);
+                size_t allocation_size = static_cast<size_t>(modified_allocate_info->allocationSize);
 
                 host_pointer_size =
                     util::platform::GetAlignedSize(allocation_size, util::platform::GetSystemPageSize());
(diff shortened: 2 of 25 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4440,6 +4445,27 @@ VkResult VulkanReplayConsumerBase::OverrideAllocateMemory(
...
                                                &allocator_data);
         }
+        else if (import_fd_found)
+        {
+            VkMemoryAllocateInfo modified_allocate_info = (*replay_allocate_info);
+
+            VkBaseOutStructure* current_struct = reinterpret_cast<VkBaseOutStructure*>(&modified_allocate_info);
+            while (current_struct->pNext)
+            {
+                if (current_struct->pNext->sType == VK_STRUCTURE_TYPE_IMPORT_MEMORY_FD_INFO_KHR)
```
MarkY-LunarG, line 4455: So I did this and I got a crash on binding the memory to either an image or a buffer because the memory formats and indices didn't match.  Check out the top commit on [my fork](https://github.com/MarkY-LunarG/gfxreconstruct/) branch `marky_test_remove_fd`.

framework/graphics/vulkan_struct_get_pnext.h
```diff
@@ -107,6 +107,42 @@ static T* vulkan_struct_get_pnext(Parent_T* parent)
...
+ * Searches through the parent's pNext-chain for the first struct with the requested struct_type.
+ * @p T and @p Parent_T must be Vulkan structure-types. Returns nullptr if no matching struct could be found.
+ *
+ * @tparam  T           the structure-type to remove from pNext-chain
+ * @tparam  Parent_T    implicit type of provided structure
+ * @param   parent      pointer to a non-const vulkan-structure containing a pNext-chain.
+ * @return  a typed pointer to a structure removed from the pNext-chain or nullptr.
+ */
+template <typename T, typename Parent_T>
+static T* vulkan_struct_remove_pnext(Parent_T* parent)
```
fabian-lunarg, line 123: nice :)

framework/decode/vulkan_default_allocator.cpp
```diff
@@ -645,7 +645,7 @@ void VulkanDefaultAllocator::ReportBindIncompatibility(const VkMemoryRequirement
             auto     memory_alloc_info = reinterpret_cast<MemoryAllocInfo*>(allocator_memory_data);
             uint32_t memory_type_index = memory_alloc_info->memory_type_index;
 
-            if ((requirements[i].memoryTypeBits & (1 << memory_type_index)) != 0)
+            if ((requirements[i].memoryTypeBits & (1 << memory_type_index)) == 0)
```
fabian-lunarg, line 648: ok, this looks nasty, was this a bug in `VulkanDefaultAllocator::ReportBindIncompatibility`?
who might have been affected?

antonio-lunarg: Not sure, CI for this PR was failing for this reason though.

fabian-lunarg: ok, odd. looks like you fixed a 5 year old bug, because you were first affected? :)

framework/util/thread_data.h
```diff
@@ -0,0 +1,70 @@
...
+
+  private:
+    static format::ThreadId GetThreadId();
+
+  private:
+    static std::mutex                                     count_lock_;
+    static format::ThreadId                               thread_count_;
+    static std::unordered_map<uint64_t, format::ThreadId> id_map_;
+
+  private:
```
fabian-lunarg, line 62: the repetition of `private:` / `public:` sections is a tad confusing. I get it you're separating by method/member/static etc. but I had to look 3x to spot the `GetThreadId()` method - is this governed by a rule or best-practice?

antonio-lunarg: Yep, we should follow [this rule](https://google.github.io/styleguide/cppguide.html#Declaration_Order).
I messed this up while refactoring :D

fabian-lunarg (approved):
looks good!

framework/encode/output_stream_writer.h
```diff
@@ -0,0 +1,71 @@
...
+#include "util/compressor.h"
+#include "util/thread_data.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(encode)
+
+/**
+ * @brief Common class with utilities to write to an output stream.
+ */
+class OutputStreamWriter
```
bradgrantham-lunarg, line 36: I think this replaces the previous use of CombineAndWrite, is that correct?  This feels like a bigger abstraction than warranted if it just allows not passing in `ThreadData` and allows polymorphism for `OutputStreamWrite`.  It's not much of a burden to pass in `GetThreadData()` as a parameter to `CombineAndWrite`.  Do you anticipate in the future we may have capture managers with different implementations of `OutputStreamWrite`?  Can you tell me more about the reasoning to apply?

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -1800,6 +1825,228 @@ void VulkanCaptureManager::PostProcess_vkEnumeratePhysicalDeviceGroups(
...
+{
+    if (result == VK_SUCCESS)
+    {
+        if (IsCaptureModeTrack())
+        {
+            GFXRECON_ASSERT(state_tracker_ != nullptr);
+            state_tracker_->TrackBufferMemoryBinding(device, buffer, memory, memoryOffset);
+        }
+
+        if (IsCaptureModeWrite())
```
bradgrantham-lunarg, line 1942: Is the logic in this PR applicable to AHB and other external memory backing?  Not as part of this PR but that maybe a useful thing to bring up in our AHB conversations.

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_resource_initializer.cpp
```diff
@@ -77,9 +77,20 @@ VulkanResourceInitializer::~VulkanResourceInitializer()
         resource_allocator_->FreeMemoryDirect(staging_memory_, nullptr, staging_memory_data_);
     }
 
-    device_table_->DestroySampler(device_, draw_sampler_, nullptr);
-    device_table_->DestroyDescriptorPool(device_, draw_pool_, nullptr);
-    device_table_->DestroyDescriptorSetLayout(device_, draw_set_layout_, nullptr);
+    if (draw_sampler_ != VK_NULL_HANDLE)
```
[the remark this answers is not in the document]
antonio-lunarg: Yes, I noticed that sampler, pool, and set_layout where not initialized in the path I introduced for initializing a buffer with external memory, so I added these guards to prevent destruction of null handles.

framework/encode/capture_manager.h
```diff
@@ -52,7 +55,7 @@ GFXRECON_BEGIN_NAMESPACE(encode)
 class ApiCaptureManager;
 
 // The CommonCaptureManager provides common functionality referenced API specific capture managers
-class CommonCaptureManager
+class CommonCaptureManager : public OutputStreamWriter
```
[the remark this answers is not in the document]
fabian-lunarg: yes, that's a valid point

framework/encode/output_stream_writer.h
antonio-lunarg: Thank you for your comments, Brad. This one made me think a bit, and I realized the main reason for this was the capture manager doing something special _before_ and _after_ writing to the output stream, so I made a decorator class for the file stream in `CommonCaptureManager` to make it clear. This approach is still flexible enough to allow using `CommandWriter` from both `CommonCaptureManager` and `VulkanStateWriter`.

bradgrantham-lunarg (approved):

PR #1921, opened 2024-12-16
WIP -- Zulauf vulkan codegen dev with consumer dispatch from marky

jzulauf-lunarg:
for test and @MarkY-LunarG  sanity check


jzulauf-lunarg:
Mooted

PR #1928, opened 2024-12-19
Track vkSetDebugUtilsObject[Name|Tag]EXT

panos-lunarg:


framework/encode/vulkan_state_writer.cpp
```diff
@@ -4338,5 +4340,583 @@ void VulkanStateWriter::WriteExecuteFromFile(const std::string& filename, uint32
     blocks_written_ += n_blocks + 1;
 }
 
+void VulkanStateWriter::WriteDebugUtilsState(const VulkanStateTable& state_table)
+{
+    // state_table.VisitWrappers([&](const vulkan_wrappers::BufferWrapper* wrapper) {})
+    state_table.VisitWrappers([&](const vulkan_wrappers::InstanceWrapper* wrapper) {
```
bradgrantham-lunarg, line 4346: Is every one of these visits the same code?  Can the body be made into a template lambda that is then called in the passed lambda?  E.g. something like
```
    auto write_debug_utils_calls = [&](const auto *wrapper) {
        assert(wrapper != nullptr);
        if (wrapper->debug_name_create_parameters)
        {
            WriteFunctionCall(format::ApiCall_vkSetDebugUtilsObjectNameEXT, wrapper->debug_name_create_parameters.get());
        }

        if (wrapper->debug_tag_create_parameters)
        {
            WriteFunctionCall(format::ApiCall_vkSetDebugUtilsObjectTagEXT, wrapper->debug_tag_create_parameters.get());
        }
    };

    state_table.VisitWrappers([&](const vulkan_wrappers::InstanceWrapper* wrapper) { write_debug_utils_calls(wrapper); });
    state_table.VisitWrappers([&](const vulkan_wrappers::PhysicalDeviceWrapper* wrapper) { write_debug_utils_calls(wrapper); });
    ...
```

bradgrantham-lunarg (changes requested):

framework/encode/vulkan_state_writer.cpp
panos-lunarg: done

framework/encode/vulkan_state_tracker.cpp
```diff
@@ -3231,5 +3231,688 @@ void VulkanStateTracker::TrackBeginRendering(VkCommandBuffer commandBuffer, cons
...
 }
 
+static void AssignDebugToObject(VkObjectType                        object_type,
+                                uint64_t                            object_handle,
+                                vulkan_state_info::CreateParameters object_name_parameter_buffer,
+                                vulkan_state_info::CreateParameters object_tag_parameter_buffer)
+{
+    switch (object_type)
+    {
+        case VK_OBJECT_TYPE_INSTANCE:
```
bradgrantham-lunarg, line 3241: This seems like a lot of identical code except for class names.  Will you tell me what you think about adding this to the generators?

bradgrantham-lunarg (approved):
I approve but I think if `AssignDebugToObject` is something that could be generated in just a day's work or so, I think that's a good idea to do as part of this PR.  If it's going to be very difficult, I'm okay with merging this as it is so those who need it can have it in `dev` but please make an issue to generate this function and let's schedule that in the short term.

framework/encode/vulkan_state_tracker.cpp
panos-lunarg: My initial approach was since the new members (`debug_name_create_parameters` and `debug_tag_create_parameters`) are in the base class to use polymorphism but failed to find a way because the base class is a template. 

I didn't look into autogenerating code for two reasons:
- The enumerates defined in `VkDebugReportObjectTypeEXT` are a superset of the objects we track so the autogen script will still need some logic to ignore those we don't handle.
- We already have hand written code with a switch statement per vulkan handle we handle in: [`framework/encode/vulkan_handle_wrapper_util.cpp`](https://github.com/LunarG/gfxreconstruct/blob/dev/framework/encode/vulkan_handle_wrapper_util.cpp)

bradgrantham-lunarg: It sounds like this should be a later PR.

Just to be clear, you said "The enumerates defined in `VkDebugReportObjectTypeEXT`", but this function takes an `VkObjectType`.  Is that what it's supposed to take?

For the future, it looks like there are cases in this function for every object type except the video handles and the handles in `VK_NVX_binary_import`.  Since the registry specifies the VkObjectType for every handle, this seems like it could be generated.  But it looks like we don't already have generated code for state tracker, like you said.  So maybe vulkan_state_tracker and vulkan_state_writer and vulkan_handle_wrapper are just good candidates to hit with a code generator at some point.

bradgrantham-lunarg: Yeah, it sounds like generating this should be a later PR.

Just to be clear, you said "The enumerates defined in `VkDebugReportObjectTypeEXT`", but this function takes an `VkObjectType`.  Is that what it's supposed to take?

For the future, it looks like there are cases in this function for every object type except the video handles and the handles in `VK_NVX_binary_import`.  Since the registry specifies the VkObjectType for every handle, this seems like it could be generated.  But it looks like we don't already have generated code for state tracker, like you said.  So maybe vulkan_state_tracker and vulkan_state_writer and vulkan_handle_wrapper are just good candidates to hit with a code generator at some point.

bradgrantham-lunarg (approved):

panos-lunarg:
Created https://github.com/LunarG/gfxreconstruct/issues/1933 to track the code generation part

PR #1929, opened 2024-12-19
Add CPU masking/pinning option

marius-pelegrin-arm:
Add an option to specify which CPU cores the replayer is able to use.


MarkY-LunarG:
I rebased and tested this.  I passed in a mask of 00100 and it set it, but reported a mask of 0001.  I'm fixing the "Get" reporting  and then I'll push up a fix.

MarkY-LunarG:
Modified description so that the order used was clearer and modified the example as well.

PR #1930, opened 2024-12-19
test: External memory FD

antonio-lunarg:
Actually two apps. One app acts as exporter, exporting memory to an opaque fd containing data about its color render target, while the other app acts as importer. The exporter waits for the importer to connect via UNIX domain socket and then it sends the opaque fd. The importer connects to the exporter and retrieves the fd to use for importing memory.


bradgrantham-lunarg:
Needs guards in CMakeLists.txt to build only on Linux

bradgrantham-lunarg (approved):
Needs guards in CMakeLists.txt to build only on Linux

Otherwise LGTM!

bradgrantham-lunarg (approved):

PR #1905, opened 2024-12-04
Vulkan dump resources: Drop astc format

panos-lunarg:

PR #1935, opened 2024-12-20
Vulkan dump resources: No separate alpha for images dumped raw

panos-lunarg:


framework/decode/vulkan_replay_dump_resources_json.cpp
```diff
@@ -187,7 +188,9 @@ void VulkanReplayDumpResourcesJson::InsertImageInfo(nlohmann::ordered_json& json
         json_entry["scaleFailed"] = true;
     }
 
-    if (separate_alpha && vkuFormatHasAlpha(image_format))
+    const bool raw_image = !util::filepath::GetFilenameExtension(filename).compare(".bin");
```
wpaul-samsung, line 191: We see the same with .astc files.

dfriederich:
this does fix that we don't get an alpha entry in the json for a bin file :-)
Does happen in a device unit test in sokatoa.

bradgrantham-lunarg (approved):
Approved but want to reach resolution on Warren's comment before merging

panos-lunarg:
> Approved but want to reach resolution on Warren's comment before merging

Warren's comment is basically addressed by https://github.com/LunarG/gfxreconstruct/pull/1905 which completely removes the option to dump images in astc format

PR #1936, opened 2024-12-20
Update Vulkan Headers to v1.4.304

nickdriscoll-lunarg:
This is another header update, but the changes are very minor. Regarding GFXR, it looks like the only effect was renaming one enumerant.


bradgrantham-lunarg:
Yes, looks like a very simple change.  I approve but I'm not in front of
GitHub right now.  Please squash merge at will.

PR #1937, opened 2024-12-22
Fix shader wrapper assert when pipeline binaries are used

ziga-lunarg:
From the spec:

> If a VkPipelineBinaryInfoKHR structure with a binaryCount greater than 0 is included in the pNext chain of any Vk*PipelineCreateInfo structure when creating a pipeline, implementations must use the data in pPipelineBinaries instead of recalculating it. Any shader module identifiers or shader modules declared in VkPipelineShaderStageCreateInfo instances are ignored.

### DIFF
--- a/framework/encode/vulkan_capture_manager.cpp
+++ b/framework/encode/vulkan_capture_manager.cpp
@@ -2779,14 +2779,18 @@ void VulkanCaptureManager::PostProcess_vkCreateGraphicsPipelines(VkResult       
                 vulkan_wrappers::GetWrapper<vulkan_wrappers::PipelineWrapper>(pPipelines[p]);
             assert(ppl_wrapper != nullptr);
 
-            for (uint32_t s = 0; s < pCreateInfos[p].stageCount; ++s)
+            const auto binary_info = graphics::vulkan_struct_get_pnext<VkPipelineBinaryInfoKHR>(&pCreateInfos[p]);
+            if (binary_info == nullptr || !binary_info->binaryCount)
             {
-                const vulkan_wrappers::ShaderModuleWrapper* shader_wrapper =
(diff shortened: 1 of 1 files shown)


bradgrantham-lunarg:
raytracingpipelinesnv doesn't need this condition?

ziga-lunarg:
> raytracingpipelinesnv doesn't need this condition?

Good question, under the `VK_KHR_pipeline_binary` extension the spec says:

> Extending VkGraphicsPipelineCreateInfo, VkComputePipelineCreateInfo, VkRayTracingPipelineCreateInfoKHR:
> - VkPipelineBinaryInfoKHR

So nothing about raytracingNV, but then there are VUIDs for raytracingNV and pipeline binaries. I guess I should add this for raytracingNV to be safe

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -2779,14 +2779,18 @@ void VulkanCaptureManager::PostProcess_vkCreateGraphicsPipelines(VkResult
                 vulkan_wrappers::GetWrapper<vulkan_wrappers::PipelineWrapper>(pPipelines[p]);
             assert(ppl_wrapper != nullptr);
 
-            for (uint32_t s = 0; s < pCreateInfos[p].stageCount; ++s)
+            const auto binary_info = graphics::vulkan_struct_get_pnext<VkPipelineBinaryInfoKHR>(&pCreateInfos[p]);
+            if (!binary_info || binary_info->binaryCount == 0)
```
panos-lunarg, line 2783: nit picking: According to our coding guidelines the checks should be the other way around:
`if (binary_info == nullptr || !binary_info->binaryCount)`
Same applies to the other 2 functions which have the same if condition

panos-lunarg (approved):

bradgrantham-lunarg (approved):
Approve with Panos' suggestion

PR #1942, opened 2025-01-04
Fix trim with external formats

### LINKED ISSUE #1901: Android games that use AHB will crash on trim
Images that are created with external formats are not handled correctly in the vulkan_state_writter.

ziga-lunarg:
Closes #1901 

Replaces #1695

This only fixes the crash, there is a `Todo` left that is waiting on #1812


bradgrantham-lunarg:
I've removed myself from reviewers for this one instance.  If @mikes-lunarg and @antonio-lunarg approve the PR, please merge it.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5113,6 +5121,19 @@ VulkanReplayConsumerBase::OverrideCreateImage(PFN_vkCreateImage
         {
             image_info->queue_family_index = 0;
         }
+
+        auto external_format_android = graphics::vulkan_struct_get_pnext<VkExternalFormatANDROID>(replay_create_info);
+        if (external_format_android && external_format_android->externalFormat != 0)
```
antonio-lunarg, line 5126: Please use `auto *` and explicitly check with `!= nullptr`

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4745,6 +4745,14 @@ VkResult VulkanReplayConsumerBase::OverrideBindImageMemory(PFN_vkBindImageMemory
             image_info->handle, image_info->allocator_data, memory_info->allocator_data);
     }
 
+    if (image_info->external_format)
```
antonio-lunarg, line 4748: Please, add this comment here as well.
```
// Memory requirements for image with external format can only be queried after the memory is bound
```

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5113,6 +5121,19 @@ VulkanReplayConsumerBase::OverrideCreateImage(PFN_vkCreateImage
         {
             image_info->queue_family_index = 0;
         }
+
+        auto external_format_android = graphics::vulkan_struct_get_pnext<VkExternalFormatANDROID>(replay_create_info);
+        if (external_format_android && external_format_android->externalFormat != 0)
+        {
+            image_info->external_format = true;
```
antonio-lunarg, line 5128: Please, add this comment here as well.
```
// Memory requirements for image with external format can only be queried after the memory is bound
```

framework/encode/vulkan_state_tracker.cpp
```diff
@@ -556,6 +556,16 @@ void VulkanStateTracker::TrackImageMemoryBinding(
     {
         wrapper->bind_pnext = vulkan_trackers::TrackStruct(bind_info_pnext, wrapper->bind_pnext_memory);
     }
+
+    // AHB image memory requirements can only be queried after the memory is bound
+    if (wrapper->external_format)
+    {
+        const VulkanDeviceTable* device_table = vulkan_wrappers::GetDeviceTable(device);
+        VkMemoryRequirements     image_mem_reqs;
+        assert(wrapper->handle != VK_NULL_HANDLE);
```
antonio-lunarg, line 565: Prefer `GFXRECON_ASSERT`, but I think you can use `image` directly.

framework/encode/vulkan_state_tracker_initializers.h
```diff
@@ -652,6 +652,12 @@ inline void InitializeState<VkDevice, vulkan_wrappers::ImageWrapper, VkImageCrea
     assert(wrapper->handle != VK_NULL_HANDLE);
     device_table->GetImageMemoryRequirements(parent_handle, wrapper->handle, &image_mem_reqs);
     wrapper->size = image_mem_reqs.size;
+
+    auto external_format_android = graphics::vulkan_struct_get_pnext<VkExternalFormatANDROID>(create_info);
+    if (external_format_android && external_format_android->externalFormat != 0)
```
antonio-lunarg, line 657: ditto

framework/encode/vulkan_state_tracker_initializers.h
```diff
@@ -652,6 +652,12 @@ inline void InitializeState<VkDevice, vulkan_wrappers::ImageWrapper, VkImageCrea
     assert(wrapper->handle != VK_NULL_HANDLE);
     device_table->GetImageMemoryRequirements(parent_handle, wrapper->handle, &image_mem_reqs);
     wrapper->size = image_mem_reqs.size;
```
antonio-lunarg, line 654: What's the result here when the format is external?

framework/encode/vulkan_state_writer.cpp
```diff
@@ -2759,17 +2763,26 @@ void VulkanStateWriter::WriteImageMemoryState(const VulkanStateTable& state_tabl
...
-                                                                                             &snapshot_info.level_sizes,
-                                                                                             true);
+                    if (wrapper->external_format)
+                    {
+                        snapshot_info.resource_size = wrapper->size;
+                        snapshot_info.level_sizes.push_back(wrapper->size);
+                    }
+                    else
+                    {
+                        snapshot_info.resource_size =
```
antonio-lunarg, line 2773: Why can't we use `wrapper->size` here?

ziga-lunarg: Because `snapshot_info.level_sizes` needs to be filled in. AHB cannot have mip levels so it is easier there

framework/encode/vulkan_state_tracker_initializers.h
ziga-lunarg: Good catch, it is not valid to call `vkGetImageMemoryRequirements` before memory is bound when you have external memory.

antonio-lunarg (approved):

PR #1943, opened 2025-01-07
Add condition to shader module assertion post pipeline creation

bartosz-muszarski-arm:
According to the specs VkPipelineShaderStageCreateInfo can specify a null shader module. If module is VK_NULL_HANDLE, the shader code is defined by the chained VkShaderModuleCreateInfo if present.
Currently in this scenario gfxreconstruct will trigger assert on missing shader_wrapper.
With this change the assert will be skipped and the shader wrapper will not be added to bound shaders list.


bartosz-muszarski-arm:
Fixed in https://github.com/LunarG/gfxreconstruct/pull/1979, closing

PR #1947, opened 2025-01-08
CommandList Reset at Invalid CommandAllocator

JerryAMD:
The content of the commandlist should be discarded when detect an invalid commandallocator


framework/encode/dx12_state_writer.cpp
```diff
@@ -1089,8 +1088,13 @@ void Dx12StateWriter::WriteCommandListCommands(const ID3D12CommandList_Wrapper*
...
 
-    // TODO: Don't write any commands, including the Reset or Close commands, if the command allocator used in the most
-    // recent Reset command no longer exists.
+    bool invalid_command_allocator =
+        (list_info->reset_command_allocator_id != format::kNullHandleId) &&
+        (state_table.GetID3D12CommandAllocator_Wrapper(list_info->reset_command_allocator_id) == nullptr);
+    if (invalid_command_allocator == true)
+    {
+        return;
+    }
```
davidd-lunarg, line 1097: Can this block be moved up above line 1084 `bool write_commands = CheckCommandListObjects(list_info.get(), state_table);`? That call to `CheckCommandListObjects` isn't needed if we are skipping all commands.

davidd-lunarg (approved):

PR #1952, opened 2025-01-13
Dx12 dump resources only after draw calls

davidlunarg:


USAGE_desktop_D3D12.md
```diff
@@ -276,77 +277,37 @@ Optional arguments:
   --pbi-all             Print all block information.
   --pbis <index1,index2>Print block information between block index1 and block index2.
 
-Windows-only:
   --fwo <x,y>           Force windowed mode if not already, and allow setting of a custom window location.
                         (Same as --force-windowed-origin)
-
```
bradgrantham-lunarg, line 282: Did you mean to delete this section and the "Windows-only" bit at line 279?

framework/decode/dx12_dump_resources.cpp
```diff
@@ -206,16 +208,21 @@ bool Dx12DumpResources::ExecuteCommandLists(DxObjectInfo*
...
+                                              graphics::dx12::Dx12DumpResourcePos::kBeforeDrawCall);
+                    }
 
                     ID3D12CommandList* ppCommandLists[] = { track_dump_resources_.split_command_sets[1].list };
                     replay_object->ExecuteCommandLists(1, ppCommandLists);
 
                     CopyDrawCallResources(replay_object_info,
                                           front_command_list_ids,
-                                          graphics::dx12::Dx12DumpResourcePos::kAfterDrawCall);
+                                          options_.dump_resources_before
```
bradgrantham-lunarg, line 223: I think this would be easier to read if `kAfterDrawCall` and `kAfterDrawCall_NoBefore` were bitmasks that were or'd, `kAfterDrawCall` and `kBeforeDrawCall` unless I've misunderstood this.

framework/decode/dx12_dump_resources.cpp
```diff
@@ -2556,8 +2564,8 @@ void DefaultDx12DumpResourcesDelegate::WriteResource(nlohmann::ordered_json&   j
     util::FieldToJson(jdata["res_id"], resource_data->source_resource_id, json_options_);
     util::FieldToJson(jdata["dimension"], util::ToString(resource_data->desc.Dimension), json_options_);
 
-    std::string suffix    = Dx12DumpResourcePosToString(resource_data->dump_position);
-    std::string json_path = suffix + "_file";
+    std::string suffix         = Dx12DumpResourcePosToString(resource_data->dump_position);
+    std::string json_path      = (suffix == "" ? "file" : (suffix + "_file"));
```
bradgrantham-lunarg, line 2568: This looks like a spot fix and not part of "dump resources only after draw calls".  Or is this a rebase issue?

bradgrantham-lunarg: Same question for other instances of filename and suffix processing below.

bradgrantham-lunarg (changes requested):

framework/decode/dx12_dump_resources.cpp
davidlunarg: On Vulkan, when --dump-resources-before-draw is specified, the bmp/bin/png files produced have the word "before" in the filename if it is a dump before the draw call and the word "after" it is is a dump after the draw call. If  --dump-resources-before-draw is not specified, the dump occurs after the draw call and file names do not include either "before" or "after". I added the "kAfterDrawCall_NoBefore"  for the case where the option is not specified and we don't want to include "before" or "after" in the file name. A bitmask would imply that two or more of these could occur at the same time, but that it not the case. The enums describe the three different cases.

framework/decode/dx12_dump_resources.cpp
davidlunarg: See my comments above regarding including "after" and "before". The strange expression is so that we can include an "_" to the file name when "after" or "before" is included in the name, but omit it otherwise. This makes the filenames consistent with Vulkan before/after filenames.

USAGE_desktop_D3D12.md
bradgrantham-lunarg: Heh, sure is D3D12-specific.  Wow, that inapplicable documentation has been in there a long time.

framework/decode/dx12_dump_resources.cpp
bradgrantham-lunarg: 👍

bradgrantham-lunarg:
I'd like @panos-lunarg to have a chance to offer comment, especially on anything that might change what Samsung sees from Vulkan dump resources.

framework/decode/dx12_dump_resources.cpp
```diff
@@ -2621,9 +2630,11 @@ void DefaultDx12DumpResourcesDelegate::TestWriteFloatResource(const std::string&
             data += "\n";
         }
 
-        const char* suffix        = Dx12DumpResourcePosToString(resource_data->dump_position);
-        std::string file_name_sub = file_name + "_sub_" + std::to_string(sub_index) + "_" + suffix + ".txt";
-        std::string file_path     = gfxrecon::util::filepath::Join(json_options_.root_dir, file_name_sub);
+        std::string suffix = Dx12DumpResourcePosToString(resource_data->dump_position);
+        std::string file_name_sub =
+            file_name + "_sub_" + std::to_string(sub_index) + (suffix == "" ? ".txt" : ("_" + suffix + ".txt"));
+        file_name_sub += (suffix == "" ? ".txt" : ("_" + suffix + ".txt"));
```
locke-lunarg, line 2636: Add ‵(suffix == "" ? ".txt" : ("_" + suffix + ".txt"))` twice.

davidlunarg: Good find. Fixed in subsequent commit. I don't have good test for this case, can you supply one?

Will squash this commit before merging to dev.

davidlunarg:
> I'd like @panos-lunarg to have a chance to offer comment, especially on anything that might change what Samsung sees from Vulkan dump resources.

This PR should not change the behavior of Vulkan dump resources. The --dump-resources-before-draw option is being added to dx12 and any difference in Vulkan dump resources would be a bug.

framework/decode/replay_options.h
```diff
@@ -71,6 +71,7 @@ struct ReplayOptions
     int32_t                      num_pipeline_creation_jobs{ 0 };
     std::string                  asset_file_path;
     std::string                  dump_resources_output_dir;
+    bool                         dump_resources_before{ false };
```
panos-lunarg, line 74: This variable is now common so it can be removed from `vulkan_replay_options.h`

panos-lunarg (approved):
> I'd like @panos-lunarg to have a chance to offer comment, especially on anything that might change what other users see from Vulkan dump resources.

These changes look fine from the Vulkan perspective

locke-lunarg (approved):

PR #1966, opened 2025-01-17
Remove retired swapchain images immediately

mikes-lunarg:
The implementation is allowed to destroy and reuse handles for non-acquired swapchain images as soon as the parent swapchain is retired. Update the handle tracking logic to treat these handles as destroyed, so that if they are reused, we don't get collisions in the handle tables.

Android is particularly sensitive to this as it will often immediately reuse image handles from retired swapchains, even before the retired swapchain is destroyed.

### DIFF
--- a/framework/encode/vulkan_capture_manager.cpp
+++ b/framework/encode/vulkan_capture_manager.cpp
@@ -1869,16 +1869,16 @@ void VulkanCaptureManager::PreProcess_vkCreateWaylandSurfaceKHR(VkInstance      
     }
 }
 
-void VulkanCaptureManager::PreProcess_vkCreateSwapchain(VkDevice                        device,
-                                                        const VkSwapchainCreateInfoKHR* pCreateInfo,
-                                                        const VkAllocationCallbacks*    pAllocator,
-                                                        VkSwapchainKHR*                 pSwapchain)
+void VulkanCaptureManager::PreProcess_vkCreateSwapchainKHR(VkDevice                        device,
+                                                           const VkSwapchainCreateInfoKHR* pCreateInfo,
+                                                           const VkAllocationCallbacks*    pAllocator,
+                                                           VkSwapchainKHR*                 pSwapchain)
 {
     GFXRECON_UNREFERENCED_PARAMETER(device);
     GFXRECON_UNREFERENCED_PARAMETER(pAllocator);
     GFXRECON_UNREFERENCED_PARAMETER(pSwapchain);
 
-    assert(pCreateInfo != nullptr);
+    GFXRECON_ASSERT(pCreateInfo != nullptr);
 
     if (pCreateInfo)
     {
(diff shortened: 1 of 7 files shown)


framework/encode/vulkan_handle_wrapper_util.h
```diff
@@ -200,9 +200,12 @@ void CreateWrappedDispatchHandle(typename ParentWrapper::HandleType parent,
         }
         if (!state_handle_table_.InsertWrapper(wrapper))
         {
-            GFXRECON_LOG_WARNING("Create a duplicated Handle: %" PRIu64
-                                 ". This wrapper can't be written into VulkanStateHandleTable.",
-                                 *handle);
+            auto existing_wrapper = state_handle_table_.GetWrapper<Wrapper>(wrapper->handle);
+            GFXRECON_LOG_WARNING("Cannot add duplicate entry to VulkanStateHandleTable for handle 0x%" PRIx64
+                                 " with ID %d. This handle is already wrapped with ID %d.",
```
panos-lunarg, line 205: Better used `PRIu64` instead of `%d` since `handle_id` is `uint64_t`. Same for similar warning in line 226

mikes-lunarg: Fixed

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -1889,6 +1889,59 @@ void VulkanCaptureManager::PreProcess_vkCreateSwapchain(VkDevice
...
+                                                            const VkAllocationCallbacks*    pAllocator,
+                                                            VkSwapchainKHR*                 pSwapchain)
+{
+    GFXRECON_UNREFERENCED_PARAMETER(device);
+    GFXRECON_UNREFERENCED_PARAMETER(pAllocator);
+    GFXRECON_UNREFERENCED_PARAMETER(pSwapchain);
+
+    GFXRECON_ASSERT(pCreateInfo != nullptr);
+
+    if (pCreateInfo)
```
ziga-lunarg, line 1904: I think this if is unnecessary

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -1889,6 +1889,59 @@ void VulkanCaptureManager::PreProcess_vkCreateSwapchain(VkDevice
...
+        // from oldSwapchain that are not acquired by the application may be freed by the implementation, which may
+        // occur even if creation of the new swapchain fails.
+
+        // The capture layer needs to be conservative and treat these images as destroyed now because the implementation
+        // is free to destroy and reuse the image handles before the retired swapchain is destroyed.
+        if (pCreateInfo->oldSwapchain != VK_NULL_HANDLE)
+        {
+            auto old_wrapper =
+                vulkan_wrappers::GetWrapper<vulkan_wrappers::SwapchainKHRWrapper>(pCreateInfo->oldSwapchain);
+            old_wrapper->retired = true;
```
ziga-lunarg, line 1916: Should there be an 
`if (!old_wrapper->retired)`
here?

What happens if `vkCreateSwapchainKHR()` is called with the same `oldSwapchain` twice?

mikes-lunarg: Passing the same `oldSwapchain` twice would be invalid

> VUID-VkSwapchainCreateInfoKHR-oldSwapchain-01933
> If oldSwapchain is not VK_NULL_HANDLE, oldSwapchain must be a non-retired swapchain associated with native window referred to by surface

We could add an `assert()` for that, what do you think?

ziga-lunarg: I didn't realize the oldSwapchain is retired even if `vkCreateSwapchainKHR()` fails, not an issue then

framework/encode/vulkan_capture_manager.cpp
mikes-lunarg: Fair, we have the assert. Even if we stop from crashing here, it will surely crash in the driver!

mikes-lunarg: Fixed

ziga-lunarg (approved):

panos-lunarg (approved):

bradgrantham-lunarg (approved):

PR #1969, opened 2025-01-18
Update SPIR-V Headers to 1.4.304

sarnex:
We should be using 1.4.304, and especially for the `vulkan-1.4.304.0` branch.

### DIFF
--- a/framework/util/spirv_helper.h
+++ b/framework/util/spirv_helper.h
@@ -1402,12 +1402,12 @@ const char* string_SpvOpcode(uint32_t opcode)
             return "OpFragmentFetchAMD";
         case spv::OpReadClockKHR:
             return "OpReadClockKHR";
-        case spv::OpFinalizeNodePayloadsAMDX:
-            return "OpFinalizeNodePayloadsAMDX";
+        case spv::OpEnqueueNodePayloadsAMDX:
+            return "OpEnqueueNodePayloadsAMDX";
         case spv::OpFinishWritingNodePayloadAMDX:
             return "OpFinishWritingNodePayloadAMDX";
-        case spv::OpInitializeNodePayloadsAMDX:
-            return "OpInitializeNodePayloadsAMDX";
+        case spv::OpAllocateNodePayloadsAMDX:
+            return "OpAllocateNodePayloadsAMDX";
         case spv::OpGroupNonUniformQuadAllKHR:
             return "OpGroupNonUniformQuadAllKHR";
         case spv::OpGroupNonUniformQuadAnyKHR:


sarnex:
Looks like CI is using old headers too. But the 1.4.304.0 branch of [this repo](https://github.com/LunarG/gfxreconstruct/tree/vulkan-sdk-1.4.304) fails to build because of this.

bradgrantham-lunarg:
Thank you for the PR!

The submodule for SPIRV-Headers in your branch https://github.com/sarnex/gfxreconstruct/tree/dev/external hasn't changed.  Is it possible you updated the submodule commit in your working copy and didn't commit and/or push that change?

bradgrantham-lunarg:
> Thank you for the PR!
> 
> The submodule for SPIRV-Headers in your branch https://github.com/sarnex/gfxreconstruct/tree/dev/external hasn't changed. Is it possible you updated the submodule commit in your working copy and didn't commit and/or push that change?

That is to say, the build for GFXReconstruct, whether in GitHub CI or in our internal Jenkins CI or for anyone who pulls your branch, uses the external/SPIRV-Headers commit.  To change the version any submodule, you'd have to go into the submodule directory, change the commit, then back out to the repo root, then commit the new hash of the submodule, then push that.

sarnex:
Thanks for the quick response.

Sorry, I didn't actually build `gfxreconstruct` using my branch. I'm one of the Gentoo Vulkan maintainers and we patch the code to use the pre-installed SPIR-V Headers and not the submodule.

Just looking at the `1.4.304.0` branch here, it seems the SPIR-V Headers commit is from around 1.3.290 (from [here](https://github.com/KhronosGroup/SPIRV-Headers/commit/eb49bb7b1136298b77945c52b4bbbc433f7885de)), while Vulkan-Headers is 1.4.304.0 as expected. 

I would expect all Vulkan SDK packages to be the same version. Can we bump the version in at least the 1.4.304 branch?

bradgrantham-lunarg:
For your convenience, I don't think we'd have any problem updating SPIRV-Headers to 304 especially if it's as simple as merging a PR.  Whether it gets cherry-picked into `vulkan-sdk-1.4.304` or would next land in `vulkan-sdk-1.4.304.1` would depend on direction from @KarenGhavam-lunarG and other staff inside LunarG.

Please update the SPIRV-Headers submodule commit in your PR branch and then we can test it in GitHub CI and our internal CI.

sarnex:
Just pushed that commit, thanks.

Just to be clear, I consider the fact that the `vulkan-sdk-1.4.304.0` branch here does not depend on SPIR-V headers 1.4.304.0 a bug, so I'd really appreciate if we could backport this PR if approved.

bradgrantham-lunarg:
Note this implies `submodule update` or `pull --recurse-submodules` or `submodule.recurse` global setting.

PR #1971, opened 2025-01-20
Enable AGS Markers to Replay

JerryAMD:


framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -1078,8 +1118,54 @@ HRESULT Dx12ReplayConsumerBase::OverrideD3D12CreateDevice(HRESULT
...
+
+        // Create AGS device for marker injection
+        AGSReturnCode ags_return = agsDriverExtensionsDX12_CreateDevice(
+            ags_marker_injector_->Context(), &creation_params, &extension_params, &returned_params);
+
+        if (ags_return == AGS_SUCCESS)
+        {
+            GFXRECON_LOG_DEBUG("Created AGS device.");
+
+            device->SetHandlePointerData(returned_params.pDevice);
```
davidd-lunarg, line 1141: I'd like to avoid modifications to `HandlePointerDecoder`, but I think this could be accomplished locally with existing `HandlePointerDecoder` methods:
```
            if (device->GetHandlePointer() != nullptr)
            {
                *device->GetHandlePointer() = returned_params.pDevice;
            }
```

framework/decode/gpu_cmd_wrapper.h
```diff
@@ -0,0 +1,56 @@
...
+#include "format/format.h"
+
+#include "decode/dx_replay_options.h"
+
+#include <d3d12.h>
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(decode)
+
+class GpuCmdWrapper
```
davidd-lunarg, line 36: For now this is an AGS specific class, so can you prefix the class and file names with AGS--`class AgsGpuCmdWrapper` and `ags_gpu_cmd_wrapper.h/cpp`?

davidd-lunarg (approved):
There is a conflict with `dev` that needs to be resolved and `handle_pointer_decoder.h` should be reverted. Otherwise this looks good and will be ready to merge with those changes.

PR #1973, opened 2025-01-21
Fix mip level calculation

panos-lunarg:


framework/util/defines.h
```diff
@@ -67,6 +67,11 @@ static constexpr uint64_t GFXRECON_PTR_TO_UINT64(T ptr)
     return static_cast<uint64_t>(reinterpret_cast<uintptr_t>(ptr));
 }
 
+static constexpr uint32_t GFXRECON_LOG2(uint32_t n)
+{
+    return (n > 1) ? 1 + GFXRECON_LOG2(n >> 1) : 0;
```
antonio-lunarg, line 72: Do you think we can use this?
```
static constexpr uint32_t GFXRECON_LOG2(uint32_t x)
{
    return 31 - __builtin_clz(x | 1);
}
```

bradgrantham-lunarg: Does that compile on all platforms?

bradgrantham-lunarg: I ran a bunch of tests of this against just calling cmath `log2` with a `static_cast` to `int` and it looks like the performance of recursive tests against a shift is substantially slower than cmath's `log2` on ARMv8-a (ARM64) (https://godbolt.org/z/qsbYzcv5j).  On x86 it looks like the function is turned into some kind of magic series of instructions (maybe a `clz`) and is slightly faster than calling `log2`.  This isn't currently in a critical path that is in an inner loop, but If you check in a `GFXRECON_LOG2` function, I think there's a chance someone will use it in an inner loop so I suspect it's a good idea to optimize this.  You can probably convince me that I'm wrong.

bradgrantham-lunarg: Or that my test code is wrong.

panos-lunarg: I could swear that `log2` was returning some wrong values causing the assertion to fail so I replaced it with `GFXRECON_LOG2`. Now I cannot reproduce this....

bradgrantham-lunarg:
Close this PR?

bradgrantham-lunarg:
Closing, can reopen if we can identify examples of math errors in our MIP size calculations

PR #1975, opened 2025-01-21
Avoid warning: Descriptor type 1000138000 not handled

fabian-lunarg:
- that's VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK and we do handle it


MarkY-LunarG (approved):
Short and simple.

PR #1985, opened 2025-01-24
fix: Add partial pipeline storage structure to prevent GPL data loss

dongcheol-samsung:
Added new structure to preserve partial pipeline data during GPL processing
instead of discarding it. This prevents tool from losing incomplete pipeline
information that may be needed later.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -2370,6 +2370,197 @@ bool VulkanReplayConsumerBase::CheckCommandBufferInfoForFrameBoundary(
     }
     return false;
 }
+// for Linking Graphics Pipelines Libraries start[--
+#if defined(VK_USE_PLATFORM_ANDROID_KHR)
+std::vector<VkPipeline> libraries(3);
+std::vector<VkPipeline> pipeLibraries(4);
+
+void VulkanReplayConsumerBase::LinkGraphicsPipelines(const VkGraphicsPipelineCreateInfo* createInfo)
+{
+    if (createInfo->flags == 0x0 && createInfo->pNext != nullptr)
+    {
+        const VkPipelineCreationFeedbackCreateInfo* feedbackInfo =
+            (reinterpret_cast<const VkPipelineCreationFeedbackCreateInfo*>(createInfo->pNext));
+        if (feedbackInfo->pNext != nullptr)
+        {
+            const VkPipelineLibraryCreateInfoKHR* pipelineLibraryInfo =
+                (reinterpret_cast<const VkPipelineLibraryCreateInfoKHR*>(feedbackInfo->pNext));
+            const VkDescriptorSetVariableDescriptorCountAllocateInfo* libraryLinkInfo =
+                (reinterpret_cast<const VkDescriptorSetVariableDescriptorCountAllocateInfo*>(
+                    pipelineLibraryInfo->pNext));
+            VkPipelineLibraryCreateInfoKHR* pTempLibraries =
+                (const_cast<VkPipelineLibraryCreateInfoKHR*>(pipelineLibraryInfo));
+            pTempLibraries->pLibraries = nullptr;
+            if (libraryLinkInfo != nullptr)
+            {
+                if (libraryLinkInfo->descriptorSetCount == 4)
+                {
+                    GFXRECON_LOG_WARNING("[replay]vkCreateGraphicsPipelines:link:descriptorSetCount=%d[%d][%d][%d][%d]",
+                                         libraryLinkInfo->descriptorSetCount,
+                                         libraryLinkInfo->pDescriptorCounts[0],
+                                         libraryLinkInfo->pDescriptorCounts[1],
+                                         libraryLinkInfo->pDescriptorCounts[2],
+                                         libraryLinkInfo->pDescriptorCounts[3]);
+                }
+                else
+                {
+                    GFXRECON_LOG_WARNING("[replay]vkCreateGraphicsPipelines:link:descriptorSetCount=%d[%d][%d][%d]",
+                                         libraryLinkInfo->descriptorSetCount,
+                                         libraryLinkInfo->pDescriptorCounts[0],
+                                         libraryLinkInfo->pDescriptorCounts[1],
+                                         libraryLinkInfo->pDescriptorCounts[2]);
+                }
+                std::vector<VkPipeline> pipelines(pipelineLibraryInfo->libraryCount);
+                if (libraryLinkInfo->pDescriptorCounts[0] >= vertexInputPipelines.size())
+                {
+                    pipelines[0] = vertexInputPipelines[vertexInputPipelines.size() - 1];
+                }
+                else
+                {
+                    for (size_t i = 0; i < vertexInputPipelines.size(); i++)
+                    {
+                        if (libraryLinkInfo->pDescriptorCounts[0] == i)
+                        {
+                            pipelines[0] = vertexInputPipelines[i];
+                            break;
+                        }
+                    }
+                }
+
+                if (libraryLinkInfo->pDescriptorCounts[1] >= shadersPipelines.size())
+                {
+                    pipelines[1] = shadersPipelines[shadersPipelines.size() - 1];
+                }
+                else
+                {
+                    for (size_t j = 0; j < shadersPipelines.size(); j++)
+                    {
+                        if (libraryLinkInfo->pDescriptorCounts[1] == j)
+                        {
+                            pipelines[1] = shadersPipelines[j];
+                            break;
+                        }
+                    }
+                }
+
+                if (pipelineLibraryInfo->libraryCount == 4)
+                {
+                    if (libraryLinkInfo->pDescriptorCounts[2] >= fragmentShaderPipelines.size())
+                    {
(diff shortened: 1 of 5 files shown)


framework/encode/custom_vulkan_api_call_encoders.cpp
```diff
@@ -542,6 +550,161 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice
...
+                    }
+                }
+                if (fragmentIdx >= fragmentOutputPipelines.size())
+                    fragmentIdx = fragmentOutputPipelines.size() - 1;
+                indexes.push_back(fragmentIdx);
+            }
+
+            linkInfo.pDescriptorCounts                   = indexes.data();
+            VkPipelineLibraryCreateInfoKHR* pLibraryInfo = (const_cast<VkPipelineLibraryCreateInfoKHR*>(libraryInfo));
+            pLibraryInfo->pNext                          = reinterpret_cast<VkBaseOutStructure*>(&linkInfo);
```
panos-lunarg, line 650: I am not sure I understand what's the purpose here.

`libraryInfo` is of type  `VkDescriptorSetVariableDescriptorCountAllocateInfo` which is not a valid structure to appear in the `VkGraphicsPipelineCreateInfo` pNext chain.
This takes place _after_ the call to `vkCreateGraphicsPipelines` has been forwarded to the driver (in line 550 above) so it seems that the intention here is to inject an invalid structure in the pnext chain with the purpose to store it in the capture file.
Additionally `linkInfo`'s is a stack object whose scope is inside the `if (libraryInfo != nullptr)` brackets'. By the time the encoder is called in line 762. `&linkInfo` will be pointing to garbage. Same applies to `indices` vector whose data pointer is used couple of lines above.

Now trying to apply an example in order to clarify the intentions here. Let's say there have been 4 calls to `vkCreateGraphicsPipelines` before, one of each creates a library pipeline (`VK_PIPELINE_CREATE_LIBRARY_BIT_KHR` is provided in the `pCreateInfo->flags`). Then a 5th call with flags equal to zero. This will populate the `VkDescriptorSetVariableDescriptorCountAllocateInfo` like this:
```
VkDescriptorSetVariableDescriptorCountAllocateInfo linkInfo
{
    .sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_VARIABLE_DESCRIPTOR_COUNT_ALLOCATE_INFO
    .pNext = nullptr
    .descriptorSetCount = 4
    .pDescriptorCounts = [0, 0, 0, 0]
}
```
Still I don't understand the intention

framework/generated/generated_vulkan_struct_encoders.cpp
```diff
@@ -5948,6 +5948,18 @@ void EncodeStruct(ParameterEncoder* encoder, const VkPipelineExecutableInternalR
 
 void EncodeStruct(ParameterEncoder* encoder, const VkPipelineLibraryCreateInfoKHR& value)
 {
+    // fix gpl issue for release build.
+#if defined(VK_USE_PLATFORM_ANDROID_KHR)
+    if (value.pNext != nullptr)
+    {
+        auto base = reinterpret_cast<const VkBaseOutStructure*>(value.pNext);
+        if (base->sType == 1)
```
panos-lunarg, line 5956: Not sure I understand what's the intention here.

This `EncodeStruct` function will be called when encoding the pNext chain of the `VkGraphicsPipelineCreateInfo`. This has already being tampered in this PR by the changes in the `CreateGraphicsPipelines()` function in `custom_vulkan_api_call_encoders.cpp`. I don't understand the purpose of the changes here.

A `VkStructureType` with the value 1 is `VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO = 1`. When encoding the pNext chain of `VkPipelineLibraryCreateInfoKHR` this would be unexpected and invalid to appear.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2370,6 +2370,197 @@ bool VulkanReplayConsumerBase::CheckCommandBufferInfoForFrameBoundary(
     }
     return false;
 }
+// for Linking Graphics Pipelines Libraries start[--
+#if defined(VK_USE_PLATFORM_ANDROID_KHR)
+std::vector<VkPipeline> libraries(3);
```
bradgrantham-lunarg, line 2375: These global variables and the others below probably need to be keyed at least per-device

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2370,6 +2370,197 @@ bool VulkanReplayConsumerBase::CheckCommandBufferInfoForFrameBoundary(
...
+{
+    if (createInfo->flags & VK_PIPELINE_CREATE_LIBRARY_BIT_KHR)
+    {
+        const VkPipelineCreationFeedbackCreateInfo* feedbackInfo =
+            (reinterpret_cast<const VkPipelineCreationFeedbackCreateInfo*>(createInfo->pNext));
+        const VkGraphicsPipelineLibraryCreateInfoEXT* libraryInfo =
+            (reinterpret_cast<const VkGraphicsPipelineLibraryCreateInfoEXT*>(feedbackInfo->pNext));
+        GFXRECON_LOG_WARNING(
+            "[replay]vkCreateGraphicsPipelines:compile:flags[0x%x], pPipelines=%p", libraryInfo->flags, *pipeline);
+        if (libraryInfo->flags == VK_GRAPHICS_PIPELINE_LIBRARY_VERTEX_INPUT_INTERFACE_BIT_EXT)
```
bradgrantham-lunarg, line 2539: Since these are `BIT` values and not an enumerant, this comparison and the ones below would need to be anded with the BIT to test first, to handle the (unlikely) case that a library contains multiple subsets.

framework/encode/custom_vulkan_api_call_encoders.cpp
```diff
@@ -542,6 +550,161 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice
...
+    {
+        const VkPipelineCreationFeedbackCreateInfo* feedbackInfo =
+            (reinterpret_cast<const VkPipelineCreationFeedbackCreateInfo*>(pCreateInfos->pNext));
+        const VkPipelineLibraryCreateInfoKHR* libraryInfo =
+            (reinterpret_cast<const VkPipelineLibraryCreateInfoKHR*>(feedbackInfo->pNext));
+        if (libraryInfo != nullptr)
+        {
+            if (libraryInfo->libraryCount == 4)
+            {
+                GFXRECON_LOG_WARNING("[capture]vkCreateGraphicsPipelines:link:libraryCount=%d,pLibraries[0]=%p,"
```
bradgrantham-lunarg, line 565: These look like debugging messages, and if so, probably they should be `GFXRECON_LOG_DEBUG`.  If these are generally for debugging and testing this PR, please consider making the PR a draft and consider removing the debugging statements before moving from draft to final.

framework/encode/custom_vulkan_api_call_encoders.cpp
```diff
@@ -542,6 +550,161 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice
...
+                if (libraryInfo->pLibraries[1] == shadersPipelines[j])
+                {
+                    shaderIdx = j;
+                }
+            }
+            if (shaderIdx >= shadersPipelines.size())
+                shaderIdx = shadersPipelines.size() - 1;
+            indexes.push_back(shaderIdx);
+
+            if (libraryInfo->libraryCount == 4)
```
bradgrantham-lunarg, line 610: Why is `4` a special case?

framework/encode/custom_vulkan_api_call_encoders.cpp
```diff
@@ -542,6 +550,161 @@ VKAPI_ATTR VkResult VKAPI_CALL CreateGraphicsPipelines(VkDevice
...
+                }
+                if (fragmentIdx >= fragmentOutputPipelines.size())
+                    fragmentIdx = fragmentOutputPipelines.size() - 1;
+                indexes.push_back(fragmentIdx);
+            }
+            else
+            {
+                for (size_t k = 0; k < fragmentOutputPipelines.size(); k++)
+                {
+                    if (libraryInfo->pLibraries[2] == fragmentOutputPipelines[k])
```
bradgrantham-lunarg, line 638: If libraryCount is < 3, this will be an out-of-bounds error.  The spec also doesn't guarantee ordering of the libraries; the bit in `flags` specifies what subset is being specified. My guess is that this hardcodes a particular app's layout of libraries.

bradgrantham-lunarg (changes requested):
Thanks for your PR!  This PR seems like it may encode the behavior of a particular app or engine, and it seems possible that the PR contains a spot fix for an application bug.    I think we're more likely to approve and merge this if you can tell us more about the intent for these changes, and what problem is actually being resolved by this PR.

bradgrantham-lunarg:
Is there a possibility you can provide a Vulkan test program that illustrates the bug?  Can you modify the Khronos `graphics_pipeline_library` sample to express the case that fails?

dongcheol-samsung:
@wpaul-samsung I tested Genshin, BlackDesertMobile, and SubwaySurfers, found no issue.

dongcheol-samsung:
https://github.com/LunarG/gfxreconstruct/pull/1994 fix the issues.

PR #1988, opened 2025-01-25
Clear CMake warnings about minimum versions and CMP135

bradgrantham-lunarg:


charles-lunarg (approved):
I definitely approve of updating the version - 3.4 is roughly from 2016 so definitely can be brought up past it.

PR #1990, opened 2025-01-25
Catch Exceptions by Reference to Avoid Slicing

ahcox:
Preserve messages from derived exceptions for more informative logging.

Catching them by value means making a copy of the thrown exception, and if it was actually derived from the type named at the catch point, that will be a slicing copy, throwing away the fields of the derived type which might contain information it uses to provide a helpful `what()` message. Indeed, the virtual `what()` function will resolve to the base class version, returning an empty string in the cases of the test apps.

[Ref: C++ Core Guidelines, E.15: Throw by value, catch exceptions from a hierarchy by reference.](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e15-throw-by-value-catch-exceptions-from-a-hierarchy-by-reference)

### DIFF
--- a/test/test_apps/host-image-copy/app.cpp
+++ b/test/test_apps/host-image-copy/app.cpp
@@ -483,7 +483,7 @@ int main(int argc, char* argv[])
         app.run("host image copy");
         return 0;
     }
-    catch (std::exception e)
+    catch (std::exception& e)
     {
         std::cout << e.what() << std::endl;
         return -1;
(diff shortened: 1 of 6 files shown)


antonio-lunarg (approved):

bradgrantham-lunarg:
The CI "Ubuntu GCC Release" failures were from a clang-format error on code not touched as part of this PR.

PR #1992, opened 2025-01-27
gfxrecon-info now prints vulkan physical device info

panos-lunarg:


framework/decode/vulkan_stats_consumer.h
```diff
@@ -56,6 +58,26 @@ class VulkanStatsConsumer : public gfxrecon::decode::VulkanConsumer
     uint64_t           GetAllocationCount() const { return allocation_count_; }
     uint64_t           GetMinAllocationSize() const { return min_allocation_size_; }
     uint64_t           GetMaxAllocationSize() const { return max_allocation_size_; }
+    void               GetPhysicalDeviceProperties(uint32_t&    api_version,
```
antonio-lunarg, line 61: not: not a big fun of output parameters, I would introduce a new struct to group these.

framework/decode/vulkan_stats_consumer.h
```diff
@@ -377,6 +399,30 @@ class VulkanStatsConsumer : public gfxrecon::decode::VulkanConsumer
         ++dispatch_count_;
     }
 
+    virtual void ProcessSetDevicePropertiesCommand(format::HandleId   physical_device_id,
```
bradgrantham-lunarg, line 402: There will be one of these for every VkPhysicalDevice so this needs to handle multiple SetDevicePropertiesCommand blocks.  I have a .gfxr here with 2 of those blocks, one for an AMD discrete GPU and another for the integrated Intel GPU, so the blocks can't be assumed to have the same values.

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_stats_consumer.h
panos-lunarg: should be fixed now

Fahien (approved):

antonio-lunarg (approved):

bradgrantham-lunarg (approved):

PR #1993, opened 2025-01-27
Portable RT: Fixes for shader-binding-table assembly (sbt data)

fabian-lunarg:
- adjust replacer compute-shaders to allow for pass-through of data
- Drop `GL_GOOGLE_include_directive` from internal compute-shader, unused and caused issues on radv
- rework num-addresses calculation, account for extra data following handles.
- Use `accelerationStructureCaptureReplay` only when allocator supports opaque addresses
- minor adjustments and fixes for edge-cases

this fixes portable replay (using `-m rebind`) of applications using shader-record buffers, sourcing data from shader-binding-tables.
in glsl e.g. `layout(shaderRecordEXT, std430) buffer ShaderRecordBuffer{...}`
in SPIRV: `StorageClassShaderRecordBufferKHR`

in presence of that feature, when re-assembling a new sbt, it's not sufficient to remap handles.
also the extra data needs to be copied/passed-through.
this is fixed in this PR.

the simplest example around is: https://github.com/SaschaWillems/Vulkan/tree/master/examples/raytracingsbtdata

also this PR fixes issues encountered with portable-Rt on radv


framework/decode/vulkan_address_replacer.cpp
```diff
@@ -497,7 +507,20 @@ void VulkanAddressReplacer::ProcessCmdTraceRays(
                     VK_ACCESS_SHADER_READ_BIT);
         }
 
-        // set previous push-constant data
+        // set previous compute-pipeline, if any
+        if (command_buffer_info->bound_pipelines.count(VK_PIPELINE_BIND_POINT_COMPUTE))
+        {
+            auto* previous_pipeline = object_table_->GetVkPipelineInfo(
+                command_buffer_info->bound_pipelines.at(VK_PIPELINE_BIND_POINT_COMPUTE));
+            GFXRECON_ASSERT(previous_pipeline);
```
antonio-lunarg, line 515: `previous_pipeline` MUST be valid?

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -811,7 +838,20 @@ void VulkanAddressReplacer::ProcessCmdBuildAccelerationStructuresKHR(
                         VK_ACCESS_SHADER_READ_BIT);
             }
 
-            // set previous push-constant data
+            // set previous compute-pipeline, if any
+            if (command_buffer_info->bound_pipelines.count(VK_PIPELINE_BIND_POINT_COMPUTE))
+            {
+                auto* previous_pipeline = object_table_->GetVkPipelineInfo(
+                    command_buffer_info->bound_pipelines.at(VK_PIPELINE_BIND_POINT_COMPUTE));
+                GFXRECON_ASSERT(previous_pipeline);
```
antonio-lunarg, line 846: ditto

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: yes. at this point we already checked the `bound_pipelines` map and know a compute-pipeline was previously bound (rare'ish case anyway, since the command-buffer is used to record a raytracing pipeline). so in fact, we only assert that it can also be mapped.

fabian-lunarg: double-checked the assumption, I think it's correct. the only caveat I could think of if a pipeline would have been bound and then unbound again. but that's [not allowed](https://registry.khronos.org/vulkan/specs/latest/man/html/vkCmdBindPipeline.html#VUID-vkCmdBindPipeline-pipeline-parameter)

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -497,7 +507,20 @@ void VulkanAddressReplacer::ProcessCmdTraceRays(
...
-        // set previous push-constant data
+        // set previous compute-pipeline, if any
+        if (command_buffer_info->bound_pipelines.count(VK_PIPELINE_BIND_POINT_COMPUTE))
+        {
+            auto* previous_pipeline = object_table_->GetVkPipelineInfo(
+                command_buffer_info->bound_pipelines.at(VK_PIPELINE_BIND_POINT_COMPUTE));
+            GFXRECON_ASSERT(previous_pipeline);
+            if (previous_pipeline != nullptr)
+            {
+                device_table_->CmdBindPipeline(
```
MarkY-LunarG, line 518: I think you should emit an info message like "Restoring previous pipeline" or some such just so people are aware of what's going on.

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -811,7 +838,20 @@ void VulkanAddressReplacer::ProcessCmdBuildAccelerationStructuresKHR(
...
-            // set previous push-constant data
+            // set previous compute-pipeline, if any
+            if (command_buffer_info->bound_pipelines.count(VK_PIPELINE_BIND_POINT_COMPUTE))
+            {
+                auto* previous_pipeline = object_table_->GetVkPipelineInfo(
+                    command_buffer_info->bound_pipelines.at(VK_PIPELINE_BIND_POINT_COMPUTE));
+                GFXRECON_ASSERT(previous_pipeline);
+                if (previous_pipeline != nullptr)
+                {
+                    device_table_->CmdBindPipeline(
```
MarkY-LunarG, line 849: Info message here as well.

MarkY-LunarG (approved):

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -365,10 +365,18 @@ void VulkanAddressReplacer::ProcessCmdTraceRays(
...
                 {
-                    num_handles_map[region] = region->size / region->stride;
+                    num_addresses_map[region] = region->size / capture_ray_properties_->shaderGroupHandleSize;
 
-                    for (uint32_t offset = 0; offset < region->size; offset += region->stride)
+                    uint32_t capture_handle_size = capture_ray_properties_->shaderGroupHandleSize;
+                    if (region->stride > capture_handle_size)
                     {
+                        uint32_t payload_size = region->stride - capture_handle_size;
+                        GFXRECON_LOG_DEBUG("Extra data in sbt: %d", payload_size);
```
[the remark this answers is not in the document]
fabian-lunarg: that's true, it's not really used. the log was just to confirm/print on that case and left over, no need for a warning. I'll remove it

framework/decode/vulkan_address_replacer.cpp
bradgrantham-lunarg: @antonio-lunarg is your concern resolved?

framework/decode/vulkan_address_replacer.cpp
bradgrantham-lunarg: @antonio-lunarg is your concern resolved?

framework/decode/vulkan_address_replacer.cpp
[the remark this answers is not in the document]
fabian-lunarg: done, only left a `// NOTE:` to hint at the potential presence of shaderRecord-buffers.

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: done

PR #1994, opened 2025-01-27
MapPNextStructHandles should traverse chain till end

panos-lunarg:


MarkY-LunarG (approved):
LGTM, but Zulauf should really look this over as well since he's actively still working the codegen.

panos-lunarg:
> LGTM, but Zulauf should really look this over as well since he's actively still working the codegen.

Saw your name in git blame on neighboring lines and summoned you.
I am a bit concerned that this wasn't detected earlier.

MarkY-LunarG:
> > LGTM, but Zulauf should really look this over as well since he's actively still working the codegen.
> 
> Saw your name in git blame on neighboring lines and summoned you. I am a bit concerned that this wasn't detected earlier.

Yes, apparently our tests aren't good enough to catch this.  This was part of the code refactor and we've caught other errors, but this is the concern that as we touch code we'll have things slip in like this.  Good catch.

bradgrantham-lunarg (approved):

PR #1997, opened 2025-01-28
Isolate higher min cmake to tests

bradgrantham-lunarg:
* lower some cmake min versions where possible
* put the higher cmake min in tests


charles-lunarg (approved):
This should work to raise the minimum where its needed without affecting build compatibility elsewhere.

Issue #1762, opened 2024-09-25
Create a utility for indexing draw calls for dump resources


davidd-lunarg:
Create a utility for turning dump resource arguments (Nth call to ExecuteCommandLists, Nth CommandList, Nth draw call) into GFXR capture file block indices. If possible make the utility generic enough to work for either Vulkan or DX12. Return a data object which allows lookup of any draw call in the capture file (not just 1)

PR #1998, opened 2025-01-29
Vulkan Dump: Convert call index to block index

locke-lunarg:
1. Change `--dump-resources` from `block indices` to `call indices`. The feature is similar to dx12.

2. Rename old `--dump-resources` to `--dump-resources-block-indices`.

3. Rename `dx12_browse_consumer.h` to `dx12_pre_process_consumer.h` and `Dx12BrowseConsumer` to `Dx12PreProcessConsumer`.

4. Add new `vulkan_pre_process_consumer.h` for convert `call indices` to `block indices`.

5. Move `parse_dump_resources_arg` to `RunVulkanPreProcessConsumer`. It runs after `VulkanPreProcessConsumer` is done.

6. `new VulkanReplayDumpResources` runs later. It runs after `parse_dump_resources_arg`.

7. `options_` in `VulkanReplayConsumerBase` becomes reference for updating dump resources args from `parse_dump_resources_arg`.

8. Set `replay_options`: `enable_vulkan` and `enable_d3d12`.  `enable_vulkan` and `enable_d3d12` were always true.

Closed: https://github.com/LunarG/gfxreconstruct/issues/1762


locke-lunarg:
However, if it renames old `--dump-resources` to `--dump-resources-block-indices` that cause too many changes for our CI and our users, I could undo this change and add new option `--dump-resources-call-indices` for this new feature. But it will make that dx12's `--dump-resources` and vulkan's `--dump-resources are different. I wish that won't confuse users.

davidlunarg:
There are three commits in this PR. Each adds a different feature. Should this PR be broken up into three PRs?

locke-lunarg:
These three commits are depended. The first commit only adds a new option, but the option feature isn't  implemented.  The second commit implements the feature, but when it runs vulkan --dump-resources, it crashes in dx12 --dump-resources, so it needed the third commit to fix it.

locke-lunarg:
The reason of the CI failed could be that `--dump-resources`  renamed to `--dump-resources-block-indices` . We might have to modify the CI tests.

framework/decode/vulkan_pre_process_consumer.h
```diff
@@ -0,0 +1,774 @@
...
+
+#include "generated/generated_vulkan_consumer.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(decode)
+
+const std::string DUMP_ARG_BEGIN_COMMAND_BUFFER = "BeginCommandBuffer";
+const std::string DUMP_ARG_DRAW                 = "Draw";
+const std::string DUMP_ARG_RENDER_PASS          = "RenderPass";
+const std::string DUMP_ARG_BEGIN_RENDE_PASS     = "BeginRenderPass";
```
davidlunarg, line 34: Why not DUMP_ARG_BEGIN_RENDER_PASS?

tools/replay/desktop_main.cpp
```diff
@@ -88,29 +89,82 @@ void WaitForExit() {}
...
+            replay_options.enable_dump_resources && replay_options.dump_resources_block_indices.empty();
+        if (process_dump_resources)
+        {
+            pre_process_consumer.EnableDumpResources(replay_options.dump_resources_target);
+        }
+
+        gfxrecon::decode::VulkanDecoder decoder;
+        decoder.AddConsumer(&pre_process_consumer);
+        file_processor.AddDecoder(&decoder);
+        file_processor.ProcessAllFrames();
```
davidlunarg, line 111: Does this mean that the whole capture file will be read in in order to generated the dump parameters? Would it be possible if the dump occurs early in the capture file to end this preprocessing step  prematurely to avoid reading the whole file?

locke-lunarg: `VulkanPreProcessConsumer pre_process_consumer` became necessary to run because it isn't only for dump, also for checking if this replay is Vulkan or dx12.  See `replay_options.enable_vulkan = pre_process_consumer.WasVulkanAPIDetected();`. 

Plus, It checks dump resources early
```
if (process_dump_resources)
{
    pre_process_consumer.EnableDumpResources(replay_options.dump_resources_target);
}
```
If it's not enabled, `IsComplete` will return true after knowing its Vulkan or dx12 replay, and quit this process.
```
virtual bool IsComplete(uint64_t block_index) override
{
    return check_vulkan_consumer_usage_complete_ && (check_dump_resources_complete_ || !enable_dump_resources_);
}
```

framework/decode/vulkan_pre_process_consumer.h
locke-lunarg: Thank you. Nice catch. It was typo. Fixed.

tools/replay/desktop_main.cpp
```diff
@@ -88,29 +89,82 @@ void WaitForExit() {}
...
+                replay_options.dump_resources_block_indices = pre_process_consumer.GetDumpResourcesBlockIndices();
+            }
+
+            if (replay_options.enable_dump_resources)
+            {
+                // Process --dump-resources-block-indices arg.
+                if (!gfxrecon::parse_dump_resources::parse_dump_resources_arg(replay_options))
+                {
+                    GFXRECON_LOG_FATAL("There was an error while parsing dump resources indices. Terminating.");
+                    return;
```
davidlunarg, line 128: GFXRECON_LOG_FATAL prints an error message and returns, then this method will return, I think to continue trying to do the replay. Should the return be changed to a call to exit()?

framework/decode/dx12_pre_process_consumer.h
```diff
@@ -20,17 +20,20 @@
...
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
+#define CHECK_DX12_CONSUMER_USAGE() \
+    if (!dx12_consumer_usage_)      \
+    {                               \
+        return;                     \
+    }
+
```
davidlunarg, line 36: This macro is can only be used inside the Dx12PreProcessConsumer class. Maybe it should be defined inside that class? I realized that it being defined inside that class doesn't prevent it from being invoked outside the class, but it does make it clear that is where it is intended to be used.)

locke-lunarg: Good idea. Moved.

tools/replay/desktop_main.cpp
locke-lunarg: Added exit(0);

davidlunarg:
Might we want to include both the command indices and the block indices in the output json file when command indices were used on the command line? It seems that would help in correlating an output file with the command line that was used to generate that output file.

locke-lunarg:
> Might we want to include both the command indices and the block indices in the output json file when command indices were used on the command line? It seems that would help in correlating an output file with the command line that was used to generate that output file.

@davidlunarg  Added in a new commit.

tools/tool_settings.h
```diff
@@ -1114,13 +1115,44 @@ GetVulkanReplayOptions(const gfxrecon::util::ArgumentParser&           arg_parse
...
     }
 
-    replay_options.dump_resources              = arg_parser.GetArgumentValue(kDumpResourcesArgument);
+    const std::string& dump_resources = arg_parser.GetArgumentValue(kDumpResourcesArgument);
+    if (!dump_resources.empty() && dump_resources.find_first_not_of("0123456789,") == std::string::npos)
+    {
+        // If the option parameter does not split into three comma separated numbers, consider
+        // it a Vulkan option and ignore it. If it does split into three comma separated numbers,
+        // the arg is for dx12 and should have already been validated in the Vulkan option parsing.
+        // In that case, we simply extract and save the values here.
```
davidlunarg, line 1124: This comment is incorrect - an arg of 3 numbers separated by commas can now by used for Vulkan.

davidlunarg:
Some of the documentation changes could use rewording. I have provided Locke with my proposed changes.

tools/tool_settings.h
locke-lunarg: This is a outdated comment.  Removed. Thanks.

davidlunarg (approved):

PR #2001, opened 2025-01-29
Implement dx12 dump resources if in modifiable state

davidlunarg:


locke-lunarg (changes requested):

framework/decode/dx12_dump_resources.h
```diff
@@ -60,6 +60,13 @@ enum class Dx12DumpResourceType : uint32_t
     kComputeRootParameters,
 };
 
+const uint64_t modifiableTransitionStates =
+    D3D12_RESOURCE_STATE_RENDER_TARGET | D3D12_RESOURCE_STATE_DEPTH_WRITE | D3D12_RESOURCE_STATE_STREAM_OUT |
+    D3D12_RESOURCE_STATE_COPY_DEST | D3D12_RESOURCE_STATE_RESOLVE_DEST |
+    D3D12_RESOURCE_STATE_RAYTRACING_ACCELERATION_STRUCTURE | D3D12_RESOURCE_STATE_ALL_SHADER_RESOURCE |
+    D3D12_RESOURCE_STATE_VIDEO_DECODE_WRITE | D3D12_RESOURCE_STATE_VIDEO_PROCESS_WRITE |
+    D3D12_RESOURCE_STATE_VIDEO_ENCODE_WRITE;
+
```
locke-lunarg, line 69: Is there any document to show what states are modifiable?

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -4554,6 +4554,18 @@ void Dx12ReplayConsumerBase::PreCall_ID3D12GraphicsCommandList_ResourceBarrier(
...
+            if (options_.enable_dump_resources)
+            {
+                GFXRECON_ASSERT(dump_resources_);
+                if (state.transition.StateAfter & modifiableTransitionStates)
+                {
+                    dump_resources_->ModifiableResourceAdd(barriers[i].Transition->pResource);
+                }
+                else
+                {
+                    dump_resources_->ModifiableResourceRemove(barriers[i].Transition->pResource);
```
locke-lunarg, line 4566: Here might not be a good place to record states.  `command_list->ResourceBarrier` don't change on the resource states  until the `command_list` is executed on `ExecuteCommandLists`.  A `ResourceBarrier` could be before the target `ExecuteCommandLists`, but its `ExecuteCommandLists` is after the target `ExecuteCommandLists`. In this case, changing state doesn't happen before dump resources.

Getting a correct resource state is complicated. It finds the correct resource states for copying resources on `Dx12DumpResources::CopyResourceAsyncQueue`. Maybe you could decide if the resource should be dumped there.  But from what I remember, it's not prefect. It triggered validations for wrong states sometimes.

@davidd-lunarg Do you think that the wrong states issues still happen sometimes?

framework/decode/dx12_dump_resources.h
davidd-lunarg: `D3D12_RESOURCE_STATE_COMMON` is a special state which also may be writeable. A resource state can be promoted from `D3D12_RESOURCE_STATE_COMMON` to other resource states--both read or write states--depending on how the resource is used.

framework/decode/dx12_dump_resources.cpp
```diff
@@ -2092,8 +2095,38 @@ void Dx12DumpResources::CopyResourceAsyncRead(graphics::dx12::ID3D12FenceComPtr
...
+        {
+            resourceIsModifiable =
+                ((modifiableTransitionStates && it->states) || it->states == D3D12_RESOURCE_STATE_COMMON);
+        }
+
+        // Dump the resource if is it in modifiable state
+        if (resourceIsModifiable)
+        {
+            active_delegate_->DumpResource(copy_resource_data);
+        }
```
locke-lunarg, line 2126: Should `D3D12_RESOURCE_STATE_COMMON` also be added in `modifiableTransitionStates`? Or should it stay separately?

Here is to dump the resource if the whole subresources are modifiable. Should it dump the part modifiable subresources?

davidlunarg: modifiableTransitionStates is a bitmask of the stated that would trigger a dump. Unfortunately, D3D12_RESOURCE_STATE_COMMON is defined as 0, so can't add this value to modifiableTransitionStates. It gets handled as a special case.

locke-lunarg: That's good. Thanks.

USAGE_desktop_D3D12.md
```diff
@@ -215,7 +215,9 @@ Usage:
                         [--dump-resources <submit-index,command-index,draw-call-index>]
                         [--dump-resources-before-draw]
                         [--dump-resources-dir <dir>]
+                        [--dump-resources-modifiable-state-only ]
                         [--pbi-all] [--pbis <index1,index2>] <file>
+                        <file>
```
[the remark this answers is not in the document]
davidlunarg: Fixed.

davidlunarg:
CI failure above is not unexpected - this comment changes the output dump resource json file.

tools/replay/replay_settings.h
```diff
@@ -375,6 +377,18 @@ static void PrintUsage(const char* exe_name)
...
     GFXRECON_WRITE_CONSOLE("          \t\t100 means use all available system and GPU memory");
+    GFXRECON_WRITE_CONSOLE("  --dump-resources <submit-index,command-index,drawcall-index>");
+    GFXRECON_WRITE_CONSOLE("          \t\tOutput binaray resources for a specific drawcall.");
+    GFXRECON_WRITE_CONSOLE("          \t\tInclude vertex, index, const buffer, shader resource, render target,");
+    GFXRECON_WRITE_CONSOLE("          \t\tand depth stencil. And for before and after drawcall.");
+    GFXRECON_WRITE_CONSOLE("          \t\tArguments becomes three indices, submit index, command index,");
+    GFXRECON_WRITE_CONSOLE("          \t\tdrawcall index. The command index is based on its in ExecuteCommandLists.");
+    GFXRECON_WRITE_CONSOLE("  --dump-resources-dir <dir>");
+    GFXRECON_WRITE_CONSOLE("          \t\tDirectory to write dump resources output files.");
+    GFXRECON_WRITE_CONSOLE("          \t\tDefault is the current working directory.");
```
locke-lunarg, line 388: Those two options' descriptions are duplicated.  It has above.

framework/decode/dx12_dump_resources.cpp
```diff
@@ -2092,8 +2095,33 @@ void Dx12DumpResources::CopyResourceAsyncRead(graphics::dx12::ID3D12FenceComPtr
...
+    auto source_resource_object_info = get_object_info_func_(copy_resource_data->source_resource_id);
+    auto source_resource             = reinterpret_cast<ID3D12Resource*>(source_resource_object_info->object);
+    auto source_resource_extra_info  = GetExtraInfo<D3D12ResourceInfo>(source_resource_object_info);
+    std::vector<graphics::dx12::ResourceStateInfo> res_infos = source_resource_extra_info->resource_state_infos;
+
+    // Create a mask of modifiable subresources.
+    // We treat D3D12_RESOURCE_STATE_COMMON as modifiable because that state could be considered
+    // modifiable based on how the resource is used. Note that D3D12_RESOURCE_STATE_COMMON is 0,
+    // so we can't just check for its bit being set
+    uint32_t modifiableResourcesBits = 0;
```
locke-lunarg, line 2107: I'm not sure if subresources count could be more than 32 or not. `uint64_t` might be better. Or it could print an error when the subresources count is more than 32.

davidlunarg: I can change it to 64 bits. I would think that would be enough. I don't think we should print an error. But there is probably no limit. The only other solution would be to change it to a vector. The json output for the value would be a series of integers. Thoughts?

tools/replay/replay_settings.h
davidlunarg: Missed a commit when I cherry picked from another branch. Fixed.

framework/decode/dx12_dump_resources.cpp
```diff
@@ -2563,6 +2592,7 @@ void DefaultDx12DumpResourcesDelegate::WriteResource(nlohmann::ordered_json&   j
     util::FieldToJson(jdata["heap_index"], resource_data->descriptor_heap_index, json_options_);
     util::FieldToJson(jdata["res_id"], resource_data->source_resource_id, json_options_);
     util::FieldToJson(jdata["dimension"], util::ToString(resource_data->desc.Dimension), json_options_);
+    util::FieldToJson(jdata["modifiable_subresources_mask"], subResourceModifiedMask, json_options_);
```
locke-lunarg, line 2594: `subResourceModifiedMask` is a integer to indicate the all subresources that are modifiable or not. I feel it might be hard to read.  I think we could just write a boolean into .json inside `for (const auto sub_index : resource_data->subresource_indices)`.

framework/decode/dx12_dump_resources.cpp
locke-lunarg: A vector is good. `for (const auto sub_index : resource_data->subresource_indices)` is to write every subresource data. It is just below. It could write a boolean or an integer in every subresource data.

locke-lunarg (approved):
LGTM. Thank you for handling this issue.

PR #2009, opened 2025-02-06
Zulauf vulkan codegen  tranche 5

jzulauf-lunarg:
40 commits. Minimal changes to output (one file renamed).


framework/decode/struct_pointer_decoder.h
```diff
@@ -181,6 +181,67 @@ class StructPointerDecoder : public PointerDecoderBase
...
     }
 
+    // When a parent struct pointer was being decoded, determine what child is actually being used
+    // and allocate and decode the appropriate child.
+    size_t DecodeChildren(const uint8_t* buffer, size_t buffer_size)
+    {
+        size_t bytes_read = DecodeAttributes(buffer, buffer_size);
+
+        // We should only be decoding structs.
+        assert((GetAttributeMask() & format::PointerAttributes::kIsStruct) == format::PointerAttributes::kIsStruct);
```
fabian-lunarg, line 191: prefer `GFXRECON_ASSERT` here and some other places close by.
the output is a bit more expressive I think.

framework/generated/khronos_generators/khronos_consumer_header_generator.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/python3 -i
+#
+# Copyright (c) 2018 Valve Corporation
+# Copyright (c) 2018-2024 LunarG, Inc.
```
fabian-lunarg, line 4: 2018-2025

fabian-lunarg (approved):
looks good. 
codegen-codebase seems to become both more tidy and capable.
output has not really changed.
only left minor remarks.

framework/generated/khronos_generators/khronos_consumer_header_generator.py
jzulauf-lunarg: All the commits are from 2024.  While ensure any commit with data > Jan 1 is updated

jzulauf-lunarg:
@MarkY-LunarG @bradgrantham  -- addressed @fabian-lunarg 's comments rebased/testing on dev cleanly

bradgrantham-lunarg (approved):

framework/generated/khronos_generators/khronos_base_generator.py
```diff
@@ -1665,8 +1665,42 @@ def make_simple_var_name(self, type_name):
...
-        Must implement.
-        """
+        """ Method intended to be overridden. """
         raise NotImplementedError
+
+    def generate_child_struct_switch_statement(
+        self, parent_struct, value, initial_indent, switch_expression,
+        fn_emit_default, fn_emit_case
+    ):
+        """ Parent structs are abstract, need to case to specific child struct based on type """
```
MarkY-LunarG, line 1693: Nit: "need to case to specific"  Maybe "need to swithc on specific"?

framework/generated/khronos_generators/khronos_struct_decoders_header_generator.py
```diff
@@ -59,6 +65,150 @@ def endFile(self):
             )
             write(body, file=self.outFile)
 
+        api_data = self.get_api_data()
+        var_name = api_data.type_prefix.lower() + '_type'
+        struct_type = api_data.struct_type_enum
+
+        for struct in self.get_all_filtered_struct_names():
```
MarkY-LunarG, line 54: I think this section needs at least a comment.  Something like "Handle decoding children structs.  When a parent struct is encountered, at runtime we need to determine which actual child struct is referenced and allocate that struct instead of the parent struct."

MarkY-LunarG (approved):

jzulauf-lunarg:
Squashed down to 7 commits of related changes.

PR #2014, opened 2025-02-07
VDR: Initialize multiview flag

roberto-c:
Missing variable initialization was causing multiview structure to be added when creating render passes even when there was not multiview in play.


roberto-c:
Without the multiview variable initialization, a crash could happen because a `VkRenderPassMultiviewCreateInfo ` was being added incorrectly when creating a render pass in `CloneRenderPass` due invalid vaue.

https://github.com/LunarG/gfxreconstruct/blob/7716e6bf9a7108f33a8b21fbe04a53f4663bdaeb/framework/decode/vulkan_replay_dump_resources_draw_calls.cpp#L2144

PR #2016, opened 2025-02-07
Map depth image in Unorm16 for full range

dfriederich:
The mapping from Unorm16 encoded depth images did map half of the range to the max color value, hence causing that the same gray tones were used twice, for values < 0.5 (0x8000 before unorm), or >= 0.5. Fixing the normalization to linearly map the whole range.

### DIFF
--- a/framework/util/image_writer.cpp
+++ b/framework/util/image_writer.cpp
@@ -273,8 +273,8 @@ static const uint8_t* ConvertIntoTemporaryBuffer(uint32_t    width,
                 for (uint32_t x = 0; x < width; ++x)
                 {
                     const uint32_t normalized_depth = bytes_u32[x] & 0x00FFFFFF;
-                    const float    float_depth      = static_cast<float>(normalized_depth) / 8388607.0f;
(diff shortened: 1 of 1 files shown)


dfriederich:
With this fix:
image

dfriederich:
I tried copying how it looked before, but somehow I fail (maybe some company firewall, but it did let me add the image above, odd). Anyhow, before this PR, the middle grays in the image above were dark black, then immediately afterwards some white area. While it is interesting to see the specific depth of 0x8000 so clearly, it did confuse me, I did not understand where this black to white border in some of the part of the image is from.

dfriederich:
(restarted the browser. This is how it looks without this PR)

image

bradgrantham-lunarg:
This failed because of a reference image difference, as expected.  I'll
leave this to Panos for his Monday to approve and merge and update
reference files or I can as well.  -Brad

panos-lunarg (approved):

panos-lunarg:
Merged and updated reference file

PR #2017, opened 2025-02-10
Fix Replay Crash at CreateSwapChainForComposition

JerryAMD:
Gfxr replay redirect CreateSwapChainForComposition to CreateSwapChainForHwnd. But CreateSwapChainForComposition don't have DXGI_SWAP_CHAIN_FULLSCREEN_DESC parameter. So that the replay should handle the case.


locke-lunarg (approved):

locke-lunarg:
It's good to merge.

PR #2018, opened 2025-02-10
Don't check output format for raw format requests

dfriederich:
For a dump image raw request, don't fail if we don't know how to encode the image data, as the raw write is bypassing the write data.


dfriederich:
Closing in lieu of similar fix
https://github.com/LunarG/gfxreconstruct/pull/2020

PR #2021, opened 2025-02-12
Skip unnecessary wait that might cause deadlock

bradgrantham-lunarg:
Skip unnecessary wait that might cause deadlock


bradgrantham-lunarg:
I think but I'm not certain that internal CI build #6053 failed because of commits after this one in `dev` that had associated changes to the "dump resources" reference files.

PR #2025, opened 2025-02-17
Align capture and replay swapchain transformations

panos-lunarg:


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -3302,10 +3302,13 @@ VkResult VulkanReplayConsumerBase::OverrideGetPhysicalDeviceSurfaceCapabilitiesK
     VkSurfaceKHR     surface                     = surface_info->handle;
     auto             replay_surface_capabilities = pSurfaceCapabilities->GetOutputPointer();
 
+    auto meta                                                   = pSurfaceCapabilities->GetMetaStructPointer();
+    surface_info->capture_surface_capabilities[physical_device] = *meta->decoded_value;
```
bradgrantham-lunarg, line 3306: At this line, `*meta->decoded_value` is the **captured** surface capabilities but `physical_device` is the **replay** handle.  This mapped a replay handle to captured data and I was confused (I guess I think capture data to replay data) until I traced this with a debugger.  Can `capture_surface_capabilities` be keyed off the capture PD handleID or `physical_device_info` through the captured surface handleID or something like that?

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_replay_consumer_base.cpp
panos-lunarg: Apologies but I took (yet) another approach for this. The logic of the new change is in `OverrideCreateSwapchain`:
```
#if defined(VK_USE_PLATFORM_ANDROID_KHR)
            // On Android, SetSwapchainWindowSize will take care of non identity transformations by requesting the
            // appropriate orientation with setRequestedOrientation
            modified_create_info.preTransform = VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR;
#endif
```

The changes in `generated_vulkan_replay_consumer.cpp` are still valid (although now kind of unrelated) because the check `if (pSurfaceInfo->GetPointer()->surface == VK_NULL_HANDLE) { return; }` was done before the call to `MapStructHandles` so the surface handle was always null

bradgrantham-lunarg (approved):

PR #2028, opened 2025-02-17
Fix CMake warning CMP0175

mikes-lunarg:
Started getting this warning on CMake 3.31+:

```
CMake Warning (dev) at tools/capture-vulkan/CMakeLists.txt:3 (add_custom_command):
  The following keywords are not supported when using
  add_custom_command(TARGET): DEPENDS.

  Policy CMP0175 is not set: add_custom_command() rejects invalid arguments.
  Run "cmake --help-policy CMP0175" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

From `cmake --help-policy CMP0175`:
> CMake 3.30 and earlier silently ignored unsupported keywords and missing or
> invalid arguments for the different forms of the ``add_custom_command()``
> command. CMake 3.31 implements more rigorous argument checking and will flag
> invalid or missing arguments as errors.

The simplest fix seems to be just adding the copy command to the custom target and getting rid of the custom command.

### DIFF
--- a/tools/capture-vulkan/CMakeLists.txt
+++ b/tools/capture-vulkan/CMakeLists.txt
@@ -1,8 +1,6 @@
-add_custom_target(gfxrecon-capture-vulkan.py ALL)
-
-add_custom_command(TARGET gfxrecon-capture-vulkan.py
-                   DEPENDS ${CMAKE_SOURCE_SOURCE_DIR}/gfxrecon-capture-vulkan.py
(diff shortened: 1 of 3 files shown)


bradgrantham-lunarg:
Did you test this against anything earlier than CMake 3.31?

mikes-lunarg:
> Did you test this against anything earlier than CMake 3.31?

I also tested on 3.22.1 (Ubuntu 22.04 LTS) and 3.16.3 (cmake_minimum_required)

bradgrantham-lunarg (approved):

PR #2030, opened 2025-02-19
Vulkan Dump: Convert call index to block index

locke-lunarg:
It's the same to this PR https://github.com/LunarG/gfxreconstruct/pull/1998, but added a commit to rename `--dump-resources-block-indices` to `--dump-resources`. Since `--dump-resources` checks `dump_resources.find_first_not_of("0123456789,") == std::string::npos`, it could recognize the different between call indices and block indices.


davidlunarg:
Looks good. I provided Locke with some trivial doc changes that were missed in  [PR #1998.](https://github.com/LunarG/gfxreconstruct/pull/1998)

davidlunarg (approved):

PR #2032, opened 2025-02-20
replay: Fix instance info

antonio-lunarg:
Make sure messenger_create_info memory is still valid when it goes out of scope since the modified_create_info holds a pointer to it and it needs it when calling the create instance function.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -2545,7 +2545,6 @@ void VulkanReplayConsumerBase::ModifyCreateInstanceInfo(
     }
 
     // Enable validation layer and create a debug messenger if the enable_validation_layer replay option is set.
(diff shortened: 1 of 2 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2545,7 +2545,7 @@ void VulkanReplayConsumerBase::ModifyCreateInstanceInfo(
     }
 
     // Enable validation layer and create a debug messenger if the enable_validation_layer replay option is set.
-    VkDebugUtilsMessengerCreateInfoEXT messenger_create_info{};
+    static VkDebugUtilsMessengerCreateInfoEXT messenger_create_info{};
```
bradgrantham-lunarg, line 2548: Can this be added to CreateInstanceInfoState?  Then this function doesn't have to concern itself with the lifetime of this struct.

bradgrantham-lunarg (changes requested):

bradgrantham-lunarg (approved):

PR #2033, opened 2025-02-20
Skip Invalid PlacedResource

JerryAMD:
Problem:
Application might release heap before it release PlacedResource.

Solution:
GFXR should discard all invalid resource while dump resource.

### DIFF
--- a/framework/encode/dx12_state_writer.cpp
+++ b/framework/encode/dx12_state_writer.cpp
@@ -568,12 +568,18 @@ void Dx12StateWriter::WriteResourceCreationState(
         assert(resource_wrapper->GetWrappedObject() != nullptr);
         assert(resource_wrapper->GetObjectInfo() != nullptr);
         assert(resource_wrapper->GetObjectInfo()->create_parameters != nullptr);
+        assert(resource_wrapper->GetObjectInfo()->create_object_id != format::kNullHandleId);
 
         auto        resource      = resource_wrapper->GetWrappedObjectAs<ID3D12Resource>();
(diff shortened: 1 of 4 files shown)


framework/encode/dx12_state_writer.cpp
```diff
@@ -1302,6 +1308,20 @@ bool Dx12StateWriter::CheckDescriptorObjects(const DxDescriptorInfo& descriptor_
...
+    {
+        case format::ApiCall_ID3D12Device_CreatePlacedResource:
+        case format::ApiCall_ID3D12Device8_CreatePlacedResource1:
+        case format::ApiCall_ID3D12Device10_CreatePlacedResource2:
+            // Placed resource have to have valid Heap object.
+            return (state_table.GetID3D12Heap_Wrapper(resource_info->heap_id) != nullptr);
+        default:
+            return true;
+    }
+}
```
locke-lunarg, line 1323: Could you please explain why only placed resources need to check heap id? The others don't need. That could help me to understand this PR. Thank you.

JerryAMD: This func is for checking if resource object valid or not. But currently, it only check PleacedResource. We could add more checking if found other resource invalid. For PleaceResource, it have to be created on top of Heap, if the Heap was released, then the PlacedResource was invalid and won't be used anymore. GFXR should discard the resource during tracking.

locke-lunarg: Thank you for answering.

locke-lunarg (approved):

PR #2036, opened 2025-02-21
Zulauf vulkan codegen tranche 6

jzulauf-lunarg:
All of the core refactor commits for Vulkan, with cleanup commits supporting 23 of 33 OpenXR files successfully generating.

Expect one more small set of changes to account for the last few OpenXR porting support issues.


fabian-lunarg:
reason for fail: clang-format check reports 3 locations in `framework/encode/vulkan_state_writer.cpp`

bradgrantham-lunarg:
I think it's possible the header update has caused some small generation differences, those showed up in the Ubuntu Release CI build on GitHub. With those resolved I thikn this is ready to go.

jzulauf-lunarg:
@MarkY-LunarG  just awaiting your +1

framework/generated/khronos_generators/khronos_base_generator.py
```diff
@@ -196,6 +200,8 @@ def __init__(
     ):
         self.api_name = api_name
         self.api_class_prefix = api_class_prefix
+        self.camel_guard = camel_guard
```
MarkY-LunarG, line 203: Is `camel_guard` necessary?  It's the same as `api_class_prefix`.

MarkY-LunarG (approved):

framework/generated/khronos_generators/khronos_base_generator.py
jzulauf-lunarg: Not always.

jzulauf-lunarg:
Semi-squash pushed.  Identical to unsquashed.

PR #2040, opened 2025-02-24
Track & Sanitize usage of buffer references in pipelines

fabian-lunarg:
The goal of this PR is to allow usage of buffer-references in shaders (`SPV_KHR_physical_storage_buffer`)
when using the `-m rebind` option.

This aligns with previous work for portable raytracing and follows identical principles.

We already parsed spirv shader-code for usage of SPV_KHR_physical_storage_buffer.
here we start using the extracted offsets and locations and derive information about actual buffers
and buffer-addresses during runtime, when the pipeline is used.

We basically collect an array of gpu-addresses, containing locations of stale addresses. 
Those can be replaced with an injected compute-dispatch.
In case this is not possible (inside a renderpass), the replacer/sanitizer-pass is postponed until queue-submit.
(and syncing via fence there, 'could' be done via semaphore)

using this branch, I'm testing successful replay of various cases which would previously crash.

e.g.:
https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/extensions/ray_tracing_reflection

also contains:
- improved spirv-parser with cap-check, forward spirv-reflect pass, then our backward custom parse.
- unit test for `SpirVParsingUtil`
- fixed holes in buffer/acceleration-structure tracking
- use `vkSetDebugUtilsObjectNameEXT` to annotate names of shadow-resources in VulkanAddressReplacer
- switch internal hash-function from murmur32 to xxhash32. less code and faster, same results for our purpose


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -10146,18 +10150,26 @@ VkResult VulkanReplayConsumerBase::OverrideCreateComputePipelines(
...
-            assert(pipeline_info);
+            if (create_info_meta[i].stage->module != gfxrecon::format::kNullHandleId)
+            {
+                VulkanShaderModuleInfo* module_info =
+                    object_info_table_->GetVkShaderModuleInfo(create_info_meta[i].stage->module);
+                assert(module_info);
 
-            pipeline_info->shaders.insert({ VK_SHADER_STAGE_COMPUTE_BIT, *module_info });
+                auto pipeline_info = reinterpret_cast<VulkanPipelineInfo*>(pPipelines->GetConsumerData(i));
+                assert(pipeline_info);
```
MarkY-LunarG: GFXRECON_ASSERT

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -10146,18 +10150,26 @@ VkResult VulkanReplayConsumerBase::OverrideCreateComputePipelines(
...
+
+                pipeline_info->shaders.insert({ VK_SHADER_STAGE_COMPUTE_BIT, *module_info });
+            }
+            else if (auto module_create_info =
+                         graphics::vulkan_struct_get_pnext<VkShaderModuleCreateInfo>(&in_p_create_infos->stage))
+            {
+                // TODO: handle inlined SPIRV here, it should be in the pNext-chain of VkPipelineShaderStageCreateInfo
```
MarkY-LunarG: Log a warning here that this is unhandled?

framework/graphics/vulkan_check_buffer_references.cpp
```diff
@@ -27,29 +27,41 @@ GFXRECON_BEGIN_NAMESPACE(gfxrecon)
...
     for (uint32_t i = 0; i < create_info_count; ++i)
     {
         if (auto module_create_info = vulkan_struct_get_pnext<VkShaderModuleCreateInfo>(&create_infos[i].stage))
         {
-            graphics::vulkan_check_buffer_references(module_create_info->pCode, module_create_info->codeSize);
+            graphics::vulkan_check_buffer_references(
+                module_create_info->pCode, module_create_info->codeSize, out_info_structs[i]);
         }
     }
 }
```
MarkY-LunarG, line 173: Since the above is a template, why not just remove the VkGraphicsPipelineCreateInfo and have a template parameter?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -9703,9 +9703,7 @@ VulkanAddressReplacer& VulkanReplayConsumerBase::GetDeviceAddressReplacer(const
 
 bool VulkanReplayConsumerBase::UseExtraDescriptorInfo(const VulkanDeviceInfo* device_info) const
 {
-    // TODO: testing with an override here
-    return true;
-//    return options_.dumping_resources || !device_info->allocator->SupportsOpaqueDeviceAddresses();
+    return options_.dumping_resources || !device_info->allocator->SupportsOpaqueDeviceAddresses();
```
MarkY-LunarG: This is specific to the general "opaque address support"?  Or does it look at all of the feature flags?  My concern is that it might be present for one case, but not all the cases you use this check to cover.

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -1451,7 +1465,8 @@ void VulkanAddressReplacer::run_compute_replace(const VulkanCommandBufferInfo*
 
     auto& pipeline_context_bda = pipeline_context_map_[command_buffer_info->handle];
 
-    if (!create_buffer(pipeline_context_bda.hashmap_storage, hashmap_bda_.get_storage(nullptr)))
+    if (!create_buffer(
+            pipeline_context_bda.hashmap_storage, hashmap_bda_.get_storage(nullptr), 0, 0, true, "hashmap_storage_bda"))
```
MarkY-LunarG, line 1624: Should our buffer names start with "gfxr_" to identify them differently than user buffers on debugging?

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -306,6 +306,40 @@ void VulkanAddressReplacer::UpdateBufferAddresses(const VulkanCommandBufferInfo*
...
+                // find addresses in push-constant memory and replace in-place.
+                auto* address = reinterpret_cast<VkDeviceAddress*>(static_cast<uint8_t*>(data) + offset +
+                                                                   buffer_ref_info.buffer_offset);
+
+                auto* buffer_info = address_tracker.GetBufferByCaptureDeviceAddress(*address);
+                GFXRECON_ASSERT(buffer_info != nullptr);
+                if (buffer_info != nullptr && buffer_info->replay_address != 0)
+                {
+                    uint32_t address_offset = *address - buffer_info->capture_address;
+                    *address                = buffer_info->replay_address + address_offset;
```
MarkY-LunarG, line 336: Should we log a "debug" level message that these have been replaced?

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -1367,7 +1455,7 @@ bool VulkanAddressReplacer::create_buffer(VulkanAddressReplacer::buffer_context_
         VkDebugUtilsObjectNameInfoEXT object_name_info = {};
         object_name_info.sType                         = VK_STRUCTURE_TYPE_DEBUG_UTILS_OBJECT_NAME_INFO_EXT;
         object_name_info.objectType                    = VK_OBJECT_TYPE_BUFFER;
-        object_name_info.objectHandle                  = reinterpret_cast<uint64_t>(buffer_context.buffer);
+        object_name_info.objectHandle                  = (uint64_t)buffer_context.buffer;
```
MarkY-LunarG, line 1458: Why did you change this?  Should it be a static_cast instead?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -9576,6 +9750,16 @@ VulkanAddressReplacer& VulkanReplayConsumerBase::GetDeviceAddressReplacer(const
...
 }
 
+bool VulkanReplayConsumerBase::UseExtraDescriptorInfo(const VulkanDeviceInfo* device_info) const
+{
+    return options_.dumping_resources || UseAddressReplacement(device_info);
+}
+
+bool VulkanReplayConsumerBase::UseAddressReplacement(const VulkanDeviceInfo* device_info) const
+{
+    return !device_info->allocator->SupportsOpaqueDeviceAddresses();
```
MarkY-LunarG, line 9760: Ok, in case the previous comment got deleted, my concern is that should we check this per situation?  For example, the ray tracing opaque address support might not be present, but the other standard one is, but we never use ray tracing in the current app so it doesn't matter.

framework/util/spirv_parsing_util.cpp
```diff
@@ -356,14 +471,14 @@ bool SpirVParsingUtil::ParseBufferReferences(const uint32_t* const spirv_code, s
...
-                                        else if (member.op == SpvOpTypeForwardPointer)
+                                        else if (member.op == SpvOpTypePointer || member.op == SpvOpTypeForwardPointer)
                                         {
                                             num_scalar_bytes = sizeof(uint64_t);
                                         }
                                         else if (member.op == SpvOpTypeArray || member.op == SpvOpTypeRuntimeArray)
                                         {
                                             num_scalar_bytes = std::max(num_scalar_bytes, member.traits.array.stride);
-                                            assert(false); // not handled
+                                            //                                            assert(false); // not handled
```
MarkY-LunarG, line 481: Remove this?

framework/graphics/vulkan_check_buffer_references.cpp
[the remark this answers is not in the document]
MarkY-LunarG: You did, so killing this.

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: done

framework/decode/vulkan_address_replacer.cpp
MarkY-LunarG: Hmm... I don't know.  Maybe @mikes-lunarg or @bradgrantham  can help.  Otherwise, I would have to do a 2 step where you reinterpret_cast to a uint64_t* and then a static_cast to a uint64_t.

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: I added a `GFXRECON_LOG_INFO_ONCE("VulkanAddressReplacer::ProcessCmdPushConstants(): Replay is adjusting mismatching buffer-device-addresses ..."` - analog to the other entry-points that 'might' apply some replacement of resources.

MarkY-LunarG: That sounds good. Thanks!

framework/decode/vulkan_address_replacer.cpp
bradgrantham-lunarg: `static_cast `is only valid between "related types" which includes same width.  `reinterpret_cast `is identical to C-style cast (https://learn.microsoft.com/en-us/cpp/cppcx/casting-c-cx?view=msvc-170) . I think rather than `uint64_t` it should be `uintptr_t` ; there's some weirdness on 32-bit where casting to a uint64_t sign-extends.  (undefined behavior?) : https://godbolt.org/z/M3YK9W864

So I think `reinterpret_cast<uintptr_t>()` - https://github.com/LunarG/gfxreconstruct/pull/2046

bradgrantham-lunarg: I spent WAY more time on that than it warranted but now I guess I know better.

fabian-lunarg:
```vulkan_address_replacer.cpp:1485:58: error: invalid cast from type âVkBufferâ {aka âlong long unsigned intâ} to type âuintptr_tâ {aka âunsigned intâ}
 1485 |         object_name_info.objectHandle                  = reinterpret_cast<uintptr_t>(buffer_context.buffer);```

MarkY-LunarG: I know we've done this before.  I just don't remember where....

bradgrantham-lunarg: That's fun - it compiles in 32-bit on Compiler Explorer.  So what's different between CE and our compilation?

fabian-lunarg: I refactored our existing vulkan handle-casting macros into `graphics/vulkan_util.h` and
that's used now: `VK_HANDLE_TO_UINT64(buffer_context.buffer)` - thanks for pointing me at those!

MarkY-LunarG (approved):

PR #2044, opened 2025-02-25
Add waitForFence before destroying virtual swap chain resources

zongdu-arm:
The commandbuffer of virtual swapchain may not be fully executed. Add waitForFence before destroying it.

### DIFF
--- a/framework/decode/vulkan_virtual_swapchain.cpp
+++ b/framework/decode/vulkan_virtual_swapchain.cpp
@@ -106,23 +106,25 @@ void VulkanVirtualSwapchain::CleanSwapchainResourceData(const VulkanDeviceInfo* 
         auto allocator = device_info->allocator.get();
         assert(allocator != nullptr);
 
-        for (const VulkanImageInfo& image_info : swapchain_info->image_infos)
-        {
-            allocator->DestroyImageDirect(image_info.handle, nullptr, image_info.allocator_data);
-            allocator->FreeMemoryDirect(image_info.memory, nullptr, image_info.memory_allocator_data);
-        }
-
         // Delete the virtual swapchain-specific swapchain resource data
         if (swapchain_resources_.find(swapchain) != swapchain_resources_.end())
         {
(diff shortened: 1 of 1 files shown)


framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -106,16 +106,18 @@ void VulkanVirtualSwapchain::CleanSwapchainResourceData(const VulkanDeviceInfo*
...
-        {
-            allocator->DestroyImageDirect(image_info.handle, nullptr, image_info.allocator_data);
-            allocator->FreeMemoryDirect(image_info.memory, nullptr, image_info.memory_allocator_data);
-        }
-
         // Delete the virtual swapchain-specific swapchain resource data
         if (swapchain_resources_.find(swapchain) != swapchain_resources_.end())
         {
             auto& swapchain_resources = swapchain_resources_[swapchain];
+            for (auto& copy_cmd_data : swapchain_resources->copy_cmd_data)
```
bradgrantham-lunarg, line 113: I think this `copy_cmd_data` and the `fence` below can be made `const`

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_virtual_swapchain.cpp
zongdu-arm: OK, but I don't think the `fence` is needed.

bradgrantham-lunarg: I think it's less about whether its needed than about making a statement to the reader that you don't intend in this context to change the loop variable.  `fence` doesn't really need to be a reference either since it's a single handle so you could probably just do `const fence : `

bradgrantham-lunarg:
Thank you!  This is a really helpful PR!

framework/decode/vulkan_virtual_swapchain.cpp
zongdu-arm: I updated all of this.

bradgrantham-lunarg (approved):

PR #2048, opened 2025-02-25
Correct code

locke-lunarg:
This change doesn't affect anything. The original code also got the correct result, but the code wasn't correct.

### DIFF
--- a/framework/decode/dx12_dump_resources.cpp
+++ b/framework/decode/dx12_dump_resources.cpp
@@ -1043,9 +1043,8 @@ bool CaptureGPUAddrMatchDescriptorHeap(const D3D12_GPU_DESCRIPTOR_HANDLE capture
     auto increment            = (*heap_info.capture_increments)[heap_info.descriptor_type];
     auto capture_gpu_addr_end = heap_info.capture_gpu_addr_begin + heap_info.descriptor_count * increment;
 
-    bool is_match = true ? (heap_info.capture_gpu_addr_begin <= capture_gpu_addr.ptr &&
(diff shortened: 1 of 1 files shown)


framework/decode/dx12_dump_resources.cpp
```diff
@@ -1043,9 +1043,10 @@ bool CaptureGPUAddrMatchDescriptorHeap(const D3D12_GPU_DESCRIPTOR_HANDLE capture
     auto increment            = (*heap_info.capture_increments)[heap_info.descriptor_type];
     auto capture_gpu_addr_end = heap_info.capture_gpu_addr_begin + heap_info.descriptor_count * increment;
 
-    bool is_match = true ? (heap_info.capture_gpu_addr_begin <= capture_gpu_addr.ptr &&
-                            capture_gpu_addr.ptr <= capture_gpu_addr_end)
-                         : false;
+    bool is_match =
```
bradgrantham-lunarg, line 1046: Can you just use `bool is_match = heap_info.capture_gpu_addr_begin <= capture_gpu_addr.ptr && capture_gpu_addr.ptr <= capture_gpu_addr_end;`?

(same below)

bradgrantham-lunarg (changes requested):

framework/decode/dx12_dump_resources.cpp
locke-lunarg: Thank you. Fixed.

bradgrantham-lunarg (approved):

PR #2051, opened 2025-02-27
VDR: Fix in vertex buffer offset calculation

panos-lunarg:
Min max vertex indices should be initialized to zero when for non indexed draw calls.

### DIFF
--- a/framework/decode/vulkan_replay_dump_resources_common.cpp
+++ b/framework/decode/vulkan_replay_dump_resources_common.cpp
@@ -309,11 +309,11 @@ uint32_t VkIndexTypeToBytes(VkIndexType type)
     }
 }
 
-std::pair<uint32_t, uint32_t> FindMinMaxVertexIndices(const std::vector<uint8_t>& index_data,
-                                                      uint32_t                    index_count,
-                                                      uint32_t                    first_index,
-                                                      int32_t                     vertex_offset,
-                                                      VkIndexType                 type)
+MinMaxVertexIndex FindMinMaxVertexIndices(const std::vector<uint8_t>& index_data,
+                                          uint32_t                    index_count,
+                                          uint32_t                    first_index,
+                                          int32_t                     vertex_offset,
+                                          VkIndexType                 type)
 {
     switch (type)
     {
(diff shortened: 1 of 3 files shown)


dfriederich:
Works for me :-)

framework/decode/vulkan_replay_dump_resources_common.cpp
```diff
@@ -329,7 +329,8 @@ std::pair<uint32_t, uint32_t> FindMinMaxVertexIndices(const std::vector<uint8_t>
 
             if (i == index_count)
             {
-                return std::make_pair(0, 0);
+                MinMaxVertexIndex min_max_vertex_indices = MinMaxVertexIndex{ 0, 0 };
```
dfriederich, line 332: nit, I would write this as using designated initializer names

return {.min=0, .max=0};

could of course also just use return {} (and relay on the default initializer values.

dfriederich: Just to be a little bit less verbose, and with designated initializers it is also explicit what fields are being initialized with what, I like to be explicit :-). But just a nit, it does work fine as it is.

panos-lunarg: Done

panos-lunarg: MSVC builds failed with `use of designated initializers requires at least '/std:c++20'`. We support 17.
Will switch back to previous notation

dfriederich: :-(

dfriederich (approved):

dfriederich:
@bradgrantham can you review? Can we merge this to dev?

framework/decode/vulkan_replay_dump_resources_common.cpp
```diff
@@ -370,7 +370,7 @@ std::pair<uint32_t, uint32_t> FindMinMaxVertexIndices(const std::vector<uint8_t>
 
             if (i == index_count)
             {
-                return std::make_pair(0, 0);
+                return MinMaxVertexIndex{ 0, 0 };
```
davidlunarg, line 373: Initializers are redundant, but  leave as is for readability.

davidlunarg (approved):

PR #2054, opened 2025-03-05
Fix issues around async pipeline-handle assignment/cleanup

fabian-lunarg:
changes to pipeline handle-synchronization when using async compilation.

- fixes issues in combination with `-m rebind` option and address-replacement
- fixes validation-errors for pipeline-handle leaks when using asynchronous pipeline-compilation (`--pcj X`)
- found and removed some dead code
- added debug-labels to internal VkPipelines in VulkanAddressReplacer during debugging

to reproduce the reported issue with address-replacement this sample can be used:
https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/extensions/ray_tracing_reflection
`gfxrecon-replay -m rebind --pcj -1 ray_tracing_reflection.gfxr`

### DIFF
--- a/framework/decode/vulkan_object_info_table_base.h
+++ b/framework/decode/vulkan_object_info_table_base.h
@@ -50,6 +50,23 @@ struct has_handle_future<T, decltype((void)T::future, 0)>
 template <typename T>
 inline constexpr bool has_handle_future_v = has_handle_future<T>::value;
 
+/**
+ * @brief   sync_handle will block and wait for pending handle-creation operations, then populate the object's handle.
+ *
+ * @tparam  T       object-type
+ * @param   object  a provided pointer to an asynchronous object-wrapper
+ */
+template <typename T>
+static inline void sync_handle(T* object)
+{
+    static_assert(has_handle_future_v<T>);
+    if (object != nullptr && object->handle == VK_NULL_HANDLE && object->future.valid())
+    {
+        const auto& [result, async_handles] = object->future.get();
+        object->handle                      = async_handles[object->future_handle_index];
+    }
+}
+
 // NOTE: There's nothing VulkanSpecific in these utilities
 // TODO: Find a better home for these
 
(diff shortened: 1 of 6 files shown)


framework/decode/vulkan_object_info_table_base.h
```diff
@@ -50,6 +50,23 @@ struct has_handle_future<T, decltype((void)T::future, 0)>
...
+ * @brief   sync_handle will block and wait for pending handle-creation operations, then populate the object's handle.
+ *
+ * @tparam  T       object-type
+ * @param   object  a provided pointer to an asynchronous object-wrapper
+ */
+template <typename T>
+static inline void sync_handle(T* object)
+{
+    static_assert(has_handle_future_v<T>);
+    if (object != nullptr && object->handle == VK_NULL_HANDLE && object->future.valid())
```
MarkY-LunarG, line 63: So, I don't know this code, but how does it perform the synchronize?  Is it all in the "object->future.valid()" call?

fabian-lunarg:
> What is the test case for this as well? I know you don't have one, but it would be nice to at least list what application showed this problem and how you tested.

we do have an internal test-case (combining `--pipeline-compile-jobs -1 -m rebind`) in our CI,
at least for leaked handles during cleanup.

to reproduce the reported issue with address-replacement this sample can be used:
https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/extensions/ray_tracing_reflection
again using `--pipeline-compile-jobs -1 -m rebind`

lastly, I'm often also testing with my `pbr_viewer` from here https://github.com/crocdialer/vierkant_projects
(there are nightlies for linux&windows).

framework/decode/vulkan_object_info_table_base.h
fabian-lunarg: the blocking call is: `object->future.get()` (future is a `std::shared_future`)

MarkY-LunarG:
> > What is the test case for this as well? I know you don't have one, but it would be nice to at least list what application showed this problem and how you tested.
> 
> we do have an internal test-case (combining `--pipeline-compile-jobs -1 -m rebind`) in our CI, at least for leaked handles during cleanup.
> 
> to reproduce the reported issue with address-replacement this sample can be used: https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/extensions/ray_tracing_reflection again using `--pipeline-compile-jobs -1 -m rebind`
> 
> lastly, I'm often also testing with my `pbr_viewer` from here https://github.com/crocdialer/vierkant_projects (there are nightlies for linux&windows).

Cool.  Can you modify the last commit with some of that info?  That's all I think is necessary.  Otherwise, LGTM.

PR #2056, opened 2025-03-05
Fix vertexInputState in output json

panos-lunarg:


davidlunarg (approved):
Looks good. Took me a while to figure out why the change.

PR #2057, opened 2025-03-06
Use unique_ptr for resource_dumper_

locke-lunarg:
It has a lot raw pointers that are used. Should we schedule to modify them?

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.h
+++ b/framework/decode/vulkan_replay_consumer_base.h
@@ -1457,7 +1457,7 @@ class VulkanReplayConsumerBase : public VulkanConsumer
 
     const VulkanReplayOptions& options_;
 
-    VulkanReplayDumpResources* resource_dumper_{ nullptr };
+    std::unique_ptr<VulkanReplayDumpResources> resource_dumper_;
 
   private:
     void RaiseFatalError(const char* message) const;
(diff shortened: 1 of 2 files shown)


panos-lunarg:
> It has a lot raw pointers that are used. Should we schedule to modify them?

Which pointers are you referring to?

locke-lunarg:
> > It has a lot raw pointers that are used. Should we schedule to modify them?
> 
> Which pointers are you referring to?

Sorry. I didn't look deep.  Like I saw  `CommonObjectInfoTable* object_info_table_;` in `VulkanReplayConsumerBase`. But it actually comes from `object_info_table_ = CommonObjectInfoTable::GetSingleton();`. Please forget it.

panos-lunarg (approved):

framework/decode/vulkan_replay_consumer_base.h
```diff
@@ -1457,7 +1457,7 @@ class VulkanReplayConsumerBase : public VulkanConsumer
 
     const VulkanReplayOptions& options_;
 
-    VulkanReplayDumpResources* resource_dumper_{ nullptr };
+    std::unique_ptr<VulkanReplayDumpResources> resource_dumper_{ nullptr };
```
bradgrantham-lunarg, line 1460: default for a `unique_ptr` is no pointer, e.g. nullptr, so you don't need an initializer here

bradgrantham-lunarg (approved):
LGTM but remove that nullptr initializer.

bradgrantham-lunarg (approved):

PR #2060, opened 2025-03-06
Do not track contents of AccelerationStructure scratch-buffers

fabian-lunarg:
during capture we store buffer-contents of last used input-buffers (triangle/aabb/instance/transform) to `vkCmdBuildAccelerationStructures`. 

we also store contents of scratch-buffers, which is not necessary and removed here.

### DIFF
--- a/framework/encode/vulkan_state_tracker.cpp
+++ b/framework/encode/vulkan_state_tracker.cpp
@@ -423,6 +423,9 @@ void VulkanStateTracker::TrackAccelerationStructureBuildCommand(
     auto cmd_buf_wrapper = vulkan_wrappers::GetWrapper<vulkan_wrappers::CommandBufferWrapper>(command_buffer);
     auto device_wrapper  = cmd_buf_wrapper->parent_pool->device;
 
+    // scratch space for buffer-addresses. we extract and keep track of associated buffers.
+    std::vector<VkDeviceAddress> to_extract;
+
     for (uint32_t i = 0; i < info_count; ++i)
     {
         const VkAccelerationStructureBuildGeometryInfoKHR& build_info = p_infos[i];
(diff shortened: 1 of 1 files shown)


fabian-lunarg:
> CI gfxreconstruct build # 6267 failed.

Due to some internal CI issue, will be back

framework/encode/vulkan_state_tracker.cpp
```diff
@@ -423,6 +423,9 @@ void VulkanStateTracker::TrackAccelerationStructureBuildCommand(
     auto cmd_buf_wrapper = vulkan_wrappers::GetWrapper<vulkan_wrappers::CommandBufferWrapper>(command_buffer);
     auto device_wrapper  = cmd_buf_wrapper->parent_pool->device;
 
+    // scratch space for buffer-addresses. we extract and keep track of associated buffers.
+    std::vector<VkDeviceAddress> to_extract;
```
panos-lunarg, line 427: So the way it was before that this vector was being reset in each for iteration was wrong?

fabian-lunarg: it's still being cleared at the top of loop (so a unrelated minor refactoring). 
the relevant change is to not init with the scratch-buffer address:

std::vector<VkDeviceAddress> to_extract ~~= { build_info.scratchData.deviceAddress }~~;

also added a nullptr check for good measure.

fabian-lunarg: but then you're right, I think it was incorrect before, but that didn't cause any issues. 
there were duplicated map-insertions of same entries, which we now avoid.

panos-lunarg (approved):

framework/encode/vulkan_state_tracker.cpp
bradgrantham-lunarg, line 427: On the principle of declaring the variable closer to its use, I think inside the loop is fine unless this loop is a per-frame critical path. (Is it?)  Otherwise a reader (like myself) may wonder "why does this variable persist between loop iterations?" and the answer is that it doesn't persist, so it could be confusing.

If this is a bleed-through of an edit you've made in a branch targeting a later PR, I suggest the change elevating this variable out of the loop be limited to that branch/PR.

bradgrantham-lunarg (approved):
LGTM but I would put `to_extract` back into the loop unless there's a compelling reason to elevate it.

framework/encode/vulkan_state_tracker.cpp
fabian-lunarg: agreed, I'll adjust that to make it clearer.

PR #2064, opened 2025-03-10
Optimizer: Mark all buffers targets of GetBufferDeviceAddress as used

panos-lunarg:
Device addresses can end up in places where the optimizer cannot (currently) detect and as a result can optimize out buffers that are actually used by an application.

### DIFF
--- a/framework/decode/vulkan_referenced_resource_consumer_base.cpp
+++ b/framework/decode/vulkan_referenced_resource_consumer_base.cpp
@@ -1158,5 +1158,39 @@ void VulkanReferencedResourceConsumerBase::PushDescriptorSetWithTemplate(format:
     }
 }
 
+void VulkanReferencedResourceConsumerBase::Process_vkGetBufferDeviceAddress(
+    const ApiCallInfo&                                       call_info,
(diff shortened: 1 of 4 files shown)


framework/decode/vulkan_referenced_resource_consumer_base.cpp
```diff
@@ -1158,5 +1158,21 @@ void VulkanReferencedResourceConsumerBase::PushDescriptorSetWithTemplate(format:
     }
 }
 
+void VulkanReferencedResourceConsumerBase::Process_vkGetBufferDeviceAddress(
```
bartosz-muszarski-arm, line 1161: I think the same processing should be added to Process_vkGetBufferDeviceAddressKHR

fabian-lunarg (approved):

framework/decode/vulkan_referenced_resource_consumer_base.cpp
fabian-lunarg: yes right. there is also an EXT flavor ...

bradgrantham-lunarg (approved):

PR #2065, opened 2025-03-10
Multi-Window Replay as secondary app

MarkY-LunarG:
Apply the multi-window replay application from @qiq-2022 to the "dev" tree, but as a secondary replay application instead of replacing the existing replay  application.

The reasoning behind this is temporary: the multi-window replay app does not properly handle surface rotations on pre-Android 14 devices.  Since we still test such devices in our CI, and end-users are still using pre-Android 14 OS versions, we will support both until Android 14 overtakes the rest of our testing.

For now, you may use the `gfxrecon.py` script with the argument `replay` to trigger the old replay application and `multiwin-replay` to trigger the new multi-window replay application.


MarkY-LunarG:
I've extended this with #1912 to include the option to replay using either the old application or your new one.  I just have to wait for review, but I will discuss this with @per-mathisen-arm and @bartosz-muszarski-arm in our next meeting.

framework/application/android_context.cpp
```diff
@@ -108,5 +108,47 @@ void AndroidContext::SetOrientation(ScreenOrientation orientation)
     }
 }
 
+void AndroidContext::requestNativeWindow(int width, int height)
+{
+#ifdef GFXR_MULTI_WINDOW_REPLAY
```
antonio-lunarg, line 113: It looks good to me. I just wonder if two apps are necessary. Can we use gfxrecon replay app Android settings here to request the multi-window functionality?

antonio-lunarg (approved):

PR #2072, opened 2025-03-14
ci: Use clang-format-14 in github actions

charles-lunarg:


bradgrantham-lunarg:
Should `CONTRIBUTING.md` also be updated?

charles-lunarg:
Ah yes, that is a good catch! I've pushed an update to CONTRIBUTING.md to use clang-format-14 where it makes sense.

.github/workflows/ci_build.yml
```diff
@@ -41,15 +41,15 @@ jobs:
...
+        sudo apt-get install -y libx11-xcb-dev libxcb-keysyms1-dev libwayland-dev libxrandr-dev liblz4-dev libzstd-dev clang-format-14
     - uses: lukka/get-cmake@latest
       with:
         cmakeVersion: 3.24
     - name: Run build script
       run: |
         if [ "${{ matrix.config.type }}" == "release" ]
         then
-          clang-format --version # A check for version 9 should be added.
+          clang-format-14 --version # A check for version 9 should be added.
```
panos-lunarg, line 52: The comment can be removed

CONTRIBUTING.md
```diff
@@ -215,7 +215,7 @@ style defined by the
 
 C++ Code formatting is managed with a custom ClangFormat configuration file.
 This is the `.clang-format` file found at the base of the repo tree.
-It is intended for use with **ClangFormat version 9.0.1** (see
+It is intended for use with **ClangFormat version 14.0.1** (see
```
panos-lunarg, line 218: The run in this commit actually used `14.0.6`
Screenshot from 2025-03-15 08-12-25

charles-lunarg: I think specifying the patch version is a bit excessive and probably only going to be out of date as the patch version gets incremented.

charles-lunarg:
Something to note - this makes clang-format-14 a requirement when using the python build script. Meaning this PR will flush out people using a different clang-format version. I think that's good, but wanted to be clear that it'll happen.

mikes-lunarg (approved):

PR #2077, opened 2025-03-17
VDR: For loop was mistakenly nested

panos-lunarg:
OverrideUpdateDescriptorSets was nesting the descriptor count foor loop twice by mistake, causing the function to take exceptionally long for descriptor updates that updated large arrays of descriptors.


fabian-lunarg (approved):
very nice find, that's cut quite some redundant work for large sets.

bradgrantham-lunarg (approved):

PR #2079, opened 2025-03-18
LoadData was being called twice from InitializeImage

panos-lunarg:
AcquireInitializedStagingBuffer will call LoadData so InitializeImage does not need to explicitly call it again.

### DIFF
--- a/framework/decode/vulkan_resource_initializer.cpp
+++ b/framework/decode/vulkan_resource_initializer.cpp
@@ -181,52 +181,45 @@ VkResult VulkanResourceInitializer::InitializeImage(VkDeviceSize             dat
 
     if (result == VK_SUCCESS)
     {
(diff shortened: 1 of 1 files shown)


framework/decode/vulkan_resource_initializer.cpp
```diff
@@ -181,8 +181,6 @@ VkResult VulkanResourceInitializer::InitializeImage(VkDeviceSize             dat
 
     if (result == VK_SUCCESS)
     {
-        result = LoadData(data_size, data, staging_buffer_data);
-
         if (result == VK_SUCCESS)
```
antonio-lunarg, line 184: I think you can remove the inner `if`.

panos-lunarg: right

panos-lunarg:
This is included in #2081

PR #2082, opened 2025-03-19
Final codegen cleanups for OpenXr support

jzulauf-lunarg:
JSON Child type support
Fix decode node and default values
Fix multi-api next en|decode
Clean up mappers and related types
type util fix for multi-api
Only a comment and whitespace change to code.

### DIFF
--- a/framework/generated/khronos_generators/khronos_base_generator.py
+++ b/framework/generated/khronos_generators/khronos_base_generator.py
@@ -42,8 +42,8 @@ import os
 import re
 import sys
 import json
+import copy
 from collections import OrderedDict
-from copy import deepcopy
 from generator import GeneratorOptions, OutputGenerator, noneStr, regSortFeatures
 
 def write(*args, **kwargs):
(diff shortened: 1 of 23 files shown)


framework/generated/khronos_generators/khronos_base_generator.py
```diff
@@ -1915,25 +1944,6 @@ def is_core_destroy_command(self, command_name):
...
-                type_with_prefix = api_data.struct_type_prefix + upper_type
-
-        type_with_prefix = type_with_prefix.replace('_OPEN_GLES', '_OPENGL_ES_')
-        type_with_prefix = type_with_prefix.replace('_OPEN_GL', '_OPENGL_')
-        type_with_prefix = type_with_prefix.replace('_D3_D11', '_D3D11')
-        type_with_prefix = type_with_prefix.replace('_D3_D12', '_D3D12')
-        type_with_prefix = type_with_prefix.replace('_EGL', '_EGL_')
-        type_with_prefix = type_with_prefix.replace('Device_IDProp', 'Device_ID_Prop')
-        return type_with_prefix
-
```
MarkY-LunarG, line 1936: You removed this block, but what fixed it so that it was no longer required?  Is it now OpenXR specific?

MarkY-LunarG (approved):

framework/generated/khronos_generators/khronos_base_generator.py
bradgrantham-lunarg: It looks to me like this was a function to mangle a CamelCase struct name to "screaming snake case" sType.  It looks like sType is now fetched from the XML for the struct.  I think that's handled by the change to the call site for `generate_structure_stype` which now just uses the `values` field, which in the spec looks like it is now always the `sType`.  I'm not sure when this changed but the original code Dustin checked in for this was from 2018 (8a9edb103e), so it's possible it's just that old and the registry has been fixed since then.

bradgrantham-lunarg (approved):

PR #2083, opened 2025-03-20
Filter resources with pipeline layout

panos-lunarg:


framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
```diff
@@ -1089,30 +1091,198 @@ bool DispatchTraceRaysDumpingContext::IsRecording() const
...
-        const uint32_t desc_set_index = desc_set.first;
-        for (const auto& desc : desc_set.second.descriptors)
+        for (const auto& desc_set : bound_descriptor_sets_compute)
         {
-            const uint32_t desc_binding_index                                    = desc.first;
-            tr_params.referenced_descriptors[desc_set_index][desc_binding_index] = desc.second;
+            const uint32_t desc_set_index = desc_set.first;
+
+            // Check against pipeline layout
+            if (bound_pipeline_trace_rays->desc_set_layouts.size() <= desc_set_index)
```
bradgrantham-lunarg, line 1203: Under what circumstances can this happen?  Is binding a set out of range of the available sets invalid usage?

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
```diff
@@ -1140,18 +1310,34 @@ VkResult DispatchTraceRaysDumpingContext::DumpImmutableDescriptors(uint64_t qs_i
...
+    else
+    {
+        const auto tr_params_entry = trace_rays_params.find(cmd_index);
+        GFXRECON_ASSERT(tr_params_entry != trace_rays_params.end());
+
+        disp_params = nullptr;
+        tr_params   = &tr_params_entry->second;
+    }
+
+    if (is_dispatch)
```
bradgrantham-lunarg, line 1333: Why is this separate from the previous `if-else` clause?

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
```diff
@@ -1228,18 +1414,16 @@ VkResult DispatchTraceRaysDumpingContext::DumpImmutableDescriptors(uint64_t qs_i
             }
         }
     }
-    else
+    else if (!is_dispatch)
```
bradgrantham-lunarg, line 1417: Isn't `!is_dispatch` implied by the `else` here?

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
panos-lunarg: The purpose of those new checks here is to deduce which from the currently bound descriptors are compatible with the current pipeline. For example:
- A pipeline is bound that allows descriptors in sets from 0 to 2.
- The bindings are populated with `vkCmdBindDescriptorSets`
- A different pipeline is bound that only uses set 0.
- The previous bindings still remain bound and are valid but the shaders used by the second pipeline are not allowed to access descriptors not described in the pipeline layout (only from set 0). Same applies for bindings within the sets.
- The new checks filter the descriptors based on the pipeline layout.

Previously, shader reflection basically did the filtering "automatically" but when it was removed resources that were not actually compatible with the active pipeline were being dumped just because there were bound for a previous pipeline.

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
panos-lunarg: yes

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
panos-lunarg: I think some leftovers from this PR's iterations. They can be merged

framework/decode/vulkan_object_info.h
```diff
@@ -439,6 +438,29 @@ struct VulkanShaderModuleInfo : public VulkanObjectInfo<VkShaderModule>
...
+
+using DescriptorBindingsLayout = std::unordered_map<uint32_t, DescriptorBindingLayout>;
+
+struct VulkanDescriptorSetLayoutInfo : public VulkanObjectInfo<VkDescriptorSetLayout>
+{
+    // One per descriptor binding
+    DescriptorBindingsLayout bindings_layout;
+};
+
+using DescriptorSetsLayout = std::vector<DescriptorBindingsLayout>;
```
fabian-lunarg, line 456: those nested `using` declarations are hard to read and not intuitive.
maybe use:
`using DescriptorBindingLayoutMap = std::unordered_map<uint32_t, DescriptorBindingLayout>;`

and leave out `using DescriptorSetsLayout = std::vector<DescriptorBindingsLayout>;`,
just use `std::vector<DescriptorBindingLayoutMap>` ?

then next questions, why is the latter an array, not also a map (`// One per descriptor set` would also use a map, since it could be sparse?)

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -11141,5 +11140,45 @@ void VulkanReplayConsumerBase::TrackNewPipelineCache(const VulkanDeviceInfo* dev
...
 
+VkResult VulkanReplayConsumerBase::OverrideCreatePipelineLayout(
+    PFN_vkCreatePipelineLayout                                func,
+    VkResult                                                  original_result,
+    const VulkanDeviceInfo*                                   device_info,
+    StructPointerDecoder<Decoded_VkPipelineLayoutCreateInfo>* pCreateInfo,
+    StructPointerDecoder<Decoded_VkAllocationCallbacks>*      pAllocator,
+    HandlePointerDecoder<VkPipelineLayout>*                   pPipelineLayout)
+{
+    VkResult result = func(device_info->handle,
```
fabian-lunarg, line 11151: other overrides do a nullptr/GFX_ASSERT check for `device_info`, maybe add that here too

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -11141,5 +11140,45 @@ void VulkanReplayConsumerBase::TrackNewPipelineCache(const VulkanDeviceInfo* dev
...
+                           pCreateInfo->GetPointer(),
+                           GetAllocationCallbacks(pAllocator),
+                           pPipelineLayout->GetHandlePointer());
+
+    if (result == VK_SUCCESS && UseExtraDescriptorInfo(device_info))
+    {
+        auto meta_info = pCreateInfo->GetMetaStructPointer();
+        assert(meta_info->decoded_value != nullptr);
```
fabian-lunarg, line 11159: prefer `GFXRECON_ASSERT` over `assert` within the framework

framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -1009,6 +986,10 @@ void VulkanReplayDumpResourcesBase::OverrideCmdBindPipeline(const ApiCallInfo&
     if (dispatch_rays_command_buffer != VK_NULL_HANDLE)
     {
         func(dispatch_rays_command_buffer, pipelineBindPoint, pipeline->handle);
+
+        DispatchTraceRaysDumpingContext* context = FindDispatchRaysCommandBufferContext(original_command_buffer);
+        assert(context != nullptr);
```
fabian-lunarg, line 991: assert -> GFXRECON_ASSERT

framework/decode/vulkan_object_info.h
panos-lunarg: > those nested using declarations are hard to read and not intuitive.
maybe use:
using DescriptorBindingLayoutMap = std::unordered_map<uint32_t, DescriptorBindingLayout>;
>
>and leave out using DescriptorSetsLayout = std::vector<DescriptorBindingsLayout>;,
just use std::vector<DescriptorBindingLayoutMap> ?

Done.

>then next questions, why is the latter an array, not also a map (// One per descriptor set would also use a map, since it could be sparse?)

Unless I am misunderstanding something, the sets in the pipeline layout are not sparse. `VkPipelineLayoutCreateInfo` takes only a count (`VkPipelineLayoutCreateInfo`) so sets start from 0 and are consecutive.

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
```diff
@@ -1089,30 +1091,198 @@ bool DispatchTraceRaysDumpingContext::IsRecording() const
...
-void DispatchTraceRaysDumpingContext::SnapshotBoundDescriptors(DispatchParameters& disp_params)
+void DispatchTraceRaysDumpingContext::SnapshotDispatchState(DispatchParameters& disp_params)
 {
-    for (const auto& desc_set : bound_descriptor_sets_compute)
+    // Copy descriptors that are compatible with the current pipeline layout
+    if (bound_pipeline_compute != nullptr)
     {
-        const uint32_t desc_set_index = desc_set.first;
-        for (const auto& desc : desc_set.second.descriptors)
+        for (const auto& desc_set : bound_descriptor_sets_compute)
```
fabian-lunarg, line 1099: a structured binding would be more readable than the pairs' first/second members:
e.g. `for (const auto& [set_idx, desc_info] : bound_descriptor_sets_compute)`

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
```diff
@@ -1089,30 +1091,198 @@ bool DispatchTraceRaysDumpingContext::IsRecording() const
...
-            tr_params.referenced_descriptors[desc_set_index][desc_binding_index] = desc.second;
+            const uint32_t desc_set_index = desc_set.first;
+
+            // Check against pipeline layout
+            if (bound_pipeline_trace_rays->desc_set_layouts.size() <= desc_set_index)
+            {
+                continue;
+            }
+
+            for (const auto& desc : desc_set.second.descriptors)
```
fabian-lunarg, line 1208: here too: consider structured binding for readability

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
```diff
@@ -1714,5 +1876,22 @@ void DispatchTraceRaysDumpingContext::EndCommandBuffer()
     device_table->EndCommandBuffer(DR_command_buffer);
 }
 
+void DispatchTraceRaysDumpingContext::BindPipeline(VkPipelineBindPoint bind_point, const VulkanPipelineInfo* pipeline)
+{
+    switch (bind_point)
+    {
+        case VK_PIPELINE_BIND_POINT_COMPUTE:
+            bound_pipeline_compute = pipeline;
```
fabian-lunarg, line 1884: maybe use a `map<VkPipelineBindPoint, const VulkanPipelineInfo*> bound_pipelines;` over individual members (bound_pipeline_compute / bound_pipeline_trace_rays)?

framework/decode/vulkan_replay_dump_resources_delegate.cpp
```diff
@@ -360,8 +410,9 @@ DefaultVulkanDumpResourcesDelegate::GenerateBufferDescriptorFilename(const Vulka
 {
     std::stringstream filename;
 
-    filename << capture_filename_ << "_" << "buffer_" << resource_info.buffer_info->capture_id << "_qs_"
-             << resource_info.qs_index << "_bcb_" << resource_info.bcb_index << "_rp_" << resource_info.rp << ".bin";
+    filename << capture_filename_ << "_"
+             << "buffer_" << resource_info.buffer_info->capture_id << "_qs_" << resource_info.qs_index << "_bcb_"
+             << resource_info.bcb_index << "_rp_" << resource_info.rp << ".bin";
```
fabian-lunarg, line 415: is the naming adhoc or is there some blueprint/doc somewhere for composing those file-names? I can only assume they need to match / stay in sync with some consumer

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
panos-lunarg: To be honest I wasn't aware of that notation and unfortunately I have used the current one in many places. From now on I'll try switching to this one which admittedly looks better

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
```diff
@@ -1142,16 +1312,16 @@ VkResult DispatchTraceRaysDumpingContext::DumpImmutableDescriptors(uint64_t qs_i
 
     if (is_dispatch)
     {
-        const auto disp_params = dispatch_params.find(cmd_index);
-        assert(disp_params != dispatch_params.end());
-        for (const auto& desc_set : disp_params->second.referenced_descriptors)
+        const auto disp_params_entry = dispatch_params.find(cmd_index);
+        GFXRECON_ASSERT(disp_params_entry != dispatch_params.end());
```
fabian-lunarg, line 1316: if `cmd_index` 'could' eventually be missing from the map, the assert would not be enough.
better add as condition `if(disp_params_entry != dispatch_params.end()){...)`?

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
fabian-lunarg: cool! structured-binding was one of my favorite additions with c++17 :)

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
panos-lunarg: Not sure. Since they are just two don't know it justifies using a `map`, and it would still need to be checked before accessing. Can change if you insist though, no strong opinion

framework/decode/vulkan_replay_dump_resources_delegate.cpp
panos-lunarg: Sorry this change is due to conflicts of my local `clang-format` and the one used in github. Those lines should not be affected by this PR. I'll revert them back

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
panos-lunarg: Actually it can't be null. The indices should be the same from the start of the execution so `find()` should not fail

fabian-lunarg (approved):
that's all I got in terms of remarks, none of them was really controversial

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
fabian-lunarg: yes right, it does not add real value changing here

bradgrantham-lunarg (approved):

PR #2089, opened 2025-03-25
Batch queue-submissions in ProcessBufferMemory / ProcessImageMemory

fabian-lunarg:
Instead of sequentially processing image/buffer-copies (each time doing a queue-submit + sync),
we can batch them together using a provided target-size for a combined staging-buffer.

This can drastically reduce the number of queue-submissions and improves scalability 
when processing larger quantities of textures and/or buffers.

The target-size for staging-buffers is set to `min(128Mb, total_num_staging_bytes)`, which might increase
memory-footprint during state-writes (don't think this a problem).

additional:
- Add debug-lables to internal resources in VulkanResourceUtil via `vkSetDebugUtilsObjectNameEXT`
- Remove dead code in VulkanResourcesUtil
- Add a nullptr check in VulkanAddressReplacer
- Add const-qualifiers to util::Compressor's API

### DIFF
--- a/framework/encode/vulkan_state_writer.cpp
+++ b/framework/encode/vulkan_state_writer.cpp
@@ -2074,6 +2074,8 @@ void VulkanStateWriter::ProcessBufferMemory(const vulkan_wrappers::DeviceWrapper
         }
     };
 
+    size_t num_staging_bytes = 0;
+
     for (const auto& snapshot_entry : buffer_snapshot_infos)
     {
         const vulkan_wrappers::BufferWrapper*       buffer_wrapper = snapshot_entry.buffer_wrapper;
@@ -2090,6 +2092,7 @@ void VulkanStateWriter::ProcessBufferMemory(const vulkan_wrappers::DeviceWrapper
         if (snapshot_entry.need_staging_copy)
         {
             buffer_resources.push_back(buffer_resource);
+            num_staging_bytes += buffer_resource.size;
         }
         else
         {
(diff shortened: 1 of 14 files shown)


framework/encode/vulkan_state_writer.cpp
```diff
@@ -2551,6 +2564,10 @@ void VulkanStateWriter::ProcessImageMemoryWithAssetFile(const vulkan_wrappers::D
             }
         }
     }
+
+    // batch process image-downloads requiring staging, use 128MB staging-mem
+    constexpr uint32_t staging_buffer_size = 128U << 20U;
```
panos-lunarg, line 2569: In the replay side I decided instead of using some arbitrarily predefined size for the stage buffer to stick to the size calculated by the state writer, which essentially is the size of the biggest resource.

fabian-lunarg: yes, this is generally hard to guess as it depends on the exact scenario. 
that value still seemed reasonable, my thinking was:

- using the largest resource-size degrades for large quantities of small resources.
- 128 Mb is (probably) never too large for a temporary staging-arena

fabian-lunarg: ideally we would measure the difference across a set of different scenarios and find some sweet-spot heuristic.

framework/graphics/vulkan_resources_util.h
```diff
@@ -95,66 +82,89 @@ class VulkanResourcesUtil
...
+    /**
+     * @brief   ReadImageResources processes an array of ImageResources in batches,
+     *          downloads data from GPU-memory using a staging-buffer and provides that data via callback-function.
+     *
+     * @param   image_resources     an array of ImageResource-structs
+     * @param   call_back           a callback-function, consuming data from staging-buffer
+     * @param   staging_buffer_size target size for the staging-buffer in bytes. we might allocate a larger buffer,
+     *                              depending on largest resource-size
+     */
+    VkResult ReadImageResources(const std::vector<ImageResource>&   image_resources,
```
fabian-lunarg, line 127: @panos-lunarg  the best way to read this PR is to start with this API change here,
as most other things just follow after that

framework/encode/vulkan_state_writer.cpp
fabian-lunarg: I went back and now cap the staging-buffer size with total num bytes, like this:
`size_t staging_buffer_size = std::min<size_t>(128U << 20U, num_staging_bytes);`

panos-lunarg: sounds good

framework/graphics/vulkan_resources_util.cpp
```diff
@@ -1034,163 +1082,97 @@ void VulkanResourcesUtil::DestroyStagingBuffer()
...
             GFXRECON_LOG_ERROR("Failed to create a command buffer for resource memory snapshot");
+            return VK_NULL_HANDLE;
         }
-    }
 
-    if (result == VK_SUCCESS)
-    {
-        result = BeginCommandBuffer();
+        // Because this command buffer was not allocated through the loader, it must be assigned a dispatch table.
+        *reinterpret_cast<void**>(command_asset.command_buffer) = *reinterpret_cast<void**>(device_);
```
antonio-lunarg, line 1124: Can this be a helper function?

framework/graphics/vulkan_resources_util.cpp
```diff
@@ -1034,163 +1082,97 @@ void VulkanResourcesUtil::DestroyStagingBuffer()
...
+            // NOTE: VK_HANDLE_TO_UINT64 fails here for 32bit (while it does work for VkBuffer), open for suggestions
+            object_name_info.objectHandle = (uint64_t)command_asset.command_buffer;
+            object_name_info.pObjectName  = "VulkanResourcesUtil internal command-buffer";
+            set_debug_utils_object_name_fn_(device_, &object_name_info);
+        }
     }
 
-    return result;
+    // begin + return handle
+    BeginCommandBuffer(command_asset.command_buffer);
```
antonio-lunarg, line 1140: Mmh, how about naming this function `CreateCommandBufferAndBegin()`?

framework/graphics/vulkan_resources_util.cpp
```diff
@@ -1451,493 +1437,478 @@ VkResult VulkanResourcesUtil::ResolveImage(VkImage           image,
     create_info.initialLayout         = VK_IMAGE_LAYOUT_UNDEFINED;
 
     VkResult result = device_table_.CreateImage(device_, &create_info, nullptr, resolved_image);
-    if (result == VK_SUCCESS)
+    if (result != VK_SUCCESS)
```
antonio-lunarg, line 1440: I honestly prefer early returns and avoid multiple nested happy paths.

framework/graphics/vulkan_resources_util.cpp
```diff
@@ -773,6 +775,42 @@ bool NextRowTexelCoordinates(VkImageType       imageType,
...
+VulkanResourcesUtil::VulkanResourcesUtil(VkDevice                                device,
+                                         VkPhysicalDevice                        physical_device,
+                                         const encode::VulkanDeviceTable&        device_table,
+                                         const encode::VulkanInstanceTable&      instance_table,
+                                         const VkPhysicalDeviceMemoryProperties& memory_properties) :
+    device_(device),
+    device_table_(device_table), physical_device_(physical_device), instance_table_(instance_table),
+    memory_properties_(memory_properties)
+{
+    assert(device != VK_NULL_HANDLE);
```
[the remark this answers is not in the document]
fabian-lunarg: applying for all asserts in that file

framework/graphics/vulkan_resources_util.cpp
fabian-lunarg: in this context we require it exactly once. no need for breaking it out (yet)? - plus, I wasn't entirely convinced we require this at all for our purposes, but not sure

framework/graphics/vulkan_resources_util.cpp
fabian-lunarg: moved all cleanup for resolve/scaled-images into `~image_resource_tmp_data_t()` - so we cannot leak handles when returning early

framework/graphics/vulkan_resources_util.cpp
fabian-lunarg: done

framework/graphics/vulkan_resources_util.cpp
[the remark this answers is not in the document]
fabian-lunarg: done

framework/encode/vulkan_state_writer.cpp
```diff
@@ -2343,47 +2387,110 @@ void VulkanStateWriter::ProcessImageMemory(const vulkan_wrappers::DeviceWrapper*
...
 {
-    assert(device_wrapper != nullptr);
-    assert(asset_file_stream_ != nullptr);
-
+    GFXRECON_ASSERT(device_wrapper != nullptr);
+    GFXRECON_ASSERT(asset_file_stream_ != nullptr);
     const VulkanDeviceTable* device_table = &device_wrapper->layer_table;
 
+    using ImageResource = graphics::VulkanResourcesUtil::ImageResource;
+    std::vector<ImageResource> image_resources;
```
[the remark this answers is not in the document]
fabian-lunarg: done

wpaul-samsung:
I'm seeing around 20x improvement in write state performance with this PR!  I haven't done a lot of testing with it, but I haven't noticed any obvious regressions.

framework/graphics/vulkan_resources_util.cpp
```diff
@@ -1451,502 +1437,521 @@ VkResult VulkanResourcesUtil::ResolveImage(VkImage           image,
...
+        } // current batch, consume staging-buffer
+
+        UnmapStagingBuffer();
+
+        auto batch_micros = std::chrono::duration_cast<std::chrono::microseconds>(
+                                std::chrono::high_resolution_clock::now() - start_time)
+                                .count();
+        cpu_micros += batch_micros - gpu_micros;
+    } // batch_ranges
+    GFXRECON_LOG_DEBUG("gpu: %d ms - cpu: %d ms", gpu_micros / 1000, cpu_micros / 1000);
```
fabian-lunarg, line 1931: @wpaul-samsung I left those timing probes in place in `ReadImageResources`. 
on linux I'd set the corresponding env-variable like this: `export GFXRECON_LOG_LEVEL=debug`
currently asking for the exact android equivalent ...

antonio-lunarg (approved):

fabian-lunarg:
> CI gfxreconstruct build # 6517 failed.

unrelated CI-error / false negative:
```
CMake Error at C:/Program Files/CMake/share/cmake-3.31/Modules/FetchContent.cmake:1918 (message):
  Build step for ags_sdk failed: 1
```
running again ...

PR #2094, opened 2025-03-26
Fix JSON conversion of `PointerDecoder`

marius-pelegrin-arm:
When a `PointerDecoder` didn't contain any value (because it was a nullptr) the value was still read by `FieldToJson` and written to the output of `gfxrecon-convert`.

The reason to that bug is that the test was:
```
if (len > 1)
    output a JSON array
else
    output a JSON integer
```
leaving begind the case where `len` is `0`.

This commit fixes this issue by implementing the same condition as in `HandleToJson` for `HandlePointerDecoder`.

For the following struct:
```
{
    int* x = {1, 2, 3};
    int* y = {1};
    int* z = nullptr;
}
```
The output was:
```
{
    "x": [1, 2, 3],
    "y": 1,
    "z": some garbage value
}
```
The output is now:
```
{
    "x": [1, 2, 3],
    "y": [1],
    "z": null
}
```

### DIFF
--- a/framework/decode/decode_json_util.cpp
+++ b/framework/decode/decode_json_util.cpp
@@ -96,14 +96,14 @@ void FieldToJson(nlohmann::ordered_json&                   jdata,
     {
         const auto decoded_value = data.GetPointer();
         const auto length        = data.GetLength();
-        if (length > 1)
+        if (data.IsArray())
         {
             for (size_t i = 0; i < length; ++i)
             {
                 jdata[i] = decoded_value[i];
             }
         }
-        else
+        else if (length == 1)
         {
             jdata = *decoded_value;
         }
@@ -119,14 +119,14 @@ void FieldToJson(nlohmann::ordered_json&                 jdata,
     {
         const auto decoded_value = data.GetPointer();
         const auto length        = data.GetLength();
-        if (length > 1)
+        if (data.IsArray())
         {
             for (size_t i = 0; i < length; ++i)
             {
                 jdata[i] = decoded_value[i];
             }
         }
-        else
+        else if (length == 1)
         {
             jdata = *decoded_value;
         }
@@ -142,14 +142,14 @@ void FieldToJson(nlohmann::ordered_json&                   jdata,
     {
         const auto decoded_value = data.GetPointer();
         const auto length        = data.GetLength();
-        if (length > 1)
+        if (data.IsArray())
         {
             for (size_t i = 0; i < length; ++i)
             {
                 jdata[i] = decoded_value[i];
             }
         }
-        else
+        else if (length == 1)
         {
             jdata = *decoded_value;
         }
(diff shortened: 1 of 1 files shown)


fabian-lunarg:
from internal CI:

```
ERROR: convert json files are different at line/block 150
Result:	{'index': 149, 'function': {'name': 'vkQueueSubmit', 'thread': 1, 'sub_index': 2, 'return': 'VK_SUCCESS', 'args': {'queue': 6, 'submitCount': 1, 'pSubmits': [{'sType': 'VK_STRUCTURE_TYPE_SUBMIT_INFO', 'waitSemaphoreCount': 1, 'pWaitSemaphores': [8], 'pWaitDstStageMask': [1024], 'commandBufferCount': 1, 'pCommandBuffers': [42], 'signalSemaphoreCount': 1, 'pSignalSemaphores': [9], 'pNext': None}], 'fence': 7}}}

Reference:	{'index': 149, 'function': {'name': 'vkQueueSubmit', 'thread': 1, 'sub_index': 2, 'return': 'VK_SUCCESS', 'args': {'queue': 6, 'submitCount': 1, 'pSubmits': [{'sType': 'VK_STRUCTURE_TYPE_SUBMIT_INFO', 'waitSemaphoreCount': 1, 'pWaitSemaphores': [8], 'pWaitDstStageMask': 1024, 'commandBufferCount': 1, 'pCommandBuffers': [42], 'signalSemaphoreCount': 1, 'pSignalSemaphores': [9], 'pNext': None}], 'fence': 7}}}
```
the difference is `'pWaitDstStageMask': [1024]` vs. `'pWaitDstStageMask': 1024` - which matches the description above.

MarkY-LunarG:
@marius-pelegrin-arm, did you see Fabian's message above?

marius-pelegrin-arm:
@fabian-lunarg yes, but he says it matches my description. Yes, I propose to change the output to something that makes more sense - an array of size 1 is still an array, and should be displayed as such.

MarkY-LunarG:
Oh wait, I was thinking it was backwards.  Yes, `pWaitDstStageMask: [1024]` should be what I expect (because it is an array).  I'm sorry, I though Fabian said it was the other way around.  So I'll see what needs to be done for our internal CI.

MarkY-LunarG (approved):

MarkY-LunarG:
@davidlunarg , will this modify our dump resources output?  @bradgrantham-lunarg mentioned it might be good for you to review since it could affect the results for our other customers.  What is the dump resources commit suite that I can use to run through CI manually?

Originally I was going to just update the few JSONs I knew about, but Brad mentioned the further issues.

For now, this is going to stall because we have to:
 * Replace all our CI JSONs
 * May have to update our Dump Resources JSON
 * May have to coordinate with end-users who are using our JSON output

PR #2095, opened 2025-03-27
Fix Android multi-window surface recreation

marius-pelegrin-arm:
This commit fixes two bugs.

The first is a simple forgotten call to `ANativeWindow_release` on window destruction.

The second is a synchronization issue in the Java code: During window recreation, a destroy call and a create call closely follow each other. Each of these calls is sent to the UI thread, that will then trigger the callbacks `surfaceDestroyed` and `surfaceCreated` later.
If the call to `addNewView` happens before the UI thread calls `surfaceDestroyed`, then `mSurface` is set to the destroyed surface (which is not null), which allows `addNewView` to return with the incorrect surface set, which results in a failed surface recreation.
I propose to solve this bug by simply not setting `mSurface` in `surfaceDestroyed` and `surfaceChanged` callbacks, as `mSurface` is exclusively used to detect newly created surfaces.

### DIFF
--- a/framework/application/android_jni.cpp
+++ b/framework/application/android_jni.cpp
@@ -34,7 +34,7 @@ extern "C" JNIEXPORT void JNICALL Java_com_lunarg_gfxreconstruct_replay_ReplayAc
                                                                                                   jobject surface)
 {
     gfxrecon::application::tmp_window = ANativeWindow_fromSurface(env, surface);
-    GFXRECON_LOG_INFO("Created new window %p", gfxrecon::application::tmp_window);
+    GFXRECON_LOG_INFO("Created new window %p from surface %p", gfxrecon::application::tmp_window, surface);
 }
 
 #endif // GFXR_MULTI_WINDOW_REPLAY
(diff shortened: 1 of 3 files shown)


framework/application/android_jni.cpp
```diff
@@ -34,7 +34,7 @@ extern "C" JNIEXPORT void JNICALL Java_com_lunarg_gfxreconstruct_replay_ReplayAc
                                                                                                   jobject surface)
 {
     gfxrecon::application::tmp_window = ANativeWindow_fromSurface(env, surface);
-    GFXRECON_LOG_INFO("Created new window %p", gfxrecon::application::tmp_window);
+    GFXRECON_LOG_INFO("Created new window %pfrom surface %p\n", gfxrecon::application::tmp_window, surface);
```
antonio-lunarg, line 37: Missing space after first `%p` and no need for `\n`

marius-pelegrin-arm: Good catch ! Solved =)

antonio-lunarg (approved):

Issue #2071, opened 2025-03-14
[replay] Replayer exposes all symbols


bartosz-muszarski-arm:
@MarkY-LunarG 
Change https://github.com/LunarG/gfxreconstruct/pull/1739 introduces following line in cmake:
`target_link_options(gfxrecon-replay PUBLIC "-rdynamic")`  to expose `SetInjectedCommandCallbacks` symbol.
However, this exposes all symbols from gfxrecon-replay, which may result in symbol collisions depending on the environment.
If collisions do happen, resulting issues are hard to track and identify.
Symbols exposed by the replay binary should be limited and controlled.

panos-lunarg:
> @panos-lunarg , do you have a list of additional symbols you need exported so we can cut down on the symbol leakage and define the above export files?

No, I think those 3 symbols that you list are of them that need to be exposed

MarkY-LunarG:
Some symbols reduced.  Still reducing more.

bradgrantham-lunarg:
The changes in #2154 block symbols in the stack trace on Android in logcat.  It's likely we will need to revert them.  If there is another way to accomplish the same thing, I want to hear it.

bartosz-muszarski-arm:
This issue affects builds that statically link libraries - libz specifically - without restricting symbol visibility. The solution is to limit the visibility of symbols pulled from statically linked libraries, to prevent unintended symbol overrides during dynamic linking.

PR #2098, opened 2025-03-27
layer: Limit layer library's exposed symbols

MarkY-LunarG:


MarkY-LunarG:
This is to resolve Issue #2071

charles-lunarg (commented):
It would be  good to add this to the top level CMakeLists.txt file. 

```
set(CMAKE_CXX_VISIBILITY_PRESET "hidden")
set(CMAKE_VISIBILITY_INLINES_HIDDEN "YES")
```

Otherwise the change looks appropriate. Not sure what symbols are needed for the dx12 component, so good to start with the Vulkan layer.

dfriederich:
Just as note/suggestions, this PR controls the exported symbols via linker file, good improvement. But for Linux/Android this does not add the same benefits as also compiling with hidden symbols by default. Hidden symbols allow the compiler to inline much more than he can otherwise. We use the gfxr build.py with this option added
 --cmake-extra=CMAKE_CXX_VISIBILITY_PRESET=hidden
 for this.

MarkY-LunarG:
> Just as note/suggestions, this PR controls the exported symbols via linker file, good improvement. But for Linux/Android this does not add the same benefits as also compiling with hidden symbols by default. Hidden symbols allow the compiler to inline much more than he can otherwise. We use the gfxr build.py with this option added --cmake-extra=CMAKE_CXX_VISIBILITY_PRESET=hidden for this.

Thanks @dfriederich , made the dcorresponding changes.

bradgrantham-lunarg:
Did [tests: Fix external memory test race condition](https://github.com/LunarG/gfxreconstruct/pull/2098/commits/8c92a4e51ec058ecbfcdc60013a24940a107091f) get pulled in by mistake?

layer/VkLayer_gfxreconstruct.exp
```diff
@@ -0,0 +1,5 @@
+_vkGetInstanceProcAddr
+_vkGetDeviceProcAddr
+_vkEnumerateInstanceLayerProperties
+_vkEnumerateInstanceExtensionProperties
+_vkNegotiateLoaderLayerInterfaceVersion
```
bradgrantham-lunarg, line 5: Why isn't this list the same as the lists for Linux and Windows and Android?

MarkY-LunarG: Those are Android-specific, however they still may be used by emulators.  So I went ahead and added them as well now.  It was really an oversight.  Thanks for catching this.

MarkY-LunarG:
> Did [tests: Fix external memory test race condition](https://github.com/LunarG/gfxreconstruct/pull/2098/commits/8c92a4e51ec058ecbfcdc60013a24940a107091f) get pulled in by mistake?

@bradgrantham-lunarg, it keeps failing for me for some reason.  Beau originally noticed that it might have been because I tured on hidden symbols for all of GFXR which included the Mock ICD.  However, moving it to the layer only seems to have worked once.   I figured it might be a race condition in the test, so I tried modifying it, but it still seems unhappy.

bradgrantham-lunarg:
> > Did [tests: Fix external memory test race condition](https://github.com/LunarG/gfxreconstruct/pull/2098/commits/8c92a4e51ec058ecbfcdc60013a24940a107091f) get pulled in by mistake?
> 
> @bradgrantham-lunarg, it keeps failing for me for some reason. Beau originally noticed that it might have been because I tured on hidden symbols for all of GFXR which included the Mock ICD. However, moving it to the layer only seems to have worked once. I figured it might be a race condition in the test, so I tried modifying it, but it still seems unhappy.
> 
> Oh, and it passes on my machine which is Pop-OS (Ubuntu base) with the test changes.

What I mean is, why is this commit part of this PR?  Presumably this PR is just the symbol limit.

test/test_apps/external-memory-fd/export/app.cpp
```diff
@@ -162,36 +162,51 @@ int App::get_exportable_fd()
...
 }
 
 void App::send_exportable_fd(int exportable_fd)
 {
     // Need to send the fd to the importer process
     int external_socket = socket(PF_UNIX, SOCK_STREAM, 0);
-    GFXRECON_ASSERT(external_socket >= 0)
+    if (external_socket < 0)
+    {
+        GFXRECON_LOG_ERROR("Export App Failed to create socket (%d)", external_socket);
```
bradgrantham-lunarg, line 175: I imagine these should all be fatal and return a failure status for the app.

test/test_apps/external-memory-fd/import/app.cpp
```diff
@@ -274,8 +285,14 @@ bool App::frame(const int frame_num)
 
 void App::cleanup()
 {
+    usleep(30000);
```
bradgrantham-lunarg, line 288: I'd prefer something like the import app holds the socket open until `cleanup`, then the export app `read`s the socket or `select`s on the socket after sending the fd and waits for the `read` to return 0 bytes, which it should do when the import app closes the socket.  Then no worries about race condition.

(I'm surprised there isn't a reference count on that memory but I can't find anythign in the spec about it.)

MarkY-LunarG:
> It would be good to add this to the top level CMakeLists.txt file.
> 
> ```
> set(CMAKE_CXX_VISIBILITY_PRESET "hidden")
> set(CMAKE_VISIBILITY_INLINES_HIDDEN "YES")
> ```
> 
> Otherwise the change looks appropriate. Not sure what symbols are needed for the dx12 component, so good to start with the Vulkan layer.

Originally I did, but Beau suggested moving it in case this is what caused the failures with the Mock ICD.  The subsequent run passed, so it seemed like it had solved the problem.  However, it came back.  So we could move this back to the top, but I'm reluctant to make any other changes until we resolve the test failure race condition.

MarkY-LunarG:
@bradgrantham-lunarg , should I do what @charles-lunarg suggests and move the hidden symbol change to the top-level?  It would automatically hide symbols on all our other items, and I don't know if that's something you'd be interested in by default since it might cause breakage somewhere.

bradgrantham-lunarg:
> @bradgrantham-lunarg , should I do what @charles-lunarg suggests and move the hidden symbol change to the top-level? It would automatically hide symbols on all our other items, and I don't know if that's something you'd be interested in by default since it might cause breakage somewhere.

Does this change **without** Charles's suggestion fix the issue as reported by Bartosz?  Then I would go with that for this PR.  I don't understand the implication of adding `-fvisibility-inlines-hidden` at the top level and will need to get that; can we make another PR for `set(CMAKE_CXX_VISIBILITY_PRESET "hidden")  set(CMAKE_VISIBILITY_INLINES_HIDDEN "YES")` that we can reason about?

bradgrantham-lunarg (approved):

MarkY-LunarG:
Just verified OpenXR continues to function as well with the restricted entrypoints.

PR #2100, opened 2025-03-28
Add a `--bare` option to `gfxrecon-convert`

marius-pelegrin-arm:
When comparing different traces that are very similar, we may want to convert the traces using the `jsonl` format to look directly at the call differences.
However, some data in the output of `gfxrecon-convert` tends to "pollute" the difference, like the block index at the begining of each line.

This commit implements a `--bare` option to make the output of `gfxrecon-convert` more "diff-friendly".


bartosz-muszarski-arm:
That's actually how this change started, as --no-index.
But since the goal in the end is to make the output diff-friendly, not remove "index" in particular, but anything that might introduce any noise in the diffs in the future, it was renamed to reflect its actual purpose.
That way we can rely on this in automation today.
Would it be really useful to have separate options for these fields anyway?

marius-pelegrin-arm:
Hello ! So I put aside the `--bare` option for the moment and just renamed it `--no-index` =)

bradgrantham-lunarg:
You can strip `index` from all the records in a `jsonl` with:
```
jq -c 'del(.index)'
```
Because this is so well-handled by a composable pipeline (you should be able to `convert --log-level error --output stdout` and pipe through `jq`, I don't think I would recommend we merge this PR.

PR #2102, opened 2025-03-28
Fix generic handle ids in the json consumer

mikes-lunarg:
Instead of hard-coding handle names in the json generators, use the `is_generic_cmd_handle_value`/`is_generic_struct_handle_value` from the `KhronosBaseGenerator`. This fixes handle values from VK_EXT_debug_report and VK_EXT_debug_marker which were previously displayed incorrectly in the json output.


locke-lunarg (approved):

bradgrantham-lunarg (approved):

mikes-lunarg:
I brought this branch up to date and added a test case

PR #2104, opened 2025-04-01
Change loop iterator var to const ref

mporth-brcm:
Removes compiler warning from our "strict" environment.
Probably better code generation too!


nickdriscoll-lunarg (approved):
I can't think of a reason not to accept this, and codegen seems better on MSVC and gcc.
https://godbolt.org/z/cG6M4oe9s

PR #2108, opened 2025-04-01
Update nlohmann-json to 3.11.3

bradgrantham-lunarg:
Our current nlohmann-json is a hacked-down version.  This PR moves to using an unmodified install directory tree instead.

CMakeLists.txt for Android replay and multi-win-replay have workarounds for `find_package` because the NDK appears to override/disable the use of CMAKE_PREFIX_PATH inside the project CMakeLists.txt.

Making nlohmann_json be a submodule is an option but would more than double the download for our repo including the other submodules so I opted to go this route instead.


android/layer/CMakeLists.txt
```diff
@@ -31,6 +34,7 @@ target_link_libraries(VkLayer_gfxreconstruct
                       gfxrecon_graphics
                       gfxrecon_format
                       gfxrecon_util
+                      nlohmann_json::nlohmann_json
```
antonio-lunarg, line 37: By adding this to gfxrecon-util's link-libraries, then we have a transitive dependency, so maybe we can drop this line (and the ones associated to gfxrecon-replay) ?

locke-lunarg (approved):

android/layer/CMakeLists.txt
bradgrantham-lunarg: Yes!

MarkY-LunarG (approved):

PR #2110, opened 2025-04-02
Add support for offset-addresses in VulkanAddressReplacer::UpdateBuffer (-m rebind only)

fabian-lunarg:
This PR improves replay of captures using `SPV_KHR_physical_storage_buffer` together with `-m rebind`.
 
Adds a new compute-shader with different approach for address-lookup.
so far address-replacement in GPU-buffers was using a hashmap
(containing all buffer base-addresses and mapping from capture -> replay time.)

this was sufficient when only pointers to those base-addresses were used in shaders.
e.g. https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/extensions/ray_tracing_reflection
worked fine.

but often pointers are calculated using a base-address and an offset.
e.g. https://github.com/SaschaWillems/Vulkan/tree/master/examples/raytracinggltf
did NOT work.

to support lookup of ranges this PR switches from a hashmap to a binary-search based lookup.
to avoid issues with overlapping capture<->replay ranges and successive wrong double replacements,
we combine this lookup with a hashmap, keeping track of  addresses which have already been replaced

so the PR fixes the existing caveat, in shaders we now support addresses with arbitrary offsets.

it also removes two long-standing warnings:
- when using `SPV_KHR_physical_storage_buffer` in shaders
- when using `vkGetBufferDeviceAddress` together with `-m rebind`


framework/decode/vulkan_address_replacer.cpp
```diff
@@ -1366,6 +1378,9 @@ bool VulkanAddressReplacer::init_queue_assets()
         return false;
     }
 
+    // Because this command buffer was not allocated through the loader, it must be assigned a dispatch table.
+    *reinterpret_cast<void**>(command_buffer_) = *reinterpret_cast<void**>(device_);
```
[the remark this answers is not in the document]
fabian-lunarg: I'll consolidate this here or in the other PR (and rebase here) to put the logic into a vulkan-util header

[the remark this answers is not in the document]
fabian-lunarg: done, added as `graphics::copy_dispatch_table_from_device`

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -115,6 +115,23 @@ struct replacer_params_t
...
+    uint32_t        pad;
+};
+
+struct replacer_params_binary_t
+{
+    sorted_array_t address_array;
+    hashmap_t      address_blacklist;
+
+    // input-/output-arrays can be identical when sbt-alignments/strides match
+    VkDeviceAddress input_handles, output_handles;
```
MarkY-LunarG, line 131: I think we want these split into 2 separate lines.

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -115,6 +115,23 @@ struct replacer_params_t
     uint32_t        num_handles;
 };
 
+struct sorted_array_t
```
MarkY-LunarG, line 118: Can you put up a comment explaining more what these 2 structs are supposed to be for?  The type says "sorted_array_t" but it contains address, size and pad.  Is it host, local.  Are the contents sorted, or is this an element to a sorted array.

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -1663,24 +1681,59 @@ void VulkanAddressReplacer::run_compute_replace(const VulkanCommandBufferInfo*
...
-        GFXRECON_LOG_WARNING_ONCE("VulkanAddressReplacer::run_compute_replace(): internal pipeline-creation failed",
-                                  __func__)
+        GFXRECON_LOG_WARNING_ONCE("VulkanAddressReplacer::run_compute_replace(): internal pipeline-creation failed")
         return;
     }
 
-    auto& pipeline_context_bda = pipeline_context_map_[command_buffer_info->handle].emplace_back();
+    // blacklist hashmap storage  >= 16 bytes * 2^20 slots
+    constexpr uint32_t hashmap_elem_size = 2 * sizeof(VkDeviceSize);
+    uint32_t           hashmap_num_slots = std::max<uint32_t>(1U << 20, util::next_pow_2(num_addresses * 4));
```
MarkY-LunarG, line 1690: This seems pretty big for mobile.  Is there a way we can do smaller and grow?  At least in the future?

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -1663,24 +1681,59 @@ void VulkanAddressReplacer::run_compute_replace(const VulkanCommandBufferInfo*
...
+        barrier(command_buffer_info->handle,
+                hashmap_storage_bda_binary_.buffer,
+                VK_PIPELINE_STAGE_TRANSFER_BIT,
+                VK_ACCESS_TRANSFER_WRITE_BIT,
+                VK_SHADER_STAGE_COMPUTE_BIT,
+                VK_ACCESS_SHADER_READ_BIT | VK_ACCESS_SHADER_WRITE_BIT);
+    }
+
+    auto&    pipeline_context_bda = pipeline_context_map_[command_buffer_info->handle].emplace_back();
+    uint32_t storage_size         = std::max<uint32_t>(sizeof(bda_element_t) * storage_bda_binary_.size(), 1U << 10U);
```
MarkY-LunarG, line 1721: Same here wit the 1 to the 10th.  Also, I think these values should be defined as constants somewhere.

framework/decode/vulkan_address_replacer_shaders.h
```diff
@@ -147,11 +147,14 @@ void main()
...
 #endif // g_replacer_sbt_comp: original GLSL source
 
-// g_replacer_bda_comp: original GLSL source
+// g_replacer_bda_binary_comp: original GLSL source
 #if 0
 #version 460
 #extension GL_EXT_buffer_reference2 : require
+#extension GL_EXT_scalar_block_layout : require
 #extension GL_EXT_shader_explicit_arithmetic_types_int64 : require
+#extension GL_EXT_shader_atomic_int64 : require
```
MarkY-LunarG, line 156: Is this extension supported on Arm?

per-mathisen-arm: All the latest mobile GPUs should support GL_EXT_shader_atomic_int64 in their latest generation of products.

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: I'll consolidate those two wrapper-structs, they serve the same purpose:
- they wrap a GPU-array (device-address, size, capacity)
- this array is used as storage for either a (sorted) array OR a linear hashmap

I think using `gpu_array_t` for both and adding some clarifying doc should clear things up a bit.

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: yes, it's large. basically I wanted to buy some time until we support resizing this global cache.
it's possible but resizing will require a couple missing things:

- atomically increment and read back the hashmap.size member
- read back the gpu-memory to host and rehash into a bigger array (OR grow and then rehash using another compute-shader)

let me get back to that after PTO, but you're right, there is something missing
and this part is honestly still more in a proof-of-concept state.

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: done

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: done

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: I added a `constexpr uint32_t min_storage_size` and motivating comment. the reason this was not exposed more publicly is the actual number is not so important and could reasonably be anywhere between 0 ... 16kB.

framework/decode/vulkan_address_replacer_shaders.h
fabian-lunarg: that was my conclusion as well, but I'm relieved that Per is able to confirm, thank you!
it's in fact a crucial feature for the gpu-hashmap insertions introduced in this PR.

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: I added a corresponding `TODO`, we will need this for scalability

fabian-lunarg: done, implemented resizing and rehashing :)
we now start with an initial storage size of 64kB and grow 2x when the load-factor reaches 0.25.
that's all hardcoded in `constexpr`s because I think we'll prob. never going to change it.
a newly added compute-shader is responsible for rehashing from old -> new hasmap.

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: redone without magic number

MarkY-LunarG (approved):

PR #2114, opened 2025-04-03
Remove VK dependency in SPIRV Parser

mporth-brcm:
Remove unused function `SpvReflectToVkDescriptorType` from `spirv_parsing_util.cpp` which also means that `encode/vulkan_state_info.h` does not have to be included in `spirv_parsing_util.h`.


framework/util/spirv_parsing_util.cpp
```diff
@@ -644,52 +644,5 @@ std::vector<SpirVParsingUtil::BufferReferenceInfo> SpirVParsingUtil::GetBufferRe
     return ret;
 }
 
-static VkDescriptorType SpvReflectToVkDescriptorType(SpvReflectDescriptorType type)
```
panos-lunarg, line 647: The same function is also defined in `framework/decode/vulkan_replay_consumer_base.cpp` which can also be removed

bradgrantham-lunarg: @panos-lunarg was your feedback integrated?  Can you resolve this thread?

PR #2116, opened 2025-04-03
Fix external FD replay

MarkY-LunarG:
We need to check the VkExternalMemoryHandleTypeFlags during both vkCreateImage and vkCreateBuffer calls and if they use FD, strip out the handle type from the corresponding
"VkExternalMemory..CreateInfo" struct.  If there are no other flags after that is stripped, just remove the external memory struct from the create info pNext chain.

Also, converted a few asserts to GFXRECON_ASSERT and added some additional logging info.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -5273,17 +5273,40 @@ VulkanReplayConsumerBase::OverrideCreateBuffer(PFN_vkCreateBuffer               
 {
     GFXRECON_UNREFERENCED_PARAMETER(original_result);
 
-    assert((device_info != nullptr) && (pCreateInfo != nullptr) && (pBuffer != nullptr) && !pBuffer->IsNull() &&
-           (pBuffer->GetHandlePointer() != nullptr));
+    GFXRECON_ASSERT((device_info != nullptr) && (pCreateInfo != nullptr) && (pBuffer != nullptr) &&
+                    !pBuffer->IsNull() && (pBuffer->GetHandlePointer() != nullptr));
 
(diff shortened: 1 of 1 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5452,50 +5475,70 @@ VulkanReplayConsumerBase::OverrideCreateImage(PFN_vkCreateImage
...
     {
         // In this case, the image has been sampled at capture time and format is now RGBA8_UNORM.
-        create_info->format             = VK_FORMAT_R8G8B8A8_UNORM;
+        modified_create_info.format     = VK_FORMAT_R8G8B8A8_UNORM;
         external_format->externalFormat = 0;
     }
 
-    VkResult result = allocator->CreateImage(
-        pCreateInfo->GetPointer(), GetAllocationCallbacks(pAllocator), capture_id, replay_image, &allocator_data);
+    if (external_memory != nullptr && external_memory->handleTypes & VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD_BIT)
```
antonio-lunarg, line 5492: I think you can remove this if and its body altogether.

MarkY-LunarG: This is somehow bad merge code.  Good catch, I'll just strip out this block.  I have no idea how it created it from the above if contents but the below if check.

MarkY-LunarG: Code now removed.

antonio-lunarg (approved):

ziga-lunarg (approved):

PR #2119, opened 2025-04-04
tests: Fix external memory test capture file race condition

bradgrantham-lunarg:
Both apps were being run with identical capture settings.  We were lucky that the import app was run second, so it won. Based on my experiments, I think the two apps had file pointers into the same file, and the export app was ahead of the import app, so the contents of the capture file ended up being the import apps capture followed by whatever junk was already written by the export app at that location in the file.

This changes forces the "background" app (export app) to capture to "background.gfxr".

This is a brute-force fix to get past a strange regression in #2098 in which the export app's capture somehow started later and that's what was compared to the import app's reference .gfxr.

At a later time @beau-lunarg and @antonio-lunarg and I should get together and figure out the ideal behavior for multi-app tests.


antonio-lunarg (approved):

test/test_apps/external-memory-fd/export/app.cpp
```diff
@@ -162,36 +162,51 @@ int App::get_exportable_fd()
...
 }
 
 void App::send_exportable_fd(int exportable_fd)
 {
     // Need to send the fd to the importer process
     int external_socket = socket(PF_UNIX, SOCK_STREAM, 0);
-    GFXRECON_ASSERT(external_socket >= 0)
+    if (external_socket < 0)
+    {
+        GFXRECON_LOG_ERROR("Export App Failed to create socket (%d)", external_socket);
```
antonio-lunarg, line 175: `GFXRECON_LOG_ERROR` is just logging the message without aborting. I think it would be useful to return early after these checks, otherwise I think we'll read the other messages from the lines below which might be misleading.

test/test_apps/external-memory-fd/import/app.cpp
```diff
@@ -276,11 +287,17 @@ void App::cleanup()
...
+
+    close(import_socket_);
 }
 
 void App::setup()
 {
-    create_buffer_from_fd(receive_importable_fd());
+    import_socket_ = create_import_socket();
+    int import_fd = receive_importable_fd(import_socket_);
+    if(import_fd >= 0) {
```
antonio-lunarg, line 298: I think you can already close the socket after receiving the `fd`

test/test_apps/external-memory-fd/import/app.cpp
```diff
@@ -276,11 +287,17 @@ void App::cleanup()
...
     init.disp.freeMemory(imported_memory_, nullptr);
+
+    close(import_socket_);
 }
 
 void App::setup()
 {
-    create_buffer_from_fd(receive_importable_fd());
+    import_socket_ = create_import_socket();
+    int import_fd = receive_importable_fd(import_socket_);
```
antonio-lunarg, line 297: I think we want to check both `import_socket_` and `import_fd` and abort here. Running in a testing environment, we want to make sure that the return value of this process is not `0`, otherwise it will count as a _success_.

test/test_apps/external-memory-fd/export/app.cpp
bradgrantham-lunarg: We need a GFXRECON_LOG_FATAL that is actually fatal, maybe that `throw`s.  In the meantime I'll add a `throw` after these unrecoverable errors.

test/test_apps/external-memory-fd/import/app.cpp
bradgrantham-lunarg: Added - Good enough?

PR #2123, opened 2025-04-08
Add table to indicate support for various features

MarkY-LunarG:
This table indicates GFXR support on various platforms for various features.
This change incorporates feedback from other engineers as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,51 @@ submit a pull request.
 Please see the [CONTRIBUTING](CONTRIBUTING.md) file in this repository for
 more details
 
+## Supported Capabilities
+
+| Capability     |   Windows  |  Linux  |  Mac  |   Android   |
+| -------------- | :--------: | :-----: | :-------: | :-----: |
+| **Vulkan Support** |||||
+| Capture | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
+| Replay  | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
+| gfxrecon-compress | :white_check_mark: | :white_check_mark: | :white_check_mark: | :large_orange_diamond: |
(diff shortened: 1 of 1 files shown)


per-mathisen-arm:
Not sure what DX12 replay on Linux means to you here, but we use it regularly with DX emulation layers to try out desktop content on linux and android. What would also be useful, I think, would be to indicate which APIs and platforms are supported in a portable manner (ie Vulkan only for now, i guess).

MarkY-LunarG:
> Not sure what DX12 replay on Linux means to you here, but we use it regularly with DX emulation layers to try out desktop content on linux and android. What would also be useful, I think, would be to indicate which APIs and platforms are supported in a portable manner (ie Vulkan only for now, i guess).

Thanks for the feedback.  For now, I wasn't including portability items.  Another similar scenario is that  some Android traces replay on desktop with certain settings, but I know that's kind of hidden in the current table right now.  Let me think about how to display this.

MarkY-LunarG:
Updated with a new section describing issues with portability.

README.md
```diff
@@ -35,6 +35,51 @@ submit a pull request.
...
+
+**Legend**
+* :white_check_mark: : Supported
+* :x: : Not Suported
+* :large_orange_diamond: : Not supported on Android, but Desktop tools support modifying Android captures
+* :construction: : Early Preview / Alpha / Beta
+
+### Cross-Platform Capture/Replay Support
+
+Cross platform support, such as capturing on one operating system and
```
bradgrantham-lunarg, line 66: I suggest inverting the sense of this sentence, maybe something like "The table above only represents replay on the same platform and driver version on which the file was captured."

bradgrantham-lunarg (approved):
LGTM except the cross-platform sentence and I don't feel super strongly about that.

PR #2124, opened 2025-04-08
fix: Use dst_format when reading images

antonio-lunarg:
The dst_format attribute should be used when is NOT undefined.


antonio-lunarg:
Closing this as it's been included in https://github.com/LunarG/gfxreconstruct/pull/2164

PR #2128, opened 2025-04-09
Resolve issues around DebugUtilsCallback()

### LINKED ISSUE #2112: print EXT_debug_utils messages on replay
Certain layers/tools/apps emit debug messages through these channels. RenderDoc, for example, emits all its error messages through here, including the errors when it fails to load its capture layer. gfxr should print these messages on replay for improved debugging.

nickdriscoll-lunarg:
Resolves #2112 

`DebugUtilsCallback` will silently omit messages that don't set `pCallbackData->pMessageIdName`. This causes an issue with layers that don't set that field, such as RenderDoc

This PR makes the function print `(nullptr)` in place of `pCallbackData->pMessageIdName` when it is null.

Additionally, this PR unconditionally registers the debug callback so that messages from layers are displayed even when the captured app didn't set the debug callback, and sets the callback severity based on a new option `--debug-messenger-level`

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -121,6 +121,8 @@ static VKAPI_ATTR VkBool32 VKAPI_CALL DebugUtilsCallback(VkDebugUtilsMessageSeve
     if (pCallbackData->pMessageIdName != nullptr)
         message_id_name = pCallbackData->pMessageIdName;
 
+    GFXRECON_LOG_INFO("WE ARE IN THE DEBUG CALLBACK UNCONDITIONAL SECTION");
+
     if ((pCallbackData != nullptr) && (pCallbackData->pMessage != nullptr))
     {
         if (messageSeverity & VK_DEBUG_UTILS_MESSAGE_SEVERITY_ERROR_BIT_EXT)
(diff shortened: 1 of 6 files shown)


nickdriscoll-lunarg:
Converting this to a draft as it turns out we want replay to unconditionally register a debug callback so that layers can print their debug messages, and that's a nontrivial change.

framework/decode/vulkan_replay_consumer_base.h
```diff
@@ -1715,6 +1715,8 @@ class VulkanReplayConsumerBase : public VulkanConsumer
     util::ThreadPool main_thread_queue_;
     util::ThreadPool background_queue_;
 
+    VkDebugUtilsMessengerEXT debug_messenger;
```
antonio-lunarg, line 1718: nit: `debug_messenger_`

antonio-lunarg (approved):

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -116,24 +116,28 @@ static VKAPI_ATTR VkBool32 VKAPI_CALL DebugUtilsCallback(VkDebugUtilsMessageSeve
 {
     GFXRECON_UNREFERENCED_PARAMETER(pUserData);
 
-    if ((pCallbackData != nullptr) && (pCallbackData->pMessageIdName != nullptr) &&
-        (pCallbackData->pMessage != nullptr))
+    // Allow pCallbackData->pMessageIdName to be nullptr by defining a default string for message id name
+    const char* message_id_name = "(nullptr)";
+    if (pCallbackData->pMessageIdName != nullptr)
```
bradgrantham-lunarg, line 121: Please (always) put braces around the content of the blocks in an if-then-else.

bradgrantham-lunarg (commented):
Looks good to me.  (pending CI passing)

bradgrantham-lunarg (approved):

PR #2130, opened 2025-04-10
Add non-standard YUV format Android hardware buffer support

zehao-jing-arm:
Convert the non-standard YUV format to a standard RGBA format when capturing

Change-Id: Ida31e147721ba363edd2106edd7f150e7dd41dbb


MarkY-LunarG:
@ziga-lunarg , I know you modified AHB behavior lately so can you look this over and let us know what needs to be updated here?

framework/util/ahardwarebuffer_format_converter.cpp
```diff
@@ -0,0 +1,1014 @@
...
+    262203,     8,          9,          3,          589849,     10,         6,          1,          0,
+    0,          0,          1,          0,          196635,     11,         10,         262176,     12,
+    0,          11,         262203,     12,         13,         0,          262167,     15,         6,
+    2,          262176,     16,         1,          15,         262203,     16,         17,         1,
+    327734,     2,          4,          0,          3,          131320,     5,          262205,     11,
+    14,         13,         262205,     15,         18,         17,         327767,     7,          19,
+    14,         18,         196670,     9,          19,         65789,      65592
+};
+
+VkResult AHardwareBufferFormatConverter::ConvertImageFormat(VkQueue                                   queue,
```
antonio-lunarg, line 152: Is this doing the same thing done here? https://github.com/LunarG/gfxreconstruct/pull/2130/files#diff-a5b5ad66c57c3c303336581e4024b728c144b1ac6a8427bc295d2a821ec25443R241

I like this PR for two reasons:
1. it defers the image conversion operation at vkCreateImage time, instead of doing it at AllocateMemory time.
2. the whole process is encapsulated in its own helper class.

But, we need to think thoroughly about doing all of this at capture time, because converting every structure means that we are going to lose information about the original app behaviour. I'd like other opinions on that matter: @bradgrantham , @mikes-lunarg , @ziga-lunarg , @MarkY-LunarG

antonio-lunarg: Also, I am not sure doing the conversion _while capturing_ is the right way to go. We're still on the original device, which supports the current non-standard YUV format, therefore converting it to RGBA8_UNORM might be unnecessary and slow down the application.

MarkY-LunarG:
From a GFXReconstruct standpoint, we really should not be messing with the behavior during capture if we can avoid it. 
 GFXR can, and does, change behavior during replay, but we try to capture the contents as they were presented at capture time.  If this was all done at replay time, I would consider moving forward, but the fact that it's capture-time seems to be a no-go for me.  @bradgrantham-lunarg , do you agree?

bradgrantham-lunarg:
> From a GFXReconstruct standpoint, we really should not be messing with the behavior during capture if we can avoid it. GFXR can, and does, change behavior during replay, but we try to capture the contents as they were presented at capture time. If this was all done at replay time, I would consider moving forward, but the fact that it's capture-time seems to be a no-go for me. @bradgrantham-lunarg , do you agree?

Yes, capture should attempt to pass through and save the actual commands the application wrote unless we absolutely can't (like in the case of non-CPU-readable AHB).  If the AHB contents can't be fetched and stored in capture, then this PR may be the solution.  But if we can convert the contents on replay, that's a better place to do it.  Another option is to make *additional* Vulkan commands in capture if we will need further information to replay, like we do (I think) with memory heaps, and then store that information in MetaData blocks.  But any change to Vulkan commands or structs or parameters hides the application's original operation and also causes the capture file to represent less the application behavior when run without capture.

MarkY-LunarG:
If this is a convert that requires HW during the capture time, I would suggest capturing the content as an RGBA surface separately (like with a FillMemory metadata command).  Or maybe we need a new command indicating a conversion occurred and here's the converted data.  But that would be more of an informative block before the FillMemory.

MarkY-LunarG:
We recently renewed this discussion and I am leaning towards making a change to this.  However, I would like a "mode" we can set the capture layer to via a setting, like a `force_portable` mode. This way the conversion behavior is only done when the user knowingly wants this capture to be saved  in a portable manner, and they are willingly doing something that could change the behavior.

Does this sound reasonable @zehao-jing-arm  (@per-mathisen-arm )?

[Update]
Forget that.  After talking to Antonio he said:
 1) This is needed because we can't get data any other way
 2) We know how to identify this scenario and only do it then.

So carry  on.

antonio-lunarg:
> We recently renewed this discussion and I am leaning towards making a change to this. However, I would like a "mode" we can set the capture layer to via a setting, like a `force_portable` mode. This way the conversion behavior is only done when the user knowingly wants this capture to be saved in a portable manner, and they are willingly doing something that could change the behavior.
> 
> Does this sound reasonable @zehao-jing-arm (@per-mathisen-arm )?
> 
> [Update] Forget that. After talking to Antonio he said:
> 
> 1. This is needed because we can't get data any other way
> 2. We know how to identify this scenario and only do it then.
> 
> So carry on.

@zehao-jing-arm , please make sure whether gfxreconstruct already provides the non-standard formats related to this PR before continuing on this path. I would gladly see a design refactor, but make sure the app is still using the original formats while capturing. Refer to this PR for anything that gfxreconstruct is doing already: https://github.com/LunarG/gfxreconstruct/pull/2130

PR #2133, opened 2025-04-11
Add `--screenshot-interval` option

marius-pelegrin-arm:
In some case people want to be able to take a screenshot "every N frame" over long frame ranges or over the entire capture.

For now the only way to do this was either to take all screenshots and discard unused ones afterwards - which increase significantly the run time - or to write a very long list manually or assisted with a script, which is not very user-friendly.

This commit implements this with an option `--screenshot-interval N` where a screenshot is taken every `N` frame within screenshot ranges.

The default behavior stays the same if `--screenshot-interval` is `1`, which is the default value.

Example: If the screenshot ranges are `10-15,20-30` and the screenshot interval is `3`, then screenshots will be taken at frames 10, 13, 20, 23, 26 and 29.

### DIFF
--- a/USAGE_android.md
+++ b/USAGE_android.md
@@ -706,6 +706,7 @@ usage: gfxrecon.py replay [-h] [--push-file LOCAL_FILE] [--version]
                           [--paused] [--screenshot-all] [--screenshots RANGES]
                           [--screenshot-format FORMAT] [--screenshot-dir DIR]
                           [--screenshot-prefix PREFIX] [--screenshot-scale SCALE]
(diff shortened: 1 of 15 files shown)


MarkY-LunarG:
Ugh, GitHub asked me to fix the file in the GUI.  Didn't turn out so well.  I'll fix it on my machine and re-push.

MarkY-LunarG:
Ok.  Fixed the "merge commit" issue that was created by using GitHub to fix the issues with USAGE.md.  Sigh.

PR #2135, opened 2025-04-14
encode: Always store info about acquired images

antonio-lunarg:
The capture layer is only storing this info when _tracking_. With this, we always store info about acquired images.
Test app and little test fix included as well.


framework/encode/vulkan_capture_manager.h
```diff
@@ -508,11 +508,34 @@ class VulkanCaptureManager : public ApiCaptureManager
...
-            state_tracker_->TrackAcquireImage(*index, swapchain, semaphore, fence, 0);
+            if (IsCaptureModeTrack())
+            {
+                assert((state_tracker_ != nullptr) && (index != nullptr));
+                state_tracker_->TrackSemaphoreSignalState(semaphore);
+                state_tracker_->TrackAcquireImage(*index, swapchain, semaphore, fence, 0);
+            }
+            else
+            {
+                auto wrapper = vulkan_wrappers::GetWrapper<vulkan_wrappers::SwapchainKHRWrapper>(swapchain);
```
mikes-lunarg, line 521: Can we just call TrackAcquireImage here instead of duplicating the contents?

framework/encode/vulkan_capture_manager.h
```diff
@@ -521,15 +544,40 @@ class VulkanCaptureManager : public ApiCaptureManager
...
+                state_tracker_->TrackSemaphoreSignalState(pAcquireInfo->semaphore);
+                state_tracker_->TrackAcquireImage(*index,
+                                                  pAcquireInfo->swapchain,
+                                                  pAcquireInfo->semaphore,
+                                                  pAcquireInfo->fence,
+                                                  pAcquireInfo->deviceMask);
+            }
+            else
+            {
+                auto wrapper =
```
mikes-lunarg, line 562: Same question as above

test/test_apps/common/test_app_base.cpp
```diff
@@ -2759,7 +2760,8 @@ create_surface_headless(VkInstance instance, vkb::InstanceDispatchTable& disp, V
 void create_swapchain(SwapchainBuilder& swapchain_builder, Swapchain& swapchain)
 {
     auto new_swapchain = swapchain_builder.set_old_swapchain(swapchain).build();
-    destroy_swapchain(swapchain);
+    // Do not destroy old swapchain right away, as its images might be currently acquired and are still valid
+    // destroy_swapchain(swapchain);
```
mikes-lunarg, line 2764: When will old swapchain get destroyed for other test apps which rely on this behavior? I think it will create a validation error for them when the device is destroyed.

test/test_apps/acquired-image/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+# IN THE SOFTWARE.
+#
+# Author: LunarG Team
+# Description: CMake script for acquired image test app
+###############################################################################
+
+set(TEST_NAME acquired-image)
```
mikes-lunarg, line 27: This is cool trick, I'm going to steal this for the test I'm writing

framework/encode/vulkan_capture_manager.h
antonio-lunarg: Maybe not in the state tracker object? I see it's [only created when tracking](https://github.com/LunarG/gfxreconstruct/blob/f296bb1f219ad3d6a2051b8595e48dbe243fd189/framework/encode/capture_manager.cpp#L167)
Do you think we can extract this somewhere else? Can it be a method of `SwapchainKHRWrapper`?

test/test_apps/common/test_app_base.cpp
antonio-lunarg: Yes, need to fix this. I was thinking about a `SwapchainBuilder` parameter to configure that..

mikes-lunarg (commented):
I've thought about it a little more and can see are 2 additional issues:
- This change doesn't add the tracking for present operations, i.e. how `is_acquired` gets set back to `false`
- I think there is an interaction with VK_KHR_shared_presentable_image that unfortunately makes this even more complex, as that extension allows the application to access a swapchain image after it has been presented.

test/test_apps/common/test_app_base.cpp
antonio-lunarg: Done

framework/encode/vulkan_capture_manager.h
antonio-lunarg: Extracted in a subset function

mikes-lunarg:
> > I think there is an interaction with VK_KHR_shared_presentable_image that unfortunately makes this even more complex, as that extension allows the application to access a swapchain image after it has been presented.
> 
> Do you think we should solve this potential issue in this PR?

I would say only if the capture/issue in question is using that extension, otherwise probably best to address separately.

mikes-lunarg (approved):

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -2289,6 +2289,107 @@ void VulkanCaptureManager::PostProcess_vkCreateSwapchainKHR(VkResult
...
+            assert((state_tracker_ != nullptr) && (index != nullptr));
+            state_tracker_->TrackSemaphoreSignalState(semaphore);
+            state_tracker_->TrackAcquireImage(*index, swapchain, semaphore, fence, 0);
+        }
+        else
+        {
+            TrackAcquireImageSubset(swapchain, *index);
+        }
+    }
+}
```
fabian-lunarg, line 2326: I know those are separate entry-points and is certainly clean. but we do repeat the same logic.
we 'could' just assemble an adhoc `VkAcquireNextImageInfoKHR` here and pass to `PostProcess_vkAcquireNextImage2KHR`, which would avoid some duplication (but might be more confusing!?).

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -2289,6 +2289,107 @@ void VulkanCaptureManager::PostProcess_vkCreateSwapchainKHR(VkResult
     }
 }
 
+// Performs a subset of the state tracking performed by VulkanStateTracker::TrackAcquireImage,
+// only storing values needed for non-tracking capture.
+void TrackAcquireImageSubset(VkSwapchainKHR swapchain, size_t image_index)
+{
+    auto* wrapper = vulkan_wrappers::GetWrapper<vulkan_wrappers::SwapchainKHRWrapper>(swapchain);
+    GFXRECON_ASSERT(wrapper != nullptr);
```
fabian-lunarg, line 2297: is an additional check `if(wrapper != nullptr)` justified here (or in general when we only have `GFXGFXRECON_ASSERT`)? I mean, it's good we eventually run into those asserts in Debug, but would crash in Release.

antonio-lunarg:
I've integrated the test in the CI.

wpaul-samsung:
@bradgrantham-lunarg Is this ready to merge?

bradgrantham-lunarg (approved):

PR #2142, opened 2025-04-16
Support MetaCommand for Trim

JerryAMD:
GFXR have to track the initialize meta command parameters during capture. New meta data block added to support the feature.
During replay, gfxr have to inject helper commandlist to call InitializeMetaCommand()


framework/encode/dx12_state_writer.cpp
```diff
@@ -696,6 +696,52 @@ void Dx12StateWriter::WriteResourceCreationState(
...
+            init_meta_command.capture_id                          = wrapper->GetCaptureId();
+            init_meta_command.initialization_parameters_data_size = wrapper_info->initialize_parameters->GetDataSize();
+            init_meta_command.total_number_of_initializemetacommand = metacommand_wrappers.size();
+            init_meta_command.block_index                           = ++block_index;
+
+            output_stream_->Write(&init_meta_command, sizeof(init_meta_command));
+            output_stream_->Write(wrapper_info->initialize_parameters->GetData(),
+                                  wrapper_info->initialize_parameters->GetDataSize());
+        }
+    }
```
locke-lunarg, line 742: `InitializeMetaCommand` could happen after starting to write. It should check if  the MetaCommand is initialized before writing it.

JerryAMD: If the `InitializeMetaCommand` happen after state writing, GFXR will dump it to trace file with its CommandList. So, we won't lost it.

JerryAMD:
Any idea why Ubuntu release build failed, but debug build passed?

locke-lunarg:
> Any idea why Ubuntu release build failed, but debug build passed?

These errors are about code format checking. Your code format didn't match clang format.

JerryAMD:
> r



> > Any idea why Ubuntu release build failed, but debug build passed?
> 
> These errors are about code format checking. Your code format didn't match clang format.

I did format the code in VS before commit the change. And why is debug build OK?

locke-lunarg:
> I did format the code in VS before commit the change. And why is debug build OK?
Because debug build didn't check format. I'm not sure why your format was failed. It might be the different clang version.

You could use git tool and this error log to fix the format. 
1. Go to the failed CI run (e.g. "Ubuntu GCC Release")
2. Click the gear at the upper right and "View Raw Logs"
3. Copy the URL in the address bar
4. Run this command in Git Bash from the root of your repo with the command-line on your branch: 
`cd /c/gfxreconstruct`
`curl 'the-url-you-copied' |  cut -c 30- | git apply -p0 -`

framework/encode/dx12_state_writer.cpp
locke-lunarg: I mean that `wrapper_info->initialize_parameters` could be empty since it hasn't initialized yet.

JerryAMD: Added a flag to identify the MetaCommand has been initialized.

locke-lunarg: Thanks.

locke-lunarg (approved):

PR #2147, opened 2025-04-16
 openxr tutorial4 support

jzulauf-lunarg:
Additional openxr capabilities to support Openxr Tutorial 4

### DIFF
--- a/framework/decode/openxr_replay_consumer_base.cpp
+++ b/framework/decode/openxr_replay_consumer_base.cpp
@@ -893,6 +893,106 @@ void OpenXrReplayConsumerBase::Process_xrCreateDebugUtilsMessengerEXT(
     CheckResult("xrCreateDebugUtilsMessengerEXT", returnValue, replay_result, call_info);
 }
 
+void OpenXrReplayConsumerBase::Process_xrEnumerateEnvironmentBlendModes(
+    const ApiCallInfo&                      call_info,
+    XrResult                                returnValue,
+    format::HandleId                        instance,
+    format::HandleId                        systemId,
+    XrViewConfigurationType                 viewConfigurationType,
+    uint32_t                                environmentBlendModeCapacityInput,
+    PointerDecoder<uint32_t>*               environmentBlendModeCountOutput,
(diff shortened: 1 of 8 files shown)


framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -70,10 +70,70 @@ void OpenXrReplayConsumerBase::SetVulkanReplayConsumer(VulkanReplayConsumerBase*
...
+static XRAPI_ATTR XrResult XRAPI_CALL
+ReplayXrTermSubmitDebugUtilsMessageEXT(XrInstance                                  instance,
+                                       XrDebugUtilsMessageSeverityFlagsEXT         messageSeverity,
+                                       XrDebugUtilsMessageTypeFlagsEXT             messageTypes,
+                                       const XrDebugUtilsMessengerCallbackDataEXT* callbackData)
+{
+    // This is a stub to prevent crashes during replay, see AddInstanceTable below.
+    return XR_SUCCESS;
+}
+
```
MarkY-LunarG, line 102: Don't we also need stubs for the remaining entrypoints in that extension?  Otherwise they may get triggered and crash:
 * xrSessionBeginDebugUtilsLabelRegionEXT
 * xrSessionEndDebugUtilsLabelRegionEXT
 * xrSessionInsertDebugUtilsLabelEXT

jzulauf-lunarg: For some reason, the OpenXR loader doesn't terminate them, I mirrored its behavior.  I can add them if preferred.

MarkY-LunarG: I think we should, just in case the loader gets updated since they've shown a desire to do it for the other entrypoints of the same extension.  Just to be safe.

bradgrantham-lunarg (approved):
Approved pending resolution of https://github.com/LunarG/gfxreconstruct/pull/2147#discussion_r2049276975

MarkY-LunarG (approved):

PR #2148, opened 2025-04-16
Marky fix disable openxr

MarkY-LunarG:
Two commits:

```
openxr: Add ability to disable OpenXR in gradle

Add the ability to disable OpenXR when building the GFXR layer
in Gradle.
Also add some better error messages.
```

```
Fix android test build with Gradle

The build failed for me because some components were built with
AndroidX packages and others were not.
```

### DIFF
--- a/android/layer/build.gradle
+++ b/android/layer/build.gradle
@@ -28,6 +28,7 @@ android {
                 arguments "-DANDROID_TOOLCHAIN=clang", "-DANDROID_STL=c++_static"
                 if (project.hasProperty("DisableOpenXR")) {
                     logger.info('GFXReconstruct layer built with NO OpenXr Support! [DisableOpenXR found]')
+                    arguments.add("-DGFXRECON_ENABLE_OPENXR=OFF")
                 }
                 else {
                     logger.info('GFXReconstruct layer built with OpenXr Support.')
--- a/android/tools/quest_replay/CMakeLists.txt
+++ b/android/tools/quest_replay/CMakeLists.txt
@@ -65,5 +65,6 @@ target_link_libraries(
         android
         log
         OpenXR::openxr_loader)
-
+else()
+    message(WARNING "OpenXR support disabled!")
 endif()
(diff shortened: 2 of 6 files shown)


jzulauf-lunarg (approved):
LGTM

beau-lunarg (approved):

android/layer/build.gradle
```diff
@@ -28,6 +28,7 @@ android {
                 arguments "-DANDROID_TOOLCHAIN=clang", "-DANDROID_STL=c++_static"
                 if (project.hasProperty("DisableOpenXR")) {
```
bradgrantham-lunarg, line 29: Is this documented anywhere?  Like, how do I set "DisableOpenXR" and rebuild so that it is not included?

android/tools/quest_replay/CMakeLists.txt
```diff
@@ -65,5 +65,6 @@ target_link_libraries(
         android
         log
         OpenXR::openxr_loader)
-
+else()
```
bradgrantham-lunarg, line 68: I think this might be better conditionally included in android/settings.gradle, if that's possible.

bradgrantham-lunarg (approved):
Okay to get this in as coded but want to come back around to conditionally including android/tools/quest_replay/CMakeLists.txt instead of wrapping all its contents in an `if()`.

android/layer/build.gradle
MarkY-LunarG: Ugh, no.  Good idea, though.  I'll update the BUILD.md file with information on disabling the OpenXR build in the various ways.

android/tools/quest_replay/CMakeLists.txt
MarkY-LunarG: Are you talking about updating the settings.gradle like this?

```
if (project.hasProperty("DisableOpenXR")) {
    logger.info('GFXReconstruct Quest Replay app skipped [DisableOpenXR found]')
}
else {
    include ':quest_replay'
    project(':quest_replay').projectDir = file('tools/quest_replay')
}
```

That did not work.  Unless you know how to make it work.

android/layer/build.gradle
MarkY-LunarG: Added documentation.

PR #2149, opened 2025-04-17
Update support table for OpenXR

MarkY-LunarG:


jzulauf-lunarg (approved):
Matches my understanding of the state of OXR

PR #2154, opened 2025-04-17
Restrict replay symbols exposed.

MarkY-LunarG:
Complete issue #2071.  This continues some of the work started with #2098 by extending the symbol removal to tools.


MarkY-LunarG:
@fabian-lunarg / @bradgrantham can one of you verify these changes on your project since I try to expose only the symbols  you need?  Also, it's only for Android replayer, right?  Or do we need to do the same for desktop?

bartosz-muszarski-arm:
The issue I reported https://github.com/LunarG/gfxreconstruct/issues/2071 was discovered on a desktop build, haven't tried android.

bradgrantham-lunarg:
I don't have a way to verify at the moment.  I'll have to request that from @fabian-lunarg or @mikes-lunarg or @antonio-lunarg .

fabian-lunarg:
this looks fine to me and works locally. a rebase will be required to pass CI (due to some unrelated dump-resource baseline change).

> can one of you verify these changes on your project since I try to expose only the symbols you need?

I think it does, but maybe I'm missing something :) is there anything specific I should verify?

MarkY-LunarG:
Good catch @fabian-lunarg, I needed to add ANativeActivity_onCreate to the map.  However, it also seems that adding symbols to the Map alone was not enough.  Some symbols need to be both exported from the Map and from the CMakelists.txt.  I updated all the apps with .map files and symbols where appropriate.

fabian-lunarg (approved):

bradgrantham-lunarg (approved):
Test out by Samsung, seems to work for them.

MarkY-LunarG:
Thanks @bradgrantham-lunarg

bartosz-muszarski-arm:
I encountered the issue on the arm64 desktop version of the replayer when used with a particular instrumentation stack.
The symptom is a `double free or corruption (!prev)` error originating from the glibc malloc implementation.
I need to generate new data to get more details on the exact symbol and will get back to you later this week.
I haven't tested the changes in this PR yet, but if they limit visibility on the desktop replayer, we should be good.

bradgrantham-lunarg:
@bartosz-muszarski-arm These changes block symbols in the stack trace on Android in logcat.  It's likely we will need to revert them.  If you can think of another way to accomplish the same thing, I want to hear it.

PR #2156, opened 2025-04-17
WIP: openxr: Fail Xr negotiation when vulkan not enabled

jzulauf-lunarg:
This is the xrNegotiateLayer... logic implemented, but for some reason the __system_property_find / __system_property_read_callback sequence always returns empty strings.  

Not sure if this is because I'm trying to read global properties or ??  Also, the parallel implementation of this in the XrLoader (PlatformUtilsGetAndroidSystemProperty) has and additional "failover" path I haven't checked.

@MarkY-LunarG @bradgrantham @mikes-lunarg  -- any ideas?

### DIFF
--- a/layer/trace_layer.cpp
+++ b/layer/trace_layer.cpp
@@ -21,6 +21,8 @@
 ** DEALINGS IN THE SOFTWARE.
 */
 
+#include "util/platform.h"
+
 #include PROJECT_VERSION_HEADER_FILE
 
 #include "layer/trace_layer.h"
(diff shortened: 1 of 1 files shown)


mikes-lunarg:
`enable_gpu_debug_layers` and `gpu_debug_layers` are settings and not properties.

It's the difference between
```
adb shell getprop debug.gfxrecon.capture_file
```
and
```
adb shell settings get global enable_gpu_debug_layers
```

@MarkY-LunarG and I were talking about this the other day -- there doesn't appear to be a native way to read the settings like there is for the properties.

mikes-lunarg:
I just looked at the fallback path in OXR and `__system_property_get` won't help either, its just an older version of the same api.

mikes-lunarg:
The old popen hack we used to use for properties doesn't seem to work for settings either:
```
FILE* pipe = popen("settings get global enable_gpu_debug_layers", "r");
```
It doesn't produce any output and generates a permissions error in logcat:
```
ActivityManager: Permission Denial: getCurrentUser() from pid=5897, uid=10413 requires android.permission.INTERACT_ACROSS_USERS
```

I wonder if you could create a temporary VkInstance and use it to query for the capture layer with VK_EXT_tooling_info. Kind of a pain, and it will generate noise in the capture file, but it would tell you whether or not the Vulkan capture layer is active.

jzulauf-lunarg:
This approach won't work.  Looking at other, ways for OpenXR capture to fail with informative error messaging and without crashing.

PR #2172, opened 2025-04-28
check that generated OpenXR code isn't out of date

bradgrantham-lunarg:


framework/generated/khronos_generators/openxr_generators/openxr_base_generator.py
```diff
@@ -171,7 +171,7 @@ def __init__(
             extra_manual_commands=extra_manual_commands,
         )
 
-        self.begin_end_file_data.guards.append(('ifdef', 'ENABLE_OPENXR_SUPPORT'))
```
mikes-lunarg, line 174: Wow looks like you already caught something!

mikes-lunarg (approved):

PR #2178, opened 2025-04-29
Refactor Vulkan trace layer

davidd-lunarg:


framework/generated/generated_vulkan_layer_func_table.h
```diff
@@ -47,696 +47,700 @@
...
-    { "vkCmdSetRayTracingPipelineStackSizeKHR",                                                              reinterpret_cast<PFN_vkVoidFunction>(encode::vkCmdSetRayTracingPipelineStackSizeKHR) },
-    { "vkCmdDrawMeshTasksEXT",                                                                               reinterpret_cast<PFN_vkVoidFunction>(encode::vkCmdDrawMeshTasksEXT) },
-    { "vkCmdDrawMeshTasksIndirectEXT",                                                                       reinterpret_cast<PFN_vkVoidFunction>(encode::vkCmdDrawMeshTasksIndirectEXT) },
-    { "vkCmdDrawMeshTasksIndirectCountEXT",                                                                  reinterpret_cast<PFN_vkVoidFunction>(encode::vkCmdDrawMeshTasksIndirectCountEXT) },
-    { "vk_layerGetPhysicalDeviceProcAddr",                                                                   reinterpret_cast<PFN_vkVoidFunction>(vulkan_entry::GetPhysicalDeviceProcAddr) },
-};
+GFXRECON_BEGIN_NAMESPACE(vulkan_entry_layer)
+static const std::unordered_map<std::string, PFN_vkVoidFunction> GetVulkanFuncTableLayer() {
```
jzulauf-lunarg: Should this be inline instead?  Won't every includer have to initialize the map?

framework/generated/khronos_generators/khronos_layer_func_table_generator.py
```diff
@@ -30,32 +30,51 @@ class KhronosLayerFuncTableGenerator():
...
+            skip_func_namespace += '_{}'.format(table_name_suffix.lower())
+            get_table_name += '{}'.format(table_name_suffix)
+
+        # Write the static function to get the table.
         write(
-            'const std::unordered_map<std::string, {}> {}_func_table = {{'.
-            format(api_data.void_func_pointer_type, api_data.api_name.lower()),
-            file=self.outFile
-        )
+            'static const std::unordered_map<std::string, {}> {}() {{'.format(
```
jzulauf-lunarg: This is probably the correct place to make the "shouldn't this be inline" comment?  

Not for this PR, but if this is common included, a .cpp location would be yet better.

layer/vulkan_entry_layer.cpp
```diff
@@ -742,9 +678,9 @@ extern "C"
         // Fill in the function pointers if our version is at least capable of having the structure contain them.
         if (pVersionStruct->loaderLayerInterfaceVersion >= 2)
         {
-            pVersionStruct->pfnGetInstanceProcAddr       = gfxrecon::vulkan_entry::GetInstanceProcAddr;
-            pVersionStruct->pfnGetDeviceProcAddr         = gfxrecon::vulkan_entry::GetDeviceProcAddr;
-            pVersionStruct->pfnGetPhysicalDeviceProcAddr = gfxrecon::vulkan_entry::GetPhysicalDeviceProcAddr;
+            pVersionStruct->pfnGetInstanceProcAddr       = gfxrecon::vulkan_entry_layer::GetInstanceProcAddr;
```
jzulauf-lunarg: I'm finding the naming quite confusing. The exact same name is both a virtual member of the  VulkanEntryLayer class and a raw VKAPI function in the vulkan_entry_layer namespace.  

Naming that differs exactly by only CamelCase vs. snake_case... not a fan.  Names should be more distinct (and permutations  i.e. layer_entry_vulkan  don't count c.f. https://en.wikiquote.org/wiki/Colossal_Cave_Adventure -- "all different maze"). Either guardstring the function (e.g. VkApi<foo> for the VKAPI_* version) or choose a namespace that doesn't match the class name except for case (api_entrypoints_vulkan might work).

jzulauf-lunarg (commented):
Some preferences for changes, for you consideration.  The naming really threw me for several minutes, though perhaps in the IDE it would have become clearer, faster.

layer/vulkan_entry_layer.cpp
davidd-lunarg: I propose the following naming scheme:

```
gfxrecon:: (namespace)
    encode:: (namespace)
        class VulkanEntryBase (in encode/vulkan_entry_base.h/cpp)
        
    vulkan_layer:: (namespace)
        class LayerVulkanEntry (in layer/layer_vulkan.entry.h/cpp)
        static inline ... GetVulkanLayerFuncTable (in generated/generated_vulkan_layer_func_table.h)
        
    // For context, in the future the following will be added:
    vulkan_recapture:: (namespace)
        RecaptureVulkanEntry (derived class, in tools/replay/recapture_vulkan.entry.h/cpp)
        static inline ... GetVulkanRecaptureFuncTable (in generated/generated_vulkan_recapture_func_table.h)
```
Where:
- The `VulkanEntry` classes encapsulate the logic used to customize behavior for Vulkan entry points. There will be separate entry point behavior between the capture layer and the `replay --recapture` feature.
- The `vulkan_layer` namespace contains Vulkan layer specific code and is only compiled into the `VkLayer_gfxreconstruct` project
- The `vulkan_recapture` namespace contains Vulkan recapture specific code and is only compiled into the `gfxrecon-replay` project

davidd-lunarg: The code has been changed to the above naming scheme. Also the `openxr_entry` namespace was renamed to `openxr_layer`.

framework/generated/khronos_generators/khronos_layer_func_table_generator.py
davidd-lunarg: Added inline to the function signature. Each instance of the generated file (`generated_vulkan_layer_func_table.h`, `generated_openxr_layer_func_table.h`, etc) should only be included once at the top-level compilation unit (dll or exe).

bradgrantham-lunarg (approved):

PR #2187, opened 2025-05-01
Fix D3D12 write-watch flag handling in capture

ryannecas:
D3D12_HEAP_FLAG_ALLOW_WRITE_WATCH is added to certain heaps upon creation during capture. However, ID3D12Heap::GetDesc() was both encoding the flag to the capture file and returning it back to the application. Fix with an override implementation

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -393,13 +393,13 @@ codesign -dvvv libVkLayer_gfxreconstruct.dylib`
 Apple's developer information about code-signing can be found here:
 https://developer.apple.com/library/archive/documentation/Security/Conceptual/CodeSigningGuide/Introduction/Introduction.html
 
-### Disabling OpenXR Inclusing For Desktop Builds
+### Disabling OpenXR Inclusion For Desktop Builds
 
 If there are any concerns about using the OpenXR content in the resulting components of GFXReconstruct,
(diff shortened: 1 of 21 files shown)


davidd-lunarg:
Thank you for the submission!

> ID3D12Heap::GetDesc() was both encoding the flag to the capture file

I agree this is wrong and is fixed by this PR

> . . . and returning it back to the application

Isn't the flag currently removed from the desc in `PostProcess_ID3D12Heap_GetDesc` before it is returned to the application? So this PR correctly preserves that behavior, and from the captured application's perspective there is no change.

> ID3D12Heap::GetDesc()

Is a similar change needed for `ID3D12Resource::GetHeapProperties`? If so, would you mind including a fix for that with this change?

ryannecas:
PostProcess_ID3D12Heap_GetDesc does remove the flag given by Dispatch. However, Dispatch is getting the D3D12_HEAP_DESC passed by value rather than reference, so the effect of clearing the flag doesn't make it back to ID3D12Heap_Wrapper::GetDesc

davidd-lunarg:
The CI check failed due to formatting that was modified in `dev`. Rebasing onto the latest `dev` commit should fix it.

davidd-lunarg (approved):

Issue #2166, opened 2025-04-23
VulkanRebindAllocator::WriteBoundResourceStaging hardcodes VK_IMAGE_ASPECT_COLOR_BIT


zmike:
This causes the `functions_.cmd_copy_buffer_to_image` call to be invalid for any depth/stencil rebinds.

zmike:
There should probably be a sweep to ensure there are no other cases of this.

locke-lunarg:
https://github.com/LunarG/gfxreconstruct/blob/dev/framework/decode/vulkan_rebind_allocator.cpp#L1684
Is here the place that you mention?

Excuse me, could you please share more information, or even a simple? For copy buffer to image in `VulkanRebindAllocator::WriteBoundResourceStaging`, it's used for `BindImageMemory`, and `BindImageMemory` usually happen before `CreateImageView`. In this case, the image shouldn't have aspect mask info. So I need more information to fix it. Thank you.

locke-lunarg:
@zmike Could you please test this PR https://github.com/LunarG/gfxreconstruct/pull/2192 to see if it's fixed? Thanks.

zmike:
I've tested, and the linked PR does not fix the issue I'm seeing.

RADV backtrace:

```
#0  0x00007ffff787ff54 in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007ffff7826f9e in raise () from /lib64/libc.so.6
#2  0x00007ffff780e942 in abort () from /lib64/libc.so.6
#3  0x00007ffff780e85e in __assert_fail_base.cold () from /lib64/libc.so.6
#4  0x00007ffff781f0c7 in __assert_fail () from /lib64/libc.so.6
#5  0x00007fffd7e74190 in vk_format_get_aspect_format (format=VK_FORMAT_D32_SFLOAT_S8_UINT, aspect=6) at ../src/vulkan/util/vk_format.c:605
#6  0x00007fffd78acba3 in vk_image_buffer_copy_layout (image=0x5f86040, region=0x7fffffffb2e0) at ../src/vulkan/runtime/vk_image.c:332
#7  0x00007fffd76799ab in copy_memory_to_image (cmd_buffer=0x30916a0, buffer_addr=18446603348374192128, buffer_size=42467336, src_copy_flags=0, 
[...]

PR #2190, opened 2025-05-02
Resolve image aspect in rebind allocator resource initialization

bartosz-muszarski-arm:
Currently the image aspect for copy operation is hardcoded as VK_IMAGE_ASPECT_COLOR_BIT. This can casue issues with processing depth and stencil data. This commit fixes the problem for single-planar formats by resolving aspect based on image format.


locke-lunarg:
This PR should solve this issue https://github.com/LunarG/gfxreconstruct/issues/2166

PR #2192, opened 2025-05-05
Get aspect mask by its format

### LINKED ISSUE #2166: VulkanRebindAllocator::WriteBoundResourceStaging hardcodes VK_IMAGE_ASPECT_COLOR_BIT

locke-lunarg:
Use GetFormatAspectMask, instead of VK_IMAGE_ASPECT_COLOR_BIT

closed: https://github.com/LunarG/gfxreconstruct/issues/2166

### DIFF
--- a/framework/decode/vulkan_rebind_allocator.cpp
+++ b/framework/decode/vulkan_rebind_allocator.cpp
@@ -59,6 +59,7 @@
 #include "util/alignment_utils.h"
 #include "util/platform.h"
 #include "graphics/vulkan_util.h"
+#include "graphics/vulkan_resources_util.h"
 #include "graphics/vulkan_struct_get_pnext.h"
 
 #include "generated/generated_vulkan_enum_to_string.h"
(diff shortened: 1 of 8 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -556,28 +557,35 @@ void VulkanReplayConsumerBase::ProcessFillMemoryCommand(uint64_t       memory_id
...
+                                                 image_info.extent,
+                                                 aspect,
+                                                 image_info.samples,
+                                                 image_info.usage,
+                                                 image_info.initialLayout,
+                                                 final_layout,
+                                                 image_info.arrayLayers,
+                                                 1,
+                                                 &copy_region);
+                }
```
locke-lunarg, line 588: Could someone please helps me to verify whether this change is correct.

fabian-lunarg: ok, this is guarded by `VK_USE_PLATFORM_ANDROID_KHR`, took me a while to realize :)

fabian-lunarg: I tried several android captures I had but none is exercising this code-path.
it seems this handling an AndroidHardBuffer case?

fabian-lunarg (approved):
the changes look good to me, overall cleaner, all CI tests pass.
I checked several cases on Desktop and Android and did not spot any differences
in terms of correctness or validation-warnings.
just curious when/where this caused problems -
is there a testcase for this or could we create a simple one?

bradgrantham-lunarg:
> just curious when/where this caused problems - is there a testcase for this or could we create a simple one?

Just reported in #2166.  We don't have a test case.  Locke, can you think for a minute whether there may be a capture file already in CI that exercises one of these code paths?  Or are there a few lines we could add to vkcube to make the code path run?  I don't think you have to go as far as making a sample program.

locke-lunarg:
@fabian-lunarg I couldn't write a testcase for it. I don't know how to make it happen. For issue https://github.com/LunarG/gfxreconstruct/issues/2166, the bug could happen with `-m rebind` in two cases. 

1. `VulkanReplayConsumerBase::ProcessFillMemoryCommand (for image)`, and the memory type is not `VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT (!resource_alloc_info->is_host_visible)`. 

2. `VulkanRebindAllocator::BindImageMemory`, and the memory type is not `VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT (!resource_alloc_info->is_host_visible)`, and the memory had been written by `ProcessFillMemoryCommand (memory_alloc_info->original_content != nullptr)`.

Although a depth image memory isn't host visible, but gfxr doesn't write depth image memory. `ProcessFillMemoryCommand` isn't for depth images. I couldn't write a testcase for it.

locke-lunarg:
https://github.com/LunarG/gfxreconstruct/pull/2190
This PR did something like I did. I will override some this PR.

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: found an android capture that exercises that code-path.
there were no apparent issues here, both with and w/o `-m rebind`

bradgrantham-lunarg (approved):
LGTM, just need to resolve those merge conflicts.

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -1678,37 +1679,23 @@ void VulkanRebindAllocator::WriteBoundResourceStaging(
...
-                    default:
-                        aspect = VK_IMAGE_ASPECT_COLOR_BIT;
-                        break;
+                    VkImageAspectFlags aspect_flags = aspect;
+
+                    VkBufferImageCopy region{};
+                    region.bufferOffset      = 0;
+                    region.bufferRowLength   = 0;
+                    region.bufferImageHeight = 0;
+                    region.imageSubresource  = { aspect_flags, 0, 0, 1 };
```
locke-lunarg, line 1693: Added this change. The aspectMask member of imageSubresource must only have a single bit set.

framework/decode/vulkan_replay_consumer_base.cpp
locke-lunarg: Thank you for testing.

PR #2197, opened 2025-05-07
Dump-Resources: Fix empty vertex-buffers

fabian-lunarg:
Use correct vertex-input data when using `VK_DYNAMIC_STATE_VERTEX_INPUT_EXT`

- retrieve information about dynamic_vertex_input from optional `VkPipelineLibraryCreateInfoKHR`

### DIFF
--- a/framework/decode/vulkan_replay_dump_resources.cpp
+++ b/framework/decode/vulkan_replay_dump_resources.cpp
@@ -2243,27 +2243,33 @@ void VulkanReplayDumpResourcesBase::DumpGraphicsPipelineInfos(
             }
         }
 
-        // Graphics pipeline library
-        const auto gpl_info =
+        // handle optional VkGraphicsPipelineLibraryCreateInfoEXT
+        const auto* gpl_info =
             graphics::vulkan_struct_get_pnext<VkGraphicsPipelineLibraryCreateInfoEXT>(in_p_create_infos);
         if (gpl_info != nullptr)
         {
             pipeline_info->gpl_flags = gpl_info->flags;
         }
 
-        const auto pl_info = GetPNextMetaStruct<Decoded_VkPipelineLibraryCreateInfoKHR>(create_info_meta->pNext);
-        if (pl_info != nullptr)
+        // handle optional VkPipelineLibraryCreateInfoKHR
+        const auto* pipeline_library_info =
+            GetPNextMetaStruct<Decoded_VkPipelineLibraryCreateInfoKHR>(create_info_meta->pNext);
+        if (pipeline_library_info != nullptr)
         {
-            const uint32_t          library_count = pl_info->pLibraries.GetLength();
-            const format::HandleId* ppl_ids       = pl_info->pLibraries.GetPointer();
+            const uint32_t          library_count = pipeline_library_info->pLibraries.GetLength();
+            const format::HandleId* ppl_ids       = pipeline_library_info->pLibraries.GetPointer();
+
             for (uint32_t lib_idx = 0; lib_idx < library_count; ++lib_idx)
             {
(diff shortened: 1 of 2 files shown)


fabian-lunarg:
> Could the application "forget" to do so when it creates the pipeline and just calls vkCmdSetVertexInputEXT?

that might be the case, looking into it. 
in the repro-case I have here, there are 'some' pipelines decorated with `VK_DYNAMIC_STATE_VERTEX_INPUT_EXT`
and their `pipeline_info->dynamic_vertex_input` is set. 
either this flag is missing for other pipelines (e.g. due to weird application behavior or legit spec, ...)
or this gets lost somewhere.

in the end I also think, when there is valid data from `vkCmdSetVertexInputEXT` we should just use it in any case.

panos-lunarg:
> that might be the case, looking into it.

Reading `vkCmdSetVertexInputEXT`'s description in the spec: 

> This command sets the vertex input attribute and vertex input binding descriptions state for subsequent drawing commands when drawing using [shader objects](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#shaders-objects), or when the graphics pipeline is created with VK_DYNAMIC_STATE_VERTEX_INPUT_EXT set in [VkPipelineDynamicStateCreateInfo](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#VkPipelineDynamicStateCreateInfo)::pDynamicStates. Otherwise, this state is specified by the [VkGraphicsPipelineCreateInfo](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#VkGraphicsPipelineCreateInfo)::pVertexInputState values used to create the currently active pipeline.

I interpret this as `vkCmdSetVertexInputEXT` should only be applied when the currently bound pipeline has been created with `VK_DYNAMIC_STATE_VERTEX_INPUT_EXT`.
The application can issue a `vkCmdSetVertexInputEXT` and bind a relative pipeline but this should apply only to draw calls done with that pipeline. If later a pipeline without dynamic vertex input is bound then the input from `vkCmdSetVertexInputEXT` should be ignored.
So I'm thinking that maybe `pipeline_info->dynamic_vertex_input` is correct and set in the right place.

fabian-lunarg:
I was digging through a converted capture .json and think I can see the issue.
there, we create a pipeline, bind it and issue `vkCmdSetVertexInput`. this pipeline does not have `VK_DYNAMIC_STATE_VERTEX_INPUT_EXT` set. 
but it does use a `VK_STRUCTURE_TYPE_PIPELINE_LIBRARY_CREATE_INFO_KHR` during creation, supplying 4 other pipelines.
in one of those `VK_DYNAMIC_STATE_VERTEX_INPUT_EXT` is indeed present, so we are supposed to inherit it from there.

panos-lunarg:
> I was digging through a converted capture .json and think I can see the issue. there, we create a pipeline, bind it and issue `vkCmdSetVertexInput`. this pipeline does not have `VK_DYNAMIC_STATE_VERTEX_INPUT_EXT` set. but it does use a `VK_STRUCTURE_TYPE_PIPELINE_LIBRARY_CREATE_INFO_KHR` during creation, supplying 4 other pipelines. in one of those `VK_DYNAMIC_STATE_VERTEX_INPUT_EXT` is indeed present, so we are supposed to inherit it from there.
> 
> so during pipeline-creation, we should probably iterate this pipeline-library in order to retrieve correct flags.

Good catch!

panos-lunarg:
There is logic for handling pipeline libraries in `VulkanReplayDumpResourcesBase::DumpGraphicsPipelineInfos`, just a few lines below where `pipeline_info->dynamic_vertex_input` is set. Perhaps this case can be handled there

panos-lunarg (approved):

bradgrantham-lunarg (approved):

PR #2198, opened 2025-05-07
Fix handling of portability extensions

### LINKED ISSUE #2132: VK_KHR_portability_subset should only be enabled on specific drivers
VK_KHR_portability_subset is (probably?) only required for MoltenVK usage. gfxreconstruct should check the driver id before enabling this extension.

nickdriscoll-lunarg:
Resolves #2132 

This PR implements two changes to `replay`:

-  If the replay device's `VkPhysicalDeviceDriverProperties::driverID` is _not_ MoltenVK, we omit the `VK_KHR_portability_subset` extension
-   If `vkCreateInstance` fails with `VK_ERROR_EXTENSION_NOT_PRESENT`, we remove the `VK_KHR_portability_enumeration` extension and try again. This is important because some tools, like RenderDoc, do not support even the enumeration extension.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -2885,19 +2885,14 @@ VulkanReplayConsumerBase::OverrideCreateInstance(VkResult original_result,
         create_state.modified_create_info.flags &= ~VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR;
 
         std::vector<const char*>& modified_extensions = create_state.modified_extensions;
-        for (int i = 0; i < modified_extensions.size(); ++i)
-        {
-            const char* name = modified_extensions[i];
-            if (!util::platform::StringCompare(name, VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME))
-            {
-                const size_t last_idx         = modified_extensions.size() - 1;
-                modified_extensions[i]        = modified_extensions[last_idx];
-                modified_extensions[last_idx] = name;
-                break;
-            }
-        }
-        create_state.modified_extensions.pop_back();
-        create_state.modified_create_info.enabledExtensionCount -= 1;
+        modified_extensions.erase(std::remove_if(modified_extensions.begin(),
+                                                 modified_extensions.end(),
+                                                 [](const char* ext) -> bool {
+                                                     return !util::platform::StringCompare(
+                                                         ext, VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME);
+                                                 }),
+                                  modified_extensions.end());
+        create_state.modified_create_info.enabledExtensionCount = modified_extensions.size();
 
         // Try to create instance again
         result = create_instance_proc_(
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -2697,8 +2697,7 @@ void VulkanReplayConsumerBase::ModifyCreateInstanceInfo(
     std::vector<VkExtensionProperties> available_extensions;
     if (feature_util::GetInstanceExtensions(instance_extension_proc, &available_extensions) == VK_SUCCESS)
     {
-        // Always enable portability enumeration if on Mac
-        #ifdef __APPLE__
+        // Always enable portability enumeration if available
         modified_create_info.flags &= ~VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR;
         for (const VkExtensionProperties& extension : available_extensions)
         {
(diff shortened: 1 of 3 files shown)


fabian-lunarg (approved):

fabian-lunarg:
> CI gfxreconstruct build # 6826 failed.

checked the failed CI run, was an unrelated issue (we swapped a build-machine), all good now

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2865,6 +2865,34 @@ VulkanReplayConsumerBase::OverrideCreateInstance(VkResult original_result,
...
+    if (result == VK_ERROR_EXTENSION_NOT_PRESENT)
+    {
+        // Assume we weren't able to create instance because of VK_KHR_portability_enumeration
+        GFXRECON_LOG_WARNING(
+            "Assuming unable to create instance due to VK_KHR_portability_enumeration. Attempting creation "
+            "without that extension...");
+        create_state.modified_create_info.flags &= ~VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR;
+
+        std::vector<const char*>& modified_extensions = create_state.modified_extensions;
+        for (int i = 0; i < modified_extensions.size(); ++i)
```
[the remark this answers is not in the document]
fabian-lunarg: agree, loop could be replaced with:

modified_extensions.erase(std::remove_if(modified_extensions.begin(),
                                                                            modified_extensions.end(),
                                                                            [](const char* ext) -> bool {
                                                                                return !util::platform::StringCompare(
                                                         ext, VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME);
                                                                            }),
                                                  modified_extensions.end());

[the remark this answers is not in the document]
fabian-lunarg: and then `create_state.modified_create_info.enabledExtensionCount = modified_extensions.size()` ?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2865,6 +2865,34 @@ VulkanReplayConsumerBase::OverrideCreateInstance(VkResult original_result,
...
+                modified_extensions[i]        = modified_extensions[last_idx];
+                modified_extensions[last_idx] = name;
+                break;
+            }
+        }
+        create_state.modified_extensions.pop_back();
+        create_state.modified_create_info.enabledExtensionCount -= 1;
+
+        // Try to create instance again
+        result = create_instance_proc_(
```
fabian-lunarg, line 2892: maybe here it makes sense to check if we actually removed `VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME`? (check if `enabledExtensionCount` has changed since last attempt)

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2865,6 +2865,34 @@ VulkanReplayConsumerBase::OverrideCreateInstance(VkResult original_result,
...
+            if (!util::platform::StringCompare(name, VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME))
+            {
+                const size_t last_idx         = modified_extensions.size() - 1;
+                modified_extensions[i]        = modified_extensions[last_idx];
+                modified_extensions[last_idx] = name;
+                break;
+            }
+        }
+        create_state.modified_extensions.pop_back();
+        create_state.modified_create_info.enabledExtensionCount -= 1;
```
fabian-lunarg, line 2889: `enabledExtensionCount` is decremented unconditionally, maybe better just use `modified_extensions.size()`?

framework/decode/vulkan_replay_consumer_base.cpp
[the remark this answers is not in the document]
nickdriscoll-lunarg: Made this change.

framework/decode/vulkan_replay_consumer_base.cpp
nickdriscoll-lunarg: What should be done if there was no change? Just log an error and let it crash on the next attempt?

fabian-lunarg: just logging an error seems appropriate. I mean we know we added the portability-extension, so 'expect' it gets removed here. the log-error 'could' remind us of that assumption if that changes one day.

bradgrantham-lunarg (approved):

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: this case basically does not exist. we always start with `VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME` in the extension list. 
I was thinking if that 'could' change one day, we shouldn't get confused here. 
so, not too relevant, was not an actual concern.

PR #2199, opened 2025-05-07
openxr: Remove relative time logic from decoders

jzulauf-lunarg:
Relative time logic is present in frame loop replay logic which should be sufficient.

Just capturing this as a WIP.  Haven't tested it, beyond showing that converts are now consistent run to run.

### DIFF
--- a/framework/decode/value_decoder.h
+++ b/framework/decode/value_decoder.h
@@ -89,36 +89,9 @@ class ValueDecoder
 #endif
 
 #if ENABLE_OPENXR_SUPPORT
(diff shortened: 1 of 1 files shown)


MarkY-LunarG (approved):

bradgrantham-lunarg (approved):
Has this been tested by running `convert` twice and making sure the results are identical for `XrTime`?

bradgrantham-lunarg:
> Haven't tested it, beyond showing that converts are now consistent run to run.

This sounds like it was tested.  I will approve (again) based on that.

bradgrantham-lunarg (approved):

PR #2204, opened 2025-05-09
Handle xrEnumerateDisplayRefreshRatesFB incomplete

MarkY-LunarG:
Handle a different amount of display rates returned in the replay from the capture for the xrEnumerateDisplayRefreshRatesFB command.

### DIFF
--- a/framework/decode/openxr_replay_consumer_base.cpp
+++ b/framework/decode/openxr_replay_consumer_base.cpp
@@ -993,6 +993,94 @@ void OpenXrReplayConsumerBase::Process_xrEnumerateEnvironmentBlendModes(
     }
 }
 
+void OpenXrReplayConsumerBase::Process_xrEnumerateDisplayRefreshRatesFB(
+    const ApiCallInfo&        call_info,
+    XrResult                  returnValue,
+    format::HandleId          session,
+    uint32_t                  displayRefreshRateCapacityInput,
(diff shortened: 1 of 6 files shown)


framework/decode/openxr_replay_consumer_base.cpp
```diff
@@ -993,6 +993,94 @@ void OpenXrReplayConsumerBase::Process_xrEnumerateEnvironmentBlendModes(
...
+            if (replay_count < displayRefreshRateCapacityInput)
+            {
+                GFXRECON_LOG_FATAL(
+                    "xrEnumerateDisplayRefreshRatesFB failed to find as many items during replay as during capture");
+                return;
+            }
+
+            // Allocate the blend mode array and get all the values
+            std::vector<float> display_refresh_rates;
+            display_refresh_rates.resize(replay_count);
```
ziga-lunarg, line 1030: You could do this in one step with `std::vector<float> display_refresh_rates(replay_count);`

ziga-lunarg (approved):
Fixes the issue I was seeing

framework/decode/openxr_replay_consumer_base.cpp
bradgrantham-lunarg: And use `display_refresh_rates.size()` in the call to `pfn_enum_dis_refresh_rates_fb` instead of `replay_count`

PR #2207, opened 2025-05-12
android: Fix orientation when pre-rotation is set

antonio-lunarg:
We currently assume identity orientation of capture and replay device match and it is portrait, then when pre-rotation is requested, we set the orientation accordingly.

It might still show some issues with devices that have a landscape native orientation (either capture or replay device, or both), so my guess is that it will need more testing on different devices.


mikes-lunarg (approved):
LGTM. I retested both replay apps against the original report titles and the display is correct.

PR #2209, opened 2025-05-14
Align async pipeline creation with pipeline caches

fabian-lunarg:
Fix interop of CLI-options:

we had this warning since external pipeline-caches were introduced:

>Requested both asynchronous pipeline-creation (--pipeline-creation-jobs) and 
explicit pipeline-caches (--save-pipeline-cache | --load-pipeline-cache). This is
currently not supported and will prevent usage of pipeline-caches.

This PR consolidates the behavior across regular and asynchronous pipeline-creation
and removes the warning.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -227,14 +227,6 @@ VulkanReplayConsumerBase::VulkanReplayConsumerBase(std::shared_ptr<application::
             num_threads += (int32_t)std::thread::hardware_concurrency();
         }
         background_queue_.set_num_threads(std::clamp<uint32_t>(num_threads, 0, std::thread::hardware_concurrency()));
-
-        if (options_.add_new_pipeline_caches || !options_.save_pipeline_cache_filename.empty() ||
-            !options_.load_pipeline_cache_filename.empty())
-        {
-            GFXRECON_LOG_WARNING("Requested both asynchronous pipeline-creation (--pipeline-creation-jobs) and "
-                                 "explicit pipeline-caches (--save-pipeline-cache | --load-pipeline-cache). This is "
-                                 "currently not supported and will prevent usage of pipeline-caches.");
-        }
     }
 
     // If we want to save a pipeline cache file, we do this to be sure the file exists, is empty, and optionally, is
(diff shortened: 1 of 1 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -11044,11 +11054,26 @@ std::function<handle_create_result_t<VkPipeline>()> VulkanReplayConsumerBase::As
...
+                                pipeline_handles = out_pipelines.data(),
+                                num_pipelines    = out_pipelines.size(),
+                                handle_deps      = std::move(handle_deps)] {
+            // asynchronous operation is done. clear tracked handles, call deferred deletes
+            ClearAsyncHandles(handle_deps);
+
+            // if a pipeline cache was created, track it to know when to destroy it/save it to file
+            if (cache_pipeline_id != format::kNullHandleId)
+            {
+                TrackNewPipelineCache(device_info, cache_pipeline_id, pipeline_cache, pipeline_handles, num_pipelines);
```
MarkY-LunarG, line 11074: Is this necessary for graphics pipelines as well?  This commit seems to only touch compute.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -10620,21 +10612,16 @@ VkResult VulkanReplayConsumerBase::OverrideCreateGraphicsPipelines(
...
+                                  pipeline_cache,
                                   create_info_count,
                                   maybe_replaced_create_infos,
                                   in_p_allocation_callbacks,
                                   out_pipelines);
 
     // If a pipeline cache was created, track it to know when to destroy it/save it to file
-
-    if (in_pipeline_cache != override_pipeline_cache && replay_result == VK_SUCCESS)
+    if (cache_pipeline_id != format::kNullHandleId)
```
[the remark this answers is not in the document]

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: yes, it's done the same for graphics/compute sync/async now.

fabian-lunarg: maybe just I did graphics in first commit, then added compute in another

PR #2210, opened 2025-05-14
Codegen enumerate API calls

MarkY-LunarG:
Codegen any api calls that perform an enumerate style behavior that can also return an incomplete result.  This is because during replay we may have more items than the capture found and our current setup requires a manual implementation to check that what was in the capture appears in the replay.

I do not do this if:
 - The returned list is a struct
 - The returned list is a handle, atom or opaque item.

While I called "generate" for Vulkan, it did not generate any new code.  This only generated code specifically for OpenXR>


MarkY-LunarG:
@bradgrantham-lunarg you're just on here for visibility.

framework/generated/khronos_generators/khronos_replay_consumer_body_generator.py
```diff
@@ -65,6 +65,28 @@ def generate_custom_extended_struct_handling(self, struct, struct_type):
 
         return None
 
+    def is_enum_with_insfficient(self, api_data, cmd_name, values):
```
bradgrantham-lunarg, line 68: insufficient

framework/generated/khronos_generators/khronos_base_generator.py
```diff
@@ -1635,6 +1641,15 @@ def genCmd(self, cmdinfo, name, alias):
 
         # Create the declaration for the function prototype
         proto = cmdinfo.elem.find('proto')
+
+        # If there are error codes, add them to the list for this command
+        errorcodes = cmdinfo.elem.get('errorcodes')
+        if errorcodes is not None:
+            errors = []
```
bradgrantham-lunarg, line 1648: `self.all_cmd_errors[name] = errorcodes.split(',')` can be used in place of these four lines of code if you'd like.

jzulauf-lunarg (changes requested):
I think the enumerant discrepency code is more complex than needed, and should be templated not repeatedly generated.

framework/generated/generated_openxr_replay_consumer.cpp
```diff
@@ -146,6 +146,70 @@ void OpenXrReplayConsumer::Process_xrGetSystemProperties(
...
+    uint32_t* out_environmentBlendModeCountOutput = environmentBlendModeCountOutput->IsNull() ? nullptr : environmentBlendModeCountOutput->AllocateOutputData(1, static_cast<uint32_t>(0));
+    XrEnvironmentBlendMode* out_environmentBlendModes = environmentBlendModes->IsNull() ? nullptr : environmentBlendModes->AllocateOutputData(environmentBlendModeCapacityInput);
+
+    XrResult replay_result = XR_SUCCESS;
+    if (!environmentBlendModes->IsNull() && returnValue == XR_SUCCESS)
+    {
+        // Get actual count from replay
+        uint32_t replay_count = 0;
+        replay_result = GetInstanceTable(in_instance)->EnumerateEnvironmentBlendModes(in_instance, in_systemId, viewConfigurationType, 0, &replay_count, nullptr);
+        if (replay_result != XR_SUCCESS || replay_count < environmentBlendModeCapacityInput)
```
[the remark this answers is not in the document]
MarkY-LunarG: How do you wrap a template around this since the calls may be formatted differently for each enumerate call?  Some have more parameters than others.  It's not just a matter of these are the same except for the queried types.  It seems like overkill for this at this time when the codegen is sufficient for a near-term fix.

MarkY-LunarG:
Closing this PR for now.  I can re-open when we have funds to continue  it.

PR #2212, opened 2025-05-16
OpenXR: Reentry Rework

jzulauf-lunarg:
A rework of the openxr capture reentry control which solve the multithread issues found in openxr-tutorial, and includes a work-around to enable more applications.

Note that reentry control has three algorithms present, but a code modification is currently required to change them, as openxr support is still experimental and *developers* may still want to play with them (or others).  One of them is backwards-compatible, the other two survive multithread basic testing. Likely only the currently active one will be retained.


framework/generated/util_generators/generate_api_call_id_strings.py
```diff
@@ -0,0 +1,71 @@
...
+    */
+
+    ''')
+
+    for line in in_file:
+        if re.search(r'^\s+ApiCall\S+\s+=', line):
+            pair = re.sub(r'^\s+ApiCall_(\S+)\s+=.*$', r'    {ApiCall_\1, "\1"},', line)
+            out_file.write(pair)
+
+    out_file.write('    // clang-format on\n')
```
MarkY-LunarG, line 71: I don't see a clang-format off command

framework/encode/openxr_capture_manager.cpp
```diff
@@ -194,6 +194,82 @@ void OpenXrCaptureManager::EndFramePreDispatch(XrSession session, const XrFrameE
     WriteViewRelativeLocationMetadata(session, *frameEndInfo);
 }
 
+OpenXrCaptureManager::OpenXrCaptureManager() : ApiCaptureManager(format::ApiFamilyId::ApiFamily_OpenXR)
+{
+    reentry_state_.reentry_mode = ReentryMode::kThreadDisable; // Note make this configurable in future?
```
MarkY-LunarG, line 199: I would make a TODO to add this as an option.  I want it configurable just to verify if a bug goes away.  Otherwise, we should only have this one option.

framework/encode/custom_openxr_api_call_encoders.cpp
```diff
@@ -338,17 +361,32 @@ XRAPI_ATTR XrResult XRAPI_CALL xrCreateVulkanDeviceKHR(XrInstance
...
+            auto       instance_wrapper = openxr_wrappers::GetWrapper<openxr_wrappers::InstanceWrapper>(instance);
+            VkInstance vulkanInstance   = instance_wrapper->vkgipa_vkinstance_map[createInfo->pfnGetInstanceProcAddr];
+
+            vulkan_wrappers::CreateWrappedHandle<vulkan_wrappers::InstanceWrapper,
+                                                 vulkan_wrappers::NoParentWrapper,
+                                                 vulkan_wrappers::DeviceWrapper>(
+                vulkanInstance,
+                vulkan_wrappers::NoParentWrapper::kHandleValue,
+                vulkanDevice,
+                VulkanCaptureManager::GetUniqueId);
```
MarkY-LunarG, line 384: Add a GFXR_LOG_ERROR here .  I think it should show up as a "hey dummy this is going on".

jzulauf-lunarg: Don't want to "engine shame" until we talk with the developers of the library that is doing this.  With the work around, we're functional.

jzulauf-lunarg:
@bradgrantham-lunarg  -- addressed most of the comments above, but moved to DRAFT as Android testing was inconclusive.

PR #2213, opened 2025-05-20
Fix Vulkan WSI instance extensions

marius-pelegrin-arm:
This commit fixes 3 bugs:
- `VK_EXT_HEADLESS_SURFACE_EXTENSION_NAME` missing in `kSurfaceExtensions`
- `cli_wsi_extension_` not being set (nor possible to set) when replaying on Android
- If offscreen content is captured and replayed specifying a WSI, it was possible to end up with instance extensions containing a surface extension but no `VK_KHR_surface`

### DIFF
--- a/framework/application/application.h
+++ b/framework/application/application.h
@@ -46,7 +46,10 @@ class Application final
   public:
     Application(const std::string& name, decode::FileProcessor* file_processor);
 
-    Application(const std::string& name, const std::string& cli_wsi_extension, decode::FileProcessor* file_processor);
+    Application(const std::string&     name,
+                decode::FileProcessor* file_processor,
+                const std::string&     cli_wsi_extension,
+                void*                  platform_specific_wsi_data);
 
     ~Application();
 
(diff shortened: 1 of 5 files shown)


framework/application/application.h
```diff
@@ -46,7 +46,10 @@ class Application final
   public:
     Application(const std::string& name, decode::FileProcessor* file_processor);
 
-    Application(const std::string& name, const std::string& cli_wsi_extension, decode::FileProcessor* file_processor);
+    Application(const std::string&     name,
+                decode::FileProcessor* file_processor,
+                const std::string&     cli_wsi_extension,
+                void*                  platform_specific_wsi_data);
```
MarkY-LunarG, line 52: you could just initialize this here to = nullptr and remove the changes to application.cpp line 61, and the last bit of desktop_main.cpp line 156.

marius-pelegrin-arm:
@bradgrantham-lunarg Typically, capture an application that renders to a headless surface and try to replay it on Android. The first bug will cause `VK_EXT_headless_surface` to not be removed from the queried extensions and the second one will cause `VK_KHR_android_surface` to not be added, even if you tried to specify `--wsi`. The replayer will crash at instance creation.

PR #2217, opened 2025-05-26
CMakeLists.txt: Disable AGS support in Windows on ARM64 builds

### LINKED ISSUE #1262: Windows ARM64 builds broken
d3d12_capture_manager.h unconditionally includes ags_dispatch_table.h, which unconditionally includes amd_ags.h
amd_ags.h isn't in the precompiled folder for win64-arm

fgiancane8:
AMD provides AGS libraries only in x86/64 variants so in case of ARM64 builds AGS should be always disabled.

Fixes #1262.


fgiancane8:
Keeping the PR in draft state as I need to test it extensively. Will mark as ready to merge when testing is done.

fgiancane8:
Tested on an x86_64 Windows box with ARM64 cross toolchain, using VS2022 generators and latest SDKs.

bradgrantham-lunarg (approved):

PR #2220, opened 2025-05-29
test: Remove SDL dependency

antonio-lunarg:
Use the `application` library instead.


bradgrantham-lunarg:
@beau-lunarg can you find a moment to look through this since you know more about the test framework than I do?  If you think it looks good then I will approve it based on your review.

test/test_apps/common/test_app_base.cpp
```diff
@@ -326,12 +324,149 @@ class VulkanFunctions
...
+        GFXRECON_ASSERT(inst != VK_NULL_HANDLE);
+        GFXRECON_ASSERT(ptr_vkGetInstanceProcAddr != nullptr);
+        instance = inst;
+        graphics::LoadVulkanInstanceTable(ptr_vkGetInstanceProcAddr, inst, &inst_table);
+
+        get_inst_proc_addr(fp_vkCreateDevice, "vkCreateDevice");
+        get_inst_proc_addr(fp_vkGetDeviceProcAddr, "vkGetDeviceProcAddr");
+    }
+
+    bool load_vulkan_library()
```
mikes-lunarg, line 184: There is some platform code for loading libraries that we might be able to reuse here: `graphics::InitializeLoader()` / `util::platform::GetProcAddress`

antonio-lunarg:
> This is a great cleanup, my local linux build goes from 912 build steps down to just 274!
> 
> The test apps run fine with `GFXRECON_TESTAPP_HEADLESS` (and with run_tests.sh) but I get an assert if I try to run them individually from the launcher, I think because `Application::InitializeWsiContext` is never called.
> 
> ```shell
> $ env -C install/test/res ../test_apps/launcher/gfxrecon-test-launcher triangle
> [gfxrecon] FATAL - Assertion failed: 'wsi_context' (/home/mike/Projects/gfxreconstruct2/test/test_apps/common/test_app_base.cpp:2877)
> gfxrecon-test-launcher: /home/mike/Projects/gfxreconstruct2/test/test_apps/common/test_app_base.cpp:2877: gfxrecon::decode::Window* gfxrecon::test::create_window(gfxrecon::application::Application&, bool, int, int): Assertion `wsi_context' failed.
> fish: Job 1, 'env -C install/test/res ../test…' terminated by signal SIGABRT (Abort)
> ```

Thanks for testing this!
I added the `--wsi` command line argument, so we can specify a platform at command line now.
There's still a problem with `auto` as it does not select any platform, hence no WSI context is initialiazed.
I tried with automatic selection of the first available platform when `auto` is selected, but that triggers a bug with the replayer. 🤔

test/test_apps/common/test_app_base.cpp
antonio-lunarg: Cool! Updated.

antonio-lunarg:
> > This is a great cleanup, my local linux build goes from 912 build steps down to just 274!
> > The test apps run fine with `GFXRECON_TESTAPP_HEADLESS` (and with run_tests.sh) but I get an assert if I try to run them individually from the launcher, I think because `Application::InitializeWsiContext` is never called.
> > ```shell
> > $ env -C install/test/res ../test_apps/launcher/gfxrecon-test-launcher triangle
> > [gfxrecon] FATAL - Assertion failed: 'wsi_context' (/home/mike/Projects/gfxreconstruct2/test/test_apps/common/test_app_base.cpp:2877)
> > gfxrecon-test-launcher: /home/mike/Projects/gfxreconstruct2/test/test_apps/common/test_app_base.cpp:2877: gfxrecon::decode::Window* gfxrecon::test::create_window(gfxrecon::application::Application&, bool, int, int): Assertion `wsi_context' failed.
> > fish: Job 1, 'env -C install/test/res ../test…' terminated by signal SIGABRT (Abort)
> > ```
> 
> Thanks for testing this! I added the `--wsi` command line argument, so we can specify a platform at command line now. There's still a problem with `auto` as it does not select any platform, hence no WSI context is initialiazed. I tried with automatic selection of the first available platform when `auto` is selected, but that triggers a bug with the replayer. 🤔

@mikes-lunarg , so I opted for a different behaviour for the test launcher: `auto` here means "first available platform". On the other hand, the replay app keeps ignoring WSI on `auto`.

beau-lunarg (approved):

mikes-lunarg (approved):

PR #2221, opened 2025-05-29
Ignore files generated in the assets folder

antonio-lunarg:


MarkY-LunarG (approved):

antonio-lunarg:
> This seems like a great addition but if we ever make an "assets" directory anywhere (e.g. test/assets) people will be surprised to find this message:
> 
> ```
> The following paths are ignored by one of your .gitignore files:
> test/assets
> hint: Use -f if you really want to add them.
> hint: Disable this message with "git config advice.addIgnoredFile false"
> ```
> 
> Is there a way we can limit this to `android/`? Like maybe put it in `android/`?

I've refined the path to `android/layer/assets`.

bradgrantham-lunarg (approved):

MarkY-LunarG:
Yikes, good catch @bradgrantham-lunarg, I didn't notice we had another assets folder.

bradgrantham-lunarg:
> Yikes, good catch @bradgrantham-lunarg, I didn't notice we had another assets folder.

We don't, but it seemed like a good candidate for someone to eventually make in the future.

PR #2225, opened 2025-05-30
Fix Composition/CoreWindow SwapChain replay

dgraves:
D3D12 replay currently converts CreateSwapChainForComposition and CreateSwapChainForCoreWindow to CreateSwapChainForHwnd, but the HWND based swapchain does not support some composition features available to the Composition and CoreWindow swapchains, resulting in swapchain creation failure when composition is enabled. The conversions will also always specify the same window ID of 0, leading to swapchain creation failure when multiple swapchains are created because they all attempt to use the same window. This change addresses these issues with the following:
- Clear values related to composition from DXGI_SWAP_CHAIN_DESC1 prior to calling CreateSwapChainForHwnd.
- Specify a unique window ID for CreateSwapChainForCompostion and CreateSwapChainForCoreWindow conversions to CreateSwapChainForHwnd.

For these cases, the original application would have composed images from multiple swapchains into a single window, while replay will display the swapchain images in separate windows.

### DIFF
--- a/framework/decode/dx12_replay_consumer_base.cpp
+++ b/framework/decode/dx12_replay_consumer_base.cpp
@@ -97,7 +97,6 @@ static void ReplayMessageFunc(D3D12_MESSAGE_CATEGORY category,
     }
 }
 
-
 void InitialResourceExtraInfo(HandlePointerDecoder<void*>* resource_decoder,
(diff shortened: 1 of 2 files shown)


bradgrantham-lunarg:
All testing has passed except Android, so going to merge this.

PR #2227, opened 2025-06-02
VDR: Fix compute immutable resources

panos-lunarg:


framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -2185,6 +2185,72 @@ void VulkanReplayDumpResourcesBase::DumpGraphicsPipelineInfos(
...
+    uint32_t                                                         createInfoCount,
+    HandlePointerDecoder<VkPipeline>*                                pPipelines)
+{
+    const auto* create_info_meta = pCreateInfos->GetMetaStructPointer();
+    for (uint32_t i = 0; i < createInfoCount; ++i)
+    {
+        VulkanPipelineInfo* pipeline_info = reinterpret_cast<VulkanPipelineInfo*>(pPipelines->GetConsumerData(i));
+
+        // Copy pipeline layout information
+        if (create_info_meta != nullptr)
```
bradgrantham-lunarg: I think this if-statement should be outside the loop.  (Unless I'm misunderstanding and there's a hidden side effect in the loop which alters that pointer.)

framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -2185,6 +2185,72 @@ void VulkanReplayDumpResourcesBase::DumpGraphicsPipelineInfos(
     }
 }
 
+void VulkanReplayDumpResourcesBase::DumpComputePipelineInfos(
+    const StructPointerDecoder<Decoded_VkComputePipelineCreateInfo>* pCreateInfos,
+    uint32_t                                                         createInfoCount,
+    HandlePointerDecoder<VkPipeline>*                                pPipelines)
+{
+    const auto* create_info_meta = pCreateInfos->GetMetaStructPointer();
```
bradgrantham-lunarg: The body of this function is the same as the other functions below - their bodies should be extracted and made into a single templated function to avoid cut-and-paste errors later.

bradgrantham-lunarg (dismissed):

bradgrantham-lunarg (changes requested):
factor out identical function bodies and elevate if-statement out of loop

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -10710,6 +10729,16 @@ VkResult VulkanReplayConsumerBase::OverrideCreateComputePipelines(
...
+
+    // Information is stored in the created PipelineInfos only when the dumping resources feature is in use
+    if (replay_result == VK_SUCCESS)
+    {
+        if (options_.dumping_resources)
+        {
+            resource_dumper_->DumpComputePipelineInfos(pCreateInfos, create_info_count, pPipelines);
+        }
+    }
+
```
fabian-lunarg, line 10741: we do not (yet) have any async flavor for raytracing-pipelines,
but here we will require the same in `AsyncCreateComputePipelines` (analog to `AsyncCreateGraphicsPipelines`).

framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -2185,6 +2185,72 @@ void VulkanReplayDumpResourcesBase::DumpGraphicsPipelineInfos(
...
+        if (create_info_meta != nullptr)
+        {
+            const auto ppl_layout_info = object_info_table_->GetVkPipelineLayoutInfo(create_info_meta[i].layout);
+            if (ppl_layout_info != nullptr)
+            {
+                pipeline_info->desc_set_layouts = ppl_layout_info->desc_set_layouts;
+            }
+        }
+    }
+}
```
fabian-lunarg: function-bodies are identical. easiest way seems to make `Decoded_VkRayTracingPipelineCreateInfoKHR`/`Decoded_VkRayTracingPipelineCreateInfoNV` a template param with a static_assert for the type. alternatively an std::variant could be used as param, but likely more cumbersome

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: meaning, could you also add to `AsyncCreateComputePipeline`:

```
...
graphics::populate_shader_stages(...)// after this, before the async call is assembled

if (returnValue == VK_SUCCESS && options_.dumping_resources)
{
    resource_dumper_->DumpComputeRayTracingPipelineInfos(pCreateInfos, createInfoCount, pPipelines);
}
```

panos-lunarg: right, got it. I misinterpreted this as being a todo for when we add the async for raytracing pipelines

fabian-lunarg (approved):

Issue #2223, opened 2025-05-29
OpenXR loader should be dynamically found and loaded like Vulkan loader


bradgrantham-lunarg:
`decode` is linked against `$<$<BOOL:${OPENXR_SUPPORT_ENABLED}>:OpenXR::openxr_loader` in `framework/decode/CMakeLists.txt`.  On Ubuntu that means that `gfxrecon-replay` run directly from the unpacked SDK `.tar.xz` expects to find `libopenxr_loader.1` but somehow doesn't know to go to the executable's directory to get it.  So as of 313.0, one must add the directory containing `libopenxr_loader.so.1` to `LD_LIBRARY_PATH` (or `PATH` maybe?) in order to run `replay` without first running `setup_env.sh`.

The Vulkan loader is explicitly loaded by Consumers that need to make Vulkan calls, see framework/graphics/vulkan_util.cpp and `framework/decode/vulkan_replay_consumer_base.cpp:    loader_handle_ = graphics::InitializeLoader();`

Need to build similar plumbing for OpenXR's loader, probably split "InitializeLoader" into "InitializeVulkanLoader" and "InitializeOpenXrLoader", probably only call "InitializeOpenXrLoader" in OpenXR consumers.

MarkY-LunarG:
Another option is to build the static OpenXR loader.  This is an option that Vulkan does not allow and would simplify the process. 

To do this, we need to update the two places that add OpenXR-SDK as a subdirectory and precede the calls with:

```
option(DYNAMIC_LOADER "" OFF)
```

These two cases are in GFXR top CMakeLists.txt file (around line 388) and the tools/quest_replay application CMakeLists.txt around line 33.  If we want to go that route.

bradgrantham-lunarg:
Is there static initialization that happens in the OpenXR loader?  Then statically linking will induct all of that into `replay` and it would be nice to not do *any* OpenXR work in `replay` until we know we need an OpenXR instance.

I think going the dynamic route may not be much work at all since we already have it in there for Vulkan.  But will that work on Android?  Or do Android apps expect to be statically linked?

lunarpapillo:
Another option is to use `CMAKE_BUILD_RPATH` and `CMAKE_INSTALL_RPATH` to add a relative path to the `gfxrecon-replay` executable telling the loader where to find the dependent libraries.

I mention this just for completeness - I actually prefer either of the other two alternatives cited; but for completeness all alternatives should be investigated on their own merits.

[...]

PR #2230, opened 2025-06-05
[openxr] Add support for dynamic loader

MarkY-LunarG:
Load the OpenXR loader dynamically not statically.  This prevents having to incorporate it into the standard GFXR replay app since most people will only need OpenXR when using OpenXR applications.  So it's an unnecessary inclusion.

@jzulauf-lunarg, it would be good to know if this continues to work with your Monado.

### DIFF
--- a/framework/decode/openxr_resource_tracking_consumer.cpp
+++ b/framework/decode/openxr_resource_tracking_consumer.cpp
@@ -62,11 +62,12 @@ void OpenXrResourceTrackingConsumer::InitializeLoader()
     {
         XrResult result = get_instance_proc_addr_(
             XR_NULL_HANDLE, "xrCreateInstance", reinterpret_cast<PFN_xrVoidFunction*>(create_instance_function_));
-        if (create_instance_function_ == nullptr)
-        {
-            GFXRECON_LOG_FATAL("Failed to load OpenXR runtime library; please ensure that the path to the OpenXR "
-                               "loader has been added to the appropriate system path");
-        }
+    }
+
+    if (create_instance_function_ == nullptr)
+    {
+        GFXRECON_LOG_FATAL("Failed to load OpenXR runtime library; please ensure that the path to the OpenXR "
+                           "loader has been added to the appropriate system path");
     }
 }
 
--- a/framework/decode/openxr_resource_tracking_consumer.cpp
+++ b/framework/decode/openxr_resource_tracking_consumer.cpp
@@ -23,6 +23,7 @@
 #if ENABLE_OPENXR_SUPPORT
 
 #include "decode/openxr_resource_tracking_consumer.h"
+#include "graphics/khronos_util.h"
 
 #include <algorithm>
 #include <cassert>
@@ -30,16 +31,6 @@
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
-
-const std::vector<std::string> kLoaderLibNames = {
-#if defined(WIN32)
-    "openxr-1.dll"
-#elif defined(__APPLE__)
-    "libopenxr.dylib", "libopenxr.1.dylib", "libMoltenVK.dylib"
-#else
-    "libopenxr.so", "libopenxr.so.1"
-#endif
-};
 
 OpenXrResourceTrackingConsumer::OpenXrResourceTrackingConsumer(
     const OpenXrReplayOptions& options, OpenXrTrackedObjectInfoTable* tracked_object_info_table) :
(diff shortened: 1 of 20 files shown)


MarkY-LunarG:
Possible fix for #2223.

jzulauf-lunarg:
@MarkY-LunarG -- as it doesn't build on windows, I'm guessing, no.

MarkY-LunarG:
Squashed all commits into one.

MarkY-LunarG:
Also verified OpenXR Tutorial 4 still replays on Meta Quest Pro after these changes.

framework/decode/openxr_resource_tracking_consumer.cpp
```diff
@@ -77,9 +64,8 @@ void OpenXrResourceTrackingConsumer::InitializeLoader()
             XR_NULL_HANDLE, "xrCreateInstance", reinterpret_cast<PFN_xrVoidFunction*>(create_instance_function_));
         if (create_instance_function_ == nullptr)
         {
-            GFXRECON_LOG_FATAL("Failed to load OpenXr runtime library; please ensure that the path to the OpenXr "
-                               "loader (eg. %s) has been added to the appropriate system path",
-                               kLoaderLibNames[0].c_str());
+            GFXRECON_LOG_FATAL("Failed to load OpenXR runtime library; please ensure that the path to the OpenXR "
+                               "loader has been added to the appropriate system path");
         }
```
bradgrantham-lunarg: This can fall through with failure but not print anything because the control flow is slightly different from the replay consumer base InitializeLoader.  Add a print in else for `get_instance_proc_addr_ != nullptr`.

framework/graphics/khronos_util.h
```diff
@@ -42,19 +42,16 @@ GFXRECON_BEGIN_NAMESPACE(graphics)
...
-#if defined(WIN32)
-    "vulkan-1.dll"
-#elif defined(__APPLE__)
-    "libvulkan.dylib", "libvulkan.1.dylib", "libMoltenVK.dylib"
-#else
-    "libvulkan.so.1", "libvulkan.so"
-#endif
+enum KhronosLoaderType : uint16_t
+{
+    KhronosLoader_None,
```
bradgrantham-lunarg: Is anything ever going to use `KhronosLoader_None` or is there a reason to special case `0`?  If not, then I think you can drop this enumerant.

framework/graphics/khronos_util.cpp
```diff
@@ -20,19 +20,50 @@
...
+util::platform::LibraryHandle InitializeKhronosLoader(KhronosLoaderType type)
 {
-    return util::platform::OpenLibrary(kLoaderLibNames);
+    switch (type)
+    {
+        case KhronosLoader_Vulkan:
+            return util::platform::OpenLibrary(kVulkanLoaderLibNames);
+        case KhronosLoader_OpenXR:
+            return util::platform::OpenLibrary(kOpenXrLoaderLibNames);
+        default:
```
bradgrantham-lunarg, line 57: Calling without a valid enumerant implies corruption to me, or calling with the wrong enumerant type. 
 maybe assert or abort for default?

bradgrantham-lunarg (changes requested):

jzulauf-lunarg (changes requested):
gfxrecon-replay fails in `OpenXrReplayConsumerBase::Process_xrCreateApiLayerInstance` with `get_instance_proc_addr_ == nullptr`.  This happens because `util::platform::LibraryHandle InitializeKhronosLoader(KhronosLoader_OpenXR)`  tries to load a loader from a list of names in vector `kOpenXrLoaderLibNames` which only contains the name  `"openxr_loader-1.dll"`  and that file does not exist anywhere in the build tree.

MarkY-LunarG:
> gfxrecon-replay fails in `OpenXrReplayConsumerBase::Process_xrCreateApiLayerInstance` with `get_instance_proc_addr_ == nullptr`. This happens because `util::platform::LibraryHandle InitializeKhronosLoader(KhronosLoader_OpenXR)` tries to load a loader from a list of names in vector `kOpenXrLoaderLibNames` which only contains the name `"openxr_loader-1.dll"` and that file does not exist anywhere in the build tree.

@jzulauf-lunarg, what is the name of the loader DLL on your platform?

jzulauf-lunarg:
@MarkY-LunarG there is not a system level dll.  It needs to be built local to the application.  There are no instructions in OpenXR-SDK to do dynamic building.  There are instructions in OpenXR-SDK-Source [BUILDING.md](https://github.com/KhronosGroup/OpenXR-SDK-Source/blob/main/BUILDING.md).  As the release openxrloader.lib is only 16K, not clear of the advantage of dynamic linking.

bradgrantham-lunarg:
> @MarkY-LunarG there is not a system level dll. It needs to be built local to the application. There are no instructions in OpenXR-SDK to do dynamic building. There are instructions in OpenXR-SDK-Source [BUILDING.md](https://github.com/KhronosGroup/OpenXR-SDK-Source/blob/main/BUILDING.md). As the release openxrloader.lib is only 16K, not clear of the advantage of dynamic linking.

Please see linked issue.

MarkY-LunarG:
@bradgrantham-lunarg , to keep it all in this issue, I pointed out to Zulauf that the GFXR build does build a .DLL for the loader (under `dbuild\windows\x64\output` for my build called `openxr_loaderd.dll`).  How that can get used and if it works with Monado remains unknown right now.

framework/decode/openxr_resource_tracking_consumer.cpp
MarkY-LunarG: This matches the general behavior of what `VulkanResourceTrackingConsumer` does.  I moved the LOG_FATAL out to a `if  (create_instance_function_ == nullptr)`  check just like the Vulkan side.  Should I do more?

framework/graphics/khronos_util.cpp
MarkY-LunarG: Added Fatal log and assert.

framework/decode/openxr_resource_tracking_consumer.cpp
bradgrantham-lunarg: LGTM

PR #2231, opened 2025-06-06
Filter primaries that reference invalid secondaries

panos-lunarg:


framework/encode/vulkan_handle_wrappers.h
```diff
@@ -435,6 +435,20 @@ struct CommandBufferWrapper : public HandleWrapper<VkCommandBuffer>
...
+        kRecording,
+        kExecutable,
+        kInvalid
+    };
+
+    CommandBufferState state{ kInitial };
+
+    // This is set to true when a command buffer is began with VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT
+    // Used to track command buffer life cycle
+    bool one_time_submision{ false };
```
bradgrantham-lunarg: submission

bradgrantham-lunarg:
I think this looks good but I don't feel confident about my assessment - @fabian-lunarg would you please give it a once-over?

framework/encode/vulkan_handle_wrappers.h
```diff
@@ -273,7 +273,7 @@ struct DeviceMemoryWrapper : public HandleWrapper<VkDeviceMemory>
     // This is the device which was used to allocate the memory.
     // Spec states if the memory can be mapped, the mapping device must be this device.
     // The device wrapper will be initialized when allocating the memory. Some handling
-    // like StateTracker::TrackTlasToBlasDependencies may use it before mapping
+    // like StateTracker::TrackCommandBuffersSubmision may use it before mapping
```
fabian-lunarg, line 276: much better name :)

framework/encode/vulkan_state_writer.cpp
```diff
@@ -4339,7 +4353,10 @@ bool VulkanStateWriter::CheckCommandHandle(vulkan_state_info::CommandHandleType
         case vulkan_state_info::CommandHandleType::BufferViewHandle:
             return IsBufferViewValid(handle_id, state_table);
         case vulkan_state_info::CommandHandleType::CommandBufferHandle:
-            return (state_table.GetVulkanCommandBufferWrapper(handle_id) != nullptr);
+        {
+            const vulkan_wrappers::CommandBufferWrapper* wrapper = state_table.GetVulkanCommandBufferWrapper(handle_id);
+            return (wrapper != nullptr && wrapper->state == vulkan_wrappers::CommandBufferWrapper::kExecutable);
```
fabian-lunarg, line 4358: doesn't this change the (expected) behavior ?
before `CheckCommandHandle` would just tell if the handle is contained in the state-table.

now it also checks if it was recorded before (and was not submitted if it's a one-time-submit).
this 'could' change behavior somewhere else too?
what about command-buffers that have not yet been recorded (`kInitial`)?
those would become `kInvalid` too?

fabian-lunarg:
the changes look good. guess just some minor questions, for clarity.

framework/encode/vulkan_state_writer.cpp
panos-lunarg: The purpose here is to not dump commands of command buffers which are referencing invalid secondaries. Previously this was limited to checking whether the secondary is actually a valid object (has not been deleted). The new code also checks if this is correct according to the spec.

> what about command-buffers that have not yet been recorded (kInitial)?
those would become kInvalid too?

According to spec this would be invalid usage as the secondary must be in executable or pending state.

By the time the state writer triggers the primary - secondary combination can be invalid.
For instance:
- Secondary allocated
- Secondary recorded
- Primary allocated
- Primary recorded + references secondary
- ...
- Secondary is reset (is set to the initial state)
- Capture is triggered -> state writer executes
  
If we dump the primary's commands then during replay we are basically violating the spec

PR #2233, opened 2025-06-08
VDR: Fixes for secondaries for compute/rt

panos-lunarg:


wpaul-samsung:
This works for us.  Please initiate a review.

framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -105,24 +105,15 @@ VulkanReplayDumpResourcesBase::VulkanReplayDumpResourcesBase(const VulkanReplayO
                 std::piecewise_construct,
                 std::forward_as_tuple(bcb_index),
                 std::forward_as_tuple((options.Dispatch_Indices.size() && options.Dispatch_Indices[i].size())
-                                          ? options.Dispatch_Indices[i]
-                                          : DispatchIndices(),
+                                          ? &options.Dispatch_Indices[i]
```
bradgrantham-lunarg, line 108: Passing pointers through here sets off some alarms in my head (here and in other places in these files that I didn't catch when originally submitted as PR) - is this going into a constructor that dereferences these pointers or are the pointers being stored?

panos-lunarg: Yes, they are passed into the constructor. If they are not null they are dereferenced and the content they point to (vectors) is copied by value.

bradgrantham-lunarg (approved):

PR #2234, opened 2025-06-09
Allowlist for different return code from capture

panos-lunarg:
Creating a list of Vulkan api calls which should be allowed to return a different error code during replay


bradgrantham-lunarg (approved):

wpaul-samsung:
This works for us.  Thanks!

PR #2238, opened 2025-06-09
VDR: Option to dump unreferenced vertex bindings

panos-lunarg:
The new replay option --dump-resources-dump-unused-vertex-bindings can be used to dump vertex bindings which are not referenced by a vertex attribute


fabian-lunarg:
just curious. we already have plenty of CLI options for dump-resources. we also have defined a json-file to contain settings for dump-resources. would it be possible (or a good idea!?) to add new parameters to this json-schema instead of adding new CLI options?

panos-lunarg:
Actually this sounds like a good idea.
The only "obstacle" I see is that we also support passing dump resources indices as cli options only, which I think has limitations compared to the json input, so the rest of parameters need to be passed as other cli options as well.
I think adding the VDR options in the input json and then deprecate the CLI options will decluttering the already numerous options.
I'll create a ticket to track this and see what other people think about it

fabian-lunarg (approved):

fabian-lunarg:
> I'll create a ticket to track this and see what other people think about it

sounds good!

bradgrantham-lunarg:
Issue #2239

bradgrantham-lunarg (approved):

PR #2242, opened 2025-06-11
[openxr] Ignore refer space change events

MarkY-LunarG:
Reference space changed events are not critical, and sometimes do not re-occur during the replay.  So add them to the list of events we're allowed to skip.

Also fix a bug caught by Brad where we were spitting out the incorrect event information when storing it to the pending event array.


MarkY-LunarG:
Passed with Jenkin's re-run # 7018.

PR #2250, opened 2025-06-16
Split phys dev surface formats

panos-lunarg:
Physical device surface formats
(vkGetPhysicalDeviceSurfaceFormats(2)KHR) and capabilities
(vkGetPhysicalDeviceSurfaceCapabilities(2)KHR) are tracked by the state
tracker. State writer groups all queried results and uses the 2KHR
version of those commands to store them in the state setup. This assumes
that the replay device implements those functions which some times is
not the case.

This patch makes the state tracker to distinguish which functions are
called and the state writer will use these instead of hard coding the
2KHR versions.

Also handle vkGetPhysicalDeviceSurfaceCapabilities2EXT in state tracker
in similar fashion as the other 2 functions


framework/encode/vulkan_handle_wrappers.h
```diff
@@ -506,10 +506,20 @@ struct CommandPoolWrapper : public HandleWrapper<VkCommandPool>
...
+        k2KHR,
+        k2EXT
+    } function_version;
+
+    VkSurfaceKHR                    surface;
+    VkPhysicalDeviceSurfaceInfo2KHR surface_info{};
     std::unique_ptr<uint8_t[]>      surface_info_pnext_memory;
 
-    VkSurfaceCapabilities2KHR  surface_capabilities;
+    VkSurfaceCapabilitiesKHR   surface_capabilities;
```
bradgrantham-lunarg, line 520: If these are always keyed from `FunctionVersion`, should they be a union or maybe even a C++17 variant?

panos-lunarg:
Abandoning this in favor of https://github.com/LunarG/gfxreconstruct/pull/2265

PR #2251, opened 2025-06-16
Fix D3D SECURITY_ATTRIBUTES struct decode

dgraves:
D3D struct encoding was using the custom SECURITY_ATTRIBUTES encoder defined for Vulkan, but decoding was using an incompatible generated decoder, resulting in an access violation during the decoding process. This change updates the code generator to use the custom SECURITY_ATTRIBUTES decoder, with the same logic used to process the LARGE_INTEGER struct, which also has a custom decoder, applied to the SECURITY_ATTRIBUTES struct.

A fix for a potential null dereference when decoding the SECURITY_DESCRIPTOR member of SECURITY_ATTRIBUTES, which impacts both D3D and Vulkan, is also included.


framework/decode/custom_vulkan_struct_decoders.cpp
```diff
@@ -296,22 +296,27 @@ size_t DecodeStruct(const uint8_t* buffer, size_t buffer_size, Decoded_VkPushDes
...
-    // so we can subtract 4 bytes from sid_authority_size.
-    size_t   allocation_size = sizeof(SID) + (sid_authority_size - sizeof(uint32_t));
-    uint8_t* unpacked_memory = DecodeAllocator::Allocate<uint8_t>(allocation_size);
-
-    SID* sid               = reinterpret_cast<SID*>(unpacked_memory);
-    sid->Revision          = bytes[0];
-    sid->SubAuthorityCount = bytes[1];
-    memcpy(sid->IdentifierAuthority.Value, &bytes[2], 6);
-    memcpy(sid->SubAuthority, &bytes[8], sid_authority_size);
+    uint8_t* unpacked_memory = nullptr;
```
bradgrantham-lunarg: If this isn't Vulkan-specific, I think it maybe helpful for it to be factored out to an API-independent header now that D3D is going to use it.  Having `custom_dx12_struct_decoders.h` include `custom_vulkan_struct_decoders.h` may be confusing to some readers.  (E.g. me )

bradgrantham-lunarg (changes requested):
Just the one thought about factoring out non-API-specific bits

bradgrantham-lunarg:
Looks pretty good and I think I have just the concern about factoring out the functionality.  Thank you @dgraves !

PR #2252, opened 2025-06-17
Revert "CMakeLists.txt: Disable AGS support in Windows on ARM64 builds"

bradgrantham-lunarg:
Reverts LunarG/gfxreconstruct#2217


bradgrantham-lunarg:
Actually I think the fix is pretty simple, closing.

PR #2253, opened 2025-06-17
handle case of no generator specified

bradgrantham-lunarg:


fabian-lunarg (approved):

fgiancane8:
Hey @bradgrantham-lunarg ,
I saw my previous PR (#2217) broke some of the default values. Glad you merged this fix. I think there is still one case that would be unhandled. (arm64 build on arm64).

Will open a follow-up PR soon.

PR #2254, opened 2025-06-17
Report frame number in debug callback during replay

### LINKED ISSUE #1859: Add frame number to messages received through debug utils
This would allow a user of CDL to localize a crash to a short sequence of frames, make a trimmed recapture of those frames, and use that recapture to more quickly investigate bugs and make repro case.  @jeremyg-lunarg 

GFXR also may not be setting up debug utils messengers correctly, needs investigation as part of this task.

nickdriscoll-lunarg:
Fixes #1859 

This PR adds logic to propagate the frame number down from the FileProcessor to the consumers, so that the frame number can appear in messages emitted by the debug callback.

### DIFF
--- a/framework/decode/api_decoder.h
+++ b/framework/decode/api_decoder.h
@@ -199,6 +199,7 @@ class ApiDecoder
 
     virtual void SetCurrentBlockIndex(uint64_t block_index){};
 
+    // Expects zero-based frame_number to match the way FileProcessor::current_frame_number_ works
     virtual void SetCurrentFrameNumber(uint64_t frame_number){};
 
     virtual void SetCurrentApiCallId(format::ApiCallId api_call_id){};
--- a/framework/decode/api_decoder.h
+++ b/framework/decode/api_decoder.h
@@ -199,7 +199,7 @@ class ApiDecoder
 
     virtual void SetCurrentBlockIndex(uint64_t block_index){};
 
-    virtual void SetCurrentFrameNumber(uint64_t frame_number) {};
+    virtual void SetCurrentFrameNumber(uint64_t frame_number){};
 
     virtual void SetCurrentApiCallId(format::ApiCallId api_call_id){};
 
--- a/framework/decode/api_decoder.h
+++ b/framework/decode/api_decoder.h
@@ -199,6 +199,8 @@ class ApiDecoder
 
     virtual void SetCurrentBlockIndex(uint64_t block_index){};
 
+    virtual void SetCurrentFrameNumber(uint64_t frame_number) {};
+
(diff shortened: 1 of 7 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -119,7 +119,8 @@ static VKAPI_ATTR VkBool32 VKAPI_CALL DebugUtilsCallback(VkDebugUtilsMessageSeve
                                                          const VkDebugUtilsMessengerCallbackDataEXT* pCallbackData,
                                                          void*                                       pUserData)
 {
-    GFXRECON_UNREFERENCED_PARAMETER(pUserData);
+    VulkanReplayConsumerBase* replay_consumer = static_cast<VulkanReplayConsumerBase*>(pUserData);
+    uint64_t frame_number = replay_consumer->GetFrameNumber() + 1; // Add 1 because of one-based indexing of frames
```
antonio-lunarg: I would move this `+ 1; // Add 1 because of one-based indexing of frames` to the `GetFrameNumber()` implementation.

framework/decode/vulkan_replay_consumer_base.h
```diff
@@ -307,6 +309,8 @@ class VulkanReplayConsumerBase : public VulkanConsumer
 
     void InitializeReplayDumpResources();
 
+    uint64_t GetFrameNumber();
```
antonio-lunarg: this can be `const`

framework/decode/vulkan_replay_consumer_base.cpp
nickdriscoll-lunarg: Of course

framework/decode/vulkan_replay_consumer_base.h
nickdriscoll-lunarg: Oh ok. TIL what `const` means in the context of method declaration.

framework/decode/api_decoder.h
```diff
@@ -199,6 +199,8 @@ class ApiDecoder
 
     virtual void SetCurrentBlockIndex(uint64_t block_index){};
 
+    virtual void SetCurrentFrameNumber(uint64_t frame_number){};
```
antonio-lunarg, line 203: So the getter is returning a 1-based frame number. But the setter is expecting 0-based, right? I would add documentation for that to avoid confusion.

nickdriscoll-lunarg: Good point. Documented it.

antonio-lunarg (approved):

bradgrantham-lunarg (approved):

PR #2257, opened 2025-06-19
Add ApiCallId for VK_ARM_tensors

naomipappe:
Adds `api_call_id`'s for https://registry.khronos.org/vulkan/specs/latest/man/html/VK_ARM_tensors.html extension


fabian-lunarg:
this is straight-forward, adding API-call Ids for two extensions.
rebased because branch was affected by an intermediate + unrelated issue.

I'm realizing `VK_ARM_tensors` "interacts" with `VK_EXT_descriptor_buffer`, but GFXR doesn't support it yet.

fabian-lunarg (approved):

PR #2259, opened 2025-06-20
[openxr] Force static linkage on all platforms to openxr loader

MarkY-LunarG:
This is an alternative to PR #2230.


MarkY-LunarG:
Forgot to rebase on required change in dev.

bradgrantham-lunarg (approved):

PR #2260, opened 2025-06-23
VDR: Merge comp and RT outputs into descriptors

panos-lunarg:
When dumping descriptors for compute and ray tracing there was an
attempt to distinguish between "outputs" (storage descriptors) and the
rest of the descriptors. This commit merges outputs into descriptors.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -8837,7 +8837,7 @@ VkResult VulkanReplayConsumerBase::OverrideCreateRayTracingPipelinesKHR(
     {
         if (options_.dumping_resources)
         {
-            resource_dumper_->DumpComputeRayTracingPipelineInfos(pCreateInfos, createInfoCount, pPipelines);
+            resource_dumper_->DumpRayTracingPipelineInfos(pCreateInfos, createInfoCount, pPipelines);
         }
     }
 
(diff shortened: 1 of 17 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5543,6 +5543,34 @@ VulkanReplayConsumerBase::OverrideCreateBuffer(PFN_vkCreateBuffer
...
+        {
+            auto in_p_create_meta       = pCreateInfo->GetMetaStructPointer();
+            auto buffer_view_info       = reinterpret_cast<VulkanBufferViewInfo*>(pBufferView->GetConsumerData(0));
+            buffer_view_info->buffer_id = in_p_create_meta->buffer;
+            buffer_view_info->offset    = in_p_create_info->offset;
+            buffer_view_info->range     = in_p_create_info->range;
+        }
+    }
+
+    return result;
```
fabian-lunarg, line 5568: this will behave differently compared to capture when initializing like this: `VkResult result = VK_SUCCESS;`

why is the call to `func` under the condition `if(original_result == VK_SUCCESS)` ? (also this pattern seems different from  other OverrideCreate** calls)

panos-lunarg: Right. Changed it to follow the logic from the rest of the OverrideCreate calls

bradgrantham-lunarg (approved):

PR #2261, opened 2025-06-23
Various validation fixes

fabian-lunarg:
this PR contains accumulated minor fixes for various warning and validation-errors: 

capture-layer:

related to raytracing/acceleration-structures
`[gfxrecon] WARNING - vulkan_wrappers::GetWrappedId() couldn't find Handle: 0x2813a1bfc50's wrapper. It might have been destroyed`

replay:
```
[gfxrecon] ERROR - DEBUG MESSENGER: UNASSIGNED-GeneralParameterError-ExtensionNotEnabled: vkGetPhysicalDeviceSurfaceCapabilities2KHR(): function required extension VK_KHR_get_surface_capabilities2 which has not been enabled.

[gfxrecon] ERROR - DEBUG MESSENGER: UNASSIGNED-GeneralParameterError-ExtensionNotEnabled: vkGetPhysicalDeviceSurfaceFormats2KHR(): function required extension VK_KHR_get_surface_capabilities2 which has not been enabled.

[gfxrecon] ERROR - DEBUG MESSENGER: VUID-vkCmdCopyBuffer-dstBuffer-00120: vkCmdCopyBuffer(): dstBuffer (VkBuffer 0x1e100000001e1) was created with VK_BUFFER_USAGE_2_TRANSFER_SRC_BIT|VK_BUFFER_USAGE_2_STORAGE_BUFFER_BIT|VK_BUFFER_USAGE_2_SHADER_DEVICE_ADDRESS_BIT but requires VK_BUFFER_USAGE_TRANSFER_DST_BIT.

[gfxrecon] ERROR - DEBUG MESSENGER: VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03667: vkCmdBuildAccelerationStructuresKHR(): pInfos[0].mode is VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR, srcAccelerationStructure must have been previously built.

[gfxrecon] ERROR - DEBUG MESSENGER: VUID-vkCmdDraw-None-09600: vkQueueSubmit2(): pSubmits[0].pCommandBufferInfos[0].commandBuffer command buffer VkCommandBuffer 0x65455dfa07a0 expects VkImage 0x18e000000018e (subresource: aspectMask = VK_IMAGE_ASPECT_COLOR_BIT, mipLevel = 0, arrayLayer = 0) to be in layout VK_IMAGE_LAYOUT_PRESENT_SRC_KHR--instead, current layout is VK_IMAGE_LAYOUT_UNDEFINED.
```
- raytracing: changes to trim state-loading for default case (not `-m rebind`)
- additionally some changes to buffer/address tracking and replacement aim to improve performance


framework/decode/vulkan_device_address_tracker.h
```diff
@@ -122,12 +122,20 @@ class VulkanDeviceAddressTracker
      */
     [[nodiscard]] std::unordered_map<VkDeviceAddress, VkDeviceAddress> GetAccelerationStructureDeviceAddressMap() const;
 
+    //! aggregate to group an address and size
+    struct device_address_item_t
```
antonio-lunarg: I would call this some like "address range"

fabian-lunarg: sounds good. then also rename 'size' to 'range' for consistency?

antonio-lunarg: I would say that a range is either a (start, finish) pair, or a an (address, size) pair. So, keeping `size` is the right option to me.

fabian-lunarg: done

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2727,6 +2727,13 @@ void VulkanReplayConsumerBase::ModifyCreateInstanceInfo(
                 available_extensions, &modified_extensions, VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME);
         }
 
+        // NOTE: capturing unconditionally uses 'vkGetPhysicalDeviceSurfaceCapabilities2KHR' for restoring surface state
```
bradgrantham-lunarg: I thought @mikes-lunarg and/or @panos-lunarg were working on that, is that resolved yet?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -8856,7 +8869,7 @@ VulkanReplayConsumerBase::OverrideDeferredOperationJoinKHR(PFN_vkDeferredOperati
     if (!deferred_operation_info->pending_state)
     {
         // The deferred operation object has no deferred command or its deferred command has been finished.
-        return VK_SUCCESS;
+        return original_result;
```
bradgrantham-lunarg, line 8865: Was this tested with a capture file?

bradgrantham-lunarg (changes requested):
Only question I had is whether the `original_result` in deferred_operations was tested.

framework/decode/vulkan_replay_consumer_base.cpp
mikes-lunarg: #2265

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: yes, this was encountered capturing/replaying a current AAA title using vkDeferredOperation. it is my understanding though this is not specific to this capture/title. when using a deferred operation, all but one worker-thread rather whould return `VK_THREAD_DONE` while only one will return VK_SUCCESS. so this caused quite the log-spam and did not reflect what was actually captured

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: ok, perfect. so I should be able to rollback this specific change? @mikes-lunarg

bradgrantham-lunarg (approved):

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: removed the section here, we won't need it

PR #2262, opened 2025-06-23
CMake: Make platform detection on Windows more robust

fgiancane8:
After #2217, it is required to pass `-T` and `-A` for CMake to be able to correctly compute target platforms and thus select proper tools/libs to use at compile time. If these settings are not honoured, failures may happen at CMake configure stage and wrong DXC package is fetched (or no package at all).

Try to mitigate failures by implementing more robust platform detection strategies.


fgiancane8:
My previous PR broke some defaults for GFXR users (thanks @dgraves for reporting) and thus we reworked the code a little bit so to better handle all cases on Windows/Windows on ARM64.

Test results on CMake 4.0.1:

- [X] Host: `x64` Target: `empty`
- [X] Host: `x64` Target: `x64`
- [X] Host `x64` Target `ARM64`
- [X] Host: `ARM64` Target: `empty`
- [X] Host: `ARM64` Target: `x64`
- [X] Host `ARM64` Target `ARM64`

Build succeeds in all of these scenarios.

fgiancane8:
@bradgrantham-lunarg can someone review this PR when they have time? I can't add you to reviewers...

fgiancane8:
gentle reminder.

fgiancane8:
looks like tip is broken 🥲 let me check if I can run my tests again...

bradgrantham-lunarg:
> looks like tip is broken 🥲 let me check if I can run my tests again...

Very strange - both `dev` and your branch compile for me on macOS, so I think we need to look into this some more early in the week.

fgiancane8:
> > looks like tip is broken 🥲 let me check if I can run my tests again...
> 
> Very strange - both `dev` and your branch compile for me on macOS, so I think we need to look into this some more early in the week.

checked the CI errors -- looks like it is a C++ template instantiation error on test, which I assume are not touched by my changes. I'll give it a look more thoroughly on Monday too.
Keep you posted!

fgiancane8:
thanks! @bradgrantham-lunarg

PR #2264, opened 2025-06-25
VDR: Handle vkCreateRenderPass2

panos-lunarg:
- Handle vkCreateRenderPass2
- Handle pNext structs in both vkCreateRenderPass functions

### DIFF
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -647,33 +647,16 @@ struct VulkanRenderPassInfo : public VulkanObjectInfo<VkRenderPass>
     std::vector<VkImageLayout>           attachment_description_final_layouts;
     std::vector<VkAttachmentDescription> attachment_descs;
 
-    struct SubpassReferences
+    std::vector<uint8_t> create_info;
+    enum FuncVersion
     {
-        VkSubpassDescriptionFlags          flags;
-        VkPipelineBindPoint                pipeline_bind_point;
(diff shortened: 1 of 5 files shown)


framework/decode/vulkan_object_info.h
```diff
@@ -647,33 +647,10 @@ struct VulkanRenderPassInfo : public VulkanObjectInfo<VkRenderPass>
...
-        std::vector<VkAttachmentReference> resolve_att_refs;
-        std::vector<uint32_t>              preserve_att_refs;
-
-        bool                  has_depth;
-        VkAttachmentReference depth_att_ref;
-    };
-
-    // The attachment references per subpass
-    std::vector<SubpassReferences> subpass_refs;
+    std::vector<uint8_t> create_info;
```
bradgrantham-lunarg, line 650: I guess there's not much cost to having two vectors here if one is always empty, but could this just be one vector?  Is whether one is empty ever used in conditionals?

panos-lunarg: > Is whether one is empty ever used in conditionals?

Yes. The one that is not empty decides which function to use, `vkCreateRenderPass` or `vkCreateRenderPass2`. One of the vectors can be replaced with a boolean though.

panos-lunarg: Actually, after some experimentation, I discovered that it matters whether we call `vkCreateRenderPass2` vs `vkCreateRenderPass2KHR`. So I added an enum to keep track of the appropriate version of the function

PR #2267, opened 2025-06-26
clear replacers before allocator and device destroy

locke-lunarg:
Replacers have many Vulkan objects that need to be destroyed.  If it's cleared in `~VulkanReplayConsumerBase`,  it will happen errors because allocator and device have been destroyed.
`SaschaWillems/raytracingbasic` with `-m rebind` could trigger these errors.
https://github.com/SaschaWillems/Vulkan/tree/master/examples/raytracingbasic

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -261,9 +261,6 @@ VulkanReplayConsumerBase::~VulkanReplayConsumerBase()
     // Idle all devices before destroying other resources.
     WaitDevicesIdle();
 
-    // free replacer internal vulkan-resources
-    _device_address_replacers.clear();
-
     // process queued async tasks
     background_queue_.join_all();
     main_thread_queue_.poll();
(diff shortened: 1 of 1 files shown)


antonio-lunarg (approved):

crocdialer (commented):
nice!

fabian-lunarg:
I had a look at the outcome of the PR seems to have shifted the problem, while also creating another. 
before it was working well with trimming and that's what I tested with mostly, using the same sample.
now the cleanup for a trimmed capture yields:

```
[gfxrecon] ERROR - /home/crocdialer/dev/trunks/lunarg/gfxreconstruct/external/VulkanMemoryAllocator/include/vk_mem_alloc.h:10495 "Some allocations were not freed before destruction of this memory block!"
...
[gfxrecon] ERROR - DEBUG MESSENGER: Frame #501: VUID-vkDestroyDevice-device-05137: vkDestroyDevice(): Object Tracking - For VkDevice 0x57d3868f4f80, VkCommandBuffer 0x57d38ac2f640 has not been destroyed.
```

also cleanup in `OverrideDestroyDevice` should only cleanup for that particular device.
we have to make sure to test both with and without trimming. trying to provide another fix and get both cases right.

PR #2271, opened 2025-06-30
Fix cleanup for VulkanAddressreplacer

fabian-lunarg:
- account both for trimming and non-trimming case
- cleanup only for relevant device(s)

follow up on https://github.com/LunarG/gfxreconstruct/pull/2267

tested using SaschaWillems/raytracinggltf using `-m rebind` with and without trimming.


locke-lunarg (approved):
Thank you for fixing it.

PR #2272, opened 2025-06-30
Fake support for `VK_GOOGLE_display_timing`

marius-pelegrin-arm:
This commit add fake support for `VK_GOOGLE_display_timing` extension at replay time.
If the extension is supported by the replay device, then it will be properly querried and used normally. If the extension is not supported, then functions will be "faked" by just returning `VK_SUCCESS` silently - as the returned values are not used by the replayer anyway.

This "extension faking" system helps to sort through extensions that can actually be harmful to the replay when not supported, and "harmless" unsupported extensions.

Extending this system also allows to implement custom support for required extensions that could be harmful to the replay.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -3103,8 +3103,8 @@ void VulkanReplayConsumerBase::ModifyCreateDeviceInfo(
                     GFXRECON_LOG_WARNING("faking extension-support for: %s", ext_name);
                     modified_extensions.erase(iter);
                     faked_extensions_.push_back(ext_name);
+                    return true;
                 }
-                return true;
             }
             return false;
         };


lunarpapillo:
The CI failure above appears to be due to an Adnroid failure.  We're working on the issue and will restart the run once we've resolved it.

fabian-lunarg:
I'm not sure about the consequences of adding this mechanism around `IsExtensionBeingFaked(...)`.
while this is occurring on replay-side it might still have side-effects for trimming ranges and alter derived capture-files.

"faking" the extension in this PR is implemented as doing nothing and return VK_SUCCESS.
a recapture would contain the call-ID+data for `vkGetPastPresentationTimingGOOGLE` and report VK_SUCCESS,
but not contain any valid data.

```
// Provided by VK_GOOGLE_display_timing
VkResult vkGetPastPresentationTimingGOOGLE(
    VkDevice                                    device,
    VkSwapchainKHR                              swapchain,
    uint32_t*                                   pPresentationTimingCount, <- not populated
    VkPastPresentationTimingGOOGLE*             pPresentationTimings); <- not populated
```

so the faking would at least need to populate with reasonable defaults, which raises the next question(s).
like, would `pPresentationTimingCount` -> 0,  `pPresentationTimings` -> nullptr work and still qualify as VK_SUCCESS etc.

the biggest issue I'm seeing is this implied inconsistency in derived captures, which does not align well with
gfxr's principles (.md version currently discussed in [this PR](https://github.com/LunarG/gfxreconstruct/pull/2329))

edit: according to https://registry.khronos.org/vulkan/specs/latest/man/html/vkGetPastPresentationTimingGOOGLE.html
is should be valid to return VK_SUCCESS and set `pPresentationTimingCount` -> `0`,  `pPresentationTimings` -> `nullptr`

fabian-lunarg:
@marius-pelegrin-arm can you help me understand or remind me, what was the reason you chose to add the faking-mechanism? i.e. we have a `--remove-unsupported` flag to deal with extensions not being present at replay time.

I'd like to better understand why you preferred this solution,
like is it a spot-solution for `VK_GOOGLE_display_timing` or something generally useful, 
are there other extension that come to mind?

fabian-lunarg:
@marius-pelegrin-arm I added some refactoring and comments, mostly to better motivate what exactly 'faking' means in this context. guess I'm happy with the PR at this point

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -3086,34 +3086,39 @@ void VulkanReplayConsumerBase::ModifyCreateDeviceInfo(
...
+        auto sanitize_faked_extension =
+            [this, &available_extensions, &modified_extensions](const char* ext_name) -> bool {
+            if (graphics::feature_util::IsSupportedExtension(modified_extensions, ext_name) &&
+                !graphics::feature_util::IsSupportedExtension(available_extensions, ext_name))
+            {
+                auto iter = std::find_if(
+                    modified_extensions.begin(), modified_extensions.end(), [ext_name](const char* extension) {
+                        return util::platform::StringCompare(ext_name, extension) == 0;
+                    });
+                modified_extensions.erase(iter);
```
MarkY-LunarG: Should we add an assert, just in case we get an invalid iter here (even though we shouldn't)?

framework/decode/vulkan_replay_consumer_base.h
```diff
@@ -1875,6 +1892,10 @@ class VulkanReplayConsumerBase : public VulkanConsumer
     std::vector<uint32_t>                   capture_image_indices_;
     std::vector<VulkanSwapchainKHRInfo*>    swapchain_infos_;
 
+    //! faked extensions is a list of currently bypassed extensions.
```
MarkY-LunarG: Nit, not sure what the exclamation points are for, so I recommend removing.

framework/decode/vulkan_replay_consumer_base.cpp
MarkY-LunarG: Can we also add a log warning here that we're faking an extension?

framework/decode/vulkan_replay_consumer_base.h
fabian-lunarg: guess a habit. I think that resolves to a one-line doxygen-comment. removed the !

PR #2279, opened 2025-07-03
Add missing 'cmd_index' for various entry-points to json-exports

fabian-lunarg:
closes 2258


fabian-lunarg:
the change/fix is straight-forward but we'll have to adjust our CI's json-groundtruth.
this is also straight-forward but will likely cause some disruption for other branches ...

PR #2280, opened 2025-07-03
Json converter emmit VK_SHADER_UNUSED_KHR instead of int

panos-lunarg:


framework/decode/custom_vulkan_struct_to_json.cpp
```diff
@@ -563,5 +563,58 @@ void FieldToJson(nlohmann::ordered_json& jdata, const Decoded_VkImageToMemoryCop
...
+
+        FieldToJson(jdata["sType"], decoded_value.sType, options);
+        FieldToJson(jdata["type"], decoded_value.type, options);
+        if (decoded_value.generalShader != VK_SHADER_UNUSED_KHR)
+        {
+            FieldToJson(jdata["generalShader"], decoded_value.generalShader, options);
+        }
+        else
+        {
+            jdata["generalShader"] = "VK_SHADER_UNUSED_KHR";
```
bradgrantham-lunarg, line 583: Does this mean that sometimes the output will be an int for a handle and sometimes will be `SHADER_UNUSED`?

panos-lunarg: yes. it should be either an index (not a handle) or SHADER_UNUSED

PR #2281, opened 2025-07-03
settings: Variable Expansion & VkLayerSettingsEXT

antonio-lunarg:
This pull request introduces two main features to GFXReconstruct:

1. **Path Variable Expansion for Dynamic File Paths**
Capture file paths now support variable patterns such as `${AppName}` (application or package name) and `${InternalDataPath}` (Android only).
This allows dynamic construction of file paths based on application or environment-specific information.
Documentation for the capture_file option has been updated across Android, Vulkan, and D3D12 usage guides to describe this new capability.

2. **Vulkan Layer Configuration via VkLayerSettingsEXT**
The Vulkan capture layer can now be configured programmatically using the `VkLayerSettingsEXT` mechanism.
Settings (such as `capture_file`) can be passed directly in the `pNext` chain of `VkInstanceCreateInfo`, enabling configuration in environments where environment variables or settings files are not available.
Usage documentation has been updated with code examples and details on supported settings.


bradgrantham-lunarg (changes requested):
Assuming that `("/data/data/" + (info.AppName + ("/files/" + base_filename)))` is correct, LGTM.  Can this PR also add a note to USAGE_Android noting this behavior?

wpaul-samsung:
Daniel is currently blocked from commenting here due to an IT issue.  Here's his feedback:

I wonder if we should append the files folder in here. I get that the android java getFilesDir function returns that path
[https://developer.android.com/reference/android/content/ContextWrapper.html#getFilesDir()](https://urldefense.com/v3/__https:/developer.android.com/reference/android/content/ContextWrapper.html*getFilesDir()__;Iw!!KUh5zVML9r9m!w-HpWIZbgQ_n6q6qJzbjPgBsuPCHsHZXGWZJJK-1_GurO7C4JzxNnVY0wSlrWtDkiFMY0-KXH1HgWAwNq-PKWlhRD2Y$)
but it literally has files in its name.
For our case, I don't want to place the gfxr inside of the applications files folder, as IMHO the application should have full control and knowledge of the files. Hence we locate the gfxr in the main app folder.
Anyway, I would recommend not to append /files, and I will tests it out without the /files addition.
Just as notes, for completeness
•	looks like the current directory for most applications and services on Android remains on /. So 'relative' paths like data/data//file.gfxr did work before this change. I don't think anyone should use relative paths, but if we want to avoid this possible/theoretical issue we could use some specific pattern such as '${InternalData}/file.gfxr, and only replace that pattern. If we only support the pattern at the start of the path, then it also does not change any use of an absolute path (which IMHO are all the reasonable ones)
•	Android since 4.2 supports multiple users, and for those the folder is in /data/user// (and possibly /files). We don't support this for now, and even with this PR in this form, we still could support it for the normal short package ids. So I think not supporting this is not an real issue, we can support it if it shows up, not that hard anyhow.
•	this changes it only for the package name, not for the json file defined by the property. We could apply the same change (via shared function, of course) there. We are using this property, so this note is just for completeness.

framework/util/file_path.cpp
```diff
@@ -595,6 +595,62 @@ std::string FindModulePath(const std::string& target_module, bool case_sensitive
...
+        while (pos != std::string::npos)
+        {
+            std::string replacement = "";
+
+            switch (static_cast<PathVariable>(i))
+            {
+                case PathVariable::kAppName:
+                    replacement = info.AppName;
+                    break;
+                case PathVariable::kInternalDataPath:
```
dfriederich, line 637: nit maybe should have some #ifdef __ANDROID__?

dfriederich:
I will adapt my change to use VkLayerSettingsEXT, hope to report if it works for me tomorrow.

USAGE_android.md
```diff
@@ -361,6 +361,50 @@ in the GFXReconstruct GitHub repository at `layer/vk_layer_settings.txt`. Most
 binary distributions of the GFXReconstruct software will also include a sample
 settings file.
 
+#### Layer Settings via VkLayerSettingsEXT
+
+A new way to configure the GFXReconstruct Vulkan capture layer is via the Vulkan
+`VkLayerSettingsEXT` mechanism, which allows settings to be passed directly through the
```
dfriederich: VkLayerSettingsEXT->VkLayerSettingEXT

USAGE_android.md
```diff
@@ -361,6 +361,50 @@ in the GFXReconstruct GitHub repository at `layer/vk_layer_settings.txt`. Most
 binary distributions of the GFXReconstruct software will also include a sample
 settings file.
 
+#### Layer Settings via VkLayerSettingEXT
+
+A new way to configure the GFXReconstruct Vulkan capture layer is via the Vulkan
```
mikes-lunarg: nit: Prefer "alternate" to "new" since nobody is going to come back and remove the new when it isn't new anymore.

USAGE_android.md
```diff
@@ -361,6 +361,50 @@ in the GFXReconstruct GitHub repository at `layer/vk_layer_settings.txt`. Most
...
+of your `VkInstanceCreateInfo`, and include settings for the
+`VK_LAYER_LUNARG_gfxreconstruct` layer. For example, to set the capture file name:
+
+```c
+VkLayerSettingEXT capture_file_setting = {
+    .pLayerName = "VK_LAYER_LUNARG_gfxreconstruct",
+    .pSettingName = "capture_file",
+    .type = VK_LAYER_SETTING_TYPE_STRING_EXT,
+    .valueCount = strlen("my_capture.gfxr"),
+    .pValues = "my_capture.gfxr"
```
mikes-lunarg: `valueCount` should be 1 and `pValues` should be an array of `char*`

See https://github.com/KhronosGroup/Vulkan-Utility-Libraries/blob/main/docs/layer_configuration.md#code-example-to-configure-the-validation-layer-programmatically for some existing examples

USAGE_android.md
```diff
@@ -361,6 +361,50 @@ in the GFXReconstruct GitHub repository at `layer/vk_layer_settings.txt`. Most
 binary distributions of the GFXReconstruct software will also include a sample
 settings file.
 
+#### Layer Settings via VkLayerSettingEXT
```
mikes-lunarg: I think we should refer to this mechanism by the extension name `VK_EXT_layer_settings`

framework/encode/capture_settings.cpp
```diff
@@ -300,12 +190,16 @@ CaptureSettings::CaptureSettings(const TraceSettings& trace_settings)
...
+void CaptureSettings::LoadSettings(CaptureSettings* settings, const CaptureLayerSettings* layer_settings)
 {
     if (settings != nullptr)
     {
         OptionsMap capture_settings;
 
+        if (layer_settings != nullptr)
+        {
+            LoadOptionsLayer(&capture_settings, layer_settings);
+        }
```
mikes-lunarg: Instead of maintaining and populating a new settings struct, would it be possible to load everything we find in `VkLayerSettingsCreateInfoEXT` into a dumb key/value store (e.g. `OptionsMap`) and then reuse the existing processing code? This is essentially how `LoadOptionsFile` works, it just grabs whatever is in the file and lets `ProcessOptions` sort it out.

framework/encode/vulkan_capture_layer_settings.h
```diff
@@ -0,0 +1,103 @@
...
+                    }
+                    else if (setting.valueCount == 0 || setting.pValues == nullptr)
+                    {
+                        GFXRECON_LOG_ERROR("Expected non-empty value for '%s' setting, but got empty value",
+                                           CAPTURE_FILE_NAME_LOWER);
+                    }
+                    else
+                    {
+                        layer_settings.capture_filename =
+                            std::string(static_cast<const char*>(setting.pValues), setting.valueCount);
```
mikes-lunarg: Just another reminder that `pValues` is an array of null-terminated `char*` and `valueCount` is the length of the array (1)

mikes-lunarg (changes requested):
Since we are implementing `VK_EXT_layer_settings` it needs to go in the layer manifest file like this: https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/layers/VkLayer_khronos_validation.json.in#L28-L31

I don't remember if there are additional steps required to make it discoverable by applications via `vkEnumerateInstanceExtensionProperties` but it is worth testing.

USAGE_android.md
antonio-lunarg: Fixed and added some unit tests.

framework/encode/capture_settings.cpp
antonio-lunarg: Great point! I implemented a solution which relies on `GetDefaultTraceSettings()`, but I had to call `InitSingleton()` before `CreateInstance()` (and I discovered [this bug](https://github.com/LunarG/gfxreconstruct/issues/2299)!). Let me know your thoughts.

dfriederich:
I tried again with the latest PR, but now it fails for me to pass the gfxr filename via VkLayerSettingsEXT
 :-(.

dfriederich (changes requested):
not exactly sure, but the current revision does not work for me. Could be my fault, but it worked just fine with a prior revision with my changes.

dfriederich:
I think it is my bug, I did misunderstand the VK_LAYER_SETTING_TYPE_STRING_EXT setting. I took it to mean that string's value is defined in VkLayerSettingEXT are  valueCount characters starting at where pValues points to, but I guess it is defined that pValues points to an array of 'char*' pointers, each pointing to a zero terminated string , and valueCount should be 1 in our case as we just pass one string. Will update the sokatoa side PR.

dfriederich (approved):
Issue was my fault, my misunderstanding of the VK_LAYER_SETTING_TYPE_STRING_EXT format. It works with long path's as it is.

USAGE_android.md
```diff
@@ -361,6 +361,52 @@ in the GFXReconstruct GitHub repository at `layer/vk_layer_settings.txt`. Most
 binary distributions of the GFXReconstruct software will also include a sample
 settings file.
 
+#### Layer Settings via VK_EXT_layer_settings
+
+An alternative way to configure the GFXReconstruct Vulkan capture layer is via the Vulkan
+`VK_EXT_layer_settings` layer, which allows settings to be passed directly through the
```
bradgrantham-lunarg: ^ VK_EXT_layer_settings layer^VK_EXT_layer_settings extension^ ?

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -80,6 +81,8 @@ bool VulkanCaptureManager::CreateInstance()
 
 VulkanCaptureManager* VulkanCaptureManager::InitSingleton()
 {
+    std::lock_guard<std::mutex> instance_lock(instance_lock_);
```
bradgrantham-lunarg, line 84: Can you give me a rundown of why you added this?  If there's a potential path for two threads to call this function then we need to probably also protect the other capture manager `InitSingleton`s (OpenXR, DX12) and maybe elevate it to the parent class.

antonio-lunarg: Thank you for this question, @bradgrantham-lunarg ! I created an issue just for that: https://github.com/LunarG/gfxreconstruct/issues/2299. In short, the parent class is already using a lock, but that lock is not properly guarding the creation of the other capture managers.

The change I introduce here is not addressing this bug (well, it does it indirectly). I made this change for a matter of initialization order, as I needed to make sure that `VulkanCaptureManager::layer_settings_` is initialized before the function call below: `CreateInstance()`.

The common capture manager can get the underlying API capture manager layer settings by calling `ApiCaptureManager::GetDefaultTraceSettings()`, therefore I wanted to make use of this functionality.

In practice this will happen:
- VCM: VulkanCaptureManager
- CCM: CommonCaptureManager
```mermaid
%%{init: {'themeVariables': { 'fontSize': '11px'}}}%%
flowchart TD
    Z[VCM::InitSingleton] -->|VCM::singleton_ now ready| Y
    Y[GetVulkanLayerTraceSettings] -->|VCM::layer_settings_ now ready| A
    A[VCM::CreateInstance] --> CCM::CreateInstance

    subgraph CCM::CreateInstance
      direction TB
      B[CCM::CreateInstance < VCM >] --> C
      C[VCM::InitSingleton] -->|VCM::singleton_ already initialized| D
      D[CCM::CreateInstance] --> E
      E[CCM::LockedCreateInstance] --> F
      F[VCM::GetDefaultTraceSettings] -->|returns VCM::layer_settings_| G
      G@{ shape: sm-circ, label: "Small start" }
    end
  
```

bradgrantham-lunarg: I think all called functions are protected by a lock in `vkCreateInstance`, so I still don't understand why this is necessary.  But it seems safe enough to me, so I'm going to approve it and merge, then we'll discuss later.

PR #2283, opened 2025-07-07
VulkanAddressReplacer: wait for timeline-semaphores when inside renderpass

fabian-lunarg:
Fixes an error encountered when using `-m rebind` in a non-trivial device-address sample:

when performing address-replacement we try to inject a compute-shader dispatch
into the same command-buffer(A) that will consume the address-data.

by spec this is not possible inside renderpasses though and we fallback to a local submit + fence-sync,
right before queue-submission of A.

in this case we shall inherit all wait-semaphores of A before starting address-replacement,
which is fixed in this PR.

also add an additional (pre-) pipeline-barrier into VulkanAddressReplacer, for good measure.


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4088,11 +4099,19 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit2(PFN_vkQueueSubmit2     f
                 addresses_to_replace.insert(addresses_to_replace.end(),
                                             command_buffer_info->addresses_to_replace.begin(),
                                             command_buffer_info->addresses_to_replace.end());
+                sync_wait_semaphores |= !command_buffer_info->addresses_to_replace.empty();
```
antonio-lunarg, line 4155: The corresponding line in `OverrideQueueSubmit` does not have a `!`. Is that on purpose?

fabian-lunarg: no, will fix :)

fabian-lunarg: this location was correct, corrected the other

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4078,8 +4086,10 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit2(PFN_vkQueueSubmit2     f
 
         for (uint32_t i = 0; i < submitCount; i++)
         {
-            uint32_t num_command_buffers = submit_info_data[i].pCommandBufferInfos->GetLength();
-            auto*    cmd_buf_info_metas  = submit_info_data[i].pCommandBufferInfos->GetMetaStructPointer();
+            uint32_t num_command_buffers  = submit_info_data[i].pCommandBufferInfos->GetLength();
+            auto*    cmd_buf_info_metas   = submit_info_data[i].pCommandBufferInfos->GetMetaStructPointer();
+            bool     sync_wait_semaphores = false;
+
```
MarkY-LunarG, line 4143: Don't we need to strip out the semaphores for the original submit as well?  Otherwise aren't we going to hit a possible wait when un-signaled issue?  Like maybe inside the call to `SyncWaitSemaphores` we end  up removing the waited semaphores from the `submit_infos`?

fabian-lunarg: good point, looking into it

fabian-lunarg: done! complete rewrite of the related synchronization-solution. all done by re-wiring wait-semaphores to an intermediate replacer-submission and added signal-semaphore. no more fence-sync, no wait-before-signal hazards. lifetime of internal replacer-resources properly managed.

MarkY-LunarG (approved):

PR #2285, opened 2025-07-07
docs: Update about SDK OpenXR support

MarkY-LunarG:


KarenGhavam-lunarG:
@MarkY-LunarG 
The README.md is not pulled into the SDK for the SDK documentation. The USAGE_desktop_Vulkan is what is pulled into the SDK documentation. I think that file should be updated. And if there is one for Android, an update should be done there as well.

PR #2286, opened 2025-07-08
Allocator supports more functions

locke-lunarg:
Test code for QueueBindSparse: `gfxreconstruct\test\test_apps\sparse-resources` and  https://github.com/SaschaWillems/Vulkan/tree/master/examples/texturesparseresidency

A lot of functions that have memory parameters needs to be supported by allocator.

For rebind, `vkAllocateMemory` didn't create a memory, only save the memory info. It creates an allocation for memory when a object binds a memory. One allocation is for one binding case. It uses `allocation_info.offset` to bind, instead of original offset. 

`VideoSession` and `BindSparse` are different cases that one object could bind multiple memories. In this case, `ResourceAllocInfo` becomes have a `vector<BoundMemoryInfo>`.

Plus, a object could re-bind the same memory through `BindSparse`, so finding existing allocation first before creating a new one.  If it re-bind the same memory, but different offset, it needs to create a new allocation for it. 

I think we could have `VulkanResourceAllocatorBase` by auto generation. When we update headers, it writes virtual functions that need to be handled by allocator in `VulkanResourceAllocatorBase`. Therefore, we could know what we need to do. Added one issue https://github.com/LunarG/gfxreconstruct/issues/2289.

### DIFF
--- a/framework/decode/vulkan_default_allocator.h
+++ b/framework/decode/vulkan_default_allocator.h
@@ -362,15 +362,12 @@ class VulkanDefaultAllocator : public VulkanResourceAllocator
                                      VkFence                 fence,
                                      ResourceData*           allocator_buf_datas,
                                      const MemoryData*       allocator_buf_mem_datas,
-                                     const format::HandleId* buf_mem_capture_ids,
                                      VkMemoryPropertyFlags*  bind_buf_mem_properties,
                                      ResourceData*           allocator_img_op_datas,
                                      const MemoryData*       allocator_img_op_mem_datas,
-                                     const format::HandleId* img_op_mem_capture_ids,
                                      VkMemoryPropertyFlags*  bind_img_op_mem_properties,
                                      ResourceData*           allocator_img_datas,
                                      const MemoryData*       allocator_img_mem_datas,
-                                     const format::HandleId* img_mem_capture_ids,
                                      VkMemoryPropertyFlags*  bind_img_mem_properties) override;
 
     virtual uint64_t GetDeviceMemoryOpaqueCaptureAddress(const VkDeviceMemoryOpaqueCaptureAddressInfo* info,
--- a/framework/decode/vulkan_default_allocator.h
+++ b/framework/decode/vulkan_default_allocator.h
@@ -76,11 +76,11 @@ class VulkanDefaultAllocator : public VulkanResourceAllocator
                                         const VkAllocationCallbacks*       allocation_callbacks,
                                         format::HandleId                   capture_id,
                                         VkVideoSessionKHR*                 session,
-                                        std::vector<ResourceData>*         allocator_datas) override;
+                                        ResourceData*                      allocator_data) override;
 
     virtual void DestroyVideoSession(VkVideoSessionKHR            session,
                                      const VkAllocationCallbacks* allocation_callbacks,
-                                     std::vector<ResourceData>    allocator_datas) override;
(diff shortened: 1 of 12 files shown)


framework/decode/vulkan_default_allocator.h
```diff
@@ -305,12 +327,76 @@ class VulkanDefaultAllocator : public VulkanResourceAllocator
...
+    virtual void DestroyAccelerationStructureNV(VkAccelerationStructureNV    acc_str,
+                                                const VkAllocationCallbacks* allocation_callbacks,
+                                                ResourceData                 allocator_data) override;
+
+    virtual void
+    GetAccelerationStructureMemoryRequirementsNV(const VkAccelerationStructureMemoryRequirementsInfoNV* info,
+                                                 VkMemoryRequirements2KHR* memory_requirements,
+                                                 ResourceData              allocator_data) override;
+
+    virtual VkResult BindAccelerationStructureMemoryNV(uint32_t                                       bind_info_count,
```
fabian-lunarg, line 351: two questions wrt. AccelerationStructureNV

- is this also required for KHR? (maybe also, how relevant is NV vs. KHR looking forward?)
- if required to have in allocator's Baseclass-API, could this be a delegate? (at adds a lot of API and makes default_allocator harder to read)

locke-lunarg: `VkAccelerationStructureNV` and `VkAccelerationStructureKHR` are different. 

`VkAccelerationStructureNV`  is from `VK_NV_ray_tracing`
https://registry.khronos.org/vulkan/specs/latest/man/html/VK_NV_ray_tracing.html

`VkAccelerationStructureKHR` is from `VK_KHR_acceleration_structure`
`VkAccelerationStructureKHR` doesn't  bind a memory.
https://registry.khronos.org/vulkan/specs/latest/man/html/VK_KHR_acceleration_structure.html

fabian-lunarg: ok, then the other part:

why does it make sense to support `VkAccelerationStructureNV`? 

e.g. nvidia's latest vendor-extensions build on top of `VkAccelerationStructureKHR` not their old NV counterpart.
-> e.g. https://registry.khronos.org/vulkan/specs/latest/man/html/VK_NV_cluster_acceleration_structure.html
it is my impression `VkAccelerationStructureNV` has been replaced for good by the almost identical KHR-variant.
this would leave only backwards-compatibility as reason to support it?
(I know we do have some code supporting it, but what is the reason to add new code supporting it?)

locke-lunarg: I don't have a good reason to support it. I just found what we missed and added it. I can remove it, if we all agree it's better not to support it. @bradgrantham  What do you think?

locke-lunarg: Removed code about `VkAccelerationStructureNV`.

fabian-lunarg (approved):
the changes make sense and look good. with that, I'm curious to try out some queue bind-sparse cases I'm currently seeing issues with.

locke-lunarg:
@fabian-lunarg Excuse me. I added a commit to skip function if the memory is null. Please review again. Yes, please test it for the issue of queue bind-sparse.  Let me know if you see any issues. Thank you.

fabian-lunarg (approved):

bradgrantham-lunarg (approved):

PR #2290, opened 2025-07-09
Correct Pipeline Cache UUID output in gfxrecon-info

jzulauf-lunarg:
Noticed I was getting inconsistent values output from gfxrecon-info for the pipeline cache UUID field.  Turns out we were writing the address of the array in gfxrecon-info.  Changed info to do the same thing gfxrecon-convert does.


tools/info/main.cpp
```diff
@@ -525,7 +525,9 @@ void PrintVulkanStats(const gfxrecon::decode::FileProcessor&       file_processo
             GFXRECON_WRITE_CONSOLE("\tVendor ID: 0x%x", props.second.vendorID);
             GFXRECON_WRITE_CONSOLE("\tDevice ID: 0x%x", props.second.deviceID);
             GFXRECON_WRITE_CONSOLE("\tDevice type: %u", props.second.deviceType);
-            GFXRECON_WRITE_CONSOLE("\tPipeline cache UUID: 0x%x", props.second.pipelineCacheUUID);
+            constexpr size_t uuid_size   = sizeof(props.second.pipelineCacheUUID);
```
fabian-lunarg: this works because pipeCacheUUID is a `uint8_t[VK_UUID_SIZE]` - it would be slightly more readable,
to just use `uuid_size = VK_UUID_SIZE` or just use `VK_UUID_SIZE` below.

jzulauf-lunarg: Agreed.

bradgrantham-lunarg (approved):

PR #2294, opened 2025-07-10
Optional environment variables

### LINKED ISSUE #2193: Environment capture may be saving personally-identifying information
Environment capture may be saving personally-identifying information such as user ID and steam username, etc.

Need to have allow-list for environment variables/properties to be captured.  Potentially add an option to allow additional variables for capture or capture all.

mikes-lunarg:
Switch from capturing all environment variables to only those which are specified through the GFXRECON_CAPTURE_ENVIRONMENT option

fixes #2193


bradgrantham-lunarg:
The reason for the original feature (#1663) was to save `MESA_SHADER_CACHE_DIR` to enable it being used in replay.  Can we leave MESA_SHADER_CACHE_DIR in there or do we believe that is inappropriate without explicit specification?  (If I recall correctly this came out of a request for replay on Mesa reusing the cache created in capture - @zmike?)

zmike:
Hm I think you probably should change it to be an explicit opt-in. We definitely have good reason to want to capture `MESA_SHADER_CACHE_DIR`, but I don't think we always want it to be captured.

zmike:
(so this change is probably fine)

mikes-lunarg:
My reasoning was that since replay substitution is already explicit, making the capture side explicit isn't much more of a burden, and now it matches

antonio-lunarg (approved):

bradgrantham-lunarg (approved):
Approved but I think the GitHub CI tests should pass before merging.

mikes-lunarg:
> Approved but I think the GitHub CI tests should pass before merging.

That seems like a reasonable request :stuck_out_tongue_winking_eye:

CI is now passing. Also, since the new functionality runs fine on macOS, I think we can close #1983

PR #2303, opened 2025-07-16
script: Build generator

antonio-lunarg:
With this, `build.py` accepts a CMake generator as input, e.g.:
```console
python3 ./scripts/build.py --generator Ninja
```


fabian-lunarg (approved):
good change

scripts/build.py
```diff
@@ -240,6 +244,9 @@ def cmake_generate_options(args):
             '-DBUILD_LAUNCHER_AND_INTERCEPTOR={}'.format(
                 'OFF' if not args.build_launcher else 'ON'))
         generate_options.extend('-D' + arg for arg in args.cmake_extra)
+        if args.generator:
+            generate_options.append('-G{}'.format(args.generator))
+            generate_options.append("-DCMAKE_CXX_FLAGS=-fdiagnostics-color=always")
```
bradgrantham-lunarg: I'm not sure this should always be on if a generator is provided - is this line a leftover from changes local to your machine?

antonio-lunarg: Good point. I think it's valid for gcc and clang, but might not be recognized by other compilers, and it should not be tied to a generator anyway.

PR #2304, opened 2025-07-16
Reserve ParentToChild Metacommand ids

laurentiu-erhan-arm:


fabian-lunarg:
@laurentiu-erhan-arm could you share some information about the purpose for the reserved meta-command Ids?

laurentiu-erhan-arm:
Sure, these 2 are used for handling the compaction of acceleration structures and opacity micromaps in the raytracing postprocessing approach ( https://github.com/ARM-software/gfxreconstruct-arm/blob/dev-arm/framework/format/format.h )

fabian-lunarg (dismissed):
corresponds to enum-values present in ARM's branch:

```
enum ParentToChildDependencyType : uint32_t
{
    kUnknownDependency                         = 0,
    kAccelerationStructuresDependency          = 1,
    kMicromapCompactionDependency              = 2,
    kAccelerationStructureCompactionDependency = 3
};
```

bradgrantham-lunarg:
Agree with @fabian-lunarg 's assessment.  This seems (in absence of more context) like feature-specific item that we are unlikely to add to LunarG/gfxreconstruct so would only add maintenance burden without adding functionality.

fabian-lunarg:
closing this PR with the explanation given above. we cannot guarantee this type will stay on our side. 
to still avoid potential collisions of enum-values in the future, our suggestion would be to issue enum-values with the highest bit set.

PR #2306, opened 2025-07-17
Set device info for deduplicate device

locke-lunarg:
Some discussion need for it.  Should the second device info  be from its create info? or from the source device info? This PR chose its create info because it passed test, but the source device info failed during replay. I didn't look deeply the reason.

Plus, should we add more requirements for deduplicate device? Like both create info should be similar and their instances should be the same. Is it possible to have deduplicate instance option?

I added two issues about it.
https://github.com/LunarG/gfxreconstruct/issues/2311
https://github.com/LunarG/gfxreconstruct/issues/2312

### DIFF
--- a/USAGE_desktop_Vulkan.md
+++ b/USAGE_desktop_Vulkan.md
@@ -631,6 +631,7 @@ gfxrecon-replay         [-h | --help] [--version] [--cpu-mask <binary-mask>] [--
                         [--dump-resources-dump-all-image-subresources] <file>
                         [--pbi-all] [--pbis <index1,index2>]
                         [--pipeline-creation-jobs | --pcj <num_jobs>]
+                        [--deduplicate-device]
 
 
 Required arguments:
@@ -898,6 +899,9 @@ Optional arguments:
                         `--load-pipeline-cache`.
   --quit-after-frame
               Specify a frame after which replay will terminate.
+
+  --deduplicate-device
+              If set, at most one VkDevice will be created for each VkPhysicalDevice.
 ```
 
 ### Key Controls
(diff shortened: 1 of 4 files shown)


zmike:
I would think gfxr should be able to analyze the device create infos and then merge them so that the combined extensions+features for both devices are enabled on the deduplicated device.

bradgrantham-lunarg:
> I would think gfxr should be able to analyze the device create infos and then merge them so that the combined extensions+features for both devices are enabled on the deduplicated device.

The layer has already opened the first device by the time it sees the second CreateDevice, so I think the best it can do is de-duplicate if the second CreateDevice is a subset of the first.

bradgrantham-lunarg (approved):

locke-lunarg:
I changed that the second device info was copied from the source device info. I attempted making them to have the same memory address of device info to avoid data sync issues. But some their data have to be different, like capture_id and  duplicate_source_id. In this case, they have individual device info. I'm not sure if they will have data sync issues in somewhere,

@bradgrantham-lunarg  please review again.

locke-lunarg:
For dx12 object info, the object info have DxObjectInfo and DxObjectExtraInfo for the basic data and extra data. For deduplicate device, they could have individual DxObjectInfo, but the same DxObjectExtraInfo to avoid data sync issues. But Vulkan object info doesn't have that design.

locke-lunarg:
I added one PR to change to check `physicalDevice` , instead of `pipelineCacheUUID`.  `pipelineCacheUUID` could be the same, even if the `physicalDevice` or `deviceName ` are different. Using `physicalDevice` could avoid different instance issues.

PR #2307, opened 2025-07-17
android, build: Do not build 32b Android by default

ncesario-lunarg:
Removes 32 bit builds from the default gradle configurations.

Fixes LunarG/Projects#969.


MarkY-LunarG (approved):
LGTM

PR #2308, opened 2025-07-18
convert: Fix binary out dir when stdout used

### LINKED ISSUE #2202: gfxrecon-convert dumps binaries to 'stdout' directory
Command:
`gfxrecon-convert --output stdout --include-binaries trace.gfxr`
will currently send json data to stdout and create "stdout" directory in the trace directory for the binary data.

The expected behavior is the binaries are created in the directory that matches the trace file name, as by default.

Additionally, for automation purposes it would be useful to have a `--binary-output` parameter allowing to specify a separate binary directory.

MarkY-LunarG:
When stdout is set for the output, and a "--include-binaries" option is used, we were outputting to a folder called "stdout". This should not have been the case because according to our own usage info, it should have reverted to the GFXR capture file name.

Fixes #2202

### DIFF
--- a/framework/decode/json_writer.cpp
+++ b/framework/decode/json_writer.cpp
@@ -224,7 +224,7 @@ void RepresentBinaryFile(JsonWriter&             writer,
     {
         std::string filename = writer.GenerateFilename(filename_base);
         std::string basename = gfxrecon::util::filepath::Join(json_options.data_sub_dir, filename);
-        std::string filepath = gfxrecon::util::filepath::Join(json_options.root_dir, basename);
+        std::string filepath = gfxrecon::util::filepath::Join(json_options.root_dir, filename);
         if (writer.WriteBinaryFile(filepath, data_size, data))
         {
             FieldToJson(jdata, basename, json_options);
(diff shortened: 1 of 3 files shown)


fabian-lunarg:
wrt. compilation-issue. we'll need to rebase after https://github.com/LunarG/gfxreconstruct/pull/2309 has been merged.

framework/decode/json_writer.cpp
```diff
@@ -223,11 +223,16 @@ void RepresentBinaryFile(JsonWriter&             writer,
...
         if (writer.WriteBinaryFile(filepath, data_size, data))
         {
-            FieldToJson(jdata, basename, json_options);
+            std::string file_only_name = filename;
+            size_t      last_slash_pos = file_only_name.find_last_of("/\\");
+            if (last_slash_pos >= 0)
+            {
+                file_only_name.replace(0, last_slash_pos + 1, "");
+            }
+            FieldToJson(jdata, file_only_name, json_options);
```
fabian-lunarg: we could use [STL](https://en.cppreference.com/w/cpp/filesystem/path/filename.html) like `std::string file_only_name = std::filesystem::path(filepath).filename().string();` ?

MarkY-LunarG: I forgot we're up to 17 at least now.  Thanks for reminding me!

fabian-lunarg:
95% sure the CI fail is due to recent .jsonl groundtruth-update. in this case a rebase should fix it

PR #2315, opened 2025-07-23
VDR: Add option to compress files that are dumped raw

panos-lunarg:
`--dump-resources-binary-file-compression-type` chooses between the available options. Default is uncompressed

### DIFF
--- a/framework/decode/vulkan_replay_dump_resources_json.cpp
+++ b/framework/decode/vulkan_replay_dump_resources_json.cpp
@@ -20,6 +20,7 @@
 ** DEALINGS IN THE SOFTWARE.
 */
 
+#include "format/format_util.h"
 #include "Vulkan-Utility-Libraries/vk_format_utils.h"
 #include "util/file_path.h"
 #include PROJECT_VERSION_HEADER_FILE
@@ -55,6 +56,8 @@ VulkanReplayDumpResourcesJson::VulkanReplayDumpResourcesJson(const VulkanReplayO
     dr_options["dumpResourcesDumpRawImages"]            = options.dump_resources_dump_raw_images;
     dr_options["dumpResourcesDumpSeparateAlpha"]        = options.dump_resources_dump_separate_alpha;
     dr_options["dumpResourcesDumpUnusedVertexBindings"] = options.dump_resources_dump_unused_vertex_bindings;
+    dr_options["dumpResourcesBinaryFileCompressionType"] =
+        format::GetCompressionTypeName(options.dump_resources_binary_file_compression_type);
 };
 
 bool VulkanReplayDumpResourcesJson::InitializeFile(const std::string& filename)
--- a/framework/decode/vulkan_replay_dump_resources_json.cpp
+++ b/framework/decode/vulkan_replay_dump_resources_json.cpp
@@ -21,8 +21,11 @@
 */
 
 #include "format/format_util.h"
+#include "util/file_path.h"
 #include "Vulkan-Utility-Libraries/vk_format_utils.h"
 #include "util/file_path.h"
+#include "util/logging.h"
+#include <cstdint>
 #include PROJECT_VERSION_HEADER_FILE
 #include "generated/generated_vulkan_enum_to_string.h"
 #include "vulkan_replay_dump_resources_json.h"
@@ -165,17 +168,19 @@ nlohmann::ordered_json& VulkanReplayDumpResourcesJson::GetCurrentSubEntry()
     return current_entry != nullptr ? *current_entry : json_data_;
 }
 
-void VulkanReplayDumpResourcesJson::InsertImageSubresourceInfo(nlohmann::ordered_json& json_entry,
-                                                               VkFormat                image_format,
-                                                               VkImageType             image_type,
-                                                               format::HandleId        image_id,
-                                                               const VkExtent3D&       extent,
-                                                               const std::string&      filename,
-                                                               VkImageAspectFlagBits   aspect,
-                                                               uint32_t                mip_level,
-                                                               uint32_t                array_layer,
-                                                               bool                    separate_alpha,
-                                                               const std::string*      filename_before)
+void VulkanReplayDumpResourcesJson::InsertImageSubresourceInfo(nlohmann::ordered_json&      json_entry,
+                                                               VkFormat                     image_format,
+                                                               VkImageType                  image_type,
+                                                               format::HandleId             image_id,
+                                                               const VkExtent3D&            extent,
+                                                               const std::string&           filename,
+                                                               VkImageAspectFlagBits        aspect,
+                                                               uint32_t                     layer_count,
+                                                               uint32_t                     mip_level,
+                                                               uint32_t                     array_layer,
+                                                               const std::vector<uint64_t>* subresource_sizes,
+                                                               bool                         separate_alpha,
+                                                               const std::string*           filename_before)
 {
     const std::string aspect_str_whole(util::ToString<VkImageAspectFlagBits>(aspect));
     const std::string aspect_str(aspect_str_whole.begin() + 16, aspect_str_whole.end() - 4);
@@ -216,6 +221,21 @@ void VulkanReplayDumpResourcesJson::InsertImageSubresourceInfo(nlohmann::ordered
         else
         {
             json_entry["file"] = filename;
+        }
+    }
+
+    if (subresource_sizes != nullptr)
+    {
+        const std::string file_extension      = util::filepath::GetFilenameExtension(filename);
+        const bool        is_image_dumped_raw = !file_extension.compare(".bin");
+
+        if (is_image_dumped_raw)
+        {
+            const uint32_t sub_res_idx = mip_level * layer_count + array_layer;
+            GFXRECON_ASSERT(sub_res_idx < subresource_sizes->size());
+
+            const uint64_t subresource_size = (*subresource_sizes)[sub_res_idx];
+            json_entry["size"]              = subresource_size;
         }
     }
 }
--- a/framework/graphics/vulkan_resources_util.h
+++ b/framework/graphics/vulkan_resources_util.h
@@ -61,8 +61,8 @@ class VulkanResourcesUtil
     // resource and the size of the biggest staging buffer necessary is known in advance.
     VkResult CreateStagingBuffer(VkDeviceSize size);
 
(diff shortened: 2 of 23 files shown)


tools/replay/replay_settings.h
```diff
@@ -357,6 +357,9 @@ static void PrintUsage(const char* exe_name)
     GFXRECON_WRITE_CONSOLE("          \t\tDump all available mip levels and layers when dumping images.");
     GFXRECON_WRITE_CONSOLE("  --dump-resources-dump-unused-vertex-bindings");
     GFXRECON_WRITE_CONSOLE("          \t\tDump a vertex binding even if no vertex attributes references it.");
+    GFXRECON_WRITE_CONSOLE("  --dump-resources-binary-file-compression-type");
+    GFXRECON_WRITE_CONSOLE("          \t\tCompress files that are dumped as binary. Available compression types");
+    GFXRECON_WRITE_CONSOLE("          \t\tare: [none, lz4, zlib, zlib, zstd]. Default is none (no compression).");
```
dfriederich: nit 2x zlib

tools/tool_settings.h
```diff
@@ -203,6 +205,11 @@ const char kSwapchainOffscreen[] = "offscreen";
 const char kScreenshotFormatBmp[] = "bmp";
 const char kScreenshotFormatPng[] = "png";
 
+const char kCompressionTypeNone[] = "none";
```
bradgrantham-lunarg, line 209: I think these should live in `gfxrecon::format` next to the CompressionType enumerant, maybe even a compile-time map from string to enum?

csteinberg-samsung:
one thing that has come up in attempting to integrate this is that the output of the file names in the output json is still being called .bin when it has a different extension now. Either the path should be updated to the new file extension based on compression format or the compression format should be listed in the output json.  (ideally the paths are updated)

bradgrantham-lunarg:
> one thing that has come up in attempting to integrate this is that the output of the file names in the output json is still being called .bin when it has a different extension now. Either the path should be updated to the new file extension based on compression format or the compression format should be listed in the output json. (ideally the paths are updated)

It should be in the output JSON: https://github.com/LunarG/gfxreconstruct/blob/ed21aebcfa3b48118f826a700b1707120ecce7e3/framework/decode/vulkan_replay_dump_resources_json.cpp#L59

csteinberg-samsung:
ah yes we see that now. We don't copy those headers into our api-cache.json so we missed it. We have a workaround now so I think this is fine.

dfriederich:
The zlib format worked for me so far. I did run into some issues I had to address on our side, and did not test much but very simple index/vertex buffers, but it works.
I did just start to implement lz4 support, and that worked for one case, but in a second case I get an error from the decode library I'm using. Might be my bug (always most likely), might be the decode library, might be gfxr, don't know yet :-)
```
sample_hdr_15/Screenshots/Api/1/vulkan_sample_hdr_indexBuffer_qs_434_bcb_290_dc_303_uint32.lz4' succeeded
2025-07-24T01:48:26.597Z root WARN Failed to read binary: the offset to copy is not contained in the decompressed buffer
```
Will have to continue implementing/testing tomorrow.

dfriederich:
My changes are in https://github.com/android-graphics/sokatoa/pull/3588

dfriederich:
the lz4 file which the lib I'm using chokes on.
[error.zip](https://github.com/user-attachments/files/21398402/error.zip)

tools/tool_settings.h
panos-lunarg: I put these here in an attempt to follow the coding style of this file which also contains other strings in the previous 20 lines or so, with the same purpose (compare them against cli arguments).

Also `format_util.cpp` defines `GetCompressionTypeName` which converts the enum into a string.
So it feels kind of a duplicate to put these short strings in `format.h` which contains definitions strictly related to the gfxr file format.

panos-lunarg:
> one thing that has come up in attempting to integrate this is that the output of the file names in the output json is still being called .bin when it has a different extension now. Either the path should be updated to the new file extension based on compression format or the compression format should be listed in the output json. (ideally the paths are updated)

The file extension hasn't changed. It's still `.bin`.
Also as Brad mentioned a new entry has been added in the json header which states the compression method used.

tools/replay/replay_settings.h
panos-lunarg: fixed

panos-lunarg:
@dfriederich gfxr uses pre-compiled binaries for the compression libraries. Could the problems you are seeing be mismatches/conflicts between the library gfxr used to compress and the library you are using to decompress?

wpaul-samsung:
> @dfriederich gfxr uses pre-compiled binaries for the compression libraries. Could the problems you are seeing be mismatches/conflicts between the library gfxr used to compress and the library you are using to decompress?

It looks like GFXR uses LZ4_compress_fast and LZ4_decompress_safe.  According to https://github.com/lz4/lz4/issues/1504, LZ4_compress_fast uses the legacy format (equivalent to lz4 -l1) from the CLI.  With lz4 CLI version 1.10.0, I'm unable to decompress these files compressed with LZ4_compress_fast().  I get `Error 44 : Unrecognized header : file cannot be decoded`.

I don't see a way to decompress the legacy format from the CLI.  We're actually using lz4-napi, which implements LZ4 in Rust.  I'm not sure how we should proceed here.

dfriederich:
> > @dfriederich gfxr uses pre-compiled binaries for the compression libraries. Could the problems you are seeing be mismatches/conflicts between the library gfxr used to compress and the library you are using to decompress?
> 
> It looks like GFXR uses LZ4_compress_fast and LZ4_decompress_safe. According to [lz4/lz4#1504](https://github.com/lz4/lz4/issues/1504), LZ4_compress_fast uses the legacy format (equivalent to lz4 -l1) from the CLI. With lz4 CLI version 1.10.0, I'm unable to decompress these files compressed with LZ4_compress_fast(). I get `Error 44 : Unrecognized header : file cannot be decoded`.
> 
> I don't see a way to decompress the legacy format from the CLI. We're actually using lz4-napi, which implements LZ4 in Rust. I'm not sure how we should proceed here.

GFXR is not using the LZ4 frame format, which most tools, including CLI expect, but the block format. Will try to see if we can decode it directly.
Usually in the block format, the uncompresses size is transferred separately, and I guess also in the GFXR binary case. But for the JSON setup we don't have the uncompressed size, the JSON did always lack this info (I did think it should have contained it, but it does not). Not sure if we need it to be added or if we can decompress without the size, without too much of an issue or not.

framework/decode/vulkan_replay_dump_resources_json.cpp
```diff
@@ -55,6 +56,8 @@ VulkanReplayDumpResourcesJson::VulkanReplayDumpResourcesJson(const VulkanReplayO
     dr_options["dumpResourcesDumpRawImages"]            = options.dump_resources_dump_raw_images;
     dr_options["dumpResourcesDumpSeparateAlpha"]        = options.dump_resources_dump_separate_alpha;
     dr_options["dumpResourcesDumpUnusedVertexBindings"] = options.dump_resources_dump_unused_vertex_bindings;
+    dr_options["dumpResourcesBinaryFileCompressionType"] =
```
wpaul-samsung, line 62: We it be possible to also get the uncompressed data size added here as well?

panos-lunarg: I'll look into it

panos-lunarg: Pushed a couple more commits

dfriederich:
I wonder if GFXR could/should document explicitly that it uses LZ4 Block format, and does not add the commonly used LZ4 Frame format around the blocks as well.
https://github.com/lz4/lz4/blob/dev/doc/lz4_Frame_format.md
https://github.com/lz4/lz4/blob/dev/doc/lz4_Block_format.md

tools such as the lz4 command line utilities expect the LZ4 Frame format, so being explicit to use LZ4 Block format only might help others too :-)

wpaul-samsung:
We are happy with the branch.  Can we get it reviewed?

bradgrantham-lunarg:
Feature needs description in `vulkan_dump_resources.md`

framework/decode/vulkan_replay_dump_resources_json.cpp
bradgrantham-lunarg: @wpaul-samsung please resolve this uncompressed data size was added adequately

framework/graphics/vulkan_resources_util.h
```diff
@@ -71,8 +71,7 @@ class VulkanResourcesUtil
     // all_layers_per_level boolean determines if all array layer per mip map level will be accounted as one.
     //
     // Return value is the total size of the image.
-    uint64_t GetImageResourceSizesOptimal(VkImage                image,
-                                          VkFormat               format,
+    uint64_t GetImageResourceSizesOptimal(VkFormat               format,
```
bradgrantham-lunarg, line 74: Update the comment describing the function since it doesn't take a `VkImage` anymore

bradgrantham-lunarg (approved):

framework/decode/vulkan_replay_dump_resources_json.cpp
wpaul-samsung: This is resolved from our perspective, but I don't have the ability to resolve this myself.

framework/graphics/vulkan_resources_util.h
panos-lunarg: This should be addressed now

vulkan_dump_resources.md
```diff
@@ -310,6 +310,9 @@ Dump resources feature can be control in several ways. To do so, a number of par
               When enabled all image resources will be dumped verbatim as raw bin files.
   --dump-resources-dump-separate-alpha
               When enabled alpha channel of dumped images will be dumped in a separate file.
+  --dump-resources-binary-file-compression-type
+              Compress files that are dumped as binary. Available compression types are: [none, lz4 (block format), zlib,
```
panos-lunarg, line 314: @bradgrantham-lunarg I am not sure if this can be considered adequate  documentation for the lz4 compression

dfriederich: I think the '(block format)' part helps a lot to understand that it is not lz4 frame format, fine with me :-).

Issue #2121, opened 2025-04-07
add device deduplication


zmike:
A tooling interaction which is extremely useful is gfxreconstruct + renderdoc. Specifically, creating a trace with gfxreconstruct and then grabbing renderdoc captures off the trace replay. This enables distributing a single trace which exhibits multiple issues vs distributing multiple renderdoc captures (which require far more storage space).

Unfortunately, this interaction doesn't always work. Games running through DXVK will (stupidly) create 2 VkDevices for any VkPhysicalDevice in use, which breaks renderdoc capture even though the second device is never actually used. It would be very helpful if gfxreconstruct had an option to deduplicate VkDevice creation, i.e., treat VkDevice as a singleton per VkPhysicalDevice.

bradgrantham-lunarg:
Hey, @nickdriscoll-lunarg I think this could be a good optional replay feature with a CLI option, that we maybe could keep a map of DeviceCreateInfos and created devices and use the same replay device for different capture devices that had the same createinfos.  May require some thinking about how we manage mapping of capture handleIDs to replay handles.

zmike:
I think if this hasn't been started then it can be bumped down in priority. The issue in question can be worked around by creating gfxr traces with the renderdoc capture layer loaded. It's not great, but it's not particularly difficult to set up.

jesliwang:
> I think if this hasn't been started then it can be bumped down in priority. The issue in question can be worked around by creating gfxr traces with the renderdoc capture layer loaded. It's not great, but it's not particularly difficult to set up.


I'm stuck on this problem. I can't figure out how you did it----could you describe the basic idea? @zmike

zmike:
Run the app like you are going to do a renderdoc capture and then also add the gfxr layer.

jesliwang:
> Run the app like you are going to do a renderdoc capture and then also add the gfxr layer.

What's your device? On android devices, it tends to get  stuck during loading the layer - Whether renderdoc before gfxr, or reverse.

PR #2325, opened 2025-07-31
--deduplicate-device for RenderDoc and DXVK

locke-lunarg:
Source: https://github.com/LunarG/gfxreconstruct/issues/2121
Outdated PR: https://github.com/LunarG/gfxreconstruct/pull/2306

Here is the detail of my investigation for using `RenderDoc` and `DXVK`.

1. When it uses `DXVK`, it will create two duplicate devices for one physical device. Plus, we can see `pEngineName` in `vkCreateInstance` is `DXVK`.

2. When `RenderDoc` launches `gfxrecon-replay` to replay this file, it gets these errors:
`[gfxrecon] ERROR - DEBUG MESSENGER: Frame #1: (nullptr): RenderDoc does not support multiple simultaneous logical devices.`
`[gfxrecon] FATAL - API call at index: 2020 thread: 1 vkCreateDevice returned error value VK_ERROR_INITIALIZATION_FAILED that does not match the result from the capture file: VK_SUCCESS. Replay cannot continue.`
`RenderDoc` support only one device, and the second create device is failed. The purpose of `--deduplicate-device` is to fix this issue. It will use the same device for the two duplicate devices.

3. It could get fixed in another way. Capturing the game with `RenderDoc`. The capture file could be replayed with `RenderDoc` successfully.
Add `export VK_INSTANCE_LAYERS=VK_LAYER_LUNARG_gfxreconstruct:VK_LAYER_RENDERDOC_Capture` and `steam steam://run/app-id` could capture the game with `RenderDoc`. `RenderDoc` causes the second create device failed. GFXR captures the failed result `VK_ERROR_INITIALIZATION_FAILED`. In this case, it won't happen the replay results don't match to cause stopping replaying.

5. If we change the layer order to be `export VK_INSTANCE_LAYERS=VK_LAYER_RENDERDOC_Capture:VK_LAYER_LUNARG_gfxreconstruct`, the second create device won't be captured, and `pApplicationName` and `pEngineName` in `vkCreateInstance` are `RenderDoc Capturing App` and `RenderDoc`.  It looks that `RenderDoc` could wipe out `DXVK` effect.

6. For No. 3 and No. 4 cases,  they don't have duplicate devices issue and don't need `--deduplicate-device`.

7. But for No. 4 case, `RenderDoc` layer is before `GFXR` layer, so `GFXR` captures some `RenderDoc` code, like `GetDeviceMemoryOpaqueCaptureAddress`. Some capture code from `RenderDoc` might cause unknown errors during replay.

8. If two devices are created by different instances, `RenderDoc` could run it successfully, so this case also doesn't need `--deduplicate-device`.

### DIFF
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -333,7 +333,7 @@ struct VulkanDeviceInfo : public VulkanObjectInfo<VkDevice>
     std::vector<VkPhysicalDevice> replay_device_group;
 
     // For use with device deduplication
-    bool is_duplicate{ false };
+    format::HandleId duplicate_source_id{ format::kNullHandleId };
 };
 
 struct VulkanQueueInfo : public VulkanObjectInfo<VkQueue>
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -335,7 +335,7 @@ struct VulkanDeviceInfo : public VulkanObjectInfo<VkDevice>
     // For use with device deduplication
     format::HandleId duplicate_source_id{ format::kNullHandleId };
 
-    void copy(VulkanDeviceInfo* source_info)
+    void copy_characteristics(const VulkanDeviceInfo* source_info)
     {
         parent                     = source_info->parent;
         allocator                  = source_info->allocator;
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -307,17 +307,17 @@ struct VulkanPhysicalDeviceInfo : public VulkanObjectInfo<VkPhysicalDevice>
 struct VulkanDeviceInfo : public VulkanObjectInfo<VkDevice>
 {
     VkPhysicalDevice                         parent{ VK_NULL_HANDLE };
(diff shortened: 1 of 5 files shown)


framework/decode/vulkan_object_info.h
```diff
@@ -333,7 +333,22 @@ struct VulkanDeviceInfo : public VulkanObjectInfo<VkDevice>
     std::vector<VkPhysicalDevice> replay_device_group;
 
     // For use with device deduplication
-    bool is_duplicate{ false };
+    format::HandleId duplicate_source_id{ format::kNullHandleId };
+
+    void copy(VulkanDeviceInfo* source_info)
```
bradgrantham-lunarg: This doesn't copy all the members of the struct.  Could this be called `copy_characteristics` or something more descriptive of its purpose?

Also is there a benefit for this function not being `(const VulkanDeviceInfo& source_info)`?

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_object_info.h
locke-lunarg: changed

bradgrantham-lunarg (approved):

PR #2329, opened 2025-08-01
Add Architectural Principles

bradgrantham-lunarg:
Add Architectural Principles

### DIFF
--- a/docs/GFXReconstruct Architectural Principles.md
+++ b/docs/GFXReconstruct Architectural Principles.md
@@ -76,7 +76,7 @@ It may contain more commands represented by Metadata commands but it should be p
 Capture should avoid changing the content of the application’s graphics commands.  Some exceptions are necessary in order to save information required for replay or modify data so that replay is possible.
 
 * Implications:  
-  * It *may* be acceptable for the user to *request* an annotated or altered capture.  (e.g. A user may wish to save the GPU buffers used by drawing commands at capture time.  However, the preferred technique for altering the application’s graphics command stream and saving additional information for a specialization of GFXR is to make an additional upchain or downchain layer separate from GFXR. For APIs without a formal third-party layering procedure (e.g. DirectX 12), we may implement a plugin architecture for pre- or post- function and method call alteration of commands.)  
+  * It *may* be acceptable for the user (through capture layer configuration from, as examples, environment variables, Android properties, or calls into the capture layer from another API layer or the application being captured) to *request* an annotated or altered capture.  (e.g. A user may wish to save the GPU buffers used by drawing commands at capture time.  However, the preferred technique for altering the application’s graphics command stream and saving additional information for a specialization of GFXR is to make an additional upchain or downchain layer separate from GFXR. For APIs without a formal third-party layering procedure (e.g. DirectX 12), we may implement a plugin architecture for pre- or post- function and method call alteration of commands.)
   * Combined with the above principle “A capture file should contain as much as possible of all the graphics API calls made by the application”, a corollary is that API calls modified in capture should be stored in their original form in the capture file; a MetaData Command may encode additional information to signal to the replay application to also modify the calls on replay.  
 * Benefits:  
   * GFXR limits the differences between running the application with and without the capture layer.  GFXR increases confidence that replay of the capture file will closely match the application’s execution without the capture layer.  
@@ -92,7 +92,9 @@ Capture should avoid changing the content of the application’s graphics comman
 
 **Replay without user intervention in the **same** **environment** as capture is an explicit goal. **
 
-Replay on the same GPU model as capture, on the same operating system, with the same driver revision, under the same or lower load conditions without additional user-provided options **is** an explicit goal.
+Replay on the same GPU model as capture, on the same operating system, with the same driver revision, under the same or lower load conditions without additional user-provided options **is** an explicit goal.  
+
+Replay on GPUs, operating systems, OS revisions, and driver revisions that differ from the capture environment, while a lower priority than replay on the same environment, is highly desired.  Therefore capture must not make portable replay impossible.
 
 * Implications:  
   * Include modifications to the sequence of commands without which replay *in the same environment* would be unlikely to succeed.  An example is “Virtual Swapchain”, without which swapchain image indices are likely to differ between capture and replay for all presentation modes except FIFO.  


docs/GFXReconstruct Architectural Principles.md
```diff
@@ -0,0 +1,220 @@
...
+
+**Developers should apply their best effort to adhere to the spirit of the principles**
+
+If an exception is made to a principle, there should be a well-considered, documented reason.  These principles are listed in rough order of decreasing importance so that a reasoned decision can be made if new development would necessarily conflict with one or the other of two principles.
+
+## RECONSTRUCT 
+
+**A capture file should contain as much as possible all of the unmodified graphics API calls made by the application for the captured, trimmed range of frames.**
+
+It may contain more commands represented by Metadata commands but it should be possible to**reconstruct** the original app’s sequence of API commands from the captured “api call” blocks.
```
fabian-lunarg: missing space 'to **reconstruct**'

docs/GFXReconstruct Architectural Principles.md
```diff
@@ -0,0 +1,220 @@
...
+
+* Implications:  
+  * With no options enabled, capture should record as much as possible a “vanilla” capture that does not record additional graphics API calls.  Graphics calls to retrieve additional information should be represented as metadata command blocks (e.g. existing FillMemoryCommandHeader “MetaData Command”) so that it is differentiated from the app’s calls. Graphics calls that must be modified to enable replay should be represented as or augmented with metadata command blocks (e.g. SetOpaqueAddressCommand).  
+* Benefits:  
+  * It is possible to analyze an application’s API calls using the content of the capture file without the application.   The user doesn’t have to determine what calls the capture layer modified or added.  
+* Exceptions:  
+  * Because of the nature of trimming, Graphics API calls issued before the beginning of the trim range are likely to be rearranged and many omitted.  The individual calls to create objects and change object state should remain as close to the original commands issued by the application as possible.  
+  *
```
fabian-lunarg: empty bullet

docs/GFXReconstruct Architectural Principles.md
```diff
@@ -0,0 +1,220 @@
...
+  * Capture enables the Vulkan device features bufferDeviceAddressCaptureReplay, rayTracingPipelineShaderGroupHandleCaptureReplay, and accelerationStructureCaptureReplay during capture of Vulkan applications (and in replay) to allow replay on the same device and driver of ray-tracing workloads.  
+    * Capture may alter function calls (e.g. add USAGE flags to memory creation) that enable the use of those device features  
+  * Before a trim range and, in particular, immediately before the first function within a trim range:  
+    * GFXReconstruct may make API calls to query state, read memory contents, and synchronize with the underlying implementation.  
+    * GFXReconstruct may remove some functions that pertained only to previous frame contents that are not included in the trim range.  An example is if vkQueuePresentKHR was omitted from state setup, then vkWaitForFences  
+  * Additionally, any changes in the future that may alter original application behavior should be configurable with a capture-time option. 
+
+## REPLAY 
+
+**Replay without user intervention in the**same** **environment** as capture is an explicit goal. **
```
fabian-lunarg: missing space, 'the **same**'

docs/GFXReconstruct Architectural Principles.md
```diff
@@ -0,0 +1,220 @@
...
+
+Replay on the same GPU model as capture, on the same operating system, with the same driver revision, under the same or lower load conditions without additional user-provided options**is** an explicit goal.
```
fabian-lunarg: missing space, 'options **is**'

docs/GFXReconstruct Architectural Principles.md
```diff
@@ -0,0 +1,220 @@
...
+  * Combined with the above principle “A capture file should contain as much as possible of all the graphics API calls made by the application”, a corollary is that API calls modified in capture should be stored in their original form in the capture file; a MetaData Command may encode additional information to signal to the replay application to also modify the calls on replay.  
+* Benefits:  
+  * GFXR limits the differences between running the application with and without the capture layer enabled are limited.  GFXR increases confidence that replay of the capture file will closely match the application’s execution without the capture layer.
```
bartosz-muszarski-arm: `GFXR limits the differences between running the application with and without the capture layer enabled are limited`
Perhaps should be:
`GFXR limits the differences between running the application with and without the capture layer enabled`

docs/GFXReconstruct Architectural Principles.md
```diff
@@ -0,0 +1,220 @@
...
+  - [Interface](#interface)
+  - [Compatibility](#compatibility)
+  - [Invocation](#invocation)
+  - [Convert-JSON](#convert-json)
+
+---
+
+# Purpose
+
+The purpose of GFXReconstruct is recording application graphics API calls to a file (hereafter called a “capture file”), replaying the API calls from a capture file, and processing capture files.  The design, implementation, and use of the tools should be simple. The file format should preserve the application’s calls thoroughly and accurately.  Replay should honor the application’s intent.
```
per-mathisen-arm: "file format" -> "capture"? The actual file format should be orthogonal to what is capture inside it.

"Replay should by default attempt to accurately reproduce the application's intent" might be more clear?

docs/GFXReconstruct Architectural Principles.md
```diff
@@ -0,0 +1,220 @@
...
+
+We recognize broad categories of use for GFXReconstruct that we describe below.  The definition of these “use cases” guides the definition of the principles.
+
+# Use Cases
+
+## USE-CASE-REPRODUCTION
+
+GFXReconstruct is intended primarily for reproduction of an application’s API calls for analysis, bug reporting, and testing.  This use case prioritizes the integrity and fidelity of the captured data and reproduction of the data when replayed.  While file storage and performance is valued, representing the application’s original API calls and intent by default is paramount.  This use case prioritizes using GFXR through one or more of its command-line tools or scripts and does not prioritize use of the internal classes or objects within the code itself.
```
per-mathisen-arm: I'll just note that accurate capture and capturing intent are not the same thing and sometimes are in conflict with each other, since "explicit" APIs like Vulkan are not be set up to capture intent, so you may have to tease out intent by introducing additional choices for the application, which gives you intent at the cost of accuracy of capture.

docs/GFXReconstruct Architectural Principles.md
```diff
@@ -0,0 +1,220 @@
...
+* Implications:  
+  * Include modifications to the sequence of commands without which replay *in the same environment* would be unlikely to succeed.  An example is “Virtual Swapchain”, without which swapchain image indices are likely to differ between capture and replay for all presentation modes except FIFO.  
+* Benefits:  
+  * Users are not surprised by “replay” failing for a capture just taken.  
+* Exceptions:  
+  * Android replay may require “--surface-index” parameter app to force selection of a particular surface.
+
+## WORKLOAD 
+
+**Replay should result in a GPU workload as similar to capture as possible unless requested by the user.**
```
per-mathisen-arm, line 108: Perhaps add a point about the CPU overhead of replay being as low as possible, or at least not heavier than that of the original app? While we cannot reproduce the original CPU workload of the app, we can at least try not to introduce CPU bottlenecks that weren't there in the original.

docs/GFXReconstruct Architectural Principles.md
```diff
@@ -0,0 +1,220 @@
...
+
+We accept that the capture layer will reduce system performance and increase use of system memory during capture of the application.  All components should be designed and implemented with an eye to performance, especially the capture layer.
+
+* Implications:  
+  * Avoid locks and allocations  
+  * Make large initial allocations, avoid repeated allocations  
+  * Prefer grouping gpu-transfer operations, limit added queue submissions and synchronization (try to limit stalls when initiating trimmed capture)  
+* Benefits:  
+  * Ease of use of the tools  
+* Current Exceptions
```
per-mathisen-arm, line 150: "Current limitations"?

docs/GFXReconstruct Architectural Principles.md
```diff
@@ -0,0 +1,220 @@
...
+We accept that the capture layer will reduce system performance and increase use of system memory during capture of the application.  All components should be designed and implemented with an eye to performance, especially the capture layer.
+
+* Implications:  
+  * Avoid locks and allocations  
+  * Make large initial allocations, avoid repeated allocations  
+  * Prefer grouping gpu-transfer operations, limit added queue submissions and synchronization (try to limit stalls when initiating trimmed capture)  
+* Benefits:  
+  * Ease of use of the tools  
+* Current Exceptions  
+  * We may have lots of STL maps doing small incremental heap allocations of their internal redblack tree nodes protected by locks (grep for std::map).
```
per-mathisen-arm, line 151: Drop "may"? It is a statement of fact after all ;-)

docs/GFXReconstruct Architectural Principles.md
```diff
@@ -0,0 +1,220 @@
...
+    * GFXReconstruct may make API calls to query state, read memory contents, and synchronize with the underlying implementation.  
+    * GFXReconstruct may remove some functions that pertained only to previous frame contents that are not included in the trim range.  An example is if vkQueuePresentKHR was omitted from state setup, then vkWaitForFences  
+  * Additionally, any changes in the future that may alter original application behavior should be configurable with a capture-time option. 
+
+## REPLAY 
+
+**Replay without user intervention in the**same** **environment** as capture is an explicit goal. **
+
+Replay on the same GPU model as capture, on the same operating system, with the same driver revision, under the same or lower load conditions without additional user-provided options**is** an explicit goal.
+
```
per-mathisen-arm, line 98: No mention of portable replay as a goal at all?

fabian-lunarg: good point. we should mention it here

docs/GFXReconstruct Architectural Principles.md
fabian-lunarg: I think we address this below under "PERFORMANCE" ("All components but especially capture ...").
maybe we make it more clear that we also very much aim for good replay performance :)

docs/GFXReconstruct Architectural Principles.md
fabian-lunarg: I think it would be good to maintain such a list, but 'could' become exhaustive when not structured properly, would need to be revisited in regular intervals and ideally track roadmaps and state of individual limitations. 
just here, I think it doesn't belong.
imo this should be a rather stable document, making generic, underlying principles transparent.

docs/GFXReconstruct Architectural Principles.md
bradgrantham-lunarg: How is "accurately reproduce" different from "honor"?

docs/GFXReconstruct Architectural Principles.md
bradgrantham-lunarg: Interesting - I meant this section to be about the quantities and form of work requested from the GPU, not about performance exactly.  So I'll change WORKLOAD to "GPU WORKLOAD"

docs/GFXReconstruct Architectural Principles.md
bradgrantham-lunarg: I see what you're saying, but I meant here that the items under "Exceptions" are code that has snuck in that impacts performance and I imagine can still be refactored/redesigned to be performant.

docs/GFXReconstruct Architectural Principles.md
bradgrantham-lunarg: Yes.

docs/GFXReconstruct Architectural Principles.md
bradgrantham-lunarg: This document is meant to make strong statements against which we can measure a change to the code.

While LunarG and contributors to the project do put a **lot** of effort into attempting to replay on implementations that differ from the capture-time implementation, I think I would stop short of saying it is an explicit **architectural principle**.

I think portable replay is more like "we want it if it's possible practically within the engineering resource available".  Maybe it's worth saying something like "while portable replay is not a strong design principle, it is highly desired, therefore capture must not make portable replay impossible" or some such.  I'll think about how to rephrase that.

docs/GFXReconstruct Architectural Principles.md
```diff
@@ -0,0 +1,219 @@
...
+  * Combined with the above principle “A capture file should contain as much as possible of all the graphics API calls made by the application”, a corollary is that API calls modified in capture should be stored in their original form in the capture file; a MetaData Command may encode additional information to signal to the replay application to also modify the calls on replay.  
+* Benefits:  
+  * GFXR limits the differences between running the application with and without the capture layer.  GFXR increases confidence that replay of the capture file will closely match the application’s execution without the capture layer.  
+* Exceptions:  
+  * Capture enables the Vulkan device features bufferDeviceAddressCaptureReplay, rayTracingPipelineShaderGroupHandleCaptureReplay, and accelerationStructureCaptureReplay during capture of Vulkan applications (and in replay) to allow replay on the same device and driver of ray-tracing workloads.
```
per-mathisen-arm, line 84: Perhaps add an explicit exception here of things added because of explicit communication/agreement between the app and the capture layer? The capture layer can offer additional functionality where the use of it assumes that the application (to some extent) is aware that it is debugged or captured. This includes things like debug annotations, explicit host updates, explicit frame boundaries, etc..

PR #2330, opened 2025-08-02
VDR: Delegate refactoring

panos-lunarg:


framework/decode/vulkan_replay_dump_resources_common.cpp
```diff
@@ -557,18 +563,15 @@ VkResult DumpImageToFile(const VulkanImageInfo*               image_info,
...
                     }
                     else
                     {
                         scaled_extent = image_info->extent;
                     }
 
-                    scaled_extent.width  = std::max(1u, scaled_extent.width >> mip);
-                    scaled_extent.height = std::max(1u, scaled_extent.height >> mip);
-                    scaled_extent.depth  = std::max(1u, scaled_extent.depth >> mip);
+                    scaled_extent                    = ScaleToMipLevel(scaled_extent, mip);
```
bradgrantham-lunarg: Does this line immediately replace the value of `scaled_extent` set in the previous `if`-`else` statement?

framework/decode/vulkan_replay_dump_resources_common.cpp
```diff
@@ -627,8 +630,22 @@ VkResult DumpImageToFile(const VulkanImageInfo*               image_info,
...
 
-                    util::bufferwriter::WriteBuffer(
+                    const size_t bytes_written = util::bufferwriter::WriteBuffer(
                         filename, offsetted_data, subresource_sizes[sub_res_idx], compressor);
+
+                    dumped_subresource.size = subresource_sizes[sub_res_idx];
+
+                    if (compressor != nullptr && bytes_written != subresource_sizes[sub_res_idx])
+                    {
+                        if (!before_command)
```
bradgrantham-lunarg: Are `before_{thing}` or `{thing}` ever used at the same time?  It looks like to me in a quick scan it's only ever one or the other, and it's selected by `before_command` in this function and in other places.  I think that makes it a place where errors are more likely.  Could filename, size, and compressed_size be put in a struct and selected, or maybe even just have just the one variant without `before_` in `Dumped*` and put `before_command` in `Dumped*`?

framework/decode/vulkan_replay_dump_resources_common.h
```diff
@@ -63,6 +65,269 @@ struct MinMaxVertexIndex
 
 using BoundDescriptorSets = std::unordered_map<uint32_t, VulkanDescriptorSetInfo::VulkanDescriptorBindingsInfo>;
 
+enum class DumpResourcesCommandType
+{
+    kUnknown,
```
bradgrantham-lunarg: This means the object doesn't yet contain information about a command, right?  This and the other `kUnknown` below seem more to me like `kNone`.

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
```diff
@@ -1232,130 +1212,130 @@ VkResult DispatchTraceRaysDumpingContext::DumpMutableResources(uint64_t bcb_inde
...
-        res_info.set                    = mutable_resources_clones.images[i].desc_set;
-        res_info.binding                = mutable_resources_clones.images[i].desc_binding;
-        res_info.array_index            = mutable_resources_clones.images[i].array_index;
-        res_info.stages                 = mutable_resources_clones.images[i].stages;
-        auto res                        = delegate_.DumpResource(res_info);
-        if (res != VK_SUCCESS)
-        {
-            return res;
-        }
+        // Stop pointing to the stack allocated VulkanImageInfo
```
bradgrantham-lunarg: I think it would be good to avoid any situation in which we need to be careful about a stack-allocated object since if the author of a new function forgets to do the following two lines we will likely read incorrect memory later.

framework/decode/vulkan_replay_dump_resources_delegate.cpp
```diff
@@ -64,89 +64,229 @@ bool DefaultVulkanDumpResourcesDelegate::IsImageDumpable(const graphics::VulkanI
...
+            dumped_buffer = std::get_if<DumpedBuffer>(&dumped_buffer_desc->dumped_resource);
+            GFXRECON_ASSERT(dumped_buffer != nullptr);
+        }
+        break;
+
+        default:
+            GFXRECON_LOG_ERROR("%s(): Unexpected resource type", __func__)
+    }
+
+    dumped_buffer->filename = std::invoke(filename_generator, *this, *dumped_resource, false);
```
bradgrantham-lunarg: What is the value of using invoke here and not just calling a function in the `if-else` chain above?

framework/decode/vulkan_replay_dump_resources_draw_calls.h
```diff
@@ -402,6 +394,77 @@ class DrawCallsDumpingContext
         }
     }
 
+    static const format::ApiCallId DrawCallTypeApiCallId(DrawCallType type)
```
bradgrantham-lunarg: Were these two functions added for debugging and should be removed?  They're not used anywhere as far as I can tell.

framework/decode/vulkan_replay_dump_resources_common.h
```diff
@@ -63,6 +65,269 @@ struct MinMaxVertexIndex
...
+
+struct DumpedBuffer : DumpedFile
+{
+    DumpedBuffer() = default;
+
+    DumpedBuffer(const VulkanBufferInfo* bi, VkDeviceSize o, size_t s) : DumpedFile(s), buffer_info(bi), offset(o) {}
+
+    DumpedBuffer(const std::vector<uint8_t> data) : DumpedFile(data.size()), inline_uniform_data(data) {}
+
+    DumpedBuffer(const DumpedBuffer& other)
```
bradgrantham-lunarg: Should this also copy `inline_uniform_data`?

framework/decode/vulkan_replay_dump_resources_common.cpp
panos-lunarg: `before_command` specifies whether we are dumping resources before the execution of the dumped command or after. When we are dumping "before" (`--dump-resources-before-draw` cl option is set) we are also dumping "after" the execution of the command. So in this case `DumpImageToFile` and `DumpBufferToFile` functions will be called once when dumping the resources before (`before_command` == `true`) and once after (`before_command` == `false`).
If we are doing a "normal" dump, dumping only after the execution of the commands, then the `before_command` will always be false.

framework/decode/vulkan_replay_dump_resources_common.cpp
panos-lunarg: Yes. Will change it

framework/decode/vulkan_replay_dump_resources_common.h
panos-lunarg: Sounds good

framework/decode/vulkan_replay_dump_resources_draw_calls.h
panos-lunarg: good catch

framework/decode/vulkan_replay_dump_resources_delegate.cpp
panos-lunarg: In this function just fewer lines of code.
In `DumpImage`, where each filename generator is called for each image subresource, I think it benefits more as it simplifies the code.

framework/decode/vulkan_replay_dump_resources_common.h
panos-lunarg: Good catch

framework/decode/vulkan_replay_dump_resources_draw_calls.h
bradgrantham-lunarg: DrawCallTypeToApiCallId too

framework/decode/vulkan_replay_dump_resources_common.cpp
bradgrantham-lunarg: Sent an email for a subsequent revision about this, resolving for this PR.

framework/util/buffer_writer.cpp
```diff
@@ -45,30 +45,35 @@ bool WriteBuffer(const std::string& filename, const void* data, size_t size, con
...
+    size_t bytes_written;
     if (compressor != nullptr)
     {
         std::vector<uint8_t> compressed_data;
         size_t compressed_size = compressor->Compress(size, static_cast<const uint8_t*>(data), &compressed_data, 0);
-        if (compressed_size > 0)
+
+        // Regardless of the compression method, if compressed output is larger than the input then use the uncompressed
+        // data
+        if (compressed_size && compressed_size < size)
```
dfriederich: I do think that if we do compress, then we should still be doing it even if the result is slightly bigger. For zlib, we can uncompress without knowing the size, with this we cannot.

framework/util/buffer_writer.cpp
```diff
@@ -45,30 +45,35 @@ bool WriteBuffer(const std::string& filename, const void* data, size_t size, con
...
-    bool success;
+    bool   success;
+    size_t bytes_written;
     if (compressor != nullptr)
     {
         std::vector<uint8_t> compressed_data;
         size_t compressed_size = compressor->Compress(size, static_cast<const uint8_t*>(data), &compressed_data, 0);
-        if (compressed_size > 0)
+
+        // Regardless of the compression method, if compressed output is larger than the input then use the uncompressed
```
dfriederich: Off by 1 in comment bug :-).
If the sizes are equal, this uses uncompressed too. It has to, as otherwise the equal size case would be ambiguous.

framework/util/buffer_writer.cpp
```diff
@@ -29,7 +29,7 @@ GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(util)
 GFXRECON_BEGIN_NAMESPACE(bufferwriter)
 
-bool WriteBuffer(const std::string& filename, const void* data, size_t size, const Compressor* compressor)
+size_t WriteBuffer(const std::string& filename, const void* data, size_t size, const Compressor* compressor)
 {
     assert(data);
     assert(size);
```
dfriederich, line 35: Just as note, I'm a bit surprised that a 0 sized buffer is not legal to be written. Not wrong, just unexpected at least for me :-)

framework/util/buffer_writer.cpp
panos-lunarg: I was planning on changing the json output a bit by adding a new entry.
`size` will list the uncompressed size and `compressedSize` will list the compressed size. If there's no compression then there won't be a `compressedSize`.
You have a point that it might cause confusion if not compressing all the time, I'll switch back to always using the compressed output, regardless its size, while keeping the json entries.

framework/util/buffer_writer.cpp
panos-lunarg: I think since [the spec does not allow creating 0 sized buffers](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#VUID-VkBufferCreateInfo-size-00912), a zero size here would be suspicious

framework/decode/vulkan_replay_dump_resources_common.h
```diff
@@ -63,15 +67,336 @@ struct MinMaxVertexIndex
...
+
+    DumpedBuffer(const VulkanBufferInfo* bi, VkDeviceSize o, VkDeviceSize s) : DumpedFile(s), buffer_info(bi), offset(o)
+    {}
+
+    const VulkanBufferInfo* buffer_info{ nullptr };
+    VkDeviceSize            offset{ VK_WHOLE_SIZE };
+
+    void CopyDumpedInfo(const DumpedBuffer& other)
+    {
+        filename        = other.filename;
```
MarkY-LunarG, line 121: Don't you need to copy over the offset as well?

framework/decode/vulkan_replay_dump_resources_common.h
```diff
@@ -63,15 +67,336 @@ struct MinMaxVertexIndex
...
+
+struct DumpedImage
+{
+    DumpedImage() = default;
+
+    DumpedImage(const VulkanImageInfo* i_f) : image_info(i_f), scaling_failed(false), dumped_raw(false) {}
+
+    const VulkanImageInfo* image_info{ nullptr };
+
+    bool     scaling_failed{ false };
```
MarkY-LunarG, line 138: Can you add a comment what you mean by "scaling_failed" here?

framework/decode/vulkan_replay_dump_resources_common.h
```diff
@@ -63,15 +67,336 @@ struct MinMaxVertexIndex
...
+        dumped_format       = other.dumped_format;
+        dumped_subresources = other.dumped_subresources;
+    }
+};
+
+struct DumpedResourceBase
+{
+    DumpedResourceBase() = default;
+
+    DumpedResourceBase(DumpResourceType t, uint64_t bcb, uint64_t cmd, uint64_t qs) :
```
MarkY-LunarG, line 174: What is BCB?  I think a description of the items here would be helpful.  I'm assuming it is BeginCmdBuffer Index?

framework/decode/vulkan_replay_dump_resources_common.h
```diff
@@ -63,15 +67,336 @@ struct MinMaxVertexIndex
...
+        DumpedResourceBase(t, bcb, cmd, qs, rp, sp),
+        location(l), dumped_image(img_info)
+    {
+        if (before)
+        {
+            dumped_image_before = DumpedImage(img_info);
+        }
+    }
+
+    DumpedImage dumped_image;
```
MarkY-LunarG, line 399: I'm not sure what the difference is between the `dumped_image` and `dumped_image_before`, can you clarify?

framework/decode/vulkan_replay_dump_resources_common.h
```diff
@@ -63,15 +67,336 @@ struct MinMaxVertexIndex
...
+    DumpedImage dumped_image_before;
+
+    uint32_t location{ 0 };
```
MarkY-LunarG, line 406: What is the location?  Is this render target index?

framework/decode/vulkan_replay_dump_resources_common.cpp
```diff
@@ -525,111 +555,26 @@ VkResult DumpImageToFile(const VulkanImageInfo*               image_info,
...
             for (uint32_t layer = 0; layer < image_info->layer_count; ++layer)
             {
-                const std::string& filename = filenames[f++];
+                const VkExtent3D subresource_extent        = ScaleToMipLevel(image_info->extent, mip);
+                const VkExtent3D subresource_scaled_extent = ScaleToMipLevel(scaled_extent, mip);
 
-                // We don't support stencil output yet
-                if (aspects[i] == VK_IMAGE_ASPECT_STENCIL_BIT)
-                    continue;
+                dumped_image.dumped_subresources.emplace_back(
```
MarkY-LunarG, line 565: Don't we need more info if we're going to determine what this sub-resource is, like command buffer index etc?  Or is this tracking not meant to identify images at specific commands/locations?

framework/decode/vulkan_replay_dump_resources_common.cpp
```diff
@@ -642,9 +587,60 @@ VkResult DumpImageToFile(const VulkanImageInfo*               image_info,
...
+    GFXRECON_ASSERT(buffer_info != nullptr);
+
+    GFXRECON_ASSERT(dumped_buffer.size);
+    GFXRECON_ASSERT(dumped_buffer.size != VK_WHOLE_SIZE);
+    GFXRECON_ASSERT(dumped_buffer.offset != VK_WHOLE_SIZE);
+
+    const uint32_t transfer_queue_index = FindTransferQueueFamilyIndex(device_info->enabled_queue_family_flags);
+    if (transfer_queue_index == VK_QUEUE_FAMILY_IGNORED)
+    {
+        GFXRECON_LOG_ERROR("Failed to find a transfer queue")
```
MarkY-LunarG, line 632: Should we add a TODO (and maybe an issue) to implement this in the future if a transfer queue isn't available?  Or is that a scenario we will never care about?

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
```diff
@@ -238,7 +236,7 @@ bool DispatchTraceRaysDumpingContext::MustDumpDispatch(uint64_t index) const
         return false;
     }
 
-    for (size_t i = dump_resources_before_ ? current_dispatch_index_ / 2 : current_dispatch_index_;
+    for (size_t i = options_.dump_resources_before ? current_dispatch_index_ / 2 : current_dispatch_index_;
```
MarkY-LunarG, line 239: Why are these " / 2" if the _before option is present?

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
```diff
@@ -1246,130 +1225,173 @@ VkResult DispatchTraceRaysDumpingContext::DumpMutableResources(uint64_t bcb_inde
...
-            res_info.binding     = mutable_resources_clones_before.buffers[i].desc_binding;
-            res_info.array_index = mutable_resources_clones_before.buffers[i].array_index;
-            res_info.stages      = mutable_resources_clones_before.buffers[i].stages;
-            res                  = delegate_.DumpResource(res_info);
+            return res;
+        }
+
+        delegate_.DumpResource(res_info);
+
+        // Dump the "before" resource
```
MarkY-LunarG, line 1279: Why dump the "before" items after the "after" items? Maybe a comment here would be helpful.

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
```diff
@@ -1246,130 +1225,173 @@ VkResult DispatchTraceRaysDumpingContext::DumpMutableResources(uint64_t bcb_inde
...
+                            options_.dump_resources_dump_raw_images,
+                            options_.dump_resources_dump_all_image_subresources,
+                            dumped_image_data.data,
+                            device_info,
+                            device_table_,
+                            instance_table_,
+                            object_info_table_);
             if (res != VK_SUCCESS)
             {
+                GFXRECON_LOG_ERROR("Reading from image descriptor %" PRIu64 " failed (%s)",
```
MarkY-LunarG, line 1302: In all the log messages in this file, should you identify these as compute/ray-tracing workload items?

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
```diff
@@ -1421,17 +1447,72 @@ VkResult DispatchTraceRaysDumpingContext::DumpDescriptors(uint64_t qs_index,
...
+                            if (img_info == nullptr)
+                            {
+                                continue;
+                            }
+
+                            if (!IsImageDumpable(instance_table_, object_info_table_, img_info))
                             {
-                                image_descriptors.insert(img_info);
-                                dumped_descriptors.image_descriptors.insert(img_info);
+                                continue;
```
MarkY-LunarG, line 1452: Maybe an info or debug message here?

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
```diff
@@ -1421,17 +1447,72 @@ VkResult DispatchTraceRaysDumpingContext::DumpDescriptors(uint64_t qs_index,
...
+                                                         device_table_,
+                                                         instance_table_,
+                                                         object_info_table_);
+                                if (res != VK_SUCCESS)
+                                {
+                                    GFXRECON_LOG_ERROR("Reading from image descriptor %" PRIu64 " failed (%s)",
+                                                       img_info->capture_id,
+                                                       util::ToString(res).c_str());
+
+                                    dumped_resources.dumped_descriptors.pop_back();
```
MarkY-LunarG, line 1493: Is this right?  Are we sure it's always the last at this point?

framework/decode/vulkan_replay_dump_resources_delegate.cpp
```diff
@@ -28,345 +28,494 @@
...
+        case VK_FORMAT_B8G8R8_UNORM:
+            return util::imagewriter::DataFormats::kFormat_BGR;
 
-    if (output_image_format != KFormatRaw)
+        case VK_FORMAT_B8G8R8A8_SRGB:
+        case VK_FORMAT_B8G8R8A8_UNORM:
+            return util::imagewriter::DataFormats::kFormat_BGRA;
+
+        case VK_FORMAT_D32_SFLOAT:
+        case VK_FORMAT_D32_SFLOAT_S8_UINT:
```
MarkY-LunarG, line 203: Are the stencil format items correct here?  Won't a D32 S8 for example have 40-bits per pixel?

MarkY-LunarG (commented):
In general things look good, lots of questions and some comments  would be greatly helpfu.  Additionally, I think some where describing how you identify common resources would be useful because it wasn't obvious to me.

framework/decode/vulkan_replay_dump_resources_common.h
fabian-lunarg: `scaling_failed` indicates that scaling was requested (scale != 1.f) but not supported.

it's set like this in .cpp:
`dumped_image.scaling_failed = (scale != 1.0f && !scaling_supported);`

we support scaling via `vkCmdBlit` and that 'can' be unsupported for certain formats (e.g. like depth/stencil)

framework/decode/vulkan_replay_dump_resources_common.h
panos-lunarg: Yes, good catch

framework/decode/vulkan_replay_dump_resources_common.h
panos-lunarg: Yes

framework/decode/vulkan_replay_dump_resources_common.h
panos-lunarg: Yes. Added a comment

framework/decode/vulkan_replay_dump_resources_common.h
panos-lunarg: `dumped_image_before` is the dumped resource when we are also dumping the resources before the execution of the command (draw, dispatch, etc).
Added a comment

framework/decode/vulkan_replay_dump_resources_common.h
panos-lunarg: Yes. Added a comment

framework/decode/vulkan_replay_dump_resources_common.cpp
panos-lunarg: The sub resources belong to `DumpedImage` (`DumpedImage::dumped_subresources`).
The resources we dump currently that are images are `DumpedDescriptor` and `DumpedRenderTarget` which both inherit `DumpedResourceBase` which hold the information that you describe.

framework/decode/vulkan_replay_dump_resources_common.cpp
panos-lunarg: I think this scenario is fairly unlikely since transfer operations can be performed by queues that have `VK_QUEUE_TRANSFER_BIT` set. In the absence of such a queue we search for queues that have `VK_QUEUE_GRAPHICS_BIT | VK_QUEUE_COMPUTE_BIT` bits set, which according to spec, can also perform transfer operations. So in order for us to return `VK_QUEUE_FAMILY_IGNORED` the implementation must be lacking a queue with any of these 3 flags

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
panos-lunarg: There is a counter for each dumped command (Dispatch in this case). When we are also dumping the resources before the execution of the command this counter is incremented twice per command. So in some cases, for example in this function, we need to divide by to find which command we are right now.

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
panos-lunarg: The order is not important. The dumped resource is constructed, in this particular case at line 1240, the "after" resource is dumped and then the "before". It could have been the other way around

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
panos-lunarg: Don't think it will add some valuable information. Also these warnings/errors come from functions that are shared between ray tracing/dispatch code and are agnostic from which one they are called

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
panos-lunarg: Yes

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
panos-lunarg: `dumped_resources.dumped_descriptors.emplace_back()` is called in all cases so it should be the last, right?

framework/decode/vulkan_replay_dump_resources_delegate.cpp
panos-lunarg: We don't support stencil atm, so if we dump only the depth aspect it should be correct

bradgrantham-lunarg (approved):

PR #2332, opened 2025-08-04
Fix for trimming AccelerationStructures

fabian-lunarg:
Simplify acceleration-structure (AS) tracking and trimming by **not** differentiating between build-modes: 'build' and 'update', just treat everything as 'build'.
 
when replaying a trimmed capture, during state-init, we rebuild acceleration-structures from scratch.
`VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR` would rely on a pre-existing AS, which does not apply here.

associated scratch-buffers for an 'update' are normally smaller compared to 'build', so they could not be used when rebuilding from scratch.
for this reason we relax a condition and allow VulkanAddressReplacer to substitute those missing (or not large enough) buffers, also without using `-m rebind`.

**Fixed in this PR**:

capture:
`[gfxrecon] WARNING - vulkan_wrappers::GetWrappedId() couldn't find Handle: 0x46d6c0e27d0's wrapper. It might have been destroyed`

replay:
```
[gfxrecon] ERROR - DEBUG MESSENGER: Frame #1: VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03667: vkCmdBuildAccelerationStructuresKHR(): pInfos[0].mode is VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR, srcAccelerationStructure must have been previously built.
The Vulkan spec states: For each element of pInfos, if its mode member is VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR, its srcAccelerationStructure member must have previously been constructed with VK_BUILD_ACCELERATION_STRUCTURE_ALLOW_UPDATE_BIT_KHR

[gfxrecon] ERROR - DEBUG MESSENGER: Frame #1: VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03674: vkCmdBuildAccelerationStructuresKHR(): pInfos[0].scratchData.deviceAddress (0xe00a34b000) has no buffer(s) associated that are valid.
The following buffers are missing VK_BUFFER_USAGE_2_STORAGE_BUFFER_BIT usage flag: ...

and/or consecutive

[gfxrecon] ERROR - DEBUG MESSENGER: Frame #1: VUID-vkResetFences-pFences-01123: vkResetFences(): pFences[0] (VkFence 0xa4f0000000a4f) is in use.
The Vulkan spec states: Each element of pFences must not be currently associated with any queue command that has not yet completed execution on that queue (https://vulkan.lunarg.com/doc/view/1.4.321.1/linux/antora/spec/latest/chapters/synchronization.html#VUID-vkResetFences-pFences-01123)
[gfxrecon] ERROR - DEBUG MESSENGER: Frame #1: VUID-vkBeginCommandBuffer-commandBuffer-00049: vkBeginCommandBuffer(): on active VkCommandBuffer 0x653113198190 before it has completed. You must check command buffer fence before this call.
The Vulkan spec states: commandBuffer must not be in the recording or pending state (https://vulkan.lunarg.com/doc/view/1.4.321.1/linux/antora/spec/latest/chapters/cmdbuffers.html#VUID-vkBeginCommandBuffer-commandBuffer-00049)
```

in effect, this PR enabled me to successfully capture/replay scenarios with many animated AS (general using `VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR`) with trimming, that would have failed before.

example scene used for testing AS-trimming, captured on steamdeck (combined w. test portability/address-sanitizer):
screenshot_frame_100

### DIFF
--- a/framework/encode/vulkan_state_tracker.cpp
+++ b/framework/encode/vulkan_state_tracker.cpp
@@ -513,21 +513,14 @@ void VulkanStateTracker::TrackAccelerationStructureBuildCommand(
         }
 
         // track all AS builds as regular builds, we'll have no AS to 'update'
-//        if (build_info.mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_BUILD_KHR)
+        if (wrapper->latest_build_command_)
         {
-            if (wrapper->latest_build_command_)
-            {
-                // TODO: keep original scratch-buffer, check if that makes sense
-                dst_command.geometry_info.scratchData = wrapper->latest_build_command_->geometry_info.scratchData;
-            }
-            dst_command.geometry_info.mode = VK_BUILD_ACCELERATION_STRUCTURE_MODE_BUILD_KHR;
-            dst_command.geometry_info.srcAccelerationStructure = VK_NULL_HANDLE;
-            wrapper->latest_build_command_ = std::move(dst_command);
+            // TODO: keep original scratch-buffer, check if that makes sense
+            dst_command.geometry_info.scratchData = wrapper->latest_build_command_->geometry_info.scratchData;
         }
-//        else if (build_info.mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR)
-//        {
-//            wrapper->latest_update_command_ = std::move(dst_command);
-//        }
+        dst_command.geometry_info.mode                     = VK_BUILD_ACCELERATION_STRUCTURE_MODE_BUILD_KHR;
+        dst_command.geometry_info.srcAccelerationStructure = VK_NULL_HANDLE;
+        wrapper->latest_build_command_                     = std::move(dst_command);
 
         wrapper->blas.clear();
 
--- a/framework/encode/vulkan_state_tracker.cpp
+++ b/framework/encode/vulkan_state_tracker.cpp
@@ -512,14 +512,22 @@ void VulkanStateTracker::TrackAccelerationStructureBuildCommand(
                                                  pp_buildRange_infos[i] + build_info.geometryCount);
         }
 
-        if (build_info.mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_BUILD_KHR)
+        // track all AS builds as regular builds, we'll have no AS to 'update'
+//        if (build_info.mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_BUILD_KHR)
         {
+            if (wrapper->latest_build_command_)
+            {
+                // TODO: keep original scratch-buffer, check if that makes sense
+                dst_command.geometry_info.scratchData = wrapper->latest_build_command_->geometry_info.scratchData;
+            }
+            dst_command.geometry_info.mode = VK_BUILD_ACCELERATION_STRUCTURE_MODE_BUILD_KHR;
+            dst_command.geometry_info.srcAccelerationStructure = VK_NULL_HANDLE;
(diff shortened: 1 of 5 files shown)


framework/encode/vulkan_state_tracker.cpp
```diff
@@ -512,14 +512,15 @@ void VulkanStateTracker::TrackAccelerationStructureBuildCommand(
...
-            wrapper->latest_build_command_ = std::move(dst_command);
-        }
-        else if (build_info.mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR)
-        {
-            wrapper->latest_update_command_ = std::move(dst_command);
+            // TODO: keep original scratch-buffer, check if that makes sense
+            dst_command.geometry_info.scratchData = wrapper->latest_build_command_->geometry_info.scratchData;
         }
+        dst_command.geometry_info.mode                     = VK_BUILD_ACCELERATION_STRUCTURE_MODE_BUILD_KHR;
+        dst_command.geometry_info.srcAccelerationStructure = VK_NULL_HANDLE;
```
antonio-lunarg, line 517: Would it be possible to keep the update command once we store the first build command in the capture file?

fabian-lunarg: I think that was the intention, but we prob. never fully implemented that. this logic is dedicated for restoring trim-states, so we mostly care about the final outcome. we wouldn't gain anything by doing a build, immediately followed by an update. also the 'update' most likely was not done with the originally built AS as `srcAccelerationStructure` but with a previously updated AS. so storing the first 'build' and last 'update' is not really correct or what happened (and you store multiple vertex/index-buffers, build 2x etc.)

antonio-lunarg: Right, how about keeping the other update commands once the first update has been converted and stored as a build command?

fabian-lunarg: in effect, I think that's already the case here. e.g. imagine during capture, the **application** builds acceleration-structure A and update it several times (not sure if that's typical usage, most code I've seen would create new handles each time):

`vkCmdBuildAccelerationStructuresKHR` (**vertex-buffer 1** - w. `VK_BUILD_ACCELERATION_STRUCTURE_MODE_BUILD_KHR`)
`vkCmdBuildAccelerationStructuresKHR` (**vertex-buffer 2** w. `VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR`)
...
`vkCmdBuildAccelerationStructuresKHR` (**vertex-buffer N** w. `VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR`)

for the **trim-state**, we end up storing only one build, using the latest supplied vertex/index-buffer.
-> `vkCmdBuildAccelerationStructuresKHR` (**vertex-buffer N** w. `VK_BUILD_ACCELERATION_STRUCTURE_MODE_BUILD_KHR`)

we do not store all build- or update commands, only the latest, so we can reconstruct the end-result.
each build/update command comes with associated buffers we need to track and store in the capture-file.

so, why keep the other (intermediate?) update commands, what would we gain?

antonio-lunarg: I see now! Thanks for the clarification :)

bradgrantham-lunarg (approved):

PR #2336, opened 2025-08-05
VulkanAddressReplacer: Align dispatch gpu read/write-access to 8byte

fabian-lunarg:
applies only when using `-m rebind`.

changes to internal compute-shader spirv used by VulkanAddressReplacer.

Fix issues when addresses stored in buffers are not aligned to 8-byte.
Check (and correct) alignment, optionally read with a padding-helper.

e.g. a buffer structured like this caused issues:

```
layout(buffer_reference, std430) buffer VertexBufferPtr{ packed_vertex_t v[]; };
layout(buffer_reference, std430) buffer IndexBufferPtr{ uint v[]; };

struct params_struct_t
{
    uint num_indices;
    VertexBufferPtr vertex_buffer; -> 4byte aligned
    IndexBufferPtr index_buffer; -> 4byte aligned
}
```

relevant part from spec, I think is here https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_physical_storage_buffer.asciidoc#326-memory-access

-> "Valid values are defined by the execution environment"
which seems to be `8` for me (on nvidia 575.57.08)


framework/decode/vulkan_address_replacer_shaders.h
```diff
@@ -292,26 +291,50 @@ layout(push_constant, std430) uniform pc
     replacer_params_bda_t params;
 };
 
+layout(buffer_reference, std430) buffer BufferDeviceAddressPtr{ uint64_t buffer_device_address; };
+layout(buffer_reference, std430) buffer PaddedBufferDeviceAddressPtr{ uint pad; uint64_t buffer_device_address; };
+
+//! helper to read a VkDeviceAddress from address 'ptr', read-access is 8-byte aligned
+uint64_t read_address(uint64_t ptr)
+{
+    return (ptr & 7UL) == 4 ? PaddedBufferDeviceAddressPtr(ptr & ~7).buffer_device_address :
```
antonio-lunarg: Can this be `!= 0` like the one below?

fabian-lunarg: yes, let's consolidate. actually this value 'could' also be 1 or 2 (when using e.g. uint8_t or uint_16_t types).
which would require additional padding helper-types and more logic.
I was thinking we skip the 1 and 2-byte cases, at least for the moment (they should be very rare I assume)

fabian-lunarg: done

PR #2337, opened 2025-08-05
VDR: Filter semaphores on same submission

panos-lunarg:
When dumping both draw calls and dispatch/trace rays from the same queue submit draw calls are dumped first. If wait semaphores are used in the submission then the first submission (draw calls) will sucessfully wait for the semaphore while the second submission (dispatch) will hang.


dfriederich (approved):
Fixed the reported issues for me. Don't think we need it in for this release as we did disable the thumbnail retrieval which did show it in the first place. But its a bug, we may enable thumbnails for compute nodes in the future, and it is possible to happen for the 1 scaled requests as well :-)

MarkY-LunarG (approved):
Looks good.

PR #2348, opened 2025-08-07
Add maximum enum-value for format::MetaDataType

fabian-lunarg:
reserving high-bit for 'special purposes' (like testing, experimentation, specializations, etc.)

- the intention is to provide a vendor-neutral way to have non-conflicting custom IDs

### DIFF
--- a/framework/decode/file_processor.cpp
+++ b/framework/decode/file_processor.cpp
@@ -2257,7 +2257,7 @@ bool FileProcessor::ProcessMetaData(const format::BlockHeader& block_header, for
     }
     else
     {
-        if (meta_data_type >= format::MetaDataType::ADDRESS_RANGE_END ||
+        if (meta_data_type >= format::MetaDataType::kBeginExperimentalReservedRange ||
             meta_data_type == format::MetaDataType::kReserved23 || meta_data_type == format::MetaDataType::kReserved25)
         {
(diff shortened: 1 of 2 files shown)


MarkY-LunarG (approved):

bradgrantham-lunarg:
How will this solve the problem of vendor IDs?  Won't they still have conflicts but just within the reserved space?  Could we have something like ranges reserved explicitly through PRs?

fabian-lunarg:
> How will this solve the problem of vendor IDs? Won't they still have conflicts but just within the reserved space? Could we have something like ranges reserved explicitly through PRs?

didn't think so far, as to partition this space, etc. - would rather treat it as completely opaque to gfxr.
my stance here was this will guarantee upstream-gfxr won't ever collide with the reserved space. 
this allows 'anyone' to add custom metadata-IDs, create capture-files using those, and not risk collisions with upstream.

for any addition coming from vendors, when functionality is added back to gfxr,
we are back at reserving individual metadata IDs (in the monotonic increasing enum-list), it's the cleanest way.

what we gain is basically: 
we can offer stability for custom IDs without knowing about them, while reducing maintenance and unnecessary gatekeeping.

bradgrantham-lunarg (approved):
Maybe change the name from `ADDRESS_RANGE_END` to something matching the others like `kBeginExperimentalReservedRange` .

PR #2352, opened 2025-08-07
Add unfinished meta json functions

locke-lunarg:
It added them to json, like
```
{
  "index": 259055,
  "meta": {
    "name": "BuildVulkanAccelerationStructuresMetaCommand (Unfinished)",
    "args": null
  }
},
```

I tried to change functions in `MetadataConsumerBase` to pure virtual functions, so we knew what they were missed in `MetadataJsonConsumer`. But it needed to change too much.


locke-lunarg:
It made a lot of mismatch for .json in CI. This change seems to be not worth, so closed it.

PR #2355, opened 2025-08-08
Fix `CommonCaptureManager::IsCaptureModeDisabled()`

marius-pelegrin-arm:
The current implementation of `CommonCaptureManager::IsCaptureModeDisabled()` cannot work because `kModeDisabled` is 0. This causes the function to always return `true` no matter the value of `GetCaptureMode()`.

This causes capture to exit instantly, before any frame is captured, when `GFXRECON_QUIT_AFTER_CAPTURE_FRAMES` is set.


fabian-lunarg (approved):
thank you!

bradgrantham-lunarg (approved):

PR #2359, opened 2025-08-12
VulkanAddressReplacer: Support for inline-uniform-blocks

fabian-lunarg:
relevant for `-m rebind`:

- add missing support for inline-uniform-blocks to `VulkanAddressReplacer::ProcessCmdBindDescriptorSets`
- fix a range-end calculation when replacing in runtime descriptor-arrays

### DIFF
--- a/framework/decode/vulkan_address_replacer.cpp
+++ b/framework/decode/vulkan_address_replacer.cpp
@@ -481,7 +481,7 @@ void VulkanAddressReplacer::ProcessCmdPushConstants(const VulkanCommandBufferInf
     GFXRECON_ASSERT(command_buffer_info != nullptr && data != nullptr);
     for (const auto& [bind_point, pipeline_id] : command_buffer_info->bound_pipelines)
     {
-        auto* pipeline_info = object_table_->GetVkPipelineInfo(pipeline_id);
+        const auto* pipeline_info = object_table_->GetVkPipelineInfo(pipeline_id);
         GFXRECON_ASSERT(pipeline_info != nullptr);
         if (pipeline_info != nullptr)
         {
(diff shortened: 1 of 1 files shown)


fabian-lunarg:
realizing the correction needs to be applied earlier, prior to `vkUpdateDescriptorSets`. I'll rework the solution

edit:
turned out we were in the right place (we know which pipeline/SPIRV is used only when the set is bound).
the only way to apply a correction seems to be an injected/repeated call to `vkUpdateDescriptorSet`.

fabian-lunarg:
confirmed this works now. 
can provide a capture, containing buffer-device-addresses in an inline-uniform-block,
currently failing replay (with -m rebind) on dev and fixed by this PR.

locke-lunarg (approved):

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -529,71 +531,127 @@ void VulkanAddressReplacer::ProcessCmdBindDescriptorSets(VulkanCommandBufferInfo
             continue;
         }
         GFXRECON_ASSERT(buffer_ref_info.set <= descriptorSetCount);
-        auto* descriptor_set_info =
-            object_table_->GetVkDescriptorSetInfo(pDescriptorSets->GetPointer()[buffer_ref_info.set]);
-        if (descriptor_set_info == nullptr)
+
+        // we need a mutable pointer, to allow for in-place corrections
+        auto* descriptor_set_info_mut = const_cast<VulkanDescriptorSetInfo*>(
```
bradgrantham-lunarg: Is this casting away a const?  Are you modifying data that other users of `object_table_->GetVkDescriptorSetInfo` will continue to view as const (and therefore assume it won't change)?

fabian-lunarg: yes and no.
the const_cast was necessary to allow modification of contained `descriptor_set_binding_info.inline_uniform_block.data()` in-place.

the reason we needed to const_cast was we only got hold of a `const decode::CommonObjectInfoTable& object_table` in ctor (this was sufficient so far).
I found we can get rid of casting by changing this reference to non-const.

fabian-lunarg: after the mentioned change, I made other structs queried from 
`decode::CommonObjectInfoTable* object_table_` 
explicitly `const auto*` (they were implicitly `const` before)

PR #2362, opened 2025-08-13
WIP: Initial pass at traitisfying FileProcessor::ProcessBlocks. 

jzulauf-lunarg:
Looking at modifying FileProcessor to prototype header size reductions, noticed that there are currently three "twisty little ProcessBlocks, all different." Looking to DRY this to aid in prototyping, and avoid missed changes (like the file stack in the preloading variant).

Notes:
- Looking for feedback on traitsifying scheme, maintainability, readability, etc.
- Nothing is is the right files.
- The interface changes to make the Process<foo>Block should be propagated to avoid the extra layer of specialization
- The rationalization of the combinatoric differences with the <Derived>::ProcessBlocks hasn't been done.

### DIFF
--- a/framework/decode/file_processor.cpp
+++ b/framework/decode/file_processor.cpp
@@ -293,6 +293,101 @@ void FileProcessor::DecrementRemainingCommands()
     }
 }
 
+template <typename SubBlockType>
+struct SubBlockTypeDefault
+{};
+template <>
+struct SubBlockTypeDefault<format::ApiCallId>
+{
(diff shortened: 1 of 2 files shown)


davidd-lunarg:
A base/derived class hierarchy for the file processor classes would be more natural for me. Is that feasible? What trade offs are there compare to the template approach?

jzulauf-lunarg:
@davidd-lunarg  -- base derived is what we have, with each derived class being combinatorically different in ProcessBlocks, with missing coverage (for example file_stack_ handling) in some.

This is maintenance fraught, especially as we evolve the schema for future features and encoding efficiency.

PR #2363, opened 2025-08-13
ci: Specify internal test suite commit in file

ncesario-lunarg:
Adds a file (TEST_SUITE_REF) to store a refspec in ci-gfxr-suites. This is to give developers control over which version of the suites file should be associated with their code. Specifically, this should be updated when changes need to be made to reference files in the test suite.


bradgrantham-lunarg:
Is this PR obsolete or needs updating?  The test suite ref is in `test_suite.ref`

ncesario-lunarg:
Yes, this is superseded by pipelines, as one can just modify https://github.com/LunarG/gfxreconstruct/blob/5fb702a9ad02c8c1e634ffc75c572e9ddf401756/ci/runJob.sh#L13 to use a custom suites ref if wanted/needed.

Issue #2349, opened 2025-08-07
Crash if ignored pointer is invalid


ziga-lunarg:
If the command buffer in `vkBeginCommandBuffer()` is a primary command buffer then `pInheritanceInfo` must be ignored. 

There was already an attempt to fix that in #1922

But as the code got changed it did not fix the original problem.

Capturing CTS test `dEQP-VK.api.command_buffers.bad_inheritance_info_random` would reproduce this.

The code that handles this is in vulkan_capture_manager.cpp.cpp:

```
VkResult VulkanCaptureManager::OverrideBeginCommandBuffer(VkCommandBuffer                 commandBuffer,
                                                          const VkCommandBufferBeginInfo* pBeginInfo)
{
    auto handle_unwrap_memory = VulkanCaptureManager::Get()->GetHandleUnwrapMemory();
    auto pBeginInfo_unwrapped = vulkan_wrappers::UnwrapStructPtrHandles(pBeginInfo, handle_unwrap_memory);

    auto modified_begin_info = (*pBeginInfo_unwrapped);

    const auto command_buffer_wrapper =
        vulkan_wrappers::GetWrapper<vulkan_wrappers::CommandBufferWrapper>(commandBuffer);

    // If command buffer level is primary, pInheritanceInfo must be ignored
    if (command_buffer_wrapper && command_buffer_wrapper->level == VK_COMMAND_BUFFER_LEVEL_PRIMARY &&
        modified_begin_info.pInheritanceInfo != nullptr)
    {
        modified_begin_info.pInheritanceInfo = nullptr;
    }

    return vulkan_wrappers::GetDeviceTable(commandBuffer)->BeginCommandBuffer(commandBuffer, &modified_begin_info);
}
```

But the crash happens already in:
`auto pBeginInfo_unwrapped = vulkan_wrappers::UnwrapStructPtrHandles(pBeginInfo, handle_unwrap_memory);`

PR #2366, opened 2025-08-14
Fix crash when primary cmd-buf has invalid inheritance-data

fabian-lunarg:
- move correction from OverrideBeginCommandBuffer() to a CustomEncoderPreCall
- cleanup override, not required anymore

closes [#2349](https://github.com/LunarG/gfxreconstruct/issues/2349)

### DIFF
--- a/framework/encode/custom_vulkan_encoder_commands.h
+++ b/framework/encode/custom_vulkan_encoder_commands.h
@@ -1805,6 +1805,16 @@ struct CustomEncoderPostCall<format::ApiCallId::ApiCall_vkImportFenceFdKHR>
 };
 
 template <>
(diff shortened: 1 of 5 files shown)


ziga-lunarg (changes requested):
This change is not enough, in `generated_vulkan_api_call_encoders.cpp` the `pBeginInfo` in `vkBeginCommandBuffer` must also be modified.

fabian-lunarg:
> This change is not enough, in `generated_vulkan_api_call_encoders.cpp` the `pBeginInfo` in `vkBeginCommandBuffer` must also be modified.

I found a better place to apply the correction was as a `CustomEncoderPreCall`. 
should make sure our generated (or any subsequent) code doesn't crash when attempting to unwrap the struct.

MarkY-LunarG (approved):

PR #2367, opened 2025-08-19
replay: Wait when GetStatus is expected to be VK_SUCCESS

antonio-lunarg:
Previously, we called vkGetFenceStatus in a tight loop until the fence would be signaled, but that might prevent other threads from doing their job.

Calling vkGetFenceStatus instead of waiting for the fence to be signaled is regarded as a good practice (https://developer.samsung.com/galaxy-gamedev/resources/articles/usage.html). While vkWaitForFences stalls execution wasting useful CPU cycles, calling vkGetFenceStatus allows the CPU to perform small jobs before querying for the fence status again.

By looking at this from the perspective of the captured application, these small jobs might involve doing other Vulkan work. For this reason, we call vkWaitForFences here, giving other threads a chance to run.


bradgrantham-lunarg:
Can you say more about how choosing `vkWaitForFences` is different here from the app point of view than a loop calling `vkGetFenceStatus`?  Either way this function doesn't return until the Fence is signaled, and either way other threads can be scheduled on other CPUs and either way this thread can be pre-empted.  Is the loop starving other threads that are waiting on access to the GPU or something?

antonio-lunarg:
> Can you say more about how choosing `vkWaitForFences` is different here from the app point of view than a loop calling `vkGetFenceStatus`? Either way this function doesn't return until the Fence is signaled, and either way other threads can be scheduled on other CPUs and either way this thread can be pre-empted. Is the loop starving other threads that are waiting on access to the GPU or something?

The only difference is that there's no busy loop. I have a case where one frame would originally call `vkGetFenceStatus() -> VK_SUCCESS` 5 times, but during replay these become more than 5 thousand, which is just unuseful work and IMHO it does not reflect the original behaviour of the application just as much as calling one `vkWaitForFences()`.

That said, I do not know if the busy loop determines other bad side-effects from a GPU driver perspective.

per-mathisen-arm:
A wait can have an indirect effect on the GPU, so from our standpoint it is much closer to the original behaviour to convert a vkGetFenceStatus to a vkWaitForFences than to have potentially thousands of vkGetFenceStatus calls, which also lead to higher CPU contention.

fabian-lunarg:
> Can you say more about how choosing `vkWaitForFences` is different here from the app point of view than a loop calling `vkGetFenceStatus`? Either way this function doesn't return until the Fence is signaled, and either way other threads can be scheduled on other CPUs and either way this thread can be pre-empted. Is the loop starving other threads that are waiting on access to the GPU or something?

we discussed yesterday and I agree, it's seems not to be about starving other threads, cannot see how. the most plausible explanation-attempt is that our previous synchronization (busy-waiting with `vkGetFenceStatus`) produces negative side-effects, symptom looks like missing synchronization. also for other reasons I think the `vkWaitForFences` is the cleaner solution: yields a deterministic command-stream and is way more cpu-efficient.

fabian-lunarg (approved):

fabian-lunarg:
@antonio-lunarg  maybe better to trim the comment down a bit? it reads a bit back-forth confusing and in the end does not explain much, only guesses. imo better to describe what we actually do. e.g.` // inject vkWaitForFences, synchronize fence-status`

bradgrantham-lunarg (approved):

antonio-lunarg:
> @antonio-lunarg maybe better to trim the comment down a bit? it reads a bit back-forth confusing and in the end does not explain much, only guesses. imo better to describe what we actually do. e.g.` // inject vkWaitForFences, synchronize fence-status`

Done

PR #2368, opened 2025-08-20
Enhanced rv annotation from capture

zongdu-arm:
Enhanced rv annotation for ray tracing, insert valid bits in the high bits of GPU VA and shader identifier to ensure their uniqueness.


framework/encode/d3d12_capture_manager.cpp
```diff
@@ -1462,7 +1462,17 @@ void D3D12CaptureManager::PreProcess_ID3D12Resource_Unmap(ID3D12Resource_Wrapper
                                             size,
                                             offset);
                                     }
+
                                     WriteFillMemoryCmd(memory_id, offset, size, start_address);
+
+                                    if (RvAnnotationActive() == true)
```
fabian-lunarg: couple of statement like this `if (RvAnnotationActive() == true)`
but this is a tautological way to write `if (RvAnnotationActive())` so we should simplify

zongdu-arm: I fixed it.

lunarpapillo:
The LunarG CI failure was due to a machine configuration issue.  I'm re-running.

zongdu-arm:
Why add this modification?
1. Some judgment conditions are missing or not rigorous, it sometimes causes crash.
2. The high bits of annotation is restored for remapping during replaying and reusing optimization.
3. From the following spec description, the shader identifier is not unique, so adds high bits.

> The data itself globally identifies the shader, so even if the shader appears in a different
> state object with same associations, like any root signatures, it will have the same
> identifier.

zongdu-arm:
This PR was created a long time ago and hasn't been updated with the latest changes. 
Merging it might cause problems, so I will  hold off on merging it for now. 
It will be merged later if needed.

PR #2369, opened 2025-08-20
VulkanAddressReplacer: Fix gpu-stalls on AMD by adjusting sync-stage

fabian-lunarg:
relevant for buffer-device-addresses and `-m rebind`:

- fixes recent regression, very poor performance of gpu-dispatch on AMD/radv  
- replace `VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT` with `VK_PIPELINE_STAGE_ALL_COMMANDS_BIT`

### DIFF
--- a/framework/decode/vulkan_address_replacer.cpp
+++ b/framework/decode/vulkan_address_replacer.cpp
@@ -388,11 +388,8 @@ VkSemaphore VulkanAddressReplacer::UpdateBufferAddresses(
         {
             if (!wait_semaphores)
             {
-                run_compute_replace(command_buffer_info,
-                                    addresses,
-                                    num_addresses,
-                                    address_tracker,
-                                    VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT);
+                run_compute_replace(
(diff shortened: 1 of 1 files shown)


MarkY-LunarG:
@fabian-lunarg, can you provide more details on what tests you ran to verify this did not cause other regressions?  What exactly was the performance hit we were seeing?

fabian-lunarg:
> @fabian-lunarg, can you provide more details on what tests you ran to verify this did not cause other regressions? What exactly was the performance hit we were seeing?

this PR is fixing a regression to a pipeline-barrier's synchronization-stage. it was introduced couple weeks ago, here https://github.com/LunarG/gfxreconstruct/pull/2283.

the performance hit was severe. qualitatively stalling replay. I tested this on steamdeck using radv and bisected to be sure. other changes in linked PR were fine and I cannot recall why I even adjusted the sync-stage back then.
I did the changes and tested replay with `-m rebind` using only nvidia hardware back then and there were no issues there.
think that's why I refer to AMD in the title, while in fact it's a rather generic regression-fix.

PR #2371, opened 2025-08-20
Fix `VulkanRebindAllocator::AllocateAHBMemory`

marius-pelegrin-arm:
According to the Vulkan specification:

(VUID-VkMemoryAllocateInfo-allocationSize-02383)
If the parameters define an import operation and the external handle type is `VK_EXTERNAL_MEMORY_HANDLE_TYPE_ANDROID_HARDWARE_BUFFER_BIT_ANDROID`, `allocationSize` must be the size returned by
`vkGetAndroidHardwareBufferPropertiesANDROID` for the Android hardware buffer.

As per the current state of the code, this function can only be reached if `memory_alloc_info->ahb` is not `nullptr`, so this applies. Thus `allocationSize` must neither be the one returned by `vkGetImageMemoryRequirements` nor the capture time allocation size.

I also added a `GFXRECON_ASSERT` to make sure that `memory_alloc_info->ahb` cannot be `nullptr`.

### DIFF
--- a/framework/decode/vulkan_rebind_allocator.cpp
+++ b/framework/decode/vulkan_rebind_allocator.cpp
@@ -843,6 +843,8 @@ VkResult VulkanRebindAllocator::BindBufferMemory2(uint32_t                      
 
 VkResult VulkanRebindAllocator::AllocateAHBMemory(MemoryAllocInfo* memory_alloc_info, const VkImage image)
 {
+    GFXRECON_ASSERT(memory_alloc_info->ahb != nullptr);
+
     VkImportAndroidHardwareBufferInfoANDROID importAHBInfo;
     importAHBInfo.sType  = VK_STRUCTURE_TYPE_IMPORT_ANDROID_HARDWARE_BUFFER_INFO_ANDROID;
     importAHBInfo.pNext  = nullptr;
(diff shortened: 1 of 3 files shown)


framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -843,6 +843,8 @@ VkResult VulkanRebindAllocator::BindBufferMemory2(uint32_t
 
 VkResult VulkanRebindAllocator::AllocateAHBMemory(MemoryAllocInfo* memory_alloc_info, const VkImage image)
 {
+    GFXRECON_ASSERT(memory_alloc_info->ahb != nullptr);
```
fabian-lunarg, line 846: to also be safe for Release-build, better to extend this assert to an if(...) statement, wrapping the call `functions_.get_android_hardware_buffer_properties` - I assume there would be little to do if we see a `nullptr`, so we could log a warning an return.

marius-pelegrin-arm: There IS something to do. If ahb is nullptr, then it is not an external memory import and the memory requirements can be queried normally. However it is assumed that if you call "AllocateAHBMemory" you already checked that you are in fact importing an AndroidHardwareBuffer. Currently, all places that call this function are checking this, but in order for it to be "future proof" I added this assert.

The goal is not to make it acceptable to call "AllocateAHBMemory" without a valid AHB.

fabian-lunarg: ok, sounds good. we must not call `AllocateAHBMemory` without existing `memory_alloc_info->ahb`,
already check/assert that in all places calling it. so adding this `GFXRECON_ASSERT` as additional check seems enough.

ziga-lunarg (approved):
LGTM

fabian-lunarg (approved):

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -853,17 +855,20 @@ VkResult VulkanRebindAllocator::AllocateAHBMemory(MemoryAllocInfo* memory_alloc_
     dedicatedAllocateInfo.pNext = &importAHBInfo;
     dedicatedAllocateInfo.image = image;
 
-    VkMemoryRequirements memoryRequirements;
-    functions_.get_image_memory_requirements(device_, image, &memoryRequirements);
+    VkAndroidHardwareBufferPropertiesANDROID androidHardwareBufferProperties;
+    androidHardwareBufferProperties.sType = VK_STRUCTURE_TYPE_ANDROID_HARDWARE_BUFFER_PROPERTIES_ANDROID;
+    androidHardwareBufferProperties.pNext = nullptr;
+    functions_.get_android_hardware_buffer_properties(
```
MarkY-LunarG, line 861: Should we assert if `function_.get_android_hardware_buffer_properties` is `nullptr`?  Otherwise, looks good to me.

MarkY-LunarG (approved):

framework/decode/vulkan_rebind_allocator.cpp
marius-pelegrin-arm: I don't think it is possible to be in a situation where an AHardwareBuffer exists at replay time without the extension being querried and available at replay time as well.

PR #2379, opened 2025-08-21
Avoid redundant active_files_ find ops

jzulauf-lunarg:
Cache a copy of the active_file_ entry iterator in the ActiveFileContext, and then never lookup by filename again.

For long file names hash and find were integer % of profiler cycles.

### DIFF
--- a/framework/decode/file_processor.cpp
+++ b/framework/decode/file_processor.cpp
@@ -218,7 +218,7 @@ bool FileProcessor::ProcessFileHeader()
     bool               success = false;
     format::FileHeader file_header{};
 
-    ActiveFiles& active_file = active_files_[file_stack_.front().filename];
+    assert(file_stack_.front().active_file != active_files_.end());
 
     if (ReadBytes(&file_header, sizeof(file_header)))
     {
@@ -537,7 +537,7 @@ bool FileProcessor::ReadCompressedParameterBuffer(size_t  compressed_buffer_size
 
 bool FileProcessor::ReadBytes(void* buffer, size_t buffer_size)
 {
-    auto file_entry = active_files_.find(file_stack_.back().filename);
+    const auto& file_entry = file_stack_.back().active_file;
     assert(file_entry != active_files_.end());
 
     if (util::platform::FileRead(buffer, buffer_size, file_entry->second.fd))
@@ -550,7 +550,7 @@ bool FileProcessor::ReadBytes(void* buffer, size_t buffer_size)
 
 bool FileProcessor::SkipBytes(size_t skip_size)
 {
-    auto file_entry = active_files_.find(file_stack_.back().filename);
+    const auto& file_entry = file_stack_.back().active_file;
     assert(file_entry != active_files_.end());
 
     bool success = util::platform::FileSeek(file_entry->second.fd, skip_size, util::platform::FileSeekCurrent);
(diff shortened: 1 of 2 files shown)


framework/decode/file_processor.cpp
```diff
@@ -218,7 +218,7 @@ bool FileProcessor::ProcessFileHeader()
     bool               success = false;
     format::FileHeader file_header{};
 
-    ActiveFiles& active_file = active_files_[file_stack_.front().filename];
+    assert(file_stack_.front().active_file != active_files_.end());
```
panos-lunarg: `GFXRECON_ASSERT` instead of `assert`

framework/decode/file_processor.h
```diff
@@ -193,6 +193,18 @@ class FileProcessor
     }
 
   private:
+    // Must be define before the s
```
panos-lunarg: Must be define before the _stack_

panos-lunarg (approved):

framework/decode/file_processor.h
```diff
@@ -250,24 +262,14 @@ class FileProcessor
...
     struct ActiveFileContext
     {
-        ActiveFileContext(std::string filename_) : filename(std::move(filename_)){};
-        ActiveFileContext(std::string filename_, bool execute_till_eof_) :
-            filename(std::move(filename_)), execute_till_eof(execute_till_eof_){};
+        ActiveFileContext(const ActiveFileIterator& active_file_, bool execute_til_eof_ = false) :
+            active_file(active_file_), execute_till_eof(execute_til_eof_){};
 
-        std::string filename;
+        ActiveFileIterator active_file;
```
panos-lunarg: Perhaps it would be safer to use a pointer or a reference instead of an iterator.
`unordered_map` iterators can be invalidated by an `emplace` if it causes a `rehash`. On the other hand, pointers and references are are only invalidated by erasing the element they are pointing/reference, even when the corresponding iterator is invalidated.

jzulauf-lunarg: I had thought iterators were safe until erasure, but I'm apparently wrong.  However reshash will also do moves of the elements themselves, contents pointers are just as at risk.  Best to just cache off a copy of the FD itself, that's safe.


Operations | Invalidated
-- | --
All read only operations, swap, std::swap | Never
clear, rehash, reserve, operator= | Always
insert, emplace, emplace_hint, operator[] | Only if causes rehash
erase | Only to the element erased

https://en.cppreference.com/w/cpp/container/unordered_map.html

jzulauf-lunarg:
@panos-lunarg -- here's a safe smart pointer version that will survive rehashing.  I also added refcount/close/cleanup for unreferenced open FD's.  I can split that piece out in a separate PR if you prefer.

framework/decode/file_processor.h
```diff
@@ -193,6 +193,54 @@ class FileProcessor
...
+            std::string GetFilename() const;
+            FILE*       GetFile() const;
+            bool        FileSeek(int64_t offset, util::platform::FileSeekOrigin origin);
+
+          private:
+            ActiveFiles& active_file;
+            friend class ActiveFiles;
+            Ref(ActiveFiles& active_file_);
+        };
+        using RefPtr = std::unique_ptr<Ref>;
```
panos-lunarg: Can't we just use a simple pointer or reference instead of this new class and the `unique_ptr`.

From https://en.cppreference.com/w/cpp/container/unordered_map.html:
```
References and pointers to either key or data stored in the container are only invalidated
by erasing that element, even when the corresponding iterator is invalidated.
```

So if I understand this correctly pointers or references are safe and should work for our case.
`SetActiveFile` functions then can just `file_stack_.emplace_back(file_entry->second, execute_till_eof);` and this new class will not be necessary. `file_stack_.pop_back()` should take care of freeing resources without the need of smart pointers

jzulauf-lunarg: Got rid of the RefPtr, but the ActiveFiles::Ref object allows for automated reference counting without having to use strong and weak shared pointers.  ActiveFileContext now HasA Ref, instead of a RefPtr.

The automated ref count allows for automated file closure (the file open logic is robust to filename robust at this point as well).  The ActiveFiles map may have closed files, but at least we aren't leaving all our open files opened until the FileProcessor is destroyed now.

panos-lunarg (approved):

PR #2380, opened 2025-08-26
Performance improvement of WriteBoundResourceStaging

laurentiu-erhan-arm:
Previously for every staging write, gfxreconstruct was introducing a vkQueueWaitIdle to ensure the data is correct when being read. This would have the result of a performance penalty if multiple staging writes would be chained.

The solution issues vkQueueSubmit for each of the staging buffer copies and waits for their fences on the next capture time vkQueueSubmit.

Change-Id: Icea25403432ce380a1a365a30e141008e409485d


laurentiu-erhan-arm:
To add more info, this is just a performance improvement of a previous commit/pull request: https://github.com/LunarG/gfxreconstruct/pull/1275

PR #2381, opened 2025-08-26
Check rebind allocate fail

locke-lunarg:


framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -780,6 +780,11 @@ VkResult VulkanRebindAllocator::BindBufferMemory(VkBuffer
                                 *bind_memory_properties);
             }
         }
+        else
+        {
+            GFXRECON_LOG_WARNING("AllocateMemory is failed: %s in Rebind BindBufferMemory.",
```
fabian-lunarg: better say 'AllocateMemory failed: ...'

fabian-lunarg (approved):

framework/decode/vulkan_rebind_allocator.cpp
locke-lunarg: fixed

PR #2383, opened 2025-08-27
Fix ID3D10Blob trimmed capture

dgraves:
ID3D10Blob creation commands are currently written to the trimmed file state snapshot before root signature and PSO creation. This is done to handle the case where blob objects are used to create root signatures.

There is a second use case for blobs that is not currently handled correctly. This case is for cached PSO blobs returned by PSO objects. With the current order, replay processing for these blobs will fail because they are being written to the capture file before their parent object.

This change splits blobs processing into two passes, with the first pass processing root signature blobs before root signature/PSO creation and the second pass processing cached PSO blobs after PSO creation.


lunarpapillo:
Sorry, aborted CI by mistake.  Restarting CI.

davidd-lunarg (approved):

PR #2385, opened 2025-08-28
Fix in ReadImageResources

panos-lunarg:
In case of multi sampled images the resolved image was never used. Instead the CopyImageToBuffer was attempting to copy directly from the multi sampled image causing visual artifacts.

### DIFF
--- a/framework/graphics/vulkan_resources_util.cpp
+++ b/framework/graphics/vulkan_resources_util.cpp
@@ -1788,7 +1788,7 @@ VkResult VulkanResourcesUtil::ReadImageResources(const std::vector<ImageResource
                 command_buffer = cmd_buf_it->second;
             }
 
-            VkImage copy_image = tmp_data[i].resolve_image != VK_NULL_HANDLE ? tmp_data[i].resolve_image : img.image;
+            VkImage copy_image = img.image;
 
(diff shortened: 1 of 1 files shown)


fabian-lunarg (approved):

framework/graphics/vulkan_resources_util.cpp
```diff
@@ -1842,14 +1844,15 @@ VkResult VulkanResourcesUtil::ReadImageResources(const std::vector<ImageResource
...
 
-                copy_image = tmp_data[i].scaled_image != VK_NULL_HANDLE ? tmp_data[i].scaled_image : copy_image;
-
                 if (result != VK_SUCCESS)
                 {
                     // free temporary resource, continue
                     tmp_data[i] = {};
                     continue;
                 }
+
```
MarkY-LunarG, line 1865: Two questions:
1) Why did you move it after the result check above?
2) Is this fixing anything from the old check or was this just a code readability change?

panos-lunarg: 1. If `result != VK_SUCCESS` then there is no point updating `copy_image` variable at all.
2. In contrast to the change with `tmp_data[i].resolve_image`, this specific change does not fix something, just readability and uniformity with the `resolve_image` counterpart

MarkY-LunarG (approved):

PR #2386, opened 2025-08-28
Set extension properties of physical_device_info

locke-lunarg:
A crash occurred in `VulkanReplayConsumerBase::OverrideGetRayTracingShaderGroupHandlesKHR` because `physical_device_info->replay_device_info->raytracing_properties` was null. 

The reason was the system had two GPUs, A and B, and the device used GPU B, but when the capture replayed `vkGetPhysicalDeviceProperties2`, it retrieved GPU A's properties instead. As a result, GPU B's properties were never set. In this case, GPU B's properties should have been set in `VulkanDeviceUtil::GetReplayDeviceProperties`.

It should run `VulkanDeviceUtil::GetReplayDeviceProperties` when any property in `replay_device_info` is null , not just when the main `properties` is null.

### DIFF
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -172,6 +172,13 @@ struct VulkanReplayDeviceInfo
     std::optional<VkPhysicalDeviceDriverProperties>                   driver_properties;
     std::optional<VkPhysicalDeviceRayTracingPipelinePropertiesKHR>    raytracing_properties;
     std::optional<VkPhysicalDeviceAccelerationStructurePropertiesKHR> acceleration_structure_properties;
+
+    bool IsPropertiesNull()
+    {
+        // Not include memory properties.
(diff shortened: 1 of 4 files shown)


framework/decode/vulkan_object_info.h
```diff
@@ -172,6 +172,13 @@ struct VulkanReplayDeviceInfo
     std::optional<VkPhysicalDeviceDriverProperties>                   driver_properties;
     std::optional<VkPhysicalDeviceRayTracingPipelinePropertiesKHR>    raytracing_properties;
     std::optional<VkPhysicalDeviceAccelerationStructurePropertiesKHR> acceleration_structure_properties;
+
+    bool IsProperiesNull()
```
fabian-lunarg: typo in `IsProperiesNull`

fabian-lunarg (approved):

PR #2387, opened 2025-09-01
Add sccache to reuse intermediates across ci runs

mikes-lunarg:
This is an experiment to test using [sccache](https://github.com/mozilla/sccache) to speed up GitHub CI builds by caching intermediate build artifacts.

This change adds sccache to the Linux, Android, and macOS builds. It should be possible to add Windows in the future, but it will likely require extra steps ([example: MSVC debug format](https://deepwiki.com/mozilla/sccache/1.2-installation-and-setup#msvc-specific-configuration)).

### DIFF
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -38,6 +38,8 @@ jobs:
       uses: actions/checkout@v4
       with:
         submodules: 'recursive'
+    - name: Setup sccache
+      uses: mozilla-actions/sccache-action@v0.0.9
     - name: Install build dependencies
       run: |
         sudo apt-get update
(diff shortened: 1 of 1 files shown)


mikes-lunarg:
### Timing results:

Run 1 (pre cache)
Screenshot 2025-08-31 224748

Run 2 (post cache)
Screenshot 2025-08-31 230718

This represents the best-case speedup since no source files changed between the two runs. I expect that the real gains will not be quite as dramatic. The biggest winner is the macOS runner which normally struggles to build some of the large generated files. Using this method, those large files will only need to be rebuilt during header updates or when a specific generator changes.

PR #2389, opened 2025-09-02
Set C++ dialect to c++20, fix compilation-issues around wchar_t

fabian-lunarg:


bradgrantham-lunarg:
@fabian-lunarg Some notes from an interested party:
* CMakeLists.txt in `android/` need updating
* Probably should add a warning in a prominent source file checking the compiler's C++ language support to make sure it's >= 20
* On Android the clang support is tied to the NDK version so probably need to rev the NDK version

fabian-lunarg:
> @fabian-lunarg Some notes from an interested party:
> 
>     * CMakeLists.txt in `android/` need updating
> 
>     * Probably should add a warning in a prominent source file checking the compiler's C++ language support to make sure it's >= 20
> 
>     * On Android the clang support is tied to the NDK version so probably need to rev the NDK version

I handled 1+2 but would be happy about advise on 3 'rev the NDK version'
we had some input/discussion:

_NDK r23+ (Clang 12+, libc++ 13+), but safer with NDK r25+ (Clang 14+) → ensures std::format, ranges, coroutines, modules, etc. work without caveats_

but then also the observation that github-CI compiles fine for Android/c++20 using NDK 22.1.7171670.
so for the migration NDK 22+ 'should' suffice.

android/framework/cmake-config/PlatformConfig.cmake
```diff
@@ -1,4 +1,4 @@
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
```
roberto-c, line 1: Also adjustments to Gradle build files may be required as they also set the standard to use. If possible, I would prefer to remove the compiler options in the Gradle build script to here to avoid unexpected overrides. 

https://github.com/LunarG/gfxreconstruct/blob/c365eecd5341daade8ffe411cfe984cb6a75c34b/android/layer/build.gradle#L26-L28

fabian-lunarg: done, removed occurrences of -std=c++XX from gradle-files. I'm not sure those were previously picked up at all or  overwritten by cmake-config. in any case the duplication wasn't nice.

bradgrantham-lunarg: @roberto-c thumbs-upped, so I resolved

PR #2390, opened 2025-09-02
Make --preserve-capture-data keep Handle IDs

davidd-lunarg:
When using `gfxrecon-replay --capture --preserve-capture-data`, GFXR
will reuse the original captured Handle ID so that the objects in the
resulting capture file will have the same ID as the source capture file.
The `--preserve-capture-data` feature is under development and its
usage or behavior may change.


bartosz-muszarski-arm:
Shouldn't it be a default behavior in `--capture` mode? Looks very useful for tracking and debugging.

davidd-lunarg:
> Shouldn't it be a default behavior in `--capture` mode? Looks very useful for tracking and debugging.

I agree. I have updated the PR to make this the behavior for `gfxrecon-replay --capture`.

framework/encode/capture_manager.h
```diff
@@ -69,7 +69,18 @@ class CommonCaptureManager
...
+
+  public:
+    static void SetDefaultUniqueIdOffset(format::HandleId offset) { default_unique_id_offset_ = offset; }
+
+    static format::HandleId GetUniqueId();
+    static void             PushUniqueId(const format::HandleId id);
+    static void             ClearUniqueIds();
+
+    // Set to true to force capture manager to generate new, default IDs instead of using the unique ID stack.
+    static void SetForceDefaultUniqueId(bool force) { force_default_unique_id_ = force; }
```
bradgrantham-lunarg, line 83: Does this need to be atomic or guarded or per-thread in the case that multiple threads call `*InjectedCommands`?  (e.g. in the case of `--pcj`)

bradgrantham-lunarg (approved):
I have the one concern about multi-threaded access to the introduced `force` function.

framework/encode/capture_manager.h
davidd-lunarg: This is a good point. The new vars to control UID generation may need to be thread local or otherwise made thread-safe. I'll look more into this.

davidd-lunarg:
> I have the one concern about multi-threaded access to the introduced `force` function.

I've added commit https://github.com/LunarG/gfxreconstruct/pull/2390/commits/b4ebf14c53d040739c73716c77bc893a2c813aa8 that addresses this issue. The `unique_id_stack_` and `force_default_unique_id_` vars are thread local and the async methods for creating pipelines push the unique ids from the tread creating the pipeline.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -11609,11 +11668,14 @@ std::function<handle_create_result_t<VkPipeline>()> VulkanReplayConsumerBase::As
                  in_pAllocator,
                  createInfoCount,
                  create_info_data = std::move(create_info_data),
-                 handle_deps      = std::move(handle_deps)]() mutable -> handle_create_result_t<VkPipeline> {
+                 handle_deps      = std::move(handle_deps),
+                 pipeline_ids     = std::move(pipeline_ids)]() mutable -> handle_create_result_t<VkPipeline> {
         std::vector<VkPipeline> out_pipelines(createInfoCount);
         auto     create_infos  = reinterpret_cast<const VkComputePipelineCreateInfo*>(create_info_data.data());
+        PushRecaptureHandleIds(pipeline_ids.data(), pipeline_ids.size());
```
davidd-lunarg, line 11807: Moving this comment to the PR from an obscure commit line, sorry for the duplicate:

@fabian-lunarg Could you look over these changes to the async pipeline creation code (commit https://github.com/LunarG/gfxreconstruct/commit/31168929a772382dc3719409dd68587d6d6bff0b)? No need to review the full PR unless you want to. When running `gfxrecon-replay --pcj N --capture <...>`, `PushRecaptureHandleIds` is used by replay to tell the CaptureManager what Handle IDs to assign the next objects it creates.

fabian-lunarg (approved):

PR #2391, opened 2025-09-02
WIP: Prototype of memory mapped file input

jzulauf-lunarg:
Prototypes use of memory mapped file reading on Window.  The posix path is defined, but not even build tested.  Known broken is the "Preloading" file processor.  Parameters are not tuned.  Page unmapping is asynchronous as that was profiled as slow (a known windows issue), a preliminary "prefetching" implementation is present on the Windows path, but didn't appear to improve performance. Needs further evaluation.

Update: pushed rebase with #if 0'd dead code removed.  Commits aren't quite right, but the code is what is benchmarked on Windows.  Next step: clean compile on POSIX -- test gfxinfo on WSL2.


per-mathisen-arm:
What is the benefit of this, considering that the file content is (mostly, usually) compressed anyways?

PR #2393, opened 2025-09-04
Add forgotten documentation for `GFXRECON_CAPTURE_PROCESS_NAME`

marius-pelegrin-arm:


fabian-lunarg (approved):

USAGE_android.md
```diff
@@ -266,6 +266,18 @@ adb shell settings put global gpu_debug_layers VK_LAYER_LUNARG_gfxreconstruct
 adb shell settings put global gpu_debug_layer_app com.lunarg.gfxreconstruct.replay
 ```
 
+You can also restrict the layer to a specific application using these three steps:
+1. Push the GFXReconstruct capture layer to `/data/local/debug/vulkan` directory.
```
bradgrantham-lunarg, line 270: Does this directory work for all Android installations?  Will this require rooting for some devices?

marius-pelegrin-arm: According to the Android documentation (https://source.android.com/docs/core/graphics/implement-vulkan), the method with the application folder is supposed to be used for non-debuggable apps, and `/data/local/debug/vulkan` is supposed to be used for debuggable apps.

So this path is available on all Android installation and does not require rooting... in theory. In practice I wouldn't be surprise to encounter different behaviors. The Android specification has always been considered more as a suggestion than an actual rule for vendors.

But if you want to stick to the Android spec, you can remove the line above: `Make sure the target application is debuggable` and replace it by an "if the target application is/is not debuggable"

bradgrantham-lunarg (approved):

PR #2394, opened 2025-09-05
Correct read errors with preload enabled

jzulauf-lunarg:
Fixes regression to preload caused by file processor IO cleanup.


jzulauf-lunarg:
@beau-lunarg: A test for this feature would be the replay of any content enabling the feature with the flags set... `--measurement-frame-range <start, end>  --preload-measurement-range`  enable where `end > start + 1`. AFAICT preload wasn't effectively enabled for end = start + 1, though my testing of this corner case was minimal.  

Advise use of `start, start+2`.

fabian-lunarg (approved):

panos-lunarg (approved):

PR #2397, opened 2025-09-09
Fileprocessor cleanups

jzulauf-lunarg:
Two substantial cleanups of the FileProcessor

 1: Refactor FileProcessor file access

    Reroute all file access though ActiveFiles::Ref such that active_files_ can be cleaned up and closed at EOF
    Isolate file input operations to a file input stream class

2:    Rewrite PreloadPreloadFileProcessor::ProcessBlocks 

   Eliminates a ProcessBlock non-dry variant
   Change data accumulation from N-squared to N
   
Tested again CI extended trace list for performance and functionality on gfxinfo and replay.  For non-preload, no noted performance impact.  For --preload-measurement frames, was able to demonstrate linear preload performance after commit vs. measured N-square performance prior.  Was able to preload 100's of frames for non-trivial trace.


framework/util/zlib_compressor.h
```diff
@@ -41,10 +41,10 @@ class ZlibCompressor : public Compressor
                     std::vector<uint8_t>* compressed_data,
                     size_t                compressed_data_offset) const override;
 
-    size_t Decompress(size_t                      compressed_size,
-                      const std::vector<uint8_t>& compressed_data,
-                      size_t                      expected_uncompressed_size,
-                      std::vector<uint8_t>*       uncompressed_data) const override;
+    size_t Decompress(size_t                compressed_size,
+                      const uint8_t*        compressed_data,
```
MarkY-LunarG, line 45: I'm curious, why this change?  Does it help that much to pass the data in as a pointer versus the reference?

framework/decode/preload_file_processor.cpp
```diff
@@ -27,42 +27,18 @@
...
 
 void PreloadFileProcessor::PreloadNextFrames(size_t count)
 {
-    status_ = PreloadStatus::kRecord;
     while (--count != 0U)
     {
-        ProcessNextFrame();
+        DoProcessNextFrame([this]() { return this->PreloadBlocksOneFrame(); });
     }
-    status_ = PreloadStatus::kReplay;
```
MarkY-LunarG, line 39: Why did you remove  this?

framework/decode/preload_file_processor.h
```diff
@@ -38,75 +40,80 @@ class PreloadFileProcessor : public FileProcessor
...
-        // Copies provided object of type T into the preload buffer
-        // Returns a pointer to inserted object in the container
         template <typename T>
-        inline void* Add(T* data)
+        bool Read(T& value, size_t at) const
         {
-            return &*container_.insert(
-                container_.end(), reinterpret_cast<char*>(data), reinterpret_cast<char*>(data) + sizeof(T));
+            // Ensure that this isn't being misused.
+            static_assert(std::is_trivially_copyable_v<T>, "Read<T> requires a trivially copyable type");
```
MarkY-LunarG, line 71: Do we not have a GFXR static assert macro?

MarkY-LunarG (approved):
These are big changes, I would suggest you test and then pass back to ARM for testing once you're done before you merge.

framework/decode/file_processor.h
```diff
@@ -175,28 +176,41 @@ class FileProcessor
...
+        {
+            return true;
+        }
+        return file_stack_.back().active_file.IsEof();
     }
 
   private:
     // Must be define before the Seek calls below
-    class ActiveFiles
+    class ActiveFiles : public FileInputStream
```
davidd-lunarg: Can a `shared_ptr` be used instead of the `ActiveFiles` class? For example, something like:

1. Replace `using ActiveFilePtr = std::unique_ptr<ActiveFiles>;` with `using ActiveFilePtr = std::shared_ptr<FileInputStream>;`
1. Add a destructor to `FStreamFileInputStream` to `Close()` the file if open
2. Remove `class ActiveFiles`

jzulauf-lunarg: I'd need to see how the whole ref counting to close would fit in.  FStreamInputFileStream knows nothing of the file_stack... I'll take a look.  I'm all for simpler.

framework/util/zlib_compressor.h
jzulauf-lunarg: Spans will give us pointers when I merge them... so this change is a bit forward looking.

jzulauf-lunarg: @MarkY-LunarG  -- I'm thinking I'll need to split this PR then, but the clean up is needed for the further cleanup WIP to move forward.  Full mutated clones of ProcessBlocks complicate things dramatically.  Especially when the clones are feature incomplete.

framework/decode/preload_file_processor.cpp
jzulauf-lunarg: there is no longer any "mode of operation" in ProcessBlocks which was the reason for it.  In fact there is no longer a PreloadFileProcessor::ProcessBlocks override.  PreLoadBlocks replaces it more DRYly.

framework/decode/preload_file_processor.h
jzulauf-lunarg: No, AFAICT

framework/decode/file_processor.h
jzulauf-lunarg: @davidd-lunarg -- yes.  The whole thing can be simplified.  The Open/SetActiveFile would be merged and the "active_files_" map would just go away in favor of a shared_ptr in the ActiveFileContext (which I haven' seen if I can *also* eliminate)

jzulauf-lunarg: @davidd-lunarg -- Updated as reccomended.  ActiveFileContext has specific additional information not suited for the FileInputStream.

jzulauf-lunarg:
@fabian-lunarg Could you take a look  at the Preload changes?

framework/decode/file_processor.cpp
```diff
@@ -561,10 +512,10 @@ bool FileProcessor::ReadBytes(void* buffer, size_t buffer_size)
 
 bool FileProcessor::SkipBytes(size_t skip_size)
 {
-    auto& file_entry = file_stack_.back().active_file;
-    assert(file_entry);
+    const auto& active_file = file_stack_.back().active_file;
+    assert(active_file);
```
davidd-lunarg: Since this `assert` is being modified, it could be updated to use `GFXRECON_ASSERT`. Same for a couple other `assert`s in this PR.

fabian-lunarg (approved):
looks good!

framework/decode/preload_file_processor.h
bradgrantham-lunarg: What would the benefit of a GFXR static assert macro be?  We use GFXRECON_ASSERT to wrap a test with logging, but a compile-time assert wouldn't invoke logging.  I suspect bare C++ `static_assert` is sufficient.

jzulauf-lunarg:
@bradgrantham-lunarg @davidd-lunarg @panos-lunarg -- I added back the ability to keep track of the "active" set of open files.  turns out the kExecuteBlocksFromFile trace examples we have keep going back to the same file, s.t. without a cache we have close/open churn.

Note ClockCache use vs. the ad hoc ActiveFiles of the original.

@mikes-lunarg  or @fabian-lunarg -- you want to give a good look at ClockCache (it's a classic old algorithm, my only twist is that the lookup key *isn't* the value... )

panos-lunarg:
> @bradgrantham-lunarg @davidd-lunarg @panos-lunarg -- I added back the ability to keep track of the "active" set of open files. turns out the kExecuteBlocksFromFile trace examples we have keep going back to the same file, s.t. without a cache we have close/open churn.

Right. Now I remember from a discussion with Brad I encountered the same problem. Trimmed captures, especially games with large number of resources, utilizing an asset file had a significant increase in their frame state setup times.
An example I can find from back then. A capture from a game with ~58100 `ExecuteBlocksFromFile` in its frame state setup.
1. Closing and opening the asset file for each `ExecuteBlocksFromFile` - > 155 seconds to load
2. Keeping the files open -> 10 seconds

davidd-lunarg (approved):
Looks good to me,

davidd-lunarg:
> I added back the ability to keep track of the "active" set of open files. turns out the kExecuteBlocksFromFile trace examples we have keep going back to the same file, s.t. without a cache we have close/open churn.

Out of curiosity, how many files are expected to be open at once? Or how many different files referenced from one capture file?

panos-lunarg:
> Out of curiosity, how many files are expected to be open at once? Or how many different files referenced from one capture file?

For asset/index files I think it should be 2 - 3 max

jzulauf-lunarg:
> > Out of curiosity, how many files are expected to be open at once? Or how many different files referenced from one capture file?
> 
> For asset/index files I think it should be 2 - 3 max

CacheSize is pow and currently 3 (i.e. 8 entries) and set in a `using` statement for the Cache.  ActiveFiles was open ended, and never closed files until ~FileProcessor

PR #2395, opened 2025-09-09
VDR: Ability to request dumping of specific subresources

panos-lunarg:
This commit gives the ability for a more fine-grained control of which descriptors are dumped with each draw call / dispatch / trace rays command.

Each descriptor can be distinguished by a descriptor set index, binding index and array index tuple. A list of such tuples can now be provided with each command index in order to specify exactly which descriptors to dump.

Additionally, for image descriptors, there’s is now an option to provide a VkImageSubresourceRange describing which of the image’s subresources to be dumped.

A more detailed example can be found in vulkan_dump_resources.md

Issue #2239, opened 2025-06-10
VDR: Pass dump resource related options through the input json instead of CLI options


panos-lunarg:
The CLI options for the Vulkan dump resources feature are already numerous and might need to add more in the future. @fabian-lunarg  proposed a possibly better alternative of passing these options through the input json file instead. A side effect of this will be that it won't be possible to pass these options when not using the input json and just passing the indices via CLI.

Objectives:
- Implement passing VDR options via the input json
- Apart from `--dump-resources`, deprecate rest of VDR CLI options

PR #2398, opened 2025-09-10
VDR: acceleration structures

panos-lunarg:
This commit adds support for dumping acceleration structures that are
referenced as descriptors from draw calls / trace rays commands.

Currently the buffers that should be dumped:
- Input buffers used in vkCmdBuildAccelerationStructuresKHR which are:
  - Vertices, Indices and transformation matrices
  - AABBs
  - Instance buffers used for building TLAS
- The serialized acceleration structures

Currently only VkAccelerationStructureKHR is supported. Also host side acceleration
structures commands are not supported.


MarkY-LunarG:
I'm concerned we're starting to have a ton of "dump-resources" related arguments.  It's getting kind of crazy..  Should we step back and evaluate having a "control file" instead?  @bradgrantham-lunarg what do you think?

csteinberg-samsung:
> I'm concerned we're starting to have a ton of "dump-resources" related arguments. It's getting kind of crazy.. Should we step back and evaluate having a "control file" instead? @bradgrantham-lunarg what do you think?

There is already the request json file itself. I'm not sure I like the idea of two files to manage, but the options could be baked into the request json perhaps? 

I had suggested a few months ago that the request json format be expanded to be more explicit in what gets dumped instead of high level operations that control broad categories of things to dump. I had provided an early proposal for that a while back. I will admit it might have been a bit too verbose however. 

The sub resource ranges kind of move in that direction already, explicitly defining which immutable resources (descriptors) get dumped and what range of data is dumped (when they are images), but its more limited in its aggressive verboseness compared to my original proposal. 

Maybe there is some happy medium here where the options can be provided in the request json. I don't know if they should be top level still or applied on a per draw/dispatch/trace ray event level?

csteinberg-samsung:
any change to how the flags are supplied would be a significant refactor for us just fyi. I would suggest maybe just adding the flag here to include dumping the build inputs as part of this pr and any rework for the control flags be done in a future pr to break up the integration effort?

panos-lunarg:
> I'm concerned we're starting to have a ton of "dump-resources" related arguments. It's getting kind of crazy.. Should we step back and evaluate having a "control file" instead? @bradgrantham-lunarg what do you think?

Yes I think we should (https://github.com/LunarG/gfxreconstruct/issues/2239). But not as part of this PR

bradgrantham-lunarg:
The option name `--dump-resources-dump-build-acceleration-structures-input-buffers` does feel almost like a quiet protest against the dump-resources argument naming so far.  It's 65 characters long.  `--dr-as-inputs` probably would have sufficed.

There's already an issue stating we should probably move the options into the input file: https://github.com/LunarG/gfxreconstruct/issues/2239

@MarkY-LunarG wrote:
> I'm concerned we're starting to have a ton of "dump-resources" related arguments. It's getting kind of crazy.. Should we step back and evaluate having a "control file" instead? @bradgrantham-lunarg what do you think?

There's already a JSON input file for dump-resources.  I think that would be a fine place for these options.

@csteinberg-samsung wrote:
> any change to how the flags are supplied would be a significant refactor for us just fyi.

Just to be clear, you're saying it's a significant refactor to change from providing the options on the command-line to providing the options in the JSON dump-resources specification?  That's interesting to me - is creating the JSON strongly separated from the invocation of `gfxrecon-replay` in your software?

> I would suggest maybe just adding the flag here to skip dumping the build inputs as part of this pr and any rework for the control flags be done in a future pr to break up the integration effort?

I think it's problematic to add dump outputs without an enable or with only an explicit disable because other users of dump-resources will suddenly get more data than they expected, implying new JSON output and disk storage and runtime and transfer time.

If we were to move this into the JSON and not provide through a CLI option, from here forward it would only be possible to fully specify dump-resources inputs using the JSON.

I think we **should** move to specifying input options in the JSON file and remove them from the command-line.  If we push this off it just won't get done.  @csteinberg-samsung I need to know how much of an effort it will be to move from booleans on the command-line to booleans in the JSON file.  (also @roberto-c I need to know how much of an impact that is on you.)

@panos-lunarg can you sketch out what it might look like to bring all the CLI options into the input JSON?  At least the ones that make sense.

csteinberg-samsung:
> Just to be clear, you're saying it's a significant refactor to change from providing the options on the command-line to providing the options in the JSON dump-resources specification? That's interesting to me - is creating the JSON strongly separated from the invocation of `gfxrecon-replay` in your software?

Yes they are two different parts of a process, one is constructing the dump resources request file, the other constructs the command line arguments to invoke a gfxr replay which includes more than just the dump resources options. We can merge/move them, I'm happy enough to do that, but its still an effort and seems unrelated to the work of dumping acceleration structures itself.

csteinberg-samsung:
> I think it's problematic to add dump outputs without an enable or with only an explicit disable because other users of dump-resources will suddenly get more data than they expected, implying new JSON output and disk storage and runtime and transfer time.

Yes I think I agree with this as well, I think the flag enables additional data though, it doesn't skip/disable it. Acceleration structures are just a piece of immutable data (descriptors) that was not being dumped originally. There could be a separate flag to explicitly enable dumping acceleration structures but then there will be two new flags instead of one.

bradgrantham-lunarg:
> > I think it's problematic to add dump outputs without an enable or with only an explicit disable
> 
> Yes I think I agree with this as well, I think the flag enables additional data though, it doesn't skip/disable it. Acceleration structures are just a piece of immutable data (descriptors) that was not being dumped originally. There could be a separate flag to explicitly enable dumping acceleration structures but then there will be two new flags instead of one.

Earlier you said "_I would suggest maybe just adding the flag here to skip dumping the build inputs as part of this pr_".

csteinberg-samsung:
> Earlier you said "_I would suggest maybe just adding the flag here to skip dumping the build inputs as part of this pr_".

That was a mistake, I had edited it immediately afterwards, the flag Panos added is to ADD the build inputs to the dump, not skip it. It skips it by default.

csteinberg-samsung:
note that it only controls the build input addition though, it will currently still dump the serialized acceleration structure bin if its part of the descriptors. The immutable resources flag controls dumping that currently. It skips dumping all descriptors by default, then dumps all descriptors when the immutable flag is set, note that the descriptors field was added in pr https://github.com/LunarG/gfxreconstruct/pull/2395 which if provided acts as a white list for descriptors to dump, but you would need to supply ALL descriptors minus acceleration structures if you wanted to filter them out specifically. That would match the current behavior seen by apps today after this pr is merged if no additional flag is provided to INCLUDE dumping acceleration structure descriptor sets when the immutable flag is set. 

Summary of the settings:
* no flags set - No descriptors including no Acceleration Structures (current behavior)
* immutable flag set - Dumps all descriptors including acceleration structure (with this PR - did not dump AS before this pr)
* immutable flag + build inputs flag set - Dumps all descriptors including accelerations structures + the build inputs of acceleration structures (with this PR - did not dump AS before this pr and the build input flags did not exist)
* immutable flag + Descriptors array - dumps only the descriptors listed in the descriptors list (plus only sub-resource ranges of image type descriptors if provided - current behavior as of #2395)

bradgrantham-lunarg:
@panos-lunarg and I discussed - this PR will remain the addition of the CLI flag.  A subsequent PR (which I would argue should be the next VDR pull request) should pull the CLI options into the JSON input specification.

csteinberg-samsung:
sounds good, ill start preping to merge our handling of the CLI operations into the code that generates the request json on our end.

framework/decode/vulkan_device_address_tracker.cpp
```diff
@@ -229,5 +237,42 @@ VulkanDeviceAddressTracker::GetBufferDeviceAddressMap() const
     return address_lookup_helper_map_;
 }
 
+VulkanPerDeviceAddressTrackers* VulkanPerDeviceAddressTrackers::singleton_ = nullptr;
```
bradgrantham-lunarg: I think using a pointer for this object opens it up to error.  You could put this as a static inside GetDeviceAddressTracker and then it doesn't need to be a pointer and would look more like the previous implementation in VulkanReplayConsumerBase and then not expose the singleton.

framework/decode/vulkan_replay_consumer_base.h
```diff
@@ -1871,8 +1870,8 @@ class VulkanReplayConsumerBase : public VulkanConsumer
     std::string                                                                    screenshot_file_prefix_;
     graphics::FpsInfo*                                                             fps_info_;
 
-    std::unordered_map<const decode::VulkanDeviceInfo*, decode::VulkanDeviceAddressTracker> _device_address_trackers;
-    std::unordered_map<const decode::VulkanDeviceInfo*, decode::VulkanAddressReplacer>      _device_address_replacers;
+    VulkanPerDeviceAddressTrackers*                                                    _device_address_trackers;
```
bradgrantham-lunarg: Why bring the map from device to address tracker into a struct but not do the same for the address replacers?

framework/decode/vulkan_replay_dump_resources.h
```diff
@@ -574,6 +600,9 @@ class VulkanReplayDumpResourcesBase
 
     std::string capture_filename;
 
+    DumpResourcesAccelerationStructuresContext acceleration_structures_context_;
+    bool                                       dump_as_build_intput_buffers_;
```
bradgrantham-lunarg: Please change to `dump_as_build_input_buffers_`.

panos-lunarg: Fixed

framework/decode/vulkan_device_address_tracker.cpp
panos-lunarg: Should be fixed now

framework/decode/vulkan_replay_consumer_base.h
panos-lunarg: I don't need the address replacer in vdr code and since it's separate from the address tracker I didn't touch this. Should I do the same for uniformity?

panos-lunarg: Converted address replacer into a singleton as well.

framework/decode/vulkan_device_address_tracker.h
```diff
@@ -164,6 +169,19 @@ class VulkanDeviceAddressTracker
     std::unordered_map<VkDeviceAddress, device_address_range_t> address_lookup_helper_map_;
 };
 
+class VulkanPerDeviceAddressTrackers
```
fabian-lunarg: why is it necessary to manage the objects with static lifetime? 
I can see benefits when accessing an instance from multiple locations, but the  resulting ownership/lifetime isn't really clear to me. when are the resources cleaned up?

`VulkanPerDeviceAddressTrackers` is just wrapping a map of objects. wouldn't a ~~static~~ shared-pointer to a map work the same?

panos-lunarg: Good question.

Currently on dev the destruction of the objects held in the map needs to be [done manually](https://github.com/LunarG/gfxreconstruct/blob/fcd47b13740c746ec392e13731e575ebc79a2551/framework/decode/vulkan_replay_consumer_base.cpp#L274). This makes sense as the class holds vulkan objects which need to be destroyed before the parent objects are destroyed. So I am not sure if letting the destruction to be handled automatically (when the shared pointer gets out of scope from all holders) is safe as it needs to have some timing.

In this PR the destruction is still handled by the destructor of [`VulkanReplayConsumerBase` ](https://github.com/LunarG/gfxreconstruct/pull/2398/files#diff-899abc3a63198f8500fe8c411f311b26dd976249b5cf9f923561180da755e4deR277)

fabian-lunarg: ok right, we cannot rely on RAII and the cleanup was done manually before too, in order to achieve proper timing of resource cleanup. I still wish we wouldn't introduce more singletons or I'm not really seeing the benefit in this case. it's additional code to do the same per-device management we already had and putting the per-device logic next to the class definitions of VulkanAddressFoo reduces orthogonality.

framework/decode/vulkan_replay_dump_resources_common.h
```diff
@@ -93,6 +97,43 @@ enum class DumpResourceType
...
+        DumpedRawData index_buffer;
+        DumpedRawData transform_buffer;
+    };
+
+    struct AABBS
+    {
+        DumpedRawData aabb_buffer;
+    };
+
+    std::vector<std::variant<std::monostate, Triangles, AABBS>> build_data;
```
fabian-lunarg, line 121: seeing this the 1st time :) - does the variant contain `std::monostate` so `build_data` is default-constructible?

framework/decode/vulkan_device_address_tracker.h
panos-lunarg: OK let me see if I can pass a reference from `VulkanReplayConsumerBase` down to the other users and avoid the singleton

framework/decode/vulkan_replay_dump_resources_copy_array_of_pointers.h
```diff
@@ -0,0 +1,224 @@
...
+    ASInstanceKHR_ptr out_buffer_base_address;
+    uint              count;
+}
+registers;
+
+void main(void)
+{
+    for (int i = 0; i < registers.count; ++i)
+    {
+        registers.out_buffer_base_address.instance[i].data = registers.array_of_pointers_base_address.pointers[i].instance[0].data;
```
fabian-lunarg, line 61: nice solution for array of pointers

framework/decode/vulkan_replay_dump_resources_common.h
panos-lunarg: Yes so it helps identify uninitialized variants (or at least that's why I used it in this case)

framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
```diff
@@ -1598,6 +1611,70 @@ VkResult DrawCallsDumpingContext::DumpDescriptors(uint64_t qs_index, uint64_t bc
...
+                            desc_binding.desc_type,
+                            desc_set_index,
+                            desc_binding_index,
+                            array_index,
+                            as_info,
+                            options_.dump_resources_dump_build_AS_input_buffers,
+                            DumpResourcesCommandType::kGraphics);
+
+                        auto& new_dumped_as =
+                            std::get<DumpedTopLevelAccelerationStructure>(new_dumped_desc.dumped_resource);
```
fabian-lunarg, line 1639: this 'could' throw [std::bad_variant_access](https://en.cppreference.com/w/cpp/utility/variant/bad_variant_access.html) - alternatively use `std::get_if` and check for nullptr? (safe and avoids try/catch)

fabian-lunarg (approved):
looks good. my biggest concern was the singleton. not too important, just nicer without imo

framework/decode/vulkan_device_address_tracker.h
panos-lunarg: Done. Singletons have been reverted and references to the trackers map is passed down to the users. I did some interactive rebasing and squashed the commits into 1 commits

framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
panos-lunarg: Good catch but I think it is safe because the `emplace_back` in line 1623 above will call the constructor in [framework/decode/vulkan_replay_dump_resources_common.h at line 500-515](https://github.com/LunarG/gfxreconstruct/pull/2398/files#diff-626b29e50e3121c036ae6628a8ff4f5b7d34e26b147ae4f14c96985224763964R500) which will initialize the variant accordingly

wpaul-samsung:
@panos-lunarg It looks like this has been approved.  Is it ready to merge?

PR #2399, opened 2025-09-10
Rework tracking of AccelerationStructures

fabian-lunarg:
our tracking and internal book-keeping of raytracing acceleration-structures
has been based on `VkAccelerationStructureKHR` handles and assumed a 1:1 relation of handles and corresponding `VkDeviceAddress`.

this assumption can be easily violated though, e.g.: 
- multiple `VkAccelerationStructureKHR` over same buffer/offset -> same AS `VkDeviceAddress`
- deleted handles, but buffer-content 'can' still be used e.g. as referenced bottom-lvl structure during top-lvl builds
- so the relation `VkAccelerationStructureKHR` <-> `VkDeviceAddress` is not (1:1) but rather (0-N:1)

this is reworked in this PR: 
`VkAccelerationStructureKHR` treated more like a lightweight 'view' on a resource. 
we fully realize and use the close relationship of buffer-device-addresses and acceleration-addresses.
one can be deduced from the other. we use this when necessary and enforce with asserts.

acceleration-structure tracking is now based on `VkDeviceAddress` and the
lifetime of associated internal assets is now tied to the referenced `VkBuffer`, not the `VkAccelerationStructureKHR`


CMakeLists.txt
```diff
@@ -25,7 +25,7 @@
 # Author: AMD Developer Tools Team
 # Description: CMake script for framework util target
 ###############################################################################
-cmake_minimum_required(VERSION 3.16.3)
+cmake_minimum_required(VERSION 3.16.3...${CMAKE_VERSION})
```
MarkY-LunarG: Is this a typo?

framework/encode/vulkan_handle_wrappers.h
```diff
@@ -481,6 +481,8 @@ struct CommandBufferWrapper : public HandleWrapper<VkCommandBuffer>
         // The offset from the above address to start reading the VkAccelerationStructureInstanceKHR structures
         uint32_t offset;
     };
+
+    // TODO: Fabian, check wtf
```
MarkY-LunarG: Nit: Don't forget to remove.

CMakeLists.txt
fabian-lunarg: thanks for catching. it's an experimental change that slipped in (we had some CI issues all around). going to rebase that out again.

framework/encode/vulkan_handle_wrappers.h
fabian-lunarg: puh, that was close :) btw. I found out what is was in the end, but not related to this PR in the end.

PR #2402, opened 2025-09-11
ci: Fix permissions and MacOS dependencies

antonio-lunarg:
This PR provides two commits to fix two distinct issues in our CI.

1. The test app script has been run with a test user created on the fly because the Vulkan loader worked differently as an admin user, and that prevented the tests from controlling the environment the way it needed to.
Unfortunately, this technique triggers a permission error on GitHub ubuntu runners, but using the current runner user instead makes it happy.

2. After github upgraded macos runners [0] with cmake 4.*, our dependencies in the macos ci script failed to build since they were based on a deprecated cmake minimum version. The fix updates lz4 version which has an updated cmake script, and adds `CMAKE_POLICY_VERSION_MINIMUM=3.5` to the other two dependencies which have no similar cmake script update.

[0] https://github.com/actions/runner-images/issues/12934

### DIFF
--- a/.github/workflows/scripts/build-dependencies-macos.sh
+++ b/.github/workflows/scripts/build-dependencies-macos.sh
@@ -7,7 +7,7 @@ export MACOSX_DEPLOYMENT_TARGET=11.0
 INSTALLDIR="${INSTALLDIR:-"$HOME/deps"}"
 NPROCS="${NPROCS:-$(getconf _NPROCESSORS_ONLN)}"
 ZSTD=1.5.5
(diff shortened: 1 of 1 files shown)


bradgrantham-lunarg:
Can you provide a little more context in the description?  Why change how the tests are run and why upgrade the LZ4 version?

antonio-lunarg:
> @bradgrantham-lunarg  Can you provide a little more context in the description? Why change how the tests are run and why upgrade the LZ4 version?

Updated both commit messages and PR description.

PR #2404, opened 2025-09-15
Fix a few errors and warnings found by clang-check

per-mathisen-arm:
These are fairly random fixes of things I noticed while going through the output of clang-check from clang 22. Its warnings are getting quite useful, although the amount of false positives and plain noise is still very high.

### DIFF
--- a/framework/decode/vulkan_offscreen_swapchain.cpp
+++ b/framework/decode/vulkan_offscreen_swapchain.cpp
@@ -300,12 +300,6 @@ VkResult VulkanOffscreenSwapchain::SignalSemaphoresFence(const VulkanQueueInfo* 
                                                          const VkSemaphore*     signal_semaphores,
                                                          VkFence                fence)
 {
-    uint32_t queue_family_index = default_queue_family_index_;
-    if (queue_info)
-    {
-        queue_family_index = queue_info->family_index;
-    }
-
     VkPipelineStageFlags wait_stage  = VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT;
     VkSubmitInfo         submit_info = { VK_STRUCTURE_TYPE_SUBMIT_INFO };
 
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -6860,8 +6860,6 @@ VkResult VulkanReplayConsumerBase::OverrideCreateShaderModule(
 
     // Replace shader in 'override_info'
     std::unique_ptr<char[]> file_code;
-    const uint32_t* const   orig_code = original_info->pCode;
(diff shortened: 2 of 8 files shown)


framework/decode/vulkan_offscreen_swapchain.cpp
```diff
@@ -300,12 +300,6 @@ VkResult VulkanOffscreenSwapchain::SignalSemaphoresFence(const VulkanQueueInfo*
                                                          const VkSemaphore*     signal_semaphores,
                                                          VkFence                fence)
 {
-    uint32_t queue_family_index = default_queue_family_index_;
-    if (queue_info)
-    {
-        queue_family_index = queue_info->family_index;
-    }
-
```
fabian-lunarg, line 308: could you explain why this isn't required anymore?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -8033,7 +8031,9 @@ VulkanReplayConsumerBase::OverrideQueuePresentKHR(PFN_vkQueuePresentKHR
         // If a swapchain was removed, pNext stucts that reference the swapchain need to be modified as well.
         if (!removed_swapchain_indices_.empty())
         {
-            const VkBaseInStructure* next = reinterpret_cast<const VkBaseInStructure*>(modified_present_info.pNext);
+            VkBaseInStructure* next =
+                reinterpret_cast<VkBaseInStructure*>(const_cast<void*>(modified_present_info.pNext));
+            VkBaseInStructure* prev = reinterpret_cast<VkBaseInStructure*>(&modified_present_info);
```
fabian-lunarg, line 8154: this 'could' be a case where maybe [our existing helpers](https://github.com/LunarG/gfxreconstruct/blob/dev/framework/graphics/vulkan_struct_get_pnext.h) could simplify the iteration.

framework/decode/vulkan_offscreen_swapchain.cpp
fabian-lunarg: ah, I see. we didn't use `queue_family_index`, but assigned to it. then our default compiler-warnings won't brag about _Local variable 'queue_family_index' is only assigned but never accessed_

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: I'll get back to that idea and not mix up with the changes in this PR

fabian-lunarg (approved):

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: I put the changes I had in mind into a dedicated PR https://github.com/LunarG/gfxreconstruct/pull/2434

PR #2405, opened 2025-09-16
Rework tracking of AccelerationStructures (part2)

### LINKED ISSUE #2347: Acceleration structure tracking issues
The current gfxreconstruct implementation tracks built acceleration structures using their VkAccelerationStructureKHR handle. However, Vulkan permits two valid cases that violate the assumption of one-to-one mapping between handle and AS data:

    Multiple handles over the same memory
    An application may create several AS handles with identical creation parameters targeting the same device memory. Each call to vkCreateAccelerationStructureKHR yields a distinct handle, yet once built using any of these handles, they all reference the same underlying AS data.

    Destruction of handles without invalidating memory
    After building, an app may call vkDestroyAccelerationStructureKHR on a handle. This destroys the handle but does not reclaim or invalidate the GPU memory or device-addressed AS data. The AS remains valid and can still be referenced via its vkDeviceAddress e.g. in an instance buffer.

Because gfxreconstruct relies on AS handles to track and store object data, both of these valid Vulkan patterns cause it to:

    Lose track of AS data when a handle is destroyed (even though the AS is still alive).

    Multiplication of internal records when multiple handles alias the same AS, leading to mismatching or missing information.

These tracking errors manifest as missing or mis-reconstructed BLAS/TLAS in portable trace replays and trimmed replays on the same device.

Sample:
https://github.com/ARM-software/tracetooltests/blob/main/src/vulkan_as_6.cpp

Steps to reproduce portable replay issues:
1. Capture vulkan_as_6 - no parameters needed. 
2. Replay with -m rebind

Steps to reproduce same-device trimming issues:
1. Capture `vulkan_as_6 --blas-build-frame --tlas-build-frame`
2. Trim frame 2 or 3
3. Replay trimmed trace with no replay parameters

Steps to reproduce handle aliasing issues:
Add "--blas-alias" parameter to the above vulkan_as_6 launch commands.

fabian-lunarg:
follow-up after https://github.com/LunarG/gfxreconstruct/pull/2399

this PR reworks tracking of acceleration-structures during **capturing**.
this fixes trimming cases where input bottom-lvl acceleration-structures deleted their handles,
but associated buffers are still alive and used for `vkCmdBuildAccelerationStructureKHR`

fixes #2347

also:
replay/trim: better sanity-check, avoid validation-issues wrt. `minAccelerationStructureScratchOffsetAlignment`

### DIFF
--- a/framework/decode/vulkan_address_replacer.cpp
+++ b/framework/decode/vulkan_address_replacer.cpp
@@ -1068,9 +1068,8 @@ void VulkanAddressReplacer::ProcessCmdBuildAccelerationStructuresKHR(
             // scratch-buffer: check for alignment
             scratch_buffer_usable =
                 scratch_buffer_usable &&
-                build_geometry_info.scratchData.deviceAddress %
-                        replay_acceleration_structure_properties_->minAccelerationStructureScratchOffsetAlignment ==
-                    0;
+                (build_geometry_info.scratchData.deviceAddress %
+                 replay_acceleration_structure_properties_->minAccelerationStructureScratchOffsetAlignment) == 0;
 
             if (!as_buffer_usable || !scratch_buffer_usable)
             {
--- a/framework/encode/vulkan_state_writer.cpp
+++ b/framework/encode/vulkan_state_writer.cpp
@@ -1909,6 +1909,7 @@ void VulkanStateWriter::WriteAccelerationStructureBuildState(const gfxrecon::for
     // handle was deleted. we'll require one for rebuilding, so encode calls to create a temporary AS+buffer
     if (as_destroyed)
     {
+        GFXRECON_LOG_WARNING_ONCE("VulkanStateWriter: substituting deleted Acceleration-Structure handles");
         WriteRecreateAccelerationHandle(command);
     }
 


naomipappe:
Hey, that's very important issue you are addressing! We have a `tracetooltest` for it, namely [this one](https://github.com/ARM-software/tracetooltests/blob/main/src/vulkan_as_6.cpp). 

When you run it with argument `--blas-build-frame`, it does the following - builds the BLAS, and then destroys a handle to it, while keeping the contents of the storage buffer intact. We then insert a frame boundary there, so it is considered a different frame from what comes next.

So, you can test your tracking by capturing this program `vulkan_as_6 --blas-build-frame`, and then trimming it from frame 2. Without the fix for state tracking it should crash becase no BLAS is created in the resulting trimmed trace.

fabian-lunarg:
> Hey, that's very important issue you are addressing! We have a `tracetooltest` for it, namely [this one](https://github.com/ARM-software/tracetooltests/blob/main/src/vulkan_as_6.cpp).
> 
> When you run it with argument `--blas-build-frame`, it does the following - builds the BLAS, and then destroys a handle to it, while keeping the contents of the storage buffer intact. We then insert a frame boundary there, so it is considered a different frame from what comes next.
> 
> So, you can test your tracking by capturing this program `vulkan_as_6 --blas-build-frame`, and then trimming it from frame 2. Without the fix for state tracking it should crash becase no BLAS is created in the resulting trimmed trace.

thank you @naomipappe !

I'm indeed using `vulkan_as_6` as test-case and currently iterating on the trimming with `--blas-build-frame`.
it seems not there quite yet, but repro is solid and should be getting there

fabian-lunarg:
wrt. `vulkan_as_6`:

    Steps to reproduce same-device trimming issues:
    Capture vulkan_as_6 --blas-build-frame --tlas-build-frame
    Trim frame 2 or 3
    Replay trimmed trace with no replay parameters

-> works now

antonio-lunarg (approved):

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -1051,11 +1053,24 @@ void VulkanAddressReplacer::ProcessCmdBuildAccelerationStructuresKHR(
...
-                                         (scratch_buffer_info->usage & VK_BUFFER_USAGE_STORAGE_BUFFER_BIT);
+            uint32_t scratch_size = build_geometry_info.mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_BUILD_KHR
+                                        ? build_size_info.buildScratchSize
+                                        : build_size_info.updateScratchSize;
+
+            // scratch-buffer: check for nullptr and size
+            bool scratch_buffer_usable = scratch_buffer_info != nullptr && scratch_buffer_info->size >= scratch_size;
+
+            // scratch-buffer: check usage-flags
+            auto scratch_usage_flags = VK_BUFFER_USAGE_STORAGE_BUFFER_BIT | VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT;
```
MarkY-LunarG, line 1064: What about just a standard uniform buffer too?

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -1051,11 +1053,24 @@ void VulkanAddressReplacer::ProcessCmdBuildAccelerationStructuresKHR(
...
+            scratch_buffer_usable =
+                scratch_buffer_usable && (scratch_buffer_info->usage & scratch_usage_flags) == scratch_usage_flags;
+
+            // scratch-buffer: check for alignment
+            scratch_buffer_usable =
+                scratch_buffer_usable &&
+                build_geometry_info.scratchData.deviceAddress %
```
MarkY-LunarG: Nit add paranthesis around the % portion to be safe.

framework/encode/vulkan_state_writer.cpp
```diff
@@ -1813,6 +1901,17 @@ void VulkanStateWriter::WriteAccelerationStructureBuildState(const gfxrecon::for
...
     UpdateAddresses(command);
+
+    // check for deleted handles, create replacements
+    bool as_destroyed = vulkan_wrappers::GetWrappedId<vulkan_wrappers::AccelerationStructureKHRWrapper>(
+                            command.geometry_info.dstAccelerationStructure, false) == format::kNullHandleId;
+
+    // handle was deleted. we'll require one for rebuilding, so encode calls to create a temporary AS+buffer
+    if (as_destroyed)
+    {
+        WriteRecreateAccelerationHandle(command);
```
MarkY-LunarG, line 1913: I think this might deserve an info message.  Just so we know (oh by the way, this occurred).  Just because I can see it being important debugging tip if we see issues.

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: not required. scratch-buffers are mandatory storage-buffers with device-address

framework/encode/vulkan_state_writer.cpp
fabian-lunarg: ok, will restrict to a log_once, because if this pattern occurs, it will occur frequently. this seems to be the case for UE5.

framework/decode/vulkan_address_replacer.cpp
MarkY-LunarG: Thanks for verifying.

framework/encode/vulkan_state_writer.cpp
MarkY-LunarG: Sounds good!

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: done

framework/encode/vulkan_state_writer.cpp
fabian-lunarg: done

PR #2407, opened 2025-09-17
Add `--present-mode` option

marius-pelegrin-arm:
Add an option to adjust the swapchain's present mode used at replay time.

Present mode may need to be modified for performance measurement reasons or for portability reasons.
This commit allows to set a particular present mode to use.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -7520,9 +7520,6 @@ VkResult VulkanReplayConsumerBase::OverrideCreateSwapchainKHR(
                     break;
             }
 
-            VulkanPhysicalDeviceInfo* physical_device_info =
-                object_info_table_->GetVkPhysicalDeviceInfo(device_info->parent_id);
-
             auto instance_table = GetInstanceTable(physical_device_info->parent);
 
             uint32_t present_mode_count;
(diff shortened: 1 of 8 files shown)


fabian-lunarg (approved):
makes sense, looks good

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7432,6 +7432,65 @@ VkResult VulkanReplayConsumerBase::OverrideCreateSwapchainKHR(
...
+                physical_device_info->handle, modified_create_info.surface, &present_mode_count, nullptr);
+
+            std::vector<VkPresentModeKHR> supported_present_modes(present_mode_count);
+            instance_table->GetPhysicalDeviceSurfacePresentModesKHR(physical_device_info->handle,
+                                                                    modified_create_info.surface,
+                                                                    &present_mode_count,
+                                                                    supported_present_modes.data());
+
+            bool is_supported = false;
+            for (VkPresentModeKHR supported_present_mode : supported_present_modes)
```
MarkY-LunarG: Nit: Can't you just use "find" on the vector?

fabian-lunarg: done, replaced with `std::find`

MarkY-LunarG (approved):

PR #2412, opened 2025-09-18
VDR: Handle dynamic rendering from secondaries

panos-lunarg:
Secondary command buffers can call vkCmdBeginRendering. This was not taken into consideration which caused crashes. This PR should address this issue.

### DIFF
--- a/framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
+++ b/framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
@@ -61,7 +61,8 @@ DrawCallsDumpingContext::DrawCallsDumpingContext(const CommandIndices*          
     delegate_(delegate), options_(options), compressor_(compressor), current_render_pass_type_(kNone),
     aux_command_buffer_(VK_NULL_HANDLE), aux_fence_(VK_NULL_HANDLE),
     command_buffer_level_(DumpResourcesCommandBufferLevel::kPrimary), device_table_(nullptr), instance_table_(nullptr),
-    object_info_table_(object_info_table), replay_device_phys_mem_props_(nullptr)
+    object_info_table_(object_info_table),
+    replay_device_phys_mem_props_(nullptr), secondary_with_dynamic_rendering_{ false }
 {
     if (draw_indices != nullptr)
     {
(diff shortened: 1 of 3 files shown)


framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
```diff
@@ -3485,6 +3482,84 @@ uint32_t DrawCallsDumpingContext::RecaclulateCommandBuffers()
...
+    {
+        uint32_t sec_rts_copied = 0;
+        for (auto sec_it = rp_secondary.begin(); sec_it < rp_secondary.end(); ++sec_it)
+        {
+            if (prim_it->empty())
+            {
+                if (!sec_it->empty())
+                {
+                    rp_primary.erase(prim_it);
+                    rp_primary.insert(prim_it, *sec_it);
```
bradgrantham-lunarg: `prim_it` is invalidated by the `erase`.  https://en.cppreference.com/w/cpp/container/vector/erase - "Iterators (including the [end()](https://en.cppreference.com/w/cpp/container/vector/end.html) iterator) and references to the elements **at** or after the point of the erase are invalidated."  Same for `insert`.  I think you need to assign the return value to `prim_it`, something like:
```
                    prim_it = rp_primary.erase(prim_it);
                    prim_it = rp_primary.insert(prim_it, *sec_it);
```
If you just mean to replace the element, can you use `*prim_it = *sec_it`;?  Otherwise I think you're potentially memmoving the entire rest of `rp_primary `1 element backward when you erase, and then memmoving it forward 1 when you insert, although I understand that having it as `erase`+`insert` makes sense for consistency with the `else` clause below.

framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
```diff
@@ -3485,6 +3482,84 @@ uint32_t DrawCallsDumpingContext::RecaclulateCommandBuffers()
...
+                    ++sec_rts_copied;
+                }
+            }
+            else
+            {
+                if (!sec_it->empty())
+                {
+                    if ((*sec_it)[0] < (*prim_it)[0])
+                    {
+                        rp_primary.insert(prim_it, *sec_it);
```
bradgrantham-lunarg: Same thought about invalidating the iterator as above

framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
panos-lunarg: Good catch! I pushed some more changes which should address this

dfriederich:
I was not able to reproduce the issue anymore, but we did change in sokatoa so we are not fetching dispatch nodes for the main tree, which is I think what was different before. So works :-)

PR #2418, opened 2025-09-24
VDR: Remove unnnececary error message

panos-lunarg:
There was a forgotten error message related to whether an image can be dumped or not left by mistake. This error should be removed as the case it generates as an error is already treated as a warning inside IsImageDumpable.


panos-lunarg:
This is included in https://github.com/LunarG/gfxreconstruct/pull/2419

PR #2419, opened 2025-09-24
VDR: Report images that cannot be dumped in output json

panos-lunarg:
There are two cases where an image otherwise marked for dumping will not get dumped:

1. Format is not supported on the replay device. A capture taken on a different vulkan implementation can contain VkImages which are not supported on the replay device. When replaying we re-create these images without warnings but dumping them can be problematic and cause crashes.

2. Multisampled images that cannot be resolved. Format features of a multisampled image must contain VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT in order to do a vkCmdResolveImage. We do not dump images (nor dump them in the state setup block) that cannot be resolved directly.

Instead of ignoring the above images, this PR will add a new entry in the output json for each image along with the reason of why the image couldn't be dumped.

### DIFF
--- a/framework/decode/vulkan_replay_dump_resources_common.h
+++ b/framework/decode/vulkan_replay_dump_resources_common.h
@@ -95,6 +95,13 @@ enum class DumpResourceType
     kDispatchTraceRaysInlineUniformBufferDescriptor,
 };
 
+enum ImageDumpResult
+{
+    kCanDump,
+    kCanNotResolve,
+    kFormatNotSupported
+};
+
 struct DumpedFile
 {
     DumpedFile() = default;
@@ -129,7 +136,9 @@ struct DumpedImage
 {
     DumpedImage() = default;
 
-    DumpedImage(const VulkanImageInfo* i_f) : image_info(i_f), scaling_failed(false), dumped_raw(false) {}
+    DumpedImage(const VulkanImageInfo* i_f, ImageDumpResult cd) :
+        image_info(i_f), scaling_failed(false), dumped_raw(false), can_dump(cd)
+    {}
 
     const VulkanImageInfo* image_info{ nullptr };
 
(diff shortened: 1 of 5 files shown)


framework/decode/vulkan_replay_dump_resources_common.h
```diff
@@ -94,6 +94,13 @@ enum class DumpResourceType
     kDispatchTraceRaysInlineUniformBufferDescriptor,
 };
 
+enum CanImageBeDumped
```
bradgrantham-lunarg: I think this is the name for a calculation, not a status, can this be ImageDumpResult?

bradgrantham-lunarg (approved):
Looks good, just some nits about the naming

framework/decode/vulkan_replay_dump_resources_common.h
panos-lunarg: Fixed

framework/decode/vulkan_replay_dump_resources_delegate.cpp
```diff
@@ -802,25 +802,39 @@ void DefaultVulkanDumpResourcesDelegate::GenerateOutputJsonDrawCallInfo(
...
+                                                                options_.dump_resources_dump_separate_alpha,
+                                                                image.dumped_raw);
+                }
+            }
+        }
+        else
+        {
+            if (image.can_dump == ImageDumpResult::kCanNotResolve)
+            {
+                rt_entry["CouldNotDump"] = "CouldNotResolve";
```
dfriederich: nit, all other attribute use lowercase names. Also "couldNotDump" sounds more like the value the error attribute should have, than the name of the attribute.
Maybe 'error', 'failure', 'dumpFailure' or ... :-)
```
rt_entry["dumpFailure"] = "CouldNotResolve";
```

PR #2427, opened 2025-09-25
Zlib compressor

panos-lunarg:
It is possible, espessialy for small buffers, that the compressed output to be larger than the uncompressed. Current implementation of ZlibCompressor::Compress does not take this into account resulting in corrupted compressed buffers


panos-lunarg:
> If I understand it correctly, `deflate()` might return a value different than `Z_STREAM_END`, in which case you fill the data with zeroes and continue to compress until it's happy. LGTM but I would add some comments explaining the problem and the rationale for your changes.

The problem here is the code here assumes that allocating an out buffer with the same size as the uncompressed data should be enough but it turns out that this is not always the case. Sometimes the compressed output can be larger than the uncompressed input. So if `deflate` returns something other than `Z_STREAM_END` (perhaps I need to check what it actually returns) means that the out buffer has ran out of space. So we allocate some more (`CHUNK` bytes) and tell `deflate` to continue where it stopped. Adding some comments is a very good idea. I'll address that

MarkY-LunarG:
I fail to understand why we compress this if the compression size is larger.  Doesn't that kind of go against the whole point?

panos-lunarg:
> I fail to understand why we compress this if the compression size is larger. Doesn't that kind of go against the whole point?

Well this is the compressor itself. The caller can decide what to do in these cases. For example the vulkan capture manager and state writer compare the 2 sizes and write to the file whichever is smaller

dfriederich:
> I fail to understand why we compress this if the compression size is larger. Doesn't that kind of go against the whole point?

The dump resources extraction API only reports the format used for all bin files, not on a file per file bases. Hence the issue is that without this fix, gfxr replay creates corrupt gzip files for dump resources.

(For the gfxr file, the using code decides not to use the compressed (before this fix invalid code), but instead uses the uncompressed data, hence it does not suffer from the same issue.)

PR #2428, opened 2025-09-25
VDR: Two small fixes

panos-lunarg:
1. Fixed logic in IsBlitSupported() regarding integer formats
signedness. Also added missing [S|U]SCALED.
2. Formats B8G8R8_SRGB and B8G8R8_UNORM are removed from the logic that
decides which format to use when dumping images into bmp/png as they
might not be supported. Instead we stick to VK_FORMAT_B8G8R8A8_UNORM and
VK_FORMAT_B8G8R8A8_SRGB which are mandatory from the spec.

### DIFF
--- a/framework/decode/vulkan_replay_dump_resources_common.cpp
+++ b/framework/decode/vulkan_replay_dump_resources_common.cpp
@@ -108,7 +108,7 @@ static VkFormat ChooseDestinationImageFormat(VkFormat format)
 
     if (vkuFormatIsSRGB(format))
     {
-        dst_format = vkuFormatHasAlpha(format) ? VK_FORMAT_B8G8R8A8_SRGB : VK_FORMAT_B8G8R8_SRGB;
+        dst_format = VK_FORMAT_B8G8R8A8_SRGB;
     }
     else if (vkuFormatIsDepthOrStencil(format))
     {
@@ -118,7 +118,7 @@ static VkFormat ChooseDestinationImageFormat(VkFormat format)
     }
     else
     {
-        dst_format = vkuFormatHasAlpha(format) ? VK_FORMAT_B8G8R8A8_UNORM : VK_FORMAT_B8G8R8_UNORM;
+        dst_format = VK_FORMAT_B8G8R8A8_UNORM;
     }
 
     return dst_format;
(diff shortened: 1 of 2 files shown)


framework/decode/vulkan_replay_dump_resources_common.cpp
```diff
@@ -108,7 +108,7 @@ static VkFormat ChooseDestinationImageFormat(VkFormat format)
 
     if (vkuFormatIsSRGB(format))
     {
-        dst_format = vkuFormatHasAlpha(format) ? VK_FORMAT_B8G8R8A8_SRGB : VK_FORMAT_B8G8R8_SRGB;
+        dst_format = VK_FORMAT_B8G8R8A8_SRGB;
```
bradgrantham-lunarg, line 111: This value is used at line 453 and is then passed on to `dst_format` and through to `dumped_image.dumped_format`.  Does that mean that some images that have no alpha will end up dumped with 0 alpha?

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_replay_dump_resources_common.cpp
panos-lunarg: > Does that mean that some images that have no alpha will end up dumped with 0 alpha?

I don't think this can happen.
Images without alpha will be converted with `vkCmdBlitImage` into the R8G8B8A format that is chosen here and, if I understand this correctly, according to spec they will get an alpha value of 1

bradgrantham-lunarg: Is information lost?  The image didn't have alpha at all, but saving an RGBA image loses that and it's not clear whether the image had an alpha of 1.0 or no alpha.

panos-lunarg: Well losing information when dumping as bmp/png should be expected, no?
We convert every image (format) that can be converted by the implementation into those formats so that we can dump them in image files. This is more for convenience (you can see the dumped images without any effort). The output json should list the image's original format.
`--dump-resources-dump-raw-images` should dump images verbatim so it can be used when not losing information is desired

bradgrantham-lunarg: I mean that you lose the information that the image was RGB in memory if the image lands in the file as RGBA.

panos-lunarg: The output json should list the image with its original format so this should be evident

bradgrantham-lunarg: Excellent, thanks!

PR #2429, opened 2025-09-25
Fix color-distortion when capture uses HDR color-format

fabian-lunarg:
This PR fixes issues with captures using swapchain color-format `VK_FORMAT_A2B10G10R10_UNORM_PACK32`.
even when this format is not supported during **replay**,
GFXR would allow using it, for created `VkImageView`s onto the swapchain-images. 

symptom were color-sampling distortions due to different data-layouts:

distortion

fixed

- Add `OverrideGetPhysicalDeviceSurfaceFormatsKHR`, keep track of supported formats
- fallback to safe format if we are sure a configuration is not supported
- adjust image-views for swapchain-images if necessary

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -3904,7 +3904,8 @@ VkResult VulkanReplayConsumerBase::OverrideGetEventStatus(PFN_vkGetEventStatus  
               ((original_result == VK_EVENT_RESET) && (result == VK_EVENT_SET))) &&
              (++retries <= kMaxEventStatusRetries));
 
-    return result;
+//    return result;
+    return original_result;// tmp: silence log-spam
 }
 
 VkResult VulkanReplayConsumerBase::OverrideGetQueryPoolResults(PFN_vkGetQueryPoolResults  func,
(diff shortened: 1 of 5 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7379,6 +7401,26 @@ VkResult VulkanReplayConsumerBase::OverrideCreateSwapchainKHR(
...
+        // check if 'replay_create_info->imageFormat' is supported,
+        // do nothing if we got no information about available surfaces
+        bool surface_format_supported = !physical_device_info->surface_formats;
+        if (physical_device_info->surface_formats)
+        {
+            for (const auto& supported : *physical_device_info->surface_formats)
+            {
+                if (supported.format == replay_create_info->imageFormat)
+                {
+                    surface_format_supported = true;
```
MarkY-LunarG, line 7442: break to keep from looping.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7379,6 +7401,26 @@ VkResult VulkanReplayConsumerBase::OverrideCreateSwapchainKHR(
                 supported_extension_iterator == instance_info->util_info.enabled_extensions.end();
         }
 
+        // check if 'replay_create_info->imageFormat' is supported,
+        // do nothing if we got no information about available surfaces
+        bool surface_format_supported = !physical_device_info->surface_formats;
```
MarkY-LunarG, line 7435: Isn't this setting it to true if formats exist, but then it might not be found in the following loop so there's no "false" by default.  So maybe set to false instead?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -9892,6 +9927,11 @@ VkResult VulkanReplayConsumerBase::OverrideCreateImageView(
             modified_create_info.format = VK_FORMAT_R8G8B8A8_UNORM;
         }
     }
+    else if (img_info->is_swapchain_image && img_info->format != modified_create_info.format)
+    {
+        // for swapchain-images set image-view to actual format, avoid issues with distorted HDR-colors
+        modified_create_info.format = img_info->format;
```
MarkY-LunarG, line 9966: Maybe log a warning_once here?

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: that's what I had there :)
the idea then was to _do nothing if we got no information about available surfaces_ 
so when we init with `!physical_device_info->surface_formats` this won't change anything when `surface_formats` 
is not available.
-> this happens when it was not queried (unlikely!?) ~~or when we trim (this needs to be added/fixed)~~
-> trimming looks also fine

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: done

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: done

MarkY-LunarG (approved):

PR #2437, opened 2025-09-30
Uniform allocation callbacks behavior in `VulkanRebindAllocator`

marius-pelegrin-arm:
When using the rebind allocator, VMA handles most allocations with the allocation callbacks provided when creating the allocator.

Currently, allocations callbacks in the rebind allocator are either:
- The ones from the VMA allocator, constant during the entire replay
- The ones fed by the `VulkanReplayConsumer`
- Replaced by `nullptr`

With this commit I propose to make everything use the callbacks of the VMA allocator.


fabian-lunarg:
fyi @locke-lunarg

fabian-lunarg (approved):
agree

locke-lunarg (approved):

PR #2438, opened 2025-09-30
Fix dark colors when capture uses an SRGB swapchain-format

### LINKED ISSUE #2414: add option for RGBA -> BGRA swapchain
Many times a capture will replay with colors swizzled because the swapchain uses the wrong format. Would be nice to have an option to correct this.

fabian-lunarg:
follow-up to https://github.com/LunarG/gfxreconstruct/pull/2429

during replay, when we cannot support a requested swapchain VkFormat, we fall back to a 'safe' format.
later during runtime, when `VkImageView`s onto swapchain images are created,
we again have to apply corrections.
in both cases we also need to account for SRGB formats.

e.g. `VK_FORMAT_R8G8B8A8_UNORM` -> `VK_FORMAT_B8G8R8A8_UNORM`
and `VK_FORMAT_R8G8B8A8_SRGB` -> `VK_FORMAT_B8G8R8A8_SRGB`

also: 
during swapchain creation we check for usage of `VK_KHR_swapchain_mutable_format`
apply corrections, if necessary.

fixes #2414

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -58,6 +58,7 @@
 #include "generated/generated_vulkan_enum_to_string.h"
 #include "util/to_string.h"
 #include "vulkan/vulkan_core.h"
+#include "Vulkan-Utility-Libraries/vk_format_utils.h"
 
 #include <algorithm>
 #include <cstddef>
@@ -9993,8 +9994,9 @@ VkResult VulkanReplayConsumerBase::OverrideCreateImageView(
     }
     else if (img_info->is_swapchain_image && img_info->format != modified_create_info.format)
     {
-        // for swapchain-images set image-view to actual format, avoid issues with distorted HDR-colors
-        modified_create_info.format = img_info->format;
+        // for swapchain-images set image-view to a fallback format, avoid issues with distorted HDR/SRGB colors
+        modified_create_info.format =
+            vkuFormatIsSRGB(modified_create_info.format) ? VK_FORMAT_B8G8R8A8_SRGB : VK_FORMAT_B8G8R8A8_UNORM;
     }
 
     VkResult result = func(device, &modified_create_info, allocator, out_view);
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -7453,18 +7453,12 @@ VkResult VulkanReplayConsumerBase::OverrideCreateSwapchainKHR(
             // handle VK_KHR_swapchain_mutable_format
             if (modified_create_info.flags & VK_SWAPCHAIN_CREATE_MUTABLE_FORMAT_BIT_KHR)
             {
-                bool mutable_srgb = false;
                 if (auto* format_list_pnext =
                         graphics::vulkan_struct_get_pnext<VkImageFormatListCreateInfo>(&modified_create_info))
                 {
-                    for (uint32_t i = 0; i < format_list_pnext->viewFormatCount; ++i)
-                    {
-                        if (vkuFormatIsSRGB(format_list_pnext->pViewFormats[0]))
-                        {
-                            mutable_srgb = true;
-                            break;
-                        }
-                    }
+                    const auto* view_formats     = format_list_pnext->pViewFormats;
+                    const auto* view_formats_end = view_formats + format_list_pnext->viewFormatCount;
+                    bool        mutable_srgb     = std::any_of(view_formats, view_formats_end, vkuFormatIsSRGB);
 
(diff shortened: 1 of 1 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7445,8 +7450,33 @@ VkResult VulkanReplayConsumerBase::OverrideCreateSwapchainKHR(
...
+                    for (uint32_t i = 0; i < format_list_pnext->viewFormatCount; ++i)
+                    {
+                        if (vkuFormatIsSRGB(format_list_pnext->pViewFormats[0]))
+                        {
+                            mutable_srgb = true;
+                            break;
+                        }
+                    }
+
+                    // overwrite existing VkImageFormatListCreateInfo
```
MarkY-LunarG, line 7463: Nit, this should be a warning (maybe just once) that you are overwritting the swapchain format.

MarkY-LunarG (approved):

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7445,8 +7450,33 @@ VkResult VulkanReplayConsumerBase::OverrideCreateSwapchainKHR(
...
+            // handle VK_KHR_swapchain_mutable_format
+            if (modified_create_info.flags & VK_SWAPCHAIN_CREATE_MUTABLE_FORMAT_BIT_KHR)
+            {
+                bool mutable_srgb = false;
+                if (auto* format_list_pnext =
+                        graphics::vulkan_struct_get_pnext<VkImageFormatListCreateInfo>(&modified_create_info))
+                {
+                    for (uint32_t i = 0; i < format_list_pnext->viewFormatCount; ++i)
+                    {
+                        if (vkuFormatIsSRGB(format_list_pnext->pViewFormats[0]))
```
bradgrantham-lunarg: Should this be checking pViewFormats[i]?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7445,8 +7450,33 @@ VkResult VulkanReplayConsumerBase::OverrideCreateSwapchainKHR(
 
         if (!surface_format_supported)
         {
+            // handle VK_KHR_swapchain_mutable_format
+            if (modified_create_info.flags & VK_SWAPCHAIN_CREATE_MUTABLE_FORMAT_BIT_KHR)
+            {
+                bool mutable_srgb = false;
+                if (auto* format_list_pnext =
```
bradgrantham-lunarg, line 7456: Could this be `any_of`?  Something like
```
bool mutable_srgb = std::any_of(format_list_pnext->pViewFormats, format_list_pnext->pViewFormats + format_list_pnext->viewFormatCount, vkuFormatIsSRGB);

```

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: yes, thanks for catching!

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: already in place. couple lines below, we log "Replay adjusted unsupported surface imageFormat (%d) to VK_FORMAT_B8G8R8A8_UNORM". 

this added block is dealing with `VK_KHR_swapchain_mutable_format`, mostly in order to avoid validation-errors:
```
[gfxrecon] ERROR - DEBUG MESSENGER: Frame #1: VUID-VkSwapchainCreateInfoKHR-flags-03168: vkCreateSwapchainKHR(): pCreateInfo->flags includes VK_SWAPCHAIN_CREATE_MUTABLE_FORMAT_BIT_KHR but none of the elements of the pViewFormats member of VkImageFormatListCreateInfo match imageFormat (VK_FORMAT_B8G8R8A8_UNORM).
The Vulkan spec states: If flags contains VK_SWAPCHAIN_CREATE_MUTABLE_FORMAT_BIT_KHR then the pNext chain must include a VkImageFormatListCreateInfo structure with a viewFormatCount greater than zero and pViewFormats must have an element equal to imageFormat
```
so here, we're not really swapping the format, rather adjust the list of compatible/mutable formats that `VkImageView`s created onto the swapchain-images 'could' have.

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: yes, much nicer. done

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: -> now using `std::any_of`

PR #2440, opened 2025-10-02
Skip unsupported reserved meta-data block

JerryAMD:


framework/decode/file_processor.cpp
```diff
@@ -2464,6 +2464,9 @@ bool FileProcessor::ProcessMetaData(BlockBuffer& block_buffer, format::MetaDataI
             GFXRECON_LOG_WARNING("Skipping unrecognized meta-data block with type %" PRIu16, meta_data_type);
         }
         GFXRECON_CHECK_CONVERSION_DATA_LOSS(size_t, block_header.size);
+
+        // Skip unsupported meta-data block.
+        success = true;
```
bradgrantham-lunarg, line 2469: @jzulauf-lunarg you've refactored a bunch of this.  There used to be `success = SkipBytes(static_cast<size_t>(block_header.size) - sizeof(meta_data_id));` here.  Is the skip handled in the refactor and we only need `success = true` here now?

jzulauf-lunarg: Yes.  That's correct.  I failed to note the function's pessimism ("bool success = false;") 

I have a branch I have updated the other two locations where SkipBytes was removed to guarantee identical operation (even though, I believe success is already true in those cases). Setting to true clarifies the operational assumption that the conditions do not constitute a failure state.

jzulauf-lunarg:
@JerryAMD  -- Recommend closing in favor of https://github.com/LunarG/gfxreconstruct/pull/2442  which contains the original fix, plus two other sites with similar usage.

PR #2443, opened 2025-10-02
Fix gfxrecon-convert null dereference

dgraves:
Add null check for D3D12 enum pointers in gfxrecon-convert special case.

### DIFF
--- a/framework/generated/dx12_generators/dx12_json_consumer_body_generator.py
+++ b/framework/generated/dx12_generators/dx12_json_consumer_body_generator.py
@@ -159,7 +159,22 @@ class Dx12JsonConsumerBodyGenerator(Dx12JsonConsumerHeaderGenerator, Dx12JsonCom
             # Generate a correct FieldToJson for each argument:
             for parameter in method_info['parameters']:
                 value = self.get_value_info(parameter)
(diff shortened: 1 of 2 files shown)


framework/generated/dx12_generators/dx12_json_consumer_body_generator.py
```diff
@@ -159,7 +159,22 @@ def make_consumer_method_body(self, class_name, method_info, return_type, return
             # Generate a correct FieldToJson for each argument:
             for parameter in method_info['parameters']:
                 value = self.get_value_info(parameter)
-                code += "    " + self.make_field_to_json("args", value, "options") + "\n"
+
+                function_call = self.make_field_to_json("args", value, "options") + "\n"
+
+                ## Special case for pointers to flag sets defined by enums:
+                ## (easier than having pointer decoder versions of each flagset type's FieldToString)
+                if value.is_pointer and function_call.startswith("FieldToJson_"):
```
dgraves, line 167: This comment and conditional expression were copied from `make_field_to_json`, where the special case is initially applied to the function call formatting.

bradgrantham-lunarg (approved):

PR #2449, opened 2025-10-07
Const clean decoder and consumer interfaces

jzulauf-lunarg:
Prevent decoders and consumers from modifying data passed down from the parser.


fabian-lunarg (approved):
nice

bradgrantham-lunarg (approved):

Issue #2396, opened 2025-09-09
[gfxrecon-info] Wrong output when multiple vkCreateInstance calls in trace


laurentiu-erhan-arm:
When multiple instances (vkCreateInstance calls) are in a trace, gfxrecon-info will report "Vulkan application info" only for the latest instance, or sometimes even a combination of the instances if some fields are null.

Example/to replicate: run the tracetoolstest scenario vulkan_multiinstance (https://github.com/ARM-software/tracetooltests/blob/main/src/vulkan_multiinstance.cpp)

Relevant output of gfxrecon-info

> Vulkan application info:
>         Application name: vulkan_multiinstance_2
>         Application version: 4194304
>         Engine name: testEngine
>         Engine version: 4194304
>         Target API version: 4198400 (1.1.0)
> 

Relevant gfxrecon-convert calls 

> {"index":3,"function":{"name":"vkCreateInstance","thread":1,"return":"VK_SUCCESS","args":{"pCreateInfo":{"sType":"VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO","flags":"0x00000000","pApplicationInfo":{"sType":"VK_STRUCTURE_TYPE_APPLICATION_INFO","pApplicationName":"vulkan_multiinstance_1","applicationVersion":4194304,"pEngineName":"testEngine","engineVersion":4194304,"apiVersion":4198400,"pNext":null},"enabledLayerCount":0,"ppEnabledLayerNames":null,"enabledExtensionCount":3,"ppEnabledExtensionNames":["VK_EXT_debug_utils","VK_KHR_xcb_surface","VK_KHR_surface"],"pNext":{"sType":"VK_STRUCTURE_TYPE_DEBUG_UTILS_MESSENGER_CREATE_INFO_EXT","flags":"0x00000000","messageSeverity":"0x00001110","messageType":"0x00000007","pfnUserCallback":"0x60df5547aaa6","pUserData":"0x0","pNext":null}},"pAllocator":null,"pInstance":1}}}
[...]

PR #2451, opened 2025-10-08
Update GFXRecon-Info with a "--verbose" output option

MarkY-LunarG:
Add a new "--verbose" option that allows more detail in the info output.  This includes:

- Output for every instance created including any application/engine names associated with each
- Output enabled instance and device extensions
- Per VkDevice allocation and pipeline information.
- Improved OpenXR output (including fixing the version output which could have been wrong)

This started as a direct change for #2396, but morphed into an improved output.   I still might have to do what the issue requests if this does not work for them, but I find this output more useful than just the old standard output.

### DIFF
--- a/framework/decode/dx12_stats_consumer.h
+++ b/framework/decode/dx12_stats_consumer.h
@@ -131,6 +131,9 @@ class Dx12StatsConsumer : public Dx12Consumer
                                                 HRESULT                                           return_value,
                                                 StructPointerDecoder<Decoded_DXGI_ADAPTER_DESC3>* pDesc)
     {
+        GFXRECON_UNREFERENCED_PARAMETER(call_info);
+        GFXRECON_UNREFERENCED_PARAMETER(object_id);
+        GFXRECON_UNREFERENCED_PARAMETER(return_value);
         format::DxgiAdapterDesc new_adapter = {};
         CopyAdapterDesc(new_adapter, pDesc);
         InsertAdapter(new_adapter, app_get_desc_adapters);
@@ -142,6 +145,9 @@ class Dx12StatsConsumer : public Dx12Consumer
                                    HRESULT                                                             return_value,
                                    gfxrecon::decode::StructPointerDecoder<Decoded_DXGI_ADAPTER_DESC2>* pDesc)
     {
+        GFXRECON_UNREFERENCED_PARAMETER(call_info);
+        GFXRECON_UNREFERENCED_PARAMETER(object_id);
+        GFXRECON_UNREFERENCED_PARAMETER(return_value);
         format::DxgiAdapterDesc new_adapter = {};
         CopyAdapterDesc(new_adapter, pDesc);
         InsertAdapter(new_adapter, app_get_desc_adapters);
@@ -153,6 +159,9 @@ class Dx12StatsConsumer : public Dx12Consumer
                                    HRESULT                                                             return_value,
                                    gfxrecon::decode::StructPointerDecoder<Decoded_DXGI_ADAPTER_DESC1>* pDesc)
     {
+        GFXRECON_UNREFERENCED_PARAMETER(call_info);
+        GFXRECON_UNREFERENCED_PARAMETER(object_id);
+        GFXRECON_UNREFERENCED_PARAMETER(return_value);
         format::DxgiAdapterDesc new_adapter = {};
         CopyAdapterDesc(new_adapter, pDesc);
         InsertAdapter(new_adapter, app_get_desc_adapters);
@@ -163,14 +172,22 @@ class Dx12StatsConsumer : public Dx12Consumer
                                               HRESULT                              return_value,
                                               gfxrecon::decode::StructPointerDecoder<Decoded_DXGI_ADAPTER_DESC>* pDesc)
     {
+        GFXRECON_UNREFERENCED_PARAMETER(call_info);
+        GFXRECON_UNREFERENCED_PARAMETER(object_id);
+        GFXRECON_UNREFERENCED_PARAMETER(return_value);
         format::DxgiAdapterDesc new_adapter = {};
         CopyAdapterDesc(new_adapter, pDesc);
         InsertAdapter(new_adapter, app_get_desc_adapters);
     }
 
-    std::string GetSwapchainDimensions()
+    std::string GetSwapchainDimensionsString()
     {
         return std::to_string(swapchain_width_) + 'x' + std::to_string(swapchain_height_);
+    }
+
+    std::pair<UINT, UINT> GetSwapchainDimensions()
+    {
+        return {swapchain_width_, swapchain_height_};
     }
 
     virtual void Process_IDXGIFactory_CreateSwapChain(const ApiCallInfo&                                  call_info,
@@ -180,6 +197,10 @@ class Dx12StatsConsumer : public Dx12Consumer
                                                       StructPointerDecoder<Decoded_DXGI_SWAP_CHAIN_DESC>* pDesc,
                                                       HandlePointerDecoder<IDXGISwapChain*>*              ppSwapChain)
     {
+        GFXRECON_UNREFERENCED_PARAMETER(call_info);
+        GFXRECON_UNREFERENCED_PARAMETER(object_id);
+        GFXRECON_UNREFERENCED_PARAMETER(return_value);
+        GFXRECON_UNREFERENCED_PARAMETER(pDevice);
         if (!ppSwapChain->IsNull())
         {
             swapchain_id_ = *ppSwapChain->GetPointer();
@@ -206,6 +227,13 @@ class Dx12StatsConsumer : public Dx12Consumer
         format::HandleId                                               pRestrictToOutput,
         HandlePointerDecoder<IDXGISwapChain1*>*                        ppSwapChain)
     {
+        GFXRECON_UNREFERENCED_PARAMETER(call_info);
+        GFXRECON_UNREFERENCED_PARAMETER(object_id);
+        GFXRECON_UNREFERENCED_PARAMETER(return_value);
+        GFXRECON_UNREFERENCED_PARAMETER(pDevice);
+        GFXRECON_UNREFERENCED_PARAMETER(hWnd);
+        GFXRECON_UNREFERENCED_PARAMETER(pFullscreenDesc);
+        GFXRECON_UNREFERENCED_PARAMETER(pRestrictToOutput);
         if (!ppSwapChain->IsNull())
         {
             swapchain_id_     = *ppSwapChain->GetPointer();
@@ -229,6 +257,11 @@ class Dx12StatsConsumer : public Dx12Consumer
                                                       DXGI_FORMAT        NewFormat,
                                                       UINT               SwapChainFlags)
     {
+        GFXRECON_UNREFERENCED_PARAMETER(call_info);
+        GFXRECON_UNREFERENCED_PARAMETER(return_value);
+        GFXRECON_UNREFERENCED_PARAMETER(BufferCount);
+        GFXRECON_UNREFERENCED_PARAMETER(NewFormat);
+        GFXRECON_UNREFERENCED_PARAMETER(SwapChainFlags);
         if (swapchain_id_ == object_id)
         {
             swapchain_width_  = Width;
@@ -245,6 +278,11 @@ class Dx12StatsConsumer : public Dx12Consumer
         UINT                                                                              NumPostbuildInfoDescs,
         StructPointerDecoder<Decoded_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_DESC>* pPostbuildInfoDescs)
     {
+        GFXRECON_UNREFERENCED_PARAMETER(call_info);
+        GFXRECON_UNREFERENCED_PARAMETER(object_id);
+        GFXRECON_UNREFERENCED_PARAMETER(pDesc);
+        GFXRECON_UNREFERENCED_PARAMETER(NumPostbuildInfoDescs);
+        GFXRECON_UNREFERENCED_PARAMETER(pPostbuildInfoDescs);
         dxr_workload_ = true;
     }
 
@@ -252,6 +290,8 @@ class Dx12StatsConsumer : public Dx12Consumer
     ProcessFillMemoryResourceValueCommand(const format::FillMemoryResourceValueCommandHeader& command_header,
                                           const uint8_t*                                      data)
     {
+        GFXRECON_UNREFERENCED_PARAMETER(command_header);
+        GFXRECON_UNREFERENCED_PARAMETER(data);
         opt_fillmem_ = true;
     }
 
@@ -260,6 +300,9 @@ class Dx12StatsConsumer : public Dx12Consumer
         const std::vector<format::InitDx12AccelerationStructureGeometryDesc>& geometry_descs,
         const uint8_t*                                                        build_inputs_data)
     {
+        GFXRECON_UNREFERENCED_PARAMETER(command_header);
+        GFXRECON_UNREFERENCED_PARAMETER(geometry_descs);
+        GFXRECON_UNREFERENCED_PARAMETER(build_inputs_data);
         dxr_workload_ = true;
     }
 
@@ -272,19 +315,28 @@ class Dx12StatsConsumer : public Dx12Consumer
                                                            format::HandleId   pCountBuffer,
                                                            UINT64             CountBufferOffset)
     {
+        GFXRECON_UNREFERENCED_PARAMETER(call_info);
+        GFXRECON_UNREFERENCED_PARAMETER(object_id);
+        GFXRECON_UNREFERENCED_PARAMETER(pCommandSignature);
+        GFXRECON_UNREFERENCED_PARAMETER(MaxCommandCount);
+        GFXRECON_UNREFERENCED_PARAMETER(pArgumentBuffer);
+        GFXRECON_UNREFERENCED_PARAMETER(ArgumentBufferOffset);
+        GFXRECON_UNREFERENCED_PARAMETER(pCountBuffer);
+        GFXRECON_UNREFERENCED_PARAMETER(CountBufferOffset);
         ei_workload_ = true;
     }
 
-    void Process_IDXGISwapChain_Present(const ApiCallInfo& call_info,
-                                        format::HandleId   object_id,
-                                        HRESULT            return_value,
-                                        UINT               SyncInterval,
-                                        UINT               Flags)
+    void Process_IDXGISwapChain_Present(
+        const ApiCallInfo& call_info, format::HandleId object_id, HRESULT return_value, UINT SyncInterval, UINT Flags)
     {
+        GFXRECON_UNREFERENCED_PARAMETER(call_info);
+        GFXRECON_UNREFERENCED_PARAMETER(object_id);
+        GFXRECON_UNREFERENCED_PARAMETER(return_value);
+        GFXRECON_UNREFERENCED_PARAMETER(SyncInterval);
         if (Flags & DXGI_PRESENT_TEST)
         {
             dxgi_present_test_++;
-        }        
+        }
     }
 
     virtual void ProcessDx12RuntimeInfo(const format::Dx12RuntimeInfoCommandHeader& runtime_info_header)
@@ -295,11 +347,14 @@ class Dx12StatsConsumer : public Dx12Consumer
                                    sizeof(runtime_info_header.runtime_info));
     }
 
-    virtual void ProcessSetSwapchainImageStateCommand(format::HandleId                                    device_id,
-                                                      format::HandleId                                    swapchain_id,
-                                                      uint32_t                                            current_buffer_index,
+    virtual void ProcessSetSwapchainImageStateCommand(format::HandleId device_id,
+                                                      format::HandleId swapchain_id,
+                                                      uint32_t         current_buffer_index,
                                                       const std::vector<format::SwapchainImageStateInfo>& image_state)
     {
+        GFXRECON_UNREFERENCED_PARAMETER(device_id);
+        GFXRECON_UNREFERENCED_PARAMETER(swapchain_id);
+        GFXRECON_UNREFERENCED_PARAMETER(image_state);
         dummy_trim_frame_count_ = current_buffer_index;
     }
 
@@ -310,6 +365,10 @@ class Dx12StatsConsumer : public Dx12Consumer
                                            Decoded_GUID                 riid,
                                            HandlePointerDecoder<void*>* ppDevice)
     {
+        GFXRECON_UNREFERENCED_PARAMETER(call_info);
+        GFXRECON_UNREFERENCED_PARAMETER(return_value);
+        GFXRECON_UNREFERENCED_PARAMETER(MinimumFeatureLevel);
+        GFXRECON_UNREFERENCED_PARAMETER(riid);
         GFXRECON_ASSERT(ppDevice != nullptr);
 
         if (ppDevice != nullptr && ppDevice->GetPointer() != nullptr)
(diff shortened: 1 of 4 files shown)


framework/decode/vulkan_stats_consumer.h
```diff
@@ -101,31 +148,97 @@ class VulkanStatsConsumer : public gfxrecon::decode::VulkanConsumer
...
                 {
-                    engine_name_ = app_info->pEngineName;
+                    instance_tracker.engine_name = app_info->pEngineName;
+                }
+
+                instance_tracker.app_version    = app_info->applicationVersion;
+                instance_tracker.engine_version = app_info->engineVersion;
+                instance_tracker.api_version    = app_info->apiVersion;
+
+                for (uint32_t iii = 0; iii < create_info->enabledExtensionCount; ++iii)
```
fabian-lunarg: just curious, what is the reason for triple I 'iii' as variable name, there is no naming conflict?

laurentiu-erhan-arm:
Regarding #2396 , as far as i can tell from the output in the comments, the "conflict" between 
"Application name:    3DMark SlingShot" and 
"Application name:    android framework" 
is not really resolved in the non-verbose output, only one of them being mentioned in the default output (in this case the 3DMark SlingShot one). Wouldn't it be better to show both of them? (this being extended to the other instance fields like engine name)

framework/decode/vulkan_stats_consumer.h
MarkY-LunarG: Cleaned this up to "ext" in the extension cases and "pd" or "pdg" in other cases as necessary.   It's just an easy to search for index, honestly, but not a big deal to make clearer.

MarkY-LunarG:
> Regarding #2396 , as far as i can tell from the output in the comments, the "conflict" between "Application name: 3DMark SlingShot" and "Application name: android framework" is not really resolved in the non-verbose output, only one of them being mentioned in the default output (in this case the 3DMark SlingShot one). Wouldn't it be better to show both of them? (this being extended to the other instance fields like engine name)

The concern is people are using `gfxrecon-info` for automated testing validation and having a semi-colon list suddenly appear in these fields could cause issues.  However, we were using the last VkInstance always, and now I determine which instance appears to do the most work and show that instance's info.  This is especially important since many times there would be blank application info for some instances which  were created for querying purposes. 

The verbose output provides  this functionality in a JSON parsable way that can be easily used to find multiple app info if needed, otherwise, we  recommend using `gfxrecon-convert` and parsing it directly.

MarkY-LunarG:
Example JSON output: [test.json](https://github.com/user-attachments/files/23269453/test.json)

Also,, now uses "most important VkInstance" instead of last VkInstance.

framework/decode/vulkan_stats_consumer.h
```diff
@@ -103,31 +157,99 @@ class VulkanStatsConsumer : public gfxrecon::decode::VulkanConsumer
...
     virtual void Process_vkCreateInstance(
-        const gfxrecon::decode::ApiCallInfo&                                                    call_info,
-        VkResult                                                                                returnValue,
-        gfxrecon::decode::StructPointerDecoder<gfxrecon::decode::Decoded_VkInstanceCreateInfo>* pCreateInfo,
-        gfxrecon::decode::StructPointerDecoder<gfxrecon::decode::Decoded_VkAllocationCallbacks>*,
-        gfxrecon::decode::HandlePointerDecoder<VkInstance>*) override
+        const gfxrecon::decode::ApiCallInfo&                                                     call_info,
+        VkResult                                                                                 returnValue,
+        gfxrecon::decode::StructPointerDecoder<gfxrecon::decode::Decoded_VkInstanceCreateInfo>*  pCreateInfo,
+        gfxrecon::decode::StructPointerDecoder<gfxrecon::decode::Decoded_VkAllocationCallbacks>* allocCb,
```
bradgrantham-lunarg, line 188: If allocCb isn't used, please put GFXRECON_UNREFERENCED_PARAMETER in the function body referencing the variable.

framework/decode/openxr_stats_consumer.h
```diff
@@ -42,14 +42,23 @@
...
 
+struct OpenXrInstanceTracker
+{
+    std::string              app_name;
+    uint32_t                 app_version{ 0 };
+    std::string              engine_name;
+    uint32_t                 engine_version{ 0 };
+    uint32_t                 api_version{ 0 };
+    std::vector<std::string> enabled_extensions;
+};
```
bradgrantham-lunarg, line 60: I suggest this class (and VulkanInstanceTracker and the other Vulkan data classes) should have constructors that take the source structs and populate the members directly.

FYI enabled_extensions can be directly populated from a `std::span`, like:
 ```
std::span<const char* const> exts(create_info->enabledExtensionNames, create_info->enabledExtensionCount);
enabled_extensions( exts.begin(), exts.end());
```

File: framework/decode/openxr_stats_consumer.h [hunk 4]
```diff
@@ -58,26 +67,31 @@ class OpenXrStatsConsumer : public gfxrecon::decode::OpenXrConsumer
...
+            instance_tracker.engine_name    = app_info.engineName;
+            instance_tracker.app_version    = app_info.applicationVersion;
+            instance_tracker.engine_version = app_info.engineVersion;
+            instance_tracker.api_version    = app_info.apiVersion;
+
+            for (uint32_t ext = 0; ext < create_info->enabledExtensionCount; ++ext)
+            {
+                instance_tracker.enabled_extensions.push_back(create_info->enabledExtensionNames[ext]);
+            }
+            gfxrecon::format::HandleId inst = *const_cast<gfxrecon::format::HandleId*>(pInstance->GetPointer());
```
bradgrantham-lunarg: Can you use `const gfxrecon::format::HandleId inst = ...` instead of `const_cast`?

framework/decode/vulkan_stats_consumer.h
```diff
@@ -103,31 +157,99 @@ class VulkanStatsConsumer : public gfxrecon::decode::VulkanConsumer
...
                 {
-                    engine_name_ = app_info->pEngineName;
+                    instance_tracker.app_info.engine_name = app_info->pEngineName;
+                }
+
+                instance_tracker.app_info.app_version    = app_info->applicationVersion;
+                instance_tracker.app_info.engine_version = app_info->engineVersion;
+                instance_tracker.app_info.api_version    = app_info->apiVersion;
+
+                for (uint32_t ext = 0; ext < create_info->enabledExtensionCount; ++ext)
```
bradgrantham-lunarg: I suggest you initialize using `span` as mentioned above

framework/decode/vulkan_stats_consumer.h
```diff
@@ -170,65 +292,108 @@ class VulkanStatsConsumer : public gfxrecon::decode::VulkanConsumer
...
+
+            if (std::find(instance_info_[instance].used_physical_devices.begin(),
+                          instance_info_[instance].used_physical_devices.end(),
+                          phys_dev) == instance_info_[instance].used_physical_devices.end())
+            {
+                used_physical_devices_.insert(physicalDevice);
+                instance_info_[instance].used_physical_devices.push_back(phys_dev);
+            }
+
+            VkDevice device = const_cast<VkDevice>(*reinterpret_cast<const VkDevice*>(pDevice->GetPointer()));
```
bradgrantham-lunarg: Can you use `const VkDevice device` here and then not use const_cast?

framework/decode/vulkan_stats_consumer.h
```diff
@@ -170,65 +292,108 @@ class VulkanStatsConsumer : public gfxrecon::decode::VulkanConsumer
...
+
+            if (std::find(physical_device_info_[phys_dev].devices.begin(),
+                          physical_device_info_[phys_dev].devices.end(),
+                          device) == physical_device_info_[phys_dev].devices.end())
+            {
+                physical_device_info_[phys_dev].devices.push_back(device);
+            }
+
+            VulkanDeviceTracker dev_tracker{};
+            for (uint32_t ext = 0; ext < create_info->enabledExtensionCount; ++ext)
```
bradgrantham-lunarg: I suggest you initialize using `span` as mentioned above

tools/info/main.cpp
```diff
@@ -70,17 +70,19 @@
...
+const char kExeInfoOnlyOption[]    = "--exe-info-only";
+const char kEnvVarsOnlyOption[]    = "--env-vars-only";
 const char kFileFormatOnlyOption[] = "--file-format-only";
-const char kEnumGpuIndices[]    = "--enum-gpu-indices";
+const char kEnumGpuIndices[]       = "--enum-gpu-indices";
+const char kVerboseArgument[]      = "--verbose";
 
 const char kOptions[] =
     "-h|--help,--version,--no-debug-popup,--exe-info-only,--env-vars-only,--file-format-only,--enum-gpu-indices";
+const char kArguments[] = "--verbose";
```
bradgrantham-lunarg: What do you think about this just being an option and outputting on stdout?  Then a user can send it wherever they want or run it in Python `subprocess` and directly parse the results.   Otherwise add the ability for filename to be `"-"` meaning stdout.

bradgrantham-lunarg (changes requested):

framework/decode/openxr_stats_consumer.h
MarkY-LunarG: I don't think you can use span on a dynamically sized c-style string array.  You can just copy directly into a vector with other methods which I'll use.

framework/decode/vulkan_stats_consumer.h
MarkY-LunarG: Same comment, but I'll clean it up.

framework/decode/openxr_stats_consumer.h
bradgrantham-lunarg: https://godbolt.org/z/YjT1KjE4s

bradgrantham-lunarg: `span` takes the length as part of the constructor, not as a template parameter, so it totally works on C pointers.  I think that's one of the rationales for it.  The code I pasted in my original requested change was cut-and-pasted from a test branch I had made which I no longer have.  So I've pasted in a Compiler Explorer link demonstrating `span`.

tools/info/main.cpp
```diff
@@ -70,22 +71,66 @@
...
+const char kOptions[]   = "-h|--help,--version,--no-debug-popup,--exe-info-only,--env-vars-only,--file-format-only,--"
+                          "enum-gpu-indices,--verbose";
+const char kArguments[] = "--output";
 
 const char kUnrecognizedFormatString[] = "<unrecognized-format>";
 
 const int kDefaultIndent = 12;
 
+// Global variables defining where we should output results to
+static bool          g_write_to_console = true;
```
bradgrantham-lunarg: Just have `g_output_file` and it's either a file that's opened or it's std::cout.  If no filename option, it's cout, if filename is "-" then cout, otherwise open the filename.

I think maybe the logging output e.g. usage should always be on cout?

If you just had `g_output_file` then you'd have the benefit of writing the JSON output with `g_output_file << json_content->dump(4, ' ', true) << '\n' `.

It's unfortunate that the non-JSON info used GFXRECON_WRITE_CONSOLE, so I get why you'd need a varargs implementation to send the non-JSON to a file.

You may be able to use `std::format`, and then all the calls to `WriteOutput(fmt, a, b, c)` turn into `g_output_file << std::format(fmt, a, b, c) << '\n';`

tools/info/main.cpp
```diff
@@ -339,57 +420,94 @@ struct FileFormatInfo
...
 
-void PrintFileFormatInfo(const gfxrecon::decode::FileProcessor& file_processor)
+void PrintFileFormatInfo(const gfxrecon::decode::FileProcessor& file_processor, nlohmann::json* json_output)
 {
-    PrintFileFormatInfo(FileFormatInfo{ file_processor });
+    PrintFileFormatInfo(FileFormatInfo{ file_processor }, json_output);
 }
 
-void PrintDriverInfo(const gfxrecon::decode::InfoConsumer& driver_info_consumer)
+void PrintDriverInfo(const gfxrecon::decode::InfoConsumer& driver_info_consumer, nlohmann::json* json_output)
```
bradgrantham-lunarg: This puts control flow down into every function.  I think I think I'd prefer different versions of the function yielding polymorphism, like PrintDriverInfo that does the old style, and DriverInfoToJson which returns a nlohmann::json, and the `if` is managed above these functions.

tools/info/main.cpp
```diff
@@ -427,230 +545,479 @@ FileFormatInfo GatherFileFormatInfo(gfxrecon::decode::FileProcessor& file_proces
...
+            instance_json["application-info"]["application"]["name"]    = it.second.app_name;
+            instance_json["application-info"]["application"]["version"] = it.second.app_version;
+            instance_json["application-info"]["engine"]["name"]         = it.second.engine_name;
+            instance_json["application-info"]["engine"]["version"]      = it.second.engine_version;
+            instance_json["application-info"]["api-version"]            = GetXrVersionString(it.second.api_version);
+
+            if (it.second.enabled_extensions.size() > 0)
+            {
+                instance_json["extensions"] = nlohmann::json::array();
+                for (uint32_t ext = 0; ext < it.second.enabled_extensions.size(); ++ext)
```
bradgrantham-lunarg: Use your idiom for initializing a vector of string here and in similar loops below.

tools/info/main.cpp
```diff
@@ -339,57 +420,94 @@ struct FileFormatInfo
...
+    std::string file_desc    = info_consumer.GetFileDescription();
+    std::string product_name = info_consumer.GetProductName();
 
-    // we are combining file description and product name and presenting both only if they are not same
-    std::string app_data = info_consumer.GetFileDescription();
-    if (strcmp(info_consumer.GetProductName(), "N/A") != 0)
+    if (json_output != nullptr)
     {
-        if (strcmp(info_consumer.GetProductName(), info_consumer.GetFileDescription()) != 0)
+        (*json_output)["exe"]["name"]    = info_consumer.GetAppExeName();
```
bradgrantham-lunarg: For all these cases where you're populating a hierarchy within a `json` you should use initializer lists to avoid those repeated string lookups and make this look more idiomatic, like:
```
std::string version_str = std::to_string(exe_version[0]) + "." +
                          std::to_string(exe_version[1]) + "." +
                          std::to_string(exe_version[2]) + "." +
                          std::to_string(exe_version[3]);

nlohmann::json exe_info = {
    {"exe", {
        {"version", version_str},
        {"company", info_consumer.GetCompanyName()},
        {"product", info_consumer.GetProductName()},
        {"file-description", info_consumer.GetFileDescription()}
    }}
};
```

If you decompose these functions into `Print` and `ToJson` versions like I mention elsewhere, you could even return the json directly:
```
return nlohmann:json {
    {"exe", {
        {"version", version_str},
        {"company", info_consumer.GetCompanyName()},
        {"product", info_consumer.GetProductName()},
        {"file-description", info_consumer.GetFileDescription()}
    }}
};
```

tools/info/main.cpp
```diff
@@ -427,230 +545,479 @@ FileFormatInfo GatherFileFormatInfo(gfxrecon::decode::FileProcessor& file_proces
...
+{
 
-        if (!vulkan_stats_consumer.GetResolutions().empty())
+    uint32_t inst_count    = vulkan_stats_consumer.GetInstanceCount();
+    auto     instance_info = vulkan_stats_consumer.GetInstanceInfo();
+    auto     pd_info       = vulkan_stats_consumer.GetPhysicalDeviceInfo();
+    auto     dev_info      = vulkan_stats_consumer.GetDeviceInfo();
+    if (json_output != nullptr)
+    {
+        (*json_output)["vulkan"]["header-version"] = std::to_string(XR_VERSION_MAJOR(XR_CURRENT_API_VERSION)) + "." +
```
bradgrantham-lunarg: This probably shouldn't be XR macros

tools/info/main.cpp
MarkY-LunarG: Done in several locations.

MarkY-LunarG:
@bradgrantham-lunarg , this  change includes a switch  to using an interface class definition which is inherited for handling printing out the JSON/console messages.  It also groups those utility functions together per class.  I don't know if this is the wrong direction, but let me know what you think.

framework/decode/vulkan_stats_consumer.h
```diff
@@ -53,39 +55,108 @@
...
 
-class VulkanStatsConsumer : public gfxrecon::decode::VulkanConsumer
+class VulkanStatsConsumer : public gfxrecon::decode::VulkanConsumer, public gfxrecon::util::InfoOutputInterface
 {
+
+  private:
+    struct VulkanDeviceTracker
+    {
+        VulkanDeviceTracker() {}
+        VulkanDeviceTracker(const uint32_t ext_count, const char* const* exts)
```
bradgrantham-lunarg: You can initialize enabled_extensions here directly from exts and exts + ext_count, like
```
        VulkanDeviceTracker(const uint32_t ext_count, const char* const* exts) :
            enabled_extensions(exts, exts + ext_count)
```
The move is redundant because that `std::vector<std::string>>` is a temporary, so is an rvalue, so defaults to move.  Same in VulkanInstanceTracker below.

framework/decode/vulkan_stats_consumer.h
```diff
@@ -170,65 +296,104 @@ class VulkanStatsConsumer : public gfxrecon::decode::VulkanConsumer
...
     }
 
     virtual void Process_vkCreateGraphicsPipelines(
         const gfxrecon::decode::ApiCallInfo& call_info,
         VkResult                             returnValue,
-        gfxrecon::format::HandleId,
-        gfxrecon::format::HandleId,
-        uint32_t createInfoCount,
+        format::HandleId                     device,
+        format::HandleId                     pipelineCache,
```
bradgrantham-lunarg, line 362: Throughout this PR, if you added a name to a parameter but didn't use it, please add it with `GFXRECON_UNREFERENCED_PARAMETER`.

tools/info/main.cpp
```diff
@@ -195,631 +147,435 @@ static bool CheckOptionPrintVersion(const char* exe_name, const gfxrecon::util::
...
-    FileFormatInfo(const gfxrecon::decode::FileProcessor& file_processor)
+    std::pair<const std::string, const nlohmann::json*> GenerateJson() const override
     {
-        const gfxrecon::format::FileHeader& file_header = file_processor.GetFileHeader();
-        major_version                                   = file_header.major_version;
-        minor_version                                   = file_header.minor_version;
-        uses_frame_markers                              = file_processor.UsesFrameMarkers();
-        file_supports_frame_markers                     = file_processor.FileSupportsFrameMarkers();
-    }
+        nlohmann::json* capture_file_base = new nlohmann::json;
```
bradgrantham-lunarg: Make an nlohmann::json on the stack and populate it with construction and then return that.  Return-value-optimization will likely make that higher performance and it will remove the use of pointers.  Same guidance I provided in previous feedback https://github.com/LunarG/gfxreconstruct/pull/2451#discussion_r2561065862  For this one and the other instances of passing into a pointer to a JSON below.

Does the name of the object need to be in a pair returned from this function?  That makes a policy statement about this object that to me seems like it belongs to the caller.

tools/info/main.cpp
```diff
@@ -195,631 +147,435 @@ static bool CheckOptionPrintVersion(const char* exe_name, const gfxrecon::util::
...
+        std::string data;
+    };
 
-    if (adapters.empty() == false)
+    AnnotationRecorder()
     {
-        std::unordered_map<int64_t, std::string> adapter_workload;
-        dx12_consumer.CalcAdapterWorkload(adapter_workload, adapters);
+        // Define all possible Annotation fields and let them record the info as encountered
+        AnnotationRecorder::AnnotationInfo expected_annotations[] = {
```
bradgrantham-lunarg: I think this should probably be a statically-initialized std::map from key to desc, and you can "contains" the map in your search loop in ProcessAnnotation.  In any case avoid an assignment with magic number.

tools/info/main.cpp
```diff
@@ -195,631 +147,435 @@ static bool CheckOptionPrintVersion(const char* exe_name, const gfxrecon::util::
...
+        auto        search_result_iterator = json_obj.find(key);
+        if (search_result_iterator != json_obj.end())
         {
-            GFXRECON_WRITE_CONSOLE("\tTest present count: %u", dx12_consumer.GetDXGITestPresentCount());
+            const nlohmann::json& value = *search_result_iterator;
+            if (value.is_object())
+            {
+                out += "\n\t" + value.dump(kDefaultIndent);
+                out.pop_back();
+                out += "\t}";
```
bradgrantham-lunarg: Why is this bare ending brace in this string?

framework/util/info_output.h
```diff
@@ -0,0 +1,151 @@
...
+
+    void AddInfoOutputItem(const InfoOutputInterface* io)
+    {
+        if (io)
+        {
+            info_output_vec_.push_back(io);
+        }
+    }
+
+    void SetFileTarget(const std::string& filename)
```
bradgrantham-lunarg: Don't have this object open the file.  Rather pass in the file, and then it can be an actual file or it can be std::cout, and the thing that creates this object can make the policy decision about what it sends in.  For that matter I think initialize this object with the ofstream.

framework/util/info_output.h
```diff
@@ -0,0 +1,151 @@
...
+        else
+        {
+            GFXRECON_WRITE_CONSOLE(out_string.c_str());
+        }
+    }
+
+    std::ofstream                           output_file;
+    InfoOutputFormat                        format_{ kInfoOutputFmt_Text };
+    nlohmann::json*                         json_content_{ nullptr };
+    std::vector<const InfoOutputInterface*> info_output_vec_;
```
bradgrantham-lunarg: Why does the json content need to be batched up?  I would assume info tool would just construct a whole `nlohmann::json` and then output it.  And then why do the output strings or messages need to be batched up?  Maybe we should discuss.

tools/info/main.cpp
MarkY-LunarG: IDK. I think it was removing the one from the object output  to make this integrate with the JSON of the resulting file, but it was still wrong, so reverted it back to just a "value.dump()".

framework/util/info_output.h
MarkY-LunarG: No longer done.  Reverted.

tools/info/main.cpp
bradgrantham-lunarg: Discussed, alternate implementation is in progress

framework/decode/dx12_stats_consumer.h
```diff
@@ -131,6 +131,9 @@ class Dx12StatsConsumer : public Dx12Consumer
                                                 HRESULT                                           return_value,
                                                 StructPointerDecoder<Decoded_DXGI_ADAPTER_DESC3>* pDesc)
     {
+        GFXRECON_UNREFERENCED_PARAMETER(call_info);
```
bradgrantham-lunarg, line 134: I guess it's okay to check these in but my feedback was "Throughout this PR, if **you** added a name to a parameter but didn't use it, please add it with GFXRECON_UNREFERENCED_PARAMETER"

MarkY-LunarG: It wasn't much more and I figured why not just bring it all up to date.

MarkY-LunarG:
@bradgrantham-lunarg , anything else needed?

bradgrantham-lunarg:
We discussed using `std::format` offline and I think we agreed that is a good direction to go but first we have to figure out the interaction with logging.  Current `dev` uses `GFXRECON_WRITE_CONSOLE` directly and this PR wraps that with `WriteOutput` and that goes through the logging facility which may not be desirable.  I think we can attempt to address that in a later PR.

PR #2452, opened 2025-10-09
Fix `FpsInfo`

marius-pelegrin-arm:
These two commits aim at fixing the behavior and architecture of `FpsInfo` without any breaking change.

### First commit: Make `FpsInfo` API-agnostic

As frames are defined in an API-agnostic way in GFXReconstruct, nothing forces `FpsInfo` to be API dependent.

In particular, consumers shouldn't have access to `FpsInfo` (at least not in it's current way) and markers should be dispatched to `FpsInfo` through the file processor directly. Doing this reduces code complexity and improves readability.

Also, as a consequence, measurement options are not Vulkan dependent and should be enabled no matter the API.

### Second commit: Fix issues caused by the "frame counting stuttering"

Frame numbering in GFXReconstruct works as follows:
- There is a list of calls that when encountered are considered a "frame delimiter"
- There is a meta-command, added to the trace at capture time, `FrameEndMarker`, that is used at replay time to determinate in a platform and API agnostic way what is an and of frame
- There is a meta-command, added to the trace at trimming time (so at capture time), `StateEndMarker` that indicates the end of the loading of the trimming state, before the beginning of the first frame at replay time
    
When the replay begins, it starts the "first frame".
Later, it *may* encounter a `StateEndMarker`, that will "correct" what is actually the beginning of the first frame.
Later again, it *may* encounter one or more API calls that are "frame delimiters" and will increment the frame counter.
And at some point, it *may* encounter a `FrameEndMarker` that will "correct" what is actually the end of the first frame. If a `FrameEndMarker` is encountered, frame numbering can only be incremented by these meta-commands in the rest of the trace.
    
This whole mechanism leads to the frame numbering to "stutter" from the point of view of the replayer, but in particular from the point of view of `FpsInfo`, that will *usually* encounter the following calls:
- BeginFrame(1)
- StateEndMarker <- equivalent to BeginFrame(1)
- EndFrame(1) <- ended by API call
- BeginFrame(2)
- EndFrame(2) <- ended by `FrameEndMarker`
- BeginFrame(2) <- because `FrameEndMarker` resets the frame count
- EndFrame(2)
- BeginFrame(3)
- ...
    
For options based on the frame number, this "stuttering" can cause big issues. In particular:
- If preloading from frame 2, preloading will be called twice, that may lead in a crash
- If measurements include frame 1 or 2, it can lead to logs being displayed multiple times without explanation for the user, errors in the measurements, errors in when the replay should quit...
    
This commit aims at fixing this situation without introducing any breaking change (frame numbering change, meta-command change etc..)

### DIFF
--- a/framework/graphics/fps_info.cpp
+++ b/framework/graphics/fps_info.cpp
@@ -110,13 +110,14 @@ void FpsInfo::EndFrame(uint64_t frame)
 
         // We need to check this because of the "frame stuttering" induced by the first frame end marker...
         uint64_t index = frame - measurement_start_frame_;
-        if (index < frame_durations_.size())
-        {
-            frame_durations_[index] += frame_duration;
-        }
-        else if (index == frame_durations_.size())
+        if (index == frame_durations_.size())
         {
             frame_durations_.push_back(frame_duration);
+        }
+        else if (index + 1 == frame_durations_.size())
+        {
+            GFXRECON_LOG_WARNING("Possible incomplete last frame detected. Measurements will consider the incomplete "
+                                 "frame as the last frame. This message should appear only once !");
         }
         else
         {
@@ -148,6 +149,14 @@ void FpsInfo::EndFile(uint64_t frame)
 
     if (!ended_measurement_)
     {
+        if (frame - measurement_start_frame_ + 1 == frame_durations_.size())
+        {
+            int64_t frame_end_time = util::datetime::GetTimestamp();
+            int64_t frame_duration = util::datetime::DiffTimestamps(frame_start_time_, frame_end_time);
+            frame_durations_.push_back(frame_duration);
+            ++frame;
+        }
+
         measurement_end_boot_time_    = util::datetime::GetBootTime();
         measurement_end_process_time_ = util::datetime::GetProcessTime();
         measurement_end_time_         = replay_end_time_;


MarkY-LunarG:
@locke-lunarg / @davidd-lunarg  can you review this from a D3D standpoint?  @jzulauf-lunarg can you verify it won't cause issues with your upcoming cleanup of the loaing/pre-loading.

jzulauf-lunarg:
I think a better insertion point would be in FileProcessor::ProcessBlocksOneFrame.  There is more than one way to end a frame than the frame marker, and there is a substantial refactor to the Process<BlockType> https://github.com/LunarG/gfxreconstruct/pull/2456  where functional and state elements are removed from the block parsing code, that this will fit better with.

The next step is to rewrite the PreloadFileProcessor to preload/pre-parse/pre-decompress, and encapsulating the Fps handling at a higher level will help make this easier as well.

marius-pelegrin-arm:
@jzulauf-lunarg I think there has been a misunderstanding. Frame end marker is not considered the only "way to end a frame", I just added a function to handle better the transition from "frame delimiter calls" to frame end markers, which requires detecting this way of ending a frame in particular.
Also, general frame end detection is done in `Applicaiton`, which is even above the `FileProcessor`, so I don't understand your suggestion.

jzulauf-lunarg:
With the most recent commits, preload doesn't update frame numbers twice AFAICT anymore.  The added functionality is the ability to do a timer reset at the end of state load, while processing "live" or during replay, without doing so in the consumer, yes?

framework/graphics/fps_info.cpp
```diff
@@ -143,9 +158,54 @@ void FpsInfo::EndFile(uint64_t frame)
...
+
+        // May restart measurements immediately because now is the actual start of frame 1
+        BeginFrame(1);
+    }
+}
+
+void FpsInfo::ProcessFirstFrameEndMarker()
+{
+    // When this function is called, it means we are about to reach end of frame 1
+    // But it's possible that until now received multiple calls to {Begin/End}Frame with different frame numbers
```
jzulauf-lunarg, line 188: Is this comment still true as of current HEAD of dev?

marius-pelegrin-arm: Yes. This has nothing to do with how the file processor handles frames. It has to do with the impossibility to know in advance if a trace contains frame end markers or not.
I invite you to search for when and how `kFirstFrame` is used in the file processor.

In particular in `FileProcessor::ProcessFrameMarker`:

```
        // If the capture file contains frame markers, it will have a frame marker for every
        // frame-ending API call such as vkQueuePresentKHR. If this is the first frame marker
        // encountered, reset the frame count and ignore frame-ending API calls in
        // IsFrameDelimiter(format::ApiCallId call_id).
```

marius-pelegrin-arm:
@jzulauf-lunarg No, this stutter has nothing to do with the preload file processor. Actually it's the opposite: The stutter CAUSES the preloading to be called twice under some circumstances.

jzulauf-lunarg:
@marius-pelegrin-arm -- I see what you mean now.  Something I'll need to be sure to support correctly in my refactor.

framework/graphics/fps_info.cpp
```diff
@@ -106,7 +106,22 @@ void FpsInfo::EndFrame(uint64_t frame)
...
     {
         int64_t frame_end_time = util::datetime::GetTimestamp();
-        frame_durations_.push_back(util::datetime::DiffTimestamps(frame_start_time_, frame_end_time));
+        int64_t frame_duration = util::datetime::DiffTimestamps(frame_start_time_, frame_end_time);
+
+        // We need to check this because of the "frame stuttering" induced by the first frame end marker...
+        uint64_t index = frame - measurement_start_frame_;
+        if (index < frame_durations_.size())
+        {
+            frame_durations_[index] += frame_duration;
```
locke-lunarg: Should it be `frame_durations_[index] = frame_duration;` ?

marius-pelegrin-arm: So after checking, these are actually forgotten lines of a codepath it should not be possible to reach anymore. So I tried removing it and replacing it with an assert on `index == frame_durations_.size()` and gave a try on a few edge cases. I actually found an interesting new frame stutter.

So now there are three "frame stutter" cases.

The first one is the state end marker, causing a frame to "begin twice" and is handled in `ProcessStateEndMarker`. Measurements are "restarted" by setting `started_measurement_ = false` and calling `BeginFrame(1)` which clears the `frame_durations_` vector.

The second one is the first frame end marker, causing the counter to restart. This is handled in `ProcessFirstFrameEndMarker`, which either clears `frame_durations_` directly or set `started_measurement_ = false`, which also lead to the `frame_durations_` vector to be cleared.

The third one, the I just found, is the case where the measurement range goes beyond the number of frames of the capture and where the last call recorded in the trace is not a frame end marker and or frame delimiter. Then an "incomplete frame" ends, the frame number is not incremented, and we end up in this case again.


I suggest we consider this last incomplete frame as an "additional frame". I think the proper way to do this is to log a warning in `EndFrame` when we detect that "frame stutter", and we actually add a frame if necessary in `EndFile` (because then we are sure this frame stutter is caused by this last frame)

marius-pelegrin-arm: I added a commit to implement this fix.

locke-lunarg: Thank you for fixing. I was thinking how `if (index < frame_durations_.size())` could happen.

marius-pelegrin-arm: In theory it cannot happen, that's why in the new commit, if `index < frame_durations_.size()` an error message is logged

fabian-lunarg:
ok, so this has now conflicts with recent work. the rebase does not look trivial and requires some attention.

fabian-lunarg:
> ok, so this has now conflicts with recent work. the rebase does not look trivial and requires some attention.

apparently the recent work done in https://github.com/LunarG/gfxreconstruct/pull/2474 should
solve the same problem. 
meaning we're likely to close this PR. 
did you maybe have a chance to confirm current dev has indeed solved the issues tackled here @marius-pelegrin-arm ?

fabian-lunarg:
closing here. we assume https://github.com/LunarG/gfxreconstruct/pull/2474 has dealt with the issue. if something contained in this PR has been overlooked, we'll create a follow-up.

PR #2453, opened 2025-10-09
Fix in ahb shader copy path

panos-lunarg:
The shader copy code path that copies the content of AHB buffers via a
compute shader uses a ycbcr sampler without checking the format of the
AHB. This can cause glitches when an AHB is not a yuv format. This
commits fixes the issue.

### DIFF
--- a/framework/encode/vulkan_capture_common.cpp
+++ b/framework/encode/vulkan_capture_common.cpp
@@ -21,6 +21,7 @@
  */
 
 #include "vulkan_capture_common.h"
+#include "Vulkan-Utility-Libraries/vk_format_utils.h"
 
 #if defined(VK_USE_PLATFORM_ANDROID_KHR)
 #include <android/hardware_buffer.h>
@@ -126,6 +127,20 @@ static void CommonWriteFillMemoryCmd(format::HandleId      memory_id,
     else
     {
         vulkan_state_writer->WriteFillMemoryCmd(memory_id, 0u, size, data);
+    }
+}
+
+// Wrap vkuFormatRequiresYcbcrConversion so that we handle VK_FORMAT_UNDEFINED as a YCbCr format which applies for AHB
+// external formats
+static bool ExternalFormatRequiresYcbcrConversion(VkFormat format)
+{
+    if (format == VK_FORMAT_UNDEFINED)
+    {
+        return true;
+    }
+    else
+    {
+        return vkuFormatRequiresYcbcrConversion(format);
     }
 }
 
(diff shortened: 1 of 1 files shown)


panos-lunarg:
> Should we add an info message indicating we're adding the YCbCr conversion? Or are there already messages?

This is an internal copy we do from AHBs which are not cpu visible into a vulkan image that we can read and dump into the capture file. AHBs are usually used with yuv formats so that's my guess why had the ycbcr sampler on all the times.
I don't we should generate a message about something so specific

framework/encode/vulkan_capture_common.cpp
```diff
@@ -405,8 +406,9 @@ void CommonProcessHardwareBuffer(format::ThreadId                      thread_id
         auto ahb_image_aspect_mask = graphics::GetFormatAspects(format_properties.format);
 
         VkImageViewCreateInfo image_view_create_info;
-        image_view_create_info.sType                           = VK_STRUCTURE_TYPE_IMAGE_VIEW_CREATE_INFO;
-        image_view_create_info.pNext                           = &sampler_ycbcr_conversion_info;
+        image_view_create_info.sType = VK_STRUCTURE_TYPE_IMAGE_VIEW_CREATE_INFO;
+        image_view_create_info.pNext =
+            vkuFormatRequiresYcbcrConversion(format_properties.format) ? &sampler_ycbcr_conversion_info : nullptr;
```
antonio-lunarg: Hold on! In case of an external format, this can be `VK_FORMAT_UNDEFINED`, in which case [vkuFormatRequiresYcbcrConversion](https://github.com/KhronosGroup/Vulkan-Utility-Libraries/blob/main/include/vulkan/utility/vk_format_utils.h#L1444) would return `false`, which is probably _not_ what we want.

panos-lunarg: I am not very familiar with AHB lingo so what happens if `format_properties.format == VK_FORMAT_UNDEFINED`? 

Both the ycbcr sampler and the image view that uses that sampler are assigned `VK_FORMAT_UNDEFINED`.  But the image view gets the `ahb_image` image which in lines [298 - 301](https://github.com/LunarG/gfxreconstruct/pull/2453/files#diff-a5b5ad66c57c3c303336581e4024b728c144b1ac6a8427bc295d2a821ec25443L298-L301) is assigned the `format_properties.externalFormat` which does not correspond to a vulkan format.
Is it safe to assume that this unknown format will always require a ycbcr sampler?

antonio-lunarg: Yeah, it's a bit weird. Read the description here: https://registry.khronos.org/vulkan/specs/latest/man/html/VkSamplerYcbcrConversionCreateInfo.html#_description
> If the pNext chain includes a [VkExternalFormatANDROID](https://registry.khronos.org/vulkan/specs/latest/man/html/VkExternalFormatANDROID.html) structure with non-zero externalFormat member, the sampler Y′CBCR conversion object represents an external format conversion, and format must be VK_FORMAT_UNDEFINED.

You can imagine that this create info can at this point be called `VkExternalFormatConversionCreateInfo` :D

panos-lunarg: Aha. Fair enough. I wrapped `vkuFormatRequiresYcbcrConversion` with another function that should handle `VK_FORMAT_UNDEFINED` separately

antonio-lunarg: Thanks!

PR #2455, opened 2025-10-10
codegen: Enforce NULL pNext for VU-mandated structs

antonio-lunarg:
This commit aligns encoding with Vulkan valid-usage requirements, preventing serializing non-null pNext where forbidden.

A `generate_null_pnexts.py` is run to generate `null_pnexts.json` based on pNext-nullable structs from Vulka-Headers validusage.json.

`NULL_PNEXTS` are loaded in `KhronosBaseGenerator` and the method `must_extended_struct_be_null` is exposed, overriden in Vulkan generator.

The `KhronosStructEncodersBodyGenerator` is updated to emit `EncodeStructPtrPreamble(nullptr)` when `pNext` must be `NULL`.

### DIFF
--- a/framework/generated/khronos_generators/khronos_base_generator.py
+++ b/framework/generated/khronos_generators/khronos_base_generator.py
@@ -510,21 +510,22 @@ class KhronosBaseGenerator(OutputGenerator):
         self.encode_types['uint64_t'] = 'UInt64'
 
         # Command parameter and struct member data for the current feature
-        self.struct_names = set()                              # Set of current API's struct typenames
-        self.struct_type_names = OrderedDict()                 # Map of current API's struct type enums
-        self.all_extended_structs = dict()                     # Map of all extended struct names
-        self.feature_extended_structs = dict()                 # Map of per-feature extended struct names
-        self.children_structs = dict()                         # Map of children struct names to lists of child struct names
-        self.all_struct_members = OrderedDict()                # Map of struct names to lists of per-member ValueInfo
-        self.feature_struct_members = OrderedDict()            # Map of per-feature struct names to lists of per-member ValueInfo
-        self.all_struct_aliases = OrderedDict()                # Map of struct names to aliases
-        self.feature_struct_aliases = OrderedDict()            # Map of per-feature struct names to aliases
-        self.all_union_members = OrderedDict()                 # Map of union names to lists of per-member ValueInfo
-        self.feature_union_members = OrderedDict()             # Map of per-feature union names to lists of per-member ValueInfo
-        self.all_union_aliases = OrderedDict()                 # Map of union names to aliases
-        self.feature_union_aliases = OrderedDict()             # Map of per-feature union names to aliases
-        self.extension_structs_with_handles = OrderedDict()     # Map of extension struct names to a Boolean value indicating that a struct member has a handle type
+        self.struct_names = set()                                # Set of current API's struct typenames
+        self.struct_type_names = OrderedDict()                   # Map of current API's struct type enums
+        self.all_extended_structs = dict()                       # Map of all extended struct names
+        self.feature_extended_structs = dict()                   # Map of per-feature extended struct names
+        self.children_structs = dict()                           # Map of children struct names to lists of child struct names
+        self.all_struct_members = OrderedDict()                  # Map of struct names to lists of per-member ValueInfo
+        self.feature_struct_members = OrderedDict()              # Map of per-feature struct names to lists of per-member ValueInfo
+        self.all_struct_aliases = OrderedDict()                  # Map of struct names to aliases
+        self.feature_struct_aliases = OrderedDict()              # Map of per-feature struct names to aliases
+        self.all_union_members = OrderedDict()                   # Map of union names to lists of per-member ValueInfo
+        self.feature_union_members = OrderedDict()               # Map of per-feature union names to lists of per-member ValueInfo
+        self.all_union_aliases = OrderedDict()                   # Map of union names to aliases
+        self.feature_union_aliases = OrderedDict()               # Map of per-feature union names to aliases
+        self.extension_structs_with_handles = OrderedDict()      # Map of extension struct names to a Boolean value indicating that a struct member has a handle type
         self.extension_structs_with_handle_ptrs = OrderedDict()  # Map of extension struct names to a Boolean value indicating that a struct member with a handle type is a pointer
+        self.structs_with_null_pnexts = None                     # Set of structure names whose pNext must be null
 
         self.all_cmd_params = OrderedDict()                    # Map of cmd names to lists of per-parameter ValueInfo
         self.feature_cmd_params = OrderedDict()                # Map of cmd names to lists of per-parameter ValueInfo
@@ -1561,6 +1562,23 @@ class KhronosBaseGenerator(OutputGenerator):
     def add_struct_members(self, name, value_info):
         self.all_struct_members[name] = value_info
         self.feature_struct_members[name] = value_info
+
+    def get_or_create_structs_with_null_pnexts(self):
+        """Returns the list of structs that must have null pNext pointers."""
+        if self.structs_with_null_pnexts is None:
+            self.structs_with_null_pnexts = self.struct_names.copy()
+            # The keys in validextensionstructs are base structures that can be extended by structures found in the values.
+            for struct_extended, extension_structs in self.registry.validextensionstructs.items():
+                self.structs_with_null_pnexts.discard(struct_extended)
+                # The extension structs can appear in any order in the pNext chain, therefore they won't have a mandated
+                # NULL pNext value as well.
+                for extension_struct in extension_structs:
+                    self.structs_with_null_pnexts.discard(extension_struct)
+        return self.structs_with_null_pnexts
+
+    def must_extended_struct_be_null(self, struct_name):
+        """Determine if a struct must have a null pNext pointer."""
+        return struct_name in self.get_or_create_structs_with_null_pnexts()
 
     def genUnion(self, typeinfo, typename, alias):
         """Method override.
--- a/framework/generated/khronos_generators/khronos_struct_encoders_body_generator.py
+++ b/framework/generated/khronos_generators/khronos_struct_encoders_body_generator.py
@@ -33,6 +33,10 @@ class KhronosStructEncodersBodyGenerator():
         """Override as needed"""
         return False
 
+    def must_extended_struct_be_null(self, struct_type):
+        """Override as needed"""
+        return False
+
     def write_encoder_content(self):
         api_data = self.get_api_data()
         for struct in self.get_all_filtered_struct_names():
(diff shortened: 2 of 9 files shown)


MarkY-LunarG:
Why are we doing this?  I can see the spec says that a given pNext chain must be NULL, however vendors often do private extensions which may allow non-Null pNext chains for items.  It doesn't seem like something we want to do in GFXR which is intending to copy app behavior even if it is incorrect.  Maybe I'm missing something?

antonio-lunarg:
> Why are we doing this? I can see the spec says that a given pNext chain must be NULL, however vendors often do private extensions which may allow non-Null pNext chains for items. It doesn't seem like something we want to do in GFXR which is intending to copy app behavior even if it is incorrect. Maybe I'm missing something?

It has happened that GFXR was crashing for an uninitialized `pNext` in one title. This also exposed a bug in VVL which has been addressed. https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/10802

My idea here is to be apply a defensive approach to avoid crashing in those cases, but you made a very good point.

So, @MarkY-LunarG , I guess I might revise the approach if we agree on what is the direction to go:
1. Keep codegen as is, overriding selectively `EncodeStruct` for those structure that we want to defend against.
2. Use the defensive approach by default, overriding selectively `EncodeStruct` for those structure that use private extensions.

MarkY-LunarG:
> My idea here is to be apply a defensive approach to avoid crashing in those cases, but you made a very good point.
> 
> So, @MarkY-LunarG , I guess I might revise the approach if we agree on what is the direction to go:
> 
> 1. Keep codegen as is, overriding selectively `EncodeStruct` for those structure that we want to defend against.
> 2. Use the defensive approach by default, overriding selectively `EncodeStruct` for those structure that use private extensions.

Yeah, I'm still worried about overriding private extensions.  Maybe making this global capture variable like "FORCE_NULL_PNEXT" or  something.  @bradgrantham-lunarg, do you think we should go back to the people who brought this up and mention  these concerns or just do an option?

bradgrantham-lunarg:
@MarkY-LunarG - Antonio is working on a known crash in an app with an invalid / possibly uninitialized pNext.  While we want to capture the app behavior, in this case the app does not crash when GFXR capture is not enabled.  In that case, if reasonable, we should try to avoid a crash.

@antonio-lunarg , we had spoken about private extensions before and I thought we had agreed that we can't just assume NULL for pNext even if stated in the spec.  (because technically extensions change the spec and a private extension does too).

Is there any reasonable way to check the address in pNext for rational values, maybe limiting that test to pNext which are mandated NULL in the public spec?

framework/generated/khronos_generators/khronos_struct_encoders_body_generator.py
```diff
@@ -118,9 +122,25 @@ def makeStructBody(self, api_data, name, values, prefix):
...
             if self.is_extended_struct_definition(value):
-                body += '    Encode{}Struct(encoder, {});\n'.format(
-                    api_data.extended_struct_func_prefix, prefix + value.name
-                )
+                if self.must_extended_struct_be_null(name):
+                    # For some structs, spec mandates that pNext must be null.
+                    # On the other hand, vendor extensions may add structures to the pNext chain.
+                    # To handle this, we encode the pNext chain if it is valid, otherwise we
+                    # encode a null pNext pointer.
+                    body += '    if (util::platform::PointerIsValid({}))\n'.format(prefix + value.name)
```
jzulauf-lunarg: Do we want to generate this boilerplate everywhere, or have this case generate a function call or template?

framework/generated/khronos_generators/vulkan_generators/null_pnexts.json
```diff
@@ -0,0 +1,242 @@
+{
+    "structures": [
+        "VkAccelerationStructureBuildGeometryInfoKHR",
```
jzulauf-lunarg: Is there no way to pull this information from the registry?  If not, we can ask for it to be added.  We did this for the synchronization information.  

If this is being pulled from the spec via some scripting, at least specify this was generated from a given spec revision.   If possible give the pattern used to create this list or even check in the tool to run.

antonio-lunarg: > Is there no way to pull this information from the registry?

Yes! I'm generating this file with this script: `framework/generated/khronos_generators/vulkan_generators/generate_null_pnexts.py`
I modified it now to add a `description` and a `generated_by` field to `null_pnexts.json`.

framework/generated/khronos_generators/khronos_struct_encoders_body_generator.py
antonio-lunarg: Extracted into a helper function.

antonio-lunarg:
@MarkY-LunarG , @bradgrantham-lunarg , thanks for your review! I reworked this a bit with a new approach that, for those NULL-mandated pNext, checks whether the pointer is valid before attempting to encode it.
Unfortunately, there is no well-defined / standard way to check whether a pointer is valid, so I'd like your opinions on the new functions introduced in `platform.h`.

framework/generated/khronos_generators/vulkan_generators/generate_null_pnexts.py
```diff
@@ -0,0 +1,153 @@
...
+"""
+
+import os
+import sys
+import json
+from pathlib import Path
+
+NULL_PNEXT_TEXT = (
+    '<code>pNext</code> <strong class="purple">must</strong> be <code>NULL</code>'
+)
```
mikes-lunarg: Generating this list from the validusage.json file seems like the wrong choice here. validusage.json is itself generated from vk.xml, which we already use for codegen. If you enumerate all of the types and track the `structextends` attribute for each struct, those structs that end up with no other structs pointing to them will have NULL pNext values. The other .json files are manually edited "configuration" files and creating a generated one here adds another manual step to codegen.

mikes-lunarg (commented):
> Why are we doing this? I can see the spec says that a given pNext chain must be NULL, however vendors often do private extensions which may allow non-Null pNext chains for items. It doesn't seem like something we want to do in GFXR which is intending to copy app behavior even if it is incorrect. Maybe I'm missing something?

If we find a non-NULL private extension struct we can't do anything useful with it anyway because we won't have a generated struct encoder for the unknown type.

framework/generated/khronos_generators/vulkan_generators/generate_null_pnexts.py
antonio-lunarg: That's super useful. I modified my approach, so now I am trying to derive this "pNext must be NULL" info from the registry, but for some reason I get different results. Maybe not all structures are added to `validextensionstructs`, e.g. I could not find [VkExternalMemoryAcquireUnmodifiedEXT](https://docs.vulkan.org/refpages/latest/refpages/source/VkExternalMemoryAcquireUnmodifiedEXT.html) in that. I haven't managed to figure out why.

bradgrantham-lunarg:
> > Why are we doing this? I can see the spec says that a given pNext chain must be NULL, however vendors often do private extensions which may allow non-Null pNext chains for items. It doesn't seem like something we want to do in GFXR which is intending to copy app behavior even if it is incorrect. Maybe I'm missing something?
> 
> If we find a non-NULL private extension struct we can't do anything useful with it anyway because we won't have a generated struct encoder for the unknown type.

Good point but we do already put "unknown struct type" messages in captures when we encounter a struct type we don't know, and in order to get there we need to know with a little confidence that we can follow the pnext to that struct.

bradgrantham-lunarg:
@jzulauf-lunarg please verify your feedback was addressed and resolve open feedback threads.

framework/generated/khronos_generators/khronos_base_generator.py
```diff
@@ -1565,6 +1577,22 @@ def add_struct_members(self, name, value_info):
         self.all_struct_members[name] = value_info
         self.feature_struct_members[name] = value_info
 
+    def ensure_structs_with_null_pnexts(self):
+        """Ensure that the list of structs that must have null pNext pointers is built."""
+        if self.structs_with_null_pnexts is not None:
+            return
+        self.structs_with_null_pnexts = self.struct_names.copy()
+        for struct_name in self.struct_names:
+            for structs_extended in self.registry.validextensionstructs.values():
```
mikes-lunarg: In addition to the `validextensionstructs.values()` items (which represent the extension structs that will be found in `pNext` chains), I believe you also need to discard the `validextensionstructs.keys()` items (which represent the base structures that are being extended).

framework/generated/khronos_generators/vulkan_generators/generate_null_pnexts.py
mikes-lunarg: See https://github.com/LunarG/gfxreconstruct/pull/2455#discussion_r2457155898

There is also a [list of excluded extension](https://github.com/LunarG/gfxreconstruct/blob/dev/framework/generated/khronos_generators/vulkan_generators/vulkan_base_generator.py#L66) which effectively "don't exist" for the purposes of code generation and could explain some of the differences vs. the original validusage.json scraping technique.

framework/generated/khronos_generators/khronos_base_generator.py
antonio-lunarg: Oh, it's the other way round with respect to the XML. Got it.
Then I guess we should ignore the values, as those structures that can be found in pNext chains may be extended as well.

framework/generated/khronos_generators/vulkan_generators/generate_null_pnexts.py
antonio-lunarg: So I suppose we can safely ignore them.

framework/generated/khronos_generators/khronos_base_generator.py
```diff
@@ -1562,6 +1563,19 @@ def add_struct_members(self, name, value_info):
         self.all_struct_members[name] = value_info
         self.feature_struct_members[name] = value_info
 
+    def get_or_create_structs_with_null_pnexts(self):
+        """Returns the list of structs that must have null pNext pointers."""
+        if self.structs_with_null_pnexts is None:
+            self.structs_with_null_pnexts = self.struct_names.copy()
+            # The keys in validextensionstructs are base structures that can be extended by structures found in the values.
+            for struct_extended in self.registry.validextensionstructs.keys():
```
mikes-lunarg: I *think* we still want to discard the extension structs from `validextensionstructs.values()` (in addition to the base structs) because extension structs appear in pNext chains and therefore won't have spec mandated NULL pNext values either.

antonio-lunarg: I see what you mean now! Thanks, updated with a comment :)

framework/generated/khronos_generators/khronos_base_generator.py
```diff
@@ -1562,6 +1563,23 @@ def add_struct_members(self, name, value_info):
...
 
+    def get_or_create_structs_with_null_pnexts(self):
+        """Returns the list of structs that must have null pNext pointers."""
+        if self.structs_with_null_pnexts is None:
+            self.structs_with_null_pnexts = self.struct_names.copy()
+            # The keys in validextensionstructs are base structures that can be extended by structures found in the values.
+            for struct_extended, extension_structs in self.registry.validextensionstructs.items():
+                self.structs_with_null_pnexts.discard(struct_extended)
+                # The extension structs can appear in any order in the pNext chain, therefore they won't have a mandated
+                # NULL pNext value as well.
```
mikes-lunarg, line 1574: well said 👍

framework/generated/khronos_generators/khronos_struct_encoders_body_generator.py
```diff
@@ -118,9 +122,14 @@ def makeStructBody(self, api_data, name, values, prefix):
         for value in values:
             # pNext fields require special treatment and are not processed by typename
             if self.is_extended_struct_definition(value):
-                body += '    Encode{}Struct(encoder, {});\n'.format(
-                    api_data.extended_struct_func_prefix, prefix + value.name
-                )
+                if self.must_extended_struct_be_null(name):
```
jzulauf-lunarg, line 125: DRY'r to to set "IfValid" in an if_valid string and only emit the encode line once.  But that's preference.

bradgrantham-lunarg:
I believe at this point there are no blockers for merge.

PR #2456, opened 2025-10-11
WIP Parser/Processor

jzulauf-lunarg:
A significant architectural refactoring of the FileProcessor.  Note that to simplify review, the BlockParser::Parse<BlockType> functionality has been left in place where the FileProcessor::Process<BlockType> code was located.  In later commit (or PR) the BlockParser code will be moved to their own file(s).  Note that the ErrHandler support is placeholder (just pointing back to the FileProcessor, and that the BlockHeader parsing is still within FileProcessor. Again further cleanup steps.

### DIFF
--- a/framework/decode/api_payload.h
+++ b/framework/decode/api_payload.h
@@ -1,0 +1,879 @@
+/*
+** Copyright (c) 2025 LunarG, Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and associated documentation files (the "Software"),
+** to deal in the Software without restriction, including without limitation
+** the rights to use, copy, modify, merge, publish, distribute, sublicense,
+** and/or sell copies of the Software, and to permit persons to whom the
+** Software is furnished to do so, subject to the following conditions:
+**
+** The above copyright notice and this permission notice shall be included in
+** all copies or substantial portions of the Software.
+**
+** THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+** IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+** FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+** AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+** FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+** DEALINGS IN THE SOFTWARE.
+*/
+
+#ifndef GFXRECON_API_PAYLOAD_H
+#define GFXRECON_API_PAYLOAD_H
+
+#include "api_decoder.h" // for ApiDecoder method pointers
+
+#include <array>
+#include <memory>
+#include <string>
+#include <tuple>
+#include <variant>
+#include <vector>
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(decode)
+
+// Define threshold for "by value" vs "heap" storage in the variant
+struct CommandStoragePolicy
+{
+    // This size is picked to match FunctionCallArgs and MethodCallArgs which constitute more than 95% of total blocks
+    static constexpr size_t kInlineThreshold = 48;
+    template <typename T>
+    static constexpr bool IsLarge()
+    {
+        return sizeof(T) > kInlineThreshold;
+    }
+};
+
+template <typename Command, typename Dummy = void>
+struct DispatchHasCallId : std::false_type
+{};
+
+template <typename Command>
+struct DispatchHasCallId<Command, std::void_t<decltype(std::declval<Command>().call_id)>> : std::true_type
+{};
+
+template <typename Command, typename Dummy = void>
+struct DispatchHasMetaDataId : std::false_type
+{};
+
+template <typename Command>
+struct DispatchHasMetaDataId<Command, std::void_t<decltype(std::declval<Command>().meta_data_id)>> : std::true_type
+{};
+
+template <typename Command>
+struct DispatchHasAllocGuard : std::false_type
+{};
+
+// ---- Flags base (one place for size and per-command booleans) ----------------
+//
+// The flag traits class is a base class for each of the DispatchTraits specializations
+// defined below.  The flags control the storage and dispatch characteristics of the
+// Command type.
+template <typename Command>
+struct DispatchFlagTraits
+{
+    static constexpr bool kIsLarge       = CommandStoragePolicy::IsLarge<Command>();
+    static constexpr bool kHasCallId     = DispatchHasCallId<Command>::value;
+    static constexpr bool kHasMetaDataId = DispatchHasMetaDataId<Command>::value;
+    static constexpr bool kHasAllocGuard = DispatchHasAllocGuard<Command>::value;
+};
+
+// --- Payload structs (argument order preserved) ---
+//
+// Each of the Info structures match the type and order of an ApiDecoder dispatch call.
+// std::apply is used to pass references to the Info contents to the kDecoderMethod defined
+// in the DispatchTraits object defined for the each of the Info structs.
+struct FunctionCallArgs
+{
+    format::ApiCallId call_id;
+    ApiCallInfo       call_info;
+    const uint8_t*    buffer;
+    size_t            buffer_size;
+
+    auto GetTuple() const { return std::tie(call_id, call_info, buffer, buffer_size); }
+};
+template <>
+struct DispatchHasAllocGuard<FunctionCallArgs> : std::true_type
+{};
+
+struct MethodCallArgs
+{
+    format::ApiCallId call_id;
+    format::HandleId  object_id;
+    ApiCallInfo       call_info;
+    const uint8_t*    parameter_buffer;
+    size_t            buffer_size;
+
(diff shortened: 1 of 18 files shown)


framework/decode/api_payload.h
```diff
@@ -0,0 +1,790 @@
...
+template <typename Command>
+struct CommandFlagTraits
+{
+    static constexpr bool kIsLarge       = CommandStoragePolicy::IsLarge<Command>();
+    static constexpr bool kHasCallId     = CommandHasCallId<Command>::value;
+    static constexpr bool kHasMetaDataId = CommandHasMetaDataId<Command>::value;
+    static constexpr bool kHasAllocGuard = CommandHasAllocGuard<Command>::value;
+};
+
+// --- Payload structs (argument order preserved) ---
```
ncesario-lunarg: This might be documented elsewhere, but it might be good to have a comment in here about how `GetTuple` "ties" to downstream function invocation.

framework/decode/file_processor.h
```diff
@@ -290,17 +543,122 @@ class FileProcessor
...
+            for (auto decoder : decoders_)
+            {
+                if (DecoderSupportsCommand(*decoder, command))
+                {
+                    [[maybe_unused]] DecoderAllocGuard<CommandTraits<Command>::kHasAllocGuard> alloc_guard{};
+                    SetDecoderApiCallId(*decoder, command);
+                    auto dispatch_call = [&decoder, decode_method](auto&&... args) {
+                        (decoder->*decode_method)(std::forward<decltype(args)>(args)...);
+                    };
+                    std::apply(dispatch_call, command.GetTuple());
```
ncesario-lunarg: Pretty slick!

framework/decode/file_processor.h
```diff
@@ -121,27 +184,221 @@ class BlockBuffer
...
+        }
+    };
+
+    template <typename Visitor>
+    void Visit(Visitor& visitor) const
+    {
+        if (state_ != BlockState::kSkip)
+        {
+            GFXRECON_ASSERT((state_ == BlockState::kReady) || (state_ == BlockState::kDelayedDecompress));
+            std::visit(DereferenceVisitorWrapper<Visitor>{ *this, visitor }, args_);
```
ncesario-lunarg: Probably not necessary to add a comment, but it wasn't obvious to me why the struct was necessary (as opposed to a lambda here). I'm assuming `DereferenceVisitorWrapper` exists to handle the case of a pointer dereference (hence the overload)? If that's the case, would it instead make sense to have the overloads in `DereferenceVisitorWrapper` be based off of `std::is_pointer`, rather than be specific to `std::unique_ptr`? Why restrict this to `unique_ptr`s?

framework/decode/file_processor.h
```diff
@@ -121,27 +184,221 @@ class BlockBuffer
...
+//   * Refactor UncompressedStore:
+//       - Perhap refactor Entry to hide pool as prefix to the allocation
+//       - Perhaps hide HeapBuffer size as hidden prefix
+//       - Expected saving: 8-16 B
+//
+//   * Keep CommandArgs (56 B) for now:
+//       - Size matches ~95% of blocks.
+//       - 8-byte variant overhead simplifies anonymized usage.
+// -----------------------------------------------------------------------------
+class ParsedBlock
```
ncesario-lunarg, line 229: Since this class is new, it might make sense to pull this out into another file from the start?

framework/decode/file_processor.h
```diff
@@ -121,27 +184,221 @@ class BlockBuffer
...
+    UncompressedStore uncompressed_store_;
+
+    // Variant of all parsed results
+    CommandArgs args_; // Variant with a type decoded block
+    BlockState  state_ = BlockState::kInvalid;
+};
+
+// WIP WIP WIP:
+class FileProcessor;
+using BufferPool = util::HeapBufferPool::PoolPtr; // TODO: Get a better pool, and share with FileInputStream
```
ncesario-lunarg, line 327: https://en.cppreference.com/w/cpp/memory/synchronized_pool_resource.html

I haven't looked at your HeapBufferPool implementation, but starting with C++17, there are now pool allocators that can be backed by stack memory (though that may not be necessary).

framework/util/file_input_stream.h
```diff
@@ -44,39 +45,102 @@ struct DataRange
...
+    [[nodiscard]] constexpr size_type size() const noexcept { return size_; }
+    [[nodiscard]] constexpr bool      empty() const noexcept { return size_ == 0U; }
+
+    pointer       GetData() noexcept { return data_; }
+    const pointer GetData() const noexcept { return data_; }
+    bool          IsValid() const noexcept { return !empty(); }
+    size_type     Size() const noexcept { return size(); }
+
+    template <typename U>
+    [[nodiscard]] const U* GetDataAs() const noexcept
```
ncesario-lunarg: Are these being used? It would probably be good to stick with the STL interface. Otherwise, we'll needs some sort of wrapping if we want to switch over to the C++20 one.

framework/decode/file_processor.h
jzulauf-lunarg: The CmdStore is either the class by value (if not kIsLarge) or a unique_ptr (if kIsLarge).  This is not a general solution, but specific to the CommandArgs comprising CmdStore variants.

framework/decode/file_processor.h
jzulauf-lunarg: Yes.  The BlockParser and ParserBlocks will move out of file_processor.*

framework/decode/file_processor.h
jzulauf-lunarg: Haven't reviewed.  The whole allocator approach needs to be looked at more carefully, especially for the preload-measurement-range usage.  HeapBufferPool is effectively a placeholder until we do.  There are several open-source possibilities as well.

framework/util/file_input_stream.h
jzulauf-lunarg: Yes.  We should make a project-wide decision on naming.  STL doesn't have a "get_data_as" interface, but for lazy people like me... it's perfect :)

The Span and DataSpan (and is there a better name?) also should be moved to there own files.

jzulauf-lunarg:
Passes extended CI suite tests.

framework/decode/api_payload.h
```diff
@@ -0,0 +1,635 @@
...
+{
+    format::ThreadId thread_id;
+    format::HandleId surface_id;
+    uint32_t         width;
+    uint32_t         height;
+    uint32_t         pre_transform;
+
+    auto GetTuple() const { return std::tie(thread_id, surface_id, width, height, pre_transform); }
+};
+struct CreateHardwareBufferInfo
```
MarkY-LunarG: Should this be wrapped in #ifdef ANDROID?

MarkY-LunarG: Maybe not if we want to handle this in a portable way when replaying Android captures on desktop?

framework/decode/file_processor.h
```diff
@@ -122,28 +145,168 @@ class BlockBuffer
...
+
+    void SetFrameNumber(uint64_t frame_number) noexcept { frame_number_ = frame_number; }
+    void SetBlockIndex(uint64_t block_index) noexcept { block_index_ = block_index; }
+
+    [[nodiscard]] uint64_t GetFrameNumber() const noexcept { return frame_number_; }
+    [[nodiscard]] uint64_t GetBlockIndex() const noexcept { return block_index_; }
+
+    // Define parsers for every block and sub-block type
+    ParsedBlock ParseFunctionCall(BlockBuffer& block_buffer);
+#if 0
```
MarkY-LunarG: Nit: #if 0 without a TODO to remove.

framework/decode/api_payload.h
```diff
@@ -184,7 +226,7 @@ struct SetDevicePropertiesInfo
     uint32_t         vendor_id;
     uint32_t         device_id;
     uint32_t         device_type;
-    uint8_t          pipeline_cache_uuid;
+    uint8_t          pipeline_cache_uuid[format::kUuidSize];
```
MarkY-LunarG: Yikes, are we using this and didn't catch this?

framework/decode/api_payload.h
jzulauf-lunarg: It's not wrapped in the original code.  Keeping consistency.

framework/decode/file_processor.cpp
```diff
@@ -429,25 +429,21 @@ bool FileProcessor::ProcessBlocks()
                 }
                 else if (base_type == format::BlockType::kFunctionCallBlock)
                 {
-                    format::ApiCallId api_call_id = format::ApiCallId::ApiCall_Unknown;
-
-                    success = block_buffer.Read(api_call_id);
-
-                    if (success)
+                    block_parser.SetBlockIndex(block_index_);
```
MarkY-LunarG, line 438: NOTE: There is a change from one of the ARM guys changing the frame end functionality.  You should see what's going on there so you're aware of upcoming changes.

framework/decode/api_payload.h
jzulauf-lunarg: This is a clean up of an earlier commit in this PR.  This will be squashed out.

framework/util/file_input_stream.h
```diff
@@ -44,39 +45,102 @@ struct DataRange
...
 // NOTE: Only one of the available sources will be active at any time.
 // NOTE: This class is *NOT* copyable, only movable
 // NOTE: Additional constructors could be added to support other data sources
 //
-//
-// NOTE: Currently only non-owning subspans are supported, but require explicit use of a tag type
+// TODO: Some housekeeping
+//       Add capacity assert in DataSpan(HeapBuffer&&, size_type).
+//       Decide if empty() should reflect size_ == 0 rather than !IsValid(). Document.
 class DataSpan
```
nickdriscoll-lunarg, line 122: If the difference between DataSpan and Span is that they own and don't own the data, respectively, then would OwnedSpan be a better name for DataSpan? After all, read-only or not, they're both spans of data...

jzulauf-lunarg:
Closed in favor of cleaner branch https://github.com/LunarG/gfxreconstruct/pull/2484

PR #2458, opened 2025-10-13
Fix sparse rebind support

marius-pelegrin-arm:
Currently, sparse rebind is implemented in the following way:

When a `vkQueueBindSparse` call is encountered, each memory binding is "translated" and bound in a single `vkQueueBindSparse` call. They all wait on the original call's semaphore, and they all signal a different semaphore, which is waited on by the last bind call.

First, this implementation goes against the Vulkan specification:
"The specific signal operation waited on for each semaphore must meet the following criteria:
[...]
- for binary semaphores, no wait operation exists that happens-after the signal operation and happens-before this wait operation"

In other words, it is not allowed for a binary semaphore to be waited on multiple times by multiple `vkQueueBindSparse` calls.

Second, this implementation is wildly inneficient and performance hungry. As a single `vkQueueBindSparse` call from the application can result in thousands of calls at replay time. In addition, fixing the semaphore issue would probably lead in a forced serialization of these calls, making the performances even worse.

Both these issues can be tested on the `sparse_image` sample from the Khronos Group.

This commit aims at fixing this. The idea is to not introduce any synchronization nor new calls, and simply match one `vkQueueBindSparse` call at capture time with one "translated" call at replay time.

### DIFF
--- a/framework/util/zlib_compressor.cpp
+++ b/framework/util/zlib_compressor.cpp
@@ -21,6 +21,7 @@
 ** DEALINGS IN THE SOFTWARE.
 */
 
+#include "util/logging.h"
 #ifdef GFXRECON_ENABLE_ZLIB_COMPRESSION
 
 #include "util/zlib_compressor.h"
@@ -29,6 +30,8 @@
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(util)
+
+static constexpr size_t CHUNK = 256;
 
 size_t ZlibCompressor::Compress(const size_t          uncompressed_size,
                                 const uint8_t*        uncompressed_data,
@@ -57,16 +60,41 @@ size_t ZlibCompressor::Compress(const size_t          uncompressed_size,
     compress_stream.next_in  = const_cast<Bytef*>(uncompressed_data);
 
     GFXRECON_CHECK_CONVERSION_DATA_LOSS(uInt, compressed_data->size());
-    compress_stream.avail_out = static_cast<uInt>(compressed_data->size());
+    compress_stream.avail_out = static_cast<uInt>(compressed_data->size() - compressed_data_offset);
     compress_stream.next_out  = compressed_data->data() + compressed_data_offset;
 
-    // Perform the compression (deflate the data).
+    // Initialize the compressor
     deflateInit(&compress_stream, Z_BEST_COMPRESSION);
-    deflate(&compress_stream, Z_FINISH);
+
+    while (true)
+    {
+        // Perform the compression (deflate the data).
+        int ret = deflate(&compress_stream, Z_FINISH);
+
+        if (ret == Z_STREAM_ERROR)
+        {
+            GFXRECON_LOG_ERROR("%s() There was an error while compressing.")
+            return 0;
+        }
+
+        // This means that the compression is complete
+        if (ret == Z_STREAM_END)
+        {
+            break;
+        }
+
+        // deflate() has exhausted the out buffer. Allocate some more and call deflate again
+        const size_t current_size = compressed_data->size();
+        compressed_data->resize(current_size + CHUNK);
+        compress_stream.avail_out = static_cast<uInt>(CHUNK);
+        compress_stream.next_out  = compressed_data->data() + current_size;
+    }
+
     deflateEnd(&compress_stream);
 
     // Determine the size of data from the stream
     copy_size = compress_stream.total_out;
+    compressed_data->resize(copy_size);
 
     return copy_size;
 }


framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -3071,87 +2734,434 @@ VkResult VulkanRebindAllocator::QueueBindSparse(VkQueue                 queue,
...
+                                      original_buffer_bind_info.pBinds + original_buffer_bind_info.bindCount);
+
+            for (uint32_t buf_mem_i = 0; buf_mem_i < original_buffer_bind_info.bindCount; ++buf_mem_i)
+            {
+                const VkSparseMemoryBind& original_memory_bind = original_buffer_bind_info.pBinds[buf_mem_i];
+                VkSparseMemoryBind&       modified_memory_bind = memory_binds.back()[buf_mem_i];
+
+                auto res_alloc_info = reinterpret_cast<ResourceAllocInfo*>(allocator_buf_datas[alc_buf_i]);
+                auto mem_alloc_info = reinterpret_cast<MemoryAllocInfo*>(allocator_buf_mem_datas[alc_buf_mem_i]);
+                VmaMemoryInfo*& vma_mem_info = vma_memory_infos.emplace_back(nullptr);
```
MarkY-LunarG: What is this doing?

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -3071,87 +2734,434 @@ VkResult VulkanRebindAllocator::QueueBindSparse(VkQueue                 queue,
...
+                {
+                    VkMemoryRequirements replay_req                    = {};
+                    bool                 requires_dedicated_allocation = false;
+                    bool                 prefers_dedicated_allocation  = false;
+                    allocator_->GetBufferMemoryRequirements(original_buffer_bind_info.buffer,
+                                                            replay_req,
+                                                            requires_dedicated_allocation,
+                                                            prefers_dedicated_allocation);
+
+                    replay_req.size = original_memory_bind.size;
```
MarkY-LunarG: I noticed the alignment was modified in the old code, is this no longer valid or done somewhere else?

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -3071,87 +2734,434 @@ VkResult VulkanRebindAllocator::QueueBindSparse(VkQueue                 queue,
...
+                        continue;
+                    }
+
+                    GFXRECON_ASSERT(vma_mem_info);
+                    modified_memory_bind.memory = vma_mem_info->allocation_info.deviceMemory;
+                    modified_memory_bind.memoryOffset =
+                        GetRebindOffsetFromOriginalDeviceMemory(original_memory_bind.memoryOffset, *vma_mem_info);
+                }
+
+                if (allocator_->m_UseMutex && vma_mem_info != nullptr && vma_mem_info->allocation != nullptr &&
```
MarkY-LunarG: You have this block of code repeated 3 times.

1) Can you make a function for it?
2) Can you give more detail on what it's doing (i.e. why is it grabbing block info from VMA)?

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -3071,87 +2734,434 @@ VkResult VulkanRebindAllocator::QueueBindSparse(VkQueue                 queue,
...
+        modified_bind_info.imageBindCount       = modified_image_bind_infos[i].size();
+        modified_bind_info.pImageBinds          = modified_image_bind_infos[i].data();
+    }
+
+    for (VmaDeviceMemoryBlock* block : vma_mem_blocks)
+    {
+        block->m_MapAndBindMutex.Lock();
+    }
+
+    result = functions_.queue_bind_sparse(queue, modified_bind_infos.size(), modified_bind_infos.data(), fence);
```
MarkY-LunarG, line 3002: The old path did some semaphore magic.  Is that no longer necessary?

framework/decode/vulkan_rebind_allocator.cpp
marius-pelegrin-arm: It adds a nullptr to the vector and create a reference to it (vma_mem_info) that may then be assigned to a non-null value by VmaAllocateMemory.

This is used at the end of the function for the calls to UpdateAllocInfo

framework/decode/vulkan_rebind_allocator.cpp
marius-pelegrin-arm: Good catch, this was removed by accident.

framework/decode/vulkan_rebind_allocator.cpp
marius-pelegrin-arm: This was formerly the lock in ProcessSingleQueueBindSparse.
We still need to lock these memory blocks, but we cannot lock them one by one as we go in the for loop. In addition the same block may be used for several of these sub-allocations. So I fill a vector of "blocks to lock" and lock them just before the call to vkQueueBindSparse.

This can be factored into a function, but this function has no purpose in itself so naming it will be very opaque...

"VulkanRebindAllocator::AddAllocationBlockToLockVector" ?

How would you name such a function ?

locke-lunarg (approved):
LGTM! Thank you for fixing this issue.

MarkY-LunarG:
@marius-pelegrin-arm , I pushed up a change that reduces duplication here.

marius-pelegrin-arm:
I had a quick look and gave it a test, it seems good, thanks you =)

PR #2459, opened 2025-10-13
Convert FileProcess non-owning spans to span backport

jzulauf-lunarg:
- Add backported Span (lite) and use in file processing
- Updated contributing to add exception for STL-style for classes requiring support from range-for and STL Algorithms
- Change compressor interface, takes data() not container
- Also improve type safety of the casting getter with some type traits utilities.

### DIFF
--- a/framework/decode/file_processor.cpp
+++ b/framework/decode/file_processor.cpp
@@ -675,7 +675,7 @@ BlockBuffer::BlockSpan FileProcessor::ReadCompressedParameterBuffer(BlockBuffer&
         size_t uncompressed_size = compressor_->Decompress(compressed_buffer_size,
                                                            reinterpret_cast<const uint8_t*>(compressed_span.data()),
                                                            expected_uncompressed_size,
-                                                           &uncompressed_buffer_);
+                                                           uncompressed_buffer_.data());
         if ((0 < uncompressed_size) && (uncompressed_size == expected_uncompressed_size))
         {
             return BlockBuffer::BlockSpan(reinterpret_cast<const std::byte*>(uncompressed_buffer_.data()),
--- a/framework/decode/file_transformer.cpp
+++ b/framework/decode/file_transformer.cpp
@@ -380,7 +380,7 @@ bool FileTransformer::ReadCompressedParameterBuffer(size_t  compressed_buffer_si
         size_t uncompressed_size = compressor_->Decompress(compressed_buffer_size,
                                                            compressed_parameter_buffer_.data(),
                                                            expected_uncompressed_size,
-                                                           &parameter_buffer_);
+                                                           parameter_buffer_.data());
         if ((0 < uncompressed_size) && (uncompressed_size == expected_uncompressed_size))
         {
             *uncompressed_buffer_size = uncompressed_size;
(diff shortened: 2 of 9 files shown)


framework/util/file_input_stream.h
```diff
@@ -44,7 +45,73 @@ struct DataRange
...
+    }
+
+    Span(Span&&)                 = default;
+    Span(const Span&)            = default;
+    Span& operator=(Span&&)      = default;
+    Span& operator=(const Span&) = default;
+
+    // Add other constructors as needed (c.f. YAGNI)
+
+    constexpr iterator begin() noexcept { return data_; }
```
MarkY-LunarG, line 81: I'm not a big fan of the type name override, but I can understand your reasoning.  So not going to request a change, just mention I, personally, would rather see the original T* instead of "pointer".

framework/util/file_input_stream.h
```diff
@@ -44,7 +45,73 @@ struct DataRange
...
+    constexpr pointer       data() noexcept { return data_; }
+    constexpr const_pointer data() const noexcept { return data_; }
+
+    [[nodiscard]] constexpr size_type size() const noexcept { return size_; }
+    [[nodiscard]] constexpr bool      empty() const noexcept { return size_ == 0U; }
+
+    pointer       GetData() noexcept { return data_; }
+    const pointer GetData() const noexcept { return data_; }
+    bool          IsValid() const noexcept { return !empty(); }
+    size_type     Size() const noexcept { return size(); }
```
MarkY-LunarG: Why this and `size()`?

framework/util/file_input_stream.h
jzulauf-lunarg: this is from std::span, the typedefs are part of the interface

framework/util/file_input_stream.h
jzulauf-lunarg: We haven't made a definitive statement about our style variances from the Google Style Guide.  size is to match the backport of std:span, Size to match the styleguide.  @bradgrantham-lunarg ?

CONTRIBUTING.md
```diff
@@ -211,7 +211,11 @@ to differentiate the work from other people working in the repo.
 
 Changes to the GFXReconstruct project's C++ code should conform to the coding
 style defined by the
-[Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html).
+[Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html). The exception
```
jzulauf-lunarg: @bradgrantham-lunarg Here's the updated style language.

mikes-lunarg: This reads fine as is, but "There is an exception for custom containers..." would be a little more natural IMO. Also, there is a stray space character in ` begin`

The larger question I have is if this note is even required. The Google C++ Style Guide already lists an exception for STL-like interfaces: https://google.github.io/styleguide/cppguide.html#Exceptions_to_Naming_Rules

jzulauf-lunarg: @mikes-lunarg revised to reflect.

mikes-lunarg (approved):

PR #2463, opened 2025-10-16
VDR: Move VDR option from CLI into the input json

panos-lunarg:
All Vulkan dump resources options except from `--dump-resources-dir` are
removed from command line and are moved in the input json. A json
example:
```Json
{
    "DumpResourcesOptions": {
      "Scale": 1.0,
      "ImageFormat": "bmp",
      "OutputDir": "",
      "ColorAttachmentIndex": -1,
      "DumpBeforeCommand": false,
      "DumpDepth": false,
      "DumpVertexIndexBuffer": false,
      "DumpAllDescriptors": true,
      "DumpAllImageSubresources": false,
      "DumpRawImages": false,
      "DumpSeparateAlpha": false,
      "DumpUnusedVertexBindings": false,
      "JsonOutputPerCommand": false,
      "DumpBuildAccelerationStructuresInputBuffers": true,
      "BinaryFileCompressionType": "None"
    },

    "BeginCommandBuffer": [ 587 ],
    "TraceRays": [[ 590 ]],
    "QueueSubmit": [ 611 ]
}
```
`--dump-resources-dir` is left as a command line option (but is also
available as a json option with json having a higher priority) as it 
provides flexibility for CI testing.


bradgrantham-lunarg:
@roberto-c

roberto-c:
Looks good to me. Probably it is already being worked on, but adding a sample JSON input in the documentation may be helpful.

Giving that there are options to select which kind of resources to dump, it may be good to consider a bit-mark approach like: 

```
// Each type of resource we would like to dump OR'ed
"resource_dump_type" : "Color | Depth | Stencil | AS | IB"
// ALL_RESOURCES as shortcut
"resource_dump_type" : "ALL_RESOURCES"
```

This may help to reduce the number of arguments to specify which resources to dump and allow adding more type of resources using the same argument in the future.

panos-lunarg:
@roberto-c 

Yes the documentation will be updated accordingly.
One change that we plan on doing with this PR and still not reflected in the code is to remove the 2 of the currently 3 available ways of describing the command indices. These are
1. With the input .json (i.e. `--dump-resources input.json`)
2. With a .txt (i.e. `--dump-resources input.txt`). The txt is expected to contain a simplified format
3. Directly via the the CL options (i.e. `--dump-resources BeginCommandBuffer=105,Draw=111,BeginRenderPass=106,EndRenderPass=112,QueueSubmit=143`)

If all other dump resources cli options are moved in the input json file then options 2 and 3 suddenly become very limited (apart from their inherent limitations in describing more complex command indices inputs) as you can no longer provide the rest of the dump resources options.

If you are using any of the last two options this is a good chance to raise any objections about removing them.

The bit-mark approach for describing which kind of resources to be dumped is interesting and we can add it in a separate PR.

roberto-c:
> ... One change that we plan on doing with this PR and still not reflected in the code is to remove the 2 of the currently 3 available ways of describing the command indices. These are
> 
>     1. With the input .json (i.e. `--dump-resources input.json`)
> 
>     2. With a .txt (i.e. `--dump-resources input.txt`). The txt is expected to contain a simplified format
> 
>     3. Directly via the the CL options (i.e. `--dump-resources BeginCommandBuffer=105,Draw=111,BeginRenderPass=106,EndRenderPass=112,QueueSubmit=143`)
> 
> 
> If all other dump resources cli options are moved in the input json file then options 2 and 3 suddenly become very limited (apart from their inherent limitations in describing more complex command indices inputs) as you can no longer provide the rest of the dump resources options.
> 
> If you are using any of the last two options this is a good chance to raise any objections about removing them.
> 

I think we should be fine if the command line options are removed. We are using the input JSON file to specify what we want to dump.

panos-lunarg:
@roberto-c FYI

roberto-c:
> @roberto-c FYI

LGTM.

USAGE_android.md
```diff
@@ -975,48 +965,12 @@ options:
                         Same as "--swapchain captured". Ignored if the "--
                         swapchain" option is used.
   --dump-resources DUMP_RESOURCES
```
davidlunarg: Arg should be <filename.json>, consistent with line 784?

tools/replay/parse_dump_resources_cli.cpp
```diff
@@ -330,303 +457,103 @@ bool parse_dump_resources_arg(gfxrecon::decode::VulkanReplayOptions& vulkan_repl
         return true;
     }
 
+    // Vulkan dump resources expectes only a json file
```
davidlunarg: Spelling

davidlunarg: The variable parse_error is set to false at the start of this method, never changed, and the value is compared at the end of this method. Should probably not declare or use it.

tools/replay/replay_settings.h
```diff
@@ -334,51 +316,9 @@ static void PrintUsage(const char* exe_name)
...
-    GFXRECON_WRITE_CONSOLE("          \t\tDump a vertex binding even if no vertex attributes references it.");
-    GFXRECON_WRITE_CONSOLE("  --dump-resources-binary-file-compression-type");
-    GFXRECON_WRITE_CONSOLE("          \t\tCompress files that are dumped as binary. Available compression types");
-    GFXRECON_WRITE_CONSOLE("          \t\tare: [none, lz4 (block format), zlib, zstd]. Default is none");
-    GFXRECON_WRITE_CONSOLE("          \t\t(no compression).");
-    GFXRECON_WRITE_CONSOLE("  --dump-resources-dump-build-acceleration-structures-input-buffers");
-    GFXRECON_WRITE_CONSOLE("          \t\tDump all input buffers used in vkCmdBuildAccelerationStructures. This");
-    GFXRECON_WRITE_CONSOLE("          \t\tincludes vertex, index, transformation matrix, AABB and instance buffers.");
-    GFXRECON_WRITE_CONSOLE("          \t\tDefault is off.");
+    GFXRECON_WRITE_CONSOLE("  --dump-resources <json file>");
```
davidlunarg: Use <filename>.json to be consistent?

vulkan_dump_resources.md
```diff
@@ -315,72 +316,94 @@ Image descriptors can be fine grained further by specifying the desired subresou
...
+The rest of the dump resources options are available through the input json file. They are expected to be inside a json object named `DumpResourcesOptions`.
+The possible options are the following:
+
+* `Scale` <scale>: Scale images generated by dump resources by the given scale factor. The scale factor must be a floating point number greater than 0. Values greater than 10 are capped at 10. Default value is 1.0.
+
+* `ImageFormat` <format>:
+Image file format to use for image resource dumping. Available formats are: `bmp` and `png`. Default is `bmp`.
+
+* `DumpBeforeCommand` <`true` | `false`>:
+In addition to dumping gpu resources after the CmdDraw, CmdDispatch and CmdTraceRays calls also dump resources before those calls.
```
davidlunarg: Omit "<true|false>.   Change to: "In addition to dumping gpu resources after the CmdDraw, CmdDispatch and CmdTraceRays calls, also dump resources before those calls.

vulkan_dump_resources.md
```diff
@@ -315,72 +316,94 @@ Image descriptors can be fine grained further by specifying the desired subresou
...
+
+* `DumpDepth`
+Configures whether to dump the depth attachment when dumping draw calls. Default is disabled.
```
davidlunarg: Change to: "Enables dumping of the depth attachment when dumping draw calls. Default is disabled."

vulkan_dump_resources.md
davidlunarg: The type of the value expected for each of these keys isn't clearly defined. They can be inferred, but should be specified. Maybe the description of each should begin something like: "Boolean value specifying whether ..." or "Integer value specifiying the..."

vulkan_dump_resources.md
```diff
@@ -315,72 +316,94 @@ Image descriptors can be fine grained further by specifying the desired subresou
...
+
+* `ColorAttachmentIndex` <index>
```
davidlunarg: Omit "< index>"?

vulkan_dump_resources.md
```diff
@@ -315,72 +316,94 @@ Image descriptors can be fine grained further by specifying the desired subresou
...
+
+* `DumpUnusedVertexBindings`
+Bound vertex buffer bindings are filtered based on the currently active graphics pipeline. Setting this option to true will dump all bound vertex buffers
+
+* `JsonOutputPerCommand`
+Enables storing a json output file for each dumped command. Overrides default behavior which
+is generating one output json file that contains the information for all dumped commands.
+
+* `BinaryFileCompressionType`
+Compress files that are dumped as binary. Available compression types are: [`none`, `lz4` (block format), `zlib`, `zlib`, `zstd`]. Default is `none` (no compression).
```
davidlunarg: Change to: "Boolean value specifying whether files that are dumped as binary files are to be compressed. ..."

davidlunarg (changes requested):

vulkan_dump_resources.md
panos-lunarg: Changed it in the option's type: `compression method`

vulkan_dump_resources.md
panos-lunarg: Changed it in the option's type: `unsigned integer`

vulkan_dump_resources.md
panos-lunarg: fixed

vulkan_dump_resources.md
panos-lunarg: Very good point. I decided to add the type in all parameters so it is more clear with a quick glance. Also added at the end of each description the option's default value.

tools/replay/replay_settings.h
panos-lunarg: fixed

tools/replay/parse_dump_resources_cli.cpp
panos-lunarg: fixed

panos-lunarg: fixed

USAGE_android.md
panos-lunarg: fixed

PR #2465, opened 2025-10-16
encode: Fix buffer size calculation for image copy

antonio-lunarg:


framework/graphics/vulkan_resources_util.cpp
```diff
@@ -2647,5 +2647,113 @@ VkResult VulkanResourcesUtil::BlitImage(VkCommandBuffer       command_buffer,
...
+ * @return VkDeviceSize The number of bytes that must be available from region.pHostPointer
+ *                      to satisfy the copy described by 'region'.
+ *
+ * @details
+ * - Handles depth/stencil special cases per the Vulkan specification.
+ * - Accounts for block-compressed formats by converting to texel-block units and rounding up for partial blocks.
+ * - Returns 0 for invalid/empty copies; callers should already have guards for those cases.
+ */
+template <typename RegionCopy>
+VkDeviceSize GetBufferSizeFromCopyImage(const RegionCopy& region, uint32_t array_layers, VkFormat format)
```
panos-lunarg: This looks like a static function. Let's use `static` in the definition as well.

Nitpick: There is a similar function `GetImageResourceSizesOptimal` which calculates the size required for an image but for the gpu/implementation side. The added documentation here is great and helps a lot to tell what this new function does but I think there might still be some confusion about how the other function is different. Can we do something about that?

antonio-lunarg: I am not sure they do the same thing. This function is more focused on the host memory size used in a (e.g.) `VkMemoryToImageCopy`, while the other one it seems to me that it computes a total memory size required by a `VkImage`. And it is possible that the result might not be the same, especially if the copy region is only a portion of the total image size.

PR #2466, opened 2025-10-17
Better staging-buffer sizes for VulkanResourceInit

fabian-lunarg:
- apply changes to data stored in `BeginResourceInitCommand`
- staging-buffer size-bounds are derived from largest resource & total sum of staging-resource (vs. just largest resource)
- can yield larger and more optimal staging-buffer size, **reducing number of required queue-submits**
- query memory-heap-sizes for staging-mem, in order to perform a sanity-check

also:
- sprinkles some c++20 refactoring of loop-iteration, using **ranges**
- VulkanAddressReplacer: correct headroom for buffer-sizes w. offset-alignment.

### DIFF
--- a/framework/encode/vulkan_state_writer.cpp
+++ b/framework/encode/vulkan_state_writer.cpp
@@ -2940,18 +2940,15 @@ void VulkanStateWriter::WriteBufferMemoryState(const VulkanStateTable& state_tab
             // method requires the buffer to be bound to a single range of a single memory, which is not applicable for
             // sparse buffers. Therefore, we set the two values to use staging copy for dumping sparse buffers.
             snapshot_info.memory_properties = VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT;
-            snapshot_info.need_staging_copy = true; // Staging copy is needed for sparse buffer.
 
-            if ((*total_staging_copy_size) < wrapper->size)
-            {
-                (*total_staging_copy_size) = wrapper->size;
-            }
+            // Staging copy is needed for sparse buffer.
+            snapshot_info.need_staging_copy = true;
+            (*total_staging_copy_size) += wrapper->size;
 
-            if (snapshot_info.need_staging_copy && ((*max_staging_copy_size) < wrapper->size))
+            if ((*max_staging_copy_size) < wrapper->size)
             {
                 (*max_staging_copy_size) = wrapper->size;
             }
-
             snapshot_entry.buffers.emplace_back(snapshot_info);
         }
     });


fabian-lunarg:
the android-builds are currently failing due to a mismatch in ndk-versions for layer (26) and tools (22). fyi @bradgrantham

bradgrantham:
> the android-builds are currently failing due to a mismatch in ndk-versions for layer (26) and tools (22). fyi @bradgrantham

@bradgrantham-lunarg

framework/decode/vulkan_resource_initializer.cpp
```diff
@@ -186,15 +195,12 @@ VkResult VulkanResourceInitializer::InitializeBuffer(VkDeviceSize        data_si
                     offsetted_regions_copy_[i].srcOffset += staging_buffer_offset_;
                 }
 
-                if (result == VK_SUCCESS)
```
MarkY-LunarG, line 189: Why was this removed?

fabian-lunarg: was redundant 2x nesting of `if (result == VK_SUCCESS)`

fabian-lunarg: ah hold on, good point.

edit:
yes, was doubled up.

we got this above and no other change to `result` after:
```
result = BeginCommandBuffer(queue_family_index, &command_buffer);
if (result == VK_SUCCESS)
...
```

bradgrantham-lunarg:
👍

panos-lunarg:
I think vulkan state writer iterates over all available buffers and images before it starts generating the initialisation commands. Perhaps we can write a meta command with the total size the stage buffer (ideally) needs to be. The replay device can use that information and depending on the available memory create the "ideal" stage buffer

fabian-lunarg:
> I think vulkan state writer iterates over all available buffers and images before it starts generating the initialisation commands. Perhaps we can write a meta command with the total size the stage buffer (ideally) needs to be. The replay device can use that information and depending on the available memory create the "ideal" stage buffer

this is pretty much what I implemented in this PR. instead of adding a new meta-command I realized we didn't
make use of the two existing fields in `BeginResourceInitCommand`.
I repurposed one of them and populate with the summed size of all staging-resources.
new captures will report that total size and replay can create the staging-buffer matching that size.
for old/existing captures, which still only report size of largest resource, a minimum staging-size was introduced as fallback, which was set low'ish as 32Mb.
additional checks have been added to double-check the availability of a suitable heap.

framework/encode/vulkan_state_writer.cpp
```diff
@@ -2826,16 +2828,16 @@ void VulkanStateWriter::WriteBufferMemoryState(const VulkanStateTable& state_tab
...
+                    if (*max_staging_copy_size < wrapper->created_size)
+                    {
+                        *max_staging_copy_size = wrapper->created_size;
+                    }
 
-                if (snapshot_info.need_staging_copy && ((*max_staging_copy_size) < wrapper->created_size))
-                {
-                    (*max_staging_copy_size) = wrapper->created_size;
+                    // sum staging-copy sizes
+                    *total_staging_copy_size += wrapper->created_size;
```
fabian-lunarg, line 2839: @panos-lunarg `VulkanStateWriter` now keeps track of a `total_staging_copy_size`

panos-lunarg:
> > I think vulkan state writer iterates over all available buffers and images before it starts generating the initialisation commands. Perhaps we can write a meta command with the total size the stage buffer (ideally) needs to be. The replay device can use that information and depending on the available memory create the "ideal" stage buffer
> 
> this is pretty much what I implemented in this PR. instead of adding a new meta-command I realized we didn't make use of the two existing fields in `BeginResourceInitCommand`. I repurposed one of them and populate with the summed size of all staging-resources. new captures will report that total size and replay can create the staging-buffer matching that size. for old/existing captures, which still only report size of largest resource, a minimum staging-size was introduced as fallback, which was set low'ish as 32Mb. additional checks have been added to double-check the availability of a suitable heap.

This is great. I didn't check in detail this PR before commenting. I think this is the best thing we can do

PR #2471, opened 2025-10-21
Check pData of ShaderGroupHandles identical

locke-lunarg:
If ShaderGroupHandles' pData of the capture and the replay are different, the replay might fail.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -9591,20 +9591,42 @@ VulkanReplayConsumerBase::OverrideGetRayTracingShaderGroupHandlesKHR(PFN_vkGetRa
     assert((device_info != nullptr) && (pipeline_info != nullptr) && (pData != nullptr) &&
            (pData->GetOutputPointer() != nullptr));
 
+    auto physical_device_info = GetObjectInfoTable().GetVkPhysicalDeviceInfo(device_info->parent_id);
+
+    // in practice: always 32 bytes
+    uint32_t capture_handle_size = physical_device_info->capture_raytracing_properties->shaderGroupHandleSize;
+    uint32_t replay_handle_size =
+        physical_device_info->replay_device_info->raytracing_properties->shaderGroupHandleSize;
+
+    uint32_t replay_data_size = dataSize;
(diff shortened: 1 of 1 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -9610,20 +9610,39 @@ VulkanReplayConsumerBase::OverrideGetRayTracingShaderGroupHandlesKHR(PFN_vkGetRa
...
+        if (capture_handle_size != replay_handle_size)
+        {
+            GFXRECON_LOG_WARNING(
+                "shaderGroupHandleSize of the capture and the replay are different. The replay might fail.");
+        }
+        else if (memcmp(captured_data, output_data, replay_data_size) != 0)
+        {
+            GFXRECON_LOG_WARNING(
+                "ShaderGroupHandles' pData of the capture and the replay are different. The replay might fail.");
+        }
```
davidd-lunarg: I think this check should be skipped if replay used `-m rebind`. In that case the `VulkanAddressReplacer` should perform a mapping from capture to replay handle values. Is that right @fabian-lunarg ?

fabian-lunarg: yes, that's right. the handles are expected to match only on same device+driver. when using `-m rebind` we need to map them. there is an utility to guard that condition:  `bool UseAddressReplacement(...)`

locke-lunarg: `rebind` and `VulkanAddressReplacer` seems to not do anything for it. Is this part not finished yet?

fabian-lunarg: it is certainly there :) - SBT / handle-mapping is done as part of [VulkanAddressReplacer::ProcessCmdTraceRays](https://github.com/LunarG/gfxreconstruct/blob/2d1b0893efcbf785e39567116c32b548dfbfd626/framework/decode/vulkan_address_replacer.h#L165)
the handle-map created below is stored and used as input-parameter for the gpu sanitizer-dispatch.

locke-lunarg: Thanks, @fabian-lunarg . Added checking UseAddressReplacement().

fabian-lunarg (approved):

PR #2474, opened 2025-10-21
Frame marker versioning, detection, and patch

jzulauf-lunarg:
Because the presence of kFrameMarkerBlock/kEndMarker blocks cannot be determined prior before the first frame delimiting function or method block is encountered (e.g., kFunctionCallBlock/ApiCall_vkQueuePresentKHR ), the ProcessBlocks can will return twice with the same frame number -- causing a variety of issues.  This PR addresses this in the following ways.  This avoid all double return from ProcessBlocks on frame 1.

 * Add gfxr-version detection for frame marker support
    * Parses gfxrecon-version from the operations annotation.  Any version after v1.0.1 has frame markers.
    * Patching support for files created between Jul-10-23 and Oct-3-23.
* Add warning for frame markers in v0.0 files.
* Add option to info command to patch for frame markers
    * Detects for 0,0 file version with frame makers present and updates format to 0,1.  
    * Skips all other files.
* Add support for explicit frame marker detection
    * Update file format version from 0, 0 to 0, 1
    * Add file format version detection for frame markers.
    * Add file format and frame marker information to gfxrecon-info


jzulauf-lunarg:
@marius-pelegrin-arm -- this should address all double frame one returns, though files created between Frame Marker being added and v1.0.1 being set need an 1 byte patch (with a WARNING for affected files)

tools/info/main.cpp
```diff
@@ -317,14 +321,34 @@ void PrintAnnotations(uint32_t                          annotation_count,
     }
 }
 
-void PrintFileFormatInfo(const gfxrecon::decode::FileProcessor& file_processor)
+struct FileFormatInfo
```
MarkY-LunarG, line 320: NOTE: I like you separating this info out into it's own structure.

MarkY-LunarG (approved):

jzulauf-lunarg:
@bradgrantham-lunarg -- updated with a separate patch utility with all the features of the support inside info (which I've stripped back out).  The non-DRYness is minimal, don't see the need to refactor it out.

bradgrantham-lunarg (approved):

marius-pelegrin-arm:
I just come back from vacation and am starting to read the commit. It's heavy to unpack so I will take a bit of time to review this. However at first sight I already have a few questions:
-  If I understand correctly, a new "version" is added to the trace file in an annotation. Why not use the trace version (`GFXRECON_TRACE_VERSION_{MAJOR|MINOR}`) ?
- It also appears you create a new tool to update the version of a trace, why not use gfxrecon-optimize for this purpose ?

jzulauf-lunarg:
@marius-pelegrin-arm.  I'm not bumping the gfxrecon-version found in the trace, I'm bumping the file format version found in the header.  

I hadn't thought of putting the patch operation in optimize, as it's not really a whole file operation, but a targeted one byte patch. Originally, I had it as an option to 'info', but it really doesn't belong there. It can actually be done with a `dd` command, but that approach seemed ... unsafe, so I added a targeted patch tool.  It's smart, safe, and literally can be run against *all* files, only updating ones that need it.

PR #2475, opened 2025-10-22
Fixes for custom gfxr pipeline cache

panos-lunarg:
The replay options --save-pipeline-cache and --load-pipeline-cache do not work as on could expect if they are pointing at the same file as the --save-pipeline-cache deletes the file's content when the replay starts.

This PR changes the way the pipeline caches are internally handled and instead of dumping them immediately to the cache file, they are kept in memory until the end of execution. This way the cache file is not emptied at the begining of the execution and can be used for multiple runs, with each one effectively updating the file.


panos-lunarg:
Abandoning in favor of https://github.com/LunarG/gfxreconstruct/pull/2486

PR #2476, opened 2025-10-23
Add support for VK_DESCRIPTOR_TYPE_BLOCK_MATCH_IMAGE_QCOM

anholt:
This is part of VK_QCOM_image_processing, and they are just like a SAMPLED_IMAGE, but for only the new sampling opcodes.


anholt:
This was enough to capture and compare dump output for a couple of testcases of the extension.

PR #2484, opened 2025-10-28
Refactor parsing from file processing -- in situ

jzulauf-lunarg:
Note that the work is being done in place as much as possible to clarify review.  

The ParsedBlock, BlockBuffer, and BlockParser will be moved to their own home(s) in a future PR.

### DIFF
--- a/framework/decode/api_payload.h
+++ b/framework/decode/api_payload.h
@@ -1,0 +1,879 @@
+/*
+** Copyright (c) 2025 LunarG, Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and associated documentation files (the "Software"),
+** to deal in the Software without restriction, including without limitation
+** the rights to use, copy, modify, merge, publish, distribute, sublicense,
+** and/or sell copies of the Software, and to permit persons to whom the
+** Software is furnished to do so, subject to the following conditions:
+**
+** The above copyright notice and this permission notice shall be included in
+** all copies or substantial portions of the Software.
+**
+** THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+** IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+** FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+** AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+** FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+** DEALINGS IN THE SOFTWARE.
+*/
+
+#ifndef GFXRECON_API_PAYLOAD_H
+#define GFXRECON_API_PAYLOAD_H
+
+#include "api_decoder.h" // for ApiDecoder method pointers
+
+#include <array>
+#include <memory>
+#include <string>
+#include <tuple>
+#include <variant>
+#include <vector>
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(decode)
+
+// Define threshold for "by value" vs "heap" storage in the variant
+struct CommandStoragePolicy
+{
+    // This size is picked to match FunctionCallArgs and MethodCallArgs which constitute more than 95% of total blocks
+    static constexpr size_t kInlineThreshold = 48;
+    template <typename T>
+    static constexpr bool IsLarge()
+    {
+        return sizeof(T) > kInlineThreshold;
+    }
+};
+
+template <typename Command, typename Dummy = void>
+struct DispatchHasCallId : std::false_type
+{};
+
+template <typename Command>
+struct DispatchHasCallId<Command, std::void_t<decltype(std::declval<Command>().call_id)>> : std::true_type
+{};
+
+template <typename Command, typename Dummy = void>
+struct DispatchHasMetaDataId : std::false_type
+{};
+
+template <typename Command>
+struct DispatchHasMetaDataId<Command, std::void_t<decltype(std::declval<Command>().meta_data_id)>> : std::true_type
+{};
+
+template <typename Command>
+struct DispatchHasAllocGuard : std::false_type
+{};
+
+// ---- Flags base (one place for size and per-command booleans) ----------------
+//
+// The flag traits class is a base class for each of the DispatchTraits specializations
+// defined below.  The flags control the storage and dispatch characteristics of the
+// Command type.
+template <typename Command>
+struct DispatchFlagTraits
+{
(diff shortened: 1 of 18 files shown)


jzulauf-lunarg:
Passes "extended test suite"

framework/decode/file_processor.cpp
```diff
@@ -430,127 +433,47 @@ bool FileProcessor::ProcessBlocks()
...
-
-                    success = block_buffer.Read(marker_type);
+                    block_parser.SetBlockIndex(block_index_);
+                    block_parser.SetFrameNumber(current_frame_number_);
+                    // NOTE: upon successful parsing, the block_buffer block data has been moved to the
+                    // parsed_block, though the block header is still valid.
+                    ParsedBlock parsed_block = block_parser.ParseBlock(block_buffer);
 
-                    if (success)
+                    // We don't support delayed decompression in these visitors
```
MarkY-LunarG: Is this meant to be a TODO like you'll eventually add it or is this simply a note you are not doing any delayed decompression here?

framework/decode/file_processor.cpp
```diff
@@ -820,32 +773,168 @@ void FileProcessor::HandleBlockReadError(Error error_code, const char* error_mes
...
 {
+    // We can infer the presence of frame markers from the operations version
+    if (annotation.type == gfxrecon::format::AnnotationType::kJson &&
+        annotation.label.compare(gfxrecon::format::kAnnotationLabelOperation) == 0)
+    {
+        // This is an operations annotation containing the version of the capture tool.
+        format::GfxrVersion version = format::ParseVersionFromOperations(annotation.data.c_str());
+        if (version.SupportsFrameMarkers())
+        {
+            assert(current_frame_number_ == kFirstFrame);
```
MarkY-LunarG: GFXRECON_ASSERT

framework/decode/file_processor.h
```diff
@@ -293,6 +512,118 @@ class FileProcessor
...
+            // The block and marker type are implied by the Args type
+            is_frame_delimiter = file_processor_.ProcessFrameDelimiter(end_frame);
+            success            = true;
+        }
+
+        // I/O Control
+        void Visit(const ParsedBlock& parsed_block, const ExecuteBlocksFromFileArgs& execute_blocks)
+        {
+            // The block and marker type are implied by the Args type
+            is_frame_delimiter = false;
```
MarkY-LunarG, line 584: Do you have to redefine all these to false since they're initialized that way?

framework/decode/file_processor.h
```diff
@@ -122,27 +184,187 @@ class BlockBuffer
...
+    BlockBuffer::BlockSpan ReadParameterBuffer(BlockBuffer& block_buffer, size_t buffer_size);
+
+    // TODO: Replace with actual ErrorHandler s.t. FileTransformer can use the BlockParser
+    using ErrorHandler = FileProcessor;
+    BlockParser(ErrorHandler& err, BufferPool& pool, util::Compressor* compressor) :
+        pool_(pool), err_handler_(err), compressor_(compressor)
+    {}
+
+  private:
+    BufferPool        pool_; // TODO: Get a better pool, and share with FileInputStream
```
MarkY-LunarG, line 356: Does this need to be an issue?  Or are you planning on pulling it in with a future commit in this work?

framework/decode/file_processor.h
```diff
@@ -42,18 +45,72 @@
...
+    return true;
+}
+
+template <typename Args>
+static void SetDecoderApiCallId(ApiDecoder& decoder, const Args& args)
+{
+    if constexpr (DispatchTraits<Args>::kHasCallId)
+    {
+        decoder.SetCurrentApiCallId(args.call_id);
+    }
```
MarkY-LunarG, line 91: Should we warn if not ?  The reason I ask is you use this inside a loop that has already checked that we should support it so if we don't something bad happened.

framework/decode/file_processor.h
jzulauf-lunarg: In a later PR as part of the Parser refactor work, will will continue with the deferred decompression and file transformer and housekeeping tasks.

framework/decode/file_processor.h
jzulauf-lunarg: The constexpr is part of the template/traits resolution, that is compile time evaluated, s.t. only Args types that have the call id field set the call id, and compile to a no-op otherwise.

framework/decode/file_processor.h
jzulauf-lunarg: I was going to remove both the defaults, but then realized, there's nothing that prevents a Visit caller from reusing a Visitor object, so instead added a note to always set this state for exactly that reason.

framework/decode/preload_file_processor.cpp
```diff
@@ -91,11 +92,11 @@ bool PreloadFileProcessor::PreloadBlocksOneFrame()
 }
 
 // Grab the block data off the front of the
-bool PreloadFileProcessor::GetBlockBuffer(BlockBuffer& block_buffer)
+bool PreloadFileProcessor::GetBlockBuffer(BlockParser& block_parser, BlockBuffer& block_buffer)
 {
     // Quick escape
     if (preload_block_data_.empty())
-        return Base::GetBlockBuffer(block_buffer);
+        return Base::GetBlockBuffer(block_parser, block_buffer);
```
MarkY-LunarG, line 100: Nit: Please use braces around this.

framework/decode/file_processor.cpp
```diff
@@ -2729,5 +2673,10 @@ bool FileProcessor::HandleBlockEof(const char* operation, bool report_frame_and_
     return success;
 }
 
+void ParsedBlock::Decompress(BlockParser& parser)
+{
+    assert("Not supported" == nullptr);
```
MarkY-LunarG: Another assert.

framework/decode/file_processor.cpp
```diff
@@ -2340,27 +2326,15 @@ bool FileProcessor::ProcessMetaData(BlockBuffer& block_buffer, format::MetaDataI
...
             {
-                std::string filename = util::filepath::Join(absolute_path_, filename_c_str);
-
-                // Check for self references
-                if (!filename.compare(file_stack_.back().active_file->GetFilename()))
-                {
-                    GFXRECON_LOG_WARNING(
-                        "ExecuteBlocksFromFile is referencing itself. Probably this is not intentional.");
-                }
-
```
MarkY-LunarG, line 2351: Is this behavior not needed anymore?

jzulauf-lunarg: moved to `FileProcessor::ProcessExecuteBlocksFromFile` called from the ProcessVisitor

framework/decode/api_payload.h
```diff
@@ -0,0 +1,879 @@
...
+    auto GetTuple() const { return std::tie(thread_id, device_id, swapchain_id, last_presented_image, image_state); }
+};
+struct BeginResourceInitArgs
+{
+    format::MetaDataId meta_data_id; // Needed by DispatchVisitor, but not ApiDecoder
+
+    format::ThreadId thread_id;
+    format::HandleId device_id;
+    uint64_t         max_resource_size;
+    uint64_t         max_copy_size;
```
fabian-lunarg, line 349: collided w. recent refactoring (https://github.com/LunarG/gfxreconstruct/pull/2466/):
`max_copy_size` has been renamed (and repurposed) to `total_copy_size`. 
so we should correct the field-name here too

PR #2491, opened 2025-10-31
BlockParser deferred decompression support

jzulauf-lunarg:
Implement deferred decompression support. DRY parameter buffer decompression logic.  Tested with kAlways and kQueueOptimized modes in ProcessBlocks

### DIFF
--- a/framework/decode/api_payload.h
+++ b/framework/decode/api_payload.h
@@ -63,6 +63,30 @@ template <typename Command>
 struct DispatchHasMetaDataId<Command, std::void_t<decltype(std::declval<Command>().meta_data_id)>> : std::true_type
 {};
 
+template <typename Command, typename Dummy = void>
+struct DispatchHasData : std::false_type
+{};
+
+template <typename Command>
+struct DispatchHasData<Command, std::void_t<decltype(std::declval<Command>().data)>> : std::true_type
+{};
+
+template <typename Command, typename Dummy = void>
+struct DispatchHasDataSize : std::false_type
+{};
+
+template <typename Command>
+struct DispatchHasDataSize<Command, std::void_t<decltype(std::declval<Command>().data_size)>> : std::true_type
+{};
+
+template <typename Command, typename Dummy = void>
+struct DispatchHasCommandHeader : std::false_type
+{};
+
+template <typename Command>
+struct DispatchHasCommandHeader<Command, std::void_t<decltype(std::declval<Command>().command_header)>> : std::true_type
+{};
+
 template <typename Command>
 struct DispatchHasAllocGuard : std::false_type
 {};
@@ -75,10 +99,13 @@ struct DispatchHasAllocGuard : std::false_type
 template <typename Command>
 struct DispatchFlagTraits
 {
-    static constexpr bool kIsLarge       = CommandStoragePolicy::IsLarge<Command>();
-    static constexpr bool kHasCallId     = DispatchHasCallId<Command>::value;
-    static constexpr bool kHasMetaDataId = DispatchHasMetaDataId<Command>::value;
-    static constexpr bool kHasAllocGuard = DispatchHasAllocGuard<Command>::value;
+    static constexpr bool kIsLarge          = CommandStoragePolicy::IsLarge<Command>();
+    static constexpr bool kHasCallId        = DispatchHasCallId<Command>::value;
+    static constexpr bool kHasMetaDataId    = DispatchHasMetaDataId<Command>::value;
+    static constexpr bool kHasAllocGuard    = DispatchHasAllocGuard<Command>::value;
+    static constexpr bool kHasData          = DispatchHasData<Command>::value;
+    static constexpr bool kHasDataSize      = DispatchHasDataSize<Command>::value;
+    static constexpr bool kHasCommandHeader = DispatchHasCommandHeader<Command>::value;
 };
 
 // --- Payload structs (argument order preserved) ---
@@ -90,10 +117,10 @@ struct FunctionCallArgs
 {
     format::ApiCallId call_id;
     ApiCallInfo       call_info;
-    const uint8_t*    buffer;
-    size_t            buffer_size;
+    const uint8_t*    data;
+    size_t            data_size;
 
-    auto GetTuple() const { return std::tie(call_id, call_info, buffer, buffer_size); }
+    auto GetTuple() const { return std::tie(call_id, call_info, data, data_size); }
 };
 template <>
 struct DispatchHasAllocGuard<FunctionCallArgs> : std::true_type
@@ -104,10 +131,10 @@ struct MethodCallArgs
     format::ApiCallId call_id;
     format::HandleId  object_id;
     ApiCallInfo       call_info;
-    const uint8_t*    parameter_buffer;
-    size_t            buffer_size;
+    const uint8_t*    data;
+    size_t            data_size;
 
-    auto GetTuple() const { return std::tie(call_id, object_id, call_info, parameter_buffer, buffer_size); }
+    auto GetTuple() const { return std::tie(call_id, object_id, call_info, data, data_size); }
 };
 template <>
 struct DispatchHasAllocGuard<MethodCallArgs> : std::true_type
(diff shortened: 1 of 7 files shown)


jzulauf-lunarg:
@bradgrantham-lunarg @MarkY-LunarG @ncesario-lunarg @panos-lunarg ... review?

framework/decode/file_processor.cpp
```diff
@@ -876,25 +845,107 @@ ParsedBlock BlockParser::ParseBlock(BlockBuffer& block_buffer)
...
+    {
+        // For uncompressed blocks the data size is the parameter buffer size
+        result.uncompressed_size = block_buffer.Remainder();
+    }
+
+    if (result.success)
+    {
+        const size_t buffer_size = block_buffer.Remainder();
+        result.buffer            = block_buffer.ReadSpan(buffer_size);
+        result.success           = result.buffer.size() == buffer_size; // To correctly handle correctly empty buffers
```
ncesario-lunarg: // To correctly handle ~~correctly~~ empty buffers

Should this be done before the above `if (result.success)`, or does `result.buffer` need to be filled out even if `result.success == false`?

framework/decode/file_processor.cpp
```diff
@@ -990,75 +991,30 @@ ParsedBlock BlockParser::ParseMethodCall(BlockBuffer& block_buffer)
...
-        }
-
-        if (success)
-        {
-            return ParsedBlock(
-                std::move(block_buffer.ReleaseData()),
-                MethodCallArgs{
-                    call_id, object_id, call_info, parameter_data.GetDataAs<const uint8_t>(), parameter_buffer_size },
-                std::move(uncompressed_store));
+            return ParsedBlock(std::move(block_buffer.ReleaseData()),
```
ncesario-lunarg: Is `std::move` necessary here? `ReleaseData` is already returning an r-value and the same constructor is called above without `std::move`. I'm fine either way as long as we're consistent. Unless I'm missing something as to why this one is different?

framework/decode/file_processor.cpp
```diff
@@ -1093,28 +1049,10 @@ ParsedBlock BlockParser::ParseMetaData(BlockBuffer& block_buffer)
         success = success && block_buffer.Read(header.memory_offset);
         success = success && block_buffer.Read(header.memory_size);
 
-        if (success)
+        const char*         label       = "fill memory meta-data block";
+        ParameterReadResult read_result = ReadParameterBuffer(label, block_buffer, header.memory_size);
```
ncesario-lunarg: Do we want to do this if `!success`? If we do, then making the below condition `success && read_result.success` might be slightly more readable.

ncesario-lunarg (commented):
This LGTM in general. A couple questions, but nothing that I would hold up merging for from my perspective.

framework/decode/file_processor.cpp
jzulauf-lunarg: We only try to read the buffer (span), if no previous errors have occurred.   

The double correctly is intentional.  We can't just check size() != 0 to know if the buffer is correct, we have to check the size of the span we read vs. the span we wanted (which might be the empty span)... in order to test correctly for buffers that a correct when empty.

Do you have a suggestion for a better comment?

@ncesario-lunarg

framework/decode/file_processor.cpp
jzulauf-lunarg: Correct. Fixed

framework/decode/file_processor.cpp
jzulauf-lunarg: This one got turned inside out accidentally (success check inside result.success check)  fixed.

framework/decode/file_processor.cpp
ncesario-lunarg: Ah, I see now. I do think the "double correctly" is a little confusing and the new comment is much better (at least for me it is :)).

ncesario-lunarg (approved):
LGTM pending others' reviews.

framework/decode/api_payload.h
```diff
@@ -505,19 +535,19 @@ struct InitializeMetaArgs
...
-    auto GetTuple() const { return std::tie(header, initialization_parameters_data); }
+    auto GetTuple() const { return std::tie(command_header, data); }
 };
 struct AnnotationArgs
 {
     uint64_t               block_index;
     format::AnnotationType type;
     std::string            label;
-    std::string            data;
+    std::string            annotation_data;
```
MarkY-LunarG, line 548: Is this because you don't want it associated with the standard data?  If so, a comment would be nice.

framework/decode/file_processor.cpp
```diff
@@ -578,77 +581,43 @@ bool FileProcessor::PeekBlockHeader(format::BlockHeader* block_header)
...
     err_handler_.HandleBlockReadError(error_code, error_message);
 }
 
-BlockBuffer::BlockSpan BlockParser::ReadParameterBuffer(BlockBuffer& block_buffer, size_t buffer_size)
+bool BlockParser::DecompressWhenParsed(const ParsedBlock& parsed_block)
 {
-    return block_buffer.ReadSpan(buffer_size);
+    switch (decompression_policy_)
+    {
+        case kAlways:
```
MarkY-LunarG, line 621: Nit, you could just make this the default and it will do what you already do with the default and final return.

framework/decode/file_processor.cpp
```diff
@@ -876,25 +845,111 @@ ParsedBlock BlockParser::ParseBlock(BlockBuffer& block_buffer)
...
+
+    return parsed_block;
+}
+
+// The parameter buffer always takes up the end of the block
+// Default value for uncompressed_size is kReadSizeFromBuffer -- which is the behavior of Function and method call
+BlockParser::ParameterReadResult
+BlockParser::ReadParameterBuffer(const char* label, BlockBuffer& block_buffer, uint64_t uncompressed_size)
+{
+    ParameterReadResult result; // result.success initializes to true;
```
MarkY-LunarG, line 919: Thank you for this comment.  I would have complained about the else below without it.

framework/decode/file_processor.h
```diff
@@ -319,6 +367,18 @@ using BufferPool = util::HeapBufferPool::PoolPtr;
...
+    {
+        kAlways = 0,     // Always decompress parameter data when parsing blocks
+        kNever,          // Never decompress parameter data when parsing blocks
+        kQueueOptimized, // Decompress only "small" blocks deferring large block to JIT decompression, suitable for
+                         // block preload/preparsed content
+    };
+
+    // The threshold is based on a trade off, maximize the number blocks with no decompression at replay
+    // but minimizing the memory overhead of the non-deferred compressions.
+    static constexpr size_t kSmallThreshold = 96 + sizeof(format::BlockHeader); // 83% of blocks, 26% of bytes
```
MarkY-LunarG, line 381: I would give more detail, i.e. "I ran a histogram analysis and determined that 96 is the ...."

framework/decode/file_processor.h
```diff
@@ -273,22 +275,68 @@ class ParsedBlock
...
+    {
+        GFXRECON_ASSERT(state_ == kDeferredDecompress);
+        state_              = kReady;
+        uncompressed_store_ = std::move(from_store);
+    }
+
   private:
     template <typename ArgPayload>
-    DispatchArgs MakeDispatchArgs(ArgPayload&& payload)
+    static DispatchArgs MakeDispatchArgs(ArgPayload&& payload)
```
MarkY-LunarG, line 340: Why static and not just make method "const"?

framework/decode/file_processor.cpp
```diff
@@ -876,25 +845,111 @@ ParsedBlock BlockParser::ParseBlock(BlockBuffer& block_buffer)
...
         case format::kStateMarkerBlock:
-            return ParseStateMarker(block_buffer);
+            parsed_block = ParseStateMarker(block_buffer);
+            break;
         case format::kAnnotation:
-            return ParseAnnotation(block_buffer);
+            parsed_block = ParseAnnotation(block_buffer);
+            break;
         case format::kUnknownBlock:
+            break;
```
MarkY-LunarG, line 898: Should we have a log message here and in the default path?

framework/decode/file_processor.h
jzulauf-lunarg: Two reasons: 1) Principle of least access.  It needs none of ParsedBlocks fields, 2) it's being called while the constructor is running (i.e. before ParsedBlocks would be "valid"

framework/decode/file_processor.cpp
jzulauf-lunarg: Logged by the caller.

jzulauf-lunarg:
@MarkY-LunarG  -- I'm going to address the nits/comments in https://github.com/LunarG/gfxreconstruct/pull/2508  -- so that I don't have to migrate them across the already rehomed source.

PR #2495, opened 2025-11-05
Beau single commit

beau-lunarg:
Add initial refs for test dependency and test suite dependency


beau-lunarg:
There was a Jenkins hiccup, which is why LunarG CI Checkrun is cancelled.  Passing run is here:
http://jenkins:8080/job/gfxreconstruct/8121/

PR #2497, opened 2025-11-06
Add `--streamline-annotate` option

marius-pelegrin-arm:
Add option to annotate frames using Gator's `ANNOTATE` macros for Streamline.

For more info about Streamline: https://developer.arm.com/Tools%20and%20Software/Streamline%20Performance%20Analyzer

### DIFF
--- a/USAGE_android.md
+++ b/USAGE_android.md
@@ -1018,6 +1018,7 @@ options:
   --dump-resources-dump-separate-alpha
                         Dump image alpha in a separate image file.
   --pbi-all             Print all block information.
+  --streamline-annotate Add gator annotation tags to mark replay frames.
   --pbis RANGES         Print block information between block index1 and block
                         index2
   --pcj, --pipeline-creation-jobs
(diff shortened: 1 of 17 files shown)


marius-pelegrin-arm:
The test seems to fail on code style check. Which is weird because content in `external` shouldn't be checked. Is there a way to disable it ?

MarkY-LunarG:
Thanks @marius-pelegrin-arm , we're looking into it now.

fabian-lunarg:
we came to the conclusion we want to avoid the tight coupling demonstrated in this PR,
but we acknowledge there might be a need to provide such functionality in the future.

this sort of extension 'could' maybe be achieved by utilizing VK-layer mechanism or 'some' other plugin-mechanism (which we do not have atm). I understand this sounds rather vague, but we have to close for now and get back to it.

per-mathisen-arm:
Understood. We'll pass this feedback back to the Streamline people.

PR #2500, opened 2025-11-06
Check UseAddressReplacement for trimming

locke-lunarg:
Some ray tracing game's trimming replay with no rebind hang on 
```
  VulkanAddressReplacer::ProcessBuildVulkanAccelerationStructuresMetaCommand
       ~QueueSubmitHelper()
               device_table->WaitForFences
``` 

I think that for no rebind case, it shouldn't use `VulkanAddressReplacer`. I added `UseAddressReplacement()` in `ProcessCopyVulkanAccelerationStructuresMetaCommand` and `ProcessBuildVulkanAccelerationStructuresMetaCommand` to skip `VulkanAddressReplacer`, and then the trimming replay is successful.

@fabian-lunarg Do you think that it's good? The replay has many `VulkanAddressReplacer` and `VulkanDeviceAddressTracker` without checking `UseAddressReplacement()`. Should we check `UseAddressReplacement()` for all of them?


crocdialer:
> @fabian-lunarg Do you think that it's good? The replay has many VulkanAddressReplacer and VulkanDeviceAddressTracker without checking UseAddressReplacement(). Should we check UseAddressReplacement() for all of them?

basically that was intentional. when we skip `ProcessBuildVulkanAccelerationStructuresMetaCommand` we circumvent rebuilding of acceleration-structures for the trim-state entirely and fallback to using the existing buffer-content (which might contain an invalid AS due to driver updates etc.). this works on same device until the driver applies changes.

meaning not all cases in VulkanAddressReplacer are rebind-related (most are).
there are also relevant cases for AS-trimming in general without using `-m rebind`.

I think we should figure out why/what goes south with the submit/fence-sync in

 ```
 ~QueueSubmitHelper()
               device_table->WaitForFences
```

fabian-lunarg:
> @fabian-lunarg Do you think that it's good? The replay has many VulkanAddressReplacer and VulkanDeviceAddressTracker without checking UseAddressReplacement(). Should we check UseAddressReplacement() for all of them?

basically that was intentional. when we skip ProcessBuildVulkanAccelerationStructuresMetaCommand we circumvent rebuilding of acceleration-structures for the trim-state entirely and fallback to using the existing buffer-content (which might contain an invalid AS due to driver updates etc.). this works on same device until the driver applies changes.

meaning not all cases in VulkanAddressReplacer are rebind-related (most are).
there are also relevant cases for AS-trimming in general without using -m rebind.

I think we should figure out why/what goes south with the submit/fence-sync in

~QueueSubmitHelper()
              device_table->WaitForFences

fabian-lunarg:
@locke-lunarg 
> Some ray tracing game's trimming replay with no rebind hang on

can you share/pm a capture so I can try reproducing?

fabian-lunarg:
> I think we should figure out why/what goes south with the submit/fence-sync in
~QueueSubmitHelper()
device_table->WaitForFences

I think it's best we go ahead with your suggested changes and I'll try to figure it out along the way

fabian-lunarg (approved):

MarkY-LunarG:
@fabian-lunarg, it sounds like we should make a TODO or a new issue if you're recommending these changes go in to cover  the remaining problem.  Or were you planning on waiting to close this until then?

fabian-lunarg:
> @fabian-lunarg, it sounds like we should make a TODO or a new issue if you're recommending these changes go in to cover the remaining problem. Or were you planning on waiting to close this until then?

sounds good, I'll formulate a follow-up issue. it's even debatable if we regard that as a problem or not. 
it's the question what we expect from a 'normal' replay and under which conditions we accept a capture to become incompatible.

MarkY-LunarG:
Ok. At the very least, maybe we should we at least put a INFO message on the "else" side of these new checks indicating something may go wrong in these scenarios?

fabian-lunarg:
> Ok. At the very least, maybe we should we at least put a INFO message on the "else" side of these new checks indicating something may go wrong in these scenarios?

the else-case basically corresponds to "not -m rebind". we don't expect anything to go wrong there for replay on same device+driver. there will be a populated buffer that was valid during replay, so should really work.
we would print that log_info basically for all trimmed RT replays, no exception, so I'm not sure it's necessary.

I guess it's better to explain the other way around. we had those checks in place and I removed them a couple months ago. reason was basically to make normal replay also rebuild AS (like required for `- m rebind`) to make it more robust against driver-changes.
that's maybe even something nobody expects and we make no promises there. also it comes at some cost for state-init.

now apparently this is causing issues, so the problem it caused currently really outweighs the benefit.
chances are it is the better approach and aligns with what is expected from GFXR.

PR #2502, opened 2025-11-07
script: Add skip folder list to check_code_style.py

MarkY-LunarG:
This strips out certain directories for running clang-format on.

In this case, 'external' sub-folder should not be handled with clang-format.

### DIFF
--- a/cmake/CodeStyle.cmake
+++ b/cmake/CodeStyle.cmake
@@ -48,7 +48,12 @@ endif()
 
 macro(generate_target_source_files TARGET TARGET_SOURCE_FILES)
     get_target_property(${TARGET_SOURCE_FILES} ${TARGET} SOURCES)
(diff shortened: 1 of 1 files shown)


scripts/check_code_style.py
```diff
@@ -28,6 +28,28 @@
...
 import argparse
 import os
+from pathlib import Path
+
+def sanitize_file_list(file_list: list[str]):
+    """
+    Remove files from the list of skip folders which are folders
+    we do not want to execute clang-format analysis of the files.
+    """
+    skip_folders = [
```
bradgrantham-lunarg: Make this module-level and pass it in from check_code_style

bradgrantham-lunarg (approved):

scripts/check_code_style.py
MarkY-LunarG: Using only CMake now.

PR #2503, opened 2025-11-10
Fix variable declaration in `OverrideQueueSubmit{2}`

marius-pelegrin-arm:
The `addresses_to_replace` vector is wrongly declared BEFORE looping on the submit infos in `VulkanReplayConsumerBase::OverrideQueueSubmit`. The vector is thus filled and used during each iteration of the loop but never cleared. This is wrong - and `VulkanReplayConsumerBase::OverrideQueueSubmit2` has the correct behavior.

In addition, the `submit_info_mut` variable is not used before getting inside the `if (sync_wait_semaphores)` context. The variable is declared inside the context for one override but out of the context for the other. This does not create a bug, but itches my brain (in addition to technically being additional useless calls in some situations).


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4030,10 +4030,10 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit(PFN_vkQueueSubmit
 
     if (UseAddressReplacement(device_info) && submit_info_data != nullptr)
     {
-        std::vector<VkDeviceAddress> addresses_to_replace;
-
```
fabian-lunarg, line 4034: thanks for catching this! prob. got screwed when transferring from the `OverrideQueueSubmit2` code-path

PR #2504, opened 2025-11-10
VulkanResourceInitializer follow-up

fabian-lunarg:
- fix staging_buffer_size being overwritten
- no fence recreation, just reset

follows up on https://github.com/LunarG/gfxreconstruct/pull/2466 and corrects an overlooked assignment.

### DIFF
--- a/cmake/CodeStyle.cmake
+++ b/cmake/CodeStyle.cmake
@@ -48,7 +48,12 @@ endif()
 
 macro(generate_target_source_files TARGET TARGET_SOURCE_FILES)
     get_target_property(${TARGET_SOURCE_FILES} ${TARGET} SOURCES)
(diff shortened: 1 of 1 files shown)


framework/decode/vulkan_resource_initializer.cpp
```diff
@@ -1063,8 +1074,6 @@ VkResult VulkanResourceInitializer::AcquireStagingBuffer(VkDeviceSize size)
 
             if (result == VK_SUCCESS)
             {
-                staging_buffer_size_ = size;
```
fabian-lunarg, line 1066: this was overlooked and made us use fewer staging-mem than was allocated, preventing a previous optimization to take effect

Issue #2488, opened 2025-10-30
git commit and checkout require lengthy Android rebuilds


footballhead:
My P1G6 running Windows 11 with an i7-13800H and 64 GB of RAM can take **46s** to rebuild the Android `layer` target for `arm64-v8a` on dev HEAD (4f1fa3aa9870b00404e6597283b2032a885303b3) just by creating a new commit or changing the branch. This hampers my velocity.

From what I can tell, this is caused by many files including `project_version_<CONFIG>.h` which contains the repo SHA. Simple fixes like hiding the SHA in a static library dramatically improves build times to **6s** in the same situations. This can be dropped further by removing the SHA altogether.

I expect each Android target as well as host targets to be impacted in the same way.

My test:

```sh
# Build the layer once to generate a baseline
gradlew layer:assembleDebug -Parm64-v8a

# Generate a new SHA without code changes
git commit --allow-empty -m "test"

# Rebuild layer with only the SHA change
gradlew layer:assembleDebug -Parm64-v8a
```

`ninja -d explain` revealed that it was recompiling a handful of files because `project_version_<CONFIG>.h` changed:

```
"C:\\Users\\hitchens\\AppData\\Local\\Android\\Sdk\\cmake\\3.22.1\\bin\\ninja.exe" -C "C:\\Users\\hitchens\\Desktop\\git\\gfxreconstruct\\android\\layer\\.cxx\\Debug\\1e225de1\\arm64-v8a" VkLayer_gfxreconstruct -d explain
ninja: Entering directory `C:\\Users\\hitchens\\Desktop\\git\\gfxreconstruct\\android\\layer\\.cxx\\Debug\\1e225de1\\arm64-v8a'
ninja explain: output build.ninja older than most recent input C:/Users/hitchens/Desktop/git/gfxreconstruct/.git/HEAD (7833068202307323 vs 7833070889240476)
[0/1] Re-running CMake...
-- OpenXR Header = C:/Users/hitchens/Desktop/git/gfxreconstruct/external/OpenXR-SDK/include/openxr/openxr.h
-- Detected OpenXR Version 1.1.40
-- OpenXR support enabled
-- Vulkan Header = C:/Users/hitchens/Desktop/git/gfxreconstruct/external/Vulkan-Headers/include/vulkan/vulkan_core.h
-- Detected Vulkan Version 1.4.328
-- Configuring done
-- Generating done
[...]

PR #2506, opened 2025-11-11
Improve developer build times by hiding SHA

footballhead:
Building, especially Android, takes longer than expected when the the
git SHA changes (new commit, branch change). On my P1G6 (i7-13800H,
64GB RAM) it takes 46s to build the Android layer project for arm64-v8a:

```sh
git commit --allow-empty -m test
gradlew layer:assembleDebug -Parm64-v8a
```

A handful of C++ files include a header file containing the repo SHA.
When the SHA changes, those files are recompiled. Since some files are
large, this can take a long time.

Instead of consuming the SHA directly, isolate it into a static library
and expose it via a function call. This pushes the problem from the
compiler to the linker, reducing the number of files to recompile when
the git SHA changes.

(The only exception is version.rc which wants a string literal.)

With this change, the same test case takes 6s.

Fixes https://github.com/LunarG/gfxreconstruct/issues/2488


tools/tool_settings.h
```diff
@@ -1452,7 +1452,7 @@ static bool CheckOptionPrintVersion(const char* exe_name, const gfxrecon::util::
         }
 
         GFXRECON_WRITE_CONSOLE("%s version info:", app_name.c_str());
-        GFXRECON_WRITE_CONSOLE("  GFXReconstruct Version %s", GFXRECON_PROJECT_VERSION_STRING);
+        GFXRECON_WRITE_CONSOLE("  GFXReconstruct Version %s", GetProjectVersionString());
```
footballhead, line 1325: missed updating test-launcher. I'm surprised since I followed the instructions in BUILD.md

bradgrantham-lunarg:
To reviewers: the failure in LunarG CI Checkrun #8174 is an ongoing issue with ray tracing shader group handles and does not appear to be related to this PR.  The LunarG CI Checkrun #8174 and subsequent internal failures in "raytracing callable" should be ignored.

PR #2507, opened 2025-11-12
Add proper offscreen frame boundary handling

marius-pelegrin-arm:
**Definition**: I call "offscreen frame boundary" all Vulkan calls or structures that define a new frame and its attached resources but that are NOT a call to `vkQueuePresentKHR`. This concept could be extended to non-Vulkan calls like DX12 calls or to future GFXReconstruct meta-commands, but this is NOT handled by this commit.

Currently handled offscreen frame boundaries are:
- Calls to `vkFrameBoundaryANDROID`
- Calls to `vkQueueSubmit{2}` when the `pNext` chain of `VkSubmitInfo{2}` contains a `VkFrameBoundaryEXT` structure
- Submission of command buffers containing a recorded call to `vkCmdInsertDebugUtilsLabelEXT` or `vkCmdDebugMarkerInsertEXT` in which the label or marker name contains `graphics::kVulkanVrFrameDelimiterString`
    
Offscreen frame boundaries may or may not contain an image.
    
The goal of this commit is to provide a uniform and user-friendly way to specify what GFXReconstruct does when encountering frame boundaries (offscreen or onscreen).
    
I propose, in this commit, to add a new behavior for "presenting" offscreen frame boundary images when they exist. The idea is to consider offscreen frame boundaries as belonging to some "additional swapchain and surface" and to allow to present them or not using the existing `--swapchain` option and a new `--use-ext-frame-boundary` to force the conversion of frame boundaries to `VkFrameBoundaryEXT`.
    
This is necessary because some platforms use offscreen frame boundaries as a valid way to actually present to the screen. Making them not that "offscreen" anymore. A good example of that is Android using `vkFrameBoundaryANDROID` calls in its UI applications to present `AHardwareBuffer`-based images to the screen. At capture time, the user sees a normal application, but replaying the same calls will lead to a black screen, even on the exact same device.
    
Here is the new behavior I implement:
- The `--swapchain` option remains, and no new flag is extended, but its definition is extended, as it now also defines the behavior of offscreen frame boundaries
- The `--use-ext-frame-boundary` option is added and allows to specify the difference between "letting default offscreen frame boundaries" and "using `VkFrameBoundaryEXT` for everything". This option automatically triggers `--swapchain offscreen`.
- The `--offscreen-swapchain-frame-boundary` is deprecated and is now and alias to `--use-ext-frame-boundary`
    
Here is the new behavior of `--swapchain`:
- `captured`: All the calls are forwarded "as-is".
- `offscreen`: No `VkSwapchainKHR` object is created at all, and by default offscreen frame boundaries are just "forwarded as-is". If `--use-ext-frame-boundary` is set, everything is converted to submissions of `VkFrameBoundaryEXT`.
- `virtual`: The behavior is the same for normal Vulkan swapchains. When encountering the first offscreen frame boundary, a new window, surface and swapchain are created, and this and future offscreen frame boundaries are converted to `vkQueuePresentKHR` calls on the new swapchain.

I think this commit would be ready as a first step of improvement in the handling of offscreen frame boundaries but there are still holes in the implementation that could be addressed later:
- If an application uses multiple types of boundaries at the same time
- If an application outputs multiple images per offscreen frame boundaries (possible with `VkFrameBoundaryEXT`)
- If complex scenarios appear like using multiple devices, queues etc..

### DIFF
--- a/framework/decode/vulkan_offscreen_swapchain.cpp
+++ b/framework/decode/vulkan_offscreen_swapchain.cpp
@@ -346,10 +346,16 @@ VkResult VulkanOffscreenSwapchain::QueueSubmit(PFN_vkQueueSubmit                
         std::vector<VkFrameBoundaryEXT>   inserted_frame_boundaries;
         std::vector<std::vector<VkImage>> inserted_frame_boundaries_images;
 
+        bool capture_has_frame_boundary = false;
+
         for (uint32_t i = 0; i < submit_count; ++i)
         {
+            const bool found_frame_boundary =
+                (graphics::vulkan_struct_get_pnext<VkFrameBoundaryEXT>(&submit_infos[i]) != nullptr);
+            capture_has_frame_boundary = (capture_has_frame_boundary || found_frame_boundary);
+
             // If the submit already contains a VkFrameBoundaryEXT, we can assume we don't need to convert it
-            if (graphics::vulkan_struct_get_pnext<VkFrameBoundaryEXT>(&submit_infos[i]) == nullptr)
+            if (!found_frame_boundary)
             {
                 VkFrameBoundaryEXT*   frame_boundary = nullptr;
                 std::vector<VkImage>* images         = nullptr;
@@ -404,7 +410,7 @@ VkResult VulkanOffscreenSwapchain::QueueSubmit(PFN_vkQueueSubmit                
         }
 
         // If the queue submission contains a frame boundary, increment the frame counter
-        if (!inserted_frame_boundaries.empty())
+        if (capture_has_frame_boundary || !inserted_frame_boundaries.empty())
         {
             ++frame_boundary_.frameID;
         }
@@ -447,11 +453,16 @@ VkResult VulkanOffscreenSwapchain::QueueSubmit2(PFN_vkQueueSubmit2              
         std::vector<VkFrameBoundaryEXT>   inserted_frame_boundaries;
         std::vector<std::vector<VkImage>> inserted_frame_boundaries_images;
 
+        bool capture_has_frame_boundary = false;
+
         for (uint32_t i = 0; i < submit_count; ++i)
         {
+            const bool found_frame_boundary =
+                (graphics::vulkan_struct_get_pnext<VkFrameBoundaryEXT>(&submit_infos[i]) != nullptr);
+            capture_has_frame_boundary = (capture_has_frame_boundary || found_frame_boundary);
+
             // If the submit already contains a VkFrameBoundaryEXT, we can assume we don't need to convert it
-            if (graphics::vulkan_struct_get_pnext<VkFrameBoundaryEXT>(&submit_infos[i]) == nullptr &&
-                meta_submit_infos[i].pCommandBufferInfos != nullptr)
+            if (!found_frame_boundary && meta_submit_infos[i].pCommandBufferInfos != nullptr)
             {
                 VkFrameBoundaryEXT*   frame_boundary = nullptr;
                 std::vector<VkImage>* images         = nullptr;
@@ -507,7 +518,7 @@ VkResult VulkanOffscreenSwapchain::QueueSubmit2(PFN_vkQueueSubmit2              
         }
 
         // If the queue submission contains a frame boundary, increment the frame counter
-        if (!inserted_frame_boundaries.empty())
+        if (capture_has_frame_boundary || !inserted_frame_boundaries.empty())
         {
             ++frame_boundary_.frameID;
         }


framework/decode/vulkan_offscreen_swapchain.cpp
```diff
@@ -291,6 +270,256 @@ VkResult VulkanOffscreenSwapchain::QueuePresentKHR(VkResult
...
+        submitInfo.pCommandBuffers      = nullptr;
+        submitInfo.signalSemaphoreCount = 0;
+        submitInfo.pSignalSemaphores    = nullptr;
+
+        VkQueue queue;
+
+        decode::BeginInjectedCommands();
+        device_table->GetDeviceQueue(device_info->handle, 0, 0, &queue);
+        device_table->QueueSubmit(queue, 1, &submitInfo, VK_NULL_HANDLE);
+        decode::EndInjectedCommands();
```
MarkY-LunarG, line 311: I think we should be tracking what Queue is submitted to last in the same thread and then using that same queue to insert end frame in the "force" scenario.  Otherwise, we could end up using a Queue (in this case Queue index 0) which is scheduled differently on the hardware than we would like resulting in the "end of frame" marker occurring before/after it should during the actual replay.

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -1086,6 +1090,233 @@ void VulkanVirtualSwapchain::CmdPipelineBarrier2(PFN_vkCmdPipelineBarrier2 func,
...
+                                  application,
+                                  image_info,
+                                  semaphore_info == nullptr ? VK_NULL_HANDLE : semaphore_info->handle);
+}
+
+VkResult VulkanVirtualSwapchain::QueueSubmit(PFN_vkQueueSubmit                    func,
+                                             const VulkanQueueInfo*               queue_info,
+                                             uint32_t                             submit_count,
+                                             const VkSubmitInfo*                  submit_infos,
+                                             const Decoded_VkSubmitInfo*          meta_submit_infos,
```
fabian-lunarg, line 1115: maybe we can go without `const Decoded_VkSubmitInfo*          meta_submit_infos`
and just use `const VkSubmitInfo*                  submit_infos` to extract required information.
looking into that, it's already a lot of params passed here

framework/decode/vulkan_virtual_swapchain.h
```diff
@@ -170,6 +205,48 @@ class VulkanVirtualSwapchain : public VulkanSwapchain
 
     // Create an unordered map to associate the swapchain resource data with a particular Vulkan swapchain
     std::unordered_map<VkSwapchainKHR, std::unique_ptr<SwapchainResourceData>> swapchain_resources_;
+
+    void CreateSubmitSemaphores(const VulkanDeviceInfo* device_info, const graphics::VulkanDeviceTable* device_table);
+
+    format::HandleId FindExtFrameBoundaryImage(const void* decoded_struct);
```
fabian-lunarg: this looks brittle with `void*`

fabian-lunarg: refactored `FindExtFrameBoundaryImage` with explicit types (+ internal template-helper + static_assert)

framework/decode/vulkan_virtual_swapchain.cpp
fabian-lunarg: yeah ok, seeing the reason. a lot of handle_id -> info_struct queries are solved via `Decoded_Foo` meta-structs.
so seems there is no 'nicer' way to do it.

framework/decode/replay_options.h
```diff
@@ -58,8 +58,8 @@ struct ReplayOptions
     bool                         flush_measurement_frame_range{ false };
     bool                         flush_inside_measurement_range{ false };
     bool                         force_windowed{ false };
-    uint32_t                     windowed_width{ 0 };
-    uint32_t                     windowed_height{ 0 };
+    uint32_t                     windowed_width{ 320 };
+    uint32_t                     windowed_height{ 240 };
```
fabian-lunarg, line 62: what does that mean? where/how did we populate `windowed_width` / `windowed_height`
before?

don't think we should change here. 
this should rather define an (invalid) default-state for the struct.
and another place would be better suited to set default values windowed_width/height

android/scripts/gfxrecon.py
```diff
@@ -266,6 +267,9 @@ def MakeExtrasString(args):
     if args.offscreen_swapchain_frame_boundary:
         arg_list.append('--offscreen-swapchain-frame-boundary')
 
+    if args.use_ext_frame_boundary:
```
MarkY-LunarG, line 270: I wonder if "force_ext_frame_boundary" might be a better name for all of these.

framework/decode/vulkan_offscreen_swapchain.cpp
```diff
@@ -291,6 +270,256 @@ VkResult VulkanOffscreenSwapchain::QueuePresentKHR(VkResult
...
 }
 
+void VulkanOffscreenSwapchain::FrameBoundaryANDROID(PFN_vkFrameBoundaryANDROID           func,
+                                                    const VulkanDeviceInfo*              device_info,
+                                                    const VulkanSemaphoreInfo*           semaphore_info,
+                                                    const VulkanImageInfo*               image_info,
+                                                    VulkanInstanceInfo*                  instance_info,
+                                                    const graphics::VulkanInstanceTable* instance_table,
+                                                    const graphics::VulkanDeviceTable*   device_table,
+                                                    application::Application*            application)
```
MarkY-LunarG, line 280: application is unused?

framework/decode/vulkan_offscreen_swapchain.cpp
```diff
@@ -291,6 +270,256 @@ VkResult VulkanOffscreenSwapchain::QueuePresentKHR(VkResult
...
+                    frame_boundary->pNext      = modified_submit_infos[i].pNext;
+                    frame_boundary->imageCount = images->size();
+                    frame_boundary->pImages    = images->data();
+
+                    modified_submit_infos[i].pNext = frame_boundary;
+                }
+            }
+        }
+
+        // If the queue submission contains a frame boundary, increment the frame counter
```
MarkY-LunarG, line 412: Aren't you missing the case where it uses frameboundary already for this?  I.e. should we be tracking those as well here?

framework/decode/vulkan_offscreen_swapchain.cpp
```diff
@@ -291,6 +270,256 @@ VkResult VulkanOffscreenSwapchain::QueuePresentKHR(VkResult
...
+        if (!inserted_frame_boundaries.empty())
+        {
+            ++frame_boundary_.frameID;
```
MarkY-LunarG, line 523: Same comment about already existing frame boundary values here.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4095,7 +4114,18 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit(PFN_vkQueueSubmit
...
+                                         queue_info,
+                                         submitCount,
+                                         submit_infos,
+                                         submit_info_data,
+                                         fence_info,
+                                         instance_info,
+                                         instance_table,
+                                         device_info,
+                                         device_table,
+                                         application_.get(),
```
MarkY-LunarG, line 4127: Can't we just put this in the swapchain options since the swapchain is device specific and we should already have all this info during creation?

framework/decode/vulkan_virtual_swapchain.cpp
marius-pelegrin-arm: Yeah, I don't like it either, but I haven't found another way

framework/decode/vulkan_virtual_swapchain.h
marius-pelegrin-arm: No problem with it, as long as it still works

framework/decode/replay_options.h
marius-pelegrin-arm: This is the default place for initialization of all replay options. These magic numbers come from "magic constants" I erased at the top of `vulkan_swapchain.h`

The PR is cut into two commits and the first commit is just cleaning these default values and how the values are (not) transferred to the WSI

framework/decode/vulkan_offscreen_swapchain.cpp
marius-pelegrin-arm: In this class no, but the function is an overload common to all vulkan swapchains, and `application` is necessary to create a window in the `VulkanVirtualSwapchain`

framework/decode/vulkan_offscreen_swapchain.cpp
marius-pelegrin-arm: True, I will fix that.

framework/decode/vulkan_replay_consumer_base.cpp
marius-pelegrin-arm: The swapchain is not device specific, it is created along the VulkanReplayConsumer and there is only one no matter the number of devices

PR #2508, opened 2025-11-12
Block parser housekeeping

jzulauf-lunarg:
Two changes (in two commits)
1) rehome all BlockParser related components to their own files
2) update all visit pattern calls to be std::visit, eliminating "helper" that was more "confounder"

(based on Deferred Decompression change ... pending review) https://github.com/LunarG/gfxreconstruct/pull/2491

### DIFF
--- a/framework/decode/file_processor.cpp
+++ b/framework/decode/file_processor.cpp
@@ -277,6 +277,9 @@ bool FileProcessor::ProcessBlocks()
     // NOTE: To test deferred decompression operation uncomment next line
     // block_parser.SetDecompressionPolicy(BlockParser::DecompressionPolicy::kQueueOptimized);
 
+    ProcessVisitor  process_visitor(*this);
+    DispatchVisitor dispatch_visitor(decoders_, annotation_handler_);
+
     while (success)
     {
         PrintBlockInfo();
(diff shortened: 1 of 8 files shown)


framework/util/span.h
```diff
@@ -0,0 +1,236 @@
...
+
+    constexpr pointer       data() noexcept { return data_; }
+    constexpr const_pointer data() const noexcept { return data_; }
+
+    [[nodiscard]] constexpr size_type size() const noexcept { return size_; }
+    [[nodiscard]] constexpr bool      empty() const noexcept { return size_ == 0U; }
+
+    // Non-standard accessor
+    template <typename U>
+    [[nodiscard]] const U* GetDataAs() const noexcept
```
ncesario-lunarg, line 82: If this becomes a non-member function, what else is preventing `Span` from being replaced by `std::span`?

framework/decode/file_processor.cpp
```diff
@@ -306,18 +309,16 @@ bool FileProcessor::ProcessBlocks()
                     // parsed_block, though the block header is still valid.
                     ParsedBlock parsed_block = block_parser.ParseBlock(block_buffer);
 
-                    ProcessVisitor process_visitor(*this);
```
ncesario-lunarg, line 447: I was going to add a comment in the last PR about simplifying this, but vaguely recalled you already had something in mind :)

ncesario-lunarg (approved):
One question about (hopefully) being able to replace `Span` with `std::span` soon, but this is awesome. Thanks for doing this!

PR #2509, opened 2025-11-13
Reserve api call id 0x1348

bartosz-muszarski-arm:
Api call id reservation for an upcoming extension


fabian-lunarg:
this collided with a header-update, the assigned ID will need to get pushed further down

MarkY-LunarG (approved):

PR #2511, opened 2025-11-14
DRY FileTransformer by converting to BlockParser

jzulauf-lunarg:
Removed all duplicate block parsing code and convert to visitor style processing of the ParsedBlock Args variant.

Based on Block parser housekeeping #2508

### DIFF
--- a/framework/decode/file_transformer.cpp
+++ b/framework/decode/file_transformer.cpp
@@ -32,17 +32,13 @@ GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
 FileTransformer::FileTransformer() :
-    input_file_(nullptr), output_file_(nullptr), bytes_read_(0), bytes_written_(0),
-    error_state_(kErrorInvalidFileDescriptor), loading_state_(false)
+    input_file_(std::make_shared<FileInputStream>()), output_file_(nullptr), bytes_read_(0), bytes_written_(0),
+    error_state_(kErrorInvalidFileDescriptor), loading_state_(false), pool_(util::HeapBufferPool::Create())
 {}
 
 FileTransformer::~FileTransformer()
 {
-    if (input_file_ != nullptr)
-    {
-        fclose(input_file_);
-    }
-
+    // Note: input_file_ is closed by destructor
     if (output_file_ != nullptr)
     {
         fclose(output_file_);
@@ -59,11 +55,12 @@ bool FileTransformer::Initialize(const std::string& input_filename,
 
     bool success = false;
 
-    int32_t result = util::platform::FileOpen(&input_file_, input_filename.c_str(), "rb");
+    GFXRECON_ASSERT(input_file_ != nullptr);
+    bool open_input = input_file_->Open(input_filename.c_str());
 
-    if ((result == 0) && (input_file_ != nullptr))
+    if (open_input && input_file_->IsOpen())
     {
-        result = util::platform::FileOpen(&output_file_, output_filename.c_str(), "wb");
+        int32_t result = util::platform::FileOpen(&output_file_, output_filename.c_str(), "wb");
 
         if ((result == 0) && (output_file_ != nullptr))
         {
@@ -83,21 +80,35 @@ bool FileTransformer::Initialize(const std::string& input_filename,
 
     if (success)
     {
+        // We wait until after "ProcessFileHeader" as that is where compressor_ is initialized
+        auto err_handler = [this](BlockIOError err, const char* message) { HandleBlockReadError(err, message); };
+        block_parser_ =
+            std::make_unique<BlockParser>(BlockParser::ErrorHandler{ err_handler }, pool_, compressor_.get());
+        success = block_parser_ != nullptr;
+        if (success)
+        {
+            block_parser_->SetDecompressionPolicy(BlockParser::kNever);
+        }
+        else
+        {
+            error_state_ = kErrorOpeningFile;
+        }
+    }
+
+    if (success)
+    {
         error_state_ = kErrorNone;
     }
     else
     {
-        if (input_file_ != nullptr)
-        {
-            fclose(input_file_);
-            input_file_ = nullptr;
-        }
+        input_file_->Close();
 
         if (output_file_ != nullptr)
         {
             fclose(output_file_);
             output_file_ = nullptr;
         }
+        block_parser_.reset();
     }
 
     return success;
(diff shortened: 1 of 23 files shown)


framework/decode/parsed_block.h
```diff
@@ -73,10 +73,9 @@ class ParsedBlock
     enum BlockState
     {
         kInvalid = 0,        // Set on read error (typically block size doesn't match expected parsed size)
-        kUnknown,            // Set when block is of an unknown type (no parsing done beyond header)
+        kRawBlock,           // Set when block is of an unknown type (no parsing done beyond header)
```
MarkY-LunarG: The Raw block doesn't communicate "unknown" as well as the old value in my opinion.  In fact this seems confusing because it indicates a "decompressed" block in my verbage when I read this.

framework/decode/file_transformer.cpp
```diff
@@ -81,23 +82,34 @@ bool FileTransformer::Initialize(const std::string& input_filename,
...
 
+    if (success)
+    {
+        // We wait until after "ProcessFileHeader" as that is where compressor_ is initialized
+        auto err_handler = [this](BlockIOError err, const char* message) { HandleBlockReadError(err, message); };
+        block_parser_ =
+            std::make_unique<BlockParser>(BlockParser::ErrorHandler{ err_handler }, pool_, compressor_.get());
+        success = block_parser_ != nullptr;
+        if (success)
+            block_parser_->SetDecompressionPolicy(BlockParser::kNever);
```
MarkY-LunarG, line 90: Nit : Braces around this code under if.

tools/compress/compression_converter.cpp
```diff
@@ -403,525 +284,264 @@ bool CompressionConverter::WriteMethodCall(format::ApiCallId call_id,
...
 }
 
-bool CompressionConverter::WriteFillMemoryResourceValueMetaData(const format::BlockHeader& block_header,
-                                                                format::MetaDataId         meta_data_id)
+decode::FileTransformer::VisitResult
+CompressionConverter::WriteMetaData(const decode::InitDx12AccelerationStructureArgs& args)
 {
-    format::FillMemoryResourceValueCommandHeader rv_cmd;
-    bool                                         success = ReadBytes(&rv_cmd.thread_id, sizeof(rv_cmd.thread_id));
+    assert(format::GetMetaDataType(args.meta_data_id) == format::MetaDataType::kInitDx12AccelerationStructureCommand);
```
MarkY-LunarG: GFXRECON_ASSERT

tools/compress/compression_converter.cpp
```diff
@@ -403,525 +284,264 @@ bool CompressionConverter::WriteMethodCall(format::ApiCallId call_id,
...
-        {
-            if (!ReadParameterBuffer(data_size))
-            {
-                HandleBlockReadError(kErrorReadingBlockData,
-                                     "Failed to read init DX12 acceleration structure meta-data block");
-                return false;
-            }
-        }
+    // The header needs to be a copy as we are rewriting it below
+    format::InitSubresourceCommandHeader init_cmd = args.command_header;
```
MarkY-LunarG, line 466: Should these be references to avoid a copy?

tools/compress/compression_converter.cpp
```diff
@@ -72,190 +72,67 @@ bool CompressionConverter::WriteFileHeader(const format::FileHeader&
...
-
-        if (format::IsBlockCompressed(block_header.type))
-        {
-            success = ReadBytes(&uncompressed_size, sizeof(uncompressed_size));
-
-            if (success)
-            {
-                parameter_buffer_size -= sizeof(uncompressed_size);
-
-                GFXRECON_CHECK_CONVERSION_DATA_LOSS(size_t, uncompressed_size);
```
MarkY-LunarG, line 95: AFAICT, you're dropping a bunch of these.  I'm not sure how useful they were in the past, but wondering if you did this intentionally?

tools/compress/compression_converter.cpp
```diff
@@ -403,525 +284,264 @@ bool CompressionConverter::WriteMethodCall(format::ApiCallId call_id,
...
-                                  "Failed to write init DX12 acceleration structure meta-data block header");
-            return false;
-        }
-        if (!WriteBytes(geom_descs.data(),
-                        sizeof(format::InitDx12AccelerationStructureGeometryDesc) * geom_descs.size()))
-        {
-            HandleBlockWriteError(kErrorWritingBlockHeader,
-                                  "Failed to write init DX12 acceleration structure geometry block");
-            return false;
-        }
```
MarkY-LunarG, line 843: Are we no longer writing the geometry content?

tools/compress/compression_converter.cpp
```diff
@@ -403,525 +284,264 @@ bool CompressionConverter::WriteMethodCall(format::ApiCallId call_id,
...
 
-            if (!ReadCompressedParameterBuffer(compressed_size, data_size, &uncompressed_size))
-            {
-                HandleBlockReadError(kErrorReadingCompressedBlockData,
-                                     "Failed to read init DX12 acceleration structure meta-data block");
-                return false;
-            }
+decode::FileTransformer::VisitResult CompressionConverter::WriteMetaData(const decode::InitSubresourceArgs& args)
+{
+    assert(format::GetMetaDataType(args.meta_data_id) == format::MetaDataType::kInitSubresourceCommand);
```
MarkY-LunarG: Another "assert".  Won't point out any more.

tools/compress/compression_converter.cpp
jzulauf-lunarg: "// The header needs to be a copy as we are rewriting it below"

tools/compress/compression_converter.cpp
jzulauf-lunarg: These are are replaced by the common decompression code, now implemented (with conversion checks) in BlockParser and ParsedBlock

tools/compress/compression_converter.cpp
jzulauf-lunarg: This looks like an error.

jzulauf-lunarg: @MarkY-LunarG  -- it's just diff losing it's place.

tools/compress/compression_converter.cpp
jzulauf-lunarg: @MarkY-LunarG -- inherited most of these, updated all.

jzulauf-lunarg:
Several additional cleanups from extended review of code.

tools/optimize/file_optimizer.h
```diff
@@ -44,18 +44,22 @@ class FileOptimizer : public decode::FileTransformer
...
+    VisitResult FilterMetaData(const decode::InitBufferArgs& args);
+    VisitResult FilterMetaData(const decode::InitImageArgs& args);
+    template <typename Args>
+    VisitResult FilterMetaData(const Args& args)
+    {
+        return kNeedsPassthrough;
+    }
+
+    // bool FilterInitBufferMetaData(const format::BlockHeader& block_header, format::MetaDataId meta_data_id);
+    // bool FilterInitImageMetaData(const format::BlockHeader& block_header, format::MetaDataId meta_data_id);
```
MarkY-LunarG: Remove?

framework/decode/parsed_block.cpp
```diff
@@ -76,14 +76,18 @@ void ParsedBlock::Decompress(BlockParser& parser)
...
             auto uncompressed_size   = GetUncompressedSize(args);
-            auto uncompressed_buffer = parser.DecompressSpan(compressed_span, uncompressed_size);
-            // Patch the data buffer pointer, and shift ownership of the backing store to the parsed block
-            args.data = uncompressed_buffer.template GetAs<const uint8_t>();
-            UpdateUncompressedStore(std::move(uncompressed_buffer));
+            auto result              = parser.DecompressSpan(compressed_span, uncompressed_size);
+            if (result.success)
+            {
+                // Patch the data buffer pointer, and shift ownership of the backing store to the parsed block
+                args.data = result.decompressed_store.template GetAs<const uint8_t>();
```
MarkY-LunarG, line 83: Won't this fail if `result.decompresssed_store` is empty (block_parser.cpp line 144)?

jzulauf-lunarg: decompress_store empty is correct when compressed store is empty GetAs will return a nullptr, which is correct for an empty args.data

tools/optimize/file_optimizer.h
jzulauf-lunarg: yep. done.

PR #2512, opened 2025-11-14
Capture CreateRayTracingPipelinesKHR with Deferred Operation and VVL

locke-lunarg:
When gfxr captures CreateRayTracingPipelinesKHR  with  Deferred Operation and VVL, if `VK_INSTANCE_LAYERS` are `VK_LAYER_KHRONOS_validation;VK_LAYER_LUNARG_gfxreconstruct`,  the capture crashes because the pipeline is nullptr in `vkCmdBindPipeline`, but if `VK_INSTANCE_LAYERS` are `VK_LAYER_LUNARG_gfxreconstruct;VK_LAYER_KHRONOS_validation`, the capture is successful.

The crash reason is:
For capture, when create ray tracing pipelines with deferred operation, it will return nullptr of pipeline and save the pointer of pipeline. And then when it runs `vkDeferredOperationJoinKHR` or `vkGetDeferredOperationResultKHR`, it will update a valid handle to the pipeline, so the source program will receive the valid handle.

However, in VVL:
```
VkResult Device::CreateRayTracingPipelinesKHR(
.
.
.
    if (deferredOperation != VK_NULL_HANDLE) {
        unwrapped_pipelines = std::make_shared<std::vector<VkPipeline>>(createInfoCount);
        returned_pipelines = unwrapped_pipelines->data();
    }

    VkResult result = device_dispatch_table.CreateRayTracingPipelinesKHR(
        device, deferredOperation, pipelineCache, createInfoCount, (const VkRayTracingPipelineCreateInfoKHR *)local_pCreateInfos,
        pAllocator, returned_pipelines);
```
VVL creates a new pipeline, `returned_pipelines = unwrapped_pipelines->data();`. If VVL runs before gfxr, gfxr gets a new pipeline that VVL creates, not from the source program. In this case, the pipeline in the source program is always nullptr to crash in `vkCmdBindPipeline`. gfxr couldn't update it.

However, if gfxr runs before VVL, gfxr get the pipeline from the source program and could update a valid handle to it.

The solution is to return a valid handle of pipeline in CreateRayTracingPipelinesKHR since the driver actually returns a valid handle, even if the result is `VK_OPERATION_DEFERRED_KHR`, so gfxr doesn't need to update the handle of pipeline later. However, I wonder if the driver always returns a valid handle of pipeline when the result is `VK_OPERATION_DEFERRED_KHR`.

Test code: I added deferred operation in raytracingbasic.
https://github.com/locke-lunarg/Vulkan/tree/deferred-op

### DIFF
--- a/framework/encode/vulkan_capture_manager.cpp
+++ b/framework/encode/vulkan_capture_manager.cpp
@@ -1395,26 +1395,14 @@ VulkanCaptureManager::OverrideCreateRayTracingPipelinesKHR(VkDevice             
                                                                 deferred_operation_wrapper->p_allocator,
                                                                 deferred_operation_wrapper->pipelines.data());
 
-            if ((result == VK_OPERATION_NOT_DEFERRED_KHR) || (result == VK_SUCCESS))
+            if ((result == VK_OPERATION_NOT_DEFERRED_KHR) || (result == VK_SUCCESS) ||
+                (result == VK_OPERATION_DEFERRED_KHR))
             {
-                // VK_OPERATION_NOT_DEFERRED_KHR means the operation successfully completed immediately, so here we copy
-                // the created pipelines to original array which is used to store them by target application.
-                //
-                // Note:
-                //       VK_OPERATION_DEFERRED_KHR means the operation is successfully deferred, but the pipeline
-                //       creation might not be finished. We must therefore lean on
-                //       vkDeferredOperationJoinKHR/vkGetDeferredOperationResultKHR to get the final result. If the
-                //       result indicated the operation has finished, then the created pipelines are ready for use.
-                //
-                //       By Vulkan doc of VK_KHR_deferred_host_operations, if deferred operation object was
-                //       provided and the operation successfully completed immediately without deferring, the
-                //       command return VK_OPERATION_NOT_DEFERRED_KHR. Some hardware/driver may return VK_SUCCESS on
-                //       such case, so same handling is proceeded with VK_SUCCESS.
-                //
                 std::memcpy(
                     pPipelines, deferred_operation_wrapper->pipelines.data(), sizeof(VkPipeline) * createInfoCount);
             }
-            else if (result == VK_OPERATION_DEFERRED_KHR)
+
+            if (result == VK_OPERATION_DEFERRED_KHR)
             {
                 const std::lock_guard<std::mutex> lock(deferred_operation_mutex);
                 deferred_operation_wrapper->pending_state = true;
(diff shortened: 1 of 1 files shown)


bradgrantham-lunarg:
> If VVL runs before gfxr, gfxr gets a new pipeline that VVL creates, not from the source program. In this case, the pipeline in the source program is always nullptr to crash in vkCmdBindPipeline. gfxr couldn't update it.

According to the spec, `Parameters to the command requesting a deferred operation may be accessed by the implementation at any time until the deferred operation enters the complete state`.   That implies to me the implementation (and therefore any layers) _might not_ access the parameters until later, after CreateRayTracingPipelines has returned.  So the application, VVL, and GFXR all have to leave around any memory allocated to pPipelines until the deferred operation is complete.  If the application has an uninitialized pipeline handle at the time of CmdBindPipeline that implies to me that the application or the layers either didn't copy memory back upchain or the application or layers freed memory before the pipeline handles were written to pPipelines.  Can you detail more about what the application and VVL do at the time of vkGetDeferredOperationResultKHR?

bradgrantham-lunarg:
> > If VVL runs before gfxr, gfxr gets a new pipeline that VVL creates, not from the source program. In this case, the pipeline in the source program is always nullptr to crash in vkCmdBindPipeline. gfxr couldn't update it.
> 
> According to the spec, `Parameters to the command requesting a deferred operation may be accessed by the implementation at any time until the deferred operation enters the complete state`. That implies to me the implementation (and therefore any layers) _might not_ access the parameters until later, after CreateRayTracingPipelines has returned. So the application, VVL, and GFXR all have to leave around any memory allocated to pPipelines until the deferred operation is complete. If the application has an uninitialized pipeline handle at the time of CmdBindPipeline that implies to me that the application or the layers either didn't copy memory back upchain or the application or layers freed memory before the pipeline handles were written to pPipelines. Can you detail more about what the application and VVL do at the time of vkGetDeferredOperationResultKHR?

@locke-lunarg

locke-lunarg:
@bradgrantham-lunarg  Sorry for reply late.

Regarding VVL: Inside CreateRayTracingPipelinesKHR, VVL allocates memory for new pipelines and passes it to the CreateRayTracingPipelinesKHR call. Once the call returns valid handles, VVL immediately updates the application's pipeline handles. Later, in DeferredOperationJoinKHR or GetDeferredOperationResultKHR, VVL only updates pipeline tracking info when the deferred operation completes.

However, regarding GFXR-capture: It also allocates memory for new pipelines and passes it to the CreateRayTracingPipelinesKHR call, saving the application's pipeline memory addresses. Although the call returns valid handles immediately, GFXR leaves the application's pipeline handles as nullptr at this stage. It only copies the handles to the saved addresses later, during DeferredOperationJoinKHR or GetDeferredOperationResultKHR, once the deferred operation is complete.

The issue arises when both layers are active (App -> VVL -> GFXR): VVL passes its own temporary pipeline memory to GFXR. GFXR saves VVL's memory address but delays writing the handles until the deferred operation is complete. Since VVL sees nullptr (because GFXR hasn't written back yet), VVL cannot copy any valid handles back to the application. Even when the deferred operation is completed, GFXR copies the handles to the VVL memory, not the app's, leading to a crash if accessed.

I think GFXR-capture should write the pipeline handles to the application's pipeline memory immediately after CreateRayTracingPipelinesKHR returns valid handles, just as VVL does, instead of waiting for the deferred operation to finish

arno-lunarg:
This PR should not be needed, it was a VVL issue!

locke-lunarg:
@arno-lunarg Thank you for fixing!

arno-lunarg:
No problem, thank you for pointing out this issue!

PR #2513, opened 2025-11-14
Document the process for debugging the Android capture layer

nickdriscoll-lunarg:
This PR adds `DEBUG_android.md`, a document detailing how to use `lldb` to debug the capture layer on Android devices.

### DIFF
--- a/DEBUG_android.md
+++ b/DEBUG_android.md
@@ -17,19 +17,20 @@ on Android using lldb-server.***
 ## Index
 
 1. [Problem](#problem)
-2. [Setting up lldb-server](#lldb-server-setup)
+2. [Setting up lldb-server](#setting-up-lldb-server)
 3. [Client setup](#client-setup)
    1. [VSCode](#vscode)
    2. [CLI](#cli)
+4. [Waiting for debugger](#waiting-for-debugger)
 
 ## Problem
 
 Android apps under development can be debugged easily through Android Studio. However, the GFXR capture layer is
 a shared object file that is loaded by the, typically non-debug, application. This document illustrates a 
-method by which the `lldb-server` binary included with the Android SDK can be used for debugging
+method by which the `lldb-server` binary included with the Android NDK can be used for debugging
 a .so built in debug mode with any Android app.
 
-## lldb-server setup
+## Setting up lldb-server
 
 The first step is to locate and copy the lldb-server binary to a location on the Android device,
 making sure to give it execute permissions as well.
@@ -67,7 +68,7 @@ List of devices attached
 We document two methods for attaching a client debugger to lldb-server: VScode, and CLI
 
 ### VScode
-To setup VScode to be the client debugger, you need to set up your
+To enable VScode to be the client debugger, you need to set up your
 `launch.json` to be like the following:
 ```
 {
@@ -119,12 +120,12 @@ settings append target.exec-search-paths $PWD/android/layer/build/intermediates/
 attach $(adb shell pidof <apk name e.g. com.example.VkCube>)
 ```
 
-Then (assuming lldb is in $PATH) launch `lldb` with `lldb -s /tmp/lldbinit`
-From here `lldb` will connect, and 
+Then (assuming `lldb` is in $PATH) launch `lldb` with `lldb -s ./lldbinit`
+From here `lldb` will connect, and present you with the typical `lldb` CLI.
 
 
 
-## Wait for debugger
+## Waiting for debugger
 
 It is often desired to be able to attach the debugger to the app as soon as it launches.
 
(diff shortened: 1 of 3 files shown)


MarkY-LunarG:
I think we need to start putting more of these documents under the docs folder.  I believe our top folder is getting polluted.

DEBUG_android.md
```diff
@@ -0,0 +1,150 @@
+<!-- markdownlint-disable MD033 -->
+<!-- markdownlint-disable MD041 -->
+<p align="left"><img src="https://vulkan.lunarg.com/img/NewLunarGLogoBlack.png" alt="LunarG" width=263 height=113 /></p>
+
+[![Creative Commons][1]][2]
+
+[1]: https://i.creativecommons.org/l/by-nd/4.0/88x31.png "Creative Commons License"
+[2]: https://creativecommons.org/licenses/by-nd/4.0/
+
+Copyright &copy; 2018-2025 LunarG, Inc.
```
MarkY-LunarG: Nit: 2025

docs/DEBUG_android.md
```diff
@@ -0,0 +1,150 @@
...
+## Problem
+
+Android apps under development can be debugged easily through Android Studio. However, the GFXR capture layer is
+a shared object file that is loaded by the, typically non-debug, application. This document illustrates a 
+method by which the `lldb-server` binary included with the Android NDK can be used for debugging
+a .so built in debug mode with any Android app.
+
+## Setting up lldb-server
+
+The first step is to locate and copy the lldb-server binary to a location on the Android device,
```
MarkY-LunarG, line 36: You briefly mention that the layer is loaded as part of an application, but I would refer to another doc on how this is done.  People might end up here thinking they are debugging their application with a capture layer.

docs/DEBUG_android.md
```diff
@@ -0,0 +1,150 @@
...
+```
+> adb devices
+List of devices attached
+44021FDJH00229  device
+```
+
+We document two methods for attaching a client debugger to lldb-server: VScode, and CLI
+
+### VScode
+To enable VScode to be the client debugger, you need to set up your
```
MarkY-LunarG, line 72: @mikes-lunarg have you used this method?  I've never used it and can't comment on the completion of this documented process.

docs/DEBUG_android.md
nickdriscoll-lunarg: I reworded the Problem section to omit the word "loaded" and hopefully make it clearer that this is about debugging the capture layer with a method that applies generally to Android shared libraries.

docs/DEBUG_android.md
mikes-lunarg: Yep I have tested this method. The roundabout way of getting the target PID is a little goofy but it works!

docs/DEBUG_android.md
```diff
@@ -0,0 +1,151 @@
...
+
+Unfortunately, since we only have control over the layer, the best we can do is construct
+an arbitrary spinlock in `VulkanCaptureManager::OverrideCreateInstance()` that we release
+once the debugger has attached.
+
+At the start of `VulkanCaptureManager::OverrideCreateInstance`, add:
+```
+while (util::platform::GetEnv("debug.gfxrecon.debug_wait") == "1") {
+    usleep(100000);
+}
```
mikes-lunarg, line 144: Should we consider actually committing this loop behind a `#ifdef GFXRECON_DEBUG_BUILD` guard?

docs/DEBUG_android.md
```diff
@@ -0,0 +1,151 @@
...
+If you prefer CLI-based debugging, then create an `lldbinit` file at the root of the repo with the following contents:
+```
+platform select remote-android
+platform connect connect://<your-device-id here>:9999
+process handle -p true -s false -n false SIGSEGV
+process handle -p true -s false -n false SIGPWR
+process handle -p true -s false -n false SIGXCPU
+process handle -p true -s false -n false SIGBUS
+settings append target.exec-search-paths $PWD/android/layer/build/intermediates/cxx/Debug/<some numbers such as '3m4on72q'>/obj/<target device architecture e.g. arm64-v8a>/
+attach $(adb shell pidof <apk name e.g. com.example.VkCube>)
```
mikes-lunarg: I don't think this will work as written. LLDB won't interpret the `$PWD` env var or `$(adb shell pidof)` subshell command. In my original doc I got around this limitation with a wrapper shell script that would bake the values into the lldbinit file and then invoke the debugger.

USAGE_android.md
```diff
@@ -19,6 +19,10 @@ to one of these other documents:
  * [GFXReconstruct for Desktop Vulkan](./USAGE_desktop_Vulkan.md)
  * [GFXReconstruct for Desktop D3D12](./USAGE_desktop_D3D12.md)
 
+If you are looking for how to debug the Android capture layer, please
+refer to this document:
```
mikes-lunarg: Personal preference: This reads a little too informal, consider an alternative:
> Additional instructions for debugging the GFXReconstruct capture layer on Android can be found here:

mikes-lunarg (changes requested):

docs/DEBUG_android.md
nickdriscoll-lunarg: Understood. I assumed it wasn't working for me just bc I'm on Windows. I'll make the change

USAGE_android.md
nickdriscoll-lunarg: Fixed.

docs/DEBUG_android.md
nickdriscoll-lunarg: I don't think that's a good idea and I don't understand why we'd want that.

If you're in deep enough to need to attach the debugger to the capture layer, then you're certainly capable of inserting this loop. Additionally, if we added this and then you built GFXR in debug, unaware of the loop, you might find that the layer is hanging and you don't understand why.

Granted, that scenario implies that `debug.gfxrecon.debug_wait` was somehow set without the user's knowledge, but without a compelling reason to include the loop, I think this is a good enough reason to _not_ do it.

mikes-lunarg: Fair, the app locking up with no indication of why seems like a pretty good reason to leave this out by default

mikes-lunarg (approved):

PR #2514, opened 2025-11-14
 Prevent "partial block" false positives

jzulauf-lunarg:
Improve block read status reporting to allow correct control of error messages.

@bradgrantham-lunarg  -- this should address the issue we discussed earlier.

### DIFF
--- a/framework/decode/file_processor.cpp
+++ b/framework/decode/file_processor.cpp
@@ -494,27 +494,46 @@ bool FileProcessor::ProcessBlocks()
 // the correct sizing of the block payload are done by the caller
 bool FileProcessor::ReadBlockBuffer(BlockParser& parser, BlockBuffer& block_buffer)
 {
-    bool success = parser.ReadBlockBuffer(GetCurrentFile().active_file, block_buffer);
-    if (success)
+    bool           success = true;
+    BlockReadError status  = parser.ReadBlockBuffer(GetCurrentFile().active_file, block_buffer);
+    if (status == kErrorNone)
     {
         bytes_read_ += block_buffer.Size();
     }
     else
     {
-        HandleBlockReadError(kErrorReadingBlockData, "Failed to read block body data");
+        // Caller handles end of file on block boundaries
+        if (status != kEndOfFile)
+        {
+            HandleBlockReadError(status, "Failed to read next block");
+        }
+        success = false;
     }
     return success;
 }
 
 // Parse the block header and load the whole block into a block buffer
-bool BlockParser::ReadBlockBuffer(FileInputStreamPtr& input_stream, BlockBuffer& block_buffer)
+BlockReadError BlockParser::ReadBlockBuffer(FileInputStreamPtr& input_stream, BlockBuffer& block_buffer)
 {
     using BlockSizeType = decltype(format::BlockHeader::size);
     BlockSizeType block_size;
-    bool          success = input_stream->PeekBytes(&block_size, sizeof(block_size));
-    if (success)
+    BlockReadError status = kErrorNone;
+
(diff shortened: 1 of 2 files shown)


jzulauf-lunarg:
@panos-lunarg @bradgrantham-lunarg spot checked with a files with and without partial block.  Neither false positive, nor negative.  Clean CI.

framework/decode/file_processor.cpp
```diff
@@ -494,27 +494,46 @@ bool FileProcessor::ProcessBlocks()
...
-bool BlockParser::ReadBlockBuffer(FileInputStreamPtr& input_stream, BlockBuffer& block_buffer)
+BlockReadError BlockParser::ReadBlockBuffer(FileInputStreamPtr& input_stream, BlockBuffer& block_buffer)
 {
     using BlockSizeType = decltype(format::BlockHeader::size);
     BlockSizeType block_size;
-    bool          success = input_stream->PeekBytes(&block_size, sizeof(block_size));
-    if (success)
+    BlockReadError status = kErrorNone;
+
+    if (!input_stream->PeekBytes(&block_size, 1))
```
MarkY-LunarG, line 522: Aren't you copying a byte here on this check if it succeeds.  So aren't you doing extra work each call?  Why not just incorporate the check in PeekBytes?

jzulauf-lunarg: The way the interface is written is based on the underlying util::platform.h interface -- doesn't return how many bytes it read, just a boolean if it read all of them.  The only way to know that not even a single byte of the next block is present is to try and read a single byte.  Reading sizeof(size) could return false (and IsEof) with 0-7 bytes between readpos and eof, and eof isn't set until you try to read past the last byte.  So I have to peek 1 (which will fread 1, and then peek 8 (which will fread 7 and fill the peek buffer to 8 bytes (sizeof(size)) to let me know how big a ReadSpan to do to grab the whole block.

But the only way to know that not even 1 byte is available past readpos (with the current fread wrappers (or maybe some other FSTREAM query calls I'm not aware of) is to read one.  (Hmmmm, maybe I could stat the file and check *that* against bytes_read_...)

framework/decode/file_processor.cpp
```diff
@@ -494,27 +494,46 @@ bool FileProcessor::ProcessBlocks()
 // the correct sizing of the block payload are done by the caller
 bool FileProcessor::ReadBlockBuffer(BlockParser& parser, BlockBuffer& block_buffer)
 {
-    bool success = parser.ReadBlockBuffer(GetCurrentFile().active_file, block_buffer);
-    if (success)
+    bool           success = true;
+    BlockReadError status  = parser.ReadBlockBuffer(GetCurrentFile().active_file, block_buffer);
+    if (status == kErrorNone)
```
bradgrantham-lunarg, line 499: I feel like this flow is difficult to follow.  I imagine this could be simplified into just the case `(status != kErrorNone)` which returns `false` and otherwise returns `true` with no `success` but I don't want to block this PR.  So simplify this at the next opportunity?

jzulauf-lunarg: We have to detect between EOF (at beginning of read) > 0, None ==0, and Error < 0, in order to correctly suppress the HandleBlockReadError for Eof, but still return false, such that the callers will do the right thing.

If I was going to simplify, I'd plumb the status up to the caller and eliminate local error reporting.  Return bool is actually lossy (two answers for N > 2 states), and we're counting on the callers current logic to do the right thing

jzulauf-lunarg:
@bradgrantham-lunarg -- was there ever an issue opened on this? If so LMK and I'll update it.

PR #2520, opened 2025-11-18
Fix issues around AS track&replace (-m rebind)

fabian-lunarg:
Vulkan acceleration-structures (AS) are referenced either through a `VkAccelerationStructureKHR` handle
or their associated device-address. 
Aliasing or missing handles are (will be?) explicitly allowed, so the tracking has recently been reworked
to focus on device-addresses.

This PR fixes cases when the device address is not (yet) known and will query+update associated AS&buffer addresses adhoc, if necessary.
Symptom were sometimes missing top-lvl AS during a replay when using `-m rebind`.
 
VulkanAddressReplacer is now using replay- instead of capture-addresses for tracking internal resources.
latter 'can' always be obtained, while former 'might' be unknown at replacement-time.

Another minor fix is related to internally used host-memory, which is now using
`VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT` unconditionally.
previously we attempted to obtain memory using `VK_MEMORY_PROPERTY_HOST_CACHED_BIT`, 
which now showed issues on a previously untested config.

Finally a minor fix to cleanup/lifetime of an internally used buffer (`hashmap_control_block_bda_binary_`)

### DIFF
--- a/framework/decode/vulkan_address_replacer.h
+++ b/framework/decode/vulkan_address_replacer.h
@@ -476,8 +476,9 @@ class VulkanAddressReplacer
     std::vector<bda_element_t> storage_bda_binary_;
 
     // storage- and control-buffers for a global hashmap acting as address-filter
-    buffer_context_t hashmap_storage_bda_binary_       = {};
-    buffer_context_t hashmap_control_block_bda_binary_ = {}, hashmap_control_block_bda_binary_prev_ = {};
+    buffer_context_t hashmap_storage_bda_binary_            = {};
+    buffer_context_t hashmap_control_block_bda_binary_      = {};
+    buffer_context_t hashmap_control_block_bda_binary_prev_ = {};
 
     // pipeline-contexts per command-buffer
     std::unordered_map<VkCommandBuffer, std::vector<pipeline_context_t>> pipeline_context_map_;


framework/decode/vulkan_address_replacer.h
```diff
@@ -476,7 +477,7 @@ class VulkanAddressReplacer
 
     // storage- and control-buffers for a global hashmap acting as address-filter
     buffer_context_t hashmap_storage_bda_binary_       = {};
-    buffer_context_t hashmap_control_block_bda_binary_ = {};
+    buffer_context_t hashmap_control_block_bda_binary_ = {}, hashmap_control_block_bda_binary_prev_ = {};
```
MarkY-LunarG: Nit: We don't allow this variable grouping.  One lien per variable definition.

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -2240,7 +2255,12 @@ void VulkanAddressReplacer::update_global_hashmap(VkCommandBuffer command_buffer
             GFXRECON_ASSERT(previous_hashmap_control_block && previous_hashmap_storage_bda_binary);
             GFXRECON_ASSERT(hashmap_control_block_bda_binary_.buffer == VK_NULL_HANDLE &&
                             hashmap_storage_bda_binary_.buffer == VK_NULL_HANDLE);
-            auto& previous_control_block = *reinterpret_cast<gpu_array_t*>(previous_hashmap_control_block->mapped_data);
+
+            // defer cleanup of previous control-block
+            hashmap_control_block_bda_binary_prev_ = std::move(*previous_hashmap_control_block);
```
MarkY-LunarG, line 2260: I'm surprised this was a necessary change, shouldn't the old value have survived through the `CmdPushConstants` call?   I would think if anything hada problem it would be the `rehash_params_t` variable.

fabian-lunarg: that's right `rehash_params_t` itself is copied into push-constant and so is fine. 
just for the duration of the rehash both buffers referenced in it need to stay alive. 
then depending on **how** we submit (inject into existing cmd-buffer vs. local cmd-buffer+fence-sync),
the older buffer could end up being deleted too early. 
the re-hash is not necessary very often, since the hashmap will grow exponentially. 
so it took a while until I actually saw this in a replay

framework/decode/vulkan_address_replacer.h
fabian-lunarg: fixed

PR #2523, opened 2025-11-18
Multi CopyBufferRegion over to same DXR buffer

JerryAMD:
Game updated same DXR buffer data from one stagging buffer but different offset. If there were a few of ExecuteIndirect(DispatchRays) calls within one commandlist, GFXR only tracked the latest CopyBufferRegion 
and made the latest stagging buffer resource as DXR buffer source.
For example,

// Copy buffer1 offset 100 data over to EI_DispatchRays offset_0 
CL::CopyBufferRegion(EI_DispatchRays, offset_0, buffer1, offset_100) CL::SetStateObject1( SO_0)
CL::ExecuteIndirect(EI_DispatchRays, offset_0)

// Copy buffer1 offset 200 data over to EI_DispatchRays offset_0 
CL::CopyBufferRegion(EI_DispatchRays, offset_0, buffer1, offset_200) CL::SetStateObject1( SO_0)
CL::ExecuteIndirect(EI_DispatchRays, offset_0)

// Copy buffer1 data offset 300 over to EI_DispatchRays offset_0 
CL::CopyBufferRegion(EI_DispatchRays, offset_0, buffer1, offset_300) CL::SetStateObject1( SO_1)
CL::ExecuteIndirect(EI_DispatchRays, offset_0)
CL::Close()

// GFXR inject ResourceValueMap()
CQ::ExecuteCommandList(CL)

We have to track correct stagging buffer and its offset for each EI(), BuildAS() and DisptachRays() calls.

### DIFF
--- a/framework/decode/dx12_resource_value_mapper.cpp
+++ b/framework/decode/dx12_resource_value_mapper.cpp
@@ -501,6 +501,7 @@ void Dx12ResourceValueMapper::PostProcessExecuteIndirect(DxObjectInfo* command_l
                 command_signature_extra_info->byte_stride,
                 state_object_extra_info);
         }
+        SourceCopyResources(command_list_extra_info->resource_copies, command_list_extra_info->resource_value_info_map);
     }
 }
 
@@ -579,6 +580,8 @@ void Dx12ResourceValueMapper::PostProcessBuildRaytracingAccelerationStructure(
                 GFXRECON_LOG_ERROR("Unknown BuildRaytracingAccelerationStructure DescsLayout: %d",
                                    static_cast<int>(build_desc->Inputs.DescsLayout));
             }
+            SourceCopyResources(command_list_extra_info->resource_copies,
+                                command_list_extra_info->resource_value_info_map);
         }
         else
         {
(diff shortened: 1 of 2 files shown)


framework/decode/dx12_resource_value_mapper.cpp
```diff
@@ -501,6 +501,7 @@ void Dx12ResourceValueMapper::PostProcessExecuteIndirect(DxObjectInfo* command_l
                 command_signature_extra_info->byte_stride,
                 state_object_extra_info);
         }
+        SourceCopyResources(command_list_extra_info->resource_copies, command_list_extra_info->resource_value_info_map);
```
davidd-lunarg, line 504: Is there still a potential issue when `count_buffer` is used. In that case only one `ResourceValueType::kExecuteIndirectCountBuffer` RV is tracked?

JerryAMD: Yes. We only handled either `count_buffer` or `argument_buffer` been tracked to resource_value_info_map here. This solution is not a general one. And we also missed the case that ExecuteIndirect(DispatchRays) and CopyBuffer to the indirect DispatchRaysDesc buffer.

davidd-lunarg: Ok. I added this issue to track that additional work is needed to support indirectly referenced DXR buffers.
https://github.com/LunarG/gfxreconstruct/issues/2525
If there is any additional relevant info, please add that to the issue.

PR #2524, opened 2025-11-19
Track fence-states during capture, adjust statewrite

### LINKED ISSUE #2521: [replay] When using trimming, fence-synchronization hangs after state-init
I attempted the following:

capture: on win11, AMD Radeon 850M, driver 25.11.1 **using trimming**

replay:

- same device
-> replay works but I'm seeing a warning wrt. to `vkWaitForFences`:
```
[gfxrecon] INFO - Finished loading state for captured frame 101
================== Start timer (Frame: 2) ==================
[gfxrecon] WARNING - API call vkWaitForFences returned value VK_TIMEOUT that does not match return value from capture file: VK_SUCCESS.
```

- different device (linux, nvidia 580.82.09)
-> replay hangs at the exact location the warning was seen earlier.

checking a converted jsonl from the capture confirms the fence is initialized in un-signaled state,
followed by `vkWaitForFences` with max timeout. so this 'should' hang.

the reason the win/AMD driver skips over that issuing `VK_TIMEOUT` is a bit puzzling, 
since the used timeout is hardcoded as `std::numeric_limits<uint64_t>::max()`, but that's a driver issue imo.

but I think there is an issue with GFXR's fence-state after the trim state-init.
it looks like the fence was waiting for previous calls that are not part of the trimmed capture, 
so it cannot ever be signaled. 
in that case it needs to be created in signaled state instead, in order to avoid hanging.

used capture:
[2025_11_18_amd_850m_win_raytracingshadows_frames_101_through_600.zip](https://github.com/user-attachments/files/23605103/2025_11_18_amd_850m_win_raytracingshadows_frames_101_through_600.zip)

fabian-lunarg:
This PR adds state-tracking and adjust state-writes for `VkFence` objects.

when creating a trimmed capture, before state-write we wait via `vkDeviceWaitIdle`.
we rely on this and assume all in-flight fences have been signaled by then.
this seems to fail on certain configurations like described [here](https://github.com/LunarG/gfxreconstruct/issues/2521)

the PR fixes the issue with trimmed captures and avoids the intrinsic fence-hang in capture-files.
if a `VkFence` was either **signaled or in-flight** when a capture started, we will **initialize it in signaled state**. 

closes #2521


framework/encode/vulkan_state_writer.cpp
```diff
@@ -422,6 +422,7 @@ void VulkanStateWriter::WriteFenceState(const VulkanStateTable& state_table)
         bool                                  signaled       = wrapper->created_signaled;
 
         GetFenceStatus(device_wrapper, wrapper->handle, &signaled);
+        signaled = signaled || wrapper->in_flight;
```
fabian-lunarg, line 425: here we apply the knowledge about the fence-state

MarkY-LunarG: Should we set `wrapper->in_flight` to false here if signaled?  Also, shouldn't we check in `vkGetFenceStatus` as well because that's another place apps query the status with an explicit `vkWaitForFences`?

fabian-lunarg: > Should we set wrapper->in_flight to false here if signaled?

not sure. I think better to not write the flag here. if we were to capture via hotkey 
and create multiple captures in one session this should remain signaled (or 'in_flight') 
to avoid a potential hang in a successive state-write. so only clear that flag via `vkResetFences` or `vkWaitFences` (plus in case of successful `vkGetFenceStatus`)

>  Also, shouldn't we check in vkGetFenceStatus as well because that's another place apps query the status with an explicit vkWaitForFences?

good point, looking into that

fabian-lunarg: added a `PostProcess_vkGetFenceStatus` to clear `wrapper->in_flight` for already signaled fences

PR #2528, opened 2025-11-25
Fix issues with -m rebind, correction to use capture-time memory-indices

fabian-lunarg:
While testing various `-m rebind` scenarios (capture on X, replay on Y)
and ran into some weird issues when trying to map internal host-memory.
Reason was basically wrong usage of VulkanRebindAllocator,
feeding wrong memory-properties (we provided replay-time, but needs to be capture-time)
for internal memory-allocations.
The PR corrects this bug.


MarkY-LunarG (approved):
Looks good!

PR #2531, opened 2025-11-26
Correct differing function definitions

ryannecas:
In the d3d12/dxgi dll_main.cpp files, these release-capture functions have WINAPI in their definition, but they do not elsewhere. Adding WINAPI elsewhere to fix the conflict, which causes exceptions in x86


locke-lunarg:
Merged https://github.com/LunarG/gfxreconstruct/pull/2658. Closed this.

PR #2533, opened 2025-11-27
Reduce busy-waiting for vkGetQueryPoolResults & vkGetEventStatus

fabian-lunarg:
- currently we're busy-waiting for `vkGetQueryPoolResults` if `VK_SUCCESS` is expected.
-> can yield wrong result (running out of retries), doesn't match the original behavior and is inefficient
-> waiting seems more correct
- vkGetEventStatus is re-run in a loop, even if it already returned VK_EVENT_SET
-> this doesn't make sense imo and is corrected here

testing with a Unity/DXVK capture:
the PR reduces the amount of `vkGetEventStatus` by an order of magnitude (21Mio -> 2Mio)..

log-spam is reduced, not seeing this anymore:
`[gfxrecon] WARNING - API call vkGetQueryPoolResults returned value VK_NOT_READY that does not match return value from capture file: VK_SUCCESS.`

-> this could easily be seen as a bug. we go on with replay, after trying for `kMaxQueryPoolResultsRetries`
-> waiting seems like the better approach (similar to `vkGetFenceStatus`)
-> the overall replay-time/performance was not affected negatively, no significant change

~~still seeing lots of:~~
~~`[gfxrecon] WARNING - API call vkGetEventStatus returned value VK_EVENT_SET that does not match return value from capture file: VK_EVENT_RESET.`~~

~~which is true, just profoundly fills up the log and feels noisy.~~

-> this is now explicitly allowed and we will exclude warnings for `VK_EVENT_RESET` -> `VK_EVENT_SET`

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -1490,16 +1490,16 @@ void VulkanReplayConsumerBase::CheckResult(const char*                func_name,
     if (original != replay)
     {
         // check allow-listed functions
-        bool return_code_valid = kFunctionsAllowedToReturnDifferentCodeThanCapture.contains(func_name);
+        bool accept_return_code = kFunctionsAllowedToReturnDifferentCodeThanCapture.contains(func_name);
 
         // check allow-listed capture/replay VkResult-values
         if (const auto it = kResultValuesAllowedDifferentCodeThanCapture.find(original);
             it != kResultValuesAllowedDifferentCodeThanCapture.end())
         {
-            return_code_valid = return_code_valid || replay == it->second;
+            accept_return_code = accept_return_code || replay == it->second;
         }
 
-        if (!return_code_valid)
+        if (!accept_return_code)
         {
             if (replay < 0 && replay != VK_ERROR_FORMAT_NOT_SUPPORTED)
             {


bradgrantham-lunarg:
~~The QueryPoolResults loop came in with https://github.com/LunarG/gfxreconstruct/pull/542, which was related to DXVK.  Can you test your change against DXVK content?~~

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -1476,36 +1489,46 @@ void VulkanReplayConsumerBase::CheckResult(const char*                func_name,
 {
     if (original != replay)
     {
-        const bool is_func_allowed_to_differ = kFunctionsAllowedToReturnDifferentCodeThanCapture.count(func_name);
+        // allow-listed functions
+        bool is_func_allowed_to_differ = kFunctionsAllowedToReturnDifferentCodeThanCapture.contains(func_name);
```
MarkY-LunarG: With your new changes, I would like this variable name changed to something like `return_code_valid` instead.  I think that makes more sense than the current name.

bradgrantham-lunarg: I think a return could be "valid" and still not be allowed to differ.  The variable says exactly the meaning, which doesn't state anything about validity and states only whether GFXR cares that the return is different in replay than in capture.

fabian-lunarg: done, I settled on `accept_return_code`

bradgrantham-lunarg:
> ~The QueryPoolResults loop came in with #542, which was related to DXVK. Can you test your change against DXVK content?~

I am a little concerned that #542 added the retry count to avoid looping forever on a query pool Get.  If that condition could still happen then setting VK_QUERY_RESULT_WAIT_BIT would have the same problem our code had before #542.  (this was 4 years ago so maybe DXVK has changed so that this doesn't happen any more?  @werman )

fabian-lunarg:
> > ~The QueryPoolResults loop came in with #542, which was related to DXVK. Can you test your change against DXVK content?~
> 
> I am a little concerned that #542 added the retry count to avoid looping forever on a query pool Get. If that condition could still happen then setting VK_QUERY_RESULT_WAIT_BIT would have the same problem our code had before #542. (this was 4 years ago so maybe DXVK has changed so that this doesn't happen any more? @werman )

ok, this is good to know. this PR is in fact motivated by observations (+ a capture I have here),
which was taken using proton and I assume DXVK (the game is using DirectX11).

> If that condition could still happen then setting VK_QUERY_RESULT_WAIT_BIT would have the same problem our code had before https://github.com/LunarG/gfxreconstruct/pull/542

quickly tested replay of another 2 DXVK/proton titles and also could not observe anything negative there.
let me think a bit about a remaining concern:
maybe a trimmed capture 'could' produce such a condition that would hang in `vkGetQueryPoolResults`?

werman:
Ugh, I would need to dig up the notes where it happened exactly.... I hope I still have something. Will be able to do that in a few days.

fabian-lunarg:
> let me think a bit about a remaining concern:
maybe a trimmed capture 'could' produce such a condition that would hang in vkGetQueryPoolResults?

[doc](https://docs.vulkan.org/refpages/latest/refpages/source/vkGetQueryPoolResults.html#_description) says:

"If VK_QUERY_RESULT_WAIT_BIT is set, this command defines an execution dependency with any earlier commands that writes one of the identified queries ..."

in my understanding, at least that 'shouldn't' cause a blocking wait on cpu.
glad to wait a couple days, curious if/what exactly might have changed since then

werman:
Unfortunately I wasn't able to find any more info about which specific games behaved this way.

fabian-lunarg:
> Unfortunately I wasn't able to find any more info about which specific games behaved this way.

ok, no worries and many thanks for looking into it!

PR #2541, opened 2025-12-01
Capture in replay: preserve struct Handle IDs

davidd-lunarg:
Recapture should now preserve the Handle IDs of objects that are created and returned via output structs.


bradgrantham-lunarg:
What would be a test for this?

bradgrantham-lunarg:
Internal Jenkins build 8419 succeeded.

davidd-lunarg:
> What would be a test for this?

The `--capture-copy-data` feature added in https://github.com/LunarG/gfxreconstruct/pull/2543 relies on the preservation of the Handle IDs during recapture. In that feature the blocks within the trim range are copied directly from source to recaptured file. If the Handle IDs aren't correctly preserved for the trim state setup, the resulting capture will fail to replay. Internally there is a test plan for that feature, so this change should get decent coverage that way.

I also have some ideas if a more targeted test is needed. One idea is: CI does a recapture of a trimmed capture, converts the source and recaptured file to json, then parses the json to compare the Handle IDs of the objects created during trim state setup. Let me know if you'd like me to put together a more detailed test plan that targets just this feature.

PR #2542, opened 2025-12-02
Fix replay crash in wayland when clicking

charles-lunarg:
xdg_shell is the default shell (instead of wl_shell), but the logic to move the replay window when the user clicks was missing.

### DIFF
--- a/USAGE_android.md
+++ b/USAGE_android.md
@@ -19,6 +19,9 @@ to one of these other documents:
  * [GFXReconstruct for Desktop Vulkan](./USAGE_desktop_Vulkan.md)
  * [GFXReconstruct for Desktop D3D12](./USAGE_desktop_D3D12.md)
 
+Additional instructions for debugging the GFXReconstruct capture layer on Android can be found here:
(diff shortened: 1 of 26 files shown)


framework/application/wayland_context.cpp
```diff
@@ -367,10 +367,17 @@ void WaylandContext::HandlePointerButton(
...
-        WaylandWindow* window = entry->second;
         if ((button == BTN_LEFT) && (state == WL_POINTER_BUTTON_STATE_PRESSED))
         {
-            wl.shell_surface_move(window->GetShellSurface(), wayland_context->seat_, serial);
+            WaylandWindow* window = entry->second;
+            if (window->GetXdgToplevel() != nullptr)
+            {
+                wl.xdg->xdg_toplevel_move(window->GetXdgToplevel(), wayland_context->seat_, serial);
+            }
+            else if (window->GetXdgToplevel() == nullptr && window->GetShellSurface() != nullptr)
```
mikes-lunarg: nit: `window->GetXdgToplevel() == nullptr ` is redundant here because of the if above

mikes-lunarg (approved):

PR #2543, opened 2025-12-03
Add capture in replay option to copy source data

davidd-lunarg:
Adds replay option `--capture-copy-data`. This option is An extended version of `--capture` that can be used with trimming. The capture data within the trim range is copied directly from the source capture file into the new trimmed file. Note that replay must still run in order to generate the trim state block. Portable replay features are not supported. For example, replay should be done on the same device as capture, memory translation options are not supported, and if ray tracing is used, the device must support Vulkan's opaque capture and replay features.

This option currently forces on the option `swapchain --captured`, but this requirement is planned to be removed in future work.

Dependent on PR #2567

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -1333,6 +1333,22 @@ void VulkanReplayConsumerBase::WriteTrimBlockForRecapture(const ParsedBlock* par
     }
 }
 
+void VulkanReplayConsumerBase::SetOriginalMappedMemoryPointer(VkResult                         replay_result,
+                                                              VkDeviceMemory                   memory_handle,
+                                                              PointerDecoder<uint64_t, void*>* data_ptr_decoder)
+{
+    if (options_.capture_copy_data)
+    {
+        auto capture_manager = encode::VulkanCaptureManager::Get();
+        if (replay_result == VK_SUCCESS && capture_manager != nullptr)
+        {
+            void* original_data_ptr =
+                data_ptr_decoder->IsNull() ? nullptr : reinterpret_cast<void*>(*data_ptr_decoder->GetPointer());
+            capture_manager->SetOriginalMappedMemoryPointer(memory_handle, original_data_ptr);
+        }
+    }
+}
+
 void VulkanReplayConsumerBase::SetupForRecaptureInReplay(PFN_vkGetInstanceProcAddr get_instance_proc_addr,
                                                          PFN_vkCreateInstance      create_instance,
                                                          PFN_vkCreateDevice        create_device,
@@ -5596,14 +5612,14 @@ VkResult VulkanReplayConsumerBase::OverrideAllocateMemory(
     return result;
 }
 
-VkResult VulkanReplayConsumerBase::OverrideMapMemory(PFN_vkMapMemory         func,
-                                                     VkResult                original_result,
-                                                     const VulkanDeviceInfo* device_info,
-                                                     VulkanDeviceMemoryInfo* memory_info,
-                                                     VkDeviceSize            offset,
-                                                     VkDeviceSize            size,
-                                                     VkMemoryMapFlags        flags,
-                                                     void**                  ppData)
+VkResult VulkanReplayConsumerBase::OverrideMapMemory(PFN_vkMapMemory                  func,
+                                                     VkResult                         original_result,
+                                                     const VulkanDeviceInfo*          device_info,
+                                                     VulkanDeviceMemoryInfo*          memory_info,
+                                                     VkDeviceSize                     offset,
+                                                     VkDeviceSize                     size,
+                                                     VkMemoryMapFlags                 flags,
+                                                     PointerDecoder<uint64_t, void*>* ppData)
 {
     GFXRECON_UNREFERENCED_PARAMETER(func);
     GFXRECON_UNREFERENCED_PARAMETER(original_result);
@@ -5613,7 +5629,12 @@ VkResult VulkanReplayConsumerBase::OverrideMapMemory(PFN_vkMapMemory         fun
     auto allocator = device_info->allocator.get();
     assert(allocator != nullptr);
 
-    return allocator->MapMemory(memory_info->handle, offset, size, flags, ppData, memory_info->allocator_data);
+    auto replay_result = allocator->MapMemory(
+        memory_info->handle, offset, size, flags, ppData->GetOutputPointer(), memory_info->allocator_data);
+
+    SetOriginalMappedMemoryPointer(replay_result, memory_info->handle, ppData);
+
+    return replay_result;
 }
 
 void VulkanReplayConsumerBase::OverrideUnmapMemory(PFN_vkUnmapMemory       func,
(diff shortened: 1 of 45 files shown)


framework/generated/khronos_generators/khronos_replay_consumer_body_generator.py
```diff
@@ -818,6 +821,8 @@ def make_body_expression(self, api_data, return_type, name, values, is_override)
                                 )
                                 # If this is a struct with handles, we need to add replay mappings for the embedded handles.
                                 if value.base_type in self.structs_with_handles:
+                                    push_handleid_expr[0] = "    PushRecaptureStructArrayHandleIds({paramname}->GetMetaStructPointer(), {paramname}->GetLength(), this);\n".format(paramname=value.name)
```
MarkY-LunarG, line 859: Shouldn't we check if recapture is enabled before we try recapturing IDs?  Seems like a lot of extra work walking every structures members and then clearing values after for something that might not be needed.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -349,12 +349,18 @@ void VulkanReplayConsumerBase::ProcessStateBeginMarker(uint64_t frame_number)
...
-    GFXRECON_UNREFERENCED_PARAMETER(frame_number);
     loading_trim_state_ = false;
     if (fps_info_ != nullptr)
     {
         fps_info_->ProcessStateEndMarker(frame_number);
     }
+
+    if (options_.capture_copy_data)
+    {
+        auto vulkan_capture_manager = encode::VulkanCaptureManager::Get();
```
MarkY-LunarG, line 386: I think we should have some kind of indication that recapture trimming is started.  Log message (at least info)?

framework/encode/api_capture_manager.h
```diff
@@ -108,7 +109,11 @@ class ApiCaptureManager
 
     void EndFrame(CommonCaptureManager::ApiSharedLockT& current_lock)
     {
-        common_manager_->EndFrame(api_family_, current_lock);
+        // When doing recapture in replay with copy data, replay is responsible for calling EndFrameForRecapture.
+        if (!CommonCaptureManager::IsRecaptureCopyData())
+        {
+            common_manager_->EndFrame(api_family_, current_lock);
+        }
```
MarkY-LunarG, line 116: Regarding the comment, why exactly is this required?

framework/encode/capture_manager.cpp
```diff
@@ -142,7 +153,8 @@ bool CommonCaptureManager::LockedCreateInstance(ApiCaptureManager*           api
             GFXRECON_LOG_WARNING("Failed registering atexit");
         }
 
-        if (initialize_log_)
+        // Don't initialize logger when doing recapture. It is already initialized by replay.
+        if (!IsRecaptureInReplay())
```
MarkY-LunarG, line 182: Why not just set `initialize_log_` to false if `IsRecaptureInReplay()`?

framework/encode/capture_manager.cpp
```diff
@@ -1322,7 +1349,11 @@ void CommonCaptureManager::ActivateTrimming(std::shared_lock<ApiCallMutexT>& cur
             exclusive_api_call_lock = AcquireExclusiveApiCallLock();
         }
 
-        capture_mode_ |= kModeWrite;
+        capture_mode_ |= kModeTrim;
```
MarkY-LunarG, line 1404: This is new.  Should it be an assert verifying the value was already set?  Otherwise we are probably here for a wrong reason, right?

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -119,12 +124,17 @@ void VulkanCaptureManager::WriteTrackedStateWithAssetFile(util::FileOutputStream
                                                           util::FileOutputStream* asset_file_stream,
                                                           const std::string*      asset_file_name)
 {
+    uint32_t trim_frame = GetCurrentFrame();
+    if (CommonCaptureManager::IsRecaptureCopyData())
+    {
+        trim_frame += (GetRecaptureTrimStartFrame() - 1);
```
MarkY-LunarG, line 130: I don't understand why you're doing this?  Is this because you're going to need this state in the future at that frame?  How do you know?

framework/generated/khronos_generators/khronos_replay_consumer_body_generator.py
davidd-lunarg: Ultimately this calls `VulkanReplayConsumerBase::PushRecaptureHandleId` which does that check. I agree it would be better to also do the check further up the call stack. I avoided adding the check in the replay consumer base generator to avoid adding noise to those generated files. Instead I'll modify `PushRecaptureStructArrayHandleIds` in `common_struct_handle_mappers.h` to perform the check before doing any work. 

This change is part of PR #2541 which this PR builds on, so I will add the changes to `PushRecaptureStructArrayHandleIds` in that PR.

framework/encode/api_capture_manager.h
davidd-lunarg: This makes the frame markers in source capture file the "ground truth" for frame counting during recapture. Otherwise the capture manager will do its own frame accounting which could potentially change leading to different frame counts between the replay consumer/file processor and the capture manager.

And for a practical reason, it is done here to ensure trimming is activated by the end frame marker and not a vkQueuePresent call. If trimming is activated in the vkQueuePresent call, the subsequent end frame marker gets copied to the output capture file in `VulkanReplayConsumerBase::WriteTrimBlockForRecapture` when it shouldn't.

framework/encode/capture_manager.cpp
davidd-lunarg: `kModeTrim` is now added to capture_mode_ between Activate and Deactivate trimming, i.e., when trimming is "active". This decouples that state from `kModeWrite` which used to always be enabled between Activate and Deactivate trimming but now is not enabled when using `--capture-copy-data`. The state is then checked in `VulkanReplayConsumerBase::WriteTrimBlockForRecapture` to determine whether to pass through the current block data to the output capture file.

framework/encode/vulkan_capture_manager.cpp
davidd-lunarg: `GetCurrentFrame()` is the number of frames processed by the capture manager--the number of times `CaptureManager::EndFrame()` is called (plus 1). When running `gfxrecon-replay --capture-copy-data` using a trimmed source file, this won't correspond to the end frame markers in the source file which get copied directly to the output file. Adding the trim start frame here results in the output file's trim state block StateMarkers containing the correct frame number.

framework/encode/capture_manager.cpp
davidd-lunarg: `initialize_log_` was added to fix logging for recapture in replay in PR #2334. I replaced it with `IsRecaptureInReplay()` here because:
1. `IsRecaptureInReplay()` is useful in other places in the capture manager.
2. There is no other current use case for `initialize_log_`, and I don't know that there will be in the future.

framework/encode/vulkan_capture_manager.cpp
davidd-lunarg: As an example, if the source capture was trimmed to frames 10-20, and `gfxrecon-replay --capture-copy-data` is run with `GFXRECON_CAPTURE_FRAMES=2`, the output trim state block should indicate that it is for frame 11, not for frame 2.

davidd-lunarg:
Returning to draft status until PRs #2541 and #2567 are merged.

PR #2544, opened 2025-12-06
Modify Preload to parse before enqueing

jzulauf-lunarg:
Eliminates last duplicate block parser.

### DIFF
--- a/framework/decode/block_parser.h
+++ b/framework/decode/block_parser.h
@@ -79,17 +79,27 @@ class BlockParser
     {
         kAlways = 0,     // Always decompress parameter data when parsing blocks
         kNever,          // Never decompress parameter data when parsing blocks
-        kQueueOptimized, // Decompress only "small" blocks deferring large block to JIT decompression, suitable for
+        kQueueOptimized, // Decompress only "small" blocks, deferring large blocks to JIT decompression, suitable for
                          // block preload/preparsed content
     };
 
-    // The threshold is based on a trade off, maximize the number blocks with no decompression at replay
-    // but minimizing the memory overhead of the non-deferred compressions.
+    // Threshold data came from a histogram of numerous large traces and is chosen to
+    // reduce the number of deferred decompressions at a reasonable marginal memory cost.
+    // Memory "Penalty" or "Savings" below are based on the size of enqueued ParsedBlocks for replay.
     //
-    // Threshhold data came from a histogram of numerous large traces and is chosen as a trade-off
-    // that allows the majority of blocks to be decompressed when parsed and enqueued (for preloaded replay), while
-    // the 3/4 of the *bytes* are not decompressed when enqueued, limiting the memory impact
-    static constexpr size_t kSmallThreshold = 96 + sizeof(format::BlockHeader); // 83% of blocks, 26% of bytes
+    // For a threshold of 512 bytes:
+    // - Penalty of 74% memory vs. kNever (all deferred), performing > 99% of decompression operations at parse time.
+    // - Savings of 42% memory vs. kAlways (none deferred), requiring only < 1% decompressions at replay time.
+    //
+    // Another choice would be 256: offers 2% lower memory penalty, but doubles replay decompressions vs. 512.
+    // - Penalty of 72% memory vs. kNever, performing > 98% of decompression operations at parse time.
+    // - Savings of 43% memory vs. kAlways, requiring < 2% decompressions at replay time
+
+    // NOTE: The threshold computation was based on assumptions of compression ratios that need to validated with
+    //       additional instrumentation.
+
+    // Minimizes decompression and allocator activity during replay from queue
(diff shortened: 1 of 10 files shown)


framework/decode/block_parser.h
```diff
@@ -79,17 +79,27 @@ class BlockParser
...
+    // - Penalty of 74% memory vs. kNever (all deferred), performing > 99% of decompression operations at parse time.
+    // - Savings of 42% memory vs. kAlways (none deferred), requiring only < 1% decompressions at replay time.
+    //
+    // Another choice would be 256: offers 2% lower memory penalty, but doubles replay decompressions vs. 512.
+    // - Penalty of 72% memory vs. kNever, performing > 98% of decompression operations at parse time.
+    // - Savings of 43% memory vs. kAlways, requiring < 2% decompressions at replay time
+
+    // NOTE: The threshold computation was based on assumptions of compression ratios that need to validated with
+    //       additional instrumentation.
+
```
MarkY-LunarG, line 100: This almost seems to scream "option" to me.  Maybe make it a TODO to have this be a little configurable (like allowing a "SetDeferThreshold")?

framework/decode/file_processor.cpp
```diff
@@ -327,30 +335,34 @@ bool FileProcessor::ProcessBlocks()
...
+        else // ContinueDecoding returned false
+        {
+            process_state = ProcessBlockState::kEndProcessing;
+        }
     }
 
-    DecrementRemainingCommands();
-    return success;
+    // Update the frame number etc.
+    if (process_state == ProcessBlockState::kFrameBoundary)
```
MarkY-LunarG, line 360: Won't this mess up frame counting now by moving it out of the loop?  For example, won't line 294 be incorrect, because it will continue to process frames because kFrameBoundary > 0?

framework/decode/file_processor.h
```diff
@@ -364,6 +388,11 @@ class FileProcessor
...
         bool IsFrameDelimiter() const { return is_frame_delimiter; }
         ProcessVisitor(FileProcessor& file_processor) : file_processor_(file_processor) {}
+        void Reset()
+        {
+            is_frame_delimiter = false;
+            success            = true;
+        }
 
       private:
         bool           is_frame_delimiter = false;
```
MarkY-LunarG, line 398: Just  noticed these are class  members without the trailing underscores.

framework/decode/parsed_block.cpp
```diff
@@ -90,6 +90,16 @@ bool ParsedBlock::Decompress(BlockParser& parser)
     return IsReady();
 }
 
+void ParsedBlock::TrimBlock(BlockParser& parser)
+{
+    if (parser.ShouldTrim(*this))
```
MarkY-LunarG, line 95: We don't check the blocks state before we "trim", should we?

framework/decode/preload_file_processor.cpp
```diff
@@ -31,88 +31,131 @@ PreloadFileProcessor::PreloadFileProcessor() {}
...
-                    HandleBlockReadError(kErrorReadingBlockData, msg.c_str());
-                }
+                // Deal with the frame marker after implied frame kFunctionCallBlock frame boundary case
+                // Append the blocks leading up to the frame marker to the previous frame
+                GFXRECON_ASSERT(current_frame_number_ == (kFirstFrame + 1));
+                GFXRECON_ASSERT(!preload_frames_.empty());
+                auto& prev_frame = preload_frames_.back().blocks;
+                prev_frame.insert(prev_frame.end(),
+                                  std::make_move_iterator(pending_parsed_blocks_.begin()),
+                                  std::make_move_iterator(pending_parsed_blocks_.end()));
```
MarkY-LunarG, line 54: I don't understand where we exit this when we've completed the frame.  Are we going to keep reading to reach "count" even if it's past the end of the desired frame?

framework/decode/preload_file_processor.cpp
```diff
@@ -31,88 +31,131 @@ PreloadFileProcessor::PreloadFileProcessor() {}
...
+    DispatchFunction dispatch = [this](uint64_t block_index, ParsedBlock& block) {
+        pending_parsed_blocks_.emplace_back(block_index, std::move(block));
+        return ProcessBlockState::kRunning;
+    };
+
+    ProcessBlockState process_result = ProcessBlocks(dispatch, false /* check decoder completion */);
+    return ContinueProcessing(process_result);
+}
+
+void PreloadFileProcessor::CleanupReplay()
```
MarkY-LunarG, line 85: The naming is confusing to me.  Instead maybe call them "consumed_frames" instead of "replayed_frames" and supporting function changes as appropriate.  Replayed frames might be re-looped, so I think consumed is clearer.

framework/decode/block_parser.h
jzulauf-lunarg: Doing performance profiling on this... the impact of various threshhold values is looking to be trace specific, so, I can see that.

PR #2546, opened 2025-12-09
Add unit-test-cases for SPIRV-parsing

fabian-lunarg:
- testing buffer-device-addresses detection
- minor cleanup/refactor in parser


framework/util/spirv_parsing_util.cpp
```diff
@@ -242,6 +242,9 @@ bool SpirVParsingUtil::ParseBufferReferences(const uint32_t* const spirv_code, s
     }
 
     // forward spirv-reflect-pass
+    constexpr bool use_forward_spirv_reflect_pass = true;
```
fabian-lunarg, line 245: using this `constexpr` hook we can demonstrate the `spirv-reflect` based pass
is currently required to pass the tests.

PR #2548, opened 2025-12-10
Fix for unassisted

panos-lunarg:
On vkUnmapMemory memory wrapper's information was zeroed out before reaching the unassisted logic. This caused artifacts when replaying captures that were captured using the unassisted memory tracking method. This commit fixes the issue


ncesario-lunarg (approved):
LGTM and I verified locally that this resolves the issues I was seeing.

MarkY-LunarG (approved):

PR #2549, opened 2025-12-10
Fix a performance-issue in OverrideUpdateDescriptorSets

fabian-lunarg:
- pull `VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK`-case out of an inner loop
- minor: slightly reduce number of std::map lookups
- affected options: -m rebind and --dump-resources


MarkY-LunarG:
What's the perf improvement after this change?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -11230,12 +11247,10 @@ void VulkanReplayConsumerBase::OverrideUpdateDescriptorSets(
...
                     case VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER:
                     {
-                        dst_desc_set_info->descriptors[binding].image_info[arr_idx].image_layout =
-                            in_pDescriptorWrites[s].pImageInfo[i].imageLayout;
-
-                        dst_desc_set_info->descriptors[binding].image_info[arr_idx].image_view_info =
-                            object_info_table_->GetVkImageViewInfo(
-                                writes_meta[s].pImageInfo->GetMetaStructPointer()[i].imageView);
+                        auto& desc_image_info           = descriptor_set_binding_info.image_info[arr_idx];
+                        desc_image_info.image_layout    = in_pDescriptorWrites[s].pImageInfo[i].imageLayout;
```
MarkY-LunarG: Should we define a variable for `in_pDescriptorWrites[s]` as well out of this loop?  Or will the compiler be smart enough to handle it?  We seem to use it here and in the uniform buffer case below several times.  Also `writes_meta[s]` which could be moved up as well.

fabian-lunarg: agree, let's get rid of those as well, should also be more readable

fabian-lunarg:
> I'm not going to approve until I see actual numbers. I've seen too many optimistic improvements that actually hurt perf. Once those numbers are in, I think I can easily approve.

I'll look into measuring with the few inline-uniform captures I got here. the prime-scenario benefiting from this change are inline-uniform-buffers. the larger the more imrpovement. the PR is fixing a bug that would issue a `memcpy` for the entire range, in the inner-loop (-> O(n^2 instead of O(n))). 
so, 'should' be able to measure that if noise isn't too high.

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: done. `in_pDescriptorWrites[s]` was actually already defined as just `write`

fabian-lunarg:
got some numbers @MarkY-LunarG :)

I measured the relevant branch inside `if (UseExtraDescriptorInfo(device_info)){...}` using std::chrono probes.
this code-path is used by dump-resources and -m rebind.

the capture I was using only has some few descriptor-sets using very small inline-uniform buffers (which was optimized here),
so the average performance-gain I can show is oversee-able, but 'could' be much larger, depending on scenario.

averaged, **summed / total time** spent in above if-block:
without the fix: **32ms** vs. with fix: **16.8ms**

while this prob. doesn't blow your mind, it's fair to claim there is no performance-regression

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: `writes_meta[s]` also moved up to a variable.

MarkY-LunarG (approved):

PR #2551, opened 2025-12-10
ShaderID on Different StateObjects

JerryAMD:
GFXR query ShaderID on StateObject during replay and store the replay ShaderID to StateObject extra info struture. When replay read Shader Table and try to map the ShaderID to replay value, it use current StateObject to map value. If one Shader was referenced by two different StateObjects, the two SOs are inheritance, the RX9070 driver return different ShaderID on the two SOs. The application might use any of ShaderID to the DispatchRays Shader Table.

The solution is
For D3D12_STATE_SUBOBJECT_TYPE_EXISTING_COLLECTION inheritance, GFXR should port existing object's ShaderID map to new SO.

For AddToStateObject(), GFXR should port ShaderID map from grow object over to new SO.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -10469,18 +10469,37 @@ void VulkanReplayConsumerBase::OverrideFrameBoundaryANDROID(PFN_vkFrameBoundaryA
         util::EndInjectedCommands();
     }
 
-    CommonObjectInfoTable& object_info_table = GetObjectInfoTable();
+    // TODO: merged in PR #2623 but causing issues in extended CI -> debug remaining issues and enable this again
+    // related issue: https://github.com/LunarG/gfxreconstruct/issues/2660
+    if constexpr (false)
+    {
+        CommonObjectInfoTable& object_info_table = GetObjectInfoTable();
 
-    VulkanPhysicalDeviceInfo* physical_device_info = object_info_table.GetVkPhysicalDeviceInfo(device_info->parent_id);
-    GFXRECON_ASSERT(physical_device_info != nullptr);
+        VulkanPhysicalDeviceInfo* physical_device_info =
+            object_info_table.GetVkPhysicalDeviceInfo(device_info->parent_id);
+        GFXRECON_ASSERT(physical_device_info != nullptr);
 
-    VulkanInstanceInfo* instance_info = object_info_table.GetVkInstanceInfo(physical_device_info->parent_id);
-    GFXRECON_ASSERT(instance_info != nullptr);
+        VulkanInstanceInfo* instance_info = object_info_table.GetVkInstanceInfo(physical_device_info->parent_id);
+        GFXRECON_ASSERT(instance_info != nullptr);
 
-    const graphics::VulkanInstanceTable* instance_table = GetInstanceTable(instance_info->handle);
+        const graphics::VulkanInstanceTable* instance_table = GetInstanceTable(instance_info->handle);
 
-    swapchain_->FrameBoundaryANDROID(
-        func, device_info, semaphore_info, image_info, instance_info, instance_table, device_table, application_.get());
+        swapchain_->FrameBoundaryANDROID(func,
+                                         device_info,
+                                         semaphore_info,
+                                         image_info,
+                                         instance_info,
+                                         instance_table,
+                                         device_table,
+                                         application_.get());
+    }
+    else
+    {
+        VkDevice    device    = device_info->handle;
+        VkSemaphore semaphore = semaphore_info ? semaphore_info->handle : VK_NULL_HANDLE;
+        VkImage     image     = image_info ? image_info->handle : VK_NULL_HANDLE;
+        func(device, semaphore, image);
+    }
 }
 
 // We want to allow skipping the query for tool properties because the capture layer actually adds this extension


framework/decode/dx12_object_info.h
```diff
@@ -173,6 +173,15 @@ struct ResourceValueInfo
...
     bool operator<(const ResourceValueInfo& other) const { return offset < other.offset; }
+
+    bool operator==(const ResourceValueInfo& other) const
+    {
+        return (offset == other.offset) && (type == other.type) && (size == other.size) &&
+               (state_object == other.state_object) &&
+               (arg_buffer_extra_info.argument_buffer == other.arg_buffer_extra_info.argument_buffer) &&
+               (arg_buffer_extra_info.argument_buffer_offset == other.arg_buffer_extra_info.argument_buffer_offset) &&
+               (max_command_count == other.max_command_count);
+    }
```
davidd-lunarg: This compares all the members of `ResourceValueInfo` except for:
`(arg_buffer_extra_info.command_signature_info == other.arg_buffer_extra_info.command_signature_info)`.

Is it necessary to skip that equality check? Or could the default `operator==` be used to compare all members--and any future members that might get added?
`bool operator==(const ResourceValueInfo& other) const = default;`*

*This also requires adding the default `operator==` to `ArgumentBufferExtraInfo`:
`bool operator==(const ArgumentBufferExtraInfo& other) const = default;`

framework/decode/dx12_resource_value_mapper.cpp
```diff
@@ -645,15 +645,26 @@ void Dx12ResourceValueMapper::PostProcessCreateRootSignature(PointerDecoder<uint
 void Dx12ResourceValueMapper::PostProcessCreateStateObject(
     HandlePointerDecoder<void*>*                           state_object_decoder,
     StructPointerDecoder<Decoded_D3D12_STATE_OBJECT_DESC>* desc_decoder,
-    const std::map<std::wstring, format::HandleId>&        in_lrs_associations_map)
+    const D3D12StateObjectInfo*                            grow_from_state_object)
```
davidd-lunarg, line 648: Does the signature to `Dx12ResourceValueMapper::PostProcessCreateStateObject` need to change? If I've read it correctly, the behavior is the same as before. In both cases:
1. `lrs_associations_map = grow_from_state_object->export_name_lrs_map` when `Dx12ResourceValueMapper::PostProcessCreateStateObject` is called from `Dx12ReplayConsumerBase::OverrideAddToStateObject` and
2. `lrs_associations_map = {}` when `Dx12ResourceValueMapper::PostProcessCreateStateObject` is called from `Dx12ReplayConsumerBase::OverrideCreateStateObject`.

`grow_from_state_object` isn't used anywhere else in `Dx12ResourceValueMapper::PostProcessCreateStateObject`

framework/decode/dx12_resource_value_mapper.cpp
```diff
@@ -662,7 +673,8 @@ void Dx12ResourceValueMapper::PostProcessCreateStateObject(
                                      explicit_default_local_root_signature_id,
                                      explicit_local_root_signature_associations,
                                      hit_group_imports,
-                                     lrs_associations_map);
+                                     lrs_associations_map,
+                                     existing_shader_id_lrs_map);
```
davidd-lunarg, line 674: Is `existing_shader_id_lrs_map` used? In gets passed by reference into `GetStateObjectLrsAssociationInfo` where it gets assigned but it doesn't appear to be read anywhere.

framework/decode/dx12_object_info.h
JerryAMD: Actually, I need the override operator== for error check. The struct RsourceValueInfo default operator== could not check entire struct's equality. I'll add operator== to ArgumentbufferExtraInfo.

framework/decode/dx12_resource_value_mapper.cpp
JerryAMD: It is used for AddToStateObject() API. Given a exported name, it could get different ShaderID on the two SOs, if one of SO was build on top of older one.
This grow_from_state_object, allow the new SO inherit older SO's export names and ShaderIDs.
It's same idea to Dx12ResourceValueMapper::GetStateObjectLrsAssociationInfo() new argument existing_shader_id_lrs_map.

framework/decode/dx12_resource_value_mapper.cpp
JerryAMD: Yes. It was used by subobject_type == D3D12_STATE_SUBOBJECT_TYPE_EXISTING_COLLECTION. And it state_object_extra_info->shader_id_lrs_map.merge(existing_shader_id_lrs_map); at ln#758( I made mistake to comment it out. Will revert it).

framework/decode/dx12_resource_value_mapper.cpp
```diff
@@ -645,15 +645,26 @@ void Dx12ResourceValueMapper::PostProcessCreateRootSignature(PointerDecoder<uint
...
+    std::map<graphics::Dx12ShaderIdentifier, std::set<ResourceValueInfo>> existing_shader_id_lrs_map;
+
+    auto state_object_extra_info = GetExtraInfo<D3D12StateObjectInfo>(state_object_decoder);
+    if (grow_from_state_object != nullptr)
+    {
+        lrs_associations_map = grow_from_state_object->export_name_lrs_map;
+        /*
+        state_object_extra_info->shader_id_lrs_map.insert(grow_from_state_object->shader_id_lrs_map.begin(),
+                                                          grow_from_state_object->shader_id_lrs_map.end());
+                                                          */
```
davidd-lunarg: Remove unused, commented code.

framework/decode/dx12_object_info.h
davidd-lunarg: > The struct RsourceValueInfo default operator== could not check entire struct's equality.

Adding

`bool operator==(const ResourceValueInfo& other) const = default;` to `ResourceValueInfo` and 
`bool operator==(const ArgumentBufferExtraInfo& other) const = default;` to `ArgumentBufferExtraInfo`

creates an `operator==` that calls `operator==` on all member variables of `ResourceValueInfo` (and in turn all member variables of `ArgumentBufferExtraInfo`). If I understand correctly that is the desired behavior here. Using the default declaration ensures that if new member variables are added to `ResourceValueInfo` in the future they will be included in the equality check automatically.

Reference: https://en.cppreference.com/w/cpp/language/default_comparisons.html, section `Equality comparison`.

PR #2552, opened 2025-12-10
Add "exiting android_main" message

vaage:
Adds a message at the end of android_main to make it easier for tools, waiting for replay to end, to know when replay has finished. This only affects replay on Android; it does not affect other platforms.


tools/replay/android_main.cpp
```diff
@@ -306,6 +306,9 @@ void android_main(struct android_app* app)
     gfxrecon::util::Log::Release();
 
     gfxrecon::util::DestroyActivity(app);
+
+    GFXRECON_WRITE_CONSOLE("====== Exiting android_main");
```
MarkY-LunarG: This won't properly work.  We just released the log on line 306 so this may re-create items that should be freed.

MarkY-LunarG (changes requested):

tools/replay/android_main.cpp
vaage: We were trying to figure out where to put it. We were trying to mirror line `105` in `android_main.cpp`, these are the lines that confused us:

```cpp
GFXRECON_WRITE_CONSOLE("====== Entering android_main");
gfxrecon::util::Log::Init();
```

Because `"====== Entering android_main"` came before `gfxrecon::util::Log::Init()` we thought the other end would come after `Release`.

Should we just move it before `Release`, or is it more complicated than that?

MarkY-LunarG: Yes, it should be before the release.  I think that would be good enough.  There is still some cleanup, but I think it's mostly trivial.  Otherwise, you would have to manually add it as a logcat message if you kept it at the very end, but I think that would be too messy.

bradgrantham-lunarg: > We were trying to figure out where to put it. We were trying to mirror line `105` in `android_main.cpp` [...]
> Because `"====== Entering android_main"` came before `gfxrecon::util::Log::Init()` we thought the other end would come after `Release`.

I'm confused by that, too.  @MarkY-LunarG any idea whether that WRITE_CONSOLE before `Log::Init` is expected to work or do we just get lucky?  That WRITE_CONSOLE wouldn't obey the log settings?

MarkY-LunarG: Oh, thanks @bradgrantham-lunarg , I missed that.

It works because of defaults inside of the log class, because before you call Init, Log::Settings is still set to valid defaults that allow the GFXRECON_WRITE_CONSOLE to pass through the function and trigger the Android log message (or the "output_to_stdout" behavior).  This is really also wrong and we should fix that as well, or we should intentionally have a fallback for GFXRECON_WRITE_CONSOLE to avoid any checks if before/after logging init to just write out to stdout/logcat.   But for me, I see this as a bug.

bradgrantham-lunarg: @vaage implemented your request, going to resolve.

PR #2557, opened 2025-12-12
add generation of custom chainable Khronos structs

bradgrantham-lunarg:


MarkY-LunarG:
Yikes!  Should we also file a spec bug?

PR #2560, opened 2025-12-14
Build `project_version` as a static library

mattst88:
Fixes the build when configuring with `-DBUILD_SHARED_LIBS=ON`, which some Linux distributions have enabled by default.


mattst88:
Can we land this?

PR #2561, opened 2025-12-15
VDR: Transfer commands

panos-lunarg:
VDR: Transfer commands
It is now possible to dump resources targeted by transfer API calls. The
following transfer commands are added:
vkCmdCopyBuffer
vkCmdCopyBuffer2
vkCmdCopyBuffer2KHR
vkCmdCopyBufferToImage
vkCmdCopyBufferToImage2
vkCmdCopyBufferToImage2KHR
vkCmdCopyImage
vkCmdCopyImage2
vkCmdCopyImage2KHR
vkCmdCopyImageToBuffer
vkCmdCopyImageToBuffer2
vkCmdCopyImageToBuffer2KHR
vkCmdBlitImage
vkCmdBlitImage2
vkCmdBlitImage2KHR
vkCmdBuildAccelerationStructuresKHR
vkCmdCopyAccelerationStructureKHR

Additionaly the following GFXR transfer meta commands are also
supported:
kInitBufferCommand
kInitImageCommand
kVulkanBuildAccelerationStructuresCommand
kVulkanCopyAccelerationStructuresCommand

More details can be found in the vulkan_dump_resources.md documentation

### DIFF
--- a/framework/decode/vulkan_replay_dump_resources_delegate.cpp
+++ b/framework/decode/vulkan_replay_dump_resources_delegate.cpp
@@ -23,6 +23,7 @@
 #include "decode/vulkan_object_info.h"
 #include "decode/vulkan_replay_dump_resources_delegate.h"
 #include "decode/vulkan_replay_dump_resources_common.h"
+#include "decode/vulkan_replay_dump_resources_delegate_dumped_resources.h"
 #include "generated/generated_vulkan_enum_to_string.h"
 #include "util/buffer_writer.h"
 #include "util/image_writer.h"
@@ -2199,6 +2200,17 @@ bool DefaultVulkanDumpResourcesDelegate::DumpTransferCommandToFile(
     return true;
 }
 
+static void GenerateOutputJsonTransferImage(nlohmann::ordered_json&    json_entry,
+                                            const TransferedImageInfo& transf_img_info)
+{
+    json_entry["image"]     = transf_img_info.id;
+    json_entry["format"]    = util::ToString<VkFormat>(transf_img_info.format);
+    json_entry["extent"][0] = transf_img_info.extent.width;
+    json_entry["extent"][1] = transf_img_info.extent.height;
+    json_entry["extent"][2] = transf_img_info.extent.depth;
+    json_entry["layout"]    = util::ToString<VkImageLayout>(transf_img_info.layout);
+}
+
 void DefaultVulkanDumpResourcesDelegate::GenerateOutputJsonInitBufferCommand(const DumpedTransferCommand& cmd,
                                                                              nlohmann::ordered_json&      json_entry)
 {
@@ -2219,7 +2231,8 @@ void DefaultVulkanDumpResourcesDelegate::GenerateOutputJsonInitImageCommand(cons
     const auto* init_image = std::get_if<DumpedInitImageMetaCommand>(&cmd.dumped_resource);
     GFXRECON_ASSERT(init_image != nullptr);
 
-    json_entry["image"] = init_image->image;
+    auto& dst_image_json_entry = json_entry["dstImage"];
+    GenerateOutputJsonTransferImage(dst_image_json_entry, init_image->image);
 
     for (size_t sr = 0; sr < init_image->dumped_image.dumped_subresources.size(); ++sr)
     {
@@ -2272,9 +2285,10 @@ void DefaultVulkanDumpResourcesDelegate::GenerateOutputJsonCopyBufferToImageComm
     const auto* copy_buffer_to_image = std::get_if<DumpedCopyBufferToImage>(&cmd.dumped_resource);
     GFXRECON_ASSERT(copy_buffer_to_image != nullptr);
 
-    json_entry["srcBuffer"]      = copy_buffer_to_image->src_buffer;
-    json_entry["dstImage"]       = copy_buffer_to_image->dst_image;
-    json_entry["dstImageLayout"] = util::ToString<VkImageLayout>(copy_buffer_to_image->dst_image_layout);
+    json_entry["srcBuffer"] = copy_buffer_to_image->src_buffer;
+
+    auto& dst_image_json_entry = json_entry["dstImage"];
+    GenerateOutputJsonTransferImage(dst_image_json_entry, copy_buffer_to_image->dst_image);
 
     auto& regions_entries = json_entry["regions"];
     for (size_t i = 0; i < copy_buffer_to_image->regions.size(); ++i)
@@ -2341,10 +2355,11 @@ void DefaultVulkanDumpResourcesDelegate::GenerateOutputJsonCopyImageCommand(cons
     const auto* copy_image = std::get_if<DumpedCopyImage>(&cmd.dumped_resource);
     GFXRECON_ASSERT(copy_image != nullptr);
 
-    json_entry["srcImage"]       = copy_image->src_image;
-    json_entry["srcImageLayout"] = util::ToString<VkImageLayout>(copy_image->src_image_layout);
-    json_entry["dstImage"]       = copy_image->dst_image;
-    json_entry["dstImageLayout"] = util::ToString<VkImageLayout>(copy_image->dst_image_layout);
+    auto& src_image_json_entry = json_entry["srcImage"];
+    GenerateOutputJsonTransferImage(src_image_json_entry, copy_image->src_image);
+
+    auto& dst_image_json_entry = json_entry["dstImage"];
+    GenerateOutputJsonTransferImage(dst_image_json_entry, copy_image->dst_image);
 
     auto& regions_entries = json_entry["regions"];
     for (size_t i = 0; i < copy_image->regions.size(); ++i)
@@ -2414,9 +2429,10 @@ void DefaultVulkanDumpResourcesDelegate::GenerateOutputJsonCopyImageToBufferComm
     const auto* copy_image_to_buffer = std::get_if<DumpedCopyImageToBuffer>(&cmd.dumped_resource);
     GFXRECON_ASSERT(copy_image_to_buffer != nullptr);
 
-    json_entry["srcImage"]       = copy_image_to_buffer->src_image;
-    json_entry["srcImageLayout"] = util::ToString<VkImageLayout>(copy_image_to_buffer->src_image_layout);
-    json_entry["dstBuffer"]      = copy_image_to_buffer->dst_buffer;
+    auto& src_image_json_entry = json_entry["srcImage"];
+    GenerateOutputJsonTransferImage(src_image_json_entry, copy_image_to_buffer->src_image);
+
+    json_entry["dstBuffer"] = copy_image_to_buffer->dst_buffer;
 
     auto& regions_entries = json_entry["regions"];
     for (size_t i = 0; i < copy_image_to_buffer->regions.size(); ++i)
@@ -2463,11 +2479,13 @@ void DefaultVulkanDumpResourcesDelegate::GenerateOutputJsonBlitImageCommand(cons
     const auto* blit_image = std::get_if<DumpedBlitImage>(&cmd.dumped_resource);
     GFXRECON_ASSERT(blit_image != nullptr);
 
-    json_entry["srcImage"]       = blit_image->src_image;
-    json_entry["srcImageLayout"] = util::ToString<VkImageLayout>(blit_image->src_image_layout);
-    json_entry["dstImage"]       = blit_image->dst_image;
-    json_entry["dstImageLayout"] = util::ToString<VkImageLayout>(blit_image->dst_image_layout);
-    json_entry["filter"]         = util::ToString<VkFilter>(blit_image->filter);
+    auto& src_image_json_entry = json_entry["srcImage"];
+    GenerateOutputJsonTransferImage(src_image_json_entry, blit_image->src_image);
+
+    auto& dst_image_json_entry = json_entry["dstImage"];
+    GenerateOutputJsonTransferImage(dst_image_json_entry, blit_image->dst_image);
+
+    json_entry["filter"] = util::ToString<VkFilter>(blit_image->filter);
 
     auto& regions_entries = json_entry["regions"];
     for (size_t i = 0; i < blit_image->regions.size(); ++i)
--- a/framework/decode/vulkan_replay_dump_resources_transfer.h
+++ b/framework/decode/vulkan_replay_dump_resources_transfer.h
@@ -27,6 +27,7 @@
 #include "decode/common_object_info_table.h"
 #include "decode/vulkan_object_info.h"
 #include "decode/vulkan_replay_dump_resources_common.h"
+#include "decode/vulkan_replay_dump_resources_delegate_dumped_resources.h"
 #include "decode/vulkan_replay_options.h"
 #include "format/format.h"
 #include "generated/generated_vulkan_dispatch_table.h"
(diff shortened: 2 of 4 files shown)


framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -479,6 +500,102 @@ VulkanReplayDumpResourcesBase::FindDispatchRaysCommandBufferContext(VkCommandBuf
     return contexts;
 }
 
+std::vector<TransferDumpingContext*> VulkanReplayDumpResourcesBase::FindTransferContextCmdIndex(uint64_t cmd_index)
```
bradgrantham-lunarg: These Find functions return an array of pointers that point to elements inside map key-value pairs.  Existing map elements have to not be changed in order for the pointed objects to remain valid.  It may currently work because this PR implicitly guarantees the returned vector only exists while the map is not changed (does the PR implicitly guarantee that?).  If someone else comes along and edits this code, they may not see the pattern and may alter the map, breaking the pointers.  Is there a way to have something else in the vector like store the matching keys so the caller can lookup in the map directly, or can the values in the map be unique_ptrs?

panos-lunarg: These search functions search the maps that hold the transfer dumping contexts. They are populated inside the constructor which is called when the command indices (the indices are the maps' keys) from the input json file are consumed and they don't change after that. Unfortunately their initialization is not trivial and can't be done in the constructor initializer list and mark the variables as `const`.
Exactly the same logic is used with the other similar search functions that search the other dumping contexts (draw calls/rt/dispatch).

You are right that the fact that they do not change after that is not evident and that leaves them vulnerable. Making them store (smart) pointers instead of the contexts themselves should guard against this.

framework/decode/metadata_json_consumer.h
```diff
@@ -314,6 +315,40 @@ class MetadataJsonConsumer : public Base
...
+        auto&              jdata        = WriteMetaCommandStart("CopyVulkanAccelerationStructureMetaCommand");
+        const JsonOptions& json_options = GetJsonOptions();
+        FieldToJson(jdata["pInfo"], copy_infos, json_options);
+        WriteBlockEnd();
+    }
+
+    virtual void ProcessVulkanAccelerationStructuresWritePropertiesMetaCommand(
+        format::HandleId device_id, VkQueryType query_type, format::HandleId acceleration_structure_id) override
+    {
+        auto&              jdata        = WriteMetaCommandStart("BuildVulkanAccelerationStructuresMetaCommand");
```
bradgrantham-lunarg: Should this be `WriteMetaCommandStart("VulkanWriteAccelerationStructuresPropertiesCommand");`?

framework/decode/metadata_json_consumer.h
```diff
@@ -314,6 +315,40 @@ class MetadataJsonConsumer : public Base
         WriteBlockEnd();
     }
 
+    virtual void ProcessBuildVulkanAccelerationStructuresMetaCommand(
+        format::HandleId                                                           device_id,
+        uint32_t                                                                   info_count,
+        StructPointerDecoder<Decoded_VkAccelerationStructureBuildGeometryInfoKHR>* geometry_infos,
+        StructPointerDecoder<Decoded_VkAccelerationStructureBuildRangeInfoKHR*>*   range_infos) override
+    {
+        auto&              jdata        = WriteMetaCommandStart("BuildVulkanAccelerationStructuresMetaCommand");
```
bradgrantham-lunarg: To be consistent with the rest of the metadata JSON this should be the name from format.h, "VulkanBuildAccelerationStructuresCommand"

framework/decode/metadata_json_consumer.h
```diff
@@ -314,6 +315,40 @@ class MetadataJsonConsumer : public Base
...
+        FieldToJson(jdata["pInfos"], geometry_infos, json_options);
+        FieldToJson(jdata["ppBuildRangeInfos"], range_infos, json_options);
+        WriteBlockEnd();
+    }
+
+    void ProcessCopyVulkanAccelerationStructuresMetaCommand(
+        format::HandleId                                                  device_id,
+        StructPointerDecoder<Decoded_VkCopyAccelerationStructureInfoKHR>* copy_infos) override
+    {
+        auto&              jdata        = WriteMetaCommandStart("CopyVulkanAccelerationStructureMetaCommand");
```
bradgrantham-lunarg: To be consistent with the rest of the metadata JSON this should be the name from format.h, "VulkanCopyAccelerationStructuresCommand"

framework/decode/vulkan_replay_dump_resources_common.cpp
```diff
@@ -1571,87 +1666,224 @@ void AccelerationStructureDumpResourcesContext::ReleaseResources()
...
+    for (const auto& build_data : as_context->as_build_objects)
     {
-        GFXRECON_LOG_WARNING("Could not detect address tracker for device %" PRIu64, device_info->capture_id);
-        return VK_SUCCESS;
-    }
+        if (auto* triangles = std::get_if<AccelerationStructureDumpResourcesContext::Triangles>(&build_data))
+        {
+            auto& blas_triangles_variant = dumped_as.input_buffers.emplace_back(
+                std::in_place_type<DumpedAccelerationStructure::DumpedBuildInputTriangleBuffer>,
+                triangles->vertex_format,
```
bradgrantham-lunarg: This is almost every field in `AccelerationStructureDumpResourcesContext::Triangles` and it feels like `DumpedBuildInputTriangleBuffer` should be constructed from a `Triangles`.  Should `Triangles` be a standalone type declared earlier in the header that is used to construct `DumpedBuildInputTriangleBuffer` and then is used in `AccelerationStructureDumpResourcesContext::as_build_objects`?

And if Triangles should be a standalone struct/class, should AABBS and Instances also be standalone?

framework/decode/vulkan_replay_dump_resources_common.cpp
```diff
@@ -1668,173 +1900,1000 @@ VkResult DumpTopLevelAccelerationStructure(DumpedTopLevelAccelerationStructure&
...
+        device_table.DestroyPipelineLayout(device, *ppl_layout, nullptr);
+        GFXRECON_LOG_WARNING("vkCreateComputePipelines failed (%s)", util::ToString(res).c_str());
+    }
+
+    device_table.DestroyShaderModule(device, compute_shader, nullptr);
+
+    return VK_SUCCESS;
+}
+
+VkResult AccelerationStructureDumpResourcesContext::CloneBuildAccelerationStructuresInputBuffers(
```
bradgrantham-lunarg: These and other functions later in this commit are very long, so it's hard for me to trace execution and dependencies.  Are there opportunities to break out parts of these functions into smaller functions that make this easier to read and reason about?  Like maybe utilities for inserting buffer copies and barriers into commandbuffers, and maybe breaking out subfunctions for the geometry types?

framework/decode/vulkan_replay_dump_resources_common.h
```diff
@@ -158,21 +153,40 @@ struct DumpedBuffer : DumpedFile
...
+    struct
+    {
+        VkBuffer         handle{ VK_NULL_HANDLE };
+        format::HandleId capture_id{ format::kNullHandleId };
+    } buffer_info;
+
+    VkDeviceSize offset{ 0 };
 
     void CopyDumpedInfo(const DumpedBuffer& other)
     {
```
bradgrantham-lunarg: This feels like it should be a constructor or `operator=`.

framework/decode/vulkan_replay_dump_resources_common.h
```diff
@@ -158,21 +153,40 @@ struct DumpedBuffer : DumpedFile
 {
     DumpedBuffer() = default;
 
-    DumpedBuffer(const VulkanBufferInfo* bi, VkDeviceSize o, VkDeviceSize s) : DumpedFile(s), buffer_info(bi), offset(o)
-    {}
+    DumpedBuffer(VkBuffer buffer, format::HandleId id, VkDeviceSize o, VkDeviceSize s) : DumpedFile(s), offset(o)
+    {
+        buffer_info.handle     = buffer;
```
bradgrantham-lunarg: Since each of these constructors also construct both members of `buffer_info`, it feels to me like the anonymous struct for buffer_info should be named, have a constructor, and be constructed in the member constructor list.

framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -406,6 +406,39 @@ VulkanReplayDumpResourcesBase::FindDrawCallCommandBufferContext(uint64_t bcb_id)
     return contexts;
 }
 
+void VulkanReplayDumpResourcesBase::ReleaseDrawCallContexts(decode::Index qs_index)
+{
+    for (auto it = draw_call_contexts_.begin(); it != draw_call_contexts_.end(); ++it)
```
bradgrantham-lunarg, line 364: Can these three Release functions be range-fors?

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_replay_dump_resources_delegate.cpp
```diff
@@ -1851,5 +1992,664 @@ DefaultVulkanDumpResourcesDelegate::GenerateASDumpedBufferFilename(const DumpedR
...
+    else
+    {
+        GFXRECON_LOG_ERROR("%s(): Unexpected transfer command type", __func__);
+        GFXRECON_ASSERT(0);
+    }
+
+    return true;
+}
+
+void DefaultVulkanDumpResourcesDelegate::GenerateOutputJsonTransferInfo(
```
bradgrantham-lunarg, line 2620: This is a really long function - would it make sense to break it out into smaller utility subfunctions for readability?  Maybe just the chunks on the inside of the case blocks?

framework/decode/vulkan_replay_dump_resources_transfer.h
```diff
@@ -0,0 +1,826 @@
...
+                       bool                               bc,
+                       TransferCommandTypes               t) :
+            params(std::in_place_type<CopyBuffer>, s, d),
+            type(t), has_before_command(bc), device_table_(dt), parent_device_info_(pdi)
+        {
+            GFXRECON_ASSERT(t == TransferCommandTypes::kCmdCopyBuffer);
+
+            if (bc)
+            {
+                before_params = CopyBuffer(s, d);
```
bradgrantham-lunarg: It looks like before_params is identical to params; can you help me understand what the purpose is of having separate params and before_params?

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_replay_dump_resources_transfer.h
```diff
@@ -0,0 +1,826 @@
...
+            CopiedImage(const VulkanImageInfo* img_info) :
+                image_info(*img_info), image(VK_NULL_HANDLE), memory(VK_NULL_HANDLE)
+            {}
+
+            VkImage         image;
+            VkDeviceMemory  memory;
+            VulkanImageInfo image_info;
+        };
+
+        // kInitBufferCommand
```
bradgrantham-lunarg, line 251: I think these would be better as a polymorphic construct, like TransferParamsBase, and then subclasses - virtual destructors would remove the get_if chain and the variants wouldn't need to be maintained when the list of classes changes.  There are a lot of dependencies on things, like sometimes there's a `before_params` and sometimes not, and it would be nice if that was presented directly in the subclass and not left to being an empty reference or variant.

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_replay_dump_resources.cpp
bradgrantham-lunarg: I think I understand - if they were `const` then you could pass back references and feel safe about that, right?  I imagine there may be a way to make them be constructable but I don't know enough about the class hierachy and it might be hard to tease apart.  I made some noise about smart pointers but could you leave them in the maps with a bunch of `// XXX this mustn't be touched after initialization` and return a vector of `reference_wrapper`?

framework/decode/metadata_json_consumer.h
bradgrantham-lunarg: For this and the other two changes to the meta Json consumer, I extracted https://github.com/LunarG/gfxreconstruct/pull/2585 and fixed up the consumer names, which were not consistent with format.h.  If you want to drop out the AS commits from this PR, you could resolve these comments and then review 2585.

framework/decode/vulkan_replay_dump_resources.cpp
panos-lunarg: After looking a bit more into this I realized that in the constructor there are a couple of cases where new elements are inserted while wanting to keep previous search results valid, so I decided to keep the smart pointers.
The maps are still constant outside of the constructor but I am planning to start deleting  contexts once they are done in the new Release*Contexts() functions. So I am thinking keeping the smart pointers.

framework/decode/metadata_json_consumer.h
panos-lunarg: I think the comments regarding the json consumer are addressed after https://github.com/LunarG/gfxreconstruct/pull/2585 was merged

framework/decode/vulkan_replay_dump_resources_common.cpp
panos-lunarg: Instead of making the structures standalone I did a small refactoring where I move `AccelerationStructureDumpResourcesContext::Triangles/AABBS/Instances`  into a separate file. Same for `DumpedBuildInputTriangleBuffer` and similar structures. Now `DumpedBuildInputTriangleBuffer`'s constructor accepts a `AccelerationStructureDumpResourcesContext::Triangles` as input. Same with the other 2 structs

framework/decode/vulkan_replay_dump_resources_common.cpp
panos-lunarg: I factored out the CmdCopyBuffer + Barrier that was repeating all over the place in those two functions (in `CopyBufferAndBarrier()`. I'll see breaking down the functions into smaller ones

framework/decode/vulkan_replay_dump_resources_common.h
panos-lunarg: Done

framework/decode/vulkan_replay_dump_resources_common.h
panos-lunarg: Done

framework/decode/vulkan_replay_dump_resources.cpp
panos-lunarg: Done

framework/decode/vulkan_replay_dump_resources_delegate.cpp
panos-lunarg: Done

framework/decode/vulkan_replay_dump_resources_transfer.h
panos-lunarg: `params` (and `before_params`), apart from the call parameters, they store vulkan objects (buffers, images, and backing device memory) that the target of each transfer command is copied into. The call parameters are common but the resource copy need to be separate before and after the command.

framework/decode/vulkan_replay_dump_resources_transfer.h
panos-lunarg: Done

panos-lunarg:
@roberto-c FYI

PR #2565, opened 2025-12-16
Move "Entering android_main" message

vaage:
The "Entering android_main" message was called before the logger was initialized. This worked because Log::Settings is set to valid defaults, allowing GFXRECON_WRITE_CONSOLE to pass through the function and trigger the Android log message.


vaage:
See #2552 for additional context.

MarkY-LunarG (approved):

PR #2567, opened 2025-12-16
Set current ParsedBlock on decoders and consumers

davidd-lunarg:
These changes are made to support the changes for PR https://github.com/LunarG/gfxreconstruct/pull/2543


framework/decode/block_parser.cpp
```diff
@@ -1,1724 +1,1771 @@
-/*
-** Copyright (c) 2018 Valve Corporation
```
jzulauf-lunarg: Why is this whole file changed?

framework/decode/block_buffer.h
```diff
@@ -106,7 +106,6 @@ class BlockBuffer
   private:
     void                InitBlockHeaderFromSpan();
     size_t              read_pos_{ 0 };
-    uint64_t            block_index_{ 0U };
```
jzulauf-lunarg, line 116: The BlockParser holds the block_index so the BlockBuffer shouldn't have to, the block parser should be able to injected it into the ParsedBlock

framework/decode/block_parser.cpp
```diff
@@ -1,1724 +1,1771 @@
...
+            const size_t           data_size      = static_cast<size_t>(message_size);
+            BlockBuffer::BlockSpan parameter_data = block_buffer.ReadSpan(data_size);
+            success                               = parameter_data.size() == data_size;
+
+            if (success)
+            {
+                const char* message_start = parameter_data.GetDataAs<char>();
+                std::string message(message_start, std::next(message_start, static_cast<size_t>(message_size)));
+
+                return ParsedBlock(GetBlockIndex(),
```
jzulauf-lunarg: I've literally rewritten every single constructor call site to call one of two factory functions.  The BlockParser can inject the the BlockIndex in those functions.  So I'm wondering if you can wait < 1 week whether there's a lower churn solution that adds this support after my refactor

jzulauf-lunarg (commented):
Something is very wrong with this PR.  More than one files is diff'ing "all lines changed."

Also, migrating the block index into the ParsedBlock will be trivial after my current pending refactor of ParsedBlock creation.  I'll push a WIP of that later today.  There's a little cleanup first.

framework/decode/block_parser.cpp
davidd-lunarg: Line endings changed from `CRLF` to `LF`. I've started a discussion about GFXR's policy on EOL internally. In the meantime you can disable whitespace diff in this PR's "Files changed" page.

framework/decode/block_buffer.h
davidd-lunarg: Right. This was removed because it wasn't being used and now ParsedBlock has the block index.

framework/decode/block_parser.cpp
davidd-lunarg: Yes that is no problem.

davidd-lunarg:
@jzulauf-lunarg This PR has been rebased onto the latest changes in `dev`

framework/decode/block_parser.cpp
```diff
@@ -290,9 +292,11 @@ ParsedBlock BlockParser::ParseBlock(BlockBuffer& block_buffer)
         case format::kUnknownBlock:
         default:
             WarnUnknownBlock(block_buffer);
-            return ParsedBlock{ ParsedBlock::UnknownBlockTag(), block_index_, block_buffer.ReleaseData() };
+            return ParsedBlock{ ParsedBlock::UnknownBlockTag(), GetBlockIndex(), block_buffer.ReleaseData() };
             break;
     }
+
+    current_block_index_ = 0;
```
MarkY-LunarG: Why breaks?  You're always returning.  And this code seems unnecessary.

davidd-lunarg: This PR doesn't modify the control logic, it only changed `block_index_` to `GetBlockIndex()`. However, since the `break`s do appear to be redundant, I've removed them. https://github.com/LunarG/gfxreconstruct/pull/2567/changes/70f0465302c36e70e23f87ee84ad503d32ae1d7e

PR #2568, opened 2025-12-16
Support keyboard trigger in wayland capture

### LINKED ISSUE #2460: Can't trigger on Wayland linux
I have a 2GB gfxr file and want to trim it, with the 1.4.328.1 SDK build I have tried

Image

and neither seem to work

not sure if this is a GFXR issue or just a Wayland VkConfig issue so I tried without VkConfig and set both

```
export VK_LAYER_GFXRECON_CAPTURE_TRIGGER=TAB 
export GFXRECON_CAPTURE_TRIGGER=TAB
```

and that didn't work, so makes me think the Wayland code is broken here

charles-lunarg:


charles-lunarg:
Need to add more error handling and fix some issues described by some users who I asked for advice on creating this PR.

fabian-lunarg:
> Need to add more error handling and fix some issues described by some users who I asked for advice on creating this PR.

this PR would be awesome @charles-lunarg . what's missing?

charles-lunarg:
That was written before  I did a second pass to add proper errors and make gfxr 'independent' from the applications wayland objects. Took some effort but is well worth it. So I am ready for this to go in if it passes review.

framework/util/keyboard.cpp
```diff
@@ -195,8 +315,109 @@ bool Keyboard::GetKeyState(const std::string& key)
...
+{}
+
+void Keyboard::HandleKeyboardKey(
+    void* data, struct wl_keyboard* keyboard, uint32_t serial, uint32_t time, uint32_t key, uint32_t state)
+{
+    static const std::unordered_map<uint32_t, uint32_t> wayland_key_code_map = {
+        { KEY_F1, 0 },   { KEY_F2, 1 },   { KEY_F3, 2 },   { KEY_F4, 3 },        { KEY_F5, 4 },
+        { KEY_F6, 5 },   { KEY_F7, 6 },   { KEY_F8, 7 },   { KEY_F9, 8 },        { KEY_F10, 9 },
+        { KEY_F11, 10 }, { KEY_F12, 11 }, { KEY_TAB, 12 }, { KEY_LEFTCTRL, 13 }, { KEY_RIGHTCTRL, 14 }
+    };
```
fabian-lunarg, line 403: you're repeating this map twice. I figure you're trying to avoid a global static map due to lifetime?
could we instead provide a DRY function?

```
static const std::unordered_map<uint32_t, uint32_t>& wayland_key_code_map()
{
    static std::unordered_map<uint32_t, uint32_t> ret = {...};
    return ret;
}
```

fabian-lunarg (approved):
trigger now works beautifully for me on wayland.
code is clean throughout 
~~and I could only point out this duplicated map, which would be nice to make DRY.~~ (once the coffee starting working I realized its fine) 

let's merge :)

framework/util/keyboard.cpp
fabian-lunarg: ok, nevermind. it's two different maps :)

PR #2570, opened 2025-12-17
Jenkins Pipelines

beau-lunarg:
Extended Job:
http://jenkins:8080/job/beau-lunarg/job/gfxreconstruct-pipeline-dev-extended/13/

kicked off by per commit job
http://jenkins:8080/job/beau-lunarg/job/gfxreconstruct-pipeline-dev/job/dev/10/

This can be merged into dev without breaking existing legacy Jenkins jobs.  Pipelines will need to be enabled after this is merged

See the following for a list of new pipelines:
http://jenkins:8080/job/beau-lunarg/

### DIFF
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -64,6 +64,8 @@ pipeline {
     }
 }
 
+// extended should only be run between 8pm and midnight, only on weekdays
+// this function will calculate the amount of time to wait until the next time extended should run
 def calculateQuietPeriod() {
     def now = new Date()
     def hour = now.getHours()


ci/Jenkinsfile
```diff
@@ -0,0 +1,90 @@
...
+
+        stage('Trigger Extended Tests') {
+            when {
+                branch 'dev'
+            }
+
+            steps {
+                script {
+                    def quietPeriodSeconds = calculateQuietPeriod()
+                    build job: 'beau-lunarg/gfxreconstruct-pipeline-dev-extended',
```
lunarpapillo, line 57: The reference to `beau-lunarg` will have to change, yes?

ci/Jenkinsfile
```diff
@@ -0,0 +1,90 @@
...
+            }
+        }
+
+        stage('Trigger Extended Tests') {
+            when {
+                branch 'dev'
+            }
+
+            steps {
+                script {
```
lunarpapillo, line 55: This script will wait until the end of the "quiet period", as defined below... but it will still run once per merge to `dev`, correct?  So if a bunch of separate merges to `dev` come through, we'll immediately backlog the extended suite.

I _think_ part of the point of having nightly runs of the long-running tests is to _not_ backlog, by running only nightly with whatever is top-of-`dev`, knowing that if something fails and there was more than one commit, we'll have to do some detective work.

If this policy has changed, please ignore me. 😄

ci/Jenkinsfile
```diff
@@ -0,0 +1,90 @@
...
+    }
+}
+
+def calculateQuietPeriod() {
+    def now = new Date()
+    def hour = now.getHours()
+    def dayOfWeek = now.getDay()  // 0=Sun, 1=Mon, ..., 6=Sat
+
+    // In window (Mon-Fri 8pm-midnight): run immediately
+    if (dayOfWeek >= 1 && dayOfWeek <= 5 && hour >= 20 && hour < 24) {
```
lunarpapillo, line 75: This is acceptable.  But consider using constants like `DayOfWeek.SATURDAY` (from `java.time.DayOfWeek`) in the comparisons instead of integers, to make the code more readable and maintainable.

ci/Jenkinsfile
```diff
@@ -0,0 +1,90 @@
...
+    def target = now.clone()
+    target.setHours(20)
+    target.setMinutes(0)
+    target.setSeconds(0)
+
+    // If today is weekday and before 8pm, target is today
+    if (dayOfWeek >= 1 && dayOfWeek <= 5 && hour < 20) {
+        // target is today at 8pm, already set
+    }
+    // If weekend or after midnight, find next weekday
```
lunarpapillo, line 89: This does match what the current polling does, which avoided running tests if no one was going to be in the office to review the results.

But I'm wondering whether we should remove the weekday-only restriction and allow runs on the weekend, as we do have GFXR people who work on weekends now, and the machines are likely to be quiet anyway.

ci/Jenkinsfile
```diff
@@ -0,0 +1,90 @@
...
+    // In window (Mon-Fri 8pm-midnight): run immediately
+    if (dayOfWeek >= 1 && dayOfWeek <= 5 && hour >= 20 && hour < 24) {
+        return 0
+    }
+
+    // Calculate seconds until next 8pm Mon-Fri
+    def target = now.clone()
+    target.setHours(20)
+    target.setMinutes(0)
+    target.setSeconds(0)
```
lunarpapillo, line 83: Acceptable, but a little odd given that we reset `Seconds` to 0 but not milliseconds.

ci/Jenkinsfile.manual
```diff
@@ -0,0 +1,139 @@
...
+            description: 'Project branch to build'
+        )
+        string(
+            name: 'TEST_REPO',
+            defaultValue: 'git@github.com:LunarG/VulkanTests',
+            description: 'Test repository URL'
+        )
+        string(
+            name: 'TEST_BRANCH',
+            defaultValue: '',
```
lunarpapillo, line 22: This is a little weird, in that it doesn't match the default parameters in other Jenkins jobs.

Can you explain why this choice was made?

ci/runJob.bat
```diff
@@ -0,0 +1,61 @@
...
+git clone --verbose %TEST_SUITE_REPO% ci-gfxr-suites
+if %errorlevel% equ 0 goto :clone_suites_done
+if exist ci-gfxr-suites/ rmdir /s /q ci-gfxr-suites
+set /a clonetestloop+=1
+if %clonetestloop% gtr 3 ( 
+    echo tried to clone %TEST_SUITE_REPO% too many times, giving up
+    exit 1
+)
+waitfor forever /t 60 2>nul
+goto :clone_suites
```
lunarpapillo, line 23: This is acceptable, but I was hoping we'd eventually lose this weird `.bat` code in favor of Groovy `retry()`.

ci/runJob.groovy
```diff
@@ -0,0 +1,395 @@
...
+                        "BUILD_MODE=${buildMode}",
+                        "RESULTS_DIR=../vulkantest-results/${name}"
+                    ]) {
+                        bat 'ci/runJob.bat'
+                    }
+                }
+
+                archiveArtifacts(
+                    artifacts: 'vulkantest-results/**',
+                    excludes: 'vulkantest-results/**/*.gfxr,vulkantest-results/**/core*,vulkantest-results/**/*.jsonl',
```
lunarpapillo, line 45: need to exclude `**/*.gfxa` too (in all cases).

ci/runJob.groovy
```diff
@@ -0,0 +1,395 @@
...
+        stage(name) {
+            echo "About to allocate node with label: ${label}"
+            node(label) {
+                echo "Running on node: ${env.NODE_NAME} with label requirement: ${label}"
+
+                retry(3) {
+                    try {
+                        cleanWs(deleteDirs: true, disableDeferredWipeout: true)
+                    } catch (Exception e) {
+                        sleep(time: 5)
```
lunarpapillo, line 19: why was 5 seconds chosen?  (Is that the default?)

ci/runJob.groovy
```diff
@@ -0,0 +1,395 @@
...
+            }
+        }
+    }
+}
+
+// Derive job type from node name based on naming convention
+def getNodeType(String nodeName) {
+    if (nodeName.contains('and')) {
+        return 'android'
+    } else if (nodeName.contains('wnapl') || nodeName.contains('mac')) {
```
lunarpapillo, line 162: The old-style Mac machines are actually indicated by `dwn` (Darwin) and were not named by CI people - we just adopted the machine after it was configured.

But I'd rather rename those machines to `mac` than carry this botched naming scheme.

ci/runJob.groovy
```diff
@@ -0,0 +1,395 @@
...
+                    onlyIfSuccessful: false,
+                )
+
+            }
+        }
+    }
+}
+
+// Derive job type from node name based on naming convention
+def getNodeType(String nodeName) {
```
lunarpapillo, line 159: This is acceptable, though I'm not thrilled with choosing a node type based on the node name.  CI never intended for node names to be a supported interface (in fact, we've considered changing them all, e.g. to superhero names, to make them "pronounceable").  With this code, we'd have to formalize node name selection and prevent it from changing (a change to **VulkanTests** documentation would be sufficient).

The current jobs choose a node type based on system properties (generally the existence of particular files and directories), which has its own strengths and weaknesses.

I imagine it's possible to select a node type via the label, which is more comfortable as a supported interface... but then this doesn't work on a node with no labels (one CI workflow is to take a node "offline" by removing its labels, so we can continue to target it in Jenkins as we triage, while ensuring that it doesn't get scheduled for production jobs).

I guess I'd like to see some documentation, either inline as comments or in an external file somewhere, as to what alternatives for identifying node type were considered, and why this one was selected.

ci/runJob.groovy
```diff
@@ -0,0 +1,395 @@
+def gfxrTestWindows(
```
lunarpapillo, line 1: This is acceptable, though I'm not thrilled with the duplication of code between the `gfxrTestWindows()`, `gfxrTestLinux()`, and `gfxrTestAndroid()` functions (and with the implicit lack of `gfxrTestMac()` - perhaps `gfxrTestPosix()` might be better?).

These functions all function identically, except for the use of `bat` or `sh`, and the use of `runjob.bat` or `runjob.sh`, yes?  Is there some way to parametrize this so the common code can be factored out?

ci/runJob.groovy
```diff
@@ -0,0 +1,395 @@
...
+    } else if (nodeName.contains('ubu')) {
+        return 'linux'
+    } else if (nodeName.contains('win')) {
+        return 'windows'
+    } else {
+        return null
+    }
+}
+
+def gfxrTestWindowsManual(
```
lunarpapillo, line 173: Similar commonality here.

ci/runJob.groovy
```diff
@@ -0,0 +1,395 @@
...
+                        "BITS=${bits}",
+                        "BUILD_MODE=${buildMode}",
+                        "RESULTS_DIR=../vulkantest-results/${stageName}"
+                    ]) {
+                        sh 'ci/runJobAndroid.sh'
+                    }
+
+                    archiveArtifacts(
+                        artifacts: 'vulkantest-results/**',
+                        excludes: 'vulkantest-results/**/*.gfxr,vulkantest-results/**/core*,vulkantest-results/**/*.jsonl',
```
lunarpapillo: Probably doesn't technically need `**/*.gfxa`, but I'd include it for consistency.

ci/Jenkinsfile
```diff
@@ -0,0 +1,96 @@
...
+    stages {
+        stage('Per Commit Tests') {
+            steps {
+                script {
+                    def tests
+
+                    // Checkout and load on any available node
+                    node {
+                        checkout scm
+                        def runJob = load 'ci/runJob.groovy'
```
ncesario-lunarg, line 23: Is the idea here to eventually load this via a library that can be placed in a separate repo?

ci/Jenkinsfile
ncesario-lunarg: Should this be here? I may have misunderstood the layout, but at first glance I thought all of the extended logic (including triggers) would be in `Jenkinsfile.extended`.

ci/Jenkinsfile
```diff
@@ -0,0 +1,96 @@
...
+                          wait: false,
+                          propagate: false,
+                          quietPeriod: quietPeriodSeconds
+                }
+            }
+        }
+    }
+}
+
+def calculateQuietPeriod() {
```
ncesario-lunarg, line 69: It might be good to have a comment here indicating what this does. I _think_ the comment on line 72 describes the goal here, and I'd probably vote for having that up here. Also, since we're already including additional Groovy files, would it make sense to put this in there?

ci/Jenkinsfile
ncesario-lunarg: That's probably another reason to put this into a separate Groovy file: things like Java imports can be put in there, and the Jenkinsfile can be kept mostly (if not entirely) declarative.

ci/runJob.bat
ncesario-lunarg: I rather loathe having to deal with batch files in general, and I hope this eventually turns into a Python (or Groovy I suppose) script, but I would think that would/should be a separate change from this one, which is just moving the orchestration from Jenkins UI to pipelines.

ci/runJob.groovy
ncesario-lunarg: https://github.com/ncesario-lunarg/jenkins-testing/blob/main/vars/buildGFXR.groovy#L16-L22 is one way to conditionally run commands in different shells based on OS (and [example usage](https://github.com/ncesario-lunarg/jenkins-testing/blob/main/vars/buildGFXR.groovy#L56-L71)). It should be possible to consolidate most of these functions with something like that.

ci/Jenkinsfile.manual
```diff
@@ -0,0 +1,148 @@
...
+                        }
+
+                        selectedNodes.each { nodeName ->
+                            def type = runJob.getNodeType(nodeName)
+                            if (!type) {
+                                error "Cannot determine type for node: ${nodeName}. Name must contain 'and', 'ubu', 'mac', 'wnapl', or 'win'."
+                            }
+
+                            def closure
+                            switch (type) {
```
ncesario-lunarg, line 91: This seems a little like an "anti-pipeline pattern," but this also might be the best way to "transition" from the current "UI configuration." Ideally, I would think each of these cases is a stage which only runs on a specific set of labels, which would hopefully remove the need for all of the `selectedNodes` and `runJob.getNodeType` logic.

ci/runJob.groovy
ncesario-lunarg: Yeah, this is kind of rough. See comment about this logic in `Jenkinsfile.manual` above.

ci/Jenkinsfile.manual
```diff
@@ -0,0 +1,148 @@
...
+        stage('Run Tests') {
+            steps {
+                script {
+                    def tests = [:]
+
+                    node {
+                        checkout scm
+                        def runJob = load 'ci/runJob.groovy'
+
+                        def selectedNodes = params.NODES?.split(',')?.collect { it.trim() }?.findAll { it }
```
ncesario-lunarg, line 76: This seems a bit backwards. Usually I would think we are given jobs and select nodes based on labels defined per job. I know this is trying to keep things similar to the existing job configurations, though, so maybe this is good as a first transition step.

ncesario-lunarg (commented):
I think this is a great first step at transitioning to pipelines, and I'd vote for getting this in sooner rather than later (pending @lunarpapillo's requested changes) in order to get things ironed out quicker. I think Beau and I have done all the testing we can in a test environment and we need to actually try it out for real.

Some quick comments on @lunarpapillo's comments (commentception...)
> I'm aware of some of the potential benefits we gain from this change, but I confess I'm not currently clear on which were the most important reasons - to get repository control? To modernize? To avoid having to update the webhook server for per-configuration reporting to GitHub Actions? I'd love for this to be documented somewhere just for clarity.

One of the biggest boons here is making the CI configuration a part of the repo rather than having it completely decoupled from the repo. This allows repo authors to make changes to the configuration via PRs, giving us the ability to properly track changes to the build configuration, but--perhaps more importantly--we can trivially _test_ those configuration changes by simply creating PRs without affecting the configuration on the main branch. The "modernization" and defining jobs using code rather than UI forms are huge QoL improvements as well IMHO.

ci/Jenkinsfile
beau-lunarg: This will change once the actual jobs have been created

ci/Jenkinsfile
beau-lunarg: > This script will wait until the end of the "quiet period", as defined below... but it will still run once per merge to `dev`, correct? So if a bunch of separate merges to `dev` come through, we'll immediately backlog the extended suite.
> 
> I _think_ part of the point of having nightly runs of the long-running tests is to _not_ backlog, by running only nightly with whatever is top-of-`dev`, knowing that if something fails and there was more than one commit, we'll have to do some detective work.
> 
> If this policy has changed, please ignore me. 😄

The extended pipeline, similar to this one, have `disableConcurrentBuilds(abortPrevious: true)` set, which will always just run the latest, and will cancel the previous ones.  If multiple dev builds complete successfully, it will only run extended once that night.

beau-lunarg: > Should this be here? I may have misunderstood the layout, but at first glance I thought all of the extended logic (including triggers) would be in `Jenkinsfile.extended`.

This simply says "Trigger the extended pipeline when this pipeline completes successfully in dev".  The definition for what happens in the extended pipeline is still in the Jenkinsfile.extended.  I didn't see a way in Jenkinsfile.extended to say "start me up if this other pipeline completes successfully"

ci/Jenkinsfile
beau-lunarg: This logic is built off of the cron that I pulled from the existing extended job, which is weekdays only.  If we need to change that logic I'd rather take care of that as a separate PR once this gets in.  I'd like for this PR to be aimed at parity.  Improvement comes later

ci/runJob.groovy
beau-lunarg: Fixed

ci/Jenkinsfile
beau-lunarg: If we needed to we could.  We would have similar versioning issues like we have with ci-gfxr-suites and VulkanTests right now.  I'm not sure what the resolution would be for that.

ci/Jenkinsfile.manual
beau-lunarg: I mentioned this in the meeting a couple weeks ago, but I agree that this interface isn't the best.  But right now I'm wanting to get to parity and then improve from there.

Selecting labels rather than machines would be best, yes.  Or even just the high level combos (Mac 64bit Release, Windows AMD 32bit Debug, etc)

ci/Jenkinsfile.manual
beau-lunarg: Mentioned this in another comment, but the ideal would be to give a selection of labels or configurations.  Right now I was just aiming at parity.

beau-lunarg:
@ncesario-lunarg 
`https://github.com/ncesario-lunarg/jenkins-testing/blob/main/vars/buildGFXR.groovy#L16-L22 is one way to conditionally run commands in different shells based on OS (and [example usage](https://github.com/ncesario-lunarg/jenkins-testing/blob/main/vars/buildGFXR.groovy#L56-L71)). It should be possible to consolidate most of these functions with something like that.`

I like this, but I would have to dig in to figure out how to convert.  I think it dovetails well with another comment about not having the shell script files.  Could we make this change after we merge this in and get it running, since this is working now and these things can be a little fragile and take a bit of time to test entirely?

beau-lunarg:
@ncesario-lunarg 

`I rather loathe having to deal with batch files in general, and I hope this eventually turns into a Python (or Groovy I suppose) script, but I would think that would/should be a separate change from this one, which is just moving the orchestration from Jenkins UI to pipelines.`

These scripts are almost 100% copy/pasted from the manual legacy configuration.  You have another comment in here about turning them into proper cross-platform groovy, which i'm not opposed to.  Could that go into a PR after this one is merged?

lunarpapillo:
> One of the biggest boons here is making the CI configuration a part of the repo rather than having it completely decoupled from the repo. This allows repo authors to make changes to the configuration via PRs, giving us the ability to properly track changes to the build configuration, but--perhaps more importantly--we can trivially _test_ those configuration changes by simply creating PRs without affecting the configuration on the main branch. 

Is there more to this than "what combinations of OS, GPU, and bitness are we testing"?  How often do we expect this to change?  (I'm not looking to challenge, just to understand.)

I've mentioned before (but maybe not in any written discussion?) that moving this control to the repository removes CI's ability to reconfigure on the fly should a family of test machines suffer catastrophic failure, as recovery would require the ability to emergency force-merge a PR to any repository using 'Jenkinsfile` pipelines.  Admittedly, this is rare, especially since we now have twins of every configuration.  But it did happen in the last year despite that. 

I just want to make sure we all recognize and understand that a similar catastrophic failure will essentially shut down GFXR testing with no way for the CI administrators to recover.  I'm okay with that, but feel it important to make sure everyone recognizes that.

> The "modernization" and defining jobs using code rather than UI forms are huge QoL improvements as well IMHO.

For the record, so far I'm impressed at how much time it took to create a simple pipeline job and get it running, even with AI help, and how hard it would be to keep it running; especially as compared to UI forms which are much less powerful but easier to create and modify...  I'm willing to be convinced that pipelines will eventually be a "huge" QoL improvement, but at the moment I'm skeptical - if you're doing something that doesn't need pipeline capabilities, it still seems like using a standard **Freestyle** or **Multi-configuration** job is easier.

beau-lunarg:
> removes CI's ability to reconfigure on the fly should a family of test machines suffer catastrophic failure, as recovery would require the ability to emergency force-merge a PR to any repository using 'Jenkinsfile` pipelines.

Could you describe the catastrophic failure in more detail?  What would that look like?  What would need reconfiguring?

beau-lunarg:
> even with AI help, and how hard it would be to keep it running

I'll admit to using AI here to help educate myself.  Although it gave me lots of incorrect pieces at first.  It wasn't until I had the full script running and working that it was able to help.  I think the groovy syntax may have gone through several iterations of structural changes between versions, and so the LLM has lots of outdated information running through it.
Also, as per our AI policy, even when it gave me something correct, I had to dig into the documentation so that I could understand it.  This was even more time consuming when the AI suggestion was wrong.

ci/runJob.groovy
beau-lunarg: We use the Mac-M2 label as the selector for mac machines for gfxr.  There is a tcmac0 in our fleet that has that label.  It is an example of a machine that doesn't have dwn but is a mac.

lunarpapillo: > There is a tcmac0 in our fleet that has that label. It is an example of a machine that doesn't have dwn but is a mac.

Right - that's the correct naming convention.  `tcdwnapl` is incorrect - it was prepared by a non-CI person and adopted into CI without a name change (perhaps unwisely).

The Groovy code here should not check for `wnapl` (part of `tcdwnapl`) and should only look for `mac`.  And CI should rename the two badly named machines.

lunarpapillo:
> > removes CI's ability to reconfigure on the fly should a family of test machines suffer catastrophic failure, as recovery would require the ability to emergency force-merge a PR to any repository using 'Jenkinsfile` pipelines.
> 
> Could you describe the catastrophic failure in more detail? What would that look like? What would need reconfiguring?

A whole pile of machines - cross-platform - all started failing all VVL tests, while still passing some of the other tests.  We were able to shut down all the machines that had appeared to be affected, in order to let project CI continue while trying to figure out what had happened - it seemed similar to a GFXR PR a while back that crashed the GPUs on all machines it ran on, but the affected machines didn't respond to a restart in the same way.

It happened to be the case that the swath of machines affected included all machines that had one particular label (I can look up which label it was if you're interested).  That meant that CI _couldn't_ continue, because any job would block on the missing configuration.

We recovered by removing that configuration from all project CI in Jenkins, which allowed CI to continue but would require us to retest completely (as if adding a configuration for the first time) when the situation got resolved and the machines returned to CI use.

(If you're curious, it turned out the problem stemmed from a new **VulkanTools** test that was altering the loader configuration file on the system (bad) and never cleaning it up (worse), which meant that any machine that ran **VulkanTools** was corrupted.  We've since added code to the CI infrastructure to detect and report a compromised system immediately.  But the important thing is that any multi-machine failure can potentially require catastrophic recovery.)

If some situation like this were to happen to GFXR pipelines, we would have to kill all tests in flight, create a PR for every affected repository (only GFXR for now, but if this is the "way of the future", we have to be ready for it to apply to all repositories), get all of the PRs approved, merge all of the PRs, rebase all in-flight PRs, and then restart tests, to re-establish CI, in an urgent timeframe (quite possibly when only CI people were awake).

To get the same functionality, then, we'd need around-the-clock GFXR people with the authority to force-approve PRs ready to act on demand.  I don't think we need this functionality - this level of catastrophic failure is very rare, and deliberately so - but we should be consciously aware that we're choosing to sacrifice the ability to recover from catastrophic failure with these changes.

lunarpapillo:
> I'll admit to using AI here to help educate myself. Although it gave me lots of incorrect pieces at first. It wasn't until I had the full script running and working that it was able to help. 

Just for the record (and to give y'all a chance to educate me in a better way to do this :-), I attempted a simpler, static pipeline job to replace an existing multi-configuration job.  Even feeding the job information into AI, it took over 10x the time to get the pipeline job running than it would take with the relatively bulletproof multi-configuration forms, and the job recently failed in a surprising way that turned out to be bad code in one of the `unsuccessful` branches, which makes me fearful that there are other potential code breaks in particular conditions that haven't been exercised yet.

My conclusion for now is that we should use pipelines when we need specific features that pipelines provide; but for simpler static jobs, multi-configuration is considerably easier to create correctly and maintain.

FWIW.

lunarpapillo:
> [beau-lunarg](https://github.com/beau-lunarg) [force-pushed](https://github.com/LunarG/gfxreconstruct/compare/cee5ef396fb0d30e17b13ba09decb458bd306502..6d025c1ab11539c29be6ae5b8d63f45936931776) the beau-pipelines branch from [cee5ef3](https://github.com/LunarG/gfxreconstruct/commit/cee5ef396fb0d30e17b13ba09decb458bd306502) to [6d025c1](https://github.com/LunarG/gfxreconstruct/commit/6d025c1ab11539c29be6ae5b8d63f45936931776) [2 weeks ago](https://github.com/LunarG/gfxreconstruct/pull/2570#event-21745761472)

Yeargh... this force-push means that to review the changes, I'd have to start the whole review over again, trying to be careful to distinguish code that remains unchanged from the first review (and can be skimmed) from code that _looks_ familiar but actually has significant changes.

The original code review took hours; I'm dreading having to spend hours again just to identify what are probably a small set of changes.

Can you summarize where the significant changes are, so I can focus on just those, rather than trying to guess at where the significant changes are myself?

ci/Jenkinsfile
ncesario-lunarg: If the goal is to only trigger on success, would it be clearer to have this in a "post" section? e.g.,
```jenkinsfile
post {
  success {
    build: ...
  }
}
```

ci/Jenkinsfile
beau-lunarg: Added a comment

beau-lunarg:
> > [beau-lunarg](https://github.com/beau-lunarg) [force-pushed](https://github.com/LunarG/gfxreconstruct/compare/cee5ef396fb0d30e17b13ba09decb458bd306502..6d025c1ab11539c29be6ae5b8d63f45936931776) the beau-pipelines branch from [cee5ef3](https://github.com/LunarG/gfxreconstruct/commit/cee5ef396fb0d30e17b13ba09decb458bd306502) to [6d025c1](https://github.com/LunarG/gfxreconstruct/commit/6d025c1ab11539c29be6ae5b8d63f45936931776) [2 weeks ago](https://github.com/LunarG/gfxreconstruct/pull/2570#event-21745761472)
> 
> Yeargh... this force-push means that to review the changes, I'd have to start the whole review over again, trying to be careful to distinguish code that remains unchanged from the first review (and can be skimmed) from code that _looks_ familiar but actually has significant changes.
> 
> The original code review took hours; I'm dreading having to spend hours again just to identify what are probably a small set of changes.
> 
> Can you summarize where the significant changes are, so I can focus on just those, rather than trying to guess at where the significant changes are myself?

Its really just the comment around the calculateQuietPeriod.  The rest was force pushes to get the build working.

lunarpapillo:
> I think this is a great first step at transitioning to pipelines, and I'd vote for getting this in sooner rather than later (pending @lunarpapillo's requested changes) in order to get things ironed out quicker. I think Beau and I have done all the testing we can in a test environment and we need to actually try it out for real.

I don't mind this.  Please create a tracking issue with the concerns that are being deferred so we don't lose sight of them once this merges and closes.

lunarpapillo:
> Its really just the comment around the calculateQuietPeriod. The rest was force pushes to get the build working.

lol - probably didn't need a review just for that... I looked anyway and it's fine.

I'll approve too, contingent on a tracking issue being created.

lunarpapillo (approved):
Assuming a tracking issue is going to be made for the issues that aren't immediately being addressed.

PR #2574, opened 2025-12-19
Fix `FileProcessor` error handling

marius-pelegrin-arm:
The error handler in `FileProcessor` and `PreloadFileProcessor` is a lambda declared in the scope of the function that may call it.

When "given" to the constructor of `BlockParser`, the function is wrapped in an `std::function` object and "given" as a constant reference. The constant reference is then saved directly in the `BlockParser::err_handler_`.
Thus, the underlying `std::function` object's lifetime is theoretically only the time of the `BlockParser` constructor's call, and not the entire function like the lambda's lifetime.

This leads to the `std::function` object being destroyed before the error handler may actually be called, leading to a `SEGFAULT` if the error handler ends up being actually called.

This `SEGFAULT` has been encountered when trying to run `gfxrecon-optimize` on a DX12 trace using a linux build - instead of exiting gracefully.

This commit fixes the issue by saving the `std::function` object instead of the lambda.

### DIFF
--- a/framework/decode/block_parser.cpp
+++ b/framework/decode/block_parser.cpp
@@ -87,19 +87,18 @@ BlockIOError BlockParser::ReadBlockBuffer(FileInputStreamPtr& input_stream, Bloc
             }
         }
     }
-    else
+
+    if (status != kErrorNone)
     {
-        if (status != kEndOfFile)
-        {
-            HandleBlockReadError(status, "ReadBlockBuffer failed");
-        }
         block_buffer.Reset();
     }
+
     return status;
 }
 
 void BlockParser::HandleBlockReadError(BlockIOError error_code, const char* error_message)
 {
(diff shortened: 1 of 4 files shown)


jzulauf-lunarg (approved):
LGTM:  I literally found this yesterday and was about to push a fix.  Thanks.

marius-pelegrin-arm:
@fabian-lunarg Yes, this would be easier, safer, and probably not more costly to just copy the std::function instead of trying to use a reference.  But I just tried to keep your code "as-is". Feel free to fix the bug in a way that suits you =)

fabian-lunarg:
ok, cool. I quickly confirmed above suggestion works and would add that as another commit into the PR.

fabian-lunarg:
@jzulauf-lunarg  Added a minimal regression-test around BlockParser-init and invoking the error-handler.

PR #2575, opened 2025-12-19
Remove unnecessary payload `data_size` duplication

marius-pelegrin-arm:
The `InitDx12AccelerationStructureCommandHeader` meta command header already contains the "uncompressed data size" as a field, it is just named `inputs_data_size` instead of `data_size`.

Because of this, an additonal `data_size` field has been added to the associated API Payload structure instead of simply renaming the field.

I propose to remove this additional `data_size` field and rename `inputs_data_size` into `data_size` to be more coherent with the meta command dispatching framework.

### DIFF
--- a/framework/decode/api_payload.h
+++ b/framework/decode/api_payload.h
@@ -429,8 +429,6 @@ struct InitSubresourceArgs
 struct InitDx12AccelerationStructureArgs
 {
     format::MetaDataId meta_data_id; // Needed by DispatchVisitor, but not ApiDecoder
-    // Note: the command header has uniquely named size field... so duplicate this info for visitor use
-    size_t data_size; // Needed for deferred decompression, but not ApiDecoder
 
     format::InitDx12AccelerationStructureCommandHeader             command_header;
     std::vector<format::InitDx12AccelerationStructureGeometryDesc> geometry_descs;
@@ -548,7 +546,7 @@ struct AnnotationArgs
 
     // NOTE: The string name is intentionally *not* data to differ from the "data" fields that are uint8_t *
     // parameter data for the next level Decode operations, simplifying DispatchHasData logic
-    std::string            annotation_data;
+    std::string annotation_data;
 
     auto GetTuple() const { return std::tie(block_index, type, label, annotation_data); }
 };
--- a/framework/decode/block_parser.cpp
+++ b/framework/decode/block_parser.cpp
@@ -1223,7 +1223,7 @@ ParsedBlock BlockParser::ParseMetaData(BlockBuffer& block_buffer)
         success = success && block_buffer.Read(header.inputs_flags);
         success = success && block_buffer.Read(header.inputs_num_instance_descs);
         success = success && block_buffer.Read(header.inputs_num_geometry_descs);
-        success = success && block_buffer.Read(header.inputs_data_size);
+        success = success && block_buffer.Read(header.data_size);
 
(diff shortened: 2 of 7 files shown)


marius-pelegrin-arm:
The error looks like a network error unrelated to my commit, can I get a confirmation ?

framework/decode/block_parser.cpp
```diff
@@ -1250,15 +1250,12 @@ ParsedBlock BlockParser::ParseMetaData(BlockBuffer& block_buffer)
         if (success)
         {
             const char*         label       = "init DX12 acceleration structure meta-data block";
-            ParameterReadResult read_result = ReadParameterBuffer(label, block_buffer, header.inputs_data_size);
+            ParameterReadResult read_result = ReadParameterBuffer(label, block_buffer, header.data_size);
```
davidd-lunarg, line 1253: The third argument `uncompressed_size` to `BlockParser::ReadParameterBuffer` is ignored when compression is off, so `header.data_size` (previously `header.inputs_data_size`) is not used to set `result.uncompressed_size`.

https://github.com/LunarG/gfxreconstruct/blob/aae6ac7eb25032695b20b2cf8e4f389e91edf0d2/framework/decode/block_parser.cpp#L259-L263

Looking at how `block_buffer.Remainder()` is calculated, I think `uncompressed_size == block_buffer.Remainder()` when the `uncompressed_size` argument is non-default. @jzulauf-lunarg Should an assert be added in `ReadParameterBuffer` to enforce/verify this? For example:

```
    else
    {
        // For uncompressed blocks the data size is the parameter buffer size
+       GFXRECON_ASSERT(uncompressed_size == kReadSizeFromBuffer || uncompressed_size == block_buffer.Remainder());
        result.uncompressed_size = block_buffer.Remainder();
    }
```

davidd-lunarg (commented):
This looks good to me. I do have one comment/question. @jzulauf-lunarg Could you also review?

framework/decode/api_payload.h
```diff
@@ -429,8 +429,6 @@ struct InitSubresourceArgs
 struct InitDx12AccelerationStructureArgs
 {
     format::MetaDataId meta_data_id; // Needed by DispatchVisitor, but not ApiDecoder
-    // Note: the command header has uniquely named size field... so duplicate this info for visitor use
-    size_t data_size; // Needed for deferred decompression, but not ApiDecoder
```
jzulauf-lunarg, line 433: Since DispatchHasCommandHeader and inline size_t GetDispatchArgsDataSize(Args& args)
 combine to find the renamed field in the command_header this should be fine.  Just didn't want to go renaming other peoples fields to match the "concept" of has "data_size" or "command_header.data_size" the meta programming for decompression requires.

But this is better/cleaner

framework/decode/block_parser.cpp
jzulauf-lunarg: Certainly could assert here.  I don't know for sure that all files were created with no extra bytes in their remainder when the remainder size is given explicitly, so this might generate new errors.

jzulauf-lunarg (approved):
This is a nice cleanup.  The duplicate avoided the need to rename the command_header field and touch all the places that this PR does, but it's the right thing to do.

Not as sure about the additional ASSERT suggested.  Should be fine, can't guarantee that no files exists that would ASSERT but currently function.

PR #2576, opened 2025-12-19
Extract dx12 shader

zongdu-arm:
Extract dx12 shader


fabian-lunarg:
looks good. we're currently still looking into how we can cover the proposed functionality in CI-testing.

davidlunarg:
I built this branch on Linux. When I attempted to extract shaders from a dx12 capture, I get the following error:

> terminate called after throwing an instance of 'std::bad_function_call'
>   what():  bad_function_call

I don't think it is a requirement that the feature implemented by this PR can extract shaders from a dx12 capture file when running on Linux, but it might be a nice convenience to some folks using this tool. The terminate error should at the very least be replaced by an error message that gfxrecon-extract cannot process a dx12 capture file.

zongdu-arm:
> I built this branch on Linux. When I attempted to extract shaders from a dx12 capture, I get the following error:
> 
> > terminate called after throwing an instance of 'std::bad_function_call'
> > what():  bad_function_call
> 
> I don't think it is a requirement that the feature implemented by this PR can extract shaders from a dx12 capture file when running on Linux, but it might be a nice convenience to some folks using this tool. The terminate error should at the very least be replaced by an error message that gfxrecon-extract cannot process a dx12 capture file.

Yes, it only works with D3D12 and can be used in Windows or Linux + Wine environments.
 If you extract shaders from a dx12 capture on linux, even without this PR, the same error will be reported.

davidlunarg (approved):

PR #2581, opened 2025-12-23
Replace generated code for VkLayerSettingEXT

### LINKED ISSUE #2562: VkLayerSettingEXT is not encoded or decoded correctly
Right now we encode a void* buffer but we need to be encoding and decoding the content of `pValues` based on `type`:
```
    VK_LAYER_SETTING_TYPE_BOOL32_EXT = 0,
    VK_LAYER_SETTING_TYPE_INT32_EXT = 1,
    VK_LAYER_SETTING_TYPE_INT64_EXT = 2,
    VK_LAYER_SETTING_TYPE_UINT32_EXT = 3,
    VK_LAYER_SETTING_TYPE_UINT64_EXT = 4,
    VK_LAYER_SETTING_TYPE_FLOAT32_EXT = 5,
    VK_LAYER_SETTING_TYPE_FLOAT64_EXT = 6,
    VK_LAYER_SETTING_TYPE_STRING_EXT = 7,
```

Codegen needs to blocklist the struct type from generation and implement handwritten code to encode / decode `VkLayerSettingEXT`.

fabian-lunarg:
- Replaces problematic, generated code for `VkLayerSettingEXT` with manual treatment
- closes #2562

### DIFF
--- a/framework/encode/custom_vulkan_struct_encoders.cpp
+++ b/framework/encode/custom_vulkan_struct_encoders.cpp
@@ -429,14 +429,8 @@ void EncodeStruct(ParameterEncoder* encoder, const VkLayerSettingEXT& value)
             encoder->EncodeFloat64Array(static_cast<const double*>(value.pValues), value.valueCount);
             break;
         case VK_LAYER_SETTING_TYPE_STRING_EXT:
-        {
-            auto* string_array = static_cast<const char* const*>(value.pValues);
-            for (uint32_t i = 0; i < value.valueCount; i++)
-            {
-                encoder->EncodeString(string_array[i]);
-            }
-        }
-        break;
+            encoder->EncodeStringArray(static_cast<const char* const*>(value.pValues), value.valueCount);
+            break;
         case VK_LAYER_SETTING_TYPE_MAX_ENUM_EXT:
             break;
     }
--- a/framework/decode/custom_vulkan_struct_decoders.cpp
+++ b/framework/decode/custom_vulkan_struct_decoders.cpp
@@ -599,20 +599,13 @@ size_t DecodeStruct(const uint8_t* buffer, size_t buffer_size, Decoded_VkLayerSe
             bytes_read += wrapper->pValues.DecodeDouble(buffer + bytes_read, buffer_size - bytes_read);
             break;
         case VK_LAYER_SETTING_TYPE_STRING_EXT:
-            wrapper->string_decoders.resize(value->valueCount);
-            wrapper->string_values.resize(value->valueCount);
-
-            for (uint32_t i = 0; i < value->valueCount; ++i)
-            {
-                bytes_read += wrapper->string_decoders[i].Decode(buffer + bytes_read, buffer_size - bytes_read);
-                wrapper->string_values[i] = wrapper->string_decoders[i].GetPointer();
-            }
+            bytes_read += wrapper->string_decoder.Decode(buffer + bytes_read, buffer_size - bytes_read);
             break;
         case VK_LAYER_SETTING_TYPE_MAX_ENUM_EXT:
             break;
     }
     value->pValues = value->type == VK_LAYER_SETTING_TYPE_STRING_EXT
-                         ? static_cast<const void*>(wrapper->string_values.data())
+                         ? static_cast<const void*>(wrapper->string_decoder.GetPointer())
                          : static_cast<const void*>(wrapper->pValues.GetPointer());
     return bytes_read;
 }
--- a/framework/decode/custom_vulkan_struct_decoders.h
+++ b/framework/decode/custom_vulkan_struct_decoders.h
@@ -24,6 +24,7 @@
 #ifndef GFXRECON_DECODE_CUSTOM_STRUCT_DECODERS_H
 #define GFXRECON_DECODE_CUSTOM_STRUCT_DECODERS_H
 
+#include "string_array_decoder.h"
 #include "format/platform_types.h"
 #include "decode/custom_vulkan_struct_decoders_forward.h"
 #include "decode/descriptor_update_template_decoder.h"
@@ -299,8 +300,7 @@ struct Decoded_VkLayerSettingEXT
     PointerDecoder<uint8_t> pValues;
 
     // if type is VK_LAYER_SETTING_TYPE_STRING_EXT we need to decode an array of strings
-    std::vector<StringDecoder> string_decoders;
-    std::vector<const char*>   string_values;
+    StringArrayDecoder string_decoder;
 };
 
 GFXRECON_END_NAMESPACE(decode)


framework/encode/custom_vulkan_struct_encoders.cpp
```diff
@@ -399,5 +399,43 @@ void EncodeStruct(ParameterEncoder* encoder, const VkCopyImageToMemoryInfo& valu
...
+        case VK_LAYER_SETTING_TYPE_UINT64_EXT:
+            encoder->EncodeUInt64Array(static_cast<const uint64_t*>(value.pValues), value.valueCount);
+            break;
+        case VK_LAYER_SETTING_TYPE_FLOAT32_EXT:
+            encoder->EncodeFloatArray(static_cast<const float*>(value.pValues), value.valueCount);
+            break;
+        case VK_LAYER_SETTING_TYPE_FLOAT64_EXT:
+            encoder->EncodeFloat64Array(static_cast<const double*>(value.pValues), value.valueCount);
+            break;
+        case VK_LAYER_SETTING_TYPE_STRING_EXT:
```
bradgrantham-lunarg, line 431: The way I interpret the spec is that pValues is an array of pointers to C strings.  This would only encode a byte array, wouldn't it?

bradgrantham-lunarg (changes requested):
I think this PR consistently assumes for `VK_LAYER_SETTING_TYPE_STRING_EXT` that `pValues` is a pointer to a byte array (i.e. a single string with size encoded in `valueCount`), but `pValues` for `STRING` is an array of pointers to C-strings.  Here's the test code I used as a patch to vkcube in Vulkan-Tools, which enables best practices in validation and can be tested with `vkcube --validate`:
```
diff --git a/cube/cube.c b/cube/cube.c
index d28285de..31096be0 100644
--- a/cube/cube.c
+++ b/cube/cube.c
@@ -4219,6 +4219,30 @@ static void demo_init_vk(struct demo *demo) {
         inst_info.pNext = &dbg_messenger_create_info;
     }

+    // 1. Define the configuration you want to set
+    const char* layerName = "VK_LAYER_KHRONOS_validation";
+    const char* settingName = "enables";
+    const char* settingValue = "VK_VALIDATION_FEATURE_ENABLE_BEST_PRACTICES_EXT";
+
+    // 2. Create the Layer Setting struct
+    // Note: pValues expects a pointer to the data, even if it's a single value
+    VkLayerSettingEXT layerSetting = {};
+    layerSetting.pLayerName = layerName;
+    layerSetting.pSettingName = settingName;
+    layerSetting.type = VK_LAYER_SETTING_TYPE_STRING_EXT;
+    layerSetting.valueCount = 1;
+    layerSetting.pValues = &settingValue;
+
+    // 3. Create the Layer Settings Create Info struct
+    VkLayerSettingsCreateInfoEXT layerSettingsCreateInfo = {};
+    layerSettingsCreateInfo.sType = VK_STRUCTURE_TYPE_LAYER_SETTINGS_CREATE_INFO_EXT;
+    layerSettingsCreateInfo.pNext = NULL; // Or your existing pNext chain
+    layerSettingsCreateInfo.settingCount = 1;
+    layerSettingsCreateInfo.pSettings = &layerSetting;
+
+    layerSettingsCreateInfo.pNext = inst_info.pNext;
+    inst_info.pNext = &layerSettingsCreateInfo;
+
     err = vkCreateInstance(&inst_info, NULL, &demo->inst);
     if (err == VK_ERROR_INCOMPATIBLE_DRIVER) {
         ERR_EXIT(


```
Capturing `vkcube --validate` with this change and then replaying with `gfxrecon-replay --validate` should emit similar validation information but instead segfaults.  The output of `gfxrecon-convert` shows pValues is an array with the single value [0] but should contain an array of strings with the single value `"VK_VALIDATION_FEATURE_ENABLE_BEST_PRACTICES_EXT"`.

framework/encode/custom_vulkan_struct_encoders.cpp
fabian-lunarg: yes, got this part wrong and assumed a null-terminated string

fabian-lunarg: guess it just reads a bit confusing :) - [VK_LAYER_SETTING_TYPE_STRING_EXT](https://docs.vulkan.org/refpages/latest/refpages/source/VkLayerSettingTypeEXT.html#) specifies that the layer setting’s type is **a pointer to** a null-terminated UTF-8 string.

fabian-lunarg: done, fixed the misconception. interpreting strings `pValues` not as `const char*` but as `const char* const*`.
array of strings encoding/decoding also required some additional logic.
this tests fine now with the patched vkcube example you provided. 
during replay the layer-settings are applied correctly and best-practices are enabled.

during both capture and replay I saw this deprecation-warning but was benign:
```
[gfxrecon] WARNING - DEBUG MESSENGER: Frame #1: VALIDATION-SETTINGS: vkCreateInstance(): Application is using deprecated "enables" (VK_LAYER_ENABLES) layer settings.
Deprecated settings and new settings cannot be mixed, and deprecated ones take precedence. Consider only using the new settings:
  Deprecated: VK_VALIDATION_FEATURE_ENABLE_BEST_PRACTICES_EXT | New: "validate_best_practices" (VK_LAYER_VALIDATE_BEST_PRACTICES=1)
```

fabian-lunarg:
> The output of gfxrecon-convert shows pValues is an array with the single value [0] but should contain an array of strings with the single value "VK_VALIDATION_FEATURE_ENABLE_BEST_PRACTICES_EXT"

also looking fine now:
```
{
    "pLayerName":"VK_LAYER_KHRONOS_validation",
    "pSettingName":"enables",
    "type":"VK_LAYER_SETTING_TYPE_STRING_EXT",
    "valueCount":1,
    "pValues":["VK_VALIDATION_FEATURE_ENABLE_BEST_PRACTICES_EXT"]
}
```

framework/encode/custom_vulkan_struct_encoders.cpp
bradgrantham-lunarg: Nice!

framework/encode/custom_vulkan_struct_encoders.cpp
```diff
@@ -399,5 +398,49 @@ void EncodeStruct(ParameterEncoder* encoder, const VkCopyImageToMemoryInfo& valu
...
+        {
```
bradgrantham-lunarg: There's an `EncodeStringArray` method on `encoder` - would that have worked here?

fabian-lunarg: thanks! it's still correct and much cleaner with `EncodeStringArray` and `StringArrayDecoder`.

PR #2501, opened 2025-11-07
VDR: Same BCB different QS

panos-lunarg:
Until now each BeginCommandBuffer (BCB) index was expected to be unique
in the list. If the same command buffer was submitted multiple times in
different QueueSubmits (QS), this imposed the limitation that the
requested dumped resources to be the same for all submissions.

This PR lifts this limitation however it imposes the requirement that
the BCB and QS list to have the same length.

Example of input json with the new rules:

```Json
{
  "BeginCommandBuffer": [431, 431],
  "QueueSubmit": [563, 584],
  "Draw": [
    [{ "Index": 444 }, { "Index": 451 }, { "Index": 471 }],
    [
      { "Index": 439 },
      { "Index": 444 },
      { "Index": 451 },
      { "Index": 458 },
      { "Index": 460 },
      { "Index": 469 },
      { "Index": 471 }
    ]
  ],
  "RenderPass": [
    [
      [432, 445],
      [446, 452],
      [453, 472]
    ],
    [
      [432, 445],
      [446, 452],
      [453, 472]
    ]
  ]
}
```

PR #2582, opened 2025-12-23
Correct dump resources docs regarding multiple CB submission

cmannett85-arm:
It wasn't clear previously that the queue submission index should be duplicated for command buffers submitted simultaneously.  The secondary command buffer example had fallen into this trap too, so that has been corrected.


panos-lunarg:
Good catch. The example in the .md file you fixed used to be correct until https://github.com/LunarG/gfxreconstruct/pull/2501 got merged. After that change `BefinCommandBuffer` and `QueueSubmit` lists must have the same length.

panos-lunarg (approved):

bradgrantham-lunarg (approved):

PR #2584, opened 2025-12-27
Performance clawback I

jzulauf-lunarg:
A set of changes to reclaim some performance lost by recent refactoring. One improvement per commit.

For easiest review, commit by commit is better.  PR can also be split if required.

1. Remove extra ParsedBlock copies
2. Change heapbuffer to allocate only pow2 sizes
3. Update ReadSpan to overwrite existing spans
4. Improve read performance with better return and buffering
5. Rework constructor logic to limit parameter moves
6. Remove non-standard casting accessor from util::Span
7. Use parser local decompression buffer

**gfxrecon-info performance against 3 traces of interest:** (more negative is better)

| Description | Trace A | Trace B | Trace C |
| :--- | :--- | :--- | :--- |
| **Preregression Goal** | **0.0%** | **0.0%** | **0.0%** |
| Development Baseline | +17.8% | +17.0% | +14.8% |
| Remove extra ParsedBlock copies | +8.4% | +7.1% | +2.0% |
| Change heapbuffer to allocate only pow2 sizes | +7.7% | +6.8% | +2.0% |
| Update ReadSpan to overwrite existing spans | +5.6% | +5.2% | -0.6% |
| Improve read performance with better return and buffering | -7.8% | -6.6% | -14.0% |
| Rework constructor logic to limit parameter moves | -12.3% | -9.4% | -18.0% |
| Remove non-standard casting accessor from util::Span | -12.4% | -9.6% | -18.3% |
| **Use parser local decompression buffer** | **-13.7%** | **-11.2%** | **-19.5%** |

### DIFF
--- a/framework/decode/block_parser.h
+++ b/framework/decode/block_parser.h
@@ -115,8 +115,17 @@ class BlockParser
     void
     WarnUnknownBlock(const BlockBuffer& block_buffer, const char* sub_type_label = nullptr, uint32_t sub_type = 0U);
 
+    bool ShouldDeferDecompression(size_t block_size);
+
+    // Control use of parser local storage for decompression
+    struct UseParserLocalStorageTag
+    {};
+    bool                           DecompressSpan(const BlockBuffer::BlockSpan&   compressed_span,
+                                                  size_t                          expanded_size,
+                                                  ParsedBlock::UncompressedStore& uncompressed_buffer);
     ParsedBlock::UncompressedStore DecompressSpan(const BlockBuffer::BlockSpan& compressed_span, size_t expanded_size);
-    bool                           ShouldDeferDecompression(size_t block_size);
+    const uint8_t*
+    DecompressSpan(const BlockBuffer::BlockSpan& compressed_span, size_t expanded_size, UseParserLocalStorageTag);
 
     using ErrorHandler = std::function<void(BlockIOError, const char*)>;
     BlockParser(ErrorHandler err, BufferPool& pool, util::Compressor* compressor) :
@@ -163,6 +172,8 @@ class BlockParser
 
     uint64_t frame_number_ = 0;
     uint64_t block_index_  = 0;
+
+    ParsedBlock::UncompressedStore uncompressed_working_buffer_;
 };
 
 GFXRECON_END_NAMESPACE(decode)
--- a/framework/decode/parsed_block.h
+++ b/framework/decode/parsed_block.h
@@ -202,6 +202,8 @@ class ParsedBlock
     struct DecompressedBlockTag
     {};
     static util::DataSpan MakeDecompressedBlockData(BlockBuffer& block_buffer, BlockReferencePolicy policy) noexcept;
+
+    // For owned uncompressed store
     template <typename ArgPayload>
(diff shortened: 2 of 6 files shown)


fabian-lunarg:
very cool!

looking at the numbers/table the biggest gains came from "Remove extra ParsedBlock copies"
and "Rework constructor logic to limit parameter moves", right?

artem-lunarg:
That's impressive improvements. This almost mirrors original baseline relative goal level.

framework/decode/block_buffer.h
```diff
@@ -89,22 +88,30 @@ class BlockBuffer
...
     {
         read_pos_ = 0;
         block_span_.Reset();
     };
 
-    util::DataSpan&& ReleaseData() { return std::move(block_span_); }
-    bool             SeekForward(size_t size);
-    bool             SeekTo(size_t size);
+    const util::DataSpan&          GetData() const noexcept { return block_span_; }
+    [[nodiscard]] util::DataSpan&& ReleaseData() noexcept { return std::move(block_span_); }
```
fabian-lunarg, line 98: this looks interesting. will this only release the data when the r-value ref is actually assigned to another value? (consumed by a move-ctor, which would then move the `block_span_`) - could/should this return a `util::DataSpan` by value instead? I figure it's fine because of the `[[nodiscard]]` but it was unclear to me at first

framework/decode/block_parser.h
```diff
@@ -138,17 +150,30 @@ class BlockParser
...
     constexpr static uint64_t kReadSizeFromBuffer = std::numeric_limits<std::uint64_t>::max();
     ParameterReadResult
     ReadParameterBuffer(const char* label, BlockBuffer& block_buffer, uint64_t uncompressed_size = kReadSizeFromBuffer);
-    bool                DecompressWhenParsed(const ParsedBlock& parsed_block);
-    BufferPool          pool_; // TODO: Get a better pool, and share with FileInputStream
-    ErrorHandler        err_handler_;
-    util::Compressor*   compressor_           = nullptr;
-    DecompressionPolicy decompression_policy_ = DecompressionPolicy::kAlways;
+
+    BufferPool                        pool_; // TODO: Get a better pool, and share with FileInputStream
```
fabian-lunarg, line 167: what's the problem with the pool, are we tracking this anywhere?

framework/decode/parsed_block.h
```diff
@@ -114,57 +145,108 @@ class ParsedBlock
...
+                BlockReferencePolicy policy,
+                bool                 references_block_buffer,
+                ArgPayload&&         args) :
+        block_data_(MakeIncompressibleBlockData(block_buffer, policy, references_block_buffer)),
+        uncompressed_store_(), dispatch_args_(MakeDispatchArgs(std::forward<ArgPayload>(args))), state_(kReady)
     {}
 
-    // This is called for for when blocks are parsed with deferred decompression
+    // Create a non-compressed block of a compressible block base type
+    // TODO: Is there a clean way to static assert that a block *type* is compressible here?
```
fabian-lunarg, line 191: could 'compressible' be a c++-20 concept? could/would that be an option?
not suggesting that within this PR of course, just curious

Issue #2483, opened 2025-10-27
convert option --frame-range doesn't honor frame markers


bradgrantham-lunarg:
Converting a frame range within a trimmed capture doesn't work.

```
gfxrecon-capture-vulkan.py -f 1-100 vkcube.exe
# note capture filename in the output

gfxrecon-convert --log-level warning  --format jsonl --frame-range 60-61 --output stdout {capture-filename}
# output contains frame 60's commands

GFXRECON_CAPTURE_FRAMES=50-100 gfxrecon-replay --capture {captured-filename}
# note trimmed capture filename in the output

gfxrecon-convert --log-level warning  --format jsonl --frame-range 60-61 --output stdout {trimmed-filename}
# output contains only the header
# expect to see similar output to first convert
```

antonio-lunarg:
- `--frame-range 1` simply outputs and `EndMarker:1`.
  - Expected output should be a bunch of blocks to bring Vulkan up and at least a `vkQueuePresentKHR()`.
- `--frame-range n`, where `n>1`, outputs blocks related to frame `n+1`.
  - Expected output should be blocks related to frame `n`.

Issue #1195, opened 2023-07-13
Make sure frame numbering is 0-based and consistent everywhere


davidd-lunarg:
For untrimmed files any work done before the first frame end marker (e.g., vkQueuePresentKHR) is part of frame 1. 

`CaptureManager` initializes `current_frame_(kFirstFrame)` where `kFirstFrame = 1`. 
`FileProcessor` initialized `current_frame_number_(0)`.

Fix FileProcessor to count frame 1 correctly, fixup any dependencies which call `FileProcessor::GetCurrentFrameNumber()`, and look for and fixup any other frame counts that don't start at 1.

per-mathisen-arm:
Just for the record, we would absolutely love it if gfxreconstruct switched to be zero-based and consistent everywhere instead.

per-mathisen-arm:
I just discovered that VK_LAYER_LUNARG_screenshot is zero-indexed. So you cannot match the same frames between gfxr --screnshots and this layer, which is kind of annoying even when you know about it and may generate some nice bogus bug reports in the future :-)

bradgrantham-lunarg:
I think we are tilting to switching to 0-based in the future for all user-visible indexing across the codebase, so I have edited to match.

PR #2587, opened 2026-01-06
tools/convert: accept 0-based frame ranges

antonio-lunarg:
Align convert --frame-range with 0-based frame numbering.

Refs: #1195, #2483

### DIFF
--- a/tools/convert/main.cpp
+++ b/tools/convert/main.cpp
@@ -171,7 +171,7 @@ static bool GetFrameIndices(const gfxrecon::util::ArgumentParser& arg_parser, st
     if (!input_ranges.empty())
     {
         std::vector<gfxrecon::util::UintRange> frame_ranges =
-            gfxrecon::util::GetUintRanges(input_ranges.c_str(), "frames to be converted", true);
+            gfxrecon::util::GetUintRanges(input_ranges.c_str(), "frames to be converted", true, true);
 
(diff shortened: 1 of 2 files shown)


tools/convert/main.cpp
```diff
@@ -171,7 +171,7 @@ static bool GetFrameIndices(const gfxrecon::util::ArgumentParser& arg_parser, st
     if (!input_ranges.empty())
     {
         std::vector<gfxrecon::util::UintRange> frame_ranges =
-            gfxrecon::util::GetUintRanges(input_ranges.c_str(), "frames to be converted", true);
```
MarkY-LunarG, line 174: I think this is going to have to be an option since people are expecting it to be non-zero by default right now.

antonio-lunarg: Using 1-based frame numbering for `--frame-ranges` gives wrong results anyway. https://github.com/LunarG/gfxreconstruct/issues/2483#issuecomment-3703856945

With this PR:
- `--frame-range 0` outputs the first frame correctly with `EndMarker:1`.
- `--frame-range n`, where `n>0`, outputs blocks related to frame n where `n` is a 0-based numbering with `EndMarker:n+1` since frame numbering at capture time is 1-based.

Yes, it's a bit confusing.

The alternative is to fix this the other way round, making sure that the convert tool uses 1-based numbering correctly.

PR #2588, opened 2026-01-06
capture: unleash the chaos (aka 0-based frame numbering)

antonio-lunarg:
Heads up, this is the tiny-looking change with giant consequences. We are intentionally touching the capture output path, which means:
- golden files will drift,
- baselines will scream,
- every bot will wake up at once,
- and the CI lights are about to turn a festive shade of red.

This is not a bug, it’s a feature of progress.
Please forgive the noise while the new reality settles in.

Refs: #1195, #2483

### DIFF
--- a/framework/decode/file_processor.cpp
+++ b/framework/decode/file_processor.cpp
@@ -32,13 +32,12 @@
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
-// TODO GH #1195: frame numbering should be 1-based.
 const uint32_t kFirstFrame = 0;
 
 FileProcessor::FileProcessor() :
     current_frame_number_(kFirstFrame), error_state_(kErrorInvalidFileDescriptor), bytes_read_(0),
     annotation_handler_(nullptr), compressor_(nullptr), block_index_(0), block_limit_(0),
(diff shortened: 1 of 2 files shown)


per-mathisen-arm:
So giving up on this?

antonio-lunarg:
> So giving up on this?

Whops, branch was deleted by mistake.

I'll take the opportunity to leave an update about the state of this: definitely not trivial, we'll need to still support 1-based frame numbering for backward-compatibility, but I haven't been focusing on this PR so far as most of my energies are currently on frame-looping features.

per-mathisen-arm:
We've been thinking about this from the perspective of trimming. Currently state recreation ends up in the first frame, so if you want to replay and measure the first frame, you end up measuring all of state recreation + all the normal work in the first frame. This is not ideal. So we'd like to add a frame boundary between the state recreation and the first frame, shifting frames by one, effectively making these traces zero-indexed. I wonder if it would make sense for traces to always have a zeroth frame that is normally empty until they are trimmed?

antonio-lunarg:
> @per-mathisen-arm We've been thinking about this from the perspective of trimming. Currently state recreation ends up in the first frame, so if you want to replay and measure the first frame, you end up measuring all of state recreation + all the normal work in the first frame. This is not ideal. 

As a side note, you can try looping over the first frame: state-blocks are skipped when replaying the frame again.

PR #2591, opened 2026-01-07
Fix order of code style check in CI workflow

bradgrantham-lunarg:


bradgrantham-lunarg:
Canceled the internal Jenkins CI runs because this touches only GitHub workflow runs.

fabian-lunarg:
just curious. this changes the code-style check to run after code-generation, but we don't apply/check code-style in generated files. was there another reason to change the order?

bradgrantham-lunarg:
> just curious. this changes the code-style check to run after code-generation, but we don't apply/check code-style in generated files. was there another reason to change the order?

The code-generation is also a check, which is to say the CI run fails if there are differences between what's checked in and what the generators produce, and I had a run spend 40 minutes compiling then report codegen differences and fail.  If there are codegen differences, the job should fail before compiling.  That's the reasoning behind the change.

PR #2594, opened 2026-01-09
Build and ship x86_64 ABI for Android

emrekultursay:
1. In release builds: Build and ship Android x86_64 ABI instead of the Android armeabi-v7a ABI. This should have no impact on the CI execution time.

2. In debug builds: Build the default set of ABIs, which is {arm64, x86_64} in build.gradle. This will build a second ABI, which will double the CI execution time of the AndroidDebug workflow (but it should take same time as the AndroidRelease workflow).

### DIFF
--- a/.github/workflows/release_build.yml
+++ b/.github/workflows/release_build.yml
@@ -190,9 +190,9 @@ jobs:
         if [ "${{ matrix.config.type }}" == "release" ]
         then
           sh gradlew assembleRelease -Parm64-v8a
-          sh gradlew assembleRelease -Parmeabi-v7a
+          sh gradlew assembleRelease -Px86_64
         else
-          sh gradlew assembleDebug -Parm64-v8a
+          sh gradlew assembleDebug
         fi
     - name: Prepare artifacts
       run: |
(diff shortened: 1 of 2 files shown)


bradgrantham-lunarg:
This changes what we build on GitHub CI for our release builds (which we haven't made frequently) and our SDK builds and not the GFXR capability.  This doesn't change anything regarding our per-commit CI builds on GitHub.  Do you need us to add this to our SDK builds or our CI?  Can you tell us a little more about what problem this is intended to fix?

emrekultursay:
In a LunarG/Android meeting 1-2 months back, we had decided that we want to move our ABI focus from {arm64,arm32} to {arm64,x86_64}. 

It looks like I'm missing the one for `ci_build.yml`. Let me add that one.

emrekultursay:
My overall goal is to be able to pick the Android output artifacts of GFXR from GitHub CI (both at per-commit, and at official release boundaries) and copy it elsewhere (e.g., Sherlock profiler).

bradgrantham-lunarg:
Note: Waiting on change to ci_build.yml

PR #2596, opened 2026-01-12
Implement tracking of sparse image binding

marius-pelegrin-arm:
Currently there is a "gap" in the implementation of sparse support: Tracking of the state of bindings of a sparse image.

This tracking requires intersecting the image subresource (aspectMask, array layer, mip-map level), offset and extent being bound with the former tracked bound state.

This is what this commit implements.


framework/graphics/vulkan_resources_util.cpp
```diff
@@ -2374,6 +2374,164 @@ void UpdateSparseMemoryBindMap(std::map<VkDeviceSize, VkSparseMemoryBind>& spars
...
+    // (we'll care about <x,y,z> later)
+    // The aspect mask is a bit particular as it is a bit field...
+
+    std::vector<VkImageSubresource> intersecting_subresources;
+
+    VkImageSubresource search_key = new_bind.subresource;
+    search_key.aspectMask         = 0;
+
+    auto iterator = sparse_image_memory_bind_map.lower_bound(search_key);
+    if (iterator != sparse_image_memory_bind_map.begin())
```
MarkY-LunarG: I don't understand why you don't want the begin()?  If the first map binding is a match, isn't that ok?

framework/graphics/vulkan_resources_util.cpp
```diff
@@ -2374,6 +2374,164 @@ void UpdateSparseMemoryBindMap(std::map<VkDeviceSize, VkSparseMemoryBind>& spars
...
+        sparse_image_memory_bind_map[new_bind.subresource][new_bind.offset] = new_bind;
+        return;
+    }
+
+    // Otherwise, go over the subresources that needs to be updated
+    for (VkImageSubresource subresource : intersecting_subresources)
+    {
+        // We may need to split the subresource depending on the aspectMask
+        VkImageAspectFlags aspect_mask = (subresource.aspectMask & new_bind.subresource.aspectMask);
+        if (aspect_mask != subresource.aspectMask)
```
MarkY-LunarG, line 2494: It seems like aspect_mask could theoretically be 0.  Which would pass here but attempt a re-assign  in the later part of the code below because the original map would still be valid (or you'd overwrite it).  Maybe add an assert or a non-zero check?

framework/graphics/vulkan_resources_util.cpp
```diff
@@ -2374,6 +2374,164 @@ void UpdateSparseMemoryBindMap(std::map<VkDeviceSize, VkSparseMemoryBind>& spars
...
+        (old_offset.y + old_extent.height <= new_offset.y || new_offset.y + new_extent.height <= old_offset.y) ||
+        (old_offset.z + old_extent.depth <= new_offset.z || new_offset.z + new_extent.depth <= old_offset.z))
+    {
+        return;
+    }
+
+    // If there is an intersection, no matter the type, the old binding will have to be removed
+    removed_binds.push_back(old_offset);
+
+    // Check for each possible "slice" of the old 3D rectangular binding if the binding is overriden or must be kept.
```
MarkY-LunarG, line 2398: Couldn't we have a case that the update is a sub-region in an existing region and we wouldn't need to create another new binding?

framework/graphics/vulkan_resources_util.cpp
marius-pelegrin-arm: Yes, you're right, I mixed things up. std::map::lower_bound always confuse my mind to the highest degree...

framework/graphics/vulkan_resources_util.cpp
marius-pelegrin-arm: According to the spec:
```
VUID-VkImageSubresource-aspectMask-requiredbitmask
aspectMask must not be 0
```

framework/graphics/vulkan_resources_util.cpp
marius-pelegrin-arm: I am not sure to fully understand what you mean but:
- As we call a single `vkQueueBindSparse` over all the reconstructed binds, we don't want overlapping ranges in the bind map, as I don't think `vkQueueBindSparse` gives a guarantee on the binding order, and this would result in an undefined behavior.
- It's possible that the new bind matches perfectly with an existing bind, in this case the bind is just removed and re-inserted.

framework/graphics/vulkan_resources_util.cpp
marius-pelegrin-arm: In addition, we are going over subresources in `intersecting_subresources`, in which elements are added only if the intersection (the same used to compute `aspect_mask`) is non-zero.

So in addition to the guarantee given by the spec, the code path is just not possible.

framework/graphics/vulkan_resources_util.cpp
marius-pelegrin-arm: I pushed a commit to fix that

framework/graphics/vulkan_resources_util.cpp
MarkY-LunarG: My point is that each binding defines a square region in the texture space.  So what if we just defined a region that is already a sub-square inside a different region?  Do we need any kind of checking for that?

marius-pelegrin-arm: No, rectangles added to the binding map don't intersect with other rectangles. We guarantee that because when there is an intersection, this function is called and make sure that all new rectangles added are within the boundaries of the previous binding, don't overlap, and the previous binding is removed.

You can have overlapping rectangles but only on different subresources (aspect mask, mip level, array layer) which is perfectly fine.

fabian-lunarg:
this looks good to me. what would a minimal test-case look like?

I assume since this implements/fixes state-tracking we would need to capture using trimming and
test for the correct sparse-binding info after state-init. 
'could' this be done via `gfxrecon-convert` and don't require a replay-based test?
how did you test this @marius-pelegrin-arm ?

PR #2598, opened 2026-01-13
Add and use pack_luid() function

locke-lunarg:
```
static inline int64_t pack_luid(LUID luid)
{
    return (static_cast<uint64_t>(luid.HighPart) << 32) | luid.LowPart;
}
```
Use `pack_luid` to replace the duplicated code.

### DIFF
--- a/framework/format/format.h
+++ b/framework/format/format.h
@@ -665,7 +665,7 @@ struct DxgiAdapterDesc
 
 inline int64_t pack_luid(const format::DxgiAdapterDesc& adapter_desc)
 {
-    return (static_cast<uint64_t>(adapter_desc.LuidHighPart) << 32) | adapter_desc.LuidLowPart;
+    return static_cast<int64_t>((static_cast<uint64_t>(adapter_desc.LuidHighPart) << 32) | adapter_desc.LuidLowPart);
 }
 
 struct DxgiAdapterInfoCommandHeader
--- a/framework/format/platform_types.h
+++ b/framework/format/platform_types.h
@@ -163,7 +163,7 @@ struct LARGE_INTEGER
 
 static inline int64_t pack_luid(LUID luid)
 {
-    return (static_cast<uint64_t>(luid.HighPart) << 32) | luid.LowPart;
+    return static_cast<int64_t>((static_cast<uint64_t>(luid.HighPart) << 32) | luid.LowPart);
 }
 
 #ifndef __ANDROID__


framework/format/format.h
```diff
@@ -663,6 +663,11 @@ struct DxgiAdapterDesc
     uint32_t extra_info; // 2 bits (LSB) to store Type and 30 bits for object ID
 };
 
+inline int64_t pack_luid(const format::DxgiAdapterDesc& adapter_desc)
+{
+    return (static_cast<uint64_t>(adapter_desc.LuidHighPart) << 32) | adapter_desc.LuidLowPart;
```
MarkY-LunarG: You're casting to `uint64_t` but returning `int64_t`, won't some compilers have issues with this?  I would think we would want this to be `uint64_t` anyway.

locke-lunarg: Thank you. Fixed it.

mattst88: >  You're casting to uint64_t but returning int64_t, won't some compilers have issues with this? I would think we would want this to be uint64_t anyway.

That's totally fine.

AFAIU, LUID is int64_t. You can see this from definitions like

```c++
const int64_t adapter_luid = ...
```

but also from the definitions of `LuidLowPart` and `LuidHighPart`:

```c++
struct DxgiAdapterDesc
{
[...]
    uint32_t LuidLowPart;
    int32_t  LuidHighPart;
[...]
};
```

`LuidHighPart` is `int32_t`.

I assume the confusion that led to the `<< 31` was because some compiler (MSVC?) warned about shifting into the sign bit when they shifted by 32.

I think the code was correct as it is because: you need to shift `LuidHighPart` by 32; in order to do that you must first cast to `uint64_t`. Casting to `int64_t` means you're shifting into the sign bit, which is undefined behavior AFAIK.

In any case, the implicit cast from `uint64_t` to `int64_t` (the return type) is fine and totally expected.

ianromanick: Literally every call site assigns the value to an `int64_t`. How is returning a type other than that better?

MarkY-LunarG: I'm ok with it being `int64_t` (although that does seem  strange to me, shrug, just make it consistent and clean.

mattst88: heh, I explain the whole thing, why it's right as is, why we know luids are int64, why we need to cast to uint64_t and you begrudgingly say it's okay with you to leave it alone.

locke-lunarg: I added `static_cast<int64_t>` to avoid confusing and warning that might happen in some compilers for signed/unsigned mismatch.

MarkY-LunarG (approved):

framework/format/format.h
bradgrantham-lunarg: > heh, I explain the whole thing, why it's right as is, why we know luids are int64, why we need to cast to uint64_t and you begrudgingly say it's okay with you to leave it alone.

@mattst88 , the work on these PRs you filed are outside paying contracts and we are spending time at our discretion to land them separate from our client priorities.  We do want the code quality to improve and I appreciate your contribution, but remember that software engineers are human.  Please be cordial and choose helpful and positive words and emojis.

@locke-lunarg Matt has expressed confusion about your static_cast.  Can you give detail about why that was necessary?

mattst88: > @mattst88 , the work on these PRs you filed are outside paying contracts and we are spending time at our discretion to land them separate from our client priorities. We do want the code quality to improve and I appreciate your contribution, but remember that software engineers are human. Please be cordial and choose helpful and positive words and emojis.

Are you saying the (confused emoji) is ... non-cordial?

As to the rest of your message... lordy. I'll remind you that *I* was spending *my* time fixing issues in your project and you told me it wasn't worth the CI costs. Then you asked *me* to help land the code, with the pseudo-threat of reverting my patch that already landed independently. That's the only reason I'm involved here in this PR.

> @locke-lunarg Matt has expressed confusion about your static_cast. Can you give detail about why that was necessary?

It's not necessary. It's also not important. Y'all do whatever you like. I just tried to explain why the original code was completely fine as is.

locke-lunarg: @mattst88 Sorry for misunderstood. It's not a threat. We just prefer to keep your code in one PR, not separate into two. It helps people to understand this topic easily if they need. Thank you for your contribution. 

For `static_cast<int64_t>`, it allows us to suppress signed/unsigned mismatch warnings with zero runtime cost, while explicitly documenting that we are intentionally treating this ID as a signed type.

PR #2599, opened 2026-01-13
Expose api to externally trigger trimming

panos-lunarg:
Two new functions are implemented and should be discoverable through dlopen+dlsym:

* GFXRSetTrimmingState(bool): Controls enabling/disabling trimming. For this to work trimming must be enabled either by setting a hot key (desktop) or by enabling the android capture trigger option (debug.gfxrecon.capture_android_trigger)

* GFXRGetCaptureMode() Returns the current capture mode GFXR as an uint32_t:
- 0 (disabled): GFXR is disabled

- 1 (write): Writing mode. This state means that GFXR is in full capture mode (not trimming)

- 2 (track): GFXR is in tracking mode. Trimming can be enabled by either pressing the hot key (or setting the android trigger option) or by calling GFXRSetTrimmingState(true)

- 3 (write and track): GFXR is writing a trimmed capture. Capturing can be interrupted by either pressing again the hot key (or setting the android trigger option appropriately) or by calling GFXRSetTrimmingState(false)


wpaul-samsung:
I set trimming signal callbacks with SetSignalTrimmingCallbacks().  For a frame range capture, I get the expected results.  The callbacks happen at the expected frame.  For a manual capture, where we set debug.gfxrecon.capture_android_trigger to true to start the capture and to false when we want to stop capturing, I seem to get the start signal at the expected frame, but the end signal comes back at the very next frame, even if I capture 100 frames.

wpaul-samsung:
> I set trimming signal callbacks with SetSignalTrimmingCallbacks(). For a frame range capture, I get the expected results. The callbacks happen at the expected frame. For a manual capture, where we set debug.gfxrecon.capture_android_trigger to true to start the capture and to false when we want to stop capturing, I seem to get the start signal at the expected frame, but the end signal comes back at the very next frame, even if I capture 100 frames.

This works as expected now

bradgrantham-lunarg:
@fabian-lunarg can you give this a lookover?

framework/util/callbacks.cpp
```diff
@@ -0,0 +1,137 @@
...
+*/
+
+#include "callbacks.h"
+#include "encode/capture_manager.h"
+#include "util/logging.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(util)
+
+class CallbackBaseClass
```
fabian-lunarg: just call it `CallbackBase`?

framework/encode/capture_manager.cpp
```diff
@@ -65,6 +66,28 @@ uint64_t                                   CommonCaptureManager::default_unique_
 thread_local bool                          CommonCaptureManager::force_default_unique_id_  = false;
 thread_local std::vector<format::HandleId> CommonCaptureManager::unique_id_stack_;
 
+static bool externally_set_trimming_state_          = false;
+static bool previous_externally_set_trimming_state_ = false;
```
fabian-lunarg: is this guaranteed to be manipulated only by a single thread?
would it be possible or and make sense to use `std::atomic<bool>`?

framework/encode/capture_manager.cpp
```diff
@@ -828,6 +856,32 @@ bool CommonCaptureManager::RuntimeTriggerDisabled()
...
+{
+    if (!previous_externally_set_trimming_state_ && externally_set_trimming_state_)
+    {
+        previous_externally_set_trimming_state_ = externally_set_trimming_state_;
+        return true;
+    }
+    else
+    {
+        return false;
+    }
```
fabian-lunarg: redundant else{}

framework/util/callbacks.h
```diff
@@ -24,25 +24,37 @@
...
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
-GFXRECON_BEGIN_NAMESPACE(decode)
+GFXRECON_BEGIN_NAMESPACE(util)
 
-using PFN_BeginInjectedCommands       = void (*)(void*);
-using PFN_EndInjectedCommands         = void (*)(void*);
-using PFN_SetInjectedCommandCallbacks = void (*)(PFN_BeginInjectedCommands, PFN_EndInjectedCommands, void*);
+using PFN_EventACallBack     = void (*)(void*);
+using PFN_EventBCallBack     = void (*)(void*);
+using PFN_SetEventsCallbacks = void (*)(PFN_EventACallBack, PFN_EventBCallBack, void*);
```
fabian-lunarg: why not call those `PFN_EventBeginCallBack` and `PFN_EventEndCallBack`?
the argument-names for SetInjectedCommandCallbacks are also still `begin_fp` and `end_fp`.
I understand the callbacks became more generic than before (`PFN_BeginInjectedCommands`),
but it's still a start/end pair, right?

test/test_apps/trigger-trimming/app.cpp
```diff
@@ -0,0 +1,417 @@
...
+
+// Structure passed down to callback mechanism
+struct FrameData
+{
+    size_t                  current_frame{ 0 };
+    GetCaptureModeGFXR_fp_t GetCaptureModeGFXR_fp{ nullptr };
+};
+
+// Callback for trim start
+static void EventACallback(void* data)
```
fabian-lunarg: same comment as above, I think `EventStartCallback` would be a more descriptive name

framework/util/callbacks.cpp
panos-lunarg: sure

framework/encode/capture_manager.cpp
panos-lunarg: Good idea. I used a `mutex` instead.

framework/encode/capture_manager.cpp
panos-lunarg: fixed

framework/util/callbacks.h
panos-lunarg: sure

test/test_apps/trigger-trimming/app.cpp
panos-lunarg: done

framework/encode/capture_manager.cpp
fabian-lunarg: aah k, hold on. the mutex is fine, you're manipulating both variables under that scope. atomic would be wrong

PR #2601, opened 2026-01-14
Correct offset-calculation for nested structs in spirv_parsing_util 

fabian-lunarg:
- struct-offset calculation was plain omitted and now implemented here
- fix to matrix size calculations, correcting usage of stride and col/row majority
- de-duplicate offset calculation
- extend existing unit-test to include a check for the added/fixed functionality

### DIFF
--- a/framework/decode/custom_vulkan_struct_decoders.cpp
+++ b/framework/decode/custom_vulkan_struct_decoders.cpp
@@ -582,31 +582,59 @@ size_t DecodeStruct(const uint8_t* buffer, size_t buffer_size, Decoded_VkLayerSe
         case VK_LAYER_SETTING_TYPE_BOOL32_EXT:
         case VK_LAYER_SETTING_TYPE_UINT32_EXT:
             bytes_read += wrapper->pValues.DecodeUInt32(buffer + bytes_read, buffer_size - bytes_read);
+            value->pValues = static_cast<const void*>(wrapper->pValues.GetPointer());
             break;
         case VK_LAYER_SETTING_TYPE_INT32_EXT:
             bytes_read += wrapper->pValues.DecodeInt32(buffer + bytes_read, buffer_size - bytes_read);
+            value->pValues = static_cast<const void*>(wrapper->pValues.GetPointer());
             break;
         case VK_LAYER_SETTING_TYPE_INT64_EXT:
             bytes_read += wrapper->pValues.DecodeInt64(buffer + bytes_read, buffer_size - bytes_read);
+            value->pValues = static_cast<const void*>(wrapper->pValues.GetPointer());
             break;
         case VK_LAYER_SETTING_TYPE_UINT64_EXT:
             bytes_read += wrapper->pValues.DecodeUInt64(buffer + bytes_read, buffer_size - bytes_read);
+            value->pValues = static_cast<const void*>(wrapper->pValues.GetPointer());
             break;
         case VK_LAYER_SETTING_TYPE_FLOAT32_EXT:
             bytes_read += wrapper->pValues.DecodeFloat(buffer + bytes_read, buffer_size - bytes_read);
+            value->pValues = static_cast<const void*>(wrapper->pValues.GetPointer());
             break;
         case VK_LAYER_SETTING_TYPE_FLOAT64_EXT:
             bytes_read += wrapper->pValues.DecodeDouble(buffer + bytes_read, buffer_size - bytes_read);
+            value->pValues = static_cast<const void*>(wrapper->pValues.GetPointer());
             break;
         case VK_LAYER_SETTING_TYPE_STRING_EXT:
-            bytes_read += wrapper->string_decoder.Decode(buffer + bytes_read, buffer_size - bytes_read);
+        {
+            uint32_t attrib;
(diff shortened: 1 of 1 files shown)


framework/util/spirv_parsing_util.cpp
```diff
@@ -38,6 +39,72 @@ bool operator<(const SpirVParsingUtil::BufferReferenceInfo& lhs, const SpirVPars
...
 
+struct LayoutInfo
+{
+    uint32_t alignment = 0;
+    uint32_t size      = 0;
+};
+
+LayoutInfo compute_type_layout(const SpvReflectTypeDescription* type_description)
+{
+    uint32_t alignment = 0;
```
antonio-lunarg, line 51: This is returned but never modified.

framework/util/spirv_parsing_util.cpp
```diff
@@ -38,6 +39,72 @@ bool operator<(const SpirVParsingUtil::BufferReferenceInfo& lhs, const SpirVPars
...
+    switch (type_description->op)
+    {
+        case SpvOpTypeVector:
+            num_bytes *= type_description->traits.numeric.vector.component_count;
+            break;
+
+        case SpvOpTypeMatrix:
+            num_bytes *= type_description->traits.numeric.matrix.column_count;
+            num_bytes *= type_description->traits.numeric.matrix.row_count;
+            num_bytes = std::max(num_bytes, type_description->traits.numeric.matrix.stride);
```
antonio-lunarg: What is the meaning of stride in this context? If it is the number of bytes between columns, then this calculation might be wrong.

fabian-lunarg: I think you're right. stride is the number of bytes per column.

it should be:
`num_bytes = type_description->traits.numeric.matrix.column_count * type_description->traits.numeric.matrix.stride;`

framework/util/spirv_parsing_util.cpp
fabian-lunarg: there are still some things not considered correctly. we have to extract knowledge of the block-layout (std140, std430, scalar) and pass that into here. then we would treat e.g. a `vec3` or an `uint[12]` with different alignments. so ... we will need this eventually and I should add a clarifying TODO comment

framework/util/spirv_parsing_util.cpp
antonio-lunarg: Well, according to [the spec](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html)
> MatrixStride
Applies only to a member of a structure type. Only valid on a matrix or array whose most basic element is a matrix. Matrix Stride is an unsigned 32-bit integer specifying the stride of the rows in a RowMajor-decorated matrix or columns in a ColMajor-decorated matrix.

So there should be a check to see whether the matrix is row-major or col-major.

antonio-lunarg: Maybe:
```cpp
if ((type_description->decoration_flags & SPV_REFLECT_DECORATION_ROW_MAJOR) != 0) {
  num_bytes = type_description->traits.numeric.matrix.row_count * type_description->traits.numeric.matrix.stride;
} else if ((type_description->decoration_flags & SPV_REFLECT_DECORATION_COLUMN_MAJOR) != 0) {
  num_bytes = type_description->traits.numeric.matrix.column_count * type_description->traits.numeric.matrix.stride;
} else {
  // ehm... not sure, default to column-major?
}

Comment (fabian-lunarg) [reply framework/util/spirv_parsing_util.cpp]: ok, great. it's ... complicated :)

in spirv_reflect.h:
```
/*! @enum SpvReflectDecorationBits

NOTE: HLSL row_major and column_major decorations are reversed
      in SPIR-V. Meaning that matrices declrations with row_major
      will get reflected as column_major and vice versa. The
      row and column decorations get appied during the compilation.
      SPIRV-Reflect reads the data as is and does not make any
      attempt to correct it to match what's in the source.
```

fabian-lunarg: so I have to check for row/col major and also check if it's SpirV generated from GLSL or HLSL (what about slang?) and prob. invert cols<->rows accordingly I guess

antonio-lunarg: Also relevant: https://github.com/KhronosGroup/SPIRV-Reflect/blob/main/spirv_reflect.c#L2664

fabian-lunarg: done, added a check for `SPV_REFLECT_DECORATION_COLUMN_MAJOR` (and ROW_MAJOR) decorations.
I don't see an immediate way to figure out if the spirv was compiled from HLSL though

framework/util/spirv_parsing_util.cpp
fabian-lunarg: done, left a TODO stating the intended purpose and goal

framework/util/spirv_parsing_util.cpp
antonio-lunarg: This solution is good enough IMO

fabian-lunarg:
thanks @antonio-lunarg !

MarkY-LunarG (approved):

PR #2604, opened 2026-01-15
Disable `offscreen_swapchain_frame_boundary` swapchain option when `VK_EXT_frame_boundary` is not supported

### LINKED ISSUE #1467: VK_EXT_frame_boundary not enabled during replay when --offscreen-swapchain-frame-boundary is enabled
The --offscreen-swapchain-frame-boundary option inserts VkFrameBoundaryEXT structs to VkSubmitInfo without first checking if the device supports the VK_EXT_frame_boundary extension. This is invalid API usage.

nickdriscoll-lunarg:
Resolves #1467 

This PR adds logic to disable the `offscreen_swapchain_frame_boundary` swapchain option if `VK_EXT_frame_boundary` is unsupported by the replay device.

A method is added to allow retrieving a copy of the swapchain options in order to facilitate modifying the existing option set.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -3277,6 +3277,9 @@ void VulkanReplayConsumerBase::ModifyCreateDeviceInfo(
         // Fake VK_EXT_frame_boundary if requested, but not supported
         if (sanitize_faked_extension(VK_EXT_FRAME_BOUNDARY_EXTENSION_NAME))
         {
+            VulkanSwapchainOptions modified_options = swapchain_->GetOptions();
+            modified_options.offscreen_swapchain_frame_boundary = false;
+            swapchain_->SetOptions(modified_options);
             // also remove related feature-struct from pnext-chain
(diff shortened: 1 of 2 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -3236,6 +3236,9 @@ void VulkanReplayConsumerBase::ModifyCreateDeviceInfo(
         // Fake VK_EXT_frame_boundary if requested, but not supported
         if (sanitize_faked_extension(VK_EXT_FRAME_BOUNDARY_EXTENSION_NAME))
         {
+            VulkanSwapchainOptions modified_options             = swapchain_->GetOptions();
```
MarkY-LunarG: Nit: Add a comment to the effect of what you're doing (what you have in the commit header is good enough).

MarkY-LunarG:
@marius-pelegrin-arm , FYI.

per-mathisen-arm:
Please hard abort on failed options. Just running with different options than the user selected can generate wrong data and wrong test results, and is a very bad idea.

nickdriscoll-lunarg:
The logic has been changed so that replay will abort if the option is selected while the extension is unavailable on the replay device.

per-mathisen-arm:
Thank you!

bradgrantham-lunarg (approved):

PR #2605, opened 2026-01-15
detect bad previous VkLayerSettingEXT encoding and ignore

### LINKED ISSUE #2603: #2581 changed the encoding of layer settings struct, invalidating some pervious captures
A capture before 23f420bdecd52232a9640a3ad8cf2ea96946f8d2 that contains a VkLayerSettingsEXT struct will have encoded that struct incorrectly but the tree before that commit also decoded the struct into a compatible (wrong) representation.  As of 23f420bdecd52232a9640a3ad8cf2ea96946f8d2, VkLayerSettingsEXT is encoded and decoded correctly but will decode old encodings incorrectly and possibly crash.

bradgrantham-lunarg:
Fixes #2603

### DIFF
--- a/framework/decode/custom_vulkan_struct_decoders.cpp
+++ b/framework/decode/custom_vulkan_struct_decoders.cpp
@@ -582,31 +582,59 @@ size_t DecodeStruct(const uint8_t* buffer, size_t buffer_size, Decoded_VkLayerSe
         case VK_LAYER_SETTING_TYPE_BOOL32_EXT:
         case VK_LAYER_SETTING_TYPE_UINT32_EXT:
             bytes_read += wrapper->pValues.DecodeUInt32(buffer + bytes_read, buffer_size - bytes_read);
+            value->pValues = static_cast<const void*>(wrapper->pValues.GetPointer());
             break;
         case VK_LAYER_SETTING_TYPE_INT32_EXT:
             bytes_read += wrapper->pValues.DecodeInt32(buffer + bytes_read, buffer_size - bytes_read);
+            value->pValues = static_cast<const void*>(wrapper->pValues.GetPointer());
             break;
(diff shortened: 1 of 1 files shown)


framework/decode/custom_vulkan_struct_decoders.cpp
```diff
@@ -582,31 +582,59 @@ size_t DecodeStruct(const uint8_t* buffer, size_t buffer_size, Decoded_VkLayerSe
...
+                bytes_read += wrapper->string_decoder.Decode(buffer + bytes_read, buffer_size - bytes_read);
+                value->pValues = static_cast<const void*>(wrapper->string_decoder.GetPointer());
+            }
+            else
+            {
+                GFXRECON_LOG_INFO("Detected and discarding VkLayerSettingEXT string value incorrectly encoded as "
+                                  "void*. Settings will not be passed downstream. See commit 23f420bd.");
+                // XXX Note that it is unlikely the encoded buffer (1 uint8_t) will contain the captured layer settings
+                // settings.  Decode a uint8_t buffer to move past the data and discard the results.
+                bytes_read += wrapper->pValues.DecodeVoid((buffer + bytes_read), (buffer_size - bytes_read));
```
fabian-lunarg, line 629: I think the mismatch between previous encoded `void*` and an array
is not restricted to `VK_LAYER_SETTING_TYPE_STRING_EXT` alone.
meaning, the check would need to occur before the `switch`, handling types other than strings as well.

bradgrantham-lunarg: I think this PR just restores the previous behavior, which was that the encoded content was wrong but does not crash, since a pointer won't be followed for TYPE_STRING.  

I have to close down for the night, but if you want to take a stab at it you could PR my PR or you could make a new PR based on this one!

fabian-lunarg (approved):

PR #2606, opened 2026-01-16
Performance Clawback II: Convert file input stream to read from fread

jzulauf-lunarg:
Improved performance by locally managing a read ahead buffer and not using the fstream managed buffer.

### DIFF
--- a/framework/util/alignment_utils.h
+++ b/framework/util/alignment_utils.h
@@ -24,7 +24,6 @@
 #define GFXRECON_UTIL_ALIGNED_VALUE_H
 
 #include "util/defines.h"
-#include "util/logging.h"
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(util)
@@ -73,7 +72,7 @@ inline constexpr uint64_t next_pow_2(uint64_t v)
  */
 inline constexpr uint64_t aligned_value(uint64_t value, uint64_t alignment)
 {
-    GFXRECON_ASSERT(is_pow_2(alignment));
+    assert(is_pow_2(alignment));
     return alignment ? (value + alignment - 1) & ~(alignment - 1) : value;
 }
 
@@ -86,7 +85,7 @@ inline constexpr uint64_t aligned_value(uint64_t value, uint64_t alignment)
  */
 inline constexpr uint32_t div_up(uint32_t nom, uint32_t denom)
 {
-    GFXRECON_ASSERT(denom > 0);
+    assert(denom > 0);
     return (nom + denom - 1) / denom;
 }
 
--- a/framework/util/file_input_stream.cpp
+++ b/framework/util/file_input_stream.cpp
@@ -32,8 +32,44 @@
 #include "util/file_input_stream.h"
 #include "util/logging.h"
 
+#include <cstdlib>
+#include <new>
+
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(util)
+
+bool FStreamFileInputStream::IsOpen() const
+{
+#if FILE_INPUT_STREAM_USE_FREAD
+    return fd_ != nullptr;
+#else
+    return fd_ >= 0;
+#endif
+}
+
+bool FStreamFileInputStream::IsEof() const
+{
+    const bool can_be_eof = IsOpen() && !HasReadAhead();
+#if FILE_INPUT_STREAM_USE_FREAD
+    return can_be_eof && (feof(fd_) != 0);
+#else
+    return can_be_eof && (last_read_status_ == util::platform::FileReadStatus::kEof);
+#endif
+}
+
+bool FStreamFileInputStream::IsError() const
+{
+#if FILE_INPUT_STREAM_USE_FREAD
+    return !HasReadAhead() && IsOpen() && (ferror(fd_) != 0);
+#else
+    return !HasReadAhead() && (last_read_status_ == util::platform::FileReadStatus::kError);
+#endif
+}
+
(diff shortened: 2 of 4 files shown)


framework/util/platform.h
```diff
@@ -817,6 +914,11 @@ inline int32_t FileClose(FILE* stream)
     return fclose(stream);
 }
 
+inline int32_t FileClose(int fd)
+{
+    return close(fd);
```
antonio-lunarg: Should we use `_close(fd)` on Windows?

framework/util/file_input_stream.h
```diff
@@ -34,24 +34,32 @@
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(util)
 
+// By default don't use FSTREAMS but leave support for
+// platforms with no POSIX compatibility
+#define FILE_INPUT_STREAM_USE_FREAD 0
```
fabian-lunarg, line 39: when and how would we overwrite/enable this, did you think of e.g. a cmake-option @jzulauf-lunarg ?

framework/util/alignment_utils.h
```diff
@@ -73,7 +72,7 @@ inline constexpr uint64_t next_pow_2(uint64_t v)
  */
 inline constexpr uint64_t aligned_value(uint64_t value, uint64_t alignment)
 {
-    GFXRECON_ASSERT(is_pow_2(alignment));
+    assert(is_pow_2(alignment));
```
MarkY-LunarG, line 75: Why convert these to standard  `assert`s?

framework/util/file_input_stream.h
```diff
@@ -34,24 +34,32 @@
...
+
 class FStreamFileInputStream
 {
   public:
     using BufferPool    = HeapBufferPool;
     using BufferPoolPtr = BufferPool::PoolPtr;
 
+#if FILE_INPUT_STREAM_USE_FREAD
```
MarkY-LunarG, line 47: I also think it would be nice if there was an `info` level message saying which path is used.  It could be useful for debug in the future.  I.e. `GFXRECON_LOG_INFO_ONCE("Using FREAD for file input");` or even the "not" message.

framework/util/file_input_stream.h
```diff
@@ -67,24 +75,41 @@ class FStreamFileInputStream
     // as useful for performance if using fread.  If we go larger we should
     // go to read/_read system calls, directly and avoid the fread buffer
     // entirely.
+#if FILE_INPUT_STREAM_USE_FREAD
     constexpr static size_t kReadAheadBufferSize = 1024U;
+#else
+    constexpr static size_t        kReadAheadAlignment  = 4096U;
+    constexpr static size_t        kReadAheadBufferSize = 2U * kReadAheadAlignment;
```
MarkY-LunarG, line 82: Is this lined up properly for clang-format?

framework/util/file_input_stream.h
```diff
@@ -67,24 +75,41 @@ class FStreamFileInputStream
...
+#endif
 
     // Design assumes kMaxPeekBytes << kReadAheadBufferSize, as we move data when peeking would
     // exceed the read ahead buffer size, and we want to avoid moving large amounts of data.
```
MarkY-LunarG, line 86: TODO: Make this configurable?

framework/util/alignment_utils.h
jzulauf-lunarg: Include file ordering,  since this is now included in platform wouldn't allow including logging.h here.

framework/util/file_input_stream.cpp
```diff
@@ -42,8 +45,31 @@ bool FStreamFileInputStream::Open(const std::string& filename)
...
+#endif
+
+    fd_                = util::platform::FileOpenFd(filename.c_str(), mode);
+    const bool success = fd_ >= 0;
+    if (success)
+    {
+        last_read_status_ = util::platform::FileReadStatus::kSuccess;
+        GFXRECON_ASSERT(read_ahead_buffer_ == nullptr);
+        read_ahead_buffer_ =
+            static_cast<char*>(util::platform::AlignedAlloc(kReadAheadBufferSize, kReadAheadAlignment));
```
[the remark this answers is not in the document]
jzulauf-lunarg: Added error check for allocation failed

framework/util/file_input_stream.h
jzulauf-lunarg: If we find a platform specific need (some platform that has fread but not read) we could.  This is included to allow short term A/B testing (regression checks), if people are agreeable we can eliminate the fread path entirely at some point.

framework/util/file_input_stream.h
jzulauf-lunarg: Medium term the fread path should probably be eliminated to avoid bit rot.

framework/util/file_input_stream.h
jzulauf-lunarg: Yes.  It's a quirk of 14.7, which is what github CI uses.  Clang latest doesn't do this.

framework/util/file_input_stream.h
jzulauf-lunarg: constexpr gives the compiler the best chance at optimization

This is real reflective of programmatic use.  Currently we peek 8 bytes max, I can't imagine a use beyond this, and for performance reasons you want a peek size << buffer size.  Data movement is restricted to MaxPeek -1, but is practically tied to max bytes peeked -1, but I **_want_** anyone peeking more than a trivial amount to rethink their approach and minimize peek size. Was tempted to set this to 8, but didn't want to over constrain use.

But I can't conceive of a use to peak more than four QWORDS.  I mean, you don't a hide packet size information more than 24 bytes into a header, sensibly.

PR #2609, opened 2026-01-19
VulkanAddressReplacer: Identify/Handle chained pointer accesses

fabian-lunarg:
- Identify chained accesses by detecting duplicated (set/binding/offset) tuples.
- Add VulkanAddressReplacer::ResolveBufferAddresses routine
- use for: deferred read-back of memory-locations from host-visible buffers
- Record pointed-to locations for replacement
- Add unit-testcase for SpirVParsingUtil

This PR improves the capabilities of `VulkanAddressReplacer` and allows detecting and resolving 
shader pointer-chains to some degree. 
This is relevant when replaying captures using pointer-chaining and using `-m rebind`. 

e.g. patterns like this can now be resolved:
```
glsl:

// array of vertex-buffers
layout(buffer_reference, scalar) readonly buffer VertexBuffer { vertex_t v[]; };
layout(buffer_reference, scalar) readonly buffer VertexBufferArray { VertexBuffer v[]; };
...

layout(std140) uniform UBO
{
    camera_t camera;
    ...

    VertexBufferArray vertex_buffers; // holds an array of pointers (vertex_t*[])
} data;

void main()
{
    ...
    uint buffer_index, vertex_offset; // from somewhere else
    vertex_t vertex = data.vertex_buffers.v[buffer_index].v[vertex_offset];
    ...
}
```

basically a deferred read-back + dereferencing of `vertex_buffers` is performed before queue-submit, 
allowing to identify the buffer and record contained buffer-device-addresses for remapping/replacement.

### DIFF
--- a/framework/decode/vulkan_address_replacer.cpp
+++ b/framework/decode/vulkan_address_replacer.cpp
@@ -477,6 +477,7 @@ void VulkanAddressReplacer::ResolveBufferAddresses(VulkanCommandBufferInfo*     
         {
             GFXRECON_LOG_WARNING_ONCE("%s: is required for buffer-handle: %d -- but currently only implemented for "
                                       "host-visible buffers. replay is likely going to fail.",
+                                      __func__,
                                       buffer_info->capture_id);
             continue;
         }
@@ -490,8 +491,10 @@ void VulkanAddressReplacer::ResolveBufferAddresses(VulkanCommandBufferInfo*     
         {
             for (const auto& [offset, stride] : offset_pairs)
             {
+                // read an address from mapped pointer
                 VkDeviceAddress buffer_address =
                     *reinterpret_cast<VkDeviceAddress*>(static_cast<uint8_t*>(ptr) + offset);
+
                 auto* found_buffer_info = address_tracker.GetBufferByCaptureDeviceAddress(buffer_address);
                 bool  already_replaced  = false;
 
@@ -527,6 +530,9 @@ void VulkanAddressReplacer::ResolveBufferAddresses(VulkanCommandBufferInfo*     
                     }
                 }
             }
+
+            // unmap buffer again
+            resource_allocator_->UnmapResourceMemoryDirect(buffer_info->allocator_data);
         }
     }
 
--- a/framework/util/spirv_parsing_util.h
+++ b/framework/util/spirv_parsing_util.h
@@ -27,6 +27,7 @@
 #include <unordered_map>
 #include <map>
 #include <vector>
+#include <tuple>
 
 #include "util/defines.h"
 #include "encode/vulkan_state_info.h"


framework/decode/vulkan_address_replacer.cpp
```diff
@@ -519,6 +588,11 @@ void VulkanAddressReplacer::ProcessCmdBindDescriptorSets(VulkanCommandBufferInfo
 
     std::map<std::pair<VkDescriptorSet, uint32_t>, VkWriteDescriptorSetInlineUniformBlock> sets_requiring_update;
 
+    using set_key_t = std::tuple<uint32_t, uint32_t, uint32_t, uint32_t>;
+
+    // map used to identify chained pointer-accesses (identical set/binding/offsets)
+    std::set<set_key_t> dup_map;
```
antonio-lunarg, line 600: Does this need to be ordered? `unordered_set` should be faster in general.

framework/decode/vulkan_object_info.h
```diff
@@ -520,7 +521,7 @@ struct VulkanPipelineInfo : public VulkanObjectInfoAsync<VkPipeline>
     std::unordered_map<uint32_t, size_t> array_counts;
 
     // keep track of existing usage of buffer-references
-    std::vector<gfxrecon::util::SpirVParsingUtil::BufferReferenceInfo> buffer_reference_infos;
+    std::set<gfxrecon::util::SpirVParsingUtil::BufferReferenceInfo> buffer_reference_infos;
```
antonio-lunarg, line 524: Same for this. If you just need uniqueness guarantees, `unordered_set` might be a better choice.

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -580,18 +659,29 @@ void VulkanAddressReplacer::ProcessCmdBindDescriptorSets(VulkanCommandBufferInfo
...
-                VkDeviceAddress range_end =
-                    address + std::min<VkDeviceSize>(buffer_info->size - offset, desc_buffer_info.range);
-                command_buffer_info->addresses_to_replace.insert(address);
+                VkDeviceSize offset = desc_buffer_info.offset + buffer_ref_info.buffer_offset;
 
-                if (buffer_ref_info.array_stride)
+                // detected a chain of pointer-accesses
+                if (dup_map.contains(set_key))
+                {
+                    command_buffer_info->addresses_to_resolve[buffer_info].emplace_back(static_cast<size_t>(offset),
```
antonio-lunarg, line 673: Would it make sense to use `VkDeviceSize` directly instead of `size_t` for the map?

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -465,6 +467,73 @@ VkSemaphore VulkanAddressReplacer::UpdateBufferAddresses(
...
+                                                   const decode::VulkanDeviceAddressTracker& address_tracker)
+{
+    for (const auto& [buffer_info, offset_pairs] : command_buffer_info->addresses_to_resolve)
+    {
+        // TODO: implement read-back for device-local GPU buffer using a compute-dispatch or vkCmdCopyBuffers
+        if (!(buffer_info->memory_property_flags & VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT))
+        {
+            GFXRECON_LOG_WARNING_ONCE("%s: is required for buffer-handle: %d -- but currently only implemented for "
+                                      "host-visible buffers. replay is likely going to fail.",
+                                      buffer_info->capture_id);
```
antonio-lunarg, line 481: You wanted to use `__func__` here?

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -465,6 +467,73 @@ VkSemaphore VulkanAddressReplacer::UpdateBufferAddresses(
...
+            continue;
+        }
+
+        // map buffer
```
antonio-lunarg, line 485: Is the buffer never unmapped? Is that intended?

framework/util/spirv_parsing_util.h
```diff
@@ -76,6 +78,14 @@ class SpirVParsingUtil
     std::map<BufferReferenceInfo, std::vector<std::string>> buffer_reference_map_{};
 };
 
+// used to enable type as key for std::set/map
+inline bool operator<(const SpirVParsingUtil::BufferReferenceInfo& lhs,
+                      const SpirVParsingUtil::BufferReferenceInfo& rhs)
+{
+    return std::make_tuple(lhs.source, lhs.set, lhs.binding, lhs.buffer_offset, lhs.array_stride) <
```
antonio-lunarg, line 86: Let's include `<tuple>` for this.

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: the order is not important, but map-access is not performance critical (we expect a low-ish number of permutations per pipeline). also we would need a `std::hash` overload for the tuples for `unordered_set`

framework/decode/vulkan_object_info.h
fabian-lunarg: here we actually need the `BufferReferenceInfo` ordered. 
reason is when we later iterate the map we partially rely on that order to identify chained pointer accesses

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: was also thinking about it. but since `offset` can be applied to both host- and gpu-memory, I thought it could be confusing to use `VkDeviceSize`

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: sure did :)

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: I think I also wanted to unmap here but then realized our VulkanRebindAllocator only has an empty stub for the corresponding `UnmapResourceMemoryDirect`, so I skipped it.

fabian-lunarg: added a call to `resource_allocator_->UnmapResourceMemoryDirect(buffer_info->allocator_data);`
it's an empty stub, but communicates our intention and implementation could change in the future

framework/util/spirv_parsing_util.h
fabian-lunarg: done

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: fixed

antonio-lunarg (approved):

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -465,6 +467,73 @@ VkSemaphore VulkanAddressReplacer::UpdateBufferAddresses(
...
+                }
+                GFXRECON_ASSERT(found_buffer_info);
+
+                // ensure buffer exists and replacement actually required
+                if (found_buffer_info != nullptr && found_buffer_info->replay_address != 0 &&
+                    found_buffer_info->capture_address != found_buffer_info->replay_address)
+                {
+                    VkDeviceAddress base_address =
+                        already_replaced ? found_buffer_info->replay_address : found_buffer_info->capture_address;
+                    uint32_t address_offset = buffer_address - base_address;
```
MarkY-LunarG, line 512: Wouldn't this just be `offset`?

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -625,6 +715,8 @@ void VulkanAddressReplacer::ProcessCmdBindDescriptorSets(VulkanCommandBufferInfo
                 write_inline_uniform_block.pData    = descriptor_set_binding_info.inline_uniform_block.data();
             }
         }
+
+        dup_map.insert(set_key);
```
MarkY-LunarG, line 719: Won't this read-add something that still passes through line 665 resulting in duplication?

framework/decode/vulkan_object_info.h
```diff
@@ -691,7 +692,12 @@ struct VulkanCommandBufferInfo : public VulkanPoolObjectInfo<VkCommandBuffer>
 
     // collect buffer-device-addresses of locations to replace before submit
     std::unordered_set<VkDeviceAddress> addresses_to_replace;
-    bool                                inside_renderpass = false;
+
+    // maps buffers to (offset/stride)-pairs that need to be resolved
+    // (read back pointers, resolve additional buffers)
+    std::unordered_map<const VulkanBufferInfo*, std::vector<std::pair<size_t, uint32_t>>> addresses_to_resolve;
```
MarkY-LunarG, line 698: I was struggling with how this was different from the `addresses_to_replace`.  Then it hit me, this is buffers that you need to evaluate.  Perhaps you should call this `buffers_needing_resolved` or `datablocks_to_resolve_addresses` or something like that because once I thought of it that way, things made more sense (to me at least).

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: `offset` was the offset in 1st buffer. from there we read back a 2nd address, which points to another buffer. we resolve that by choosing correct `base_address` and calculate the `address_offset` for 2nd buffer

framework/decode/vulkan_object_info.h
fabian-lunarg: yes, I think the name could be improved. `buffers_needing_resolve` sounds more clear

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: don't think this could happen. we iterate `pipeline_info->buffer_reference_infos` (which is also a std::set, so no dups) and insert into `dup_map` once per loop-iteration. but this "detection by duplicate" is not the end of the story yet. this can only detect pointer-chains of length 1, so merely a beginning here.

PR #2610, opened 2026-01-19
Vdr maintenance

panos-lunarg:
Fixes a few vvl errors

### DIFF
--- a/framework/decode/vulkan_replay_dump_resources.cpp
+++ b/framework/decode/vulkan_replay_dump_resources.cpp
@@ -500,37 +500,15 @@ VulkanReplayDumpResourcesBase::FindTransferContext(VkCommandBuffer original_comm
     return nullptr;
 }
 
-void VulkanReplayDumpResourcesBase::ReleaseDrawCallContexts(decode::Index qs_index)
+template <typename MapOfContexts>
+void VulkanReplayDumpResourcesBase::ReleaseDumpingContexts(MapOfContexts contexts, decode::Index qs_index)
 {
-    for (const auto& [qs_bcb_pair, context] : draw_call_contexts_)
-    {
-        if (qs_bcb_pair.second == qs_index)
-        {
-            --active_contexts_;
-        }
-    }
-}
-
-void VulkanReplayDumpResourcesBase::ReleaseDispatchTraceRaysContexts(decode::Index qs_index)
-{
-    for (const auto& [qs_bcb_pair, context] : dispatch_ray_contexts_)
-    {
-        if (qs_bcb_pair.second == qs_index)
-        {
-            --active_contexts_;
-        }
-    }
-}
-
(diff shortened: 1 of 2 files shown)


framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -502,33 +502,49 @@ VulkanReplayDumpResourcesBase::FindTransferContext(VkCommandBuffer original_comm
 
 void VulkanReplayDumpResourcesBase::ReleaseDrawCallContexts(decode::Index qs_index)
 {
-    for (const auto& [qs_bcb_pair, context] : draw_call_contexts_)
+    for (auto it = draw_call_contexts_.begin(); it != draw_call_contexts_.end();)
     {
-        if (qs_bcb_pair.second == qs_index)
+        if (it->first.second == qs_index)
```
bradgrantham-lunarg: You can use `erase_if` here instead of a loop, and it returns the number of removed elements so you could just `removed_count = std::erase_if(...)` then `active_contexts_ -= removed_count`.

framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -1919,11 +1935,11 @@ VkResult VulkanReplayDumpResourcesBase::QueueSubmit(const std::vector<VkSubmitIn
             return res;
         }
 
-        for (auto [bcb_qs_pair, transf_context] : transfer_contexts_)
+        for (auto it = transfer_contexts_.begin(); it != transfer_contexts_.end();)
```
bradgrantham-lunarg: `std::erase_if` would be harder to use here.  I looked around for a good "walk this container, do work on items that match a condition, then delete them" idiom and all of them seemed awkward to me.  I think what you have is a reasonable idiom.

(std::ranges::stable_partition or std::views::filter both seem to make sense but require a manual `erase` step. There isn't a single "split this vector by condition" that I could find.  Here's what Claude and Gemini both offered but I don't like it.
```
// Partition: moves "to process" elements to the end, returns iterator to first one
auto it = std::partition(vec.begin(), vec.end(), [](const auto& x) {
    return !shouldProcess(x);  // Note: negated - keep items go first
});

// Work on the partitioned elements
for (auto work_it = it; work_it != vec.end(); ++work_it) {
    doWork(*work_it);
}

// Erase them
vec.erase(it, vec.end());
```

framework/decode/vulkan_replay_dump_resources.cpp
panos-lunarg: Done. Also merged all 3 new functions into a template one that handles all 3 cases

framework/decode/vulkan_replay_dump_resources.cpp
panos-lunarg: Rearranged things a bit and  moved the call to `ReleaseDumpingContexts()` outside of the loop

PR #2611, opened 2026-01-20
VulkanAddressReplacer: Make sure pWaitDstStageMask is valid

antonio-lunarg:


fabian-lunarg:
~~we should apply the same change to `OverrideQueueSubmit2` too~~
nevermind, you already did that :)

PR #2614, opened 2026-01-20
Use a global map to track Shader ID to LRS associations

davidd-lunarg:
Each shader identifier should be globally unique so it isn't necessary to track shader IDs per state object.

Below are supporting statements from the [DXR spec](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html):

**[Shader identifier](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#shader-identifier)**
"A shader identifier is an opaque data blob of 32 bytes that uniquely identifies (within the current device / process) one of the raytracing shaders"

"An application might create the same shader multiple times. This could be the same code but with same or different export names, potentially across separate raytracing pipelines or collections of code ([described later](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#state-object-types)). In this case the seemingly identical shaders may or may not return the same identifier depending on the implementation. Regardless, execution behavior will be consistent with the specified shader code."

**Regarding differing shader IDs on duplicate shaders:**

Update v1.24 05/30/2024 - "In GetShaderIdentifier removed the sentence: "The data itself globally identifies the shader, so even if the shader appears in a different state object (with same associations like any root signatures) it will have the same identifier.". This contradicted the text in Shader identifier that allows for duplicate shaders to have the same or different identifier depending on implementation."

### DIFF
--- a/framework/decode/dx12_object_info.h
+++ b/framework/decode/dx12_object_info.h
@@ -460,8 +460,7 @@ struct D3D12StateObjectInfo : DxObjectExtraInfo
     static constexpr char             kObjectType[] = "ID3D12StateObjectInfo";
     D3D12StateObjectInfo() : DxObjectExtraInfo(kType) {}
 
-    std::map<std::wstring, format::HandleId>                              export_name_lrs_map;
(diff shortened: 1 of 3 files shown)


davidd-lunarg:
Replaced with #2632

PR #2618, opened 2026-01-21
Fix recapture of empty handle ids on failed creation calls

bartosz-muszarski-arm:
Calls that create multiple objects can fail during capture, producing a list of null handles. Currently replayer in capture mode will attempt to reuse these null handles in the trimming process in attempt to restore the original object handles. This is an unexpected behavior and hits an assert.
This change prevents reuse of null handles generated by failing commands.


bradgrantham-lunarg:
@davidd-lunarg Can you take a look at this?

PR #2619, opened 2026-01-21
Fix descriptor copy indexing in state tracker

bartosz-muszarski-arm:
Currently descriptor copy path in state tracker uses the loop index (`d`/`i`) instead of the source array offset.
This results in incorrect descriptor_sets_bound_to relationships for images, buffers, acceleration structures, and texel views after descriptor copies, causing unexpected behavior - corruptions, crashes - during replay of trimmed traces.
Change implements look up of wrappers incrementally based on current_src_array_element for every descriptor type in the copy loop.


fabian-lunarg:
fyi @panos-lunarg 

the conclusion here was:
this was def. wrong and we seem to lack test-coverage. so we didn't notice.

fabian-lunarg:
@bartosz-muszarski-arm 

many thanks!

> ... corruptions, crashes - during replay of trimmed traces

I'd like to reproduce such case (looking into adding a regression-test).
could you point me at something that would trip over the wrong indexing?

edit:
we talked offline and cannot use their test-case for reasons. 
it should be possible to create an isolated (unit-)test to cover the descriptor-copying. 
we'll see who gets to this first :) but proceed with the PR regardless.

fabian-lunarg (approved):

PR #2620, opened 2026-01-21
Remove create pipeline cache control feature during replay

bartosz-muszarski-arm:
All pipeline types that support VK_EXT_pipeline_creation_cache_control can have VK_PIPELINE_CREATE_FAIL_ON_PIPELINE_COMPILE_REQUIRED_BIT flag enabled in the creation call. This flag allows the driver to not create pipelines if the creation would require compilation. This feature is problematic for capture/replay - gfxreconstruct will currently remove such calls during capture and always return VK_PIPELINE_COMPILE_REQUIRED to the caller. This behavior results in corruptions in new UE based content. This commit attempts to fix that by allowing the feature on capture, but removing it during replay:

Capture:
* change allow_pipeline_compile_required option to true by default This prevents capture layer from introducing any changes to the calls

Replay - on Create(Compute/Graphics/RayTracing)Pipeline:
* If creation call returned VK_PIPELINE_COMPILE_REQUIRED originally (pipeline creation failed), skip it on replay
* otherwise (pipeline creation succeeded) remove VK_PIPELINE_CREATE_FAIL_ON_PIPELINE_COMPILE_REQUIRED_BIT flag if present

This ensures the driver creates exactly the same set of objects it did during capture - it no longer has an option to fail if compilation is required.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -11765,6 +11765,7 @@ VkResult VulkanReplayConsumerBase::OverrideCreateGraphicsPipelines(
         return original_result;
     }
 
+    // remove potential VK_PIPELINE_CREATE_FAIL_ON_PIPELINE_COMPILE_REQUIRED_BIT during replay
     RemoveFailOnCompileRequiredFlags(const_cast<VkGraphicsPipelineCreateInfo*>(in_p_create_infos), create_info_count);
 
     // If there is no pipeline cache and we want to create a new one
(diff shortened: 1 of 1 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -11545,6 +11559,63 @@ void VulkanReplayConsumerBase::OverrideUpdateDescriptorSets(
...
+        {
+            handles[i] = VK_NULL_HANDLE;
+        }
+        return true;
+    }
+
+    return false;
+}
+
+template <typename CreateInfo>
```
fabian-lunarg: this would compile for all types that have a `flags` member.
we can improve this by adding `static_assert` or a simple `concept`,
in order to restrict compilation to Graphics/Compute/Raytracing create-infos.
it would also better communicate our assumptions for this routine.

framework/decode/vulkan_replay_consumer_base.h
```diff
@@ -1875,6 +1875,16 @@ class VulkanReplayConsumerBase : public VulkanConsumer
     typedef std::unordered_map<uint64_t, HardwareBufferInfo>               HardwareBufferMap;
     typedef std::unordered_map<format::HandleId, HardwareBufferMemoryInfo> HardwareBufferMemoryMap;
 
+    template <typename PipelineHandle>
```
fabian-lunarg: is there a case when `typename PipelineHandle` is not just `VkPipeline`?
meaning, is it necessary to template?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -11545,6 +11559,63 @@ void VulkanReplayConsumerBase::OverrideUpdateDescriptorSets(
...
+    {
+        if (create_infos[i].flags & VK_PIPELINE_CREATE_FAIL_ON_PIPELINE_COMPILE_REQUIRED_BIT_EXT)
+        {
+            create_infos[i].flags &= ~VK_PIPELINE_CREATE_FAIL_ON_PIPELINE_COMPILE_REQUIRED_BIT_EXT;
+            GFXRECON_LOG_WARNING("Removed VK_PIPELINE_CREATE_FAIL_ON_PIPELINE_COMPILE_REQUIRED_BIT_EXT flag from %s "
+                                 "create info index %u during replay.",
+                                 pipeline_type,
+                                 i);
+        }
+        if (create_infos[i].flags & VK_PIPELINE_CREATE_FAIL_ON_PIPELINE_COMPILE_REQUIRED_BIT)
```
fabian-lunarg, line 11679: the if-branches incl. log-warn are identical, only difference is `VK_PIPELINE_CREATE_FAIL_ON_PIPELINE_COMPILE_REQUIRED_BIT_EXT` vs. `VK_PIPELINE_CREATE_FAIL_ON_PIPELINE_COMPILE_REQUIRED_BIT`.

2nd case can be skipped, the value is the same for both (alias)

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: added a `static_assert`

framework/decode/vulkan_replay_consumer_base.h
fabian-lunarg: it was not required -> removed the template and just use `VkPipeline`

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: removed duplication for alias-values

PR #2621, opened 2026-01-22
Update test suite reference hash

beau-lunarg:


beau-lunarg:
http://jenkins:8080/job/gfxreconstruct/8678/

PR #2623, opened 2026-01-23
Add proper vkFrameBoundaryANDROID handling

marius-pelegrin-arm:
This PR is a reduction of #2507 to a more easily reviewable series of commits. It contains only handling of `vkFrameBoundaryANDROID` using the polymorphic behavior of `VulkanSwapchain` and doesn't contain any other type of offscreen frame boundary conversion.

`vkFrameBoundaryANDROID` is a non-standard Vulkan call (it does not belong to the core nor to any registered extension) used by some Android applications to present a `AHardwareBuffer` to the screen.
In particular, Android UI applications present to the screen using this call.

Currently, when the call is simply forwarded to the driver when encountered by the replayer.
The issue is that this call only actually presents to the screen when a `AHardwareBuffer`-bound image is submitted. However, at replay time, the `AHardwareBuffer` is often (always, as far as I encountered) impossible to recreate in the exact same way - and the issue gets worse when replaying on a different platform. Thus, even if the image is properly rendered and the call forwarded, the screen remains black - even on the exact same device.

This commit tries to fix this by considering the `vkFrameBoundaryANDROID` call like an additional `vkQueuePresentKHR` call.
The call is treated by the polymorphic `VulkanSwapchain` that decides whether to create a window a present to it (virtual swapchain behavior) or forward the call (offscreen/default swapchain behavior).

### DIFF
--- a/framework/decode/vulkan_virtual_swapchain.cpp
+++ b/framework/decode/vulkan_virtual_swapchain.cpp
@@ -41,10 +41,30 @@ void VulkanVirtualSwapchain::CleanDeviceResources(VkDevice device, const graphic
     {
         const auto& ofb_data = it->second;
 
+        for (auto semaphore : ofb_data.acquire_semaphores)
+        {
+            device_table->DestroySemaphore(device, semaphore, nullptr);
+        }
+
+        for (auto& img_data : ofb_data.image_datas)
+        {
+            if (img_data.copy_command_buffer != VK_NULL_HANDLE)
+            {
+                device_table->FreeCommandBuffers(device, ofb_data.command_pool, 1, &img_data.copy_command_buffer);
+            }
+
+            if (img_data.copy_semaphore != VK_NULL_HANDLE)
+            {
+                device_table->DestroySemaphore(device, img_data.copy_semaphore, nullptr);
+            }
+        }
+
         if (ofb_data.command_pool != VK_NULL_HANDLE)
         {
             device_table->DestroyCommandPool(device, ofb_data.command_pool, nullptr);
         }
+
+        ofb_data_.erase(device);
     }
 }
 


framework/decode/vulkan_offscreen_swapchain.cpp
```diff
@@ -291,6 +286,23 @@ VkResult VulkanOffscreenSwapchain::QueuePresentKHR(VkResult
...
+void VulkanOffscreenSwapchain::FrameBoundaryANDROID(PFN_vkFrameBoundaryANDROID           func,
+                                                    const VulkanDeviceInfo*              device_info,
+                                                    const VulkanSemaphoreInfo*           semaphore_info,
+                                                    const VulkanImageInfo*               image_info,
+                                                    VulkanInstanceInfo*                  instance_info,
+                                                    const graphics::VulkanInstanceTable* instance_table,
+                                                    const graphics::VulkanDeviceTable*   device_table,
+                                                    application::Application*            application)
+{
+    GFXRECON_ASSERT(device_info != nullptr);
```
fabian-lunarg, line 303: let's add `GFXRECON_UNREFERENCED_PARAMETER` here too

framework/decode/vulkan_virtual_swapchain.h
```diff
@@ -170,6 +179,33 @@ class VulkanVirtualSwapchain : public VulkanSwapchain
...
+        VkCommandBuffer copy_command_buffer{ VK_NULL_HANDLE };
+        VkSemaphore     copy_semaphore{ VK_NULL_HANDLE };
+    };
+
+    // This structure contains the custom surface, swapchain, and swapchain images data created and used by the virtual
+    // swapchain when encountering an offscreen frame boundary (like vkFrameBoundaryANDROID)
+    struct OFBData
+    {
+        VulkanSurfaceKHRInfo               surface_info{};
+        HandlePointerDecoder<VkSurfaceKHR> surface_ptr{};
```
fabian-lunarg, line 198: `ofb_data_.surface_ptr` is acquired in `VulkanVirtualSwapchain::FrameBoundaryANDROID`
but never released or cleaned up.
I assume the `VkSurface` is retrieved from- and owned by a window?

tools/replay/replay_settings.h
```diff
@@ -247,7 +247,9 @@ static void PrintUsage(const char* exe_name)
     GFXRECON_WRITE_CONSOLE("          \t\t    %s\tVirtual Swapchain of images which match", kSwapchainVirtual);
     GFXRECON_WRITE_CONSOLE("          \t\t         \tthe swapchain in effect at capture time and");
     GFXRECON_WRITE_CONSOLE("          \t\t         \twhich are copied to the underlying swapchain of the");
-    GFXRECON_WRITE_CONSOLE("          \t\t         \timplementation being replayed on. This is default.");
+    GFXRECON_WRITE_CONSOLE("          \t\t         \timplementation being replayed on. Also displays");
+    GFXRECON_WRITE_CONSOLE("          \t\t         \toffscreen frame boundaries to an additional window.");
+    GFXRECON_WRITE_CONSOLE("          \t\t         \tThis is only available and default for Vulkan.");
```
fabian-lunarg, line 252: ~~"This is default." hint got dropped~~

I misread, it's there: This is only available and default for Vulkan.

framework/decode/vulkan_offscreen_swapchain.cpp
fabian-lunarg: done

framework/decode/vulkan_virtual_swapchain.h
marius-pelegrin-arm: Yes, the surface is created specifically for `vkFrameBoundaryANDROID`.
We don't destroy the resources because there's no real way, from a `VulkanSwapchain`, to know when to destroy them. There is a `Clean` method, but it's called at replayer destruction time, when the device and resources have already been destroyed... I am not against setting up a proper cleanup though, what would you suggest ?

fabian-lunarg: aah k, think I get it. that handle is populated by `VulkanSwapchain::CreateSurface` via
`window->CreateSurface(instance_table_, instance, flags, replay_surface);`

and the only reason it's a `HandlePointerDecoder<VkSurfaceKHR>*` is,
so it fits the existing API for `VulkanSwapchain::CreateSurface`

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -1081,6 +1081,302 @@ void VulkanVirtualSwapchain::CmdPipelineBarrier2(PFN_vkCmdPipelineBarrier2 func,
...
+        device_table->GetDeviceQueue(device, 0, 0, &ofb_data_.queue);
+
+        VkCommandPoolCreateInfo command_pool_create_info;
+        command_pool_create_info.sType = VK_STRUCTURE_TYPE_COMMAND_POOL_CREATE_INFO;
+        command_pool_create_info.pNext = nullptr;
+        command_pool_create_info.flags =
+            VK_COMMAND_POOL_CREATE_TRANSIENT_BIT | VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT;
+        command_pool_create_info.queueFamilyIndex = 0;
+
+        result = device_table->CreateCommandPool(device, &command_pool_create_info, nullptr, &ofb_data_.command_pool);
```
fabian-lunarg: I cannot find any cleanup for `ofb_data_.command_pool`.
when 'should' it be cleaned up?

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -1081,6 +1081,302 @@ void VulkanVirtualSwapchain::CmdPipelineBarrier2(PFN_vkCmdPipelineBarrier2 func,
...
+        swapchain_create_info.preTransform          = VK_SURFACE_TRANSFORM_IDENTITY_BIT_KHR;
+        swapchain_create_info.compositeAlpha        = VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR;
+        swapchain_create_info.clipped               = VK_TRUE;
+        swapchain_create_info.oldSwapchain          = ofb_data_.swapchain;
+
+        switch (swapchain_options_.present_mode_option)
+        {
+            case util::PresentModeOption::kCapture:
+                // There is no "capture", but the closest is immediate as it is non-blocking.
+                swapchain_create_info.presentMode = VK_PRESENT_MODE_IMMEDIATE_KHR;
```
fabian-lunarg: if not explicitly requested, we should prob. avoid `VK_PRESENT_MODE_IMMEDIATE_KHR`
and prefer `VK_PRESENT_MODE_FIFO_KHR`,
since it's the only mode guaranteed to be available.
otherwise we would need to query if it's supported.

https://docs.vulkan.org/refpages/latest/refpages/source/VkPresentModeKHR.html#

framework/decode/vulkan_virtual_swapchain.cpp
marius-pelegrin-arm: Before device destruction. In general all resources not destroyed are meant to be kept alive "until the end" which is device destruction

fabian-lunarg: this would be here, there is already a call to the virtual `swapchain_->Clean()` -> https://github.com/LunarG/gfxreconstruct/blob/75fb9607adf7d94baf1d552f8b78aac0f2123ff1/framework/decode/vulkan_replay_consumer_base.cpp#L318

but `VulkanVirtualSwapchain` doesn't override it. so I figure we should add an override, doing the cleanup of `ofb_data_.command_pool`.

framework/decode/vulkan_virtual_swapchain.h
```diff
@@ -170,6 +179,33 @@ class VulkanVirtualSwapchain : public VulkanSwapchain
...
+        VkSwapchainKHR                     swapchain{ VK_NULL_HANDLE };
+
+        std::vector<VkSemaphore> acquire_semaphores{};
+        uint32_t                 acquire_index{ 0 };
+
+        std::vector<OFBSwapchainImageData> image_datas{};
+    };
+
+    OFBData                  ofb_data_;
+    std::vector<VkSemaphore> ofb_submit_semaphores_;
```
fabian-lunarg: `ofb_submit_semaphores_` aren't used anywhere

framework/decode/vulkan_virtual_swapchain.h
```diff
@@ -170,6 +179,33 @@ class VulkanVirtualSwapchain : public VulkanSwapchain
...
+        VkCommandPool                      command_pool{ VK_NULL_HANDLE };
+        VkSwapchainKHR                     swapchain{ VK_NULL_HANDLE };
+
+        std::vector<VkSemaphore> acquire_semaphores{};
+        uint32_t                 acquire_index{ 0 };
+
+        std::vector<OFBSwapchainImageData> image_datas{};
+    };
+
+    OFBData                  ofb_data_;
```
fabian-lunarg: OFBData groups device-specific data, but that's not reflected here.
also if we want to perform cleanup of adhoc created resources (in an override of `Clean()`),
we will require an associated `VKDevice` to do so.

I suggest we group `OFBData` in a device-map like this
`std::unordered_map<VkDevice, OFBData>               ofb_data_;`

@marius-pelegrin-arm if you don't mind I give that a shot, combining it with an override to `Clean()`

fabian-lunarg: ok, I got a commit that tracks `ofb_data_` by device, so far so good.
that allows to properly handle multiple devices and allow to cleanup adhoc resources.

the `Clean()` method override was not possible, because this is invoked after the device was destroyed already.
so I was forced to add a virtual CleanDeviceResources(VkDevice) and attach that in 
`VulkanReplayConsumerBase::OverrideDestroyDevice` and `FreeAllLiveObjects`.

so, please have a look and verify there are no issues with that 

> In particular, Android UI applications present to the screen using this call.

because I'd need to reproduce with OFB / android first.

framework/decode/vulkan_virtual_swapchain.h
fabian-lunarg: fixed

framework/decode/vulkan_virtual_swapchain.cpp
fabian-lunarg: see https://github.com/LunarG/gfxreconstruct/pull/2623/changes/24597c2ac7f470ee752a6a2ebedf8d6a8ce1bc9c

framework/decode/vulkan_virtual_swapchain.h
fabian-lunarg: https://github.com/LunarG/gfxreconstruct/pull/2623/changes/24597c2ac7f470ee752a6a2ebedf8d6a8ce1bc9c

marius-pelegrin-arm: Yes, that's what I mentionned that there was nothing to clean before device destruction. Your function seems to perfectly solve the issue. I will have a look at that =)

marius-pelegrin-arm: @fabian-lunarg I had a look. You cannnot call `encode::vulkan_wrappers::GetDeviceTable` in the virtual swapchain. This function and device table are only available at capture time, and the virtual swapchain is a replay time object.

Actually you can't really query the device table at all from there, that's why "the" device table is saved in the `VulkanSwapchain` class, which is an issue in itself.

From there, either you truly want the `VulkanSwapchain` to support multiple devices, and `VulkanSwapchain::device_table_` should be replaced with an unordered map `VkDevice -> VulkanDeviceTable` and the map should be filled during the `CreateSwapchainKHR` and `FrameBoundaryANDROID` calls, or we only want to stay in the scope of this PR and I can just add the device table to the arguments of `CleanDeviceResources`.

I wait for your answer =)

fabian-lunarg: good point about the device-table. and I agree we shouldn't attempt to fix more than necessary here :)
the point about multiple devices, while being a thing, is imo less important, but I would like to get cleanup right.

> I can just add the device table to the arguments of CleanDeviceResources

sounds like a good thing and should do the trick for the scope of this PR

fabian-lunarg: > add the device table to the arguments of CleanDeviceResources

done

framework/decode/vulkan_virtual_swapchain.cpp
fabian-lunarg: switched this to `VK_PRESENT_MODE_FIFO_KHR`, adjusted comment

framework/decode/vulkan_virtual_swapchain.h
marius-pelegrin-arm: Oh, thanks. Sorry for the delay, I am quite busy today ! x)

Looks good !

PR #2625, opened 2026-01-23
Move block parsing to preload for optimized --preload-measurement-range performance

jzulauf-lunarg:
Preload storage is vector of PreloadFrames containing vectors of ParsedBlocks moved from a deque of parsed blocks created during pre-loading.  Data construction and teardown are outside the measurement window.

Also fixes corner cases in 1st, 2nd, and stutter frames (when ProcessBlocks returns the same frame number twice), and removes extra frame when --quit-after-measurment-range is set.

### DIFF
--- a/framework/decode/file_processor.cpp
+++ b/framework/decode/file_processor.cpp
@@ -32,8 +32,6 @@
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
-const uint32_t kFirstFrame = 0;
-
 FileProcessor::FileProcessor() :
     current_frame_number_(kFirstFrame), error_state_(kErrorInvalidFileDescriptor), bytes_read_(0),
     annotation_handler_(nullptr), compressor_(nullptr), block_index_(0), block_limit_(0),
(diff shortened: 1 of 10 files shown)


framework/decode/file_processor.cpp
```diff
@@ -94,47 +91,27 @@ bool FileProcessor::Initialize(const std::string& filename)
...
 
-bool FileProcessor::DoProcessNextFrame(const std::function<bool()>& block_processor)
-{
-    bool success = IsFileValid();
-
-    if (success)
-    {
+    DispatchVisitor dispatch_visitor(decoders_, annotation_handler_);
+    // No need to decompress within the dispatch function, given kAlways policy.
+    block_parser_->SetDecompressionPolicy(BlockParser::DecompressionPolicy::kAlways);
```
MarkY-LunarG: Do we have to set this every frame?

framework/decode/file_processor.cpp
```diff
@@ -325,33 +292,61 @@ bool FileProcessor::ProcessBlocks()
...
+                    {
+                        // Invalid block. Error already logged in ParseBlock.
+                        process_state = ProcessBlockState::kError;
                     }
                 }
+                ++block_index_;
+                DecrementRemainingCommands();
             }
-            else
+            else // GetBlockBuffer failed
```
MarkY-LunarG: Do you mean `ReadBlockBuffer` failed?

framework/decode/parsed_block.h
```diff
@@ -104,15 +104,19 @@ class ParsedBlock
     using PoolEntry         = util::HeapBufferPool::Entry; // Placeholder for buffer pool
     using UncompressedStore = PoolEntry;
 
-    bool IsValid() const { return state_ != BlockState::kInvalid; }
-    bool IsReady() const { return state_ == BlockState::kReady; }
-    bool IsVisitable() const { return (state_ == BlockState::kReady) || (state_ == BlockState::kDeferredDecompress); }
-    bool IsUnknown() const { return state_ == BlockState::kUnknown; }
+    bool IsValid() const noexcept { return state_ != BlockState::kInvalid; }
```
ncesario-lunarg, line 107: Does gfxreconstruct use exceptions?

framework/decode/preload_file_processor.cpp
```diff
@@ -24,97 +24,195 @@
...
-                    const bool end_of_frame = block_buffer.IsFrameDelimiter(*this);
-                    // Record the block data for replay, Moves DataSpan out of BlockBuffer, so don't use after this.
-                    // NOTE: It is intentional to only store only the data span to make preload lightweight a possible
-                    pending_block_data_.emplace_back(std::move(block_buffer.ReleaseData()));
-                    if (end_of_frame)
-                    {
-                        // GFXRECON_LOG_INFO("Frame delimiter encountered during preload, ending frame preload.");
-                        break;
-                    }
+                    // This is really part of the non-preloaed previous (first) frame,
```
ncesario-lunarg: nit: "preloaed" -> "preload"

framework/decode/parsed_block.h
jzulauf-lunarg: It's there to give compiler the hint that it doesn't need to at guard logic...

framework/decode/preload_file_processor.cpp
```diff
@@ -24,97 +24,195 @@
...
+        if (!ContinueDecoding(block_index, true /* check decoder completion */))
+        {
+            process_state = ProcessBlockState::kEndProcessing;
+            break;
+        }
+
+        // We assume that only known, vistable blocks were preloaded
+        GFXRECON_ASSERT(queued_block.IsVisitable());
+
+        bool decompressed = false;
```
antonio-lunarg: Unused?

framework/decode/preload_file_processor.cpp
```diff
@@ -24,97 +24,195 @@
...
+            {
+                process_state = ProcessBlockState::kError;
+                break;
+            }
+            decompressed = true;
+        }
+
+        SetDecoderBlockIndex(block_index);
+        std::visit(dispatch_visitor, queued_block.GetArgs());
+        block_index++;
```
antonio-lunarg: This has no effect

framework/decode/preload_file_processor.cpp
```diff
@@ -24,97 +24,195 @@
...
-}
 
-bool PreloadFileProcessor::PreloadBlocksOneFrame()
-{
-    BlockBuffer block_buffer;
-    bool        success = true;
+    // Block processing will update current_frame_number_, so save and restore it,
+    // as callers rely on it remaining unchanged by preload.
+    const uint64_t save_current_frame = current_frame_number_;
+    bool           success            = true;
```
antonio-lunarg: This is never used

framework/decode/file_processor.h
```diff
@@ -102,6 +102,8 @@ class FileProcessor
     };
 
   public:
+    const static uint32_t kFirstFrame = 0;
```
antonio-lunarg: This can be `constexpr uint64_t`

jzulauf-lunarg:
All comments addressed, and rebased.

framework/decode/file_processor.cpp
jzulauf-lunarg: It's lightweight, and keeps me from tracking it elsewhere.   The preloader, does a save/restore, I could do it at parser construction time instead

MarkY-LunarG:
Do we need "preload" tests to make sure we don't break things?

jzulauf-lunarg:
> Do we need "preload" tests to make sure we don't break things?

Ideally, but there isn't support in the framework AFAICT.

jzulauf-lunarg:
@fabian-lunarg @antonio-lunarg ?

antonio-lunarg (approved):

PR #2627, opened 2026-01-26
Fix physical device discovery at replay time

marius-pelegrin-arm:
When the captured application enumerates physical devices multiple times, the `VulkanPhysicalDeviceInfo` instances get overwritten as many times in the table.
This leads to data acquired after the enumeration to be lost because the `VulkanPhysicalDeviceInfo` instances are not copied before being overwritten.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -3766,6 +3766,10 @@ VulkanReplayConsumerBase::OverrideEnumeratePhysicalDevices(PFN_vkEnumeratePhysic
 
                 if (old_physical_device_info->handle == new_physical_device_info->handle)
                 {
+                    GFXRECON_LOG_DEBUG("VulkanPhysicalDeviceInfo (%p) will be replaced with (%p) because of multiple "
+                                       "calls to vkEnumeratePhysicalDevices in the capture file.",
+                                       old_physical_device_info,
+                                       new_physical_device_info);
                     *new_physical_device_info = *old_physical_device_info;
                 }
             }
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -3756,6 +3756,21 @@ VulkanReplayConsumerBase::OverrideEnumeratePhysicalDevices(PFN_vkEnumeratePhysic
         SetInstancePhysicalDeviceEntries(
             instance_info, capture_device_count, capture_devices, replay_device_count, replay_devices);
 
+        // If it's not the first time we enumerate physical devices, they may already exist in the object info table.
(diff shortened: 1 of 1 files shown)


marius-pelegrin-arm:
I could detect the bug with TraceToolTest `vulkan_multidevice_1` (https://github.com/ARM-software/tracetooltests/blob/main/src/vulkan_multidevice_1.cpp)

But it was not deterministic and would succeed/fail depending on the device tested. I t can still be a good base to write a test,

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -3522,19 +3522,36 @@ VulkanReplayConsumerBase::OverrideEnumeratePhysicalDevices(PFN_vkEnumeratePhysic
...
+    // If it's not the first time we enumerate physical devices, they may already exist in the object info table.
+    // If they do, we want to copy them because otherwise they will be overwritten and capture data will be lost.
+    if (!pPhysicalDevices->IsNull())
+    {
+        for (uint32_t i = 0; i < capture_device_count; ++i)
+        {
+            VulkanPhysicalDeviceInfo* old_physical_device_info =
+                object_info_table_->GetVkPhysicalDeviceInfo(capture_devices[i]);
+            VulkanPhysicalDeviceInfo* new_physical_device_info =
+                reinterpret_cast<VulkanPhysicalDeviceInfo*>(pPhysicalDevices->GetConsumerData(i));
```
MarkY-LunarG: I don't think this is doing what you think.  The names look reversed to me.  The ConsumerData doesn't update until the very last of `Process_vkEnumeratePhysicalDevices`.  Because of that, the content at index `i` is old.  Also, what if we have completely new devices, you're overwritting a struct that has a std::vector, so you could be leaking data (not that this is a common occurrence, but desktops do support external GPUs that plug in via USB-c) and could cause breakage on this.

marius-pelegrin-arm:
I will still try to come up with a better matching between newly enumerated physical devices and the existing ones in the vulkan object info table.

marius-pelegrin-arm:
Here you go, I copy the `VulkanPhysicalDeviceInfo` based on the handle being the same as the "new" handle =)

MarkY-LunarG (approved):
LGTM.  Small request, just put a GFXRECON_LOG_DEBUG in the replace section saying "replacing x with y".  Just so we can work backwards if there are issues.

marius-pelegrin-arm:
> LGTM. Small request, just put a GFXRECON_LOG_DEBUG in the replace section saying "replacing x with y". Just so we can work backwards if there are issues.

Done =)

marius-pelegrin-arm:
I've started to repro the error internally again and I don't have time to investigate the issue for now, I will convert this ticket to draft and reopen it later.

marius-pelegrin-arm:
Re-tested locally on our pipeline and gave it a try for a few weeks. Seems to be working fine, so I re-open the PR.

PR #2628, opened 2026-01-26
Add support for `vkCmdPushConstants2{KHR}`

marius-pelegrin-arm:
There is currently support for `vkCmdPushConstants` but not for its "2" version counterparts.

The issue was detected using TraceToolTest `vulkan_compute_bda_pushconstant` (https://github.com/ARM-software/tracetooltests/blob/main/src/vulkan_compute_bda_pushconstant.cpp)


marius-pelegrin-arm:
I don't understand what you mean ? For this commit I just added two calls to the override json, which changed the generated vulkan replay consumer, which I added to the commit.

If you mean about my other PR, this is not related.

PR #2630, opened 2026-01-26
Fix bad argument forwarding in codegen

marius-pelegrin-arm:
In a generated replay consumer, if the current call is in the override list but also in the dump resource override list, and if an argument other than a `VkCommandBuffer` is a Vulkan handle, then the code generated is wrong.

This was discovered by trying to write a replay override for `vkCmdCopyBuffer` (which is currently also a dump resource override)

### DIFF
--- a/android/layer/CMakeLists.txt
+++ b/android/layer/CMakeLists.txt
@@ -58,6 +58,7 @@ target_link_libraries(VkLayer_gfxreconstruct
                       android
                       log)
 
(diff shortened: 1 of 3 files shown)


MarkY-LunarG:
Adding @panos-lunarg to look over since it's DR related.  Also, there's no resulting codegen files, did you forget to upload?

marius-pelegrin-arm:
@MarkY-LunarG No, I didn't forget to upload, it's just a case never encountered in your codebase yet.

PR #2632, opened 2026-01-27
Use a global map to track Shader ID to LRS associations

davidd-lunarg:
Each shader identifier should be globally unique so it isn't necessary to track shader IDs per state object.

Below are supporting statements from the [DXR spec](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html):

**[Shader identifier](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#shader-identifier)**
"A shader identifier is an opaque data blob of 32 bytes that uniquely identifies (within the current device / process) one of the raytracing shaders"

"An application might create the same shader multiple times. This could be the same code but with same or different export names, potentially across separate raytracing pipelines or collections of code ([described later](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#state-object-types)). In this case the seemingly identical shaders may or may not return the same identifier depending on the implementation. Regardless, execution behavior will be consistent with the specified shader code."

**Regarding differing shader IDs on duplicate shaders:**

Update v1.24 05/30/2024 - "In GetShaderIdentifier removed the sentence: "The data itself globally identifies the shader, so even if the shader appears in a different state object (with same associations like any root signatures) it will have the same identifier.". This contradicted the text in Shader identifier that allows for duplicate shaders to have the same or different identifier depending on implementation."

### DIFF
--- a/framework/decode/dx12_resource_value_mapper.h
+++ b/framework/decode/dx12_resource_value_mapper.h
@@ -168,6 +168,13 @@ class Dx12ResourceValueMapper
         std::map<uint64_t, graphics::Dx12ShaderIdentifier> mapped_shader_ids;
     };
 
+    struct ShaderIdAssociationInfo
+    {
+        std::set<format::HandleId>  state_object_ids;
+        format::HandleId            lrs_id;
+        std::set<ResourceValueInfo> resource_values;
+    };
+
     static void CopyResourceValues(const ResourceCopyInfo& copy_info, ResourceValueInfoMap& resource_value_info_map);
 
     static void CopyMappedResourceValues(const ResourceCopyInfo& copy_info);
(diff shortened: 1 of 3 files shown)


framework/decode/dx12_resource_value_mapper.h
```diff
@@ -168,6 +168,13 @@ class Dx12ResourceValueMapper
         std::map<uint64_t, graphics::Dx12ShaderIdentifier> mapped_shader_ids;
     };
 
+    struct ShaderIdAssociationInfo
+    {
+        std::set<format::HandleId>  state_object_ids;
```
MarkY-LunarG, line 173: Why are you saving this?  It doesn't appear used.

MarkY-LunarG (approved):
Changes look good  from a code standpoint, but I don't know if the functionality is correct, so I'll leave that for someone else to look over.  It appears to do the same thing as before with the only exception being the new warning.

Maybe @bradgrantham-lunarg or @locke-lunarg

framework/decode/dx12_resource_value_mapper.h
davidd-lunarg: This is saved for reference, to allow developers to determine which state object(s) contained this shader. For example, useful for debugging.

davidd-lunarg:
Thanks for taking a look.

> It appears to do the same thing as before with the only exception being the new warning.

The functional difference of this PR is contained in the first commit `Use a global map to track Shader ID - LRS data`. Previously the `shader_id_lrs_map` was stored and looked up per state object. However the IDs are global as stated in the spec, and this lead to issues where a state object can be created from other state objects (e.g., `AddToStateObject` or from an existing collection subobject) that weren't being tracked correctly. Tracking globally avoids the complication of building up the map between dependent state objects.

The second commit is for preserving information about which shader IDs came from which state objects. This doesn't have a functional difference on replay--it is for verification and debugging purposes. The shader ID to LRS mapping can be complicated to resolve, and the small amount of additional work to track this meta info should be useful to have around in the future.

framework/decode/dx12_resource_value_mapper.h
davidd-lunarg: This is saved for reference, to allow developers to determine which state object(s) contained this shader. For example, useful for debugging.

locke-lunarg (approved):
LGTM!

davidd-lunarg:
There was an instance found where two state objects referenced the same shader ID but in each state object the same shader ID had different LRS associated with it. Because of this the shader ID - LRS associations must be tracked per state object and not globally. PR #2551 retains that behavior and also addressed an issue related to creating a state object from a state object colleciton.

PR #2634, opened 2026-01-27
Fix support for `VK_KHR_pipeline_binary`

marius-pelegrin-arm:
`VkPipelineCreateInfoKHR` is a particular structure as it can take any kind of `PipelineCreateInfo` in its `pNext` chain, in addition to being required to have a non-null `pNext` member.

This was not correctly handled by the generated code because `Vk*PipelineCreateInfo` structures don't have the `structextends` flag set in the Vulkan registry. This commit aims at fixing that by blacklisting this structure and writing the encoder/decoders manually.

Ideally, this should be solved in the codegen by not considering only structs with the `structextends` flag as possible encounters in a `pNext` chain

### DIFF
--- a/framework/generated/generated_vulkan_replay_consumer.cpp
+++ b/framework/generated/generated_vulkan_replay_consumer.cpp
@@ -6467,20 +6467,19 @@ void VulkanReplayConsumer::Process_vkCreatePipelineBinariesKHR(
     StructPointerDecoder<Decoded_VkAllocationCallbacks>* pAllocator,
     StructPointerDecoder<Decoded_VkPipelineBinaryHandlesInfoKHR>* pBinaries)
 {
-    VkDevice in_device = MapHandle<VulkanDeviceInfo>(device, &CommonObjectInfoTable::GetVkDeviceInfo);
-    const VkPipelineBinaryCreateInfoKHR* in_pCreateInfo = pCreateInfo->GetPointer();
+    auto in_device = GetObjectInfoTable().GetVkDeviceInfo(device);
+
     MapStructHandles(pCreateInfo->GetMetaStructPointer(), GetObjectInfoTable());
-    const VkAllocationCallbacks* in_pAllocator = GetAllocationCallbacks(pAllocator);
-    SetStructArrayHandleLengths<Decoded_VkPipelineBinaryHandlesInfoKHR>(pBinaries->GetMetaStructPointer(), pBinaries->GetLength());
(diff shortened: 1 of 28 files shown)


framework/generated/generated_vulkan_replay_consumer.cpp
```diff
@@ -6467,20 +6467,19 @@ void VulkanReplayConsumer::Process_vkCreatePipelineBinariesKHR(
...
-    VkDevice in_device = MapHandle<VulkanDeviceInfo>(device, &CommonObjectInfoTable::GetVkDeviceInfo);
-    const VkPipelineBinaryCreateInfoKHR* in_pCreateInfo = pCreateInfo->GetPointer();
+    auto in_device = GetObjectInfoTable().GetVkDeviceInfo(device);
+
     MapStructHandles(pCreateInfo->GetMetaStructPointer(), GetObjectInfoTable());
-    const VkAllocationCallbacks* in_pAllocator = GetAllocationCallbacks(pAllocator);
-    SetStructArrayHandleLengths<Decoded_VkPipelineBinaryHandlesInfoKHR>(pBinaries->GetMetaStructPointer(), pBinaries->GetLength());
-    VkPipelineBinaryHandlesInfoKHR* out_pBinaries = pBinaries->IsNull() ? nullptr : pBinaries->AllocateOutputData(1, { VK_STRUCTURE_TYPE_PIPELINE_BINARY_HANDLES_INFO_KHR, nullptr });
+    SetStructHandleLengths(pBinaries->GetMetaStructPointer());
+    pBinaries->IsNull() ? nullptr : pBinaries->AllocateOutputData(1, { VK_STRUCTURE_TYPE_PIPELINE_BINARY_HANDLES_INFO_KHR, nullptr });
```
MarkY-LunarG, line 6474: What are you setting to nullptr here?  I think there's some code missing.  (Same for 6507 below).

MarkY-LunarG:
This seems similar to the way OpenXR handles parent/child classes (for example look at `XrEventDataBaseHeader` and it's derivatives including how `Decoded_XrEventDataBaseHeader` performs a lot of work for allocations).  Additionally, we have hooks in place to handle the proper look-ahead required and handling for static and dynamic arrays.  For example, decoding with `DecodeBaseHeader` instead of the standard decode.

Honestly, this really does seem like something the Vulkan registry is missing as well.  In OpenXR, the registry has field tags indicating when a structure has a parent structure (`parentstruct="<parent-struct-name>"`).  For example, in this case, all the other "CreateInfos" would be listed with `parentstruct="VkPipelineCreateInfoKHR"`. 
But we can focus on that later and do "custom" functionality  for now here.

MarkY-LunarG:
I made a change in PR #2668, try that out.

marius-pelegrin-arm:
Fixed by @MarkY-LunarG change to codegen

PR #2642, opened 2026-01-30
Wait before first submit

antonio-lunarg:
Add a replay option to wait for a specified number of milliseconds before submitting the first command buffer.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -4098,6 +4098,8 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit(PFN_vkQueueSubmit        
                                                        StructPointerDecoder<Decoded_VkSubmitInfo>* pSubmits,
                                                        const VulkanFenceInfo*                      fence_info)
 {
+    options_.MaybeWaitBeforeFirstSubmit();
+
     assert((queue_info != nullptr) && (pSubmits != nullptr));
 
     VkResult            result       = VK_SUCCESS;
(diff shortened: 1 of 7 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4088,6 +4088,12 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit(PFN_vkQueueSubmit
                                                        StructPointerDecoder<Decoded_VkSubmitInfo>* pSubmits,
                                                        const VulkanFenceInfo*                      fence_info)
 {
+    if (first_queue_submit_)
+    {
+        first_queue_submit_ = false;
+        options_.MaybeWaitBeforeFirstSubmit();
+    }
+
```
MarkY-LunarG: This would also have to be put in QueueSubmit2 as well.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4088,6 +4088,12 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit(PFN_vkQueueSubmit
                                                        StructPointerDecoder<Decoded_VkSubmitInfo>* pSubmits,
                                                        const VulkanFenceInfo*                      fence_info)
 {
+    if (first_queue_submit_)
```
MarkY-LunarG: Couldn't you just do a `std::call_once` if this option was enabled?

antonio-lunarg: Good point. I encapsulated everything in `VulkanReyplayOptions`

PR #2643, opened 2026-01-30
Page guard fixes

panos-lunarg:
Two commits in this PR:

- The installation of the signal handler watcher is moved so that it is installed only when Android's libsigchain is not used
- Give option to disable libsigchain


mikes-lunarg (approved):
LGTM. I'm curious what kinds of issues using libsigchain caused. Prior to this I assumed it worked the same as sigaction, just with a higher priority.

panos-lunarg:
> LGTM. I'm curious what kinds of issues using libsigchain caused. Prior to this I assumed it worked the same as sigaction, just with a higher priority.

Tbh I am not sure. There's an app that simply terminates without anything enlightening in logcat. Switching to sigaction just makes it work. It has to be related to the Android version as on a different phone with Android 15 (vs 14 ) it just works without this PR

PR #2646, opened 2026-02-02
Logging improvements

MarkY-LunarG:
These changes pave the way for improved logging:

* Logging to multiple locations simultaneously
* Each logging target can have it's own logging level eventually
* New logging targets can easily be added
* Improved initialization and destruction (now using a singleton with unique_ptr usage).

NOTE: This is squashed because it was already pre-approved in another workflow (sorry for the large chunk).

### DIFF
--- a/framework/util/logging.cpp
+++ b/framework/util/logging.cpp
@@ -36,6 +36,198 @@
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(util)
 
+constexpr char g_process_tag[] = "gfxrecon";
+
+GFXRECON_BEGIN_NAMESPACE(logging)
+
+std::unique_ptr<LoggingManager> LoggingManager::singleton_;
+std::once_flag                  LoggingManager::singleton_flag_;
+
+LoggingManager& LoggingManager::GetSingleton()
+{
+    std::call_once(singleton_flag_, [&]() { singleton_.reset(new LoggingManager()); });
+    return *singleton_;
+}
+
+LoggingManager::LoggingManager()
+{
+    const std::lock_guard<std::mutex> lock(target_mut_);
+
+    // Initialize the unique pointers to be empty
+    for (auto& tgt : logging_targets_)
+    {
+        tgt.reset(nullptr);
+    }
+}
+
+LoggingManager::~LoggingManager() {}
+
+void LoggingManager::LogMessage(LoggingSeverity severity, const std::string& message)
+{
+    const std::lock_guard<std::mutex> lock(target_mut_);
+
+    for (auto& tgt : logging_targets_)
+    {
+        if (tgt)
+        {
+            tgt->LogMessage(severity, message);
+        }
+    }
+}
+
+void LoggingManager::SetMinimumSeverity(LoggingSeverity severity)
+{
+    const std::lock_guard<std::mutex> lock(target_mut_);
+
+    for (auto& tgt : logging_targets_)
+    {
+        if (tgt)
+        {
+            tgt->SetSeverity(severity);
+        }
+    }
+}
+
+void LoggingManager::LogMessage(LoggingSeverity    severity,
+                                const std::string& indented_message,
+                                const std::string& non_indented_message)
+{
+    bool use_indent = indent_supported_ && indented_message.length() != non_indented_message.length();
+    if (!use_indent)
+    {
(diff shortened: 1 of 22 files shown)


framework/util/test/test_spirv_parsing_util.cpp
```diff
@@ -507,8 +506,7 @@ TEST_CASE("SpirVParsingUtil", "[buffer_references]")
 
 TEST_CASE("SpirVParsingUtil_offsets", "[buffer_references_with_offset]")
 {
-    // setting this to kDebugSeverity yields debug-output from parser
-    gfxrecon::util::Log::Init(gfxrecon::util::Log::kErrorSeverity);
+    gfxrecon::util::Log::Init(gfxrecon::util::LoggingSeverity::kDebug);
```
fabian-lunarg: this is not required here, same above w. `kInfo`. makes the unit-test quite verbose since `SpirVParsingUtil` uses `GFXRECON_LOG_DEBUG` to print a summary

framework/util/logging.cpp
```diff
@@ -36,6 +36,180 @@
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(util)
 
+const char g_process_tag[] = "gfxrecon";
```
fabian-lunarg: `constexpr char g_process_tag[] = "gfxrecon";` should also work

framework/util/logging.cpp
```diff
@@ -36,6 +36,180 @@
...
+    }
+    else if (logging_targets_[kTarget_StdOut])
+    {
+        // Just disable it
+        logging_targets_[kTarget_StdOut]->SetEnable(false);
+    }
+    return true;
+}
+
+bool LoggingManager::UpdateStdErrTarget(bool write_to_stderr, bool flush_after_write)
```
fabian-lunarg: `UpdateStdErrTarget` always returns `true`. should it rather return if state actually got changed? 

like:
```
bool was_using_stderr = logging_targets_[kTarget_StdErr] && logging_targets_[kTarget_StdErr]->IsEnabled(); // LoggingTargetBase::IsEnabled() does not exist, but would be useful
...
return write_to_stderr != was_using_stderr;
```

framework/util/test/test_spirv_parsing_util.cpp
MarkY-LunarG: Not sure why these were changed to Debug.  I'll change them back to kError.

framework/util/logging.cpp
MarkY-LunarG: I just removed the return type for `UpdateStdOutTarget`, `UpdateStdErrTarget`, and `UpdateDebugViewTarget`.  It didn't really make sense for those.  I did add an error message in `UpdateDebugViewTarget` if the option was enabled but the platform didn't support it.

framework/util/logging_targets.cpp
```diff
@@ -0,0 +1,260 @@
...
+    }
+    else
+    {
+        GFXRECON_LOG_FATAL("Failed opening log file %s with returned result %d", log_file_name_.c_str(), errno);
+    }
+
+    return success;
+}
+
+void LoggingTargetFile::CloseFile()
```
fabian-lunarg, line 220: LogMessage() uses the `file_mut_` mutex while writing. `OpenFile`/`CloseFile` are done outside that lock,
but also require it I think.
if multiple threads concurrently log with `leave_open == false`, this looks like a race-condition opening/closing the same `std::ofstream`.

framework/util/logging.cpp
```diff
@@ -36,6 +36,182 @@
...
+                else
+                {
+                    tgt->LogMessage(severity, non_indented_message);
+                }
+            }
+        }
+    }
+}
+
+void LoggingManager::UpdateStdOutTarget(bool write_to_console, bool write_to_stderr, bool flush_after_write)
```
fabian-lunarg, line 121: `LoggingManager::UpdateStdOutTarget` and `LoggingManager::LogMessage` both access `logging_targets_` without synchronization.

if `UpdateStdOutTarget` is only used once during init, this should be fine.
otherwise, when this can occur during runtime (or error-on-side-of-caution), would require
a mutex for `logging_targets_` I think

MarkY-LunarG: Dude, great catch.  I'm going to add a mutex to police any time we're touching the targets.

PR #2648, opened 2026-02-02
Vdr get linear image sizes

panos-lunarg:
VDR: Get linear image sizes
When dumping images as raw binaries, we dumped the whole image's size as
reported by vkGetImageMemoryRequirements. This includes implementation
imposed alignments which makes decoding of the file hard.

This commit changes this by calculating the images' sizes as tightly
packed with linear tiling not including any paddings added for
alignment.


pemgithub:
fix looks good, works with my test case (Pixel Tablet) (129 = VK_FORMAT_D24_UNORM_S8_UINT) (130 = VK_FORMAT_D32_SFLOAT_S8_UINT)

framework/graphics/vulkan_resources_util.cpp
```diff
@@ -999,6 +999,70 @@ uint64_t VulkanResourcesUtil::GetImageResourceSizesOptimal(VkFormat
     return resource_size;
 }
 
+uint64_t VulkanResourcesUtil::GetImageResourceSizesLinear(VkFormat               format,
+                                                          const VkExtent3D&      extent,
+                                                          uint32_t               mip_levels,
+                                                          uint32_t               array_layers,
+                                                          VkImageAspectFlagBits  aspect,
+                                                          std::vector<uint64_t>* subresource_offsets,
```
bradgrantham-lunarg: Will subresource_offsets or subresource_sizes ever be nullptr?  If not, making these a reference makes the intent clearer and allows you to drop the checks on those pointers in this function.

panos-lunarg: done

PR #2649, opened 2026-02-03
Initial image info for incomplete

locke-lunarg:
If the swapchain count isn't full image count, `GetSwapchainImagesKHR` will return `VK_INCOMPLETE`. It should also initialize image infos.  If it doesn't initialize, it won't be able to know whether the image info is for swapchain or not. It might also have other issues caused by lack of initialization.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -7988,7 +7988,10 @@ VkResult VulkanReplayConsumerBase::OverrideGetSwapchainImagesKHR(PFN_vkGetSwapch
         result = swapchain_->GetSwapchainImagesKHR(
             original_result, func, device_info, swapchain_info, capture_image_count, replay_image_count, replay_images);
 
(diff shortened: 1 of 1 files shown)


fabian-lunarg (approved):
makes sense

davidlunarg:
Code looks good.  Will do further testing.

davidlunarg:
This branch passed Vulkan extended and commit CI tests. The recent failure of LunarG CI Checkrun was due to an Android device running out of space on the device.

bradgrantham-lunarg:
Is there a test case we can add, either to test_cases or capture testing in Jenkins, that will succeed with this change and fail without it?

davidlunarg:
> Is there a test case we can add, either to test_cases or capture testing in Jenkins, that will succeed with this change and fail without it?

https://github.com/LunarG/VulkanTests/pull/828 found this test case. Once it it is merged, we will have a test case. PR 828 adds replay to additional surfaces, which crash without this fix.

PR #2650, opened 2026-02-03
Adjust stride of dumped AHB image

panos-lunarg:
AHBs can have a different stride than the VkImage we use to dump them from the gpu when we cannot access them from the host directly. This can lead to artifacts during replay since the binary payload we upload has the stride of the image we used to dump them.


ziga-lunarg (approved):
LGTM

MarkY-LunarG (approved):
Nice and clean.  Good comments too. Thnx Panos.

PR #2654, opened 2026-02-03
codegen: Generate any struct with a pNext/next

MarkY-LunarG:
For handle mapping and extension struct (pNext/next) struct walking during encode/decode, support ANY struct with the corresponding fields.  This is because application's may add structs in a way that appears erroneous to the spec, but may be supported by an underlying driver.


MarkY-LunarG:
Test case from @marius-pelegrin-arm :

```
git clone https://github.com/ARM-software/tracetooltests.git
cd tracetooltests
mkdir build
cd build
cmake ..
make -j

# Setup capture layer
..

# The command line to capture
./vulkan_compute_2 -b

# Replay the capture
...
```

MarkY-LunarG:
@fabian-lunarg, this needs reviewing to unblock  ARM for binary pipeline support.

MarkY-LunarG:
Squashed down to 2 commits.

PR #2666, opened 2026-02-09
VDR: fixes for AS addresses

panos-lunarg:


framework/decode/vulkan_object_info.h
```diff
@@ -426,7 +426,11 @@ struct VulkanBufferInfo : public VulkanObjectInfo<VkBuffer>
 
     // map acceleration-structure replay-addresses to existing (alias) AS-handles
     std::unordered_map<VkDeviceAddress, std::unordered_set<const VulkanAccelerationStructureKHRInfo*>>
-        acceleration_structures;
+        acceleration_structures_replay_addresses;
+
+    // map acceleration-structure capture-addresses to existing (alias) AS-handles
+    std::unordered_map<VkDeviceAddress, std::unordered_set<const VulkanAccelerationStructureKHRInfo*>>
+        acceleration_structures_capture_addresses;
```
fabian-lunarg, line 424: the two maps here bother me a bit. they have to be kept in sync, which is error-prone.
I would prefer this kind of book-keeping should, as much as possible, be done only once in `decode::VulkanAddressTracker`

panos-lunarg: Can you elaborate a bit more? Both maps are updated in `VulkanDeviceAddressTracker::TrackAccelerationStructure` and used only inside `VulkanDeviceAddressTracker` so it's not that they are scattered

framework/decode/vulkan_device_address_tracker.cpp
```diff
@@ -214,15 +218,32 @@ VulkanDeviceAddressTracker::GetBufferInfo(VkDeviceAddress
...
 const std::unordered_set<const VulkanAccelerationStructureKHRInfo*>&
 VulkanDeviceAddressTracker::GetAccelerationStructuresByCaptureDeviceAddress(VkDeviceAddress capture_address) const
 {
     // delegate query to buffer
     const auto* buffer_info = GetBufferByCaptureDeviceAddress(capture_address);
     if (buffer_info != nullptr)
     {
-        auto handle_set_it = buffer_info->acceleration_structures.find(capture_address);
-        if (handle_set_it != buffer_info->acceleration_structures.end())
+        auto handle_set_it = buffer_info->acceleration_structures_capture_addresses.find(capture_address);
```
fabian-lunarg, line 246: ok, I see, there was a bug mixing capture/replay-time addresses for this lookup.

this is the only place you use the new `buffer_info->acceleration_structures_capture_addresses`, but easier to just use the replay-address here.

```
const auto* buffer_info = GetBufferByCaptureDeviceAddress(capture_address);
if (buffer_info != nullptr)
{
     auto handle_set_it = buffer_info->acceleration_structures.find(buffer_info->replay_address);
...
}
```

framework/decode/vulkan_object_info.h
fabian-lunarg: I mean the second map is redundant, as we don't need two maps to model the mapping 
from VkBuffer to VulkanAccelerationStructure. one is enough, two are error-prone. this looks this is mixing AS-lookup with capture/replay-time address translation.

framework/decode/vulkan_device_address_tracker.cpp
panos-lunarg: `buffer_info->acceleration_structures.find(buffer_info->replay_address)` 
Will this work? An AS can be at an offset but `buffer_info->replay_address` has the base address

fabian-lunarg: you're right, the offset has to be accounted for.

```
const auto* buffer_info = GetBufferByCaptureDeviceAddress(capture_address);
if (buffer_info != nullptr)
{
     uint32_t as_offset = capture_address - buffer_info->capture_address;
     auto handle_set_it = buffer_info->acceleration_structures.find(buffer_info->replay_address + as_offset);
     ...
}
```

panos-lunarg: ah nice! :)
Still there's a conflict with `buffer_info->acceleration_structures`, `acceleration_structure_addresses_` and `VulkanAddressReplacer::DestroyShadowResources`. Pushing some more changes hopping to address this

framework/decode/vulkan_object_info.h
```diff
@@ -424,9 +424,9 @@ struct VulkanBufferInfo : public VulkanObjectInfo<VkBuffer>
     VkDeviceSize       size{ 0 };
     uint32_t           queue_family_index{ 0 };
 
-    // map acceleration-structure replay-addresses to existing (alias) AS-handles
+    // map acceleration-structure capture-addresses to existing (alias) AS-handles
     std::unordered_map<VkDeviceAddress, std::unordered_set<const VulkanAccelerationStructureKHRInfo*>>
-        acceleration_structures;
+        acceleration_structures_capture_addresses;
```
fabian-lunarg, line 424: there was a reason this was using replay-addresses here 
(for portability, we can always determine the replay-address by injecting calls, but we don't always know the capture-address).
not entirely sure if that applies only to internal resources in `VulkanAddressReplacer`.
also an existing set of regression-tests for `-m rebind` is currently not active yet,
so this makes me a bit nervous. going to run some manual tests against your branch

framework/decode/vulkan_device_address_tracker.cpp
```diff
@@ -88,6 +90,9 @@ void VulkanDeviceAddressTracker::TrackAccelerationStructure(
             acceleration_structure_addresses_[acceleration_structure_info->capture_address] =
                 acceleration_structure_info->replay_address;
 
+            buffer_info->acceleration_structures_capture_addresses[acceleration_structure_info->capture_address].insert(
```
fabian-lunarg: should be checked for `nullptr`, moved inside the `if (buffer_info != nullptr)`{...}

framework/decode/vulkan_object_info.h
fabian-lunarg: ok, this looks fine, was able to verify the mentioned set of portability test-cases using `-m rebind`
still works as expected.

fabian-lunarg (approved):
just this one `nullptr`-check is missing for me, otherwise no concerns

PR #2668, opened 2026-02-09
Treat VkPipelineCreateInfoKHR as a parent struct

MarkY-LunarG:
VkPipelineCreateInfoKHR structure is pretty much a parent struct with VkGraphicsPipelineCreateInfo, VkComputePipelineCreateInfo, and VkRayTracingPipelineCreateInfoKHR as children.

So, add a mechanism to allow for forcing the definition of parent/ child classes in python since the Vulkan spec isn't updated to handle this.


MarkY-LunarG:
This is intended for the same fix as #2634. @marius-pelegrin-arm, can you see if this resolves your issues better?

[Edit] It's not working yet, so hold off.

marius-pelegrin-arm:
@MarkY-LunarG Sure, ping me when it's ready =)

MarkY-LunarG:
@marius-pelegrin-arm, I believe this change is ready for you to try out.  It's going to fail tests until it's checked in because it is comparing the known good of the old `pipeline-binaries` test to the new value.   Once this goes in it should pass that as well.

framework/generated/khronos_generators/khronos_struct_decoders_body_generator.py
```diff
@@ -141,7 +141,7 @@ def make_decode_invocation(
                         main_body += '\n'
                         main_body += '    // For base header arrays of pointers, we need to allocate an array to the generic base type pointer first\n'
                         main_body += '    // and then read the array attributes so we can jump right in to decoding the contents\n'
-                        main_body += f'    wrapper->{value.name} = DecodeAllocator::Allocate<StructPointerDecoder<Decoded_{value.base_type}{' * '*(value.pointer_count-1)}>>();\n'
+                        main_body += f'    wrapper->{value.name} = DecodeAllocator::Allocate<StructPointerDecoder<Decoded_{value.base_type}{'*' *(value.pointer_count-1)}>>();\n'
```
marius-pelegrin-arm: Just in case, this is failing to run on Python. The "exterior format string" is defined using single quotes, so you cannot use single quotes inside the braces. I recommend you change one of them for double quotes

MarkY-LunarG: Weird that it didn't fail.  Good catch.  Fixed.

framework/generated/khronos_generators/vulkan_generators/blacklists.json
```diff
@@ -28,8 +28,7 @@
   "functions-encoder": [
     "vkCreateComputePipelines",
     "vkCreateGraphicsPipelines",
-    "vkCreateRayTracingPipelinesNV",
-    "vkCreateRayTracingPipelinesKHR"
```
MarkY-LunarG, line 32: NOTE: This was removed because it was superseded by the define in the above block on line 18.

framework/generated/khronos_generators/khronos_struct_decoders_body_generator.py
bradgrantham-lunarg: Is a Python 3.12+ change; the braces cause reinterpretation of everything between them.  I've added https://github.com/LunarG/gfxreconstruct/pull/2683 to add 3.10 to our GitHub CI to match CONTRIBUTING.md to help catch things like this.

MarkY-LunarG: This is also now fixed as well (internal section uses double quotes ["*"]).

marius-pelegrin-arm:
I had a very quick look and tried to re-run the sample but it failed to replay again. The error is different though. I guess VkPipelineCreateInfoKHR was treated correctly but the sample failed on another similar case of struct not handled in pNext. I will not be able to have a deeper look right now, but if you want to try it is pretty easy to test by yourself:

```
git clone https://github.com/ARM-software/tracetooltests.git
cd tracetooltests
mkdir build
cd build
cmake ..
make -j

# Setup capture layer
..

# The command line to capture
./vulkan_compute_2 -b

# Replay the capture
...
```

MarkY-LunarG:
> I had a very quick look and tried to re-run the sample but it failed to replay again. The error is different though. I guess VkPipelineCreateInfoKHR was treated correctly but the sample failed on another similar case of struct not handled in pNext. I will not be able to have a deeper look right now, but if you want to try it is pretty easy to test by yourself:
> 
> ```
> git clone https://github.com/ARM-software/tracetooltests.git
> cd tracetooltests
> mkdir build
> cd build
> cmake ..
> make -j
> 
> # Setup capture layer
> ..
> 
> # The command line to capture
> ./vulkan_compute_2 -b
> 
> # Replay the capture
> ...
> ```

@marius-pelegrin-arm , it looks like I misread the spec.  On my first read-through of the spec, I understood it as a parent/child relationship, but looking at the [reference manual section](https://docs.vulkan.org/refpages/latest/refpages/source/VkPipelineCreateInfoKHR.html) it does look like the structs are passed in purely via the `pNext` chain. 

So this change is incorrect.  Because of this, I'm going to close this and switch back to PR #2654.  That PR does seem to handle the capture portion without error, however, the replay crashes on my AMD system in the AMD driver right now.  I will likely close this issue if you want to try that next.

PR #2669, opened 2026-02-09
logging: Fix missing capture log file

MarkY-LunarG:
This fixes the issue where the defined log file wasn't created during capture if defined.


MarkY-LunarG:
@davidd-lunarg, did this resolve your issue?

PR #2671, opened 2026-02-10
Fix gradle options

marius-pelegrin-arm:
Since the latest Gradle version update, gradle properties are not read correctly from the command line anymore.


MarkY-LunarG:
@antonio-lunarg , can you look this over since you last updated that file?

PR #2672, opened 2026-02-10
Do not allow lazy allocation of virtual swapchain images

marius-pelegrin-arm:
In the `VulkanVirtualSwapchain`, if the captured swapchain was created with `VK_SWAPCHAIN_CREATE_DEFERRED_MEMORY_ALLOCATION_BIT_KHR`, images could not be bound to memory when `vkGetSwapchainImagesKHR` is called.

The issue is that this function, at replay time, is when GFXReconstruct transition the swapchain images to the correct layout for presentation. This is done using a `VkImageMemoryBarrier`, which cannot be used when the image is not bound to memory yet
(VUID-VkImageMemoryBarrier-image-01932).

A proper fix would be to transition the swapchain images once they are acquired instead of on `vkGetSwapchainImagesKHR`, but this would be harder to implement as we have to make sure all images are transitionned exactly once before their first presentation.

This commit implements an easier workaround that solves the crash by simply disabling the deferred memory allocation of the swapchain images.

### DIFF
--- a/framework/decode/vulkan_virtual_swapchain.cpp
+++ b/framework/decode/vulkan_virtual_swapchain.cpp
@@ -103,6 +103,16 @@ VkResult VulkanVirtualSwapchain::CreateSwapchainKHR(VkResult                    
     modified_create_info.imageUsage =
         modified_create_info.imageUsage | VK_IMAGE_USAGE_TRANSFER_SRC_BIT | VK_IMAGE_USAGE_TRANSFER_DST_BIT;
 
+    if (modified_create_info.flags & VK_SWAPCHAIN_CREATE_DEFERRED_MEMORY_ALLOCATION_BIT_KHR)
+    {
+        // Disable lazy swapchain image allocation because then swapchain images could not be bound to memory yet when
+        // vkGetSwapchainImagesKHR is called. Which means that the layout transition of swapchain images in
+        // VulkanVirtualSwapchain::CreateSwapchainResourceData would be a violation of the spec:
+        // VUID-VkImageMemoryBarrier-image-01932
+        modified_create_info.flags &= ~VK_SWAPCHAIN_CREATE_DEFERRED_MEMORY_ALLOCATION_BIT_KHR;
+        GFXRECON_LOG_INFO("Disabling VK_SWAPCHAIN_CREATE_DEFERRED_MEMORY_ALLOCATION_BIT_KHR in swapchain creation.");
+    }
+
     VkResult result = instance_table_->GetPhysicalDeviceSurfaceCapabilitiesKHR(
         physical_device, create_info->surface, &surfCapabilities);
     GFXRECON_ASSERT(result == VK_SUCCESS);


MarkY-LunarG:
@marius-pelegrin-arm, can you explain what the problem is in a little more detail?  With Virtual Swapchain, we create "virtual swapchain" images separately from the actual swapchain and we return those to the application.  So the images the replay app is using  (those returned by `vkGetSwapchainImages`) should be fully loaded virtual swapchain images.  Then, we only then BLT from the virtual image to the actual swapchain on `vkQueuePresent` (well after that image should have been loaded with `vkAcquireNextImage*`).

marius-pelegrin-arm:
> The issue is that this function, at replay time, is when GFXReconstruct transition the swapchain images to the correct layout for presentation. This is done using a `VkImageMemoryBarrier`, which cannot be used when the image is not bound to memory yet (VUID-VkImageMemoryBarrier-image-01932).

When calling `vkGetSwapchainImagesKHR`, which happens before calls to `vkAcquireNextImageKHR`, the virtual swapchain calls `VulkanVirtualSwapchain::CreateSwapchainResourceData`, which transitions the actual swapchain images to the presentation layout. I don't know how to explain more in detail, just look at this code in the function:

```
            VkImageMemoryBarrier barrier = {
                VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER, // sType
                nullptr,                                // pNext
                VK_ACCESS_NONE,                         // srcAccessMask
                VK_ACCESS_NONE,                         // dstAccessMask
                VK_IMAGE_LAYOUT_UNDEFINED,              // oldLayout
                VK_IMAGE_LAYOUT_PRESENT_SRC_KHR,        // newLayout
                VK_QUEUE_FAMILY_IGNORED,                // srcQueueFamilyIndex
                VK_QUEUE_FAMILY_IGNORED,                // dstQueueFamilyIndex
                VK_NULL_HANDLE,                         // image
                VkImageSubresourceRange{
                    graphics::GetFormatAspects(swapchain_info->format),
                    0,
                    image_create_info.mipLevels,
                    0,
                    image_create_info.arrayLayers,
                }, // subResourceRange
            };

            for (uint32_t i = 0; i < *replay_image_count; ++i)
            {
                barrier.image = swapchain_resources->replay_swapchain_images[i];
                device_table_->CmdPipelineBarrier(command_buffer,
                                                  VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT,
                                                  VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT,
                                                  0,
                                                  0,
                                                  nullptr,
                                                  0,
                                                  nullptr,
                                                  1,
                                                  &barrier);
            }
```

MarkY-LunarG:
I added support to properly handle it in this PR #2688.  Instead of removing it, I simply delay the image transition until the acquire.  This is only needed the first time each index is acquired.

marius-pelegrin-arm:
Great ! If it works, I'm all for that and I will gladly close this PR =)

MarkY-LunarG:
Verified fixed with other PR.  Closing.

PR #2676, opened 2026-02-10
Improved Settings changes round 1

MarkY-LunarG:
This change starts moving to a code generated process for all settings.  This portion of the change introduces the generation process and provides details for the capture-based settings.

The settings are generated using scripts/contents in the `scripts/settings_options` folder.  They are generated using the `generate_settings.py` Python script in that folder which looks at the settings defined in the `settings.json` file.  That file is also compared against the `settings_schema.json` to validate it prior to executing the code/doc generation process.

From these files, the following items are updated/generated:
 - Capture settings headers
 - A new settings struct header is created containing structures pertinent to all capture settings
 - The GFXReconstruct Layer manifest is updated with  newest settings improving the design and layout of settings in VkConfig.
 - The layer settings file is updated to accept all new settings automatically
 - All capture documentation is updated for possible environment variables and settings file options


MarkY-LunarG:
Also introduces a README in `scripts/settings_options` that identifies what files get generated/updated now.

framework/encode/api_capture_manager.cpp
```diff
@@ -29,10 +29,19 @@
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(encode)
 
-CaptureSettings::TraceSettings ApiCaptureManager::GetDefaultTraceSettings()
+ApiCaptureManager::ApiCaptureManager(format::ApiFamilyId api_family) : api_family_(api_family)
 {
-    // Return default trace settings.
-    return CaptureSettings::TraceSettings();
+    capture_settings_ = new CaptureSettings(api_family);
```
ncesario-lunarg: Does `capture_settings_` need to be dynamically allocated? It seems like it could be a member field and then get rid of the need for the custom destructor (and remove the need for this constructor body).

framework/util/CMakeLists.txt
```diff
@@ -134,6 +138,24 @@ target_sources(gfxrecon_util
                     $<$<BOOL:${WAYLAND_FOUND}>:${PROJECT_SOURCE_DIR}/framework/generated/generated_wayland_xdg_shell.h>
 )
 
+# NOTE: This actually generates many files, but the easiest way to get
+#       CMake to automatically trigger this automatically is to associate
+#       it with these source files.  Also `generated_settings_manager.cpp`
+#       should not be pointed to directly by CMake because it is "#include"
+#       directly into the source file.
+if (GENERATE_SETTINGS_OPTIONS)
+    add_custom_command(
```
ncesario-lunarg: Looks like the intention is to associate the `generated_settings` target with this command? Could you use the `COMMAND` parameter of [add_custom_target](https://cmake.org/cmake/help/latest/command/add_custom_target.html#command:add_custom_target) to consolidate this into a single call to `add_custom_target`? If so, that makes it a _little_ clearer from my perspective (it's still cmake though...), but not a big deal either way.

framework/util/generated_settings_manager.cpp
ncesario-lunarg, line 1: Should this be in `framework/generated`? I actually think this naming scheme makes more sense (i.e., put generated files--which are already prefixed with "generated_"--in the directories they are associated with), but it looks like there is no precedent for this.

scripts/settings_options/generate_settings.py
```diff
@@ -0,0 +1,1473 @@
...
+
+        tool_struct_names = []
+        for tool in settings_tools:
+            tool_string = tool.lower().capitalize()
+            tool_struct_name = f"Gfxr{tool_string}SettingsStruct"
+            tool_struct_names.append(
+                f"{tool_struct_name} {tool_string.lower()}_settings")
+            settings_struct_header.write(f"struct {tool_struct_name}\n")
+            settings_struct_header.write("{\n")
+            settings_struct_header.write("    // clang-format off\n")
```
ncesario-lunarg: I think you already established offline that `clang-format` does not run on any file prefixed with `generated_`, so I don't think the addition of any `// clang-format ...` comments are necessary.

framework/util/settings_common.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include "util/defines.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(util)
+GFXRECON_BEGIN_NAMESPACE(settings)
+
+enum GfxrToolType
+{
+    kGfxrToolType_Undefined = 0,
```
ncesario-lunarg: Subjective nit: since this will only be consumed by C++ rather than C, it might be less verbose to use `enum class` here and then reference these via `GfxrToolType::<name>` rather than `kGfxrToolType_<name>` or `GfxrToolType::kGfxrToolType_<name>` (same nit for enum below).

framework/util/settings_manager.cpp
```diff
@@ -0,0 +1,185 @@
...
+        case format::ApiFamily_D3D12:
+            EnableD3D12SettingsDefaults();
+            break;
+        default:
+            break;
+    }
+    ReadEnvironmentVariables();
+}
+
+SettingsManager::~SettingsManager() {}
```
ncesario-lunarg: Remove? Could use `~SettingsManager() = default;` to make it explicit in the header.

scripts/settings_options/README.md
```diff
@@ -0,0 +1,268 @@
...
+attempting to parse the contents of the `settings.json` file.
+
+
+## Dependencies
+
+In order to generate the settings content, you will need to install
+Python3 and then install the appropriate "jsonschema" package.
+
+```bash
+pip install jsonschema
```
ncesario-lunarg, line 20: May want to use a `requirements.txt` to track what's necessary to run the script.

scripts/settings_options/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+The "label" is to be formatted as a short description of the functionality
+enabled by this setting.
+It may be used as a label for the option when/if it shows up in
+the VkConfig GUI section for options related to GFXReconstruct.
+
+For example:
+
+```json
+ "label" : "My New Setting To Rule Them All",
```
ncesario-lunarg, line 128: If I have to add any new settings, this will be my label!

scripts/settings_options/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+Supported Primitive types include:
+- BOOL : Standard boolean type (true/false)
+- ENUM : An enumeration type, requires additional FLAG definitions
+- FLOAT : Standard float type
+- GROUP : This is really a group container type with one or more children
+- INT : Standard int type
+- LIST : A list of items (usually a string)
+- SAVE_FILE : the type is a file that is used to save results somewhere
+- SAVE_FOLDER : the type is a folder that will be used to generate data to
```
ncesario-lunarg, line 164: Is there something special about _SAVE_ file vs a regular file path? Same question for "_FOLDER_".

scripts/settings_options/README.md
```diff
@@ -0,0 +1,268 @@
...
+The primary data type for this setting is defined as a key within the "type" object
+labeled `primitive`.
+
+Supported Primitive types include:
+- BOOL : Standard boolean type (true/false)
+- ENUM : An enumeration type, requires additional FLAG definitions
+- FLOAT : Standard float type
+- GROUP : This is really a group container type with one or more children
+- INT : Standard int type
+- LIST : A list of items (usually a string)
```
ncesario-lunarg: It might be good to state what the format of the list is here (e.g., "comma-separated" or "JSON array").

scripts/settings_options/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+For example:
+
+```json
+            "type": {
+                "primitive": "ENUM",
+                "flags": [
+                    {
+                        "key": "BMP",
+                        "label": "BMP",787acc6470f35f520a0bc25a0efff0469102a2fe
```
ncesario-lunarg: Accidental copy/paste?

ncesario-lunarg (commented):
LGTM! I haven't gone through the python script yet, but will try to get to that sometime tomorrow. The comments I have are largely suggestive and nothing I would block merging over.

scripts/settings_options/README.md
MarkY-LunarG: These are just identifiers that VkConfig defined and I maintained them for now.  It was supposed to indicate to the GUI what kind of dialog to pop up when attempting to populate the values.

MarkY-LunarG: I'll update the comment.

scripts/settings_options/README.md
MarkY-LunarG: We don't have one right now.  Should we?

Right now, I think of 2 items we probably need overall for GFXR and that is:

```
jsonschema==4.10.3
lxml==5.2.1
````

The versions might be different, but that's what's supported on my system.  What do you think?  We should run this by the group, though.

File: framework/util/settings_common.h (hunk 4 above)
Comment (MarkY-LunarG) [reply framework/util/settings_common.h]: Also removed `GfxrSettingPrimitiveType` which was unused.

File: framework/util/generated_settings_manager.cpp
Comment (MarkY-LunarG) [reply framework/util/generated_settings_manager.cpp:1]: This is actually something I want more feedback on.  These generated source files are built from `scripts/settings_options` and yes, I agree that the generated files make more sense  to me in their relevant directory.  If people feel they belong in the `generated` folder, I will also move the script from `scripts/settings` into the generated folder.  The problem I had is none of these items are "framework" specific and it also touches files at the top level as well.

File: framework/util/CMakeLists.txt (hunk 2 above)
Comment (MarkY-LunarG) [reply framework/util/CMakeLists.txt]: Oh, that's much nicer!

File: framework/encode/api_capture_manager.cpp (hunk 1 above)
Comment (MarkY-LunarG) [reply framework/encode/api_capture_manager.cpp]: It could be defined automatically in the class without too much of a difference I guess.  I could go either way.

Comment (bradgrantham-lunarg) [reply framework/encode/api_capture_manager.cpp]: Making it a pointer makes a statement that either the lifetime is not 1:1 or the ownership is passed around.  If the member comes into existence on construction of the containing class and goes away at the end and isn't going to come and go through the lifetime of the containing class, it should just be a member field constructed with the containing class.

Comment (MarkY-LunarG) [reply framework/encode/api_capture_manager.cpp]: I'll just have to make setting the ApiFamilyId a method instead of part of the constructor.

File: USAGE_desktop_Vulkan.md [hunk 11]
```diff
@@ -282,43 +282,53 @@ Options with the BOOL type accept the following values:
...
+| Hotkey Capture Trigger | GFXRECON_CAPTURE_TRIGGER | STRING with possible values of: "", "F1", "F2", "F3", "F4", "F5", "F6", "F7", "F8", "F9", "F10", "F11", "F12", "TAB", "CONTROL" | <empty> | Specify a hotkey (any one of F1-F12, TAB, CONTROL) that will be used to start/stop capture. Example: F3 will set the capture trigger to F3 hotkey. One capture file will be generated for each pair of start/stop hotkey presses.   This, in effect, toggles `capture_dynamic_trigger`. May be the following: <ul>    <li>""  -  None (Default) </li>    <li>f1  -  F1 Key </li>    <li>f2  -  F2 Key </li>    <li>f3  -  F3 Key </li>    <li>f4  -  F4 Key </li>    <li>f5  -  F5 Key </li>    <li>f6  -  F6 Key </li>    <li>f7  -  F7 Key </li>    <li>f8  -  F8 Key </li>    <li>f9  -  F9 Key </li>    <li>f10  -  F10 Key </li>    <li>f11  -  F11 Key </li>    <li>f12  -  F12 Key </li>    <li>tab  -  TAB Key </li>    <li>control  -  CONTROL  Key </li> </ul> |
```
panos-lunarg: > This, in effect, toggles `capture_dynamic_trigger`

Looking at the code, I think this is not true

framework/util/generated_settings_struct.h
panos-lunarg, line 1: Is the purpose here to replace the settings from `framework/encode/capture_settings.h`?
As far as I understand right now it just duplicates the settings

panos-lunarg:
The android dynamic settings are not working. Taking a stab at fixing them with https://github.com/MarkY-LunarG/gfxreconstruct/pull/11

BUILD.md
```diff
@@ -530,3 +531,21 @@ The [Android Vulkan Validation Guide](https://developer.android.com/ndk/guides/g
...
 For more information and detailed examples on building and using GFXReconstruct
 on Android can be found in the [HOWTO_android.md](./HOWTO_android.md) document.
+
+
+## Generating Settings/Options
+
+Because of the many places that settings/options appear throughout the GFXReconstruct
+code base, the settings are all defined in a single JSON and that file is used to
+generate the appropriate usage of those settings throughout the code.
+The relevant files are all located in the `scripts/settingss_options` directory
```
antonio-lunarg: settings**s**_options

scripts/settings_options/generate_settings.py
```diff
@@ -0,0 +1,1468 @@
+#!/usr/bin/python3 -i
+#
+# Copyright (c) 2025 LunarG, Inc.
```
antonio-lunarg: 2025-2026

scripts/settings_options/generate_settings.py
```diff
@@ -0,0 +1,1468 @@
...
+empty_ending_string = "\");\n"
+
+# Strings used to identify Markdown/HTML docs sections that need to be replaced
+# during the codegen phanse
+capture_gen_begin_str = "<!-- CAPTURE_SETTINGS_OPTIONS TABLE CODEGEN BEGIN -->"
+capture_gen_end_str = "<!-- CAPTURE_SETTINGS_OPTIONS TABLE CODEGEN END -->"
+
+generated_source_copyright = f"""
+/*
+ * Copyright (c) 2025 LunarG, Inc.
```
antonio-lunarg: 2025-2026

framework/encode/capture_settings.cpp
```diff
@@ -41,435 +41,193 @@
...
     }
 
-    std::string trim_draw_calls = FindOption(options, kOptionKeyCaptureDrawCalls);
-    if (!trim_draw_calls.empty())
+    if (!settings_struct->capture_settings.capture_draw_calls.empty())
     {
         std::vector<util::UintRange> trim_values;
-        ParseUintRangeList(trim_draw_calls, &trim_values, "capture draw calls", false, true);
-        if (trim_values.size() == 3 || trim_values.size() == 4)
+        ParseUintRangeList(temp_string, &trim_values, "capture draw calls");
```
antonio-lunarg: `temp_string` should actually be `settings_structs->capture_settings.capture_draw_calls`?

framework/encode/capture_settings.cpp
```diff
@@ -41,435 +41,193 @@
...
         {
-            ParseUintRangeList(
-                trim_queue_submits, &settings->trace_settings_.trim_ranges, "capture queue submits", true, true);
-            if (!settings->trace_settings_.trim_ranges.empty())
+            ParseUintRangeList(settings_struct->capture_settings.capture_queue_submits,
+                               &settings->trace_settings_.trim_ranges,
+                               "capture queue submits",
+                               true,
+                               true);
+            if (settings->trace_settings_.trim_ranges.empty())
```
antonio-lunarg: Dropped the `!` by mistake?

framework/util/generated_settings_manager.cpp
ncesario-lunarg: I would vote for keeping it as you have it, but I'm curious what others think.

scripts/settings_options/README.md
ncesario-lunarg: I would say add the requirements.txt to track this, but yeah, needs more input.

framework/util/generated_settings_struct.h
MarkY-LunarG: Yes, this will eventually replace capture settings.  However, it's the first step toward  automating the process for reading and validating all the settings.  A future change will have to move from using the "TraceSettings" and utilizing these directly.  But I wanted to do baby steps (and this is a pretty  sizeable baby step).

PR #2678, opened 2026-02-10
Print shader handle when logging SPIR-V parsing warnings

### LINKED ISSUE #2535: [replay] Print shader handle when logging warning/error
I am getting 

> [gfxrecon] WARNING - Failed to track back the Function Variable OpStore, hit a OpIAdd

I run `gfxrecon-extract` and want to investigate more, it would be MUCH faster (than creating a debug build to break point) to just print out which shader is causing the issue

n-amul:
Fix #2535
Added a shader_handle parameter through the call chain:
ParseBufferReferences -> vulkan_check_buffer_references -> log
All 5 warning messages in spirv_parsing_util.cpp now include the shader handle for easier debugging:
// Before
WARNING - Failed to track back the Function Variable OpStore, hit a OpIAdd
// After
WARNING - Failed to track back the Function Variable OpStore, hit a OpIAdd (shader handle: 0x3a7f)
Existing tests pass without changes (default parameter value of 0).


fabian-lunarg:
what is the motivation for this change, when is printing the "shader-handle" (or rather a pointer) 
helpful?

fabian-lunarg:
it seems you're hitting those warnings to begin with. could you share the related SpirV or point to a case which can be reproduced?

n-amul:
> it seems you're hitting those warnings to begin with. could you share the related SpirV or point to a case which can be reproduced?

Hi, thanks for the review! I'm a new contributor and picked this up from issue #2535. Happy to adjust anything, if I am missing anything I would love to know. I don't have a specific reproducible case to share but, the motivation is that when these warnings appear during replay, there is currently no way to identify which shader caused them. The value printed is the captured handle ID (e.g. from createshadermodule), which can be used with tools like gfxrecon-extract to identify the problematic shader without needing a debug build.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7033,7 +7033,7 @@ VkResult VulkanReplayConsumerBase::OverrideCreateShaderModule(
         if (vk_res == VK_SUCCESS)
         {
             // check for buffer-references, issue warning
-            graphics::vulkan_check_buffer_references(original_info->pCode, original_info->codeSize, shader_module_info);
+            graphics::vulkan_check_buffer_references(original_info->pCode, original_info->codeSize, shader_module_info, (uint64_t)*pShaderModule->GetHandlePointer());
```
fabian-lunarg: this is not correct and got me confused first.

`*pShaderModule->GetHandlePointer()` dereferences to a `VkShaderModule` handle,
followed by a c-style cast.

more likely you want the GFXR 'handle-id' via:
`format::HandleId handle_id = *pShaderModule->GetPointer();`

fabian-lunarg:
> Happy to adjust anything, if I am missing anything I would love to know

that's cool. some things I can say:

- for the time being CI runs (github+LunarG internal) won't start automatically for your commits, as you've seen. simple reason is those CI resources are limited.
- make sure to apply clang-format on the code you've changed. github will perform a check within the GCC Release pipeline. you could check the logs which clang-format version is required and find details about the formatting-issue.
- also LunarG's capabilities for reviewing free contributions are limited, thus best effort and might be delayed. 

thank you!

framework/decode/vulkan_replay_consumer_base.cpp
n-amul: Thank you! I've fixed this to use *pShaderModule->GetPointer() to get the captured handle ID instead. Applied clang-format as well! Sorry for the confusion.

fabian-lunarg:
closing this for a couple of reason. 
- it still contains mixups of shader-module vs. format::HandleId types, both casted to uint64_t, 
so wrong and kind of brittle in the first place.
- logging of gfxr-internal ids is propagated down to the utility-module, while it should rather be the other way around.

while we want to address the underlying issue, 
we also need to account for some edge-cases like inlined spirv (when no `VkShaderModule` ever exists we use a VkPipeline handle_id instead in gfxrecon-extract). 
we should combine this with API-changes for the SpirVParsingUtil, allowing to detect+log errors outside of that class.

n-amul:
Fair points all around, thanks for the clarification! I'll work on this when I get the chance.

PR #2681, opened 2026-02-11
Undo recent change to blocking wait (vs. busy-wait) for vkGetQueryPoolResults

fabian-lunarg:
- should avoid hangs encountered in AAA-title, with non-deterministic aspect

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -71,7 +71,8 @@
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
-const size_t kMaxEventStatusRetries = 16;
+const size_t kMaxEventStatusRetries      = 16;
+const size_t kMaxQueryPoolResultsRetries = 16;
 
 const char kUnknownDeviceLabel[]  = "<Unknown>";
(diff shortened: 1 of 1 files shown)


MarkY-LunarG:
Why was the old code hanging?  Do we know what the apps were doing?  Should we make a TODO to investigate this?

MarkY-LunarG:
@fabian-lunarg / @davidd-lunarg, I would have at least put a comment describing the issue so we don't try this change again.  Something like: "While it may seem like we can simply add `   flags |= VK_QUERY_RESULT_WAIT_BIT;` to force a wait, this may not be valid for multiple queries where one or more of the queries has not been properly signalled."  I think a comment like that may go a long way to us not making this kind of change in the future.

fabian-lunarg:
> @fabian-lunarg / @davidd-lunarg, I would have at least put a comment describing the issue so we don't try this change again. Something like: "While it may seem like we can simply add ` flags |= VK_QUERY_RESULT_WAIT_BIT;` to force a wait, this may not be valid for multiple queries where one or more of the queries has not been properly signalled." I think a comment like that may go a long way to us not making this kind of change in the future.

agreed, I missed to add a hint we already had a back/forth around the right approach there.
I think in my head, still love to have a cleaner solution here, so didn't go a long way to defend the busy-wait (but ack it's our current best solution) :)
maybe that solution would just fall out automatically with a multi-threaded replay.

PR #2685, opened 2026-02-11
update test suite ref

bradgrantham-lunarg:


bradgrantham-lunarg:
Canceled Jenkins CI run because change is to extended suite.  Judgement call about test suite ref - change is only subsetting extended suite, will merge.

PR #2688, opened 2026-02-12
Support deferred virtual swapchain

MarkY-LunarG:
If deferred swapchain was enabled, the virtual swapchain would incorrectly attempt to transition invalid images.  According to the spec, the images are not actually allocated until the first call to vkAcquireNextImage* for that given index.
So, delay the transition of the image until it is valid.

### DIFF
--- a/framework/decode/vulkan_virtual_swapchain.cpp
+++ b/framework/decode/vulkan_virtual_swapchain.cpp
@@ -254,76 +254,69 @@ VkResult VulkanVirtualSwapchain::CreateSwapchainResourceData(const VulkanDeviceI
     props.resize(property_count);
     instance_table_->GetPhysicalDeviceQueueFamilyProperties(device_info->parent, &property_count, props.data());
 
-    if (!found_copy_queue_family_.contains(device))
+    uint32_t copy_queue_family_index           = VK_QUEUE_FAMILY_IGNORED;
+    bool     found_copy_queue_family_index     = false;
+    bool     found_transfer_queue_family_index = false;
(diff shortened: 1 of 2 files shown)


MarkY-LunarG:
I tested this by forcing deferred swapchain usage in several captures during replay on my AMD Linux system.

framework/decode/vulkan_virtual_swapchain.h
```diff
@@ -207,6 +215,10 @@ class VulkanVirtualSwapchain : public VulkanSwapchain
     };
 
     std::unordered_map<VkDevice, OFBData> ofb_data_;
+
+    std::unordered_map<VkDevice, bool>     found_copy_queue_family_; // false
```
fabian-lunarg: `found_copy_queue_family_` is only populated with `found_copy_queue_family_[device] = true` and only checked with `found_copy_queue_family_.contains(device)`,

so could this just be a `std::unordered_set<VkDevice>` ?

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -462,7 +481,7 @@ VkResult VulkanVirtualSwapchain::CreateSwapchainResourceData(const VulkanDeviceI
             nullptr,                                                            // pNext
             0,                                                                  // flags
             VK_IMAGE_TYPE_2D,                                                   // imageType
-            swapchain_info->format,                                             // format
+            swapchain_info->format,                                             // formatcopy_queue_family_index
```
fabian-lunarg: typo?

PR #2690, opened 2026-02-13
Add missing EndInjectedCommands

ziga-lunarg:
Some early returns were missing `util::EndInjectedCommands();`


fabian-lunarg:
nice! we should really switch usage of those to using a RAII-helper.

PR #2693, opened 2026-02-16
Fix DGC decoding

ziga-lunarg:
`VkIndirectExecutionSetInfoEXT` requires special handling, because it can use either `pPipelineInfo` or `pShaderInfo`


ziga-lunarg:
Closing in favour of #2794

PR #2696, opened 2026-02-17
Optimize gfxrecon-optimize tool

### LINKED ISSUE #2436: delete unused resources from optimized traces
optimizing traces should eliminate all unsubmitted command buffers, which should include:
* primary command buffers which are never submitted
* secondary command buffers which are executed only in primaries that are never submitted

fabian-lunarg:
currently `gfxrecon-optimize` does two things:
- removes unused `format::MetaDataType::kInitBufferCommand` and unused `format::MetaDataType::kInitImageCommand`
- only trimmed captures can be optimized

-> this does already account for the bulk of possible file-size reduction, which was the primary (only) benefit when using `gfxrecon-optimize`

this PR builds on top of existing resource-usage-tracking and adds additional capabilities:
- removes unused `VkCommandBuffer` (primary&secondary) incl. all blocks recording commands into those
- removes unused `VkPipeline`, incl. all calls to `VkCreateXXXPipelines`
- "removing" a block is done by replacing it with a label-block. this is done to keep block-indices intact and matches the previous behavior
- **trimmed and non-trimmed** captures can be processed by `gfxrecon-optimize`

closes #2436

### DIFF
--- a/framework/generated/generated_vulkan_referenced_block_consumer.cpp
+++ b/framework/generated/generated_vulkan_referenced_block_consumer.cpp
@@ -1,6 +1,6 @@
 /*
 ** Copyright (c) 2018-2023 Valve Corporation
-** Copyright (c) 2018-2023 LunarG, Inc.
+** Copyright (c) 2018-2026 LunarG, Inc.
 ** Copyright (c) 2023 Advanced Micro Devices, Inc.
 **
 ** Permission is hereby granted, free of charge, to any person obtaining a
--- a/framework/generated/generated_vulkan_referenced_resource_consumer.cpp
+++ b/framework/generated/generated_vulkan_referenced_resource_consumer.cpp
@@ -1,6 +1,6 @@
 /*
 ** Copyright (c) 2018-2023 Valve Corporation
-** Copyright (c) 2018-2023 LunarG, Inc.
+** Copyright (c) 2018-2026 LunarG, Inc.
 ** Copyright (c) 2023 Advanced Micro Devices, Inc.
 **
 ** Permission is hereby granted, free of charge, to any person obtaining a
--- a/framework/generated/generated_vulkan_api_call_encoders.cpp
+++ b/framework/generated/generated_vulkan_api_call_encoders.cpp
@@ -1,6 +1,6 @@
 /*
 ** Copyright (c) 2018-2023 Valve Corporation
-** Copyright (c) 2018-2023 LunarG, Inc.
+** Copyright (c) 2018-2026 LunarG, Inc.
 ** Copyright (c) 2023 Advanced Micro Devices, Inc.
 **
 ** Permission is hereby granted, free of charge, to any person obtaining a
--- a/framework/generated/generated_vulkan_api_call_encoders.h
+++ b/framework/generated/generated_vulkan_api_call_encoders.h
@@ -1,6 +1,6 @@
 /*
 ** Copyright (c) 2018-2023 Valve Corporation
-** Copyright (c) 2018-2023 LunarG, Inc.
+** Copyright (c) 2018-2026 LunarG, Inc.
 ** Copyright (c) 2023 Advanced Micro Devices, Inc.
 **
 ** Permission is hereby granted, free of charge, to any person obtaining a
--- a/framework/generated/generated_vulkan_command_buffer_util.cpp
+++ b/framework/generated/generated_vulkan_command_buffer_util.cpp
@@ -1,6 +1,6 @@
 /*
 ** Copyright (c) 2018-2023 Valve Corporation
-** Copyright (c) 2018-2023 LunarG, Inc.
+** Copyright (c) 2018-2026 LunarG, Inc.
 ** Copyright (c) 2023 Advanced Micro Devices, Inc.
 **
 ** Permission is hereby granted, free of charge, to any person obtaining a
--- a/framework/generated/generated_vulkan_command_buffer_util.h
+++ b/framework/generated/generated_vulkan_command_buffer_util.h
@@ -1,6 +1,6 @@
 /*
 ** Copyright (c) 2018-2023 Valve Corporation
-** Copyright (c) 2018-2023 LunarG, Inc.
+** Copyright (c) 2018-2026 LunarG, Inc.
 ** Copyright (c) 2023 Advanced Micro Devices, Inc.
 **
 ** Permission is hereby granted, free of charge, to any person obtaining a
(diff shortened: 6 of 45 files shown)


framework/generated/generated_vulkan_referenced_resource_consumer.cpp
```diff
@@ -2078,17 +2106,62 @@ void VulkanReferencedResourceConsumer::Process_vkCmdDrawClusterIndirectHUAWEI(
...
+}
+
 void VulkanReferencedResourceConsumer::Process_vkCmdPreprocessGeneratedCommandsEXT(
     const ApiCallInfo&                          call_info,
     format::HandleId                            commandBuffer,
     StructPointerDecoder<Decoded_VkGeneratedCommandsInfoEXT>* pGeneratedCommandsInfo,
     format::HandleId                            stateCommandBuffer)
 {
-    GFXRECON_UNREFERENCED_PARAMETER(pGeneratedCommandsInfo);
+    assert(pGeneratedCommandsInfo != nullptr);
```
MarkY-LunarG: GFXRECON_ASSERT?

framework/generated/generated_vulkan_referenced_block_consumer.cpp
```diff
@@ -0,0 +1,3360 @@
+/*
+** Copyright (c) 2018-2023 Valve Corporation
+** Copyright (c) 2018-2023 LunarG, Inc.
+** Copyright (c) 2023 Advanced Micro Devices, Inc.
+**
```
MarkY-LunarG, line 5: Nit: The LunarG date should be extended to 2026 in these copyrights.

framework/generated/khronos_generators/vulkan_generators/vulkan_referenced_block_consumer_body_generator.py
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/python3 -i
+#
+# Copyright (c) 2020 LunarG, Inc.
```
MarkY-LunarG: This file is brand new, so 2026.

framework/generated/khronos_generators/vulkan_generators/vulkan_referenced_block_consumer_body_generator.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+
+    # All resource and resource associated handle types to be processed.
+    RESOURCE_HANDLE_TYPES = [
+        'VkBuffer', 'VkImage', 'VkBufferView', 'VkImageView', 'VkFramebuffer',
+        'VkDescriptorSet', 'VkCommandBuffer', 'VkAccelerationStructureKHR', 'VkPipeline'
+    ]
+
+    # Handle types that contain resource and child resource handle types.
+    CONTAINER_HANDLE_TYPES = ['VkDescriptorSet']
```
MarkY-LunarG: This does not appear used.

framework/generated/khronos_generators/vulkan_generators/vulkan_referenced_block_consumer_body_generator.py
```diff
@@ -0,0 +1,147 @@
...
+
+    # Handle types that use resource and child resource handle types.
+    USER_HANDLE_TYPES = ['VkCommandBuffer']
```
MarkY-LunarG: Also not used?

framework/decode/referenced_resource_table.cpp
```diff
@@ -392,11 +384,10 @@ void ReferencedResourceTable::ProcessUserSubmission(format::HandleId user_id)
                 }
             }
 
-            for (auto& container_info : container_infos)
+            for (auto& container_info : user_info->container_infos)
             {
-                if (!container_info.second.expired())
+                if (auto container_info_ptr = container_info.second.lock())
```
MarkY-LunarG, line 399: I think this isn't completely clear.  It looks like you're performing a mutex lock here.  If it's doing something else, a comment would be nice.  (and above on line 375)

MarkY-LunarG (commented):
Some questions, also are we going to provide tests for optimize?

framework/decode/referenced_resource_table.cpp
fabian-lunarg: it's std::weak_ptr<>::lock() here and above, I'll add comments

framework/generated/khronos_generators/vulkan_generators/vulkan_referenced_block_consumer_body_generator.py
fabian-lunarg: yes, started with a c/p, will add a cleanup-commit for the generator

fabian-lunarg:
> Some questions, also are we going to provide tests for optimize?

apparently we already have some (in LunarG internal CI, both PR+extended), 
but have to confirm this (are they executed, sufficient, etc.)

framework/generated/khronos_generators/vulkan_generators/vulkan_referenced_block_consumer_body_generator.py
fabian-lunarg: done

framework/generated/khronos_generators/vulkan_generators/vulkan_referenced_block_consumer_body_generator.py
fabian-lunarg: removed

framework/generated/khronos_generators/vulkan_generators/vulkan_referenced_block_consumer_body_generator.py
fabian-lunarg: done

framework/generated/generated_vulkan_referenced_block_consumer.cpp
fabian-lunarg: agreed. only I had trouble finding the source of that license-header? it appears to be the same for all generated vulkan code, but grep'ing for that text only revealed generated files

framework/generated/generated_vulkan_referenced_resource_consumer.cpp
fabian-lunarg: done

framework/decode/referenced_resource_table.cpp
fabian-lunarg: added comments: `// attempt to lock a std::weak_ptr`
to clarify what's done there.

framework/generated/generated_vulkan_referenced_block_consumer.cpp
fabian-lunarg: found it, it was in `khronos_generators/vulkan_generators/gencode.py`
done

framework/decode/referenced_resource_table.cpp
MarkY-LunarG: Thanks.  It might be clear to most people (and not just me), but I wasn't sure what was going on there so best to avoid confusion.

PR #2698, opened 2026-02-17
Initial frame looping

antonio-lunarg:
This PR introduces experimental frame looping for replay, so a selected frame can be replayed repeatedly for analysis/profiling.

Changes:
- Add `--loop-frame` and `--loop-count` to replay options on desktop and Android, plus forwarding in `android/scripts/gfxrecon.py`.
- Add `graphics::FrameLoopInfo` to track target frame, remaining iterations, and current looping state.
-Loop orchestration lives in `Application::Run()` so loop behavior follows application frame numbering (including trimmed traces).
- Use `PreloadFileProcessor` automatically when looping is enabled, preload the target frame, and replay it by controlling advance (`SetAdvanceToNextFrame`).
- Skip state blocks during loop iterations to avoid reapplying state setup every repeat.
- Extend `VulkanFrameLoopReplayConsumer` to skip recreating already-created objects while looping (instance/device/swapchain/surfaces/AHB path).
- Default `pause_frame` to `uint32_t max` to avoid accidental pause at frame 0 when looping starts.

Notes:
- This functionality is currently experimental.
- Implemented in the Vulkan replay path.

### DIFF
--- a/framework/graphics/frame_loop_info.h
+++ b/framework/graphics/frame_loop_info.h
@@ -33,10 +33,10 @@ GFXRECON_BEGIN_NAMESPACE(graphics)
 class FrameLoopInfo
 {
   public:
-    explicit FrameLoopInfo(uint32_t loop_frame_idx  = 0,
-                           uint32_t loop_iterations = std::numeric_limits<uint32_t>::max()) :
-        loop_frame_idx_{ loop_frame_idx },
-        loop_iterations_{ loop_iterations }
+    constexpr static uint32_t INFINITE_ITERATIONS = std::numeric_limits<uint32_t>::max();
+
+    explicit FrameLoopInfo(uint32_t loop_frame_idx = 0, uint32_t loop_iterations = INFINITE_ITERATIONS) :
+        loop_frame_idx_{ loop_frame_idx }, loop_iterations_{ loop_iterations }
     {}
 
     /// Expects a 1-based frame number.
@@ -50,18 +50,27 @@ class FrameLoopInfo
 
     /// Returns true if currently looping on the loop frame, meaning that the loop frame
     /// has already been played at least once and we are currently replaying it again.
-    bool     IsLooping() const { return is_looping_; }
+    bool IsLooping() const { return is_looping_; }
 
     void     SetLooping(bool looping) { is_looping_ = looping; }
     uint32_t GetLoopFrameIdx() const { return loop_frame_idx_; }
     uint32_t GetLoopIterations() const { return loop_iterations_; }
-    void     DecrementLoopIterations() { loop_iterations_--; }
+
+    /// Decrements the number of loop iterations remaining.
+    /// If the number of iterations is infinite, this has no effect.
+    void DecrementLoopIterations()
+    {
+        if (loop_iterations_ != INFINITE_ITERATIONS)
+        {
+            --loop_iterations_;
+        }
+    }
 
   private:
     uint32_t current_frame_number_{ 0 };
     bool     is_looping_{ false };
     uint32_t loop_frame_idx_{ 0 };
-    uint32_t loop_iterations_{ std::numeric_limits<uint32_t>::max() };
+    uint32_t loop_iterations_{ INFINITE_ITERATIONS };
 };
 
 GFXRECON_END_NAMESPACE(graphics)
--- a/tools/replay/android_main.cpp
+++ b/tools/replay/android_main.cpp
@@ -145,9 +145,8 @@ void android_main(struct android_app* app)
 
         try
         {
-
             uint32_t loop_frame        = 0;
-            uint32_t loop_count        = std::numeric_limits<uint32_t>::max();
+            uint32_t loop_count        = gfxrecon::graphics::FrameLoopInfo::INFINITE_ITERATIONS;
             bool     enable_frame_loop = GetLoopFrame(arg_parser, loop_frame);
             GetLoopCount(arg_parser, loop_count);
 
--- a/tools/replay/desktop_main.cpp
+++ b/tools/replay/desktop_main.cpp
@@ -141,7 +141,7 @@ int main(int argc, const char** argv)
         std::unique_ptr<gfxrecon::decode::FileProcessor> file_processor;
 
         uint32_t loop_frame        = 0;
-        uint32_t loop_count        = std::numeric_limits<uint32_t>::max();
+        uint32_t loop_count        = gfxrecon::graphics::FrameLoopInfo::INFINITE_ITERATIONS;
         bool     enable_frame_loop = GetLoopFrame(arg_parser, loop_frame);
         GetLoopCount(arg_parser, loop_count);
 


tools/tool_settings.h
```diff
@@ -82,6 +82,8 @@ const char kOverrideGpuArgument[]                = "--gpu";
 const char kOverrideGpuGroupArgument[]           = "--gpu-group";
 const char kPausedOption[]                       = "--paused";
 const char kPauseFrameArgument[]                 = "--pause-frame";
+const char kFrameLoopArgument[]                  = "--frame-loop";
```
panos-lunarg: Nitpicking: would it make sense to introduce only one new setting with similar syntax like `GFXRECON_CAPTURE_FRAMES`? I.e. `--loop-frames N` for a single frame and `--loop-frames N-(N+x)` for frame range? No strong opinion on this, just for uniformity with other options

framework/application/application.cpp
```diff
@@ -167,6 +186,24 @@ void Application::Run()
             // PlaySingleFrame() increments this->current_frame_number_ *if* there's an end-of-frame
             PlaySingleFrame();
 
+            if (frame_loop_info_ != nullptr)
+            {
+                frame_loop_info_->SetLooping(at_loop_frame);
```
nickdriscoll-lunarg: Because we set the looping flag after the call to `PlaySingleFrame()`, the consumer will replay the looping frame one time without knowing that it is on the looping frame. This causes a problem when tracking resources like descriptor sets, as the very first set of allocations will not be tracked by the logic in the new consumer.

In my branch where I fix the `vkAllocateDesriptorSets()` validation error, I had to hack in two extra methods in `FrameLoopInfo` in order for the consumer to know if it's on the very first iteration of the looping frame.

I know that the overridden `Process_vkCreate*SurfaceKHR()` methods rely on the current behavior, but maybe the `FrameLoopInfo` could permanently remember the requested number of iterations so that the consumer can query which iteration it's on? (e.g. `uint32_t current_iteration = info.total_loop_iterations_ - info.loop_iterations_;`

antonio-lunarg: I introduced two new methods in `FrameLoopInfo` that might be useful for your case:
```cpp
    /// Returns true if the current frame number matches the loop frame index.
    bool AtLoopFrame() const { return current_frame_number_ == loop_frame_idx_; }

    /// Returns true if this is the first iteration of the loop frame.
    bool IsFirstIteration() const { return AtLoopFrame() && !IsLooping(); }
```

framework/decode/preload_file_processor.cpp
```diff
@@ -40,6 +40,10 @@ void PreloadFileProcessor::PreloadNextFrames(size_t count)
         error_state_ = CheckFileStatus();
         return;
     }
+    if (!advance_to_next_frame_)
```
jzulauf-lunarg: This feels more "contractual".  Thinking assert in additional to check.  The frame looping support state machine only current loads a single frame and only when advance_to_next_frame_ is enabled.

framework/application/application.cpp
```diff
@@ -140,6 +141,27 @@ void Application::Run()
...
             uint32_t frame_number = file_processor_->GetCurrentFrameNumber() + 1;
 
+            if (frame_loop_info_ != nullptr)
+            {
+                frame_loop_info_->SetCurrentFrameNumber(frame_number);
+
+                // Quit when frame looping has finished.
+                if (frame_loop_info_->GetLoopIterations() == 0)
+                {
+                    running_ = false;
```
jzulauf-lunarg, line 151: Is this because 

a) we never want to continue past looping frame or 
b) because we aren't confident that replay past looping is stable? (or are sure it isn't)

--quit-after-measurement-range controls this behavior for FpsInfo.. don't know if a separate --quit-after-looping would be needed.

framework/decode/preload_file_processor.cpp
jzulauf-lunarg, line 60: This mode means that we store the minimum needed.  For uncompressed block, we store the "raw block" information and have the DispatchArgs directly reference it.  However for compressed block (combined with BlockParser::DecompressionPolicy::kAlways below) we only the decompressed data and not the raw block data from the file.

jzulauf-lunarg, line 78: In order to support frame looping BlockParser::DecompressionPolicy::kAlways isn't just a performance optimization for accurate measurement *required* to prevent the the issue of having dispatch_args data members pointing at the most recently decompressed data (or stale pointers) on replay.

Probably needs some commentary about this.

jzulauf-lunarg (commented):
Preload section looks good, but I'm going to have to integrate this support in PR #2783 ... @antonio-lunarg  -- if frame looping is going to change significantly, we should probably carefully align the changes so nothing gets dropped.

framework/application/application.cpp
antonio-lunarg: a) I think we don't really care about continuing after looping, at the moment.
b) But I think replay past looping should work.

I suppose `--quit-after-looping` might be added later if someone wants it.

framework/decode/preload_file_processor.cpp
antonio-lunarg: For frame looping, I think we only care about being able to replay blocks. For compressed blocks, I would be happy do uncompress once and use the decompressed data later again and again.

antonio-lunarg: I added a comment.

framework/decode/preload_file_processor.cpp
antonio-lunarg: I added an assertion and integrated with more documentation on top of this function's signature.

framework/application/application.cpp
bradgrantham-lunarg: Then the someone adding `--quit-after-looping` would also wrap `running_ = false;` and `break` with e..g. `if(frame_loop_info_->quit_after_looping)`?

framework/application/application.cpp
```diff
@@ -167,6 +189,27 @@ void Application::Run()
...
+                {
+                    auto* preload_processor = dynamic_cast<decode::PreloadFileProcessor*>(file_processor_);
+                    GFXRECON_ASSERT(preload_processor)
+                    preload_processor->WaitDecodersIdle();
+
+                    // When replaying a frame again, skip any state blocks to avoid reapplying them again.
+                    preload_processor->SkipStateBlocks();
+
+                    GFXRECON_LOG_INFO("Looping frame (%i iterations remaining)", frame_loop_info_->GetLoopIterations());
+                    frame_loop_info_->DecrementLoopIterations();
```
bradgrantham-lunarg, line 203: Are the loop iterations abstracted behind methods as a design principle or do you have an expectation here that this may be more than just `--` at some point?

framework/application/application.h
```diff
@@ -98,6 +101,7 @@ class Application final
     decode::FileProcessor*                                       file_processor_;    ///< The FileProcessor object responsible for decoding and processing capture file data.
     bool                                                         running_;           ///< Indicates that the application is actively processing system events for playback.
     bool                                                         paused_;            ///< Indicates that the playback has been paused.  When paused the application will stop rendering, but will continue processing system events.
+    graphics::FrameLoopInfo*                                     frame_loop_info_;   ///< Indicates that playback wishes to loop a certain frame
```
bradgrantham-lunarg, line 115: I don't want to cause this PR to be retested, so don't change it for this PR, but for changes in the future and for a possible edit to looping in a future PR, I think this would be better (since it's just a few uint32's and there are only 1 `Application`s) to embed frame looping info or other Application parameters directly in `Application` with defaults that mean "no frame looping" to avoid a potential bug in future PRs not checking "frame_loop_info != nullptr".

antonio-lunarg: I actually have been talking with @jzulauf-lunarg about that. While `FrameLoopInfo` takes inspiration from `FpsInfo`, I noticed `FpsInfo` is quite useful for things related to frame-looping and I was also considering merging these two together into something like `FrameInfo`, focused on frames rather than the concept of frames-per-second.

framework/decode/preload_file_processor.cpp
```diff
@@ -176,29 +189,70 @@ bool PreloadFileProcessor::ProcessNextFrame()
...
 
     ProcessBlockState process_result = ReplayOneFrame();
+    return AdvanceToNextFrame(process_result);
+}
 
-    const bool at_end = (!replay_cursor_);
-    if (at_end)
+bool PreloadFileProcessor::AdvanceToNextFrame(ProcessBlockState process_result)
+{
+    if (advance_to_next_frame_)
```
bradgrantham-lunarg: It seems like "whether to advance to next frame" is an application preference, but it's reaching its fingers into `PreloadFileProcessor` to do it.  That is to say, `AdvanceToNextFrame` and maybe "move replay cursor" primitive make sense, but what would it mean to call `AdvanceToNextFrame` but have `advance_to_next_frame_` be `false`?  Why not have `advance_to_next_frame_` be in the caller, and then whether to call `PreloadNextFrames` and `AdvanceToNextFrame` and "move replay cursor" would be up to the caller of `FilePreloadProcessor`?

The precedent for this is that there is already a `SkipStateBlocks` that controls playback and the cursor that is a primitive an application may choose to call, and then whether to skip isn't tracked by `FilePreloadProcessor`.

framework/application/application.cpp
antonio-lunarg: I think we can already wrap a couple of these lines within a `frame_loop_info->EndFrame()`.
Or we might even wrap all of them by passing the `file_processor` as a parameter of `EndFrame(..)`.
What do you think?

framework/application/application.h
antonio-lunarg: I created an issue for this: https://github.com/LunarG/gfxreconstruct/issues/2810

framework/decode/preload_file_processor.cpp
antonio-lunarg: Please have a look at the last two commits: I replaced the `advance_to_next_frame_` mode bit with two explicit functions: `ReplayCurrentPreloadedFrame()` and `AdvancePreloadedFrame()`. When looping, Application calls only the replay function and skips the advance, keeping the cursor on the same preloaded frame for every iteration.

PR #2702, opened 2026-02-17
Covert DX12 enum string

locke-lunarg:
DX12 enum ToString dismatch `template<> ToString` in `to_string.h`.  It caused that Convert wrote int value of enum, instead of name of enum.  This PR makes these two ToString match, as generated_vulkan_enum_to_string and generated_openxr_enum_to_string.


bradgrantham-lunarg:
Can we add a capture that shows that behavior as part of this test or add the "convert" subtest to an existing capture file to get some coverage out of testing convert on DX12?

bradgrantham-lunarg:
Please make sure this goes in with a test case that is compared against a golden convert reference to have regression testing going forward.  Check with @beau-lunarg on getting that in or with me when I get back.

bradgrantham-lunarg (approved):

locke-lunarg:
@beau-lunarg  The test case has a .gfxr file (D3D12HelloTexture) and a .json file. The test converts the .gfxr file and ensure that both of the new .json file and the original .json files are the same. Is it good?

But because the .json has version, branch, env vars, path, device info, it has to skip comparing something like that.

bradgrantham-lunarg:
There is a "convert" subtest, and a key in the suite JSON for the reference
convert JSON filename, which should be put on the NAS.  There are other
examples already in the suite files.

Brad Grantham
LunarG, Inc. - 3D Graphics Software Solutions
***@***.***

beau-lunarg:
> @beau-lunarg The test case has a .gfxr file (D3D12HelloTexture) and a .json file. The test converts the .gfxr file and ensure that both of the new .json file and the original .json files are the same. Is it good?
> 
> But because the .json has version, branch, env vars, path, device info, it has to skip comparing something like that.

This code here can filter out irrelevant info that you don't want json compared
https://github.com/LunarG/gfxreconstruct/blob/dev/test/verify-gfxr.cpp#L11-L68

locke-lunarg:
@beau-lunarg  Thanks!

PR #2713, opened 2026-02-20
Frame Warm Up

antonio-lunarg:
The `--frame-warm-up-load` option controls a synthetic compute warm-up pass submitted before replay frame work. This helps ramp/stabilize GPU frequency before each frame so performance measurements are less sensitive to clock ramp-up latency.

`frame_warm_up_load` is a workload scale factor: higher values increase warm-up work by scaling compute dispatch workgroups, while `0` keeps warm-up disabled.

### DIFF
--- a/USAGE_android.md
+++ b/USAGE_android.md
@@ -1009,9 +1009,12 @@ options:
                         processing the first submit. (forwarded to replay tool)
   --idle-before-submit  Wait for the GPU to become idle before each submit.
   --frame-warm-up-spirv DEVICE_FILE
-                        Specify a SPIR-V file for the compute warm-up pass.
-                        Warm-up runs only when this option and a non-zero
-                        --frame-warm-up-load are both provided.
+                        Specify a user-provided SPIR-V compute shader for the
+                        warm-up pass. The shader must use entry point `main`
+                        and set 0, binding 0 as a storage buffer. Warm-up runs
+                        before the first submit of each replayed frame only
+                        when this option and a non-zero --frame-warm-up-load
+                        are both provided.
                         (forwarded to replay tool)
   --frame-warm-up-load LOAD
                         Specify workload scale factor for a compute dispatch warm-up pass
@@ -1077,6 +1080,61 @@ command would be:
 
 This would result in the `gfxrecon.py` Python script first pushing up the file
 to the requested location and then starting the replay.
+
+#### Frame Warm-Up
+
+Android replay forwards `--frame-warm-up-spirv` and `--frame-warm-up-load` to
+the replay tool. The shader contract is the same as on desktop:
+
+- The file must contain valid SPIR-V for a compute shader.
+- The entry point must be named `main`.
+- The shader must be compatible with set `0`, binding `0` as a storage buffer.
+- The shader must not require additional descriptors or push constants.
+- Warm-up runs before the first submit of each replayed frame.
+
+The shader is provided as an external file on purpose. Warm-up workloads are
+highly GPU- and driver-dependent, so replay does not embed a default shader and
+instead lets the user supply one that is appropriate for the target device.
+
+Example shader source:
+
+```glsl
+#version 450
+
+layout(local_size_x = 64, local_size_y = 1, local_size_z = 1) in;
+
+layout(std430, set = 0, binding = 0) buffer DataBuffer {
+  float data[];
+};
+
+void main() {
+  uint index = gl_GlobalInvocationID.x;
+  float value = data[index];
+
+  for (int i = 0; i < 1000; ++i) {
+    value = sin(value) * 0.999 + cos(float(i)) * 0.001;
+  }
+
+  data[index] = value;
+}
+```
+
+Compile it to SPIR-V on the host:
+
+```bash
+glslangValidator -V -S comp frame_warm_up.comp -o frame_warm_up.spv
+```
+
+Push the resulting SPIR-V file to the device and use it for replay:
+
+```bash
+adb push frame_warm_up.spv /sdcard/Download/frame_warm_up.spv
+
+./android/scripts/gfxrecon.py replay \
+  --frame-warm-up-spirv /sdcard/Download/frame_warm_up.spv \
+  --frame-warm-up-load 4 \
+  /sdcard/Download/android_capture.gfxr
+```
 
 
 ### Touch Controls
--- a/USAGE_desktop_Vulkan.md
+++ b/USAGE_desktop_Vulkan.md
@@ -879,13 +879,96 @@ Optional arguments:
   --serialize-render-passes
               Serialize render passes by injecting execution barriers before render pass begin during replay
   --frame-warm-up-spirv <spirv-file>
-              Specify a SPIR-V file for the compute warm-up pass.
-              Warm-up runs only when this option and a non-zero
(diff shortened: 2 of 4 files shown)


framework/graphics/vulkan_frame_warm_up.h
```diff
@@ -0,0 +1,71 @@
...
+    void WarmUp(VkDevice                                device,
+                const VkPhysicalDeviceMemoryProperties& capture_memory_properties,
+                uint32_t                                queue_family_index,
+                VkQueue                                 queue,
+                const VulkanDeviceTable*                device_table);
+
+  private:
+    uint32_t warm_up_load_{ 0 };
+
+    VkCommandPool            command_pool_{ VK_NULL_HANDLE };
```
panos-lunarg: I think other places are creating very similar temporary Vulkan objects in order to submit command buffers for execution via a QueueSubmit. Resource initializer, virtual swapchain, screenshot handler, dump resources and address replacer comes to mind without checking. I wonder if it would be a good idea to create a utility class that handles creation and destruction of such objects (of course as a separate PR and not as part of this)

framework/decode/vulkan_frame_warm_up.cpp
```diff
@@ -0,0 +1,358 @@
...
+        //     // Perform a series of arbitrary calculations in a loop to keep the GPU busy.
+        //     // This is designed to be work that the compiler can't easily optimize away.
+        //     for (int i = 0; i < 1000; i++) {
+        //         value = sin(value) * 0.999 + cos(float(i)) * 0.001;
+        //     }
+
+        //     // Write the final, modified value back into the same position in the buffer.
+        //     data[index] = value;
+        // }
+        unsigned char spirv[] = {
```
nickdriscoll-lunarg: nitpick: `uint8_t` instead of `char`?

framework/decode/vulkan_frame_warm_up.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+            VkMemoryAllocateInfo alloc_info = { VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO };
+            alloc_info.allocationSize       = mem_requirements.size;
+            bool found                      = FindMemoryTypeIndex(capture_memory_properties,
+                                             mem_requirements.memoryTypeBits,
+                                             VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT,
+                                             &alloc_info.memoryTypeIndex,
+                                             nullptr);
+
+            if (alloc_info.memoryTypeIndex != std::numeric_limits<uint32_t>::max())
```
nickdriscoll-lunarg: Replacing this expression with `if (found)` would be clearer and avoid relying on a sentinel value.

framework/decode/vulkan_frame_warm_up.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+    VkResult result = VK_SUCCESS;
+
+    if (command_pool_ == VK_NULL_HANDLE)
+    {
+        VkCommandPoolCreateInfo cmd_pool_info = { VK_STRUCTURE_TYPE_COMMAND_POOL_CREATE_INFO };
+        cmd_pool_info.queueFamilyIndex        = queue_family_index;
+        cmd_pool_info.flags                   = VK_COMMAND_POOL_CREATE_TRANSIENT_BIT;
+        result = device_table->CreateCommandPool(device, &cmd_pool_info, nullptr, &command_pool_);
+        GFXRECON_LOG_WARNING("CreateCommandPool %s.", util::ToString<VkResult>(result).c_str());
```
nickdriscoll-lunarg: I feel like most 'warning' messages in this method can be 'info' or even 'debug'. Most are not actually warning the user about something.

tools/tool_settings.h
```diff
@@ -893,6 +894,23 @@ static void GetWaitBeforeFirstSubmit(const gfxrecon::util::ArgumentParser& arg_p
...
+
+    if (!value.empty())
+    {
+        try
+        {
+            frame_warm_up_load = std::stoul(value);
+        }
+        catch (std::exception&)
+        {
+            GFXRECON_LOG_WARNING("Ignoring invalid frame warm up load option. Expected format is unsigned integer");
```
nickdriscoll-lunarg: Maybe print the invalid value as part of this warning message?

framework/graphics/vulkan_frame_warm_up.h
antonio-lunarg: I think one good step has been made with the introduction of `VulkanSubmitJob`. I believe there's room for improvement though, therefore I created an issue to keep track of that: https://github.com/LunarG/gfxreconstruct/issues/2796

framework/decode/vulkan_frame_warm_up.cpp
```diff
@@ -0,0 +1,410 @@
...
+        semaphore_values[i]  = wait_semaphores[i].timeline_value;
+    }
+
+    VkTimelineSemaphoreSubmitInfo timeline_info = {};
+    timeline_info.sType                         = VK_STRUCTURE_TYPE_TIMELINE_SEMAPHORE_SUBMIT_INFO;
+    timeline_info.waitSemaphoreValueCount       = wait_semaphores.size();
+    timeline_info.pWaitSemaphoreValues          = wait_semaphores.empty() ? nullptr : semaphore_values.data();
+
+    VkSubmitInfo submit_info = { VK_STRUCTURE_TYPE_SUBMIT_INFO };
+    // submit_info.pNext              = &timeline_info;
```
fabian-lunarg: got forgotten

framework/decode/vulkan_frame_warm_up.cpp
```diff
@@ -0,0 +1,410 @@
...
+    {
+        GFXRECON_LOG_FATAL("VulkanFrameWarmUp: Failed in vkBindBufferMemory: %s.",
+                           util::ToString<VkResult>(result).c_str());
+    }
+
+    VkDescriptorSetLayoutBinding layout_binding = {};
+    layout_binding.binding                      = 0;
+    layout_binding.descriptorType               = VK_DESCRIPTOR_TYPE_STORAGE_BUFFER;
+    layout_binding.descriptorCount              = 1;
+    layout_binding.stageFlags                   = VK_SHADER_STAGE_COMPUTE_BIT;
```
fabian-lunarg, line 185: the loaded spirv is expected to have a specific descriptor-layout (and grid-size aligned to 64?),
but this is not documented for `--frame-warm-up-spirv`

what are other restrictions for the required spirv?
why only load from external file, was an included default spirv problematic?

antonio-lunarg: `--frame-warm-up-spirv` is intentionally a mechanism for running a user-provided compute workload before replaying a frame, not a claim that gfxreconstruct knows the right warm-up shader for every GPU. We originally had a small embedded compute shader, but removed it because its effectiveness is highly GPU-, driver-, and tool-dependent. Keeping a built-in default would make that shader look endorsed as a generally good warm-up workload, when in practice it is only one arbitrary choice.

But you're right about documentation, that is lacking, I am going to address that.

antonio-lunarg: New commit pushed with docs udpate.

PR #2716, opened 2026-02-23
Enable external Vulkan layers for Android replay

mikes-lunarg:
Usually only debug builds are allowed to load external Vulkan layers on Android, which can be useful for validation/re-capture/etc. This change should allow release builds to also load external Vulkan layers for cases where performance is more important.


MarkY-LunarG:
I don't know if this will also work in the quest replay app, but the rest looks simple and easy.  Thnx Mike!

PR #2717, opened 2026-02-23
VDR: Fix implied operator precedence

panos-lunarg:


framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
jzulauf-lunarg: Should be a matching change here I think.

jzulauf-lunarg (commented):
LGTM

jzulauf-lunarg (approved):
LGTM

PR #2719, opened 2026-02-23
Fix operator precedence in GFXR version check

jzulauf-lunarg:
Misplaced ) fixed

### DIFF
--- a/framework/format/format_util.cpp
+++ b/framework/format/format_util.cpp
@@ -198,7 +198,7 @@ GfxrVersion ParseVersionFromOperations(const char* operations)
 bool GfxrVersion::SupportsFrameMarkers() const
 {
     // Supported from 1.0.1 and later builds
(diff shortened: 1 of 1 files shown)


framework/format/format_util.cpp
```diff
@@ -198,7 +198,7 @@ GfxrVersion ParseVersionFromOperations(const char* operations)
 bool GfxrVersion::SupportsFrameMarkers() const
 {
     // Supported from 1.0.1 and later builds
-    return valid && ((major > 1)) || ((major == 1) && ((minor > 0) || (patch > 0)));
+    return valid && ((major > 1) || ((major == 1) && ((minor > 0) || (patch > 0))));
```
MarkY-LunarG, line 201: LGTM, although you should be able to remove the `(major == 1) &&` as well.

PR #2722, opened 2026-02-24
jenkins: use Python venv

lunarpapillo:
As per:

    Use venv and do install for all jobs that use VulkanTests
    https://github.com/LunarG/Projects/issues/1018

this updates the Jenkins integration for GFXReconstruct pipelines to use the common boilerplate now used for all Jenkins multi-configuration jobs, including the creation and use of a Python virtual environment, as per:

    https://github.com/LunarG/VulkanTests/blob/master/doc/administrator/jenkins-jobs.md

Other than the Python virtual environment, the pattern for checking out repositories into a subdirectory has been modified as per current best practices, to use the git "-C <subdirectory>" switch to manipulate the repository in the subdirectory instead of explict "cd" commands; and artifact archival has been expanded to include the "python-venv.txt" file created during Python virtual environment initialization, so it can be examined in case of unexpected failure.

ci/runJob.bat:
-   Windows implementation

ci/runJob.sh:
-   Mac and Linux implementation

ci/runJobAndroid.sh:
-   Android on Linux host implementation

ci/runJob.groovy:
-   update all artifact archival to include python-venv.txt

### DIFF
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -3,18 +3,28 @@ pipeline {
 
     options {
         disableConcurrentBuilds(abortPrevious: true)
-        buildDiscarder(logRotator(
-            daysToKeepStr: '90',
-            numToKeepStr: '100',
-            artifactDaysToKeepStr: '90',
-            artifactNumToKeepStr: '100'
-        ))
     }
 
     stages {
         stage('Per Commit Tests') {
             steps {
                 script {
+                    // different log rotator settings for dev vs PRs
+                    def discarder = env.BRANCH_NAME == 'dev'
+                        ? buildDiscarder(logRotator(
+                            daysToKeepStr: '90',
+                            numToKeepStr: '100',
+                            artifactDaysToKeepStr: '90',
+                            artifactNumToKeepStr: '100'
+                        ))
+                        : buildDiscarder(logRotator(
+                            daysToKeepStr: '7',
+                            numToKeepStr: '5',
+                            artifactDaysToKeepStr: '7',
+                            artifactNumToKeepStr: '5'
+                        ))
+                    properties([discarder])
+
                     def tests
 
                     // Checkout and load on any available node
--- a/ci/Jenkinsfile.extended-manual
+++ b/ci/Jenkinsfile.extended-manual
@@ -43,10 +43,10 @@ pipeline {
     options {
         disableConcurrentBuilds(abortPrevious: true)
         buildDiscarder(logRotator(
-            daysToKeepStr: '90',
-            numToKeepStr: '100',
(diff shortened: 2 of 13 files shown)


ci/runJob.sh
```diff
@@ -30,13 +28,16 @@ fi
 
 git clone --verbose $TEST_REPO VulkanTests
 cd VulkanTests
```
antonio-lunarg: Forgot to remove this line?

lunarpapillo: Yes, thank you for catching it!  Fixed in the next iteration.

ci/runJobAndroid.sh
```diff
@@ -29,13 +27,16 @@ if [ -z "${TEST_COMMIT:-}" ]; then
...
+git -C VulkanTests submodule update --init --recursive
+git -C VulkanTests describe --tags --always
+
 cmake --version
 python3 --version
-python3 VulkanTests/gfxrecontest.py --os AndroidTestOS --build-mode $BUILD_MODE $ANDROID_TEST_ARGS --suite "ci-gfxr-suites/$GFXRECON_TRACE_SUBDIR/$TEST_SUITE" --trace-dir "$GFXRECON_TRACE_DIR" --result-dir "$RESULTS_DIR"
\ No newline at end of file
+
+python3 -m venv $WORKSPACE/python-venv
+$WORKSPACE/python-venv/bin/python3 -m pip install --no-cache-dir -r VulkanTests/requirements.txt > $WORKSPACE/python-venv.txt 2>&1
```
mikes-lunarg, line 40: I'm curious if there is a specific reason to use the full venv python executable path instead of using `python-venv/bin/activate`

mikes-lunarg (approved):
LGTM, just one question

lunarpapillo:
Squashing and rebasing to bring branch up-to-date

ci/runJobAndroid.sh
lunarpapillo: Here, to be as consistent as possible with other CI.

In "mainline" CI, the test script typically starts with the current directory in the repository code, but sometimes the current directory is in the workspace root.

Using the absolute path to `$WORKSPACE` works in all cases.

In a related issue... we _could_ just create the `venv` wherever we are (whether it's the workspace root or the repository root) - but it turns out creating the `venv` in the **VulkanTests** or **ci-servers** repository root breaks their tests, because their tests search for all Python files within the repository root and runs `pylint` on them.  Creating the `venv` inside the repository root taints the repository with hundreds of additional Python files that don't pass `pylint` and causes tests to fail.

So we always create the `venv` explicitly in the workspace root, and we always refer to it directly in the workspace root for consistency both with its creation and with its use across all **Jenkins** jobs.

PR #2723, opened 2026-02-24
Remove busy-wait loops for vkGetEventStatus and vkGetQueryPoolResult

fabian-lunarg:
- no more looping, no blocking-wait, just invoke func once, be happy with whatever it returns
- allow-list both functions, other return-types are expected and should be benign
- instead log debug-messages
- this behavior is more predictable and better matches GFXR guidelines imo
- log-spam is reduced significantly for replay of proton/DXVK

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -68,9 +68,6 @@
 #include <unordered_set>
 #include <future>
 
-// NOTE: still unsupported on android, as of NDK 36 (check _LIBCPP_HAS_NO_INCOMPLETE_FORMAT)
-#include <format>
-
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 


MarkY-LunarG:
I think this breaks certain replays because the game took longer for executing certain behavior at times (like AI/physics) so that the graphics was able to keep pace.  I forget which apps showed this behavior  (maybe Witcher 3?) but we had several.

What log spam was removed?  I didn't notice any messages being removed, only added.

fabian-lunarg:
> I think this breaks certain replays because the game took longer for executing certain behavior at times (like AI/physics) so that the graphics was able to keep pace. I forget which apps showed this behavior (maybe Witcher 3?) but we had several.

this time around I (again) think it shouldn't break replay :) 
the busy-wait really bothers me, as it doesn't solve any problem.

the 'actual' problem is we're not replicating the original application's concurrency pattern,
our replay is single-threaded and we don't capture/replay an application's  external synchronization.
your point about having less cpu-load is equally valid.
that's why I think it's expected those functions can&will return different return-values.
also the queried data will not be used by the replay in any way, we basically just throw that away. 
instead, when the application used the data in other GPU-related places, it will re-appear in captured buffer-data. 

observing in practice what we do there, is basically max out on `kMaxEventStatusRetries` / `kMaxQueryPoolResultsRetries` and log a warning. so we pollute both the log and don't replay what was originally captured. re-capturing our replay then spams `kMaxEventStatusRetries` invocations in a row every time the function was originally called once. that cannot be anyone's intention.

the last failed attempt to move on from busy-waiting was introducing a blocking-wait. 
that appeared to first work, but was prone to deadlock, depending on non-deterministic factors during capture/replay. (https://github.com/LunarG/gfxreconstruct/pull/2533)

this time, instead of replacing the 'busy-wait' with a 'wait', we replace with nothing.

> What log spam was removed? I didn't notice any messages being removed, only added.

the removed logs were:

```
[gfxrecon] WARNING - API call vkGetQueryPoolResults returned value VK_NOT_READY that does not match return value from capture file: VK_SUCCESS.
[gfxrecon] WARNING - API call vkGetEventStatus returned value VK_EVENT_RESET that does not match return value from capture file: VK_EVENT_SET.
```
(per frame)

the removal is implicit by adding both functions to an allow-list we already had (`kFunctionsAllowedToReturnDifferentCodeThanCapture`). this in turn circumvents the logging in  `VulkanReplayConsumerBase::CheckResult`

MarkY-LunarG:
All my local testing works.  If you can test the previously mentioned game replays and they work, I'm ok with moving forward with this.

MarkY-LunarG (approved):
This approval is dependent on verifying previous problematic trace.

per-mathisen-arm:
I don't think either of these changes are safe. For vkGetEventStatus this is used by eg Unreal Engine for resource synchronization, checking if a particular resource is ready for reuse. If you replay on the same device, you may be lucky and the timing ends up "right" but if you replay on a much slower device (eg a simulation) this will break. For vkGetQueryPoolResult, have you considered the possibility that it might be writing to a device-local buffer? I don't know of anyone who does this, but it seems a possibility, at least.

fabian-lunarg:
> I don't think either of these changes are safe. For vkGetEventStatus this is used by eg Unreal Engine for resource synchronization, checking if a particular resource is ready for reuse. If you replay on the same device, you may be lucky and the timing ends up "right" but if you replay on a much slower device (eg a simulation) this will break.

I agree that the replay-behavior around those two functions (and also other types of queries, retrieving data)
remains a problem that's not solved.
But comparing with the current behavior, which is a busy-wait with an arbitrary and separate "max-retries" variable for each function, I argue this will not introduce any regression.
In both cases we'll eventually return with a mismatching return type, because e.g. the query was ready, but was not ready during capture, or the other way around etc.

or have to ask differently:
do you expect any regression in a scenario where the current busy-wait is helpful or solves a particular problem?

> For vkGetQueryPoolResult, have you considered the possibility that it might be writing to a device-local buffer? I don't know of anyone who does this, but it seems a possibility, at least.

[vkGetQueryPoolResult](https://registry.khronos.org/VulkanSC/specs/1.0-extensions/man/html/vkGetQueryPoolResults.html) does expect a host-mem `void*` and that could also be inside a host-visible VkMemory. but my consideration was mostly: 

whatever the captured application did with the queried data, as far as it was relevant for issued gpu-commands,
would be reflected in 'some' change to a buffer. so that data would be found in a future FillMemory metacommand.
also meaning during replay the retrieved data is not used for anything.

davidd-lunarg:
DX12 has an API call `ID3D12Fence::GetCompletedValue()` that shares some similarities to `vkGetEventStatus`. For this DX12 call, when the replay result < the capture result, GFXR needs to perform a blocking wait in order to guarantee GPU execution is caught up. To do this without potentially deadlocking, additional tracking is used in `Dx12ReplayConsumerBase`. Fence and command queue states are tracked. Replay uses this information to delay the blocking wait until the correct fence signal is replayed--for example `ID3D12CommandQueue::Signal`. This prevents deadlock in single threaded replay when the call to signal the fence was captured after the call to check the value. Once the correct signal call is replayed, GFXR can perform a blocking wait.

Is a similar mechanism needed to guarantee `vkGetEventStatus` correctness? If the event is not set, but the associated API call to set the event was already replayed (for example, a command buffer with the relevant `vkCmdSetEvent` was submitted to the GPU), GFXR can immediately block until the event is set. On the other hand if the event set call hasn't been replayed yet, GFXR needs to queue the wait until it is.

`ID3D12Fence::GetCompletedValue()`:
https://github.com/LunarG/gfxreconstruct/blob/baa397ef3caf106ddb563326b417ad9672baaf40/framework/decode/dx12_replay_consumer_base.cpp#L2799-L2831

fabian-lunarg:
I guess one difference between `vkGetEventStatus` and `ID3D12Fence::GetCompletedValue()` is the former doesn't block. it can be used to e.g. query resource-loading etc., but cannot be used for synchronization on its own.
on the other hand `vkGetQueryPoolResult` 'can' block, depending on passing a flag.

having such queue and deferred execution for `vkGetQueryPoolResult` and around blocking `VkFence`-API
would then shuffle the order of execution and allow sequential, single-threaded execution.
optimally we would be able to replay using multiple threads, matching what the captured application did.
but yeah, both ways would be an improvement over now.

per-mathisen-arm:
I agree that the current implementation of vkGetEventStatus() is not terribly useful.

For vkGetQueryPoolResult(), I suppose whether it has impact here or not depends on the order of updates, which I am not clear on. Would the FillMemory command overwrite any data that vkGetQueryPoolResult would produce? This data could be passed onto the GPU where actions taken may depend on what is in this data. And I am not sure which order is actually better here. But *if* we want the query data to take precedence over stored data, *then* making sure the query completes will be more important. If not, then I guess it is irrelevant, as you say. In any case, this is a purely theoretical concern on my part, as I have not seen anyone use it this way yet.

fabian-lunarg:
> Would the FillMemory command overwrite any data that vkGetQueryPoolResult would produce?

in my understanding, when the result-data was used by the original application, then a captured FillMemory command will contain the data that `vkGetQueryPoolResult` returned during capture.
while the data returned by `vkGetQueryPoolResult` during replay is currently not used at all.
(there is at least one exception to this, e.g. for querying AS-compacted size when using portability / `-m rebind`)

> But if we want the query data to take precedence over stored data, then making sure the query completes will be more important. If not, then I guess it is irrelevant, as you say.

agreed. when we actually plan to use the data we need the query to complete. 
but currently it's as pointed out above, it should make no difference for replay.

davidd-lunarg:
Can these errors stay as `GFXRECON_LOG_WARNING_ONCE`? If I understand correctly, they are possibly indicating errors in synchronization for replay and so shouldn't be suppressed completely.

fabian-lunarg:
> Can these errors stay as `GFXRECON_LOG_WARNING_ONCE`? If I understand correctly, they are possibly indicating errors in synchronization for replay and so shouldn't be suppressed completely.

I agree they should not be suppressed completely. but I instead suggest to turn them in  `GFXRECON_LOG_DEBUG`
(and all other allow-listed cases which were not logged before).

in most cases those are benign, but when we do care, using `GFXRECON_LOG_WARNING_ONCE` would 'suppress' 
all but the first warning. that's why I think this should be a debug-log (either I need all of it or nothing)

fabian-lunarg:
> This approval is dependent on verifying previous problematic trace.

I was finally able to confirm this works.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -67,16 +67,15 @@
 #include <numeric>
 #include <unordered_set>
 #include <future>
-#include <span>
+
+// NOTE: still unsupported on android, as of NDK 36 (check _LIBCPP_HAS_NO_INCOMPLETE_FORMAT)
+#include <format>
```
MarkY-LunarG: Should you remove this?

fabian-lunarg: yes, will remove. anyway had a false negative CI run and also need to rebase again :)

PR #2724, opened 2026-02-24
Refactor address replacer

antonio-lunarg:
No functional changes. Just refactoring, deduplicating and moving code from consumer to address replacer.

I Introduced a sort of "job" system to schedule and inject pre-submit work, so now `OverrideQueueSubmit`/`OverrideQueueSubmit2` can build job plans instead of inlining semaphore injection logic.

### DIFF
--- a/framework/decode/struct_pointer_decoder.h
+++ b/framework/decode/struct_pointer_decoder.h
@@ -41,6 +41,7 @@
 #include <memory>
 #include <type_traits>
 #include <utility>
+#include <span>
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
@@ -82,7 +83,11 @@ class StructPointerDecoder : public PointerDecoderBase
 
     typename T::struct_type* GetPointer() { return struct_memory_; }
 
+    std::span<typename T::struct_type> GetSpan() { return std::span(GetPointer(), GetLength()); }
+
     const typename T::struct_type* GetPointer() const { return struct_memory_; }
+
+    const std::span<typename T::struct_type> GetSpan() const { return std::span(GetPointer(), GetLength()); }
 
     size_t GetOutputLength() const { return output_len_; }
 
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -36,6 +36,7 @@
 #include "encode/vulkan_capture_manager.h"
 #include "graphics/vulkan_feature_util.h"
 #include "decode/vulkan_object_cleanup_util.h"
+#include "decode/vulkan_submit_job.h"
 #include "format/format.h"
 #include "format/format_util.h"
 #include "generated/generated_vulkan_struct_decoders.h"
@@ -4125,9 +4126,6 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit(PFN_vkQueueSubmit        
     auto    submit_info_data = pSubmits->GetMetaStructPointer();
     VkFence fence            = VK_NULL_HANDLE;
 
-    // semaphores potentially used by replacer helper-submission(s), defined here to ensure lifetime
-    std::vector<VkSemaphore> semaphores(submitCount);
-
     if (fence_info != nullptr)
     {
         fence = fence_info->handle;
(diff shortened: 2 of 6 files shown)


framework/decode/vulkan_address_replacer.cpp
```diff
@@ -364,106 +365,106 @@ VulkanAddressReplacer::~VulkanAddressReplacer()
...
+            submit_info.waitSemaphoreCount   = wait_semaphores.size();
+            submit_info.pWaitSemaphores      = wait_semaphores.empty() ? nullptr : semaphore_handles.data();
+            submit_info.pWaitDstStageMask    = wait_semaphores.empty() ? nullptr : wait_dst_stages.data();
+            submit_info.commandBufferCount   = 1;
+            submit_info.pCommandBuffers      = &submit_asset.command_buffer;
+            submit_info.signalSemaphoreCount = 1;
+            submit_info.pSignalSemaphores    = &submit_asset.signal_semaphore;
+
+            // submit
+            device_table_->QueueSubmit(queue_, 1, &submit_info, submit_asset.fence);
```
fabian-lunarg, line 451: there is something missing now. check the dtor `~QueueSubmitHelper`,
it also did a fence-sync here

framework/decode/vulkan_address_replacer.h
```diff
@@ -344,6 +369,25 @@ class VulkanAddressReplacer
...
+     */
+    void SetAddressesToReplace(std::vector<VkDeviceAddress> addresses) { addresses_to_replace_ = std::move(addresses); }
+
+    /**
+     * @brief  SetDeviceAddressTracker allows providing a pointer to a VulkanDeviceAddressTracker, which will be used
+     * for mapping capture-time to replay-time addresses and vice versa.
+     *
+     * @param  address_tracker    a pointer to a VulkanDeviceAddressTracker.
+     */
+    void SetDeviceAddressTracker(const VulkanDeviceAddressTracker* address_tracker)
```
fabian-lunarg: this could be a good idea. yet, most API here still has parameter
`(const) VulkanDeviceAddressTracker&     address_tracker` - so it's a duplication.

I think either we should remove all parameters and switch to use the `address_tracker_` instead.
... or maybe leave that change for a future PR, so it's more oversee-able for now.

framework/decode/vulkan_address_replacer.h
```diff
@@ -509,6 +552,9 @@ class VulkanAddressReplacer
     PFN_vkGetBufferDeviceAddress       get_device_address_fn_             = nullptr;
     PFN_vkGetPhysicalDeviceProperties2 get_physical_device_properties_fn_ = nullptr;
     PFN_vkSetDebugUtilsObjectNameEXT   set_debug_utils_object_name_fn_    = nullptr;
+
+    std::vector<VkDeviceAddress>      addresses_to_replace_;
```
fabian-lunarg: not sure about making the class "statefull" by introducing this 
`std::vector<VkDeviceAddress>      addresses_to_replace_` member.

what is the main benefit?

framework/decode/vulkan_address_replacer.h
antonio-lunarg: In general, if it's `const` I believe it's fine to set it once and avoid passing it around to other functions.
If it is mutable, I would prefer keeping it a function parameter, just to avoid shooting ourselves in the foot if multithreading becomes a priority in the future.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4140,72 +4141,31 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit(PFN_vkQueueSubmit
...
-                addresses_to_replace.insert(addresses_to_replace.end(),
-                                            command_buffer_info->addresses_to_replace.begin(),
-                                            command_buffer_info->addresses_to_replace.end());
-                sync_wait_semaphores |= !command_buffer_info->addresses_to_replace.empty();
-                if (cmd_buf_info == nullptr)
-                {
-                    cmd_buf_info = command_buffer_info;
-                }
-            }
+            auto [addresses_to_replace, cmd_buf_info] =
```
fabian-lunarg, line 4151: very nice, this cleans up alot of code here

framework/decode/vulkan_address_replacer.h
antonio-lunarg: I think I experimented with it but it's not used, I am going to remove it.

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: I think I got confused in the diff, it's fine

framework/decode/struct_pointer_decoder.h
```diff
@@ -82,8 +83,12 @@ class StructPointerDecoder : public PointerDecoderBase
 
     typename T::struct_type* GetPointer() { return struct_memory_; }
 
+    std::span<typename T::struct_type> GetSpan() { return std::span(GetPointer(), GetLength()); }
```
fabian-lunarg, line 86: nice, this will be useful

framework/graphics/vulkan_util.h
```diff
@@ -75,16 +75,6 @@ static inline void copy_dispatch_table_from_device(VkDevice device, VkCommandBuf
...
 
-/**
- * @brief   StripWaitSemaphores can be used to remove all wait-semaphores for a provided VkSubmitInfo.
- *          Respective pointer in submit_info will be set to nullptr and count to zero.
- *
- * @param   submit_info     a provided VkSubmitInfo(2) struct
- * @return  an array of Semaphores that have been stripped/removed from submit_info
- */
-std::vector<std::pair<VkSemaphore, uint64_t>> StripWaitSemaphores(VkSubmitInfo* submit_info);
-std::vector<std::pair<VkSemaphore, uint64_t>> StripWaitSemaphores(VkSubmitInfo2* submit_info);
```
fabian-lunarg, line 86: there is still the `StripWaitSemaphores` implementation left in `vulkan_util.cpp`,
needs to be removed (it now lives in `vulkan_semaphore_util.cpp`)

framework/decode/vulkan_address_replacer.h
```diff
@@ -106,6 +107,30 @@ class VulkanAddressReplacer
      */
     void ResolveBufferAddresses(VulkanCommandBufferInfo*                  command_buffer_info,
                                 const decode::VulkanDeviceAddressTracker& address_tracker);
+
+    /**
+     * @brief   `ResolveBufferAddresses` can be used to identify buffers which are referenced
+     *          by buffer-device-addresses.
+     *
+     * @param   submit_info         a provided submit_info containing command-buffer-infos with locations to resolve
```
fabian-lunarg: doc needs adjustment for new params

PR #2725, opened 2026-02-24
Add memory capturing for vkMapMemory2

ziga-lunarg:
Support for VK_KHR_map_memory2 was missing. This forwards calls from new 2 functions to the old ones


fabian-lunarg (approved):

fabian-lunarg:
I double approved this :)

Issue #2709, opened 2026-02-19
[capture] Issue caused by VDR: Check for unsupported commit


OliverCho1:
**Describe the capture bug:**
I got 2 Android game issue cases.

Tiledom (Funvent Studios DMCC) - https://play.google.com/store/apps/details?id=games.spearmint.triplecrush
2 Tiles (Funvent Studios DMCC) - https://play.google.com/store/apps/details?id=games.spearmint.matchinggame
It occurs when I try to capture with trimmed capture mode. (manual capture during game playing)

However, when I capture a file from the 1st frame, the issue does not occur.

When the issue occurs, the logcat log shows the following message.
"gfxrecon: Format VK_FORMAT_D24_UNORM_S8_UINT is not supported by the implementation"

I think this commit should consider this kind of issue.

**Build Environment:**
Disable android d32 #2707

**Capture Settings:**
adb shell "setprop debug.gfxrecon.capture_android_trigger 'false'" (set true/false during the main game playing)
adb shell "setprop debug.gfxrecon.capture_compression_type 'LZ4'"
adb shell "setprop debug.gfxrecon.memory_tracking_mode 'page_guard'"
adb shell "setprop debug.gfxrecon.page_guard_copy_on_map 'true'"
adb shell "setprop debug.gfxrecon.page_guard_separate_read 'true'"
adb shell "setprop debug.gfxrecon.page_guard_persistent_memory 'false'"
adb shell "setprop debug.gfxrecon.page_guard_align_buffer_sizes 'true'"
adb shell "setprop debug.gfxrecon.page_guard_unblock_sigsegv 'false'"
adb shell "setprop debug.gfxrecon.page_guard_signal_handler_watcher 'false'"
adb shell "setprop debug.gfxrecon.page_guard_signal_handler_watcher_max_restores '1'"

**To Reproduce:**
1. Apply capture settings
2. Start game
3. Set capture_android_trigger as "true" to start capture.
4. Soon, the game finished forcibly.
5. From the logcat log, it shows the following log.
"gfxrecon: Format VK_FORMAT_D24_UNORM_S8_UINT is not supported by the implementation"

OliverCho1:
Fixed in #2727 . Closing this issue.

PR #2727, opened 2026-02-25
Turn a problematic asserting into a warning, continue loop-iteration

fabian-lunarg:
the motivation for this PR is a case where I'm able to replay a capture using:
`gfxrecon-replay -m rebind --remove-unsupported`

trying to trim this capture, using the same command for replay,  crashed somewhere deep into the capture.
debug-builds were hitting the mentioned assert.
before that, I was seeing the following logs:
`[gfxrecon] ERROR - Format VK_FORMAT_D24_UNORM_S8_UINT is not supported by the implementation`

the suggested change simply skips over the assertion, logs a warning and continues processing other loop-items.
this fixes trimming on different device for me. 
otherwise I would depend on trimming on the same device (which I don't have here)

### DIFF
--- a/framework/encode/vulkan_state_writer.cpp
+++ b/framework/encode/vulkan_state_writer.cpp
@@ -2553,7 +2553,11 @@ void VulkanStateWriter::ProcessImageMemory(const vulkan_wrappers::DeviceWrapper*
             (((image_wrapper->is_swapchain_image || image_wrapper->is_sparse_image) && memory_wrapper == nullptr) ||
              ((!image_wrapper->is_swapchain_image && !image_wrapper->is_sparse_image) && memory_wrapper != nullptr)));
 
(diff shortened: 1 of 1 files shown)


panos-lunarg:
I think https://github.com/LunarG/gfxreconstruct/issues/2709 is relevant

zmike:
I think I would expect replay to turn this into a compatible format? If VK_FORMAT_D24_UNORM_S8_UINT isn't supported then use D32S8.

fabian-lunarg:
> I think I would expect replay to turn this into a compatible format? If VK_FORMAT_D24_UNORM_S8_UINT isn't supported then use D32S8.

yes, sounds like a reasonable thing to do. looking into this, incl. trying to avoid creating new VVL-issues :)

PR #2728, opened 2026-02-25
Remove std::format

panos-lunarg:
Workaround for missing c++ libs on Ubuntu 22

### DIFF
--- a/tools/optimize/file_optimizer.cpp
+++ b/tools/optimize/file_optimizer.cpp
@@ -27,7 +27,6 @@
 #include "util/logging.h"
 
 #include <string>
-#include <format>
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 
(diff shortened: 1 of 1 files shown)


fabian-lunarg:
I'm not entirely sure here, but think we raised the minimum gcc-version to 13 recently ( maybe @MarkY-LunarG remembers the exact reason).

ubuntu 22/04 ships with gcc 11 as default - gcc12 is the max, but gcc13 'can' be installed by adding a source-repo.

our CI on github is using `ubuntu-latest`, which is 24/04.
ubuntu 24/04 ships with gcc 13 as default.
our internal CI is also setup to use gcc-13 as default. 

so, this might be fine and there are possibly still issues around `std::format` for other reasons,
like missing support even in recent NDK versions.

I just think the explanation that ubuntu-22/04 does not contain a modern enough gcc,
is not enough on its own to not use modern features. alternatively either the OS or just the compiler version can be updated.

panos-lunarg:
The problem seems to be that while you can install gcc-13 on ubuntu 22, the libc .so files are not bumped to a new enough version. So in practice you can compile but at runtime you get dynamic linking errors.

MarkY-LunarG:
I think this is only test center motivated since  Ubuntu 26.04 will be coming out shortly, but that would require us to update our test systems.  So for  now, I think we either do this, or conditionally compile the std::format code with a GCC check.

PR #2731, opened 2026-02-25
Convert memory allocation during block processing and enqueue to use a Hybrid Linear Allocator

jzulauf-lunarg:
Encapsulates all storage of "batches" of ParsedBlocks to allocate small blocks from a linear arena and large block from the heap, but owned by the allocator.  ParsedBlocks are simplified (and shrunk) to no longer owned referenced memory.  ParsedBlocks allocated with a batch's allocator are owned by the allocator and only reference memory owned by the allocator.

Performance neutral.


framework/util/hybrid_linear_allocator.h
```diff
@@ -0,0 +1,218 @@
...
+
+// #define ZZZ_INSTRUMENT
+#ifdef ZZZ_INSTRUMENT
+#include <iostream>
+#endif
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(util)
+
+class HybridLinearAllocator
```
davidd-lunarg, line 43: How different is this behavior from `MonotonicAllocator`? Could `MonotonicAllocator` be reused?

jzulauf-lunarg: It's similar, though the semantics for the HLA usage are use until full with a 1-1 relationship to a BlockBatch. We can possibly traitsify the semantics to coalesce them, but the reset logic for the HLA is performance sensitive to replay, so it would need to be zero cost.

framework/decode/block_parser.cpp
```diff
@@ -1603,15 +1617,15 @@ ParsedBlock BlockParser::ParseMetaData(BlockBuffer& block_buffer)
...
-            return MakeIncompressibleParsedBlock(
-                block_buffer,
-                SetOpaqueDescriptorDataArgs{ meta_data_id,
-                                             header.thread_id,
-                                             header.device_id,
-                                             header.object_id,
-                                             header.data_size,
-                                             reinterpret_cast<const uint8_t*>(read_result.buffer.data()) },
-                true /* references block buffer */);
+            auto* payload =
```
antonio-lunarg: What if `read_result.success` is `false`?

framework/decode/block_buffer.h
```diff
@@ -35,87 +35,86 @@ class FileProcessor;
...
+    }
+    size_t ReadPos() const
+    {
+        AssertValidPointers();
+        return cursor_ - GetData();
+    }
+    bool IsAvailable(size_t size) const noexcept
+    {
+        AssertValidPointers();
+        return (cursor_ + size) <= end_;
```
antonio-lunarg: How about checking for overflow?

framework/util/hybrid_linear_allocator.h
```diff
@@ -0,0 +1,221 @@
...
+
+    template <size_t kAlignment>
+    void* Allocate(size_t size)
+    {
+        assert(std::has_single_bit(kAlignment));
+        const uintptr_t cursor        = reinterpret_cast<uintptr_t>(alloc_cursor_);
+        uintptr_t       start         = aligned_value<kAlignment>(cursor);
+        const size_t    align_pad     = static_cast<size_t>(start - cursor);
+        const size_t    total_size    = size + align_pad;
+        DataType* const update_cursor = alloc_cursor_ + total_size;
```
antonio-lunarg, line 70: Might be a good idea to check for overflow here as well.

framework/util/hybrid_linear_allocator.h
jzulauf-lunarg: @davidd-lunarg -- spent some time trying to converge them... some performance loss to handle the different policies, so, leaving as is.

davidd-lunarg: Thanks for looking into it.

framework/decode/block_buffer.h
jzulauf-lunarg: or better, *avoiding* it `assert(cursor_ <= end_) ; return size <= (end_ - cursor_);`

framework/decode/block_parser.cpp
jzulauf-lunarg: this is just an escape... I don't know how long the "InvalidBlock" path has been missing.

framework/util/hybrid_linear_allocator.h
jzulauf-lunarg: Right.  Both for total_size and for update_cursor additions

framework/decode/preload_file_processor.cpp
```diff
@@ -206,10 +229,50 @@ FileProcessor::ProcessBlockState PreloadFileProcessor::ReplayOneFrame(PreloadedF
 
         dispatch_visitor.SetBlockIndex(block_index);
         std::visit(dispatch_visitor, queued_block.GetArgs());
```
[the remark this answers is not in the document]
jzulauf-lunarg: Added a std::monostate handler, s.t. that visitors can skip non-dispatchable blocks

framework/decode/block_parser.cpp
```diff
@@ -290,11 +253,91 @@ ParsedBlock BlockParser::ParseBlock(BlockBuffer& block_buffer)
...
+    {
+        // Conversely, retained mode always uses the block allocator
+        info.use_working_buffer = false;
+        info.mode               = BlockAllocator::AllocatorMode::kEnqueue;
+
+        // All blocks retain raw block data in BlockBatch
+        info.linear_allocation += size_considering_jumbo(raw_block_size);
+        if (format::IsBlockCompressed(type) && !ShouldDeferDecompression(raw_block_size))
+        {
+            info.linear_allocation += decompressed_size(raw_block_size);
```
antonio-lunarg, line 303: In case of a compressed block, I understand that linear allocation is increased by both compressed size and decompressed size. Is this expected? Why?

framework/decode/block_parser.cpp
```diff
@@ -290,11 +253,91 @@ ParsedBlock BlockParser::ParseBlock(BlockBuffer& block_buffer)
...
+                // as we need to retain the raw block data, to decompress later
+                info.use_working_buffer = false;
+                info.linear_allocation += size_considering_jumbo(raw_block_size);
+            }
+            else
+            {
+                // Use working buffer for raw block if immediate decompression in enqueued mode
+                // but include estimated decompressed size
+                info.use_working_buffer = true;
+                info.linear_allocation += decompressed_size(raw_block_size);
```
antonio-lunarg, line 327: In this case linear allocation is increased by only the decompressed size, ignoring the compressed size. Why?

framework/decode/parsed_block.cpp
```diff
@@ -55,19 +79,33 @@ bool ParsedBlock::Decompress(BlockParser& parser)
...
             {
-                // Patch the data buffer pointer, and shift ownership of the backing store to the parsed block
-                args.data = uncompressed_store.template GetAs<const uint8_t>();
-                UpdateUncompressedStore(std::move(uncompressed_store));
+                auto compressed_span   = GetCompressedSpan(args);
+                auto uncompressed_size = GetDispatchArgsDataSize(args);
+                working_store.ReserveDiscarding(uncompressed_size);
+                auto uncompressed_store =
+                    parser.DecompressSpan(compressed_span, uncompressed_size, working_store.get());
+                if (uncompressed_store != nullptr)
```
antonio-lunarg, line 103: Is it OK to ignore the case where `uncompressed_store` is `nullptr`?

framework/decode/preload_file_processor.cpp
```diff
@@ -197,7 +220,7 @@ FileProcessor::ProcessBlockState PreloadFileProcessor::ReplayOneFrame(PreloadedF
 
         if (queued_block.NeedsDecompression())
         {
-            if (!queued_block.Decompress(block_parser))
+            if (!queued_block.Decompress(block_parser, working_uncompressed_store_))
```
antonio-lunarg, line 239: As far as I understand, `Decompress()` now patches `args.data` to point into caller-owned scratch storage (`working_uncompressed_store_`). The current call sites seem to consume that data immediately, so this looks fine, but once that working buffer is reused, earlier `args.data` pointers would no longer be valid. It may be worth documenting this lifetime contract explicitly on `Decompress()` to make future misuse less likely.

framework/decode/block_parser.cpp
jzulauf-lunarg: It's the operation mode.  kEnqueueRetained remembers *everything* as it assumes that some consumer is interested in the raw data.

framework/decode/block_parser.cpp
jzulauf-lunarg: In operation mode kEnqueued we're trying to store the minimum.  So raw block data is loaded into a "working store" and only the decompressed data is retained  (compare vs. kEnqueueRetained)

framework/decode/parsed_block.cpp
jzulauf-lunarg: DecompressSpan has emitted error messages, the caller will detect null an convert the block to an "InvalidBlock", and ProcessBlocks will detect the InvalidBlock and abort, and the caller of *that* will exit replay on error.

framework/decode/preload_file_processor.cpp
jzulauf-lunarg: Currently Replay is single use -- but I can certainly comment that this is a destructive step and  that the block will no longer be valid.

The fact we retain the block at all is something of an artifact of the deferral of memory teardown until after the measurement loop in preload replay

PR #2733, opened 2026-02-27
VDR: Fix use after free crash

panos-lunarg:
Reserving space in the build_infos vector should fix a crash of referencing a dangling pointer when the infoCount was larger than 1.


bradgrantham-lunarg:
Does this argue that `build_infos` should be an unordered_map on int that is checked with `at` before use?

panos-lunarg:
> Does this argue that `build_infos` should be an unordered_map on int that is checked with `at` before use?

`unordered_map` feels like an overkill as `build_infos` is just an array of things that is constructed once. There's no need to search in it, or add or remove objects after it is populated.
How about moving the call to `reserve` in the constructor? This should make it more robust

PR #2736, opened 2026-02-27
Fix --wsi auto behaviour

roman-berger-arm:
Before when replaying captures from a different device --wsi <WSI> had to be specified explicitly and the --wsi auto options was ignored. Even when captures recorded on the same device were replayed the --wsi auto option was ignored and the surface from the capture device was chosen.

Now the --wsi auto option (already default behaviour) picks the surface extension from the capture device if it is available. Otherwise the first available surface extension of the replay device is used. If there was not surface extensions available on the capture device, all --wsi options are ignored and a warning is emitted. Furthermore, a check if a compositor is available is made after initializing the WSIContext.

Documentation and cli --help commands were updated to reflect these changes.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -2927,9 +2927,9 @@ void VulkanReplayConsumerBase::ModifyCreateInstanceInfo(
                             application_->InitializeWsiContext(itr_surface_extension->c_str()))
                         {
                             modified_extensions.push_back(itr_surface_extension->c_str());
-                            GFXRECON_LOG_INFO("--wsi auto: could not find surface: %s, instead using: %s",
-                                              capture_surface_extension.c_str(),
-                                              itr_surface_extension->c_str());
+                            GFXRECON_LOG_WARNING("--wsi auto: could not find surface: %s, instead using: %s",
+                                                 capture_surface_extension.c_str(),
+                                                 itr_surface_extension->c_str());
                             picked_surface = true;
                             break;
                         }


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2875,36 +2876,93 @@ void VulkanReplayConsumerBase::ModifyCreateInstanceInfo(
...
+            if (graphics::feature_util::IsSupportedExtension(available_extensions, capture_surface_extension))
+            {
+                auto itr_surface_extension = kSurfaceExtensions.find(capture_surface_extension);
+
+                // check if corresponding compositor exists on replay device
+                if (wsi_contexts.find(*itr_surface_extension) != wsi_contexts.end() ||
+                    application_->InitializeWsiContext(itr_surface_extension->c_str()))
+                {
+                    modified_extensions.push_back(itr_surface_extension->c_str());
+                    GFXRECON_LOG_INFO("--wsi auto used same surface as capture: %s", itr_surface_extension->c_str());
```
fabian-lunarg: I think this log can be skipped. we'd be seeing that on start of every regular replay. 
instead could we when `--wsi auto` is used (implicitly) and the selected WSI is different,
log something like: "`capture_surface_extension` was not available during replay, using ... instead" ?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2875,36 +2876,93 @@ void VulkanReplayConsumerBase::ModifyCreateInstanceInfo(
...
+                {
+                    auto itr_surface_extension = kSurfaceExtensions.find(current_extension.extensionName);
+                    if (itr_surface_extension != kSurfaceExtensions.end())
+                    {
+                        // check if compositor exists before pushing back surface extension
+                        if (wsi_contexts.find(*itr_surface_extension) != wsi_contexts.end() ||
+                            application_->InitializeWsiContext(itr_surface_extension->c_str()))
+                        {
+                            modified_extensions.push_back(itr_surface_extension->c_str());
+                            GFXRECON_LOG_INFO("--wsi auto picked surface: %s", itr_surface_extension->c_str());
```
fabian-lunarg: maybe include the originally captured WSI (`capture_surface_extension`) here?

framework/decode/vulkan_replay_consumer_base.cpp

framework/decode/vulkan_replay_consumer_base.cpp

fabian-lunarg:
investigating an issue our internal CI had with a replay of `vulkaninfo` ...

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2860,51 +2860,107 @@ void VulkanReplayConsumerBase::ModifyCreateInstanceInfo(
...
+                {
+                    auto itr_surface_extension = kSurfaceExtensions.find(current_extension.extensionName);
+                    if (itr_surface_extension != kSurfaceExtensions.end())
+                    {
+                        // check if compositor exists before pushing back surface extension
+                        if (wsi_contexts.contains(*itr_surface_extension) ||
+                            application_->InitializeWsiContext(itr_surface_extension->c_str()))
+                        {
+                            modified_extensions.push_back(itr_surface_extension->c_str());
+                            GFXRECON_LOG_INFO("--wsi auto: could not find surface: %s, instead using: %s",
```
MarkY-LunarG: I would make this a warning.  Just because someone might expect the exact behavior by  default.

PR #2737, opened 2026-02-27
fix: PNext handle mapping

antonio-lunarg:
Mapping handles of pNext chains is done recursively, hence remove the while loop iterating of the pNext chain in `MapNextStructHandles()`.


MarkY-LunarG (approved):
Makes sense.

PR #2738, opened 2026-02-27
android: Update build

antonio-lunarg:
Apply most of the suggestions provided by The Android Gradle plugin (AGP) Upgrade Assistant from Android Studio Panda 1.


android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.4-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.3.1-bin.zip
```
MarkY-LunarG, line 3: You're defining Gradle 9..3.1 here, but you define 9.0.1 above (in android/build.gradle).  I know this is the version the gradlew builds with so perhaps the above is just a minimum, is that the case?

MarkY-LunarG (commented):
Do we know if our customers can use these changes?  I don't know how behind their tool systems are.

android/gradle/wrapper/gradle-wrapper.properties
antonio-lunarg: I believe they're different things. In `gradle-wrapper.properties`, that's the Gradle runtime version. In `build.gradle`, that's the Android Gradle Plugin version. Just following the Android Studio automatic update procedure got me these numbers.

antonio-lunarg:
> @MarkY-LunarG  Do we know if our customers can use these changes? I don't know how behind their tool systems are.

From what I can see looking at CI: our jobs are still configured with Java 17 and pass, which indicates the build remains compatible there. Then Gradle is mostly self-contained via the wrapper, so for typical environments I suppose this update should be low risk.

MarkY-LunarG (approved):

PR #2739, opened 2026-02-27
Fix more issues discovered with pipeline binaries

MarkY-LunarG:
When using pipeline binaries, some allocations didn't occur properly or we ended up in one scenario with double allocations. This should resolve it.


ziga-lunarg (approved):
LGTM

fabian-lunarg (approved):

PR #2740, opened 2026-03-02
Fix memory leak during D3D12 capture

zongdu-arm:
Fix memory leak during D3D12 capture


zongdu-arm:
@fabian-lunarg , @davidd-lunarg, @locke-lunarg 
Could you review this PR?

PR #2751, opened 2026-03-04
Info: Split out API-specific info

MarkY-LunarG:
Update Info so that all API-specific info is separated out.  This greatly cleans up the organization of Info.

### DIFF
--- a/tools/info/info_d3d12_feature.cpp
+++ b/tools/info/info_d3d12_feature.cpp
@@ -42,11 +42,14 @@ std::string InfoD3d12Feature::CompiledHeaderVersionString() const
 #endif
 }
 
-void InfoD3d12Feature::RegisterInternalDecodeComponents(decode::FileProcessor* file_processor)
+void InfoD3d12Feature::RegisterDecodeComponents(decode::FileProcessor&      file_processor,
+                                                const decode::InfoConsumer& info_consumer)
 {
     dx12_decoder_.AddConsumer(&dx12_detection_consumer_);
     dx12_decoder_.AddConsumer(&dx12_consumer_);
-    file_processor->AddDecoder(&dx12_decoder_);
+    file_processor.AddDecoder(&dx12_decoder_);
+    file_processor_ = &file_processor;
+    info_consumer_  = &info_consumer;
 }
 
 std::string InfoD3d12Feature::GetDriverInfoString()
--- a/tools/info/info_d3d12_feature.cpp
+++ b/tools/info/info_d3d12_feature.cpp
@@ -1,0 +1,334 @@
+/*
+** Copyright (c) 2020-2026 LunarG, Inc.
+** Copyright (c) 2022-2024 Advanced Micro Devices, Inc. All rights reserved.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and associated documentation files (the "Software"),
+** to deal in the Software without restriction, including without limitation
+** the rights to use, copy, modify, merge, publish, distribute, sublicense,
+** and/or sell copies of the Software, and to permit persons to whom the
+** Software is furnished to do so, subject to the following conditions:
+**
+** The above copyright notice and this permission notice shall be included in
+** all copies or substantial portions of the Software.
+**
+** THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+** IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+** FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+** AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+** FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+** DEALINGS IN THE SOFTWARE.
+*/
+
+#if defined(D3D12_SUPPORT)
+
+#include "info_d3d12_feature.h"
+
+#include "util/feature_module_registry.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(info)
+
+// Register this class as a feature in a module registry
+GFXR_UTIL_REGISTER_FEATURE_CREATOR(InfoFeature, InfoD3d12Feature)
+
+std::string InfoD3d12Feature::CompiledHeaderVersionString() const
+{
+#if defined(D3D12SDKVersion)
+    return std::string("  D3D12 SDK Version      ") + D3D12SDKVersion;
+#else
+    return "";
+#endif
+}
+
+void InfoD3d12Feature::RegisterInternalDecodeComponents(decode::FileProcessor* file_processor)
+{
+    dx12_decoder_.AddConsumer(&dx12_detection_consumer_);
+    dx12_decoder_.AddConsumer(&dx12_consumer_);
+    file_processor->AddDecoder(&dx12_decoder_);
+}
+
+std::string InfoD3d12Feature::GetDriverInfoString()
+{
+    std::string driver_info = info_consumer_->GetDriverDesc();
+    if (!driver_info.empty())
+    {
+        return driver_info;
+    }
+    else
+    {
+        return "Not available";
+    }
+}
+
+std::string InfoD3d12Feature::GetDriverInfoText()
+{
+    std::string driver_info = info_consumer_->GetDriverDesc();
+    std::string return_val  = "\nDriver info:\n";
+    if (driver_info.length())
+    {
+        return_val += "\t" + driver_info + "\n";
+    }
+    else
+    {
+        return_val += "\tDriver info not available.\n\n";
+    }
+    return return_val;
+}
+
+std::string InfoD3d12Feature::AdapterTypeToString(format::AdapterType type)
+{
+    switch (type)
+    {
+        case format::AdapterType::kUnknownAdapter:
+            return "Unknown type (DXGI 1.0)";
+        case format::AdapterType::kSoftwareAdapter:
+            return "Software";
+        case format::AdapterType::kHardwareAdapter:
+            return "Hardware";
+        default:
+            return "Unknown";
+    }
+}
+
+std::string InfoD3d12Feature::GetRuntimeInfoText()
+{
+    format::Dx12RuntimeInfo runtime_info = dx12_consumer_.GetDx12RuntimeInfo();
+
+    std::string runtime_src = "N/A";
+    std::string runtime_ver = "N/A";
+
+    if (util::platform::StringLength(runtime_info.src) > 0)
+    {
+        runtime_src = runtime_info.src;
+        runtime_ver = std::to_string(runtime_info.version[0]) + "." + std::to_string(runtime_info.version[1]) + "." +
+                      std::to_string(runtime_info.version[2]) + "." + std::to_string(runtime_info.version[3]);
+    }
+
+    std::string return_val = "D3D12 runtime info:\n";
+    return_val += "\tVersion: " + runtime_ver + "\n";
+    return_val += "\tSource: " + runtime_src + "\n\n";
+    return return_val;
+}
+
+nlohmann::json InfoD3d12Feature::GetRuntimeInfoJson()
+{
+    format::Dx12RuntimeInfo runtime_info = dx12_consumer_.GetDx12RuntimeInfo();
+
+    std::string runtime_src = "N/A";
+    std::string runtime_ver = "N/A";
+
+    if (util::platform::StringLength(runtime_info.src) > 0)
+    {
+        runtime_src = runtime_info.src;
+        runtime_ver = std::to_string(runtime_info.version[0]) + "." + std::to_string(runtime_info.version[1]) + "." +
+                      std::to_string(runtime_info.version[2]) + "." + std::to_string(runtime_info.version[3]);
+    }
+
+    return {
+        { "version", runtime_ver },
+        { "source", runtime_src },
+    };
+}
+
+std::string InfoD3d12Feature::GetAdapterInfoText()
+{
+    std::string return_val = "D3D12 adapter info:\n";
+
+    const auto& adapters = dx12_consumer_.GetAdapters();
+
+    if (adapters.empty() == false)
+    {
+        std::unordered_map<int64_t, std::string> adapter_workload;
+        dx12_consumer_.CalcAdapterWorkload(adapter_workload, adapters);
+
+        for (const auto& adapter : adapters)
+        {
+            const int64_t luid = pack_luid(adapter);
+
+            std::string adapter_workload_pct = "";
+
+            if (adapter_workload.count(luid) > 0)
+            {
+                if (adapter_workload[luid] != "")
+                {
+                    adapter_workload_pct = "(" + adapter_workload[luid] + "% of GPU submissions)";
+                }
+            }
+            else if (adapter_workload.size() > 0)
+            {
+                adapter_workload_pct = "(0% of GPU submissions)";
+            }
+
+            std::string adapter_type = AdapterTypeToString(graphics::dx12::ExtractAdapterType(adapter.extra_info));
+
+            return_val +=
+                "\tDescription: " + util::WCharArrayToString(adapter.Description) + " " + adapter_workload_pct + "\n";
+            return_val += "\tVendor ID: " + util::to_hex_fixed_width<uint32_t>(adapter.VendorId) + "\n";
+            return_val += "\tDevice ID: " + util::to_hex_fixed_width<uint32_t>(adapter.DeviceId) + "\n";
+            return_val += "\tSubsys ID: " + util::to_hex_fixed_width<uint32_t>(adapter.SubSysId) + "\n";
+            return_val += "\tRevision: " + std::to_string(adapter.Revision) + "\n";
+            return_val += "\tDedicated Video Memory: " + std::to_string(adapter.DedicatedVideoMemory) + "\n";
+            return_val += "\tDedicated System Memory: " + std::to_string(adapter.DedicatedSystemMemory) + "\n";
+            return_val += "\tShared System Memory: " + std::to_string(adapter.SharedSystemMemory) + "\n";
+            return_val += "\tLUID LowPart: " + util::to_hex_fixed_width<uint32_t>(adapter.LuidLowPart) + "\n";
+            return_val += "\tLUID HighPart: " + util::to_hex_fixed_width<int32_t>(adapter.LuidHighPart) + "\n";
+            return_val += "\tAdapter type: " + adapter_type + "\n\n";
+        }
+    }
+    else
+    {
+        return_val += "\tAdapter info not available.\n\n";
+    }
+    return return_val;
+}
+
+nlohmann::json InfoD3d12Feature::GetAdapterInfoJson()
+{
+    const auto&    adapters      = dx12_consumer_.GetAdapters();
+    nlohmann::json adapters_json = nlohmann::json::array();
+
+    if (!adapters.empty())
+    {
+        std::unordered_map<int64_t, std::string> adapter_workload;
+        dx12_consumer_.CalcAdapterWorkload(adapter_workload, adapters);
+
+        for (const auto& adapter : adapters)
+        {
+            const int64_t luid = (static_cast<uint64_t>(adapter.LuidHighPart) << 32) | adapter.LuidLowPart;
+
+            std::string adapter_workload_pct = "";
+
+            if (adapter_workload.count(luid) > 0)
+            {
+                if (adapter_workload[luid] != "")
+                {
+                    adapter_workload_pct = "(" + adapter_workload[luid] + "% of GPU submissions)";
+                }
+            }
+            else if (adapter_workload.size() > 0)
+            {
+                adapter_workload_pct = "(0% of GPU submissions)";
+            }
+
+            std::string adapter_type = AdapterTypeToString(graphics::dx12::ExtractAdapterType(adapter.extra_info));
+
+            nlohmann::json json_adapter;
+            json_adapter["description"]["details"]          = util::WCharArrayToString(adapter.Description);
+            json_adapter["description"]["workload-percent"] = adapter_workload_pct;
+            json_adapter["vendor-id"]                       = adapter.VendorId;
+            json_adapter["device-id"]                       = adapter.DeviceId;
+            json_adapter["subsys-id"]                       = adapter.SubSysId;
+            json_adapter["revision"]                        = adapter.Revision;
+            auto& memory                                    = json_adapter["memory"];
+            memory["dedicated"]["video"]                    = adapter.DedicatedVideoMemory;
+            memory["dedicated"]["system"]                   = adapter.DedicatedSystemMemory;
+            memory["shared"]                                = adapter.SharedSystemMemory;
+            memory["luid"]["low"]                           = adapter.LuidLowPart;
+            memory["luid"]["high"]                          = adapter.LuidHighPart;
+            json_adapter["adapter-type"]                    = adapter_type;
+            adapters_json.push_back(json_adapter);
+        }
+    }
+
+    return adapters_json;
+}
+
+std::string InfoD3d12Feature::GetSwapchainInfoText()
+{
+    std::string return_val = "D3D12 swapchain info:\n";
+
+    if (dx12_consumer_.FoundSwapchainInfo())
+    {
+        return_val += "\tDimensions: " + dx12_consumer_.GetSwapchainDimensionsString() + "\n\n";
+    }
+    else
+    {
+        return_val += "\tDimensions not available.\n\n";
+    }
+
+    return return_val;
+}
+
+nlohmann::json InfoD3d12Feature::GetSwapchainInfoJson()
+{
+    auto [width, height] = dx12_consumer_.GetSwapchainDimensions();
+    return { "dimensions", { { "width", width }, { "height", height } } };
+}
+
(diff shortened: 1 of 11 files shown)


tools/info/info_api_interface.h
```diff
@@ -0,0 +1,138 @@
...
+        }
+        ss << std::setw(min_width) << value;
+        return ss.str();
+    }
+
+    bool                        api_output_override_{ false };
+    InfoOutputLevel             info_output_level_{ InfoOutputLevel::kBasic };
+    bool                        uses_frame_markers_{ false };
+    std::string                 driver_info_;
+    std::shared_ptr<InfoWriter> info_writer_;
```
ncesario-lunarg: Would it be worth adding a brief markdown doc describing the design of these components and how they interact with one another ("Container" <- "API interface" <- "API-Specific instantiation")? Nothing in-depth, just a high level overview.

My initial understanding is that the writer is owned by the container, but having it be a `shared_ptr` would allow (imply?) different ownership.

tools/info/info_container.cpp
```diff
@@ -0,0 +1,697 @@
...
+bool InfoContainer::ProcessCommandLine(int32_t argc, const char** argv)
+{
+    // Save the app name first
+    std::string app_name     = argv[0];
+    size_t      dir_location = app_name.find_last_of("/\\");
+    if (dir_location >= 0)
+    {
+        app_name.replace(0, dir_location + 1, "");
+    }
+    app_name_ = app_name;
```
ncesario-lunarg: Not sure if we're ok using `std::filesystem` in general (I think we should considering we're using C++20), but I think this can be replaced with something like:
```cpp
const std::string app_name = std::filesystem::path{argv[0]}.filename().string();
```

tools/info/info_container.cpp
```diff
@@ -0,0 +1,697 @@
...
+
+    info_writer_ = std::make_shared<InfoWriter>();
```
ncesario-lunarg: I'm still a little confused why 1) this needs to be dynamically allocated and 2) why does it need to be ref-counted?

tools/info/info_container.cpp
```diff
@@ -0,0 +1,697 @@
...
+    WriteOutput("        \t\tdisplayed when abort() is called (Windows debug only).");
+#endif
+    WriteOutput("  --verbose\t\tOutput more information in JSON format");
+    WriteOutput(
+        "  --output\t\tOutput generated information to the provided file. If not defined output goes to std::out");
+
+    std::string api_specific_usage;
+    for (auto& api_if : api_interfaces_)
+    {
+        api_if->UpdateCommandLineUsage(api_specific_usage);
```
ncesario-lunarg: General comment/question for this pattern: I'm not sure I see the benefit to concatenating a string here vs either:
1. Having the derived `UpdateCommandLineUsage`s call `info_writer_->WriteOutput(...)`, or
2. Pass a `std::ostream` to `UpdateCommandLineUsage` to (potentially) write directly to the output stream

tools/info/info_container.cpp
```diff
@@ -0,0 +1,697 @@
...
+{
+    WriteOutput("File format info:");
+    FileFormatInfo file_format_info(file_processor_);
+    WriteOutput(std::string("\tFile format version: ") + std::to_string(file_format_info.major_version) + "." +
+                std::to_string(file_format_info.minor_version));
+    WriteOutput(std::string("\tFrame delimiters: " +
+                            GetFrameMarkerString(file_format_info.uses_frame_markers, file_format_info.NeedsUpdate())));
+}
+
+nlohmann::json InfoContainer::GetFileFormatInfoJson()
```
ncesario-lunarg: I think elsewhere we're using `ordered_json` so that objects will end up printing in an "expected order." Maybe that doesn't matter as much here, though?

tools/info/info_container.h
```diff
@@ -0,0 +1,180 @@
...
+    bool                                            api_restricted_output_{ false };
+    bool                                            force_all_api_output_{ false };
+    gfxrecon::decode::FileProcessor                 file_processor_;
+    gfxrecon::decode::StatDecoderBase               stat_decoder_;
+    gfxrecon::decode::StatConsumer                  stat_consumer_;
+    gfxrecon::decode::InfoConsumer                  info_consumer_;
+    gfxrecon::decode::InfoDecoder                   info_decoder_;
+    AnnotationRecorder                              annotation_recorder_;
+    ApiAgnosticStats                                api_agnostic_stats_;
+    std::shared_ptr<gfxrecon::util::ArgumentParser> argument_parser_;
```
ncesario-lunarg: Similar question about ownership here. Does this need to be ref-counted?

ncesario-lunarg (commented):
Some minor nits and questions, but I think this is great. I don't have a good sense of how well the output of the info tool is tested in our CI (the worry being that we could potentially break somebody's downstream script that scrapes the output of the info tool), but if that's not a big concern, then LGTM!

tools/info/info_api_interface.h
MarkY-LunarG: I intended it to be shared, but I could do a unique_ptr controlled by the InfoContainer and then share a raw pointer with everything else.  What's preferable?

The documentation is a good idea.  I can put something together.

MarkY-LunarG: Actually, forget the `unique_ptr`, just the instantiation of the object in the InfoContainer.

tools/info/info_container.cpp
MarkY-LunarG: Made an integrated part of the InfoContainer class, not a pointer.

tools/info/info_container.cpp
MarkY-LunarG: IDK, I didn't use it from the beginning because I wanted the order to mirror the actual output of the normal "info" messages.

tools/info/info_container.h
MarkY-LunarG: This was original a shared pointer because I returned it to info.cpp while I was transitioning over from the old info code to the new class design.  This is a left over of that transition and should not be a shared pointer.

tools/info/info_api_interface.h
ncesario-lunarg: > Actually, forget the `unique_ptr`, just the instantiation of the object in the InfoContainer.

That's what I was thinking. The `shared_ptr` is certainly more flexible as it would decouple the Writer's lifespan from the container, but I don't know if that makes sense. I'm good either way.

tools/info/info_container.h
```diff
@@ -0,0 +1,180 @@
...
+
+    bool OutputContent();
+
+    // Temporary - only for transition (will be removed)
+    std::string GetInputFilename()
+    {
+        return argument_parser_->GetPositionalArguments().size() > 0 ? argument_parser_->GetPositionalArguments()[0]
+                                                                     : "";
+    }
+    InfoApiInterface::InfoOutputLevel GetOutputLevel() { return output_level_; }
```
MarkY-LunarG: Will remove these two things, left over transition code.

framework/decode/file_processor.cpp
```diff
@@ -67,7 +67,6 @@ bool FileProcessor::Initialize(const std::string& filename)
     }
     else
     {
-        GFXRECON_LOG_ERROR("Failed to open file %s", filename.c_str());
```
bradgrantham-lunarg: Why delete this?

tools/info/info_api_interface.h
```diff
@@ -0,0 +1,146 @@
...
+#include <nlohmann/json.hpp>
+
+#include <iomanip>
+#include <memory>
+#include <sstream>
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(info)
+
+enum class FrameMarkerTypes : std::uint32_t
```
bradgrantham-lunarg: This isn't used anywhere.  What's your expectation for this?

tools/info/info_api_interface.h
```diff
@@ -0,0 +1,146 @@
...
+    // Output methods
+    void                   SetOutputLevel(InfoOutputLevel output_level) { info_output_level_ = output_level; }
+    virtual void           PrintInfo()    = 0;
+    virtual nlohmann::json GenerateJson() = 0;
+
+  protected:
+    inline void WriteOutput(const std::string& message) { info_writer_->Print(message); }
+
+    template <typename T>
+    std::string UintToHexString(T value, uint32_t min_width = 2, bool zero_fill = true, bool show_base = true)
```
bradgrantham-lunarg: We already have variants of this in framework/util/to_string.h, please use one of those variants if they fit or make a new variant in framework/util/to_string.h.

tools/info/info_container.cpp
```diff
@@ -0,0 +1,692 @@
...
+}
+
+bool InfoContainer::RegisterApiInterface(std::unique_ptr<InfoApiInterface> api_interface)
+{
+    for (auto& api_if : api_interfaces_)
+    {
+        if (api_if->ApiFamilyId() == api_interface->ApiFamilyId())
+        {
+            WriteError((std::string("Duplicate Api Interface for API ") +
+                        std::to_string(static_cast<uint32_t>(api_interface->ApiFamilyId())) + " found!\n")
```
bradgrantham-lunarg: You don't need to cast here, std::to_string will string-ify a uint16_t by promoting it to an int.

tools/info/info_api_interface.h
```diff
@@ -0,0 +1,146 @@
...
+enum class FrameMarkerTypes : std::uint32_t
+{
+    kImplicit = 0,
+    kExplicit,
+};
+
+class InfoApiInterface
+{
+  public:
+    enum class InfoOutputLevel : std::uint32_t
```
bradgrantham-lunarg: These feel less like levels and more like options.  Could options / mask be the right abstraction here?

tools/info/info_container.cpp
```diff
@@ -0,0 +1,692 @@
...
+    }
+    return true;
+}
+
+bool InfoContainer::OutputContent()
+{
+    if (output_level_ >= InfoApiInterface::InfoOutputLevel::kApiSpecificBegin &&
+        api_agnostic_stats_.error_state != gfxrecon::decode::BlockIOError::kErrorNone)
+    {
+        WriteOutput("A failure has occurred during file processing");
```
bradgrantham-lunarg: This error condition should be detected by the logic error above, and then OutputContent shouldn't be called at all.

tools/info/info_container.cpp
```diff
@@ -0,0 +1,692 @@
...
+    WriteOutput("  --exe-info-only\tQuickly exit after extracting captured application's executable name");
+    WriteOutput("  --file-format-only\tQuickly exit after extracting file format information");
+    WriteOutput("  --env-vars-only\tQuickly exit after extracting captured application's environment variables");
+#if defined(WIN32) && defined(_DEBUG)
+    WriteOutput("  --no-debug-popup\tDisable the 'Abort, Retry, Ignore' message box");
+    WriteOutput("        \t\tdisplayed when abort() is called (Windows debug only).");
+#endif
+    WriteOutput("  --verbose\t\tOutput more information in JSON format");
+    WriteOutput(
+        "  --output\t\tOutput generated information to the provided file. If not defined output goes to std::out");
```
bradgrantham-lunarg: replace-all on std changed "stdout"

bradgrantham-lunarg (changes requested):
Some thoughts, and looking forward to our discussion about the design!  in anticipation of that, I think we would be well-served by not having the API modules allowed to access a writer at all except for logging.  GFXRECON_LOG_* already exists and has a whole mechanism for setting the level, etc, and the modules can use that.  Having an InfoWriter seems to just wrap that and not add anything the Logger doesn't already have.  I'd ideally like the API modules to return strings for the "terse" gfxrecon-info output or (even better) only return JSON and have the info tool itself distill the "terse" output from that.

framework/decode/file_processor.cpp
MarkY-LunarG: If you read the commit messages, this results in a duplicate message of the file failing to open.  I removed the one that made the most sense to me.

tools/info/info_api_interface.h
MarkY-LunarG: I had used it at one point, I'll just remove this.  I think my original design expected the API-specific sections to return this as an indication of what they thought and it ended up not being needed.

tools/info/info_container.cpp
```diff
@@ -0,0 +1,692 @@
...
+            break;
+        case InfoApiInterface::InfoOutputLevel::kEnvironmentInfo:
+            PrintEnvironmentVariableInfo();
+            break;
+        case InfoApiInterface::InfoOutputLevel::kFileInfo:
+            PrintFileFormatInfoText();
+            break;
+        case InfoApiInterface::InfoOutputLevel::kBasic:
+            PrintExeInfo();
+            if (api_agnostic_stats_.error_state != gfxrecon::decode::BlockIOError::kErrorNone)
```
MarkY-LunarG: This section is a duplicate of the above.  I'll remove it.

tools/info/info_container.cpp
MarkY-LunarG: Actually, you are correct, I had not followed it all the way to where `api_agnostic_stats_.error_state` was set.  Following it, I see that it is identical to the above.  I'll delete it.

tools/info/info_container.cpp
```diff
@@ -0,0 +1,692 @@
...
+            if (!detected_apis_.size())
+            {
+                detected_apis_.push_back("Unable to detect captured APIs");
+            }
+            json_base_["detected-apis"] = detected_apis_;
+
+            if (api_agnostic_stats_.error_state != gfxrecon::decode::BlockIOError::kErrorNone)
+            {
+                return false;
+            }
```
MarkY-LunarG: same here

tools/info/info_api_interface.h
MarkY-LunarG: Converting over to the to_strings versions (with some changes)

tools/info/info_container.cpp
MarkY-LunarG: This code will be removed.

tools/info/info_api_interface.h
MarkY-LunarG: Let me look.  It's a good thought.

MarkY-LunarG: How do you like the OutputFlags I added with  the latest commit instead?

MarkY-LunarG:
> Some thoughts, and looking forward to our discussion about the design! in anticipation of that, I think we would be well-served by not having the API modules allowed to access a writer at all except for logging. GFXRECON_LOG_* already exists and has a whole mechanism for setting the level, etc, and the modules can use that. Having an InfoWriter seems to just wrap that and not add anything the Logger doesn't already have. I'd ideally like the API modules to return strings for the "terse" gfxrecon-info output or (even better) only return JSON and have the info tool itself distill the "terse" output from that.

I updated the methods from "PrintInfo" to "GenerateInfo" and return  a string.  I still think there's nothing wrong with a dedicated writer, and I am concerned at all the string concatenation going on inside of each API generator.

I disagree on returning only JSON and processing in the info_generator right now.  That would be a huge change and the data is still already  coming from a common source and I don't see any benefit of that over the existing design.  I believe the goal of my project is met with the current design and that might be something to explore in the future.

tools/info/info_api_generator.h
```diff
@@ -0,0 +1,170 @@
...
+    // child classes with this base class.
+
+    // Define a method that can be used casts a child unique_ptr to parent
+    // so it can be included in a std::vector we'll call the registry.
+    using BaseApiGeneratorPtr = std::function<std::unique_ptr<InfoApiGenerator>()>;
+    static void RegisterGenerator(BaseApiGeneratorPtr gen_ptr) { GetRegisteredGenerators().push_back(gen_ptr); }
+    static std::vector<BaseApiGeneratorPtr>& GetRegisteredGenerators()
+    {
+        static std::vector<BaseApiGeneratorPtr> registered_Generators_;
+        return registered_Generators_;
```
ncesario-lunarg:
```suggestion
        return registered_generators_;
```

tools/info/info_generator.cpp
```diff
@@ -0,0 +1,686 @@
...
+{
+    gfxrecon::util::Log::Init();
+
+    // Query the base InfoApiGenerator class which should have a vector
+    // of all child classes that registered with it.
+    // They are unique_ptrs so we need to move control to our list.
+    for (const auto& registered_type : InfoApiGenerator::GetRegisteredGenerators())
+    {
+        api_generators_.push_back(std::move(registered_type()));
+    }
```
ncesario-lunarg: A couple questions; I'm putting these question here, but they basically revolves around needing to store a vector of lambdas and the broader "register API" design:

1. Is/should `InfoGenerator` be a singleton?
2. Should `*ApiGenerators`s be singletons?
3. Why do all of the generators need to be registered at construction time here? If that needs to be enforced, would it be clearer to have the constructor take something like an r-value vector of generators? Having a vector of lambdas makes me think we need to delay the construction of what we're registerring. That might make sense if `*ApiGenerator`s need to have a reference to `InfoGenerator`, but it doesn't look like that is the case.

tools/info/main.cpp
```diff
@@ -21,1501 +21,17 @@
...
-    else if (arg_parser.IsOptionSet(kEnvVarsOnlyOption))
-    {
-        success = GatherAndPrintEnvVars(input_filename);
-    }
-    else if (arg_parser.IsOptionSet(kFileFormatOnlyOption))
-    {
-        success = GatherAndPrintFileFormatInfo(input_filename);
-    }
-    else
+    if (info_generator.ProcessCommandLine(argc, argv) && info_generator.ProcessCapture() &&
```
ncesario-lunarg: It looks like a design goal here is that: if an `*ApiGenerator` is _built/compiled_, it will be registered. Is that correct?

tools/info/info_api_generator.h
MarkY-LunarG: Yep, copy paste after renaming them from "Interfaces"   Good catch.

tools/info/info_api_generator.h
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    enum class OutputSelectionFlags : std::uint32_t
```
ncesario-lunarg: You might be able to use `std::bitset` together with `std::bitset::reference` to create variables that reference each bit. Then you shouldn't need to overload all of the bit operators below.

framework/decode/file_processor.cpp
bradgrantham-lunarg: I don't think you can delete this here.  SetActiveFile is called from other places, e.g. when asset files are in use.  Deleting this would potentially lose information about a file open failure for an asset file.

bradgrantham-lunarg:
> > Some thoughts, and looking forward to our discussion about the design! in anticipation of that, I think we would be well-served by not having the API modules allowed to access a writer at all except for logging. GFXRECON_LOG_* already exists and has a whole mechanism for setting the level, etc, and the modules can use that. Having an InfoWriter seems to just wrap that and not add anything the Logger doesn't already have. I'd ideally like the API modules to return strings for the "terse" gfxrecon-info output or (even better) only return JSON and have the info tool itself distill the "terse" output from that.
> 
> I updated the methods from "PrintInfo" to "GenerateInfo" and return a string. I still think there's nothing wrong with a dedicated writer, and I am concerned at all the string concatenation going on inside of each API generator.
> 
> I disagree on returning only JSON and processing in the info_generator right now. That would be a huge change and the data is still already coming from a common source and I don't see any benefit of that over the existing design. I believe the goal of my project is met with the current design and that might be something to explore in the future.

There is a lot of string concatenation going on, and I agree that could be a concern. How big is the maximum string size going to be for text info output and how long will that take relative to parsing the GFXR file?

tools/info/info_api_generator.h
MarkY-LunarG: That didn't work, it treats them mostly as integers.  I went with the struct of booleans as  Brad  suggested.

tools/info/info_api_interface.h
MarkY-LunarG: Moved to struct of bools

tools/info/main.cpp
MarkY-LunarG: Yes.  I wanted it automatic because any logic here would on knowing what could be created otherwise.

tools/info/info_generator.cpp
MarkY-LunarG: Created a new class called "FeatureModuleRegistry" (framework/util/feature_module_registry.h) and that is a singelton.

MarkY-LunarG:
@bradgrantham-lunarg / @ncesario-lunarg , the changes discussed have been made. 

Because I've reverted the main.cpp back to more like the original file, it is probably best to final review the sum total of all commits instead of one at a  time.

framework/decode/file_processor.cpp
MarkY-LunarG: Restored.

MarkY-LunarG:
Removed `WIP`

MarkY-LunarG:
> There is a lot of string concatenation going on, and I agree that could be a concern. How big is the maximum string size going to be for text info output and how long will that take relative to parsing the GFXR file?

Sorry, I just saw this comment.  It's not bad since the string concatenation code is only used for the simple output (not verbose) and there isn't as much content in those paths.  So most strings would be in the <5k lines at most range by the end.

MarkY-LunarG:
@bradgrantham-lunarg anything else for this change?

tools/info/info_feature.h
```diff
@@ -0,0 +1,92 @@
...
+    // Simple "getter" style methods
+    virtual std::string Label() const = 0;
+    virtual bool        WasDetected() = 0;
+    virtual std::string CompiledHeaderVersionString() const { return ""; }
+    virtual uint32_t    GetBlankFrameCount() { return 0; }
+    virtual uint32_t    GetFrameStart() const { return 0; }
+    virtual bool        DesiresSingleLineFrameOutput() const { return false; }
+
+    // A few "setter" style methods
+    virtual void SetFrameMarkerUsage(bool found) { uses_frame_markers_ = found; }
```
bradgrantham-lunarg: This is never called but anyway I'd prefer this be passed into the functions where it's needed.

tools/info/info_feature.h
```diff
@@ -0,0 +1,92 @@
...
+    virtual void SetDriverInfoString(const std::string& driver_info) { driver_info_ = driver_info; }
```
bradgrantham-lunarg: Only D3D12 cares about DriverInfo because only D3D12 captures have a DriverInfo block...  Maybe this PR should be followed ASAP by a PR that moves processing the DriverInfo block into a dx12_consumer.  Or at least pass this in where it's needed rather than stored per-feature.  Or Maybe the API info features just have access to the InfoConsumer?

tools/info/info_feature.h
```diff
@@ -0,0 +1,92 @@
...
+        GFXRECON_UNREFERENCED_PARAMETER(arg_parser);
+        return true;
+    }
+
+    // Method to register this feature's decoder elements with the containers
+    // FileProcessor
+    virtual void RegisterDecodeComponents(gfxrecon::decode::FileProcessor& file_processor) = 0;
+
+    // Indicates that this API generator requires an API-specific output
+    bool RestrictingOutput() { return restricting_output_; }
```
bradgrantham-lunarg: This causes strange interactions between command-line arguments and the feature modules and the info loop.  Since this is only used by --enum-gpu-indices and not even dependent on the cpature file, I think the best course of action is to just check that option in main.cpp, dump the list like it was before, and exit, and don't dirty the feature interface with it.

tools/info/main.cpp
bradgrantham-lunarg: This function is no longer called and is API specific and should be deleted.

bradgrantham-lunarg: This function is no longer called and is API specific and should be deleted.

bradgrantham-lunarg: This #ifdef and function is no longer called and is API specific and should be deleted.

tools/info/info_d3d12_feature.cpp
```diff
@@ -0,0 +1,406 @@
...
+#include "util/feature_module_registry.h"
+#include "util/to_string.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(info)
+
+// Register this class as a feature in a module registry
+GFXR_UTIL_REGISTER_FEATURE_CREATOR(InfoFeature, InfoD3d12Feature)
+
+const char kEnumGpuIndices[] = "--enum-gpu-indices";
```
bradgrantham-lunarg: Along with the other comment in here about enum gpu indices, this should come out as well as the code in UpdateValidCommandLineOptionsArgs and GetCommandLineUsage in favor of a one-off option in main.cpp.

tools/info/main.cpp
bradgrantham-lunarg: This function isn't called and should be deleted.

bradgrantham-lunarg: This function isn't called and should be deleted.

tools/info/main.cpp
```diff
@@ -555,7 +506,7 @@ void PrintApiAgnosticStatsText(const gfxrecon::decode::FileProcessor& file_proce
         }
 
         // Print out the total frames and range based on the API (since we have 2 different ways of showing it)
-        if (vulkan_present)
+        if (print_single_line)
```
bradgrantham-lunarg: This is the API reaching back up to change the info print.  We're lucky that this is the last thing in the API agnostic stats.  Move it into the API-specific text function as the first thing, and may have to pass the API agnostic stats to the API-specific text function.

bradgrantham-lunarg (changes requested):

tools/info/info_feature.h
MarkY-LunarG: Fixed

tools/info/main.cpp
bradgrantham-lunarg: Ah, I see why this doesn't work.  The API-specific functions are called for each API, while the API agnostic states are called once.  🤦 Need to soak on this for a little bit.

tools/info/info_feature.h
MarkY-LunarG: Moved it.

MarkY-LunarG:
Ok, pushed up a squashed commit.

bradgrantham-lunarg (approved):

PR #2752, opened 2026-03-04
Support ARM64EC and ARM64X builds

richard-lunarg:
To fully support building gfxreconstruct as ARM64X (needed for the Vulkan SDK for Windows ARM), dependent libraries must also be built as both ARM64 and ARM64EC. The compression libraries are rebuilt both ways here and then "spliced" together. zlib, zstd, and lz4 are all built as both ARM64EC and ARM64, and for both Release and Debug modes. The libraries were renamed after building and combined as follows:

link.exe /lib /machine:arm64x /out:liblz4_static.lib liblz4_static-ARM64.lib liblz4_static-ARM64EC.lib

Etc. Etc.

I've added a README.txt file to the /external/precompiled/win64-arm folder for future reference.


bradgrantham-lunarg:
Are there sections of our internal CI Windows-ARM build log that demonstrate these are being used in our CI?

bradgrantham-lunarg:
> Are there sections of our internal CI Windows-ARM build log that demonstrate these are being used in our CI?

We can't currently test this because we can't automatically capture an x86_64 binary on Windows-ARM.

mikes-lunarg (changes requested):
Can you double check how the static libs were produced? I don't have a Win/ARM machine to test it but I did peek at it with llvm-readobj and saw a couple of strange things:

- `liblz4_staticd.lib` contains no ARM64EC object files
- `zlib.lib` and `zlibd.lib` seem to contain 2 copies of each ARM64 object file (but only 1 of each ARM64EC object file)

richard-lunarg:
I used vkConfig to do a capture of the native vkCube.exe on Windows ARM and play it back. I then copied the x86_64 vkcube.exe to the ARM machine, configured vkConfig to run IT, and it also captured a file that I could play back. This is a good test I think and I am going to add it to the manual test document since we don't have a CI test for this scenario.

richard-lunarg:
> Can you double check how the static libs were produced? I don't have a Win/ARM machine to test it but I did peek at it with llvm-readobj and saw a couple of strange things:
> 
> * `liblz4_staticd.lib` contains no ARM64EC object files
> * `zlib.lib` and `zlibd.lib` seem to contain 2 copies of each ARM64 object file (but only 1 of each ARM64EC object file)

Yes and a good exercise. Odd that I could rebuild gfxreconstruct as ARM64, ARM64EC, or ARM64X with the libraries put together wrong. They are correct now. I verified with dumpbin, and learned to use llvm-readobj.

PR #2753, opened 2026-03-04
jenkins: second effort to get pipelines using venv

lunarpapillo:
The first attempt here:

    jenkins: use Python venv
    https://github.com/LunarG/gfxreconstruct/pull/2722

was completely wrong; it modified files "runJob.bat", "runJob.sh", and "runJobAndroid.sh" that aren't actually used in CI.

This effort modifies the "runTest.bat", "runTest.sh", and "runTestAndroid.sh" files instead, which seems to be the correct way to affect CI behavior.

This PR temporarily includes a change to "test.ref" so that its testing will also evoke the latest VulkanTests changes (which require a new Python module, and will clearly fail if run without a Python virtual environment).  The test.ref changes will be kept only for testing, and will be removed before being merged.


ci/runTest.sh
```diff
@@ -1,4 +1,8 @@
 cmake --version
 python3 --version
 
-python3 VulkanTests/gfxrecontest.py --build-mode ${BUILD_MODE} --bits ${BITS} $LINUX_TEST_ARGS --suite "ci-gfxr-suites/$GFXRECON_TRACE_SUBDIR/$TEST_SUITE" --trace-dir "$GFXRECON_TRACE_DIR" --result-dir "$RESULTS_DIR"
+echo creating Python virtual environment in $WORKSPACE/python-venv...
+python3 -m venv $WORKSPACE/python-venv
+$WORKSPACE/python-venv/bin/python3 -m pip install --no-cache-dir -r VulkanTests/requirements.txt > $WORKSPACE/python-venv.txt 2>&1
```
antonio-lunarg: `runTest.bat` exits on failure here. Should we do the same on Linux/Android?

ci/runTest.bat
```diff
@@ -1,4 +1,8 @@
 cmake --version
 python --version
 
-python VulkanTests\gfxrecontest.py --compiler vs2022 --build-mode %BUILD_MODE% --bits %BITS% %WINDOWS_TEST_ARGS% --suite "ci-gfxr-suites\%GFXRECON_TRACE_SUBDIR%\%TEST_SUITE%" --trace-dir "%GFXRECON_TRACE_DIR%" --result-dir "%RESULTS_DIR%"
+echo creating Python virtual environment in %WORKSPACE%\python-venv...
+python -m venv %WORKSPACE%\python-venv
```
antonio-lunarg: Paths are unquoted (both here and in other scripts), which might be an issue if `WORKSPACE` contains spaces?

ci/runTest.sh
lunarpapillo: It does exit by default (with an error) on Linux/Android/Mac.  Windows `.bat` is unusual that it continues on this error, and the return value can be missed if a later command runs correctly.

This is a subtlety of Jenkins on Windows - would you like a comment in the script explaining the issue, or should this be considered to be Jenkins domain knowledge?

ci/runTest.bat
lunarpapillo: True.  Our Jenkins setup prohibits the Jenkins root and Jenkins job names from including spaces (which are the only ways  spaces can appear in a workspace name).  But there's no reason not to be safe "just in case" here.  Will fix.

ci/runTest.sh
lunarpapillo: I added a comment to `runTest.bat` explaining why it had to explicitly exit.

antonio-lunarg (approved):

lunarpapillo:
Passed LunarG CI: http://tcubuser.lunarg.localdomain:8080/job/gfxr-pipelines/job/gfxr/view/change-requests/job/PR-2753/

image

Will rebase, squash, force push, and merge.

PR #2763, opened 2026-03-07
jenkins: test display resolution

lunarpapillo:
DO NOT MERGE.  This is a sanity test to ensure that GFXR will run correctly (with a venv) when there's a new VulkanTests module in requirements.txt.


lunarpapillo:
Test ran successfully, demonstrating that GFXR can now run correctly with a modified `venv`.  Closing and deleting the original test branch.

PR #2766, opened 2026-03-09
Vdr rework queue submit

panos-lunarg:
Fixing two issues:
- The submission order of command buffers was not being preserved when dumping resources.
- DrawCallsDumpingContext's command buffers do not replicate the original
command buffer until the end. This can leave queries in the middle which
can lead to hangs.


hongSARC:
I have tested, vulkan sample "terrain_tessellation", and the issue seems to be fixed. Wasn't able to reproduce the issue anymore.

framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -1838,188 +1869,143 @@ VkResult VulkanReplayDumpResourcesBase::QueueSubmit(const std::vector<VkSubmitIn
...
+    std::vector<std::shared_ptr<DispatchTraceRaysDumpingContext>> dispatch_contexts;
+
     if (!output_json_per_command)
     {
         active_delegate_->DumpStart();
     }
 
-    if (pre_submit)
+    const size_t submit_count = submit_infos.size();
+    for (size_t si = 0; si < submit_count; ++si)
```
bradgrantham-lunarg: Could this just be `for(auto si: submit_infos)`?

framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -1838,188 +1869,143 @@ VkResult VulkanReplayDumpResourcesBase::QueueSubmit(const std::vector<VkSubmitIn
...
-            GFXRECON_NARROWING_ASSIGN(modified_submit_infos[s].commandBufferCount,
-                                      modified_command_buffer_handles[s].size());
-            modified_submit_infos[s].pCommandBuffers    = modified_command_buffer_handles[s].data();
-        }
-        else
-        {
-            modified_submit_infos[s].commandBufferCount = 0;
-            modified_submit_infos[s].pCommandBuffers    = nullptr;
-        }
+#define CHECK_VK_ERROR(_res_, _func_)                                                                               \
```
bradgrantham-lunarg, line 1873: Putting a macro inside a function is unusual; did you mean to put this at file scope?

framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -1838,188 +1869,143 @@ VkResult VulkanReplayDumpResourcesBase::QueueSubmit(const std::vector<VkSubmitIn
...
-            modified_submit_infos[s].signalSemaphoreCount = 0;
+        if (!submit_cbs.empty())
+        {
+            modified_submit_info.commandBufferCount = static_cast<uint32_t>(submit_cbs.size());
+            modified_submit_info.pCommandBuffers    = submit_cbs.data();
+            res = device_table.QueueSubmit(queue_info->handle, 1, &modified_submit_info, submission_fence.handle);
+            submit_cbs.clear();
         }
+
+        CHECK_VK_ERROR(res, "QueueSubmit")
```
bradgrantham-lunarg: If `submit_cbs.empty()` and the other clauses above aren't entered then `res` might be uninitialized here.  Hard to be sure; I haven't traced the logic to be sure this isn't always assigned.  But maybe assign `VK_SUCCESS`?

framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -1838,188 +1869,143 @@ VkResult VulkanReplayDumpResourcesBase::QueueSubmit(const std::vector<VkSubmitIn
...
+
+        // We need to wait for the fence in for VkSubmitInfos. Either there are transfer/dispatch dumped resources
+        // or we just need to wait before destroying it (prevent use after free)
+        res = submission_fence.Wait();
+        CHECK_VK_ERROR(res, "WaitForFences")
     }
 
-    if (!output_json_per_command)
+    for (auto transfer_context : transfer_contexts)
```
bradgrantham-lunarg: Every pass through this loop and others containing shared_ptrs adds and removes a reference on the shared_ptrs; harmless but doesn't need to happen.  Could this be `const auto&`?  I think `DumpTransferCommands()` would need to be `const` then but it seems like that might have deep consequences.

framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -1838,188 +1869,143 @@ VkResult VulkanReplayDumpResourcesBase::QueueSubmit(const std::vector<VkSubmitIn
...
+                    modified_submit_info.pWaitSemaphores      = nullptr;
+                    modified_submit_info.signalSemaphoreCount = 0;
+                    modified_submit_info.pSignalSemaphores    = nullptr;
 
-                submitted = true;
+                    // Insert original command buffer in the vector for submission. If has_transfer_or_dispatch is true
+                    // then the command buffer has already been submitted
+                    if (!has_transfer_or_dispatch)
+                    {
+                        res = submission_fence.Reset();
```
bradgrantham-lunarg: Does this always happen after the previous `submission_fence.Wait()`?  Could this be skipped? Then the next QueueSubmit in the loop would be invalid usage.

framework/decode/vulkan_replay_dump_resources.cpp
panos-lunarg: Yes. The assignment of the `last_submit_info` must be converted from
`const bool  last_submit_info  = (si == submit_count - 1);`
to
`const bool last_submit_info  = (&si == &submit_infos.back());`

framework/decode/vulkan_replay_dump_resources.cpp
panos-lunarg: No I put it there intentionally with the reasoning that is used only from this function. I understand it's unusual and not a common practice in our codebase. I can remove and copy paste the macro's definition in the places it is called if you insist.

framework/decode/vulkan_replay_dump_resources.cpp
panos-lunarg: done

framework/decode/vulkan_replay_dump_resources.cpp
panos-lunarg: I turned `auto` into `auto&` which I think should remove the overhead of changing the ref counter of the shared_ptr. Adding const is not that simple. I will look into it but I think it would be better done in a separate PR

framework/decode/vulkan_replay_dump_resources.cpp
panos-lunarg: Good point. Each SubmiInfo uses a different fence, however inside a SubmitInfo the same fence can be reused. Moved the call to `Reset` a few lines up, outside the if's scope

framework/decode/vulkan_replay_dump_resources.cpp
bradgrantham-lunarg: SGTM!

PR #2770, opened 2026-03-10
Remove 32 bit support

ziga-lunarg:


CMakeLists.txt
```diff
@@ -64,6 +64,10 @@ if (NOT DEFINED HEADLESS)
     set(HEADLESS TRUE)
 endif ()
 
+if (CMAKE_SIZEOF_VOID_P EQUAL 4)
```
MarkY-LunarG, line 67: Should we modify the message to say something like "If you want 32-bit build support, sync to version XYZ"  and also make a tag of that name before we commit this?

scripts/build.py
```diff
@@ -45,7 +45,7 @@ def is_mac():
     '''
     return 'darwin' == platform.system().lower()
 
-ARCHITECTURES = ['x64', 'x86', 'arm', 'arm64']
+ARCHITECTURES = ['x64', 'arm', 'arm64']
```
MarkY-LunarG: Shouldn't "arm" also be removed?

MarkY-LunarG (commented):
Couple of questions.

scripts/build.py
ziga-lunarg: Good catch

CMakeLists.txt
ziga-lunarg: To me that doesn't seem like a good idea, since that will stop receiving updated. I wouldn't want to see issues submitted against that old tag, when we don't support it anymore.

PR #2771, opened 2026-03-10
Revert 2740

bradgrantham-lunarg:
Probably #2740 is good, but it caused extended suite failures, so revert it and follow on with a version of 2740 with either a fix to the code or an updated capture-draw-calls reference.


bradgrantham-lunarg:
Passes at tests on just multithreaded-dr: http://jenkins:8080/job/gfxr-pipelines/job/gfxr-manual/60/pipeline-overview/.  Will force push without subsetted test suite ref..

zongdu-arm:
> Passes at tests on just multithreaded-dr: http://jenkins:8080/job/gfxr-pipelines/job/gfxr-manual/60/pipeline-overview/. Will force push without subsetted test suite ref..

Hi @bradgrantham-lunarg ,
Which test case will fail? Could you provide details of the failure? I can't open this web link.

PR #2774, opened 2026-03-11
replay: Add --idle-before-submit option

antonio-lunarg:
Add a replay option to wait for the GPU to become idle before submitting each command buffer to a queue.

### DIFF
--- a/framework/decode/vulkan_replay_options.h
+++ b/framework/decode/vulkan_replay_options.h
@@ -171,6 +171,9 @@ struct VulkanReplayOptions : public ReplayOptions
     // Milliseconds to wait before first queue submit.
     uint32_t wait_before_first_submit{ 0 };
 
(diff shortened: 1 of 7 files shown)


framework/decode/vulkan_replay_options.h
```diff
@@ -171,6 +171,9 @@ struct VulkanReplayOptions : public ReplayOptions
     // Milliseconds to wait before first queue submit.
     uint32_t wait_before_first_submit{ 0 };
 
+    /// Wait for the GPU to become idle before submitting each work.
+    bool idle_before_submit{ 0 };
```
fabian-lunarg: nit -> init with false

fabian-lunarg: and maybe revise comment "Wait for the GPU to become idle before each submit"

PR #2783, opened 2026-03-13
Async File read process and dispatch...

jzulauf-lunarg:
Async framework applied to synchronous preload.   Application to other tools in follow-on PR's.

### DIFF
--- a/framework/application/application.cpp
+++ b/framework/application/application.cpp
@@ -66,7 +66,7 @@ Application::Application(const std::string&     name,
                          const std::string&     cli_wsi_extension,
                          void*                  platform_specific_wsi_data) :
     name_(name),
-    file_processor_(file_processor), running_(false), paused_(false),
+    file_processor_(file_processor), async_processing_(false), running_(false), paused_(false),
     pause_frame_(std::numeric_limits<uint32_t>::max()), cli_wsi_extension_(cli_wsi_extension),
     fps_info_(nullptr), frame_loop_info_{ nullptr }
 {
@@ -130,6 +130,34 @@ void Application::SetFpsInfo(graphics::FpsInfo* fps_info)
 void Application::Run()
 {
     running_ = true;
+
+    if (async_processing_)
+    {
+        if (fps_info_ != nullptr)
+        {
+            auto preload_range = fps_info_->GetPreloadFrameRange();
+            if (preload_range.has_value())
+            {
+                // Fileprocessor frames are 0 based, fps_info frames are 1 based,
+                file_processor_->SetPreloadFrameRange(
+                    decode::FileProcessor::FrameRange::MakeFromOneBased(preload_range->first, preload_range->second));
+            }
+            // Prevent async "run on"
+            file_processor_->SetQuitBeforeFrame(fps_info_->GetQuitBeforeFrame() - 1);
+        }
+
+        if (frame_loop_info_ != nullptr)
+        {
+            // NOTE: Loop frame support is "experimental" and should not be used with preload (though it
+            //       is currently not prevented).  This will be resolved in future work
+            //
+            // loop_frame_idx_ is 1-based; SetQuitBeforeFrame is 0-based, a loop frame of zero is not
+            // allowed
+            file_processor_->SetQuitBeforeFrame(frame_loop_info_->GetLoopFrame() - 1);
+        }
+
+        file_processor_->StartAsyncProcessing();
+    }
 
     while (running_)
     {
--- a/framework/decode/async_processor.cpp
+++ b/framework/decode/async_processor.cpp
@@ -1,0 +1,496 @@
+/*
+** Copyright (c) 2018-2020,2022 Valve Corporation
+** Copyright (c) 2018-2020,2026 LunarG, Inc.
+** Copyright (c) 2022-2025 Advanced Micro Devices, Inc. All rights reserved.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and associated documentation files (the "Software"),
+** to deal in the Software without restriction, including without limitation
+** the rights to use, copy, modify, merge, publish, distribute, sublicense,
+** and/or sell copies of the Software, and to permit persons to whom the
+** Software is furnished to do so, subject to the following conditions:
+**
+** The above copyright notice and this permission notice shall be included in
+** all copies or substantial portions of the Software.
+**
+** THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+** IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+** FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+** AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+** FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+** DEALINGS IN THE SOFTWARE.
+*/
+
+#include "util/defines.h"
+#include "util/logging.h"
+#include "decode/async_processor.h"
+#include "decode/file_processor.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(decode)
+
+AsyncProcessor::~AsyncProcessor()
+{
+    if (async_thread_.joinable())
+    {
+        keep_alive_.store(false, std::memory_order_release);
+        throttle_cv_.notify_all();
+        async_thread_.join();
+    }
+}
+
+void AsyncProcessor::LaunchAsyncThread()
+{
+    GFXRECON_ASSERT(!async_thread_.joinable());
+    keep_alive_.store(true, std::memory_order_release);
+    async_thread_ = std::thread([this]() { this->ThreadMain(); });
+
+    // Waits until the first batch is flushed, or the queue is closed without flushing.
+    // To shorten the block, the ThreadMain() enqueues a "Continue" batch holding only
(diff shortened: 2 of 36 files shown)


jzulauf-lunarg:
@antonio-lunarg -- rebased on top of latest looping... need to commit the refactor, first.

tools/convert/main.cpp
```diff
@@ -402,6 +405,15 @@ int main(int argc, const char** argv)
             dx12_json_consumer.Initialize(&json_writer);
 #endif
 
+            if (async_processing)
+            {
+                if (frame_range_option)
+                {
+                    file_processor.SetQuitBeforeFrame(frame_indices.front() + 1);
```
antonio-lunarg, line 412: `front() + 1`? Should that be `back() + 1`?

jzulauf-lunarg:
```
        std::sort(indices.begin(), indices.end());
        std::reverse(indices.begin(), indices.end());
 ```

**Comment by jzulauf-lunarg** [comment]
Ready for review.  The changes marked "SQ4PR" will be squashed before commit but may be useful for understanding the design flow.

File: framework/graphics/fps_info.cpp [hunk 2]
```diff
@@ -83,6 +83,25 @@ bool FpsInfo::ShouldQuit(uint64_t frame)
...
+// Return exclusive quit frame;
+uint64_t FpsInfo::GetQuitBeforeFrame()
+{
+    uint64_t quit_frame = std::numeric_limits<uint64_t>::max();
+    if (quit_after_range_)
+    {
+        // end frame is exclusive
+        quit_frame = std::min(quit_frame, measurement_end_frame_);
+    }
+    if (quit_after_frame_)
```
MarkY-LunarG, line 95: Should we assert or throw a warning if both are set?

framework/application/application.cpp
```diff
@@ -131,6 +131,34 @@ void Application::Run()
...
+        }
+
+        if (frame_loop_info_ != nullptr)
+        {
+            // NOTE: Loop frame support is "experimental" and should not be used with preload (though it
+            //       is currently not prevented).  This will be resolved in future work
+            //
+            // loop_frame_idx_ is 1-based; SetQuitBeforeFrame is 0-based, a loop frame of zero is not
+            // allowed
+            file_processor_->SetQuitBeforeFrame(frame_loop_info_->GetLoopFrame() - 1);
```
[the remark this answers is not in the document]
jzulauf-lunarg: Application frame counting is 1 based.  FileProcessor frame counting is 0 based.

framework/graphics/fps_info.cpp
jzulauf-lunarg: There are several places we should revisit for flag consistency.  This is one.  Frame looping vs. preloading is another.

framework/util/thread_safe_queue.h
```diff
@@ -0,0 +1,274 @@
+
```
jzulauf-lunarg, line 1: @fabian-lunarg @MarkY-LunarG @mikes-lunarg  -- (and any other multi-thread folks) give this a careful look.

framework/decode/file_processor.cpp
```diff
@@ -55,8 +53,30 @@ void FileProcessor::WaitDecodersIdle()
     {
         decoder->WaitIdle();
     }
+}
+
+void FileProcessor::StartAsyncProcessing()
+{
+    GFXRECON_ASSERT(process_frame_number_ == kFirstFrame);
+    // async_processing_
+    async_thread_ = std::thread([this]() { this->ProcessBlocksAsync(); });
```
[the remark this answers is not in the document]
jzulauf-lunarg: The thread join is in the destructor of the AsyncProcessor .

framework/decode/async_processor.cpp
```diff
@@ -0,0 +1,495 @@
...
+
+void AsyncProcessor::ThrottleQueue()
+{
+    // Lowwater/Highwater checks
+    //  * The highwater and lowwater checks are based on pending batches
+    //  * Priority is given to lowwater, to prevent starvation of the consumer
+    //  * Preventing starvation of the consumer and the cost of one or two additional batches pending is a small memory
+    //    and batch overhead.
+    //
+    // Since last_dequeued_batch_ is set from the BlockBatch::tag it must be <= enqueued_batch_index_
```
antonio-lunarg, line 107: This sounds like a `GFXRECON_ASSERT`, worth adding it?

framework/decode/async_processor.cpp
```diff
@@ -0,0 +1,495 @@
+/*
+** Copyright (c) 2018-2020,2022 Valve Corporation
+** Copyright (c) 2018-2020,2022 LunarG, Inc.
```
antonio-lunarg: Heavy changes here, let's add 2026.

framework/util/thread_safe_queue.h
antonio-lunarg: Unit tests for this would be gold.

framework/decode/file_processor.h
```diff
@@ -431,19 +313,24 @@ class FileProcessor
     std::vector<format::FileOptionPair> file_options_;
     format::EnabledOptions              enabled_options_;
     std::vector<uint8_t>                uncompressed_buffer_;
-    uint64_t                            block_limit_;
+    uint64_t                            block_limit_{ 0 }; // No block limit by default
     bool                                pending_capture_uses_frame_markers_{ false };
```
antonio-lunarg, line 316: Which members are owned by which thread during async, and what guarantees the main thread never reads parse-thread-mutated state? Asking just to avoid any chache-line sharing here.

jzulauf-lunarg: This is a value unchanged after being set.  So MESI shared state would be worst case.  There are a bunch of state like this going into the frame processing loop.  The sync and async paths assume they don't change, but they aren't const.  (preload, frame loop, block limit are three examples). We could abstract them into a const "frame loop state"  struct, but that starts to impact all callers of the FileProcess::*Frame* interface which is broader than we want... 

I'll review for what FileProcessor members are accessed by the async thread... for example the "process_*" variables to make sure they are well commented.  Block limit isn't, clearly.

The other choice could be to package up the ProcessBlocks state and leave FileProcessor out of it... hmmm.

You think this needs cleanup for this commit or in a followup?

antonio-lunarg: I'd opt for a followup if not straightforward.

framework/util/thread_safe_queue.h
```diff
@@ -0,0 +1,274 @@
...
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(util)
+
+// Lots of details not to pollute the util namespace with so
+GFXRECON_BEGIN_NAMESPACE(thread_safe_queue_detail)
+// Queue implmentation base class
+//
+// There is no producer side throttling or queue size limitation.
+// A `size() method is intentionally non supported as it will either be inacurate or non performant
```
MarkY-LunarG: extraneous `?

framework/util/thread_safe_queue.h
```diff
@@ -0,0 +1,274 @@
...
+//
+// Further details in each derived Queue below.
+template <typename T>
+class QueueBase
+{
+  public:
+    using value_type = T;
+    using pop_type   = std::optional<value_type>;
+
+    bool live() const
```
MarkY-LunarG: Personally, I prefer queries to clearly indicate they are a query.  Like "is_alive".

framework/util/thread_safe_queue.h
```diff
@@ -0,0 +1,274 @@
...
+};
+
+// The is a multi-producer/multi-consumer safe queue.
+//
+// Any number of threads may read or write safely.
+// empty() likely not high performing as it is locking
+// For non blocking access, use `try_pop()`
+// NOTE: Using `if !empty then pop`, is slower, and may still block with multiple consumers.
+template <typename T>
+class QueueMPMC : public QueueBase<T>
```
MarkY-LunarG, line 119: I hate short names, I would prefer "QueueMultiProcMultiCons" over "QueueMPMC".  I think in the code itself, it would be clearer than the short names.

MarkY-LunarG: Oops, I see these are only used inter-file.  Forget that.

framework/util/thread_safe_queue.h
```diff
@@ -0,0 +1,274 @@
...
+
+    constexpr static bool kSingleConsumer = SingleConsumer;
+
+    using Queue = std::
+        conditional_t<kSingleConsumer, thread_safe_queue_detail::QueueMPSC<T>, thread_safe_queue_detail::QueueMPMC<T>>;
+    using value_type = typename Queue::value_type;
+    using pop_type   = typename Queue::pop_type;
+
+    bool empty() { return queue_.empty(); }
+    bool live() const { return queue_.live(); }
```
MarkY-LunarG: Same suggestion here for "live" -> "is_live".

framework/decode/file_processor.h
jzulauf-lunarg: @antonio-lunarg -- yeah, the safe solution will be to segregate the state and operation for ProcessBlocks into a "BlockProcessor" object owned by the appropriate thread with suitable interfaces.  FileProcessor is less omnibus than it used to be, but still pretty large.

framework/util/thread_safe_queue.h
jzulauf-lunarg: It's an embedded "Architecture Decision Record" for the maintainer wondering why a highly normal accessor isn't present in the class.  Will tighten this up and label it.

framework/util/thread_safe_queue.h
jzulauf-lunarg: And nobody calls it, and it wouldn't be a good idea to anyway for the same reason `size()` isn't.  Deleting.

jzulauf-lunarg: I removed the unused snapshot-style accessors rather than renaming/documenting them, clarified the single-consumer contract, and kept only the operations used by the async queue path.

MarkY-LunarG (approved):
Looks good from my standpoint.  As soon as you and Antonio work out the last bits it should be good to go.

antonio-lunarg (approved):

PR #2786, opened 2026-03-16
Present image by debug name

### LINKED ISSUE #2705: allow display of non-swapchain images during replay
When replaying e.g., an OpenVR trace, the "swapchain" image is not actually the image being displayed by the device. There should be some mechanism (cmdline option using debugutils string matching?) to select which image(s) are displayed during replay.

This should also apply to the image dumping functionality for CI purposes.

fabian-lunarg:
this PR adds the option to extract images by providing a previously set debug-utils name.

this is accomplished by adding an option: `--present-override `

in case an image is found (via pattern-matching against ``):
- create a dedicated (adhoc) swapchain with matching size and format
- i.e. HDR-images will be presented on matching HDR-swapchains, if available
- per frame:
  - blit image into adhoc swapchain
  - present adhoc swapchain
- use image as source-image for the `--screenshots` option (instead of default swapchain-image)

the behavior of the `--screenshot-scale X` option is extended, while remaining backwards-compatible.
the provided scaling factor can be 2-dimensional and supplied values can be negative.

e.g. `--screenshot-scale 1, -1` could be used to flip an image vertically.

- `--screenshot-scale` can be combined with `--present-override ` and will affect both screenshots and presentation-windows.

- the `--screenshots` option can now output images with multiple layers:
  - writes one image per layer, appending "_layer_X" prefix
  - no change for single layer case

fixes #2705 

ideas for follow-up(s) include:
- generalize to multiple images (with their own windows)
- support presenting images with multiple layers
- just supply a `VkImage`'s handle-id instead of debug-utils name
- more sophisticated lookup/filtering
- in order to be more generous/flexible with formats, create custom pipeline instead of `vkCmdBlit`

### DIFF
--- a/framework/decode/vulkan_virtual_swapchain.cpp
+++ b/framework/decode/vulkan_virtual_swapchain.cpp
@@ -41,26 +41,29 @@ void VulkanVirtualSwapchain::CleanDeviceResources(VkDevice device, const graphic
     {
         const auto& ofb_data = it->second;
 
-        for (auto semaphore : ofb_data.acquire_semaphores)
+        for (const auto& [cmd_buf, fence, semaphore] : ofb_data.frame_data)
         {
-            device_table->DestroySemaphore(device, semaphore, nullptr);
+            if (cmd_buf != VK_NULL_HANDLE)
+            {
+                device_table->FreeCommandBuffers(device, ofb_data.command_pool, 1, &cmd_buf);
+            }
+
+            if (fence != VK_NULL_HANDLE)
+            {
+                device_table->DestroyFence(device, fence, nullptr);
+            }
+
+            if (cmd_buf != VK_NULL_HANDLE)
+            {
+                device_table->DestroySemaphore(device, semaphore, nullptr);
+            }
         }
 
         for (auto& img_data : ofb_data.image_data)
         {
-            if (img_data.command_buffer != VK_NULL_HANDLE)
-            {
-                device_table->FreeCommandBuffers(device, ofb_data.command_pool, 1, &img_data.command_buffer);
-            }
-
             if (img_data.semaphore != VK_NULL_HANDLE)
             {
                 device_table->DestroySemaphore(device, img_data.semaphore, nullptr);
-            }
-
-            if (img_data.fence != VK_NULL_HANDLE)
-            {
-                device_table->DestroyFence(device, img_data.fence, nullptr);
             }
         }
 
@@ -1395,19 +1398,20 @@ void VulkanVirtualSwapchain::PresentImageAdHoc(const VulkanDeviceInfo*          
             result = device_table->QueueWaitIdle(ofb_data.queue);
             GFXRECON_ASSERT(result == VK_SUCCESS);
 
-            for (VkSemaphore acquire_semaphore : ofb_data.acquire_semaphores)
+            for (const auto& [cmd_buf, fence, acquire_semaphore] : ofb_data.frame_data)
             {
+                device_table->FreeCommandBuffers(device, ofb_data.command_pool, 1, &cmd_buf);
+                device_table->DestroyFence(device, fence, nullptr);
                 device_table->DestroySemaphore(device, acquire_semaphore, nullptr);
             }
+
             for (auto& image_data : ofb_data.image_data)
             {
                 device_table->DestroySemaphore(device, image_data.semaphore, nullptr);
-                device_table->DestroyFence(device, image_data.fence, nullptr);
-                device_table->FreeCommandBuffers(device, ofb_data.command_pool, 1, &image_data.command_buffer);
             }
             device_table->DestroySwapchainKHR(device, ofb_data.swapchain, nullptr);
 
-            ofb_data.acquire_semaphores.clear();
+            ofb_data.frame_data.clear();
             ofb_data.image_data.clear();
         }
 
@@ -1422,7 +1426,7 @@ void VulkanVirtualSwapchain::PresentImageAdHoc(const VulkanDeviceInfo*          
         result = device_table->GetSwapchainImagesKHR(device, ofb_data.swapchain, &image_count, swapchain_images.data());
         GFXRECON_ASSERT((result == VK_SUCCESS) && (swapchain_images.size() == image_count));
 
-        ofb_data.acquire_semaphores.resize(image_count);
+        ofb_data.frame_data.resize(image_count);
         ofb_data.image_data.resize(image_count);
 
         VkSemaphoreCreateInfo semaphore_create_info;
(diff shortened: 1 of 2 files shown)


framework/graphics/vulkan_resources_util.cpp
```diff
@@ -2640,6 +2639,49 @@ bool VulkanResourcesUtil::IsScalingSupported(VkFormat          src_format,
...
+                                     blit_image_params.src_img,
+                                     blit_image_params.src_layout,
+                                     VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL,
+                                     blit_image_params.aspect);
+
+    // transition dst-layout
+    TransitionImageToTransferOptimal(command_buffer,
+                                     blit_image_params.dst_img,
+                                     blit_image_params.dst_layout,
+                                     VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL,
```
antonio-lunarg, line 2655: `TransitionImageToTransferOptimal` needs to be improved: if the layout is `TRANSFER_DST` then the `dstAccessMask` should be set to `VK_ACCESS_TRANSFER_WRITE_BIT`

antonio-lunarg: Same thing for `TransitionImageFromTransferOptimal`: when current layout is `DST_OPTIMAL` then `srcAccessMask` should be `TRANSFER_WRITE`.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7430,6 +7465,43 @@ VkResult VulkanReplayConsumerBase::OverrideSetDebugUtilsObjectNameEXT(
...
+        // reverse lookup of handle-id
+        if (auto img_id_it = image_handle_id_map_.find((VkImage)info->objectHandle);
+            img_id_it != image_handle_id_map_.end())
+        {
+            format::HandleId handle_id = img_id_it->second;
+
+            if (auto* img_info = object_info_table_->GetVkImageInfo(handle_id))
+            {
+                // keep track of debug-utils name
+                img_info->debug_utils_name = info->pObjectName;
```
antonio-lunarg, line 7483: What if `pObjectName` is `nullptr`?

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -1274,24 +1267,89 @@ void VulkanVirtualSwapchain::FrameBoundaryANDROID(PFN_vkFrameBoundaryANDROID
...
+        // query available surface formats
+        uint32_t format_count = 0;
+        instance_table->GetPhysicalDeviceSurfaceFormatsKHR(
+            device_info->parent, ofb_data.surface_info.handle, &format_count, nullptr);
+        std::vector<VkSurfaceFormatKHR> surface_formats(format_count);
+        instance_table->GetPhysicalDeviceSurfaceFormatsKHR(
+            device_info->parent, ofb_data.surface_info.handle, &format_count, surface_formats.data());
+        for (const auto& [format, colorspace] : surface_formats)
+        {
+            ofb_data.surface_formats.insert(format);
```
antonio-lunarg, line 1297: How about `colorspace`? Later I see `VK_COLOR_SPACE_SRGB_NONLINEAR_KHR` is hardcoded. Maybe we can use this `colorspace` value instead?

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -1382,31 +1444,33 @@ void VulkanVirtualSwapchain::FrameBoundaryANDROID(PFN_vkFrameBoundaryANDROID
...
             GFXRECON_ASSERT(result == VK_SUCCESS);
         }
     }
 
     // Acquire next image from the swapchain
-
     VkSemaphore& acquire_semaphore     = ofb_data.acquire_semaphores[ofb_data.acquire_index];
     uint32_t     swapchain_image_index = 0;
 
     result = device_table->AcquireNextImageKHR(
```
antonio-lunarg, line 1481: What if `result` here is suboptimal or out-of-date?

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: yes, that's possible according to [spec](https://docs.vulkan.org/refpages/latest/refpages/source/VkDebugUtilsObjectNameInfoEXT.html). we have to check for it

framework/decode/vulkan_virtual_swapchain.cpp
fabian-lunarg: I think I dodged colorspace mostly for convenience (was easier for the map lookup to only match against VkFormat). also I'm not entirely sure how relevant this would be. like it could either always be `VK_COLOR_SPACE_SRGB_NONLINEAR_KHR` or we're not even interested. would keep in mind though for a follow up

framework/graphics/vulkan_resources_util.cpp
```diff
@@ -2640,6 +2639,49 @@ bool VulkanResourcesUtil::IsScalingSupported(VkFormat          src_format,
     return scale == 1.0f || is_blit_supported;
 }
 
+void VulkanResourcesUtil::BlitImage(const blit_image_params_t& blit_image_params)
+{
+    constexpr uint32_t queue_family_index = 0;
+    constexpr uint32_t queue_index        = 0;
```
MarkY-LunarG: Assuming that queue family index 0 will work here is not safe.  It is possible this could be a queue that does not support BLIT operations (i.e. a pure compute queue).  This really needs to be passed in and/or use the queue associated with the image.

framework/decode/vulkan_virtual_swapchain.cpp
fabian-lunarg: currently not handled. not entirely sure how. I haven't hit the case in practice. certain WSI allow to manually resize the window (e.g. xcb), which we don't handle gracefully and basically crash (not even reaching this point). this is also the still the case for our regular window, to the best of my knowledge. I think there are 'options' we have to become more resilient and/or flexible with the windows we open. but for this PR, best I can do is prob. add another assertion.

framework/graphics/vulkan_resources_util.cpp
fabian-lunarg: thankfully not using this anymore (was used in earlier prototype with local submit + fence-sync).
since this is actually not used and I agree with the potential problems here,
best I just remove this flavor without the `VkCommandBuffer` parameter

framework/graphics/vulkan_resources_util.cpp
fabian-lunarg: yes, getting the image-barriers 100% correct as per vulkan-spec, those helpers aren't very helpful.
in practice prob. only the pipeline-stages and to some extent the image-layout are relevant.
validation-layers also wouldn't report anything here.
I changed the access-masks to `VK_ACCESS_MEMORY_READ_BIT | VK_ACCESS_MEMORY_WRITE_BIT`
and `VK_ACCESS_TRANSFER_READ_BIT | VK_ACCESS_TRANSFER_WRITE_BIT` respectively,
which is correct for our case.

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: done

framework/decode/vulkan_virtual_swapchain.cpp
fabian-lunarg: assertion added

framework/graphics/vulkan_resources_util.cpp
fabian-lunarg: done

framework/graphics/vulkan_resources_util.cpp
antonio-lunarg: I would have opted for something more granular like:
```
switch (dst_layout) {
 case DST_OPTIMAL: dst_access_mask = TRANSFER_WRITE;
 case SRC_OPTIMAL: dst_access_mask = TRANSFER_READ;
  ...
}
```
But your solution works anyway, and if there's no need for more granularity, I am happy with the change :)

fabian-lunarg:
thank you @antonio-lunarg , @MarkY-LunarG !

PR #2794, opened 2026-03-18
Add support for `VK_EXT_device_generated_commands`

marius-pelegrin-arm:
This commit chain contains 3 commits.
Even though the commit chain's goal is to support `Vk_EXT_device_generated_commands`, the first and second commit also serves to expose different unsupported case/issues with the codegen.

## First commit

**Fix and standardize Vulkan union decoding**

This commit originates from errors detected when testing capturing and replaying of the extension `VK_EXT_device_generated_commands`.

You can test what this commit fixes by trying to capture/replay the TraceToolTests `vulkan_device_generated_commands_1` and `vulkan_compute_device_generated`.

https://github.com/ARM-software/tracetooltests

While trying to fix these samples, I realized that the behavior for encoding Vulkan unions was pretty uniform while the behavior for decoding the same unions was quite chaotic. This commit aims at fixing this.

The parent structure field that is a union, and the fields of the union should all be treated exactly the same as if they were normal structure fields:
- If it is a pointer, using `StructPointerDecoder<Decoded_X>*`
- If not, using directly `Decoded_X*`

The `Decoded_X*` of the parent structure should NOT contain a specific `decoded_type` field, as the type doesn't require any special decoding or mapping. It should instead be read from the `decoded_value`.

Finally, the `decoded_value` should always be set.

This behavior is NOT a new behavior. It was already implemented for the decoding of `VkAccelerationStructureGeometryKHR` and `VkDescriptorGetInfoEXT` for instance (though with some little errors)

Ideally, all of this should be part of the generated code, but while it is not, I think we should still attempt to have a standardized data structure.

## Second commit

**Fix `vkCreateIndirectExecutionSetEXT` handle mapping**

`vkCreateIndirectExecutionSetEXT` takes as input a `VkIndirectExecutionSetCreateInfoEXT` structure, which itself contains a `VkIndirectExecutionSetInfoEXT` union whose members can have handles to map.

Currently, union handle mapping is not handled by generated code, and this call in particular needs handles mapped manually.

This is what this commit implements.

## Third commit

**Add support for `VK_EXT_device_generated_commands`**
    
Finalize `VK_EXT_device_generated_commands` by replacing device addresses contained in `VkGeneratedCommandsInfoEXT`

### DIFF
--- a/framework/decode/vulkan_address_replacer.cpp
+++ b/framework/decode/vulkan_address_replacer.cpp
@@ -1553,6 +1553,53 @@ void VulkanAddressReplacer::ProcessUpdateDescriptorSets(uint32_t              de
     }
 }
 
+void VulkanAddressReplacer::ProcessGeneratedCommandsInfoEXT(VkGeneratedCommandsInfoEXT* pGeneratedCommandsInfo,
+                                                            const decode::VulkanDeviceAddressTracker& address_tracker)
+{
+    GFXRECON_ASSERT(pGeneratedCommandsInfo != nullptr);
+
+    auto address_remap = [&address_tracker](VkDeviceAddress& capture_address) -> bool {
+        auto buffer_info = address_tracker.GetBufferByCaptureDeviceAddress(capture_address);
+
+        // skip over null-addresses
+        if (capture_address == 0)
+        {
+            return false;
+        }
+
+        if (buffer_info != nullptr && buffer_info->replay_address != 0)
+        {
+            if (buffer_info->capture_address != buffer_info->replay_address)
+            {
+                uint64_t offset = capture_address - buffer_info->capture_address;
+
+                // in-place address-remap via const-cast
+                capture_address = buffer_info->replay_address + offset;
+                return true;
+            }
+        }
+        return false;
+    };
+
+    if (!address_remap(pGeneratedCommandsInfo->indirectAddress))
+    {
+        GFXRECON_LOG_WARNING_ONCE(
+            "VulkanAddressReplacer::ProcessGeneratedCommandsInfoEXT: indirectAddress remap failed");
+    }
+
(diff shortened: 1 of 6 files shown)


framework/decode/vulkan_address_replacer_arm.h
```diff
@@ -0,0 +1,140 @@
...
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(decode)
+
+/**
+ * @brief   VulkanAddressReplacerARM can be used to check and potentially sanitize input-parameters for various cases.
+ *
+ * Important note: all internal Vulkan-API calls performed by this class are expected to be wrapped by calls to:
+ * - decode::BeginInjectedCommands() / decode::EndInjectedCommands()
+ */
+class VulkanAddressReplacerARM : public VulkanAddressReplacerBase
```
fabian-lunarg: is it intended (and for the scope of this PR required) to introduce `VulkanAddressReplacerARM`
and `VulkanAddressReplacerBase` here?

meaning, it feels this would require some additional consolidation,
potentially a larger refactoring to keep the code-base clean and maintainable.

marius-pelegrin-arm: Absolutely not. Just me failing to adapt the commit between our internal branch and upstream. I will fix this (and the CI pipeline fail)

fabian-lunarg:
> Ideally, all of this should be part of the generated code, but while it is not, I think we should still attempt to have a standardized data structure.

this is a good point, would be nice to generalize the solutions we have found so far.

fabian-lunarg (approved):
this looks good to me. 
starting with the handling of union-decode (we can/will draw inspiration here and get back to code-gen enventually)
added overloads for handle-mapping and integration with address-replacer all clean.

PR #2798, opened 2026-03-19
Fix duplicate 'OutputDir' in DumpResourcesOptions

vaage:
Removed duplicate 'OutputDir' entry in example dump resources config.


bradgrantham-lunarg:
Will merge because this is a documentation change, will not alter testing results

PR #2799, opened 2026-03-19
Reactivate bypassed code for Offscreen Frame Boundary handling

### LINKED ISSUE #1494: Handle use of AHB on Android without QueuePresent with FrameBoundary as frame

### LINKED ISSUE #2711: vkFrameBoundaryANDROID not handled correctly
we recently merged: https://github.com/LunarG/gfxreconstruct/pull/2623
but an extended CI-run uncovered remaining issues.

the introduced code-path was left intact, but is currently side-stepped with a `constexpr bool`
this was done here: https://github.com/LunarG/gfxreconstruct/pull/2661

this issue demands to properly handle `vkFrameBoundaryANDROID`, fix remaining edge-cases
and re-activate the side-stepped code-path.

fabian-lunarg:
this seems possible now, after the newly proposed code was reworked/fixed.
a previously problematic capture replays without issue now

fixes #2711 
fixes #1494

### DIFF
--- a/framework/decode/vulkan_virtual_swapchain.cpp
+++ b/framework/decode/vulkan_virtual_swapchain.cpp
@@ -1518,7 +1518,7 @@ void VulkanVirtualSwapchain::PresentImageAdHoc(const VulkanDeviceInfo*          
             // NOTE: this is avoiding an unmotivated segfault on android
             if (ofb_data.surface_info.window->GetWsiExtension() != VK_KHR_ANDROID_SURFACE_EXTENSION_NAME)
             {
-                VkImageMemoryBarrier memory_barrier;
+                VkImageMemoryBarrier memory_barrier{};
                 memory_barrier.sType                           = VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER;
                 memory_barrier.pNext                           = nullptr;
                 memory_barrier.srcAccessMask                   = VK_ACCESS_NONE;


framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -1515,32 +1515,36 @@ void VulkanVirtualSwapchain::PresentImageAdHoc(const VulkanDeviceInfo*
...
-                                              0,
-                                              nullptr,
-                                              0,
-                                              nullptr,
-                                              1,
-                                              &memory_barrier);
+            // NOTE: this is avoiding an unmotivated segfault on android
+            if (ofb_data.surface_info.window->GetWsiExtension() != VK_KHR_ANDROID_SURFACE_EXTENSION_NAME)
+            {
+                VkImageMemoryBarrier memory_barrier;
```
antonio-lunarg: I would add `= {}` here.

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -1515,32 +1515,36 @@ void VulkanVirtualSwapchain::PresentImageAdHoc(const VulkanDeviceInfo*
...
-                                              VK_PIPELINE_STAGE_NONE,
-                                              VK_PIPELINE_STAGE_ALL_COMMANDS_BIT,
-                                              0,
-                                              0,
-                                              nullptr,
-                                              0,
-                                              nullptr,
-                                              1,
-                                              &memory_barrier);
+            // NOTE: this is avoiding an unmotivated segfault on android
```
antonio-lunarg, line 1518: Any reference about this? Maybe an external link to the description of issue?

fabian-lunarg: the details were rather platform/vendor specific, so I was unable to leave anything useful in comment or PR.
only can say this would surface in our internal/extended CI on android.
the symptom is a segfault in `vkCmdPipelineBarrier` when transitioning a new swapchain-image to `VK_IMAGE_LAYOUT_PRESENT_SRC_KHR`. validation-layers don't catch anything (they do complain when this transition is missing, so it's required by spec). this leaves the possibility of a driver-bug

framework/decode/vulkan_virtual_swapchain.cpp
fabian-lunarg: done

fabian-lunarg:
before merging, I'm running an internal/extended CI, since that fell on our feet the last time around.

PR #2804, opened 2026-03-20
Include missing header

panos-lunarg:


bradgrantham-lunarg:
> Does fix the build on Ubuntu 22.04 :-)

Ubuntu 22.04 with gcc 11?

csteinberg-samsung:
I tested it on gcc 11.4.0 on ubunutu 22.04.3 LTS and it builds successfully now.

PR #2812, opened 2026-03-24
Update VulkanMemoryAllocator: 3.1 -> 3.3

### LINKED ISSUE #2809: Update VulkanMemoryAllocator
To get `-m rebind` Vulkan 1.4 support, we need to update VMA, but some public members have become private so someone will need to investigate that and design a solution.

fabian-lunarg:
- affects '-m rebind' and bumps Vulkan-support to 1.4
- required changes to mutex-synchronization, VMA made their internal mutexes private
- fix a pass-by-value bug
- register a custom VmaDeviceMemoryCallbacks, allowing to free our external mutexes

fixes #2809

### DIFF
--- a/framework/decode/vulkan_rebind_allocator.cpp
+++ b/framework/decode/vulkan_rebind_allocator.cpp
@@ -74,6 +74,19 @@ GFXRECON_BEGIN_NAMESPACE(decode)
 constexpr format::HandleId kPlaceholderHandleId = static_cast<format::HandleId>(~0ul);
 constexpr uintptr_t        kPlaceholderAddress  = static_cast<uintptr_t>(~0ul);
 
+// Called by VMA before vkFreeMemory on each internally managed VkDeviceMemory block.
+// Removes the corresponding external mutex so block_mutexes_ doesn't grow unbounded.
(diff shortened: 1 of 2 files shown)


framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -74,11 +74,36 @@ GFXRECON_BEGIN_NAMESPACE(decode)
...
 {}
 
+std::mutex& VulkanRebindAllocator::GetOrCreateBlockMutex(VkDeviceMemory device_memory)
+{
+    std::lock_guard guard(block_mutexes_guard_);
+    auto&           mutex_ptr = block_mutexes_[device_memory];
+
+    if (!mutex_ptr)
+    {
+        mutex_ptr = std::make_unique<std::mutex>();
```
MarkY-LunarG, line 102: Should we associate this new pointer with `block_mutexes_[device_memory]`?

fabian-lunarg: already the case :)

`auto&           mutex_ptr = block_mutexes_[device_memory];`

PR #2814, opened 2026-03-24
Fix/Enhance spirv parsing

fabian-lunarg:
- Fix: inlined spirv was not parsed due to oversight
- improved detection of pointer-chains
- remove a warning that was too conservative
- Fix an indexing bug in populate_shader_stages

slang-style arrays of pointers are correctly identified now:

(slang) e.g. :

```
struct vertex_t{...};

struct params_t
{
    vertex_t **vertex_buffers; // array of pointers
    uint32_t **index_buffers;
    ....
}
```

### DIFF
--- a/framework/graphics/vulkan_check_buffer_references.cpp
+++ b/framework/graphics/vulkan_check_buffer_references.cpp
@@ -66,8 +66,8 @@ void populate_shader_stages(const decode::StructPointerDecoder<T>*    pCreateInf
                 else
                 {
                     // check potentially inlined spirv
-                    if (auto module_create_info =
-                            vulkan_struct_get_pnext<VkShaderModuleCreateInfo>(pCreateInfos->GetPointer()->pStages + s))
+                    if (auto module_create_info = vulkan_struct_get_pnext<VkShaderModuleCreateInfo>(
+                            pCreateInfos->GetPointer()[i].pStages + s))
                     {
                         graphics::vulkan_check_buffer_references(
                             module_create_info->pCode, module_create_info->codeSize, pipeline_info);
@@ -121,7 +121,7 @@ void populate_shader_stages(
         {
             // check potentially inlined spirv
             auto module_create_info =
-                vulkan_struct_get_pnext<VkShaderModuleCreateInfo>(&pCreateInfos->GetPointer()->stage);
+                vulkan_struct_get_pnext<VkShaderModuleCreateInfo>(&pCreateInfos->GetPointer()[i].stage);
             if (module_create_info != nullptr)
             {
                 graphics::vulkan_check_buffer_references(


framework/util/spirv_parsing_util.cpp
```diff
@@ -611,6 +630,31 @@ bool SpirVParsingUtil::ParseBufferReferences(const uint32_t* const spirv_code, s
...
+        uint32_t pointer_array_stride = 0;
+        {
+            const Instruction* pointee_type_insn = FindDef(type_pointer_insn->operand(1));
+            if (pointee_type_insn && pointee_type_insn->opcode() == spv::OpTypePointer &&
+                pointee_type_insn->operand(0) == spv::StorageClassPhysicalStorageBuffer)
+            {
+                for (const Instruction* decor : decorations_instructions_)
+                {
+                    if (decor->operand(0) == type_pointer_insn->resultId() &&
+                        decor->operand(1) == spv::DecorationArrayStride)
```
antonio-lunarg, line 645: I am not sure a pointer has a stride. Would it make sense to look for an array type whose element type is the pointee, and get the stride from there?

framework/graphics/vulkan_check_buffer_references.cpp
```diff
@@ -62,7 +62,9 @@ void populate_shader_stages(const decode::StructPointerDecoder<T>*    pCreateInf
                         pipeline_info->buffer_reference_infos.insert(module_info->buffer_reference_infos.begin(),
                                                                      module_info->buffer_reference_infos.end());
                     }
-
+                }
+                else
+                {
                     // check potentially inlined spirv
                     if (auto module_create_info =
```
antonio-lunarg: Unrelated to this patch but what if there are multiple create info structures in the `pCreateInfos`?

framework/util/spirv_parsing_util.cpp
fabian-lunarg: the problem is the pointer-to-pointer pattern in slang does not compile to an array-type in spirv.
instead I identified an `OpTypePointer` with `ArrayStride=8` (sizeof(VkDeviceSize aka uint64_t)) decoration. I wasn't able to detect this case using only spirv-reflect. so while I wish this code wasn't required (and might become obsolete when this would be possible using spirv-reflect), it seemed the correct (and only) way

framework/graphics/vulkan_check_buffer_references.cpp
fabian-lunarg: nice catch, big fat bug. it should use `pCreateInfos->GetPointer()[i].pStages + s`

fixing this

fabian-lunarg: done

PR #2816, opened 2026-03-24
Automatically rename player process name at runtime to captured application name

jravi-amd:
### What does this PR do? 
This PR effectively replaces the functionality in `scripts\gfxrecon-replay-renamed.py`. Currently supported only on Windows.

### How does this work? 

On Windows to get the current process name, apps call [GetModuleFilename](https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulefilenamew) with `NULL` `hModule`. This returns the current process name. Graphics drivers often rely on filename returned by this function to apply application-specific optimizations and settings.

On replay during initialization, Microsoft Detours is used to hook `GetModuleFilename` with our version. So when a graphics driver makes a call to GetModuleFilename(A/W), it gets redirected to our hooked version. 

When the replay consumer encounters a process exe name packet in the metadata, we set the original app name in process name override file via a callback from the consumer. The hooked version now returns the original app's exe name instead of `gfxrecon-replay.exe`. 

From the driver's perspective, it receives the original application's process name instead of the replay tool's name. This ensures that application-specific driver optimizations and profiles are correctly applied during replay.

### Why do this? 
Convenience. Ease of use. Invoking the player as usual, `gfxrecon-replay.exe <trace>`, should handle process name overrides.

### DIFF
--- a/framework/util/process_name_override.cpp
+++ b/framework/util/process_name_override.cpp
@@ -1,0 +1,144 @@
+/*
+** Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and associated documentation files (the "Software"),
+** to deal in the Software without restriction, including without limitation
+** the rights to use, copy, modify, merge, publish, distribute, sublicense,
+** and/or sell copies of the Software, and to permit persons to whom the
+** Software is furnished to do so, subject to the following conditions:
+**
+** The above copyright notice and this permission notice shall be included in
+** all copies or substantial portions of the Software.
+**
+** THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+** IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+** FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+** AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+** FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+** DEALINGS IN THE SOFTWARE.
+*/
+
+#include <windows.h> 
+#include <string>
+#include <filesystem>
+#include "util/to_string.h"
+#include "util/logging.h"
+#include "util/interception/hooking_detours.h"
+#include "util/process_name_override.h"
+
+namespace 
+{
+// App Name set by callback from replay consumer
+std::string s_app_name{}; 
+
+// Helper function to construct the modified application path based on the original application path and the new application name.
+std::string GetModifiedApplicationPath(DWORD (*real_getmodule_filename_a)(HMODULE, LPSTR, DWORD), std::string app_name)
+{
+    if(app_name.empty() || real_getmodule_filename_a == nullptr)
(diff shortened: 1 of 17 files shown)


tools/optimize/dx12_optimize_util.cpp
```diff
@@ -31,6 +31,11 @@
 #include "decode/dx12_resource_value_tracker.h"
 #include "decode/file_processor.h"
 
+// Process name override
+#if _WIN32
```
davidd-lunarg: Instead of `#if _WIN32`, use `#if defined(WIN32)` to be consistent with other preprocessor checks in GFXR.

framework/util/process_name_override.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+
+// Based on MSDN documentation for GetModuleFileNameA/W, if hModule is null, the function returns the path of the executable of the current process. 
+// By hooking this function, we can override the process name that is reported to the driver. 
+// https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulefilenamew
+DWORD HookGetModuleFilenameA(HMODULE hModule, LPSTR lpFilename, DWORD nSize)
+{
+    if (hModule == nullptr && !s_app_name.empty() && nSize > 0)
+    {
+        GFXRECON_LOG_INFO("Overriding process name to: %s", s_app_name.c_str());
```
davidd-lunarg: In my test this message (or the similar message below) was logged ~20 times in a replay. It would be good to log it just once as long as `s_app_name` is the same for all log messages. If `s_app_name` won't change after the first log message, `GFXRECON_LOG_INFO_ONCE` can be used. If `s_app_name` might change then a different mechanism may be needed.

tools/replay/replay_settings.h
```diff
@@ -192,12 +192,13 @@ static void PrintUsage(const char* exe_name)
     GFXRECON_WRITE_CONSOLE(
         "          \t\tThis can include vertex, index, const buffer, shader resource, render target,");
     GFXRECON_WRITE_CONSOLE("          \t\tand depth stencil resources. Resources are dumped after the drawcall.");
-    GFXRECON_WRITE_CONSOLE(
-        "  --fwo <x,y>\t\tForce windowed mode if not already, and allow setting of a custom window location.");
-    GFXRECON_WRITE_CONSOLE("          \t\t(Same as --force-windowed-origin)");
+    GFXRECON_WRITE_CONSOLE("  --fwo <x,y>                     \tForce windowed mode if not already, and allow setting of a custom window location.");
+    GFXRECON_WRITE_CONSOLE("                                  \t(Same as --force-windowed-origin)");
+    GFXRECON_WRITE_CONSOLE("  --disable-process-name-override \tDisable the process name override functionality that sets the process name");
+    GFXRECON_WRITE_CONSOLE("                                  \tto the capture application name. This is enabled by default.");
```
davidd-lunarg: I think remove or modify the statement `This is enabled by default.` Otherwise it sounds to me like `--disable-process-name-override` is enabled by default.

davidd-lunarg (changes requested):
The documentation about renaming scripts should be updated to mention the new default behavior for Windows. https://github.com/LunarG/gfxreconstruct/blob/dev/USAGE_desktop_D3D12.md#renaming-scripts

That section should also be copied to `USAGE_desktop_Vulkan.md`. The new replay flag `--disable-process-name-override` should be added to the usage documents as well.

tools/optimize/dx12_optimize_util.cpp
jravi-amd: Fixed.

tools/replay/replay_settings.h
jravi-amd: Fixed

framework/util/process_name_override.cpp
jravi-amd: Fixed.

PR #2817, opened 2026-03-24
Fix capture time crash when tracking buffer device addresses

jravi-amd:
## Problem: 

On some applications, when trim capture is enabled, the application can crash.

## Fix

In the state tracker we track buffers by device addresses. Whenever the app calls `vkGetBufferDeviceAddress`, we associate the device address to the `VkBuffer` handle. On `vkDestroyBuffer`, we remove the buffer the `VkBuffer` from the tracker but this poses a problem. `Vkbuffer`s can potentially share the same device address if they are bound to the same `VkDeviceMemory` region. After the map is updated to the new buffer with the same address, the old buffer is destroyed by the app. But when removing the buffer from the map, we don't check if the device address is still associated to the same buffer handle. So we inadvertently remove the new buffer mapping on `vkDestroyBuffer` for the old buffer. This causes tracking to crash somewhere downstream because state tracker tries to retrieve a buffer handle for the device address and it is not available anymore. The fix requires checking existing buffer handle in the map before removing the mapping.

### DIFF
--- a/android/framework/decode/CMakeLists.txt
+++ b/android/framework/decode/CMakeLists.txt
@@ -108,6 +108,8 @@ target_sources(gfxrecon_decode
                    ${GFXRECON_SOURCE_DIR}/framework/decode/vulkan_device_address_tracker.cpp
                    ${GFXRECON_SOURCE_DIR}/framework/decode/vulkan_captured_swapchain.h
                    ${GFXRECON_SOURCE_DIR}/framework/decode/vulkan_captured_swapchain.cpp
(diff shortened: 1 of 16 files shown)


fabian-lunarg:
this def. seems to improve our handling of `VkBuffer`<->`VkDeviceAddress`, but the underlying problem remains:
we still only model a 1:1 relation for buffer<->address while there could be arbitrary many aliases, so 1:N.

so when we merge this we should in any case follow-up with a PR that corrects our wrong assumption,
i.e. change the lookup `GetBufferByDeviceAddress` to return a set of handles.

jravi-amd:
@fabian-lunarg thanks for the feedback. I can look into a follow-up fix. Thanks for approving this PR.

PR #2827, opened 2026-03-27
replay: Add --serialize-queue-submissions option

antonio-lunarg:
Add a new replay option that serializes queue submissions by inserting semaphores so each submit waits for the previous, preventing concurrent execution of multiple submissions during replay.

Implement serialization in `VulkanSubmitJobExecutor` with `Serialize` overloads for `VkSubmitInfo` and `VkSubmitInfo2`, and call `Serialize` from `OverrideQueueSubmit`/`OverrideQueueSubmit2` when the option is set.

Wire the option through CLI parsers, replay_settings, and help text, and update desktop/android usage docs. Refactor submit executor storage to map submit pointers to semaphore storage to avoid multiple injections.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -4342,12 +4342,12 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit(PFN_vkQueueSubmit        
         }
     }
 
-    VulkanSubmitJobExecutor executor;
-    executor.InjectBefore(std::move(plan), pSubmits->GetSpan());
+    VulkanSubmitJobExecution execution = GetDeviceSubmitJobExecutor(device_info).CreateExecution();
+    execution.InjectBefore(std::move(plan), pSubmits->GetSpan());
 
     if (options_.serialize_queue_submissions)
     {
-        executor.SerializeExecution(pSubmits->GetSpan());
+        execution.SerializeExecution(pSubmits->GetSpan());
     }
 
     // Only attempt to filter imported semaphores if we know at least one has been imported.
@@ -4582,12 +4582,12 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit2(PFN_vkQueueSubmit2      
         }
     }
 
-    VulkanSubmitJobExecutor executor;
-    executor.InjectBefore(std::move(plan), pSubmits->GetSpan());
+    VulkanSubmitJobExecution execution = GetDeviceSubmitJobExecutor(device_info).CreateExecution();
+    execution.InjectBefore(std::move(plan), pSubmits->GetSpan());
 
     if (options_.serialize_queue_submissions)
     {
-        executor.SerializeExecution(pSubmits->GetSpan());
+        execution.SerializeExecution(pSubmits->GetSpan());
     }
 
     // Only attempt to filter imported semaphores if we know at least one has been imported.
@@ -11527,6 +11527,19 @@ VulkanFrameWarmUp& VulkanReplayConsumerBase::GetDeviceFrameWarmUp(const VulkanDe
                                                                               *object_info_table_,
                                                                               options_.frame_warm_up_spirv_path,
                                                                               options_.frame_warm_up_load) });
+    GFXRECON_ASSERT(success);
+    return new_it->second;
+}
+
+VulkanSubmitJobExecutor& VulkanReplayConsumerBase::GetDeviceSubmitJobExecutor(const VulkanDeviceInfo* device_info)
+{
+    if (auto it = device_submit_job_executors_.find(device_info); it != device_submit_job_executors_.end())
+    {
+        return it->second;
+    }
+
+    auto [new_it, success] = device_submit_job_executors_.insert(
+        { device_info, VulkanSubmitJobExecutor(device_info, GetDeviceTable(device_info->handle)) });
     GFXRECON_ASSERT(success);
     return new_it->second;
 }
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -4309,8 +4309,7 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit(PFN_vkQueueSubmit        
         GetDeviceTable(device_info->handle)->DeviceWaitIdle(device_info->handle);
     }
 
-    VulkanSubmitJobPlan     plan;
-    VulkanSubmitJobExecutor executor;
+    VulkanSubmitJobPlan plan;
 
     if (options_.frame_warm_up_load != 0 && !fps_info_->IsFirstSubmitDone())
     {
@@ -4343,7 +4342,13 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit(PFN_vkQueueSubmit        
         }
     }
 
+    VulkanSubmitJobExecutor executor;
     executor.InjectBefore(std::move(plan), pSubmits->GetSpan());
+
+    if (options_.serialize_queue_submissions)
+    {
+        executor.SerializeExecution(pSubmits->GetSpan());
+    }
 
     // Only attempt to filter imported semaphores if we know at least one has been imported.
     // If rendering is restricted to a specific surface, shadow semaphore and forward progress state will need to be
@@ -4544,8 +4549,7 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit2(PFN_vkQueueSubmit2      
         GetDeviceTable(device_info->handle)->DeviceWaitIdle(device_info->handle);
     }
 
-    VulkanSubmitJobPlan     plan;
-    VulkanSubmitJobExecutor executor;
+    VulkanSubmitJobPlan plan;
 
     if (options_.frame_warm_up_load != 0 && !fps_info_->IsFirstSubmitDone())
     {
@@ -4578,7 +4582,13 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit2(PFN_vkQueueSubmit2      
         }
     }
 
+    VulkanSubmitJobExecutor executor;
     executor.InjectBefore(std::move(plan), pSubmits->GetSpan());
+
+    if (options_.serialize_queue_submissions)
+    {
+        executor.SerializeExecution(pSubmits->GetSpan());
+    }
 
     // Only attempt to filter imported semaphores if we know at least one has been imported.
     // If rendering is restricted to a specific surface, shadow semaphore and forward progress state will need to be
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -302,7 +302,7 @@ VulkanReplayConsumerBase::~VulkanReplayConsumerBase()
     WaitDevicesIdle();
 
     // free replacer internal vulkan-resources
-    _device_address_replacers.clear();
+    device_address_replacers_.clear();
 
     // process queued async tasks
     background_queue_.join_all();
@@ -2132,7 +2132,7 @@ void VulkanReplayConsumerBase::InitializeReplayDumpResources()
     if (resource_dumper_ == nullptr)
     {
         resource_dumper_ = std::make_unique<VulkanReplayDumpResources>(
-            options_, object_info_table_, _device_address_trackers, instance_tables_, device_tables_);
+            options_, object_info_table_, device_address_trackers_, instance_tables_, device_tables_);
         GFXRECON_ASSERT(resource_dumper_);
     }
 }
@@ -3712,7 +3712,7 @@ void VulkanReplayConsumerBase::OverrideDestroyDevice(
         }
 
         // free replacer internal vulkan-resources for the device
-        _device_address_replacers.erase(device_info);
+        device_address_replacers_.erase(device_info);
 
         // free potential swapchain-resources for the device
         GFXRECON_ASSERT(swapchain_)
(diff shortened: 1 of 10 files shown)


framework/decode/vulkan_submit_job.cpp
```diff
@@ -159,12 +161,86 @@ void VulkanSubmitJobExecutor::InjectBefore(VulkanSubmitJobPlan plan, std::span<V
...
+}
+
+void VulkanSubmitJobExecutor::Serialize(std::span<VkSubmitInfo> submit_infos)
+{
+    // Each submit info should wait on semaphores signaled by the previous submit.
+    std::span<const VkSemaphore> previous_submit_signal_semaphores = {};
+
+    for (VkSubmitInfo& submit_info : submit_infos)
+    {
+        if (!previous_submit_signal_semaphores.empty())
```
antonio-lunarg: Ouch. This is still incomplete. If submit `i` has no signal semaphores, then submit `i+1` has no dependency object to wait on, so serialization stops there. The current code only propagates existing signals forward. Need fix.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -11311,11 +11321,11 @@ void VulkanReplayConsumerBase::UpdateDescriptorSetInfoWithTemplate(
 VulkanDeviceAddressTracker&
 VulkanReplayConsumerBase::GetDeviceAddressTracker(const decode::VulkanDeviceInfo* device_info)
 {
-    auto it = _device_address_trackers.find(device_info);
-    if (it == _device_address_trackers.end())
+    auto it = device_address_trackers_.find(device_info);
+    if (it == device_address_trackers_.end())
```
fabian-lunarg: nit: with c++20 we can combine variable-inits with if-statements like:

`if(auto it = device_address_trackers_.find(device_info); it == device_address_trackers_.end()){...}`

the scope of `it` is limited to the if-branch

framework/decode/vulkan_submit_job.h
```diff
@@ -127,17 +134,61 @@ class VulkanSubmitJobExecutor
...
+
+    /**
+     * @brief  Serialize submit entries within one queue-submit call.
+     *
+     * Make each submit wait on semaphores signaled by the previous submit in the same
+     * `vkQueueSubmit()` call.
+     *
+     * @param  submit_infos     submit array to mutate in place.
+     */
+    void Serialize(std::span<VkSubmitInfo> submit_infos);
```
fabian-lunarg: `Serialize` is 100% correct terms of naming this function. but since the term/meaning is overloaded with other concepts, maybe it's easier to understand with a more descriptive name. e.g. `SerializeExecution` - otherwise folks might skip reading the doc and not understand that `Serialize` is actually a mutating call.

framework/decode/vulkan_submit_job.h
```diff
@@ -127,17 +134,61 @@ class VulkanSubmitJobExecutor
...
+     *
+     * @param  submit_infos2    submit2 array to mutate in place.
+     */
+    void Serialize(std::span<VkSubmitInfo2> submit_infos2);
+
   private:
-    // Original waits stripped from each submit before injection.
-    std::vector<VulkanSubmitSemaphores> original_wait_semaphores_;
+    /// Original waits stripped from each submit before injection, keyed by either `VkSubmitInfo*` or `VkSubmitInfo2*`.
+    std::unordered_map<void*, VulkanSubmitSemaphores> original_wait_semaphores_;
```
fabian-lunarg, line 345: the `void*` as key raised the question about lifetime of pointees:

- this looks like it's pointing into DecodeAllocator memory, valid only for this block before the allocator reuses it.
-  `VulkanSubmitJobExecutor` only ever lives on the stack, so the validity of acquired VkSubmitInfo(2) pointers seems guaranteed here.

so, my verdict is this should be safe with the current usage pattern. 
it 'would' be problematic when `VulkanSubmitJobExecutor` had a persistent lifetime.

roughly correct? does this line up with what you have in mind?

fabian-lunarg: yes, you even mention it in the doc:
"It (VulkanSubmitJobExecutor) must outlive the queue submit that consumes those structs."

framework/decode/vulkan_submit_job.cpp
fabian-lunarg: the fix would be to create (+own somewhere) additional glue-semaphores?

antonio-lunarg: Yes -> https://github.com/LunarG/gfxreconstruct/pull/2827/commits/3c428af1a39cf763e19cb0fecdb1335d9d054598

framework/decode/vulkan_submit_job.h
antonio-lunarg: Reworked a bit: now the "execution" use used within the scope of the queue-submit, and the "executor" is a per-device object which keeps track of semaphores created on the fly to force synchronisation.

framework/decode/vulkan_submit_job.cpp
```diff
@@ -159,12 +162,152 @@ void VulkanSubmitJobExecutor::InjectBefore(VulkanSubmitJobPlan plan, std::span<V
...
+                                   previous_submit_signal_semaphores.begin(),
+                                   previous_submit_signal_semaphores.end());
+
+            // Update submit info with the new wait semaphores pointer and count.
+            submit_info.waitSemaphoreCount = static_cast<uint32_t>(wait_semaphores.size());
+            submit_info.pWaitSemaphores    = wait_semaphores.data();
+
+            // Ensure wait dst stage mask points to valid storage.
+            static VkPipelineStageFlags wait_stage_mask = VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT;
+            submit_info.pWaitDstStageMask               = &wait_stage_mask;
```
fabian-lunarg: `submit_info.pWaitDstStageMask` should be an array of size `wait_semaphores.size()`.

["pWaitDstStageMask is a pointer to an array of pipeline stages at which each corresponding semaphore wait will occur."](https://docs.vulkan.org/refpages/latest/refpages/source/VkSubmitInfo.html)

framework/decode/vulkan_submit_job.cpp
fabian-lunarg: pre-existing, but `submit_info.pWaitDstStageMask` should be an array (same as comment below)

framework/decode/vulkan_submit_job.cpp
```diff
@@ -113,23 +113,25 @@ void VulkanSubmitJobExecutor::InjectBefore(VulkanSubmitJobPlan plan, std::span<V
...
 
     // Gather original wait-semaphores for each submit and prepare storage for injected wait-semaphores.
     auto& submit_jobs = plan.GetSubmitJobs();
     for (uint32_t submit_index = 0; submit_index < submit_infos.size(); ++submit_index)
     {
-        original_wait_semaphores_.emplace_back();
-        injected_wait_semaphore_infos_.emplace_back();
+        VkSubmitInfo2& submit_info                   = submit_infos[submit_index];
+        auto&          original_wait_semaphores      = original_wait_semaphores_[&submit_info].semaphores;
+        auto&          injected_wait_semaphore_infos = injected_wait_semaphore_infos_[&submit_info];
```
fabian-lunarg: `auto&          injected_wait_semaphore_infos = injected_wait_semaphore_infos_[&submit_info];` is unused in this scope and performs map-lookups and implicit creation of elements. on line 148 the statement is repeated, so it looks you can drop it here

framework/decode/vulkan_submit_job.h
```diff
@@ -127,17 +140,135 @@ class VulkanSubmitJobExecutor
...
+     *
+     * @return VkSemaphore The created semaphore or `VK_NULL_HANDLE` on error.
+     */
+    VkSemaphore CreateSemaphore();
+
+  private:
+    const VulkanDeviceInfo*            device_info_  = nullptr;
+    const graphics::VulkanDeviceTable* device_table_ = nullptr;
+
+    std::vector<VkSemaphore> semaphores_;
```
fabian-lunarg: I think `std::vector<VkSemaphore> semaphores_;` is growing unbounded.
in addition to memory there will be driver-limits, so this needs attention.

every `CreateSemaphore()` call adds to it, but only the dtor clears them. 
so they live for the entire replay (because executors are owned by ReplayConsumer) and accumulate.
they should be either cleared or re-used (on queue-submits and wait-idle)

antonio-lunarg: Thank you for pointing that out. Fixing it was quite a ride. I initially wanted to destroy the binary semaphores once they were done, but Vulkan does not provide an API to check whether a binary semaphore has been signaled.

Timeline semaphores do provide that capability, so I switched the injected semaphores to timeline semaphores instead. However, timeline semaphores require special handling when used with VkSubmitInfo, so I had to iterate on the implementation a few times before getting it right. I hope the code is correct now.

framework/decode/vulkan_submit_job.cpp
```diff
@@ -159,12 +161,330 @@ void VulkanSubmitJobExecutor::InjectBefore(VulkanSubmitJobPlan plan, std::span<V
...
+
+    for (VkSubmitInfo2& submit_info : submit_infos2)
+    {
+        // Ensure this submit signals at least one semaphore so the next submit can wait on it.
+        if (submit_info.signalSemaphoreInfoCount == 0 || submit_info.pSignalSemaphoreInfos == nullptr)
+        {
+            VkSemaphoreSubmitInfo dummy_semaphore_info = { VK_STRUCTURE_TYPE_SEMAPHORE_SUBMIT_INFO };
+            dummy_semaphore_info.semaphore             = executor_.CreateTimelineSemaphore();
+            dummy_semaphore_info.value                 = 1;
+            dummy_semaphore_info.stageMask             = VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT;
```
fabian-lunarg: I think this should be `VK_PIPELINE_STAGE_2_ALL_COMMANDS_BIT` here, otherwise the signal will be issued right when the submit is added to the queue. i.e. there would be no serialization in this case

framework/decode/vulkan_submit_job.cpp
```diff
@@ -159,12 +161,330 @@ void VulkanSubmitJobExecutor::InjectBefore(VulkanSubmitJobPlan plan, std::span<V
...
+    return injected_wait_semaphores_[&submit_info];
+}
+
+std::vector<VkPipelineStageFlags>& VulkanSubmitJobExecution::GetWaitDstStageMasks(VkSubmitInfo& submit_info,
+                                                                                  bool          override)
+{
+    if (!injected_wait_dst_stage_masks_.contains(&submit_info))
+    {
+        injected_wait_dst_stage_masks_[&submit_info] = std::vector<VkPipelineStageFlags>(
+            submit_info.pWaitDstStageMask, submit_info.pWaitDstStageMask + submit_info.waitSemaphoreCount);
```
fabian-lunarg: when there were no (original) wait-semaphores `submit_info.pWaitDstStageMask` will be `nullptr`,
but coming from InjectBefore(), `submit_info.waitSemaphoreCount` was already increased, so we get UB with `nullptr + 1`.

I think this branch should be avoided when `override == true`:

if(override){...}
else if(!injected_wait_dst_stage_masks_.contains(&submit_info)){ ... }

framework/decode/vulkan_submit_job.cpp
```diff
@@ -159,12 +161,330 @@ void VulkanSubmitJobExecutor::InjectBefore(VulkanSubmitJobPlan plan, std::span<V
...
+{
+    // Iterate through tracked timeline semaphores and remove those that have reached or exceeded their signal value.
+    auto it = timeline_semaphores_.begin();
+    while (it != timeline_semaphores_.end())
+    {
+        VkSemaphore         semaphore = *it;
+        VkSemaphoreWaitInfo wait_info{ VK_STRUCTURE_TYPE_SEMAPHORE_WAIT_INFO };
+        wait_info.semaphoreCount = 1;
+        wait_info.pSemaphores    = &semaphore;
+        static uint64_t value    = 1;
```
fabian-lunarg: no need for this to be static?

framework/decode/vulkan_submit_job.h
fabian-lunarg: nice, `PruneSignaledTimelineSemaphores` will do the trick!

framework/decode/vulkan_submit_job.cpp
antonio-lunarg: Fixed

framework/decode/vulkan_submit_job.cpp
antonio-lunarg: Fixed

framework/decode/vulkan_submit_job.cpp
antonio-lunarg: I added a check for nullptr. I still want to handle the override request later, to make sure values are reset even if the stage masks have already been stored.

fabian-lunarg (approved):
happy at this point :)

antonio-lunarg:
> @fabian-lunarg  happy at this point :)

Sorry, need another round of review after figuring out we could use one single timeline semaphore for serialization.

fabian-lunarg:
> > @fabian-lunarg  happy at this point :)
> 
> Sorry, need another round of review after figuring out we could use one single timeline semaphore for serialization.

that's worth it :)

framework/decode/vulkan_submit_job.h
```diff
@@ -95,22 +98,71 @@ class VulkanSubmitJobPlan
...
+ * executor can safely destroy the semaphore.
  */
-class VulkanSubmitJobExecutor
+struct VulkanInjectedSemaphore
+{
+    VkSemaphore handle       = VK_NULL_HANDLE;
+    uint64_t    target_value = 0;
+
+    const VulkanDeviceInfo*            device_info_;
+    const graphics::VulkanDeviceTable* device_table_;
```
fabian-lunarg, line 125: NIT, struct member-naming not consistent wrt. suffix "_"

framework/decode/vulkan_submit_job.cpp
```diff
@@ -54,23 +55,86 @@ bool VulkanSubmitJobPlan::HasJobsForIndex(uint32_t submit_index) const
...
+    if (handle == VK_NULL_HANDLE)
+    {
+        return false;
+    }
+
+    VkSemaphoreWaitInfo wait_info = { VK_STRUCTURE_TYPE_SEMAPHORE_WAIT_INFO };
+    wait_info.semaphoreCount      = 1;
+    wait_info.pSemaphores         = &handle;
+    wait_info.pValues             = &target_value;
+    VkResult result               = device_table_->WaitSemaphores(device_info_->handle, &wait_info, 0);
```
fabian-lunarg: why not use [vkGetSemaphoreCounterValue](https://docs.vulkan.org/refpages/latest/refpages/source/vkGetSemaphoreCounterValue.html) here?

fabian-lunarg: or does this rely on waiting as side-effect? (wouldn't expect by its name)

framework/decode/vulkan_submit_job.h
```diff
@@ -95,22 +98,71 @@ class VulkanSubmitJobPlan
...
+
+    bool HasReachedTargetValue() const;
+
+    VulkanInjectedSemaphore(const VulkanDeviceInfo* device_info, const graphics::VulkanDeviceTable* table);
+    ~VulkanInjectedSemaphore();
```
fabian-lunarg, line 134: when you got a dtor, you will (sooner or later) also need the rest of the rule-of-5 group.
copy constructor, move constructor, (move)-assignment, ideally with copy+swap.

fabian-lunarg: meaning, copies should be prevented and moving implemented

framework/decode/vulkan_submit_job.cpp
```diff
@@ -158,13 +241,386 @@ void VulkanSubmitJobExecutor::InjectBefore(VulkanSubmitJobPlan plan, std::span<V
...
+        // Override wait semaphores pointer.
+        submit_info.pWaitSemaphores = injected_wait_semaphores_[&submit_info].data();
+    }
+    return injected_wait_semaphores_[&submit_info];
+}
+
+void VulkanSubmitJobExecution::AddWaitSemaphore(VkSubmitInfo& submit_info, const VulkanInjectedSemaphore& semaphore)
+{
+    std::vector<VkSemaphore>&          wait_semaphores                = GetWaitSemaphores(submit_info);
+    std::vector<VkPipelineStageFlags>& wait_dst_stage_masks           = GetWaitDstStageMasks(submit_info, true);
```
fabian-lunarg: `GetWaitDstStageMasks(submit_info, true);` - the overload-flag is a bit confusing and
here (A)I think it could/should be left out entirely (defaults to false, just use `GetWaitDstStageMasks(submit_info)`):

When called from `SerializeExecution` after `InjectBefore`: 
existing masks were already all `TOP_OF_PIPE_BIT` (set by InjectBefore)

When called from `SerializeExecution` standalone (no `InjectBefore`):
original application stage masks for pre-existing binary waits are preserved, and only the new timeline wait gets `TOP_OF_PIPE_BIT`

fabian-lunarg (approved):
only one minor comment remains, I'm happy (again :)

antonio-lunarg:
I've also created a test app but I pushed it on a different branch as I believe I can't reliably test this feature. https://github.com/antonio-lunarg/gfxreconstruct/tree/antonio-serialize-submissions-test

According to [vkQueueSubmit specs](https://registry.khronos.org/VulkanSC/specs/1.0-extensions/man/html/vkQueueSubmit.html?utm_source=chatgpt.com):

> Batches begin execution in the order they appear in pSubmits, but may complete out of order.
>
> The order that batches appear in pSubmits is used to determine [submission order](https://registry.khronos.org/vulkansc/specs/1.0-extensions/html/vkspec.html#synchronization-submission-order), and thus all the [implicit ordering guarantees](https://registry.khronos.org/vulkansc/specs/1.0-extensions/html/vkspec.html#synchronization-implicit) that respect it. Other than these implicit ordering guarantees and any [explicit synchronization primitives](https://registry.khronos.org/vulkansc/specs/1.0-extensions/html/vkspec.html#synchronization), these batches may overlap or otherwise execute out of order.

I was trying to _force_ out-of-order completion in the test app, but this condition can only be created on an implementation that supports overlap. If the driver serializes the queue internally, the trick used by the test app is not effective anyway.

framework/decode/vulkan_submit_job.cpp
antonio-lunarg: The overload-flag was totally confusing and unrelated to that function. I just extracted that part in `InjectBefore` directly.

PR #2830, opened 2026-03-30
Fix DX12 tight alignment

locke-lunarg:
closed: https://github.com/LunarG/Projects/issues/1125
Ref: https://microsoft.github.io/DirectX-Specs/d3d/D3D12TightPlacedResourceAlignment.html

Regularly, the alignment of the resource should be `D3D12_DEFAULT_RESOURCE_PLACEMENT_ALIGNMENT(64KB)`. But when it uses tight alignment, it becomes 8B ~ 256B. It's smaller than a page size(`system_page_size_: 4KB`). In this case, one page could have two resources.   For `WriteWatch` case, in `LoadActiveWriteStates`, if it resets `WriteWatch` on `GetWriteWatch`, another resource on the same page will miss its update. Although `GetWriteWatch` could set precise address, OS operates on page by page. It shouldn't have reset `WriteWatch` until all memory infos get updated. 

There are the other two solutions that could fix it.
1. Force `ID3D12Device_Wrapper::GetResourceAllocationInfo` to return alignment to multiple of a page size, so a page won't have two resources. But we shouldn't change the original value.

2. `GFXRECON_PAGE_GUARD_EXTERNAL_MEMORY` set false could disable `WriteWatch`. It will use shadow memory, instead of `WriteWatch`. But shadow memory consumes huge memory. And `PageGuardManager::HandleGuardPageViolation` interrupts the process to update memory info. It could impact the performance.

### DIFF
--- a/framework/util/page_guard_manager.cpp
+++ b/framework/util/page_guard_manager.cpp
@@ -807,8 +807,8 @@ void PageGuardManager::LoadActiveWriteStates(MemoryInfo* memory_info)
                 if ((modified_page_address != first_page_address) && (modified_page_address != last_page_address))
                     continue;
 
-                auto it = page_to_memory_infos_.find(modified_page_address);
-                if (it == page_to_memory_infos_.end())
+                auto it = page_to_memory_infos_for_write_watch_.find(modified_page_address);
+                if (it == page_to_memory_infos_for_write_watch_.end())
                     continue;
 
                 for (auto& other_mem_info : it->second)
@@ -818,9 +818,14 @@ void PageGuardManager::LoadActiveWriteStates(MemoryInfo* memory_info)
 
                     other_mem_info->is_modified = true;
 
+                    GFXRECON_ASSERT(static_cast<uint8_t*>(modified_page_address) >=
+                                    static_cast<uint8_t*>(other_mem_info->aligned_address));
+
                     size_t other_start_offset = static_cast<uint8_t*>(modified_page_address) -
                                                 static_cast<uint8_t*>(other_mem_info->aligned_address);
+
                     size_t other_page_index = other_start_offset >> system_page_pot_shift_;
+                    GFXRECON_ASSERT(other_page_index < other_mem_info->total_pages);
 
                     other_mem_info->status_tracker.SetActiveWriteBlock(other_page_index, true);
                 }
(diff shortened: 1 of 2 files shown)


framework/util/page_guard_manager.cpp
```diff
@@ -1325,6 +1325,17 @@ void PageGuardManager::ProcessMemoryEntry(uint64_t memory_id, const ModifiedMemo
...
 
+void PageGuardManager::ResetAllWriteWatch()
+{
+#if defined(WIN32)
+    for (auto& entry : memory_info_)
+    {
+        auto& info = entry.second;
+        ResetWriteWatch(info.aligned_address, info.mapped_range + info.aligned_offset);
+    }
+#endif
```
davidd-lunarg: Could this introduce a race condition in processing modified pages? For example:

1. `PageGuardManager::ProcessMemoryEntries` is called. Memory A is found to be modified and is processed. `ProcessMemoryEntries` continues to processing the remaining modified pages.
2. After memory A is processed, but before `ProcessMemoryEntries` calls `ResetAllWriteWatch();` (it could be processing other modified memory pages), the application completes another write to memory A.
3. `ResetAllWriteWatch();` is called, clearing the new write watch flag on memory A, causing the application's latest write to be skipped at the next call to `ProcessMemoryEntries`.

locke-lunarg:
@davidd-lunarg Upload a new solution. It updates the memory info on the same page in `PageGuardManager::LoadActiveWriteStates`, so it doesn't need to delay reset WriteWatch.

davidd-lunarg:
> @davidd-lunarg Upload a new solution. It updates the memory info on the same page in PageGuardManager::LoadActiveWriteStates, so it doesn't need to delay reset WriteWatch.

Thanks! The changes look good to me. Since this touches code paths in the `PageGuardManager` that are used by all APIs and platforms during capture, it should be well-tested before merging. A couple other questions I have are:

1. Does this have any significant affect on performance or capture file size?
2. A while back, `GFXRECON_PAGE_GUARD_ALIGN_BUFFER_SIZES` was set to true by default for Vulkan (PR #1431 and Issue #1228). Does this change that requirement? FYI @bradgrantham-lunarg 
3. Does this have any impact or side effects for capture of aliased/overlapping resources?

MarkY-LunarG:
> 2. A while back, GFXRECON_PAGE_GUARD_ALIGN_BUFFER_SIZES was set to true by default for Vulkan (PR https://github.com/LunarG/gfxreconstruct/pull/1431 and Issue https://github.com/LunarG/gfxreconstruct/issues/1228). Does this change that requirement? FYI @bradgrantham-lunarg

Yes, I believe so.  I think @ziga-lunarg was the one who ran into issues and suggested forcing it.

davidd-lunarg:
> 2. A while back, GFXRECON_PAGE_GUARD_ALIGN_BUFFER_SIZES was set to true by default for Vulkan. Does this change that requirement?

Looking into it more, I think `GFXRECON_PAGE_GUARD_ALIGN_BUFFER_SIZES` is still required for default Vulkan memory capture mode (page guard + shadow memory) to avoid application read and write to the same memory page from causing corruption. This change might make it possible to disable `GFXRECON_PAGE_GUARD_ALIGN_BUFFER_SIZES` when using `GFXRECON_PAGE_GUARD_EXTERNAL_MEMORY`. `GFXRECON_PAGE_GUARD_EXTERNAL_MEMORY` is the default mode for DX12 but it is only available on Windows.

I don't think this PR should make any changes to `GFXRECON_PAGE_GUARD_ALIGN_BUFFER_SIZES`, but I did want to make note of it.

locke-lunarg:
@davidlunarg  Fixed the bug that you messaged to me.

https://github.com/LunarG/gfxreconstruct/issues/2853 We can revisit this issue later.

panos-lunarg:
Recently I encountered a problem that sounds very similar to what is described in the original message by Locke.
An application updates a memory region (uploads vertex data) with 2 different methods:
1. Two vkCmdCopyBuffer
2. By directly writing to the mapped memory (caught by page_guard)

The problem is that the FillMemoryCommand blocks have both size and alignment of the system's page size and as a result they clobber the writes done by the CmdCopyBuffers. The following image should help visualize the problem:

573481307-795f5f3d-d2c0-4786-ab2b-58d8fbfb5b5e

Does this PR have the potential to solve this problem too?

PR #2831, opened 2026-03-30
Fix SEGFAULT in `VulkanVirtualSwapchain::FrameBoundaryANDROID`

marius-pelegrin-arm:
Introduced in commit `40ba81c24dd6b5a3094427495e43568de8d11723` The method uses `device_table_`, which is not guaranteed to be initialized or even valid if initialized, instead of `device_table` passed as an argument.

I also remove the `if` around the call, I couldn't repro any other SEGFAULT besides that, so I assume this was the SEGFAULT and had nothing to do with the WSI.. You can add it back if I am wrong.


fabian-lunarg:
good spot! I guess having both a `device_table_` and parameter `device_table` was basically asking for trouble.

PR #2835, opened 2026-03-31
Support multiple adhoc-swapchains, use for multi-layered images

fabian-lunarg:
- refactor adhoc-windowing code to support multiple windows
- present multi-layer images in `Swapchain::PresentImageAdHoc` using multiple windows
- fix missing sRGB check for fallback color-format
- add option to blit into specific layer(s)
- some work on usability, window-title/position

### DIFF
--- a/framework/decode/vulkan_virtual_swapchain.cpp
+++ b/framework/decode/vulkan_virtual_swapchain.cpp
@@ -28,6 +28,7 @@
 #include "util/callbacks.h"
 #include "graphics/vulkan_resources_util.h"
 #include "vulkan/vulkan_core.h"
+#include <sstream>
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)


framework/decode/vulkan_virtual_swapchain.h
```diff
@@ -193,43 +193,96 @@ class VulkanVirtualSwapchain : public VulkanSwapchain
...
+        ~AdhocSwapChain();
+
+        AdhocSwapChain(const AdhocSwapChain&)            = delete;
+        AdhocSwapChain& operator=(const AdhocSwapChain&) = delete;
+
+        // idiomatic copy & swap
+        AdhocSwapChain(AdhocSwapChain&& other) noexcept : AdhocSwapChain() { swap(*this, other); }
+        AdhocSwapChain& operator=(AdhocSwapChain other) noexcept
+        {
+            swap(*this, other);
```
MarkY-LunarG, line 222: Do you need "using std::swap" above this too (and above on line 219) or at least globally?

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -1321,288 +1315,329 @@ void VulkanVirtualSwapchain::PresentImageAdHoc(const VulkanDeviceInfo*
...
-            // We need to be sure that swapchain resources are not in use anymore
-            result = device_table->QueueWaitIdle(ofb_data.queue);
-            GFXRECON_ASSERT(result == VK_SUCCESS);
+            {
+                std::stringstream title_ss;
+                title_ss << image_info->debug_utils_name;
+                title_ss << " - " << window_width << " x " << window_height;
+                if (layer_count > 1)
+                {
+                    title_ss << " - layer: " << (layer + 1) << " / " << layer_count;
```
MarkY-LunarG, line 1375: Is it typical for layers to be 1 based?

framework/decode/vulkan_virtual_swapchain.h
fabian-lunarg: only used in the body of AdHocSwapChain::swap(), so should not have spilled into the namespace.

framework/decode/vulkan_virtual_swapchain.cpp
fabian-lunarg: no, but I thought it would make more intuitive sense,  when displaying as a count. e.g. for two images I wanted to avoid: 0 / 2 | 1 / 2 and preferred 1 / 2 | 2 / 2. also the resulting title-bar is not guaranteed to be even visible (e.g. on wayland we do not render window-decorations at all). so this whole string-formatting was mostly a nice to have

fabian-lunarg:
thank you @MarkY-LunarG !

PR #2836, opened 2026-03-31
Handle dx12 special array pointer 

zongdu-arm:
Handle special array pointer for D3D12_RAYTRACING_OPACITY_MICROMAP_ARRAY_DESC


zongdu-arm:
@locke-lunarg

PR #2837, opened 2026-03-31
Fix --wsi auto side effects

roman-berger-arm:
Currently `--wsi auto` (default flag) is adding all surface extensions from the replay and trying to initialize their compositors. This is done to deal with an edge case of where multiple WSIs are available in the caputre. However, when compositors are not available on the replay device, the surface extensions are not removed leading to the application breaking during the `vkCreateInstance` call with `VK_ERROR_EXTENSION_NOT_PRESENT`. This makes `--remove-unsupported` a required option to replay the trace.

The change that creates this behaviour is introduced in PR #2736, while testing the fix I could not reproduce the issue with `vulkaninfo` that originally prompted this change.

Instead of adding all the surface extensions from the replay and then removing some if they are not available. This change proposes to add them if they are available, by going through all the capture time surface extensions and checking if they are available and then adding them.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -2891,7 +2891,8 @@ void VulkanReplayConsumerBase::ModifyCreateInstanceInfo(
         }
     }
 
-    std::string capture_surface_extension;
+    // In case there is more than one replay surface
+    std::vector<std::string> capture_surface_extensions;
 
     // Transfer requested extensions to filtered extension
     for (uint32_t i = 0; i < replay_create_info->enabledExtensionCount; ++i)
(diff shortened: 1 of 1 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -2958,8 +2962,14 @@ void VulkanReplayConsumerBase::ModifyCreateInstanceInfo(
...
                             modified_extensions.push_back(itr_surface_extension->c_str());
-                            GFXRECON_LOG_WARNING("--wsi auto: could not find surface: %s, instead using: %s",
-                                                 capture_surface_extension.c_str(),
+                            std::string all_capture_surface_extensions = std::string("");
+                            for (std::string ext : capture_surface_extensions)
+                            {
+                                all_capture_surface_extensions += ext;
+                                all_capture_surface_extensions += ", ";
+                            }
+                            GFXRECON_LOG_WARNING("--wsi auto: could not find surface: %sinstead using: %s",
```
MarkY-LunarG, line 2971: You removed the comma and space after the first %s.

roman-berger-arm: Yes, because the for-loop will add a comma and space to the end of the string `all_capture_surface_extensions` during every iteration. Therefore there would be two sets of ", " at the end of that list if I didn't remove it.

PR #2858, opened 2026-04-07
Split Convert into Features for easier expansion

MarkY-LunarG:
This is the next change after #2751 (merged)

### DIFF
--- a/tools/convert/convert_d3d12_feature.cpp
+++ b/tools/convert/convert_d3d12_feature.cpp
@@ -1,0 +1,56 @@
+/*
+** Copyright (c) 2020-2026 LunarG, Inc.
+** Copyright (c) 2022-2024 Advanced Micro Devices, Inc. All rights reserved.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and associated documentation files (the "Software"),
+** to deal in the Software without restriction, including without limitation
+** the rights to use, copy, modify, merge, publish, distribute, sublicense,
+** and/or sell copies of the Software, and to permit persons to whom the
+** Software is furnished to do so, subject to the following conditions:
+**
+** The above copyright notice and this permission notice shall be included in
+** all copies or substantial portions of the Software.
+**
+** THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+** IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+** FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+** AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+** FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+** DEALINGS IN THE SOFTWARE.
+*/
+
+#if defined(D3D12_SUPPORT)
+
+#include "convert_d3d12_feature.h"
+
+#include "util/feature_module_registry.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(convert)
+
+// Register this class as a feature in a module registry
+GFXR_UTIL_REGISTER_FEATURE_CREATOR(ConvertFeature, ConvertD3d12Feature)
+
+void ConvertD3d12Feature::RegisterDecodeComponents(gfxrecon::decode::FileProcessor& file_processor)
+{
+    dx12_decoder_.AddConsumer(&dx12_json_consumer_);
+    file_processor.AddDecoder(&dx12_decoder_);
+}
+
+void ConvertD3d12Feature::RegisterJsonWriter(gfxrecon::decode::JsonWriter* json_writer)
+{
+    dx12_json_consumer_.Initialize(json_writer);
+}
+
+void ConvertD3d12Feature::UnregisterJsonWriter()
+{
+    // If CONVERT_EXPERIMENTAL_D3D12 was set, then cleanup DX12 consumer
+    dx12_json_consumer_.Destroy();
+}
+
+GFXRECON_END_NAMESPACE(info)
+GFXRECON_END_NAMESPACE(gfxrecon)
+
+#endif // D3D12_SUPPORT
--- a/tools/convert/convert_d3d12_feature.cpp
+++ b/tools/convert/convert_d3d12_feature.cpp
@@ -1,0 +1,46 @@
+/*
+** Copyright (c) 2020-2026 LunarG, Inc.
+** Copyright (c) 2022-2024 Advanced Micro Devices, Inc. All rights reserved.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and associated documentation files (the "Software"),
+** to deal in the Software without restriction, including without limitation
+** the rights to use, copy, modify, merge, publish, distribute, sublicense,
+** and/or sell copies of the Software, and to permit persons to whom the
+** Software is furnished to do so, subject to the following conditions:
+**
+** The above copyright notice and this permission notice shall be included in
+** all copies or substantial portions of the Software.
+**
+** THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+** IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+** FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+** AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+** FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+** DEALINGS IN THE SOFTWARE.
+*/
+
+#if defined(D3D12_SUPPORT)
+
+#include "convert_feature.h"
+
+#include "generated/generated_dx12_decoder.h"
+#include "generated/generated_dx12_json_consumer.h"
+#include "util/feature_module_registry.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(convert)
+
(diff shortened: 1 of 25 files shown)


MarkY-LunarG:
Ok.  Should I merge the convert_*_feature.cpp classes into the header files?  They seem so lightweight that maybe it makes sense? [@mikes-lunarg / @ncesario-lunarg ]

mikes-lunarg:
> Ok. Should I merge the convert_*_feature.cpp classes into the header files? They seem so lightweight that maybe it makes sense? [@mikes-lunarg / @ncesario-lunarg ]

+1 Personally, I prefer smaller implementations inlined in the header

MarkY-LunarG:
> > Ok. Should I merge the convert_*_feature.cpp classes into the header files? They seem so lightweight that maybe it makes sense? [@mikes-lunarg / @ncesario-lunarg ]
> 
> +1 Personally, I prefer smaller implementations inlined in the header

I just realized I still need the following in the cpp file for each one:

```
// Register this class as a feature in a module registry
GFXR_UTIL_REGISTER_FEATURE_CREATOR(ConvertFeature, ConvertD3d12Feature)
```

So if I moved everything else it would still leave this in the CPP file.  Or I just leave it as is.

tools/convert/convert_d3d12_feature.cpp
```diff
@@ -0,0 +1,56 @@
...
+}
+
+void ConvertD3d12Feature::RegisterJsonWriter(gfxrecon::decode::JsonWriter* json_writer)
+{
+    dx12_json_consumer_.Initialize(json_writer);
+}
+
+void ConvertD3d12Feature::UnregisterJsonWriter()
+{
+    // If CONVERT_EXPERIMENTAL_D3D12 was set, then cleanup DX12 consumer
```
mikes-lunarg: This comment old, D3D12 convert is not experimental anymore

tools/convert/convert_feature.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include <string>
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(convert)
+
+class ConvertFeature
+{
+  public:
+    virtual ~ConvertFeature() = default;
```
ncesario-lunarg: Would it make sense to call `Unregister*` functions here? The goal being to simplify cleanup in main. It may not make sense depending on what else `Unregister` might do in the future.

tools/convert/convert_vulkan_feature.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+// Register this class as a feature in a module registry
+GFXR_UTIL_REGISTER_FEATURE_CREATOR(ConvertFeature, ConvertVulkanFeature)
+
+std::string ConvertVulkanFeature::GetVersionString() const
+{
+    return std::to_string(VK_API_VERSION_MAJOR(VK_HEADER_VERSION_COMPLETE)) + "." +
+           std::to_string(VK_API_VERSION_MINOR(VK_HEADER_VERSION_COMPLETE)) + "." +
+           std::to_string(VK_API_VERSION_PATCH(VK_HEADER_VERSION_COMPLETE));
+}
```
mikes-lunarg: This could be constexpr

tools/convert/convert_vulkan_feature.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+void ConvertVulkanFeature::RegisterDecodeComponents(gfxrecon::decode::FileProcessor& file_processor)
+{
+    vulkan_decoder_.AddConsumer(&vulkan_json_consumer_);
+    file_processor.AddDecoder(&vulkan_decoder_);
+}
+
+void ConvertVulkanFeature::RegisterJsonWriter(gfxrecon::decode::JsonWriter* json_writer)
+{
+    vulkan_json_consumer_.Initialize(json_writer, GetVersionString());
```
mikes-lunarg: It feels a little strange to have to pass the vulkan version to the `vulkan_json_consumer_.Initialize`. It seems like the Vulkan consumer should be able to know what Vulkan version it was generated against.

tools/convert/convert_d3d12_feature.h
```diff
@@ -0,0 +1,65 @@
...
+#include "generated/generated_dx12_decoder.h"
+#include "generated/generated_dx12_json_consumer.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(convert)
+
+using Dx12JsonConsumer =
+    gfxrecon::decode::MetadataJsonConsumer<gfxrecon::decode::MarkerJsonConsumer<gfxrecon::decode::Dx12JsonConsumer>>;
+
+class ConvertD3d12Feature : public ConvertFeature
```
ncesario-lunarg: IIUC, the `ConvertFeature` for each API is effectively just a container for a consumer, decoder, and a label? If so, you could probably create a template class to remove a lot of boilerplate here, then just register of these "standard APIs" in main.

ncesario-lunarg (approved):
Some minor comments/questions but LGTM!

tools/convert/main.cpp
```diff
@@ -458,16 +421,10 @@ int main(int argc, const char** argv)
...
-#endif
-
-            // If CONVERT_EXPERIMENTAL_D3D12 was set, then cleanup DX12 consumer
-#ifdef D3D12_SUPPORT
-            dx12_json_consumer.Destroy();
-#endif
+            for (auto& feature : convert_features)
+            {
+                feature->UnregisterJsonWriter();
+            }
```
mikes-lunarg: `feature->UnregisterJsonWriter` calls `API_consumer_->Destroy()` which calls `json_writer.Destroy()`. Since the json writer can only be ended/closed once, the other ones end up doing nothing*.  Would ending the json stream be better handled up here in main.cpp since all of the consumers a sharing a single json writer?

*Except DX12, for some reason that destroy call always seems to do nothing

mikes-lunarg (commented):
Seems pretty straightforward based on the info change, just a couple of questions

mikes-lunarg:
> > > Ok. Should I merge the convert_*_feature.cpp classes into the header files? They seem so lightweight that maybe it makes sense? [@mikes-lunarg / @ncesario-lunarg ]
> > 
> > 
> > +1 Personally, I prefer smaller implementations inlined in the header
> 
> I just realized I still need the following in the cpp file for each one:
> 
> ```
> // Register this class as a feature in a module registry
> GFXR_UTIL_REGISTER_FEATURE_CREATOR(ConvertFeature, ConvertD3d12Feature)
> ```
> 
> So if I moved everything else it would still leave this in the CPP file. Or I just leave it as is.

Well shoot, I don't really have a strong preference either way then

tools/convert/convert_vulkan_feature.cpp
MarkY-LunarG: Does not work for me:

```
error: call to non-‘constexpr’ function ‘std::string std::__cxx11::to_string(int)’
```

tools/convert/convert_d3d12_feature.h
MarkY-LunarG: The problem is that the main.cpp then needs to know about each instatiation.  So it would have to know about D3D, Vulkan and OPenXR.  With this approach, main.cpp is ignorant and we can easily add new APIs while leaving main.cpp untouched.

tools/convert/convert_vulkan_feature.cpp
mikes-lunarg: Ah, your right to_string isn't constexpr because it might allocate.  It is possible to write a constexpr version, but even then you would be relying on the implementation-specific short string optimization so it might be fragile:
https://godbolt.org/z/xGsTYs33E

tools/convert/main.cpp
MarkY-LunarG: Good point!  I'll make the Json Consumers behave like D3D12 and destroy the writer in main.cpp.  Much cleaner..

tools/convert/convert_vulkan_feature.cpp
MarkY-LunarG: Great point.  I'm going to change this and the OpenXR code to have that automatically calculated inside the consumers.

MarkY-LunarG:
@mikes-lunarg / @ncesario-lunarg :  I completely modified the classes so there's only a base type and a template and all the instances use the template.  So the code is greatly simplified.  I also modified the writer to only destroy in main.cpp as Mike requested.

tools/convert/convert_feature.h
```diff
@@ -0,0 +1,109 @@
...
+            string_data[i] = in_str[i];
+        }
+    }
+};
+
+// This is the template used by all the actual API instantiations.
+// Since they are all pretty generic, the basic functionality defined
+// here is all that is needed beyond the basic types.
+template <class CON, class DEC, ConvertFixedString ST>
+class ActualConvertFeature : public ConvertFeature
```
mikes-lunarg: Consider the following naming tweak:

`class ConvertFeature : public ConvertFeatureBase`

instead of the current:

`class ActualConvertFeature : public ConvertFeature`

tools/convert/convert_feature.h
```diff
@@ -0,0 +1,109 @@
...
+{
+  public:
+    ActualConvertFeature() { label_ = ST.string_data; }
+    ~ActualConvertFeature() { UnregisterJsonWriter(); }
+
+    // Simple "getter" style methods
+    virtual std::string Label() final { return label_; }
```
mikes-lunarg: derived methods don't need to be marked `virtual` but should absolutely be marked `override`.

tools/convert/main.cpp
```diff
@@ -204,10 +182,21 @@ std::string FormatFrameNumber(uint32_t frame_number)
...
     gfxrecon::util::Log::Init();
 
+    // Query the module registry for registered modules, and
+    // call each generator here and put the unique_ptr into our
+    // internal unique_ptr vector.
+    for (const auto& registered_creator :
+         gfxrecon::util::FeatureModuleRegistry<gfxrecon::convert::ConvertFeature>::GetSingleton()
+             .GetRegisteredFeatureCreators())
+    {
+        convert_features.push_back(std::move(registered_creator()));
```
mikes-lunarg: `std::move` isn't necessary here, it's already an rvalue

mikes-lunarg (commented):
This is highly subjective, but I think the `ConvertFeature` register/register/unregister interface is a still a little odd. The initialization is split into 2 functions (`RegisterDecodeComponents` and `RegisterJsonWriter`) which both need to be called before any work can be done. Why not combine them into a single `Initialize` function? `UnregisterJsonWriter` could be renamed `Destroy` which I think matches what it is doing more closely, or even rolled into the convert feature destructor so you wouldn't have to call it at all.

tools/convert/convert_feature.h
MarkY-LunarG: Done

tools/convert/convert_feature.h
MarkY-LunarG: I used final instead of override, but removed virtual.

tools/convert/main.cpp
MarkY-LunarG: Stripped

mikes-lunarg (approved):

PR #2866, opened 2026-04-10
Add merge_group to CI build workflow for dev branch

beau-lunarg:


.github/workflows/ci_build.yml
```diff
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches:
       - dev
+  merge_group:
+    branches:
+      - dev
```
mikes-lunarg, line 11: Crazy to me that you can't tell from the docs whether the merge_group trigger has a branches filter: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#merge_group

But the vscode GHA extension is fine with it and this blog post says it's OK https://github.blog/changelog/2022-08-18-merge-group-webhook-event-and-github-actions-workflow-trigger/

mikes-lunarg (approved):

.github/workflows/ci_build.yml
beau-lunarg: Also there is the branch where I set up the merge queue tests, which is what I'm copying from for this
https://github.com/LunarG/gfxreconstruct/blob/merge-test/.github/workflows/ci_build.yml

PR #2869, opened 2026-04-13
plugin: Replay event plugin

antonio-lunarg:
Introduce a stable C ABI plugin interface that lets an external library observe frame and queue-submit related events during Vulkan replay.

- ABI: Versioned C ABI header with four events (frame begin/end, queue submit begin/end)
- `ReplayEventSink`: replay consumer submits events to the sink. The sink itself is a polymorphic type and the default sink is a no-op object.
- `PluginReplayEventSink`: validates and forward events to a library through the ABI header.
- `LoadPlugin()`: loads a library for the `PluginReplayEventSink`. This function accepts custom load operations (open_library, get_proc_address, close_library) for unit testing without a real `.so`.
- `framework/plugin/sample`: minimal print plugin demo.

The changes are limited to Vulkan replay on desktop and Android only.


framework/plugin/replay_event_sink.h
```diff
@@ -0,0 +1,103 @@
...
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(plugin)
+
+class ReplayEventSink
+{
+  public:
+    virtual ~ReplayEventSink() = default;
+
+    uint64_t QueueSubmitBegin(uint64_t queue_id);
```
fabian-lunarg: better use `format::HandleId`? it's identical type but better communicates our intention

framework/plugin/replay_event_sink.h
```diff
@@ -0,0 +1,103 @@
...
+
+  private:
+    GfxrReplayEventHeader CreateEventHeader(GfxrReplayEventType type);
+
+    bool     frame_active_        = false;
+    uint64_t current_frame_index_ = 0;
+
+    uint64_t current_submit_index_ = 0;
+    uint64_t first_submit_index_   = GFXR_REPLAY_INVALID_SUBMIT_INDEX;
+    uint64_t last_submit_index_    = GFXR_REPLAY_INVALID_SUBMIT_INDEX;
```
fabian-lunarg, line 64: since all members are integers the class "could" easily be threadsafe by using atomics.
currently it's driven by the single-threaded replay-consumer, so just curious, could e.g. `bool IsFrameActive()` be subject to concurrency situation and benefit from thread-safe design?

antonio-lunarg: Not sure. QueueSubmitBegin/End would need a mutex, probably. Anyway, using multithread at the frame-granularity sounds a bit odd to me. While multithreading is nice for everything else, I would still serialize presentation at both the CPU and GPU level.

bradgrantham-lunarg:
@fabian-lunarg As far as you are concerned has all your feedback been addressed and you would approve this PR?

docs/PLUGIN_replay_event.md
```diff
@@ -0,0 +1,458 @@
...
+--sync
+```
+
+Desktop expectations:
+
+* the option is passed directly to `gfxrecon-replay`
+* the optional parameter string is passed unchanged into `GfxrReplayPluginCreateInfo.plugin_params`
+* if loading fails, replay logs the failure and continues without plugin integration
+
+Android expectations:
```
antonio-lunarg, line 379: The method described here works on non-rooted devices.

framework/application/application.h
```diff
@@ -99,6 +100,16 @@ class Application final
         return GFXRECON_NARROWING_CAST(uint32_t, file_processor_->GetCurrentFrameNumber());
     }
 
+    plugin::ReplayEventSink* GetReplayEventSink() const
+    {
+        return replay_event_sink_.get();
+    }
+
+    void SetReplayEventSink(std::unique_ptr<plugin::ReplayEventSink> replay_event_sink)
```
fabian-lunarg, line 108: when someone would call `SetReplayEventSink(nullptr)` this could replace the valid/default `plugin::NullReplayEventSink` with nullptr. 
add a check?

framework/plugin/test/main.cpp
```diff
@@ -0,0 +1,373 @@
...
+static gfxrecon::util::platform::LibraryHandle FakeOpenLibrary(const char* library_path)
+{
+    if (strcmp(library_path, kSuccessPluginPath) == 0)
+    {
+        return kFakeLibraryHandle;
+    }
+    return nullptr; // Simulate failure for all other paths
+}
+
+static void* FakeCreatePlugin(const char* args)
```
fabian-lunarg: not sure this would ever surface as bug, but the signature here should be `GfxrReplayPluginV1* (*)(const GfxrReplayPluginCreateInfo*)` ?
so not `char*` but `const GfxrReplayPluginCreateInfo*` ?

fabian-lunarg (approved):
just left two minor comments, approved!

framework/plugin/replay_event_plugin_loader.cpp
```diff
@@ -0,0 +1,100 @@
...
+    {
+        GFXRECON_LOG_ERROR("Failed to create plugin instance: %s", load_info.library_path.c_str());
+        ops.close_library(library);
+        return nullptr;
+    }
+
+    if (!IsValidPlugin(plugin))
+    {
+        GFXRECON_LOG_ERROR("Invalid plugin instance: %s", load_info.library_path.c_str());
+        if (plugin->destroy != nullptr)
```
fabian-lunarg: ok, found one more thingy.

when `!IsValidPlugin(plugin)` that 'might' be because `plugin->struct_size < sizeof(GfxrReplayPluginV1)`
in that case `plugin->destroy` would be reading past the struct-end and yield UB.

so not sure if we should call  `plugin->destroy` on invalid plugins or at least recheck the `struct_size`?

antonio-lunarg: Added a check for struct size before attempting to call `destroy()`

PR #2870, opened 2026-04-14
loop: Light refactor to simplify future rebase.

jzulauf-lunarg:
Reduced the changes to interface and replay flow.

### DIFF
--- a/framework/application/application.cpp
+++ b/framework/application/application.cpp
@@ -139,24 +139,13 @@ void Application::Run()
         if (running_ && !paused_)
         {
             // Add one to match "trim frame range semantic"
-            uint32_t frame_number = file_processor_->GetCurrentFrameNumber() + 1;
+            uint64_t frame_number = file_processor_->GetCurrentFrameNumber() + 1;
 
-            if (frame_loop_info_ != nullptr)
+            if ((frame_loop_info_ != nullptr) && (frame_loop_info_->ShouldStartFrameLooping(frame_number)))
             {
-                frame_loop_info_->SetCurrentFrameNumber(frame_number);
-
-                // Quit when frame looping has finished.
-                if (frame_loop_info_->GetLoopIterations() == 0)
-                {
-                    running_ = false;
-                    break;
-                }
-
-                if (frame_loop_info_->IsFirstIteration())
-                {
-                    // Preload the next frame and make sure we don't advance to the next one.
-                    GetPreloadFileProcessor()->PreloadNextFrames(1);
-                }
+                // Preload the next frame and make sure we don't advance to the next one.
+                GetPreloadFileProcessor()->PreloadLoopFrame();
+                frame_loop_info_->SetLooping(true);
             }
 
             if (fps_info_ != nullptr)
@@ -184,23 +173,14 @@ void Application::Run()
             // PlaySingleFrame() increments this->current_frame_number_ *if* there's an end-of-frame
             PlaySingleFrame();
 
-            if (frame_loop_info_ != nullptr)
+            if ((frame_loop_info_ != nullptr) && (frame_loop_info_->IsLooping()))
             {
-                if (frame_loop_info_->IsFirstIteration())
+                // Quit when frame looping has finished.
+                frame_loop_info_->DecrementLoopIterations();
+                GFXRECON_LOG_INFO("Looping frame (%i iterations remaining)", frame_loop_info_->GetLoopIterations());
+                if (frame_loop_info_->GetLoopIterations() == 0)
                 {
-                    frame_loop_info_->SetLooping(true);
-                }
-
-                if (frame_loop_info_->IsLooping())
-                {
-                    auto* preload_processor = GetPreloadFileProcessor();
-                    preload_processor->WaitDecodersIdle();
-
-                    // When replaying a frame again, skip any state blocks to avoid reapplying them again.
-                    preload_processor->SkipStateBlocks();
-
-                    GFXRECON_LOG_INFO("Looping frame (%i iterations remaining)", frame_loop_info_->GetLoopIterations());
-                    frame_loop_info_->DecrementLoopIterations();
+                    running_ = false;
                 }
             }
 
@@ -241,15 +221,7 @@ bool Application::PlaySingleFrame()
 
     if (file_processor_)
     {
-        if (frame_loop_info_ != nullptr && (frame_loop_info_->IsFirstIteration() || frame_loop_info_->IsLooping()))
-        {
-            // When looping, replay the current preloaded frame without advancing.
-            success = GetPreloadFileProcessor()->ReplayCurrentPreloadedFrame();
-        }
-        else
-        {
-            success = file_processor_->ProcessNextFrame();
-        }
+        success = file_processor_->ProcessNextFrame();
 
         if (success)
         {
--- a/framework/decode/preload_file_processor.cpp
+++ b/framework/decode/preload_file_processor.cpp
@@ -26,12 +26,19 @@
 #include "util/logging.h"
 
 #include <memory>
+#include "preload_file_processor.h"
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
 PreloadFileProcessor::PreloadFileProcessor() : working_uncompressed_store_(kWorkingStoreInitialSize) {}
 
+void PreloadFileProcessor::PreloadLoopFrame()
+{
+    PreloadNextFrames(1);
+    loop_replay_      = true;
+    loop_reset_point_ = SkipStateBlocks(current_frame_number_, replay_cursor_);
+}
 void PreloadFileProcessor::PreloadNextFrames(size_t count)
 {
     size_t expected_count = count;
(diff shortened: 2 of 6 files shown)


framework/graphics/frame_loop_info.h
```diff
@@ -40,13 +40,11 @@ class FrameLoopInfo
     {}
 
     /// Expects a 1-based frame number.
-    void SetCurrentFrameNumber(uint32_t frame_number) { current_frame_number_ = frame_number; }
-
     /// Returns true if the current frame number matches the loop frame index.
-    bool AtLoopFrame() const { return current_frame_number_ == loop_frame_idx_; }
+    bool AtLoopFrame(uint64_t frame_number) const { return frame_number == loop_frame_idx_; }
 
     /// Returns true if this is the first iteration of the loop frame.
```
antonio-lunarg: This comment should be different for this function

framework/graphics/frame_loop_info.h
```diff
@@ -40,13 +40,11 @@ class FrameLoopInfo
...
-    bool IsFirstIteration() const { return AtLoopFrame() && !IsLooping(); }
```
antonio-lunarg, line 49: We still want `IsFirstIteration()`, especially when overriding some calls we want to know if this is the first iteration of the loop.

framework/decode/preload_file_processor.cpp
```diff
@@ -182,72 +189,58 @@ bool PreloadFileProcessor::ProcessNextFrame()
...
-    return AdvancePreloadedFrame();
-}
-
-bool PreloadFileProcessor::AdvanceToNextFrame(ReplayFrameResult replay_result)
-{
-    replay_cursor_ = replay_result.next_frame_cursor;
+    ProcessBlockState process_result = ReplayOneFrame();
 
     const bool at_end = (!replay_cursor_);
+    if (loop_replay_)
```
antonio-lunarg, line 195: I'd like @bradgrantham-lunarg's opinion on this, considering the comments on the initial frame looping PR https://github.com/LunarG/gfxreconstruct/pull/2698#discussion_r2976001183

antonio-lunarg: The main question is: should the file processor be aware and responsible in some way of frame looping? Should the application have this responsibility instead?

jzulauf-lunarg: @antonio-lunarg  -- With async processing (my pending change), the file processor needs to be aware of the preload range at least, and needs to stop preloading when the loop frame is encountered. Making looping a mode of frame processing keeps replay internals... internal.  I'm not even sure (after the async refactor) if PreloadFileProcessor should even be derived, as replay and queueing logic moves parent-ward.

framework/graphics/frame_loop_info.h
jzulauf-lunarg: I missed this.  The sideband communication from application to consumer seems dangerous, especially if any asynchronous operations are introduced...  

@bradgrantham-lunarg  -- this seems like it should be an antipattern moving forward, shouldn't it?

jzulauf-lunarg: @antonio-lunarg -- it's IsLooping the consumer use, changed to IsRepetition as this matches the local commentary and decouples it from the application level understanding of whether it currently IsLooping.

framework/application/application.cpp
```diff
@@ -139,24 +139,13 @@ void Application::Run()
...
-                }
-
-                if (frame_loop_info_->IsFirstIteration())
-                {
-                    // Preload the next frame and make sure we don't advance to the next one.
-                    GetPreloadFileProcessor()->PreloadNextFrames(1);
-                }
+                // Preload the next frame and make sure we don't advance to the next one.
+                GetPreloadFileProcessor()->PreloadLoopFrame();
+                frame_loop_info_->SetLooping(true);
```
jzulauf-lunarg, line 148: Left this here to keep the simplified application side logic.  Added `IsRepetition` for the consumers to call

framework/application/application.cpp
```diff
@@ -139,24 +139,13 @@ void Application::Run()
...
-                    break;
-                }
-
-                if (frame_loop_info_->IsFirstIteration())
-                {
-                    // Preload the next frame and make sure we don't advance to the next one.
-                    GetPreloadFileProcessor()->PreloadNextFrames(1);
-                }
+                // Preload the next frame and make sure we don't advance to the next one.
+                GetPreloadFileProcessor()->PreloadLoopFrame();
```
jzulauf-lunarg, line 147: After I commit async, we should come back and add correctly "empty preload" (i.e. zero complete frames) handling.

framework/application/application.cpp
```diff
@@ -184,23 +173,14 @@ void Application::Run()
...
+            if ((frame_loop_info_ != nullptr) && (frame_loop_info_->IsLooping()))
             {
-                if (frame_loop_info_->IsFirstIteration())
-                {
-                    frame_loop_info_->SetLooping(true);
-                }
-
-                if (frame_loop_info_->IsLooping())
+                // Quit when frame looping has finished.
+                frame_loop_info_->DecrementLoopIterations();
```
jzulauf-lunarg, line 179: Was half tempted to collapse all this into a "bool ContinueLooping" ...

framework/decode/preload_file_processor.cpp
```diff
@@ -26,12 +26,19 @@
...
 
 #include <memory>
+#include "preload_file_processor.h"
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
 PreloadFileProcessor::PreloadFileProcessor() : working_uncompressed_store_(kWorkingStoreInitialSize) {}
 
+void PreloadFileProcessor::PreloadLoopFrame()
```
jzulauf-lunarg, line 36: Preload for loops is always one frame, and knowing the replay mode minimizes exposing changing internals to `Application`

framework/decode/preload_file_processor.cpp
```diff
@@ -182,72 +189,58 @@ bool PreloadFileProcessor::ProcessNextFrame()
...
     {
         current_frame_number_++;
     }
-
-    return ContinueProcessing(replay_result.process_state);
+    return ContinueProcessing(process_result);
 }
 
-void PreloadFileProcessor::SkipStateBlocks()
+BlockBatch::iterator PreloadFileProcessor::SkipStateBlocks(uint64_t frame_number, BlockBatch::iterator start)
```
jzulauf-lunarg, line 223: Made this static, as the functionality as I didn't want to impose policy on replay cursor managment.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -51,7 +51,7 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateXlibSurfaceKHR(
     StructPointerDecoder<Decoded_VkAllocationCallbacks>*      pAllocator,
     HandlePointerDecoder<VkSurfaceKHR>*                       pSurface)
 {
-    if (frame_loop_info_.IsLooping())
+    if (frame_loop_info_.IsRepetition())
```
jzulauf-lunarg, line 54: Since all the comments talk about "repeating a frame" thought using "Repetition" clearer for the next maintainer (given I missed it), also isolates it from `Application` needed for state.

framework/graphics/frame_loop_info.h
jzulauf-lunarg: FirstIteration is unused. c.f. IsLooping and IsRepetetition

framework/graphics/frame_loop_info.h
```diff
@@ -40,35 +40,36 @@ class FrameLoopInfo
...
 
     void     SetLooping(bool looping) { is_looping_ = looping; }
-    uint32_t GetLoopFrameIdx() const { return loop_frame_idx_; }
     uint32_t GetLoopIterations() const { return loop_iterations_; }
 
     /// Decrements the number of loop iterations remaining.
     /// If the number of iterations is infinite, this has no effect.
     void DecrementLoopIterations()
     {
+        is_repetition_ = true;
```
jzulauf-lunarg, line 63: @antonio-lunarg  -- doing this implicitly.  Could add an SetRepetition to set explicitly if you prefer.

jzulauf-lunarg (commented):
Mostly what this change does is minimize the changes to PreloadFileProcessor that are causing rebase grief without folding these changes into another PR as a weird prefix to the PR

framework/decode/preload_file_processor.cpp
```diff
@@ -182,72 +189,58 @@ bool PreloadFileProcessor::ProcessNextFrame()
...
 
     const bool at_end = (!replay_cursor_);
+    if (loop_replay_)
+    {
+        // We're resetting the command stream, wait for all referenced resouces
+        // to be non-busy
+        WaitDecodersIdle();
+        replay_cursor_ = loop_reset_point_;
+        // We keep going as long we get the expected result
+        return process_result != ProcessBlockState::kError;
```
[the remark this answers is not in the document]
antonio-lunarg: Yeah, should this return false for `kEndProcessing` as well, @jzulauf-lunarg?

antonio-lunarg (approved):

jzulauf-lunarg:
@mikes-lunarg @jzulauf-lunarg @MarkY-LunarG @fabian-lunarg ... I've got this and another small cleanup to commit that are to enable my async work to rebase.  and it's gating #2871

framework/decode/preload_file_processor.cpp
bradgrantham-lunarg: I think my concern with #2698 was that whether things are looping or not is an application choice (maybe `Application`, re-reading my comment).  Does FileProcessor really need to have a "looping" primitive?  This is probably okay for now, but I think I much prefer a design where FileProcessor is always going forward, and a loop is created by transforming a frame's blocks into a "loopable" range of blocks, and then something above FileProcessor repeatedly plays that range of blocks.  I'll approve this so it can get in and stop blocking, but that was my preference.

bradgrantham-lunarg (approved):

PR #2871, opened 2026-04-14
Move resource create/destroy ouf of frame loop

davidlunarg:


davidlunarg:
Still testing this branch, but putting it out here for review and feedback.

davidlunarg:
Forced pushed the branch, with all my commits squashed into one commit and rebased on top of dev.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -27,236 +27,233 @@
...
-    const ApiCallInfo&                                         call_info,
-    VkResult                                                   returnValue,
-    format::HandleId                                           instance,
-    StructPointerDecoder<Decoded_VkWin32SurfaceCreateInfoKHR>* pCreateInfo,
-    StructPointerDecoder<Decoded_VkAllocationCallbacks>*       pAllocator,
-    HandlePointerDecoder<VkSurfaceKHR>*                        pSurface)
-{
-    if (frame_loop_info_.IsLooping())
+    // Skip all resource destroy commands in first and subsequent iterations.
+    if (skipDestroy.contains(call_id) && (frame_loop_info->IsFirstIteration() || frame_loop_info->IsLooping()))
```
antonio-lunarg, line 238: Once https://github.com/LunarG/gfxreconstruct/pull/2870 gets merged, you can drop `frame_loop_info->IsFirstIteration()` from this check as `frame_loop_info->IsLooping()` will include the first iteration.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -27,236 +27,233 @@
...
-    const ApiCallInfo&                                         call_info,
-    VkResult                                                   returnValue,
-    format::HandleId                                           instance,
-    StructPointerDecoder<Decoded_VkMetalSurfaceCreateInfoEXT>* pCreateInfo,
-    StructPointerDecoder<Decoded_VkAllocationCallbacks>*       pAllocator,
-    HandlePointerDecoder<VkSurfaceKHR>*                        pSurface)
-{
-    if (frame_loop_info_.IsLooping())
+    // Don't skip if we're not in the first iteration or loop.
+    if (!frame_loop_info->IsLooping())
```
antonio-lunarg, line 244: This will become `!frame_loop_info->IsRepetition()`

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -27,236 +27,233 @@
...
-        // When repeating a frame, the surface has already been created during the first iteration of the frame.
-        return;
+        return false;
     }
 
-    VulkanReplayConsumer::Process_vkCreateMetalSurfaceEXT(
-        call_info, returnValue, instance, pCreateInfo, pAllocator, pSurface);
-}
+    // Skip resource create commands after the first loop iteration.
+    if (!frame_loop_info->IsFirstIteration() && skipAllocate.contains(call_id))
```
antonio-lunarg, line 250: This `frame_loop_info->IsRepetition()`

jzulauf-lunarg:
I got a +1 from Antonio, just need one of the owner group to +1

davidlunarg:
> I got a +1 from Antonio, just need one of the owner group to +1

You mean you got a +1 on https://github.com/LunarG/gfxreconstruct/pull/2870, right?

jzulauf-lunarg:
@davidlunarg -- yes.

framework/application/application.h
```diff
@@ -99,6 +99,8 @@ class Application final
         return GFXRECON_NARROWING_CAST(uint32_t, file_processor_->GetCurrentFrameNumber());
     }
 
+    graphics::FrameLoopInfo* frame_loop_info_; ///< Indicates that playback wishes to loop a certain frame
```
antonio-lunarg: I think it would be better to have a `GetFrameLoopInfo() const` and keep this member private.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7673,7 +7673,7 @@ VkResult VulkanReplayConsumerBase::OverrideGetPhysicalDeviceSurfaceFormatsKHR(
 
     VkResult result = func(physical_device_info->handle, surface_info->handle, surface_format_count, surface_formats);
 
-    if (surface_formats != nullptr && result == VK_SUCCESS && *surface_format_count > 0)
+    if (surface_formats != nullptr)
```
antonio-lunarg: Is this change unrelated to this PR?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7696,7 +7696,7 @@ VkResult VulkanReplayConsumerBase::OverrideGetPhysicalDeviceSurfaceFormats2KHR(
     VkResult result =
         func(physical_device_info->handle, surface_info->GetPointer(), surface_format_count, surface_formats);
 
-    if (surface_formats != nullptr && result == VK_SUCCESS && *surface_format_count > 0)
+    if (surface_formats != nullptr)
```
antonio-lunarg: Unrelated as well?

framework/application/application.h
davidlunarg: Fixed in de1dbb3e42990447148460f3b716dc1b32170d64

framework/decode/vulkan_replay_consumer_base.cpp
davidlunarg: I'm not sure why this change is in this PR. Removed it in de1dbb3e42990447148460f3b716dc1b32170d64

framework/decode/vulkan_replay_consumer_base.cpp
davidlunarg: I'm not sure why this change is in this PR. Removed it in https://github.com/LunarG/gfxreconstruct/commit/de1dbb3e42990447148460f3b716dc1b32170d64

davidlunarg:
I had a bad version of Vulkan-Headers, so I rebased from latest dev, squashed all commits, and force pushed it.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -27,236 +27,233 @@
...
+// replay consumer.
 
-void VulkanReplayFrameLoopConsumer::Process_vkCreateMetalSurfaceEXT(
-    const ApiCallInfo&                                         call_info,
-    VkResult                                                   returnValue,
-    format::HandleId                                           instance,
-    StructPointerDecoder<Decoded_VkMetalSurfaceCreateInfoEXT>* pCreateInfo,
-    StructPointerDecoder<Decoded_VkAllocationCallbacks>*       pAllocator,
-    HandlePointerDecoder<VkSurfaceKHR>*                        pSurface)
+bool skip_while_looping(const graphics::FrameLoopInfo*    frame_loop_info,
```
bradgrantham-lunarg, line 40: Should this be `static`?

bradgrantham-lunarg: No, this is called from other compilation units.

bradgrantham-lunarg:
It looks to me like this change explicitly puts information and control based on looping into the non-looping VulkanReplayConsumer.  This seems like an inversion of how the responsibility should be expressed.  If a reader looks in the generated VulkanReplayConsumer, I think it will be surprising to find every function also checks to see if the function should be skipped.  I think the reader will wonder why there is also a separate VulkanReplayFrameLoopConsumer.  I think I'd prefer that VulkanReplayFrameLoopConsumer be generated and derived from VulkanReplayConsumer (which is already is) and have its members that could be skipped check the skip and then don't call the parent class method or call the parent class function if not to skip.  And anything that would never be skipped just defaults to the VulkanReplayConsumer method.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
davidlunarg: If it was static, would not be able to call it from VulkanReplayConsumer. It also could not be a member of the VulkanReplayFrameLoopConsumer class because it is called from where an instance of that class is not available.

davidlunarg:
> It looks to me like this change explicitly puts information and control based on looping into the non-looping VulkanReplayConsumer. This seems like an inversion of how the responsibility should be expressed. If a reader looks in the generated VulkanReplayConsumer, I think it will be surprising to find every function also checks to see if the function should be skipped. I think the reader will wonder why there is also a separate VulkanReplayFrameLoopConsumer. I think I'd prefer that VulkanReplayFrameLoopConsumer be generated and derived from VulkanReplayConsumer (which is already is) and have its members that could be skipped check the skip and then don't call the parent class method or call the parent class function if not to skip. And anything that would never be skipped just defaults to the VulkanReplayConsumer method.

There are many entry points that need to check skip. Rather than write many functions as was done in the original implementation, I thought I'd simply have a list of entry points and have one place where the list is checked. I was trying to avoid "bloat" - creating lots of code, all about the same, that needs to be compiled and adds to the size of the executable.

davidlunarg:
Abandoning this PR, functionality has been implemented in https://github.com/LunarG/gfxreconstruct/pull/2967.

PR #2876, opened 2026-04-16
Better check in GetPhysicalDeviceSurfaceFormats(2)KHR

fabian-lunarg:
- avoid initializing an std::optional when no information was returned
- attempts to correct wrong format-fallback-logic based on this std::optional


MarkY-LunarG (approved):
LGTM

PR #2877, opened 2026-04-16
Fix RGB8 format handling in trim/dumpresources

bartosz-muszarski-arm:
Current behavior
* image init offset logic is split across multiple paths and not consistently applied per VkBufferImageCopy region
* one path relied on a power of two alignment helper which is incorrect for some vulkan formats
* copy size command could use values that did not always map to actual tightly packed copy bytes
* this could produce invalid buffer offsets for some formats and lead to validation errors and corrupted image initialization

This change
* adds shared Vulkan copy layout helpers
 * AlignBufferOffset with modulo based align up valid for any positive alignment
 * GetBufferImageCopyOffsetAlignment covering depth stencil multiplane and regular format rules
* switches decode and replay init image offset construction to use the shared helpers
* updates copy region generation to align every region offset not only the initial staging offset
* updates image sizes and offsets to compute payload bytes from buffer image copy rules and produce aligned subresource offsets
* adds a fallback path for traces captured with previous implementation

Affects trimming and dump resource features

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -1237,6 +1237,11 @@ void VulkanReplayConsumerBase::ProcessInitImageCommand(format::HandleId         
                     copy_region.imageSubresource.baseArrayLayer = 0;
                     copy_region.imageSubresource.layerCount     = image_info->layer_count;
 
+                    const VkDeviceSize copy_alignment =
+                        graphics::GetBufferImageCopyOffsetAlignment(image_info->format, aspect);
+                    VkDeviceSize current_offset = 0;
+                    VkDeviceSize required_size  = 0;
+
                     assert(image_info->level_count == level_sizes.size());
 
                     for (uint32_t i = 0; i < image_info->level_count; ++i)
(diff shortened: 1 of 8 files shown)


fabian-lunarg:
fyi @panos-lunarg

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -1246,8 +1251,47 @@ void VulkanReplayConsumerBase::ProcessInitImageCommand(format::HandleId
                         copy_region.imageExtent.height        = std::max(1u, (image_info->extent.height >> i));
                         copy_region.imageExtent.depth         = std::max(1u, (image_info->extent.depth >> i));
 
+                        // Keep per-mip offsets valid even when level_sizes are tightly packed or
+                        // use non-power-of-two format alignments.
+                        //
+                        // For backward compatibility with older captures, this may be switched to
+                        // contiguous (legacy) offsets below when aligned offsets would exceed payload size.
+                        current_offset           = graphics::AlignBufferOffset(current_offset, copy_alignment);
```
fabian-lunarg, line 1259: the per-commit CI passed, but we 'might' have cases in extended CI that are affected.

fabian-lunarg (approved):
this is a great improvement over the previous situation.
the newly proposed logic is clean and sound, documentation too.

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: ran those additional checks, looking good.

PR #2888, opened 2026-04-21
logging: Ensure output flush of logging on close

jzulauf-lunarg:
Was seeing some unflushed stdout writes when stdout !isatty -- this change ensures that file based streams are flushed at log target destructor.


MarkY-LunarG (approved):
LGTM

Issue #2341, opened 2025-08-06
build error


zmike:
I just cloned to build, ran `cmake -G Ninja` in the build dir, and now I get this error on linux:

```
CMake Error: install(EXPORT "openxr_loader_export" ...) includes target "openxr_loader" which requires target "jsoncpp_interface" that is not in any export set.
CMake Error in external/OpenXR-SDK/src/loader/CMakeLists.txt:
  export called with target "openxr_loader" which requires target
  "jsoncpp_interface" that is not in any export set.
```

on 3c84d987d3327cbe0c1dc91a042507f38782e9c0

zmike:
Also toggling `BUILD_LOADER` to OFF breaks the build.

fabian-lunarg:
can confirm `BUILD_LOADER` OFF yields:

```
CMake Error at framework/decode/CMakeLists.txt:29 (add_library):
   Target "gfxrecon_decode" links to target "OpenXR::openxr_loader" but the
   target was not found.  Perhaps a find_package() call is missing for an
   IMPORTED target, or an ALIAS target is missing?
```
regular build with `cmake .. -GNinja && ninja` works for me though

zmike:
I can build the OpenXR SDK standalone without issues.

zmike:
I guess disabling XR support works around this for now.

PR #2889, opened 2026-04-22
Workaround OpenXR FindJsonCpp.cmake

mikes-lunarg:
When using system jsoncpp, OpenXR's fallback finder creates a local INTERFACE target that cannot be exported, causing CMake errors during install(EXPORT ...) operations.

So just use the vendored copy include with OpenXR-SDK instead...


mikes-lunarg:
I just ran into this while bringing up a new Arch Linux install

Fixes #2341

MarkY-LunarG (approved):
Cool.  Thanks for fixing Mike

PR #2894, opened 2026-04-24
Disable descriptor buffer capture replay on certain ARM driver version

tomhog-arm:
We have an issue with ARM drivers which support VK_EXT_descriptor_buffer. Actual descriptor buffers work however there is a driver bug with the descriptor buffer capture replay feature.

When descriptor buffer usage is detected in GFXR the capture replay feature is auto enabled and this causes a crash when any image view is created during replay.

This fix disabled capture replay on the effected driver version. There is a fix in upcoming driver version. 

I couldn't find any other examples of driver workarounds so am very open to suggestions on how this can be handled if this PR is not suitable.


fabian-lunarg:
> This fix disabled capture replay on the effected driver version

just curious about the benefit of this workaround. 

disabling `VK_EXT_descriptor_buffer` 's capture-replay feature sidesteps a driver-bug,
but without that you still could not replay any capture using `VK_EXT_descriptor_buffer` ? 
esp. given we haven't implemented any `-m rebind` support for the extension (and prob. never will, given there is a successor).

tomhog-arm:
Hey @fabian-lunarg 

That's a good point, I did this just to get it to stop crashing but in my basic testing it appeared the replay was successful but as mentioned that shouldn't be the case. I'll try it again and see what actually happened in this instance.

We may move toward a layer that disables the extension completely on particular driver versions as it's more flexible for us and doesn't pollute the gfxr code with work arounds.

fabian-lunarg:
> We may move toward a layer that disables the extension completely on particular driver versions as it's more flexible for us

sounds like a good approach here and more generally useful.

fabian-lunarg:
> I did this just to get it to stop crashing but in my basic testing it appeared the replay was successful but as mentioned that shouldn't be the case

without the capture-replay feature enabled we'll skip injecting `VkOpaqueCaptureDescriptorDataCreateInfoEXT`
into create-calls for various resources, e.g. [here](https://github.com/LunarG/gfxreconstruct/blob/41c7f2d964544813df5988d9689189f8520b1e2e/framework/decode/vulkan_replay_consumer_base.cpp#L6409) 
they will be created fine. 
but this should fail at the point those resources are actually used/referenced in a descriptor-buffer.
the driver-specific descriptor-blobs won't match the buffer-content stored in the capture.

fabian-lunarg:
closing for above reasons:
- overriding which extensions are available/reported (and influence gfxr-codepath) could be achieved in a more flexible way using a layer, avoiding driver-specific workarounds in gfxr.
- in this specific case: applying the workaround could only shift a driver-crash to occur at a later time.

PR #2895, opened 2026-04-24
Fix virtual swapchain memory barriers

marius-pelegrin-arm:
The virtual swapchain mechanism needs to copy the capture time swapchain image (virtual image) to the replay time swpachain imae (swapchain image).

To do this, `VulkanVirtualSwapchain` needs to do the following:
- A layout transition of the virtual image for copy (source)
- A layout transition of the swapchain image for copy (destination)
- The actual copy
- A layout transition of the virtual image back to present layout
- A layout transition of the swapchain image back to present layout

This was done correctly, but the layout transition being done using `vkCmdPipelineBarrier`, it is important to specify the correct flags for the memory dependency not to introduce additional synchronization that could slow down the replay.

This copy is mainly synchronized by the calls the application has to do at capture time in order to call `vkQueuePresentKHR`. Thus some of the flags currently used in the `vkCmdPipelineBarrier`, that would be valid in a general copy out of context, are actually unnecessary and adding potential synchronization at replay time.

### DIFF
--- a/framework/decode/vulkan_virtual_swapchain.cpp
+++ b/framework/decode/vulkan_virtual_swapchain.cpp
@@ -887,17 +887,22 @@ VkResult VulkanVirtualSwapchain::QueuePresentKHR(VkResult                       
     };
 
     final_barrier_virtual_image               = initial_barrier_virtual_image;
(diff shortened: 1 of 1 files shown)


fabian-lunarg (approved):
agreed, this is a good improvement. it wasn't correct and we over-synced before

PR #2896, opened 2026-04-24
Fix fast-forwarding of `vkTransitionImageLayout[EXT]`

marius-pelegrin-arm:
Fix `VkImage` layout tracking for which `vkTransitionImageLayout[EXT]` calls were forgotten.


fabian-lunarg (approved):
nice

PR #2897, opened 2026-04-24
Cleanup extension querying in instance creation

marius-pelegrin-arm:
Modification of Vulkan instance extensions queried by the replayer is once again getting messy. This commit aims at cleaning things a bit:
- Removing the `is_forced` boolean strategy. If an extension will be querried by the replayer anyway, it's better to still check if it is supported and if it is querried, because there may be multiple edge cases to handle (like extension being faked)
- Move "if extension supported" checks to the "if supported extensions could be queried" block. Checking if an extension is available only makes sense when available extensions query was actually successful

The following bugs are fixed by this commit:
- `VK_EXT_debug_utils` extension not being properly faked if not supported by the replay device
- `VK_KHR_get_surface_capabilities2` added twice if already queried by the application


fabian-lunarg (approved):
the logic is cleaner and I agree with the two (edge-)cases that got fixed

PR #2898, opened 2026-04-24
Limit shader buffer-reference bookkeeping to address-replacement replay

bartosz-muszarski-arm:
Currently replayer will always scan for buffer references in shaders and setup for device address replacement in some code paths even if UseAddressReplacement is disabled.

This change adds guards to skip the bookkeeping when opaque device addresses are supported  avoiding unnecessary SPIR-V parsing and metadata population.


fabian-lunarg (approved):
guarding the spirv-parsing makes total sense, as it isn't required for regular replays, as you say.  I assume this is particularly relevant for benchmarking cases.

PR #2903, opened 2026-04-25
Add missing support for vkGetImageSubresourceLayout2 and vkCmdBindDescriptorSets2

per-mathisen-arm:
These two Vulkan 1.4 commands have support for their extension variants but lack similar support for their core command variants.

### DIFF
--- a/framework/encode/vulkan_state_tracker.h
+++ b/framework/encode/vulkan_state_tracker.h
@@ -511,9 +511,6 @@ class VulkanStateTracker
                                     uint32_t               dynamicOffsetCount,
                                     const uint32_t*        pDynamicOffsets);
 
-    void TrackCmdBindDescriptorSets2KHR(VkCommandBuffer                    commandBuffer,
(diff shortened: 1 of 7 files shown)


framework/encode/vulkan_state_tracker.h
```diff
@@ -514,6 +514,9 @@ class VulkanStateTracker
     void TrackCmdBindDescriptorSets2KHR(VkCommandBuffer                    commandBuffer,
                                         const VkBindDescriptorSetsInfoKHR* pBindDescriptorSetsInfo);
```
fabian-lunarg: I think declaration for `TrackCmdBindDescriptorSets2KHR` can be dropped here, implementation exists only for `TrackCmdBindDescriptorSets2`

fabian-lunarg: done

fabian-lunarg (approved):

PR #2904, opened 2026-04-27
Fix dynamic rendering resource dumps

MartynJacques-Arm:
Prevent insertion of two EndRendering commands when dumping resources of dynamic rendering traces

Use Begin/EndRendering aliases for internal calls when one is available but the other isn't

### DIFF
--- a/framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
+++ b/framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
@@ -157,6 +157,55 @@ void DrawCallsDumpingContext::Release()
     current_cb_index_   = 0;
 }
 
+PFN_vkCmdBeginRendering DrawCallsDumpingContext::ResolveCmdBeginRendering() const
+{
+    if (device_table_->CmdBeginRendering != graphics::noop::vkCmdBeginRendering)
+    {
+        return device_table_->CmdBeginRendering;
+    }
+
+    if (device_table_->CmdBeginRenderingKHR != graphics::noop::vkCmdBeginRenderingKHR)
+    {
+        return device_table_->CmdBeginRenderingKHR;
+    }
+
+    return nullptr;
+}
+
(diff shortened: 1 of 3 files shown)


framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
```diff
@@ -156,6 +156,59 @@ void DrawCallsDumpingContext::Release()
     current_cb_index_   = 0;
 }
 
+PFN_vkCmdBeginRendering DrawCallsDumpingContext::ResolveCmdBeginRendering() const
+{
+    const VulkanDeviceInfo* device_info = object_info_table_.GetVkDeviceInfo(original_command_buffer_info_->parent_id);
+
+    auto cmd_begin_rendering = reinterpret_cast<PFN_vkCmdBeginRendering>(
+        device_table_->GetDeviceProcAddr(device_info->handle, "vkCmdBeginRendering"));
```
fabian-lunarg: what is the reason to use `device_table_->GetDeviceProcAddr` here?
doesn't `device_table_` already contain the fn-pointer?

MartynJacques-Arm: It does indeed. Thanks for pointing that out. Done.

fabian-lunarg (approved):
I agree with the fix (adding an else-condition).
the CmdBegin/EndRendering wrappers look a bit verbose but also easy to read, so fine with me.

fabian-lunarg:
@MartynJacques-Arm heej! the PR is good to go from my side, but apparently our CLA was updated in the meantime. if you could sign the new CLA, that would be cool. thank you!

MartynJacques-Arm:
Thanks @fabian-lunarg, done

MartynJacques-Arm:
Apologies, I've had to move this into a draft whilst I wait for some internal approvals. I'll let you know asap when this is done.

dnlbaines:
The CLA is approved now so this is unblocked from our side.

PR #2905, opened 2026-04-27
Fix external Vulkan layers for Android replay

tom-farmer-arm:
The injectLayers property was in the wrong part of the manifests.


mikes-lunarg (approved):
Good catch! These should in fact be application-level tags
https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/core/java/android/os/GraphicsEnvironment.java;l=84-90

PR #2908, opened 2026-04-27
break format check out of other builds

bradgrantham-lunarg:


.github/workflows/ci_build.yml
mikes-lunarg: you can remove clang-format-14 from this list

.github/workflows/ci_build.yml
```diff
@@ -145,9 +171,10 @@ jobs:
       with:
         submodules: 'recursive'
     - name: Check generated code
+      if: matrix.config.type == 'debug'
       run: |
         python3 framework/generated/generate_dx12.py # check generated code isn't out of date
-        git diff --exit-code
+        git diff --exit-code || { echo "Generated code differs from branch contents"; exit 1; }
```
mikes-lunarg: I'm surprised this runs on windows. I think the default shell is Powershell on the windows runners

mikes-lunarg (approved):

.github/workflows/ci_build.yml
```diff
@@ -11,6 +11,36 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y clang-format-14
+    - name: Clone repository
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
```
charles-lunarg, line 25: Shouldn't this be 1? Do we need any git history in order to run clang-format?

FYI the default value of `fetch-depth` is 1, whereas 0 gets the entire repository. 

EDIT: I see now that this isn't code you wrote, just moved up. Apologies!
(Still, if you happen to know if we need to checkout the full history, I'm all ears.)

.github/workflows/ci_build.yml
bradgrantham-lunarg: moved to PowerShell syntax

.github/workflows/ci_build.yml
bradgrantham-lunarg: That `diff $BASE...HEAD` does a diff against the merge-base of $BASE (dev) and HEAD (the PR branch).  If you fetch-depth 1 then I think you wouldn't have any of the previous commits including the merge-base.

charles-lunarg: Interesting, I would of thought the clang-format check works by running it against the repo, and seeing if there are any uncommitted changes. "would of thought" meaning how I would conceive of implementing this check, not that it must be implemented that way.

bradgrantham-lunarg: I think that's what this does.  The first diff is --name-only for the PR branch against the merge-base with dev, so it creates a list of changed files, and then that list is filtered through a glob of source suffixes, then that list is clang-formatted in-place, then if a second `git diff` actually displays the diffs and fails if there are any.  That's what this is supposed to do - does it fail at any of that?

PR #2910, opened 2026-04-28
Use blit for virtual swapchain image copy

marius-pelegrin-arm:
Replace `vkCmdCopyImage` with `vkCmdBlitImage` for copying virtual swapchain images to actual swapchain images.
This is theoretically costless if the size of the virtual swapchain images matches the size of the actual swapchain images, and allows to support cases where the actual swapchain size is different than the virtual swapchain image size.

### DIFF
--- a/framework/decode/vulkan_virtual_swapchain.cpp
+++ b/framework/decode/vulkan_virtual_swapchain.cpp
@@ -1034,20 +1034,31 @@ VkResult VulkanVirtualSwapchain::QueuePresentKHR(VkResult                       
                                           &initial_barrier_swapchain_image);
 
         subresource.layerCount   = swapchain_info->image_array_layers;
-        VkExtent3D  image_extent = { std::min(swapchain_resources->actual_extent.width, swapchain_info->width),
-                                     std::min(swapchain_resources->actual_extent.height, swapchain_info->height),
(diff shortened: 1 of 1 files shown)


fabian-lunarg:
I'm fine with replacing vkCmdCopyImage with vkCmdBlitImage and agree with your assessment.

> allows to support cases where the actual swapchain size is different than the virtual swapchain image size

can you maybe give an example? 
I was thinking window-resizing during replay (which we atm do not support)
'could' make use of it (other missing things wrt. sync still prevent it though).

fabian-lunarg (approved):

marius-pelegrin-arm:
> can you maybe give an example? I was thinking window-resizing during replay (which we atm do not support) 'could' make use of it (other missing things wrt. sync still prevent it though

The first example that comes to my mind is for captures that use fullscreen surface when the replay screen has a different size

fabian-lunarg:
we ran into issues, basically `vkCmdBlitImage` might be used on non-graphics queues with this PR. reverting this for the moment and come back.

what's really missing was a check, if present queue family has VK_QUEUE_GRAPHICS_BIT.
otherwise fallback to pre-existing `vkCmdCopyImage`.

PR #2911, opened 2026-04-28
[Frame looping] Reset descriptor pools that are allocated from during the looping frame

nickdriscoll-lunarg:
This PR fixes the frame looping case where descriptor sets are allocated on the looping frame.

Descriptor pools that are allocated from are tracked, then reset after syncing with the device post vkQueuePresentKHR()

### DIFF
--- a/framework/decode/vulkan_replay_frame_loop_consumer.cpp
+++ b/framework/decode/vulkan_replay_frame_loop_consumer.cpp
@@ -325,7 +325,6 @@ void VulkanReplayFrameLoopConsumer::Process_vkQueuePresentKHR(
     StructPointerDecoder<Decoded_VkPresentInfoKHR>* pPresentInfo)
 {
     // Get device
-    Decoded_VkPresentInfoKHR* meta       = pPresentInfo->GetMetaStructPointer();
     CommonObjectInfoTable&    table      = GetObjectInfoTable();
     VulkanQueueInfo*          queue_info = table.GetVkQueueInfo(queue);
     VkDevice                  device     = queue_info->parent;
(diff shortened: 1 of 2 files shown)


framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -279,5 +279,49 @@ void VulkanReplayFrameLoopConsumer::ProcessCreateHardwareBufferCommand(
...
+void VulkanReplayFrameLoopConsumer::Process_vkQueuePresentKHR(
+    const ApiCallInfo&                              call_info,
+    VkResult                                        returnValue,
+    format::HandleId                                queue,
+    StructPointerDecoder<Decoded_VkPresentInfoKHR>* pPresentInfo)
+{
+    // Get device
+    Decoded_VkPresentInfoKHR* meta           = pPresentInfo->GetMetaStructPointer();
+    CommonObjectInfoTable&    table          = GetObjectInfoTable();
+    const auto                swapchain_info = table.GetVkSwapchainKHRInfo(meta->pSwapchains.GetPointer()[0]);
```
bradgrantham-lunarg: Instead of having to reach into swapchains, I suggest get the `VulkanQueueInfo` from `GetObjectInfoTable().GetVkQueueInfo(queue);` and then the device is `GetObjectInfoTable().GetVkDeviceInfo(queue_info->parent_id);` and the handle is then `device_info->handle`

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_replay_frame_loop_consumer.cpp
nickdriscoll-lunarg: Thanks, that's a little nicer, though I opted for `queue_info->parent` instead of calling `table.GetVkDeviceInfo()`

bradgrantham-lunarg (approved):

PR #2915, opened 2026-04-29
Reduce artifact retention

beau-lunarg:


ci/Jenkinsfile
```diff
@@ -38,15 +38,11 @@ pipeline {
                     def discarder = env.BRANCH_NAME == 'dev'
                         ? buildDiscarder(logRotator(
                             daysToKeepStr: '90',
-                            numToKeepStr: '100',
-                            artifactDaysToKeepStr: '90',
-                            artifactNumToKeepStr: '100'
+                            artifactDaysToKeepStr: '14'
```
lunarpapillo, line 41: This means that every branch will save all artifacts from all runs for 14 days, right?  So if a branch has 100 commits in a week, all will be saved?

Do we really need any more than one set of artifacts, from the latest run, on any given branch?

ci/Jenkinsfile
```diff
@@ -38,15 +38,11 @@ pipeline {
...
-                            artifactNumToKeepStr: '100'
+                            artifactDaysToKeepStr: '14'
                         ))
                         : buildDiscarder(logRotator(
-                            daysToKeepStr: '7',
-                            numToKeepStr: '5',
-                            artifactDaysToKeepStr: '7',
-                            artifactNumToKeepStr: '5'
+                            daysToKeepStr: '30',
+                            artifactDaysToKeepStr: '7'
```
lunarpapillo: Won't this make the problem worse, by increasing the number of logs kept by a factor of 4?

lunarpapillo (commented):
I'd like to see a quick set of numbers describing:
- how many branches we expect to have active at once
- on average, how many commits/builds are invoked per branch
- how much space does each commit logfie take
- how much space does each extended logfile take
- how many extended runs we expect to maintain on dev
- how many extended runs we expect to maintain on non-dev branches
- etc.

so we can get some idea of what our space requirements for GFXR are.

ci/Jenkinsfile
beau-lunarg: This is for dev only, not PRs

ci/Jenkinsfile
beau-lunarg: This will keep more logs for PRs, but logs aren't the issue.  It doesn't change the number of artifacts days to keep.  But maybe this is one place where num to keep makes more sense than days.  Will change to last 3 per PR

beau-lunarg:
> I'd like to see a quick set of numbers describing:
> 
>     * how many branches we expect to have active at once
> 
>     * on average, how many commits/builds are invoked per branch
> 
>     * how much space does each commit logfie take
> 
>     * how much space does each extended logfile take
> 
>     * how many extended runs we expect to maintain on dev
> 
>     * how many extended runs we expect to maintain on non-dev branches
> 
>     * etc.
> 
> 
> so we can get some idea of what our space requirements for GFXR are.

These numbers are just looking at artifacts, not logs:
I'm also basing this on waht is in Jenkins.  If we recently purged records they won't show up here

Right now dev has 30 builds in the last two weeks, so I would anticipate keeping 30 builds with artifacts for dev
We have 20 PRs open that have any builds against them, so with 3 per PR we would have 60 builds with artifacts across all PRs
Extended should only run once per day that there is a dev build, so max 14 builds with artifacts
Extended manual currently has 20 builds in the last 3 days.  THis is a bit of a "when it rains it pours" scenario.  Also, reducing to 3 or 4 days makes things dicey around the weekends (run a build on friday, don't ahve much time to look at artifacts the next week)  So count this as 40 builds with artifacts
Manual has 12 in the past month, 8 are pinned.  The other 4 are within the past 7 days, so count it as 4 for manual.

That gives ~150 builds total (30+60+14+40+4=148).  Again, the recent purge may have altered some of these numbers

beau-lunarg:
> I'd like to see a quick set of numbers describing:
> 
>     * how many branches we expect to have active at once
> 
>     * on average, how many commits/builds are invoked per branch
> 
>     * how much space does each commit logfie take
> 
>     * how much space does each extended logfile take
> 
>     * how many extended runs we expect to maintain on dev
> 
>     * how many extended runs we expect to maintain on non-dev branches
> 
>     * etc.
> 
> 
> so we can get some idea of what our space requirements for GFXR are.

For log sizes
I ran `find . -type f -name "*.log" -printf "%s\n" | awk '{sum+=$1; n++} END {print "avg:", (n?sum/n/1024/1024:0), "MB", "count:", n}'`
within `/home/lunarg/ci-servers/jenkins/jobs/gfxr-pipelines`
and got `avg: 1.12129 MB count: 2282`

For artifact sizes:
I ran `find . -type d -path "**/builds/*/archive" -exec du -sb {} + | awk '{sum+=$1; n++} END {print
"avg:", (n?sum/n/1024/1024:0), "MB", "builds:", n}'`
within `/home/lunarg/ci-servers/jenkins/jobs/gfxr-pipelines`
and got `avg: 1987.89 MB builds: 177`

Assuming there are 13 logs per build (2282 log records and 177 artifact records, 2282/177), means about 15Mb per build for logs (2282/177*1.12MB)

Assuming the above ~150 builds is correct, that gives ~300GB for artifacts and ~3GB for logs

These number will likely go up as we add more tests or cells within our test matrix.

lunarpapillo (approved):

bradgrantham-lunarg (approved):

PR #2920, opened 2026-05-02
Vdr validate requested descriptors

panos-lunarg:


bradgrantham-lunarg:
@panos-lunarg Any idea what the performance impact is of this, if any?  Maybe just a handful of VDR runs on one heavy trace with and without this change?

panos-lunarg:
> @panos-lunarg Any idea what the performance impact is of this, if any? Maybe just a handful of VDR runs on one heavy trace with and without this change?

The expensive part of the new code is when no descriptors are requested and we have to dump all of them. The requested descriptors are stored in a different struct/format than the referenced descriptors (all bound descriptors that are valid for the current pipeline and therefore are potentially accessible by the shaders). So for each draw call we create a new map of descriptors that need to be dumped. If there are user requested descriptors they are validated and copied into the new map. Otherwise the referenced descriptor map is traversed and copied into the new map.

I think for the average use case the performance overhead will be negligible. However titles with large number of bound descriptors (i.e. utilizing the bindless extensions) can have an impact especially if multiple draw calls are requested. I am not sure if we have such a capture file so I haven't tried this with a real-world example. However I constructed a test case/benchmark to simulate this: https://godbolt.org/z/od3PEM67v

On my desktop, for 10 descriptor sets, 1000 bindings per set and 1000 array elements per binding take (10 million in total) 1.8 second to run. The above test code only benchmarks the traversal and population of the new map and does not include the searches in the object info table that need to be done for image descriptors. On the other hand I think assuming the existence of 10 million bound descriptors is a bit of an exaggeration. 

Update:
I replaced the map that holds the descriptors for dumping with a vector. This should mitigate a bit the performance overhead of `CullDescriptors`. 
New version with vector instead of a map: https://godbolt.org/z/W3EYhfs7E This takes ~500 ms instead of the 1.8 second of the map version.
Claude suggests a method for avoid using the additional map/vector but I didn't like the way it is implemented. I'll see if I can come up with something better.

bradgrantham-lunarg (approved):

PR #2921, opened 2026-05-02
Export sample count in dumped image subresource data in output JSON

cmannett85:
The sample count for an image is being collected, just not written out - this change fixes that.


panos-lunarg:
This includes the sample count in each dumped sub resource which I think is a bit redundant. I created a separate PR which output the sample count in the image's information instead: https://github.com/LunarG/gfxreconstruct/pull/2922. Please take a look and see if it suites your needs.

This change will break our internal tests since it changes the output format. I'll try to take care of this

cmannett85:
@panos-lunarg yours is much better - I'll close this one.

PR #2922, opened 2026-05-03
Vdr include sample count in output json

panos-lunarg:


framework/decode/vulkan_replay_dump_resources_json.cpp
```diff
@@ -250,6 +250,26 @@ void VulkanReplayDumpResourcesJson::InsertBufferInfo(nlohmann::ordered_json& jso
     }
 }
 
+void VulkanReplayDumpResourcesJson::InsertImageInfo(nlohmann::ordered_json& json_entry,
```
bradgrantham-lunarg, line 243: Should `InsertImageInfo` just take a `const VulkanImageInfo&`?  It's only ever called by dereferencing all the members of one of those.

framework/decode/vulkan_replay_dump_resources_delegate_dumped_resources.h
```diff
@@ -595,18 +596,18 @@ struct TransferedImageInfo
...
+    TransferedImageInfo(const TransferedImageInfo& other, VkImageLayout l) : image_info(other.image_info), layout(l)
+    {
+        GFXRECON_ASSERT(other.image_info != nullptr);
+    }
 
-    format::HandleId id;
-    VkFormat         format;
-    VkExtent3D       extent;
-    VkImageLayout    layout;
+    const VulkanImageInfo* image_info;
```
bradgrantham-lunarg: This should probably be non-pointer initialized from `const VulkanImageInfo&` so that there's never a question of use-after-free and readers won't wonder if there needs to be ownership and why it's a pointer.  You already require it in the constructor so if you made it a plain member you could also skip the assert != nullptr.

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_replay_dump_resources_delegate_dumped_resources.h
panos-lunarg: done

framework/decode/vulkan_replay_dump_resources_json.cpp
panos-lunarg: done

bradgrantham-lunarg (approved):

PR #2925, opened 2026-05-05
android: Update gradle to 9.2.0

antonio-lunarg:


MarkY-LunarG:
Why are we using 9.4.1 in one place (which only came out a few months ago) and 9.2.0 in another (which is still also pretty new).  They seem a little too recent, unless we have a hard reason for it.

antonio-lunarg:
> @MarkY-LunarG  Why are we using 9.4.1 in one place (which only came out a few months ago) and 9.2.0 in another (which is still also pretty new). They seem a little too recent, unless we have a hard reason for it.

To give some context: I try to keep my Android Studio updated, and sometimes it suggests me to update the gradle version of gfxr. When that happens I submit a PR. In other words, there's no hard reason for that, other than applying an Android Studio suggestion, but I must admit that when I do I usually do not have issues with debugging it on Android.

antonio-lunarg:
> @MarkY-LunarG Why are we using 9.4.1 in one place (which only came out a few months ago) and 9.2.0 in another (which is still also pretty new). They seem a little too recent, unless we have a hard reason for it.

About the versions:
- 9.4.1 is the Gradle version
- 9.2.0 is the Android Gradle Plugin version

bradgrantham-lunarg:
Does gradle automatically pull down these changes?  Will downstream builders of GFXR have to do anything?

antonio-lunarg:
> @bradgrantham-lunarg  Does gradle automatically pull down these changes? Will downstream builders of GFXR have to do anything?

Yes, it's completely transparent. `./gradlew` will download the new dependencies automatically.

bradgrantham-lunarg (approved):

PR #2927, opened 2026-05-05
Split Replay into Features for easier expansion

MarkY-LunarG:
This change continues on the Split previously done on both the `gfxrrecon-info` and the `gfxrecon-convert` tools to attempt to make it easier to expand new APIs into the tools without conflicting with existing changes.


tools/replay/android_main.cpp
```diff
@@ -152,118 +165,102 @@ void android_main(struct android_app* app)
...
+                bool        flush_inside_measurement_range     = false;
+                bool        preload_measurement_frame_range    = false;
+                std::string measurement_file_name;
 
-                gfxrecon::decode::VulkanDecoder vulkan_decoder;
+                auto application = std::make_shared<gfxrecon::application::Application>(
+                    kApplicationName, g_file_processor.get(), VK_KHR_ANDROID_SURFACE_EXTENSION_NAME, app);
 
-                if (replay_options.capture)
+                for (auto& feature : g_features)
```
mikes-lunarg: There is a non-trivial amount of logic duplicated between desktop_main.cpp and android_main.cpp, we should definitely pull that into a common implementation

tools/replay/android_main.cpp
```diff
@@ -152,118 +165,102 @@ void android_main(struct android_app* app)
...
-                if (IsRunPreProcessConsumer(api_replay_options))
+                for (auto& feature : g_features)
                 {
-                    RunPreProcessConsumer(filename, api_replay_options, api_replay_consumer);
+                    feature->SetMeasurementStartFrame(measurement_start_frame);
+                    feature->QueryFpsInfoOptions(quit_after_measurement_frame_range,
+                                                 flush_measurement_frame_range,
+                                                 flush_inside_measurement_range,
+                                                 preload_measurement_frame_range,
+                                                 quit_after_frame);
```
mikes-lunarg: I think it makes more sense for the replay main to parse the common options. Here, the different features duplicate work and could potentially stomp on each other.

tools/replay/android_main.cpp
```diff
@@ -152,118 +165,102 @@ void android_main(struct android_app* app)
...
+                                                     preload_measurement_frame_range,
                                                      measurement_file_name,
                                                      quit_after_frame,
                                                      quit_frame);
 
-                vulkan_replay_consumer->SetFatalErrorHandler(
-                    [](const char* message) { throw std::runtime_error(message); });
-                vulkan_replay_consumer->SetFpsInfo(&fps_info);
+                gfxrecon::graphics::FrameLoopInfo  fl_info;
+                gfxrecon::graphics::FrameLoopInfo* fl_info_ptr = nullptr;
```
mikes-lunarg: Seems a little odd to have a separate variable just for the fl_info pointer

tools/replay/android_main.cpp
```diff
@@ -152,118 +165,102 @@ void android_main(struct android_app* app)
...
+                    fl_info     = gfxrecon::graphics::FrameLoopInfo(loop_frame, loop_count);
+                    fl_info_ptr = &fl_info;
+                    application->SetFrameLoopInfo(fl_info_ptr);
+                }
 
-                vulkan_decoder.AddConsumer(vulkan_replay_consumer.get());
+                for (auto& feature : g_features)
+                {
+                    feature->CreateConsumer(g_file_processor.get(), application, fl_info_ptr);
+                    feature->SetAndroidApp(app);
```
mikes-lunarg: Would the android app handle be better encapsulated in the Application class, then the consumers that need it could get it from there without polluting the feature interface?

tools/replay/replay_vulkan_feature.cpp
```diff
@@ -0,0 +1,147 @@
...
+}
+
+void ReplayVulkanFeature::RegisterDecodeComponents(graphics::FpsInfo* fps_info)
+{
+    if (is_enabled_)
+    {
+        RegisterConsumerAndDecoder(fps_info);
+
+        file_processor_->SetPrintBlockInfoFlag(
+            replay_options_.enable_print_block_info, replay_options_.block_index_from, replay_options_.block_index_to);
```
mikes-lunarg, line 100: I think this can be moved out to replay main, it isn't vulkan specific

tools/replay/replay_vulkan_feature.cpp
```diff
@@ -0,0 +1,147 @@
...
+GFXRECON_BEGIN_NAMESPACE(replay)
+
+// Register this class as a feature in a module registry
+GFXR_UTIL_REGISTER_FEATURE_CREATOR(ReplayFeatureBase, ReplayVulkanFeature)
+
+void ReplayVulkanFeature::QueryOptions(util::ArgumentParser& arg_parser, const std::string& capture_filename)
+{
+    capture_filename_    = capture_filename;
+    replay_options_      = GetVulkanReplayOptions(arg_parser, capture_filename, &tracked_object_info_table_);
+    is_enabled_          = replay_options_.enable_vulkan;
```
mikes-lunarg: Can `is_enabled_` just be true here and `enable_vulkan_` removed from `replay_options_`? The vulkan options struct used to store the results of `WasVulkanAPIDetected` but now it seems like we are using the `is_enabled_` member of the feature class to do that.

Also I think /tools/replay/replay_pre_processing.h is a leftover.

tools/replay/replay_vulkan_feature.cpp
```diff
@@ -0,0 +1,147 @@
...
+                                            vulkan_recapture::dispatch_CreateInstance,
+                                            vulkan_recapture::dispatch_CreateDevice);
+    }
+}
+
+void ReplayVulkanFeature::CompletePreProcessingPass()
+{
+    if (needs_pre_processor_)
+    {
+        replay_options_.enable_vulkan = pre_processor_consumer_->WasVulkanAPIDetected();
```
mikes-lunarg: should this set `is_enabled_` instead?

mikes-lunarg (commented):
I can see why you didn't do this one first, lots to untangle here!

tools/replay/android_main.cpp
MarkY-LunarG: I was just making this change.  I have a pending change which pulls out a majority of the commonality between these two files.  I agree, having duplicate code in two separate files is prone to breakage.  I'll push it up when I address the other comments.

tools/replay/android_main.cpp
MarkY-LunarG: The problem is that each  API has it's own "ReplayOptions" structure that this info is stored inside.  For example, for Dx12, they use `DxReplayOptions` and Vulkan uses `VulkanReplayOptions` structures.

That's why it's a Feature call.

tools/replay/android_main.cpp
MarkY-LunarG: Good suggestion.

tools/replay/replay_vulkan_feature.cpp
MarkY-LunarG: `replay_options_` is Vulkan specific and even in the old path these were only set if Vulkan was enabled.  No change has been made to behavior here.

tools/replay/replay_vulkan_feature.cpp
MarkY-LunarG: Ah, good point.

tools/replay/android_main.cpp
MarkY-LunarG: @mikes-lunarg , the changes are in for the common code split

MarkY-LunarG:
@mikes-lunarg , I believe I've made all the changes I can based on your requests.  Can you review, especially my responses to your comments?

tools/replay/android_main.cpp
```diff
@@ -131,195 +99,25 @@ void android_main(struct android_app* app)
...
-
-                gfxrecon::decode::VulkanTrackedObjectInfoTable tracked_object_info_table;
-                gfxrecon::decode::VulkanReplayOptions          replay_options =
-                    GetVulkanReplayOptions(arg_parser, filename, &tracked_object_info_table);
-
-                std::unique_ptr<gfxrecon::decode::VulkanReplayConsumer> vulkan_replay_consumer;
+            auto make_application = [&](gfxrecon::decode::FileProcessor* fp) {
+                return std::make_shared<gfxrecon::application::Application>(
+                    kApplicationName, fp, VK_KHR_ANDROID_SURFACE_EXTENSION_NAME, app);
+            };
```
mikes-lunarg, line 117: I'm not thrilled about using a lambda for this but I guess you need to smuggle in app pointer somehow

tools/replay/android_main.cpp
mikes-lunarg: I agree that they currently all have their own options right now but was wondering if some of the common options should be factored out into a common option struct. It feels backwards to make all of the features parse the FPS info options when only the top-level application loop in the common code needs it.

mikes-lunarg (approved):
Great consolidation of the main loop, this was long overdue.

I think option handling is still the hardest part to make "pluggable". One thing I just realized is that `--help` is still hardcoded at the application level even though actual option parsing has been pushed into the features. I don't think it is critical, but long-term it would be nice is features could report their own usage flags.

PR #2930, opened 2026-05-07
VDR: Detect transfer secondaries from drawcall primaries

panos-lunarg:
This covers the case where a draw call primary dumps draw calls but also executes a secondary with transfers marked for dumping


dfriederich (approved):
Fixes the reported issue :-)

bradgrantham-lunarg (approved):

Issue #2868, opened 2026-04-13
[D3D12] Support raytracing opacity micromap for trimming


zongdu-arm:
The newly added ray tracing opacity micromap feature was not implemented after the AgilitySDK update.

We need to:

Support OMM structs in capture and replay (Already implemented in #2825)

Support OMM trimming

This issue is only used to implement the OMM trimming feature.

zongdu-arm:
Done

PR #2934, opened 2026-05-07
Support dx12 raytracing opacity micromap trimming

zongdu-arm:
Support dx12 raytracing opacity micromap trimming, track in #2868

DX12 AS trim metadata payloads now include serialized D3D12 build-input
structs to preserve OMM and OMM-array behavior:
  - Add a new meta id kInitDx12AccelerationStructureCommand, deprecate the old.
  - Encode and write full AS build inputs from state tracking and writer paths.
  - Decode and route structured build inputs through DX12 consumers.
  - Update DX12 utility and builder logic for OMM triangles and OMM-array
      input buffer entry handling.

### DIFF
--- a/framework/decode/block_parser.cpp
+++ b/framework/decode/block_parser.cpp
@@ -1276,8 +1276,13 @@ ParsedBlock& BlockParser::ParseMetaData(BlockBuffer& block_buffer)
                                  "Failed to read init subresource data meta-data block header");
         }
     }
-    else if (meta_data_type == format::MetaDataType::kInitDx12AccelerationStructureCommand)
+    else if (meta_data_type == format::MetaDataType::kInitDx12AccelerationStructureCommand_deprecated)
     {
+        GFXRECON_LOG_WARNING_ONCE(
+            "This capture contains a deprecated metacommand to create a Dx12AccelerationStructureCommand.  While still "
+            "supported, this "
+            "metacommand may not include new structure of the captured Dx12AccelerationStructureCommand.");
+
         // Parse command header.
         format::InitDx12AccelerationStructureCommandHeader header;
         success = block_buffer.Read(header.thread_id);
@@ -1287,14 +1292,15 @@ ParsedBlock& BlockParser::ParseMetaData(BlockBuffer& block_buffer)
         success = success && block_buffer.Read(header.inputs_type);
         success = success && block_buffer.Read(header.inputs_flags);
         success = success && block_buffer.Read(header.inputs_num_instance_descs);
-        success = success && block_buffer.Read(header.inputs_num_geometry_descs);
+        success = success && block_buffer.Read(header.inputs_geometry_descs_size);
         success = success && block_buffer.Read(header.inputs_data_size);
 
         // Parse geometry descs.
         std::vector<format::InitDx12AccelerationStructureGeometryDesc> geom_descs;
+        std::vector<uint8_t>                                           build_inputs_data;
         if (success)
         {
-            for (uint32_t i = 0; i < header.inputs_num_geometry_descs; ++i)
+            for (uint32_t i = 0; i < header.inputs_geometry_descs_size; ++i)
             {
                 format::InitDx12AccelerationStructureGeometryDesc geom_desc;
                 success = success && block_buffer.Read(geom_desc.geometry_type);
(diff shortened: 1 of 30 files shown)


zongdu-arm:
Hi @locke-lunarg, @bradgrantham-lunarg ,
Could you review this PR?

locke-lunarg:
The CI failed during the `replay` and `gfxrecon-info` tests for `D3D12RaytracingHelloWorld-priority-capture-1225-1235`. They got a error `[gfxrecon] ERROR - Failed to read init DX12 acceleration structure meta-data block header (frame 0 block 69)`.

I suspect the cause is the change to the trimming encoder, which would require generating new capture files for the tests to pass. Since we'd prefer not to update test assets, can we avoid modifying the encoder? or is this change necessary for the fix?

locke-lunarg:
```
ParsedBlock& BlockParser::ParseMetaData(BlockBuffer& block_buffer)
{
    ...
    else if (meta_data_type == format::MetaDataType::kInitDx12AccelerationStructureCommand)
    ...
```
Here is where the decoder is changed. The new `InitDx12AccelerationStructureCommandHeader` doesn't match the old capture files. The value of `header.inputs_size` is bad, like 188978561025, to cause failure.

Not only new `InitDx12AccelerationStructureCommandHeader`, I feel that the new parameter of    `Dx12DecoderBase::DispatchInitDx12AccelerationStructureCommand` might also cause failure for the old capture files.

locke-lunarg:
I feel that we might need a new meta data for this new solution, but keep the original meta data and original solution to be compatible with the old capture files. @bradgrantham What do you think?

zongdu-arm:
> I feel that we might need a new meta data for this new solution, but keep the original meta data and original solution to be compatible with the old capture files. @bradgrantham What do you think?

The structure data has been changed, so the old gfxr replay failed and the new FF gfxr must be re-captured.
Whether to add a new meta block ID is up to you, I can modify it at any time.

bradgrantham-lunarg:
It's important that existing captures continue to work, so this would need
to add a new revision of that metadata with new encode/decode.  (And
decoder for original metadata needs to persist)

Brad Grantham
Director of Technology
LunarG, Inc. - 3D Graphics Software Solutions
***@***.***

locke-lunarg:
I agree with @bradgrantham-lunarg . @zongdu-arm  please also adds a warning for the old meta data to recommend the user to generate a new capture file. Thanks!

zongdu-arm:
> I agree with @bradgrantham-lunarg . @zongdu-arm please also adds a warning for the old meta data to recommend the user to generate a new capture file. Thanks!

I added a new meta block ID. It is also compatible with older versions of GFXR trace.
@locke-lunarg , @bradgrantham-lunarg please review this again.

framework/decode/dx12_acceleration_structure_builder.h
```diff
@@ -41,10 +42,11 @@ class Dx12AccelerationStructureBuilder
...
-               const std::vector<format::InitDx12AccelerationStructureGeometryDesc>& init_geometry_descs,
-               const uint8_t*                                                        build_inputs_data);
+    void Build(const graphics::Dx12GpuVaMap&                                                       gpu_va_map,
+               const format::InitDx12AccelerationStructureCommandHeader&                           command_header,
+               const std::vector<format::InitDx12AccelerationStructureGeometryDesc>&               init_geometry_descs,
+               StructPointerDecoder<Decoded_D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS>* build_inputs,
+               const uint8_t*                                                                      build_inputs_data);
 
   private:
     void SetupBuild(const graphics::Dx12GpuVaMap&                                         gpu_va_map,
```
locke-lunarg: Please rename it to 'SetupBuild_deprecated' and replace all 'SetupBuild' and comment 'Deprecated. Please use SetupBuild2()'.

framework/format/format.h
locke-lunarg: Please rename it to 'kInitDx12AccelerationStructureCommand_deprecated' and replace all 'kInitDx12AccelerationStructureCommand' and comment 'Deprecated. Please use kInitDx12AccelerationStructureCommand2'.

framework/decode/block_parser.cpp
locke-lunarg, line 1280: Please add a warning here to recommend the user to generate a new capture file.

locke-lunarg:
@bradgrantham-lunarg This PR adds a new Meta Id. Our CI doesn't cover it. Should we add a new test file to our CI ? @zongdu-arm Do you know what sample code or title is good to test your PR?

zongdu-arm:
> @bradgrantham-lunarg This PR adds a new Meta Id. Our CI doesn't cover it. Should we add a new test file to our CI ? @zongdu-arm Do you know what sample code or title is good to test your PR?

I tested with 
https://github.com/microsoft/DirectX-Graphics-Samples/tree/master/Samples/Desktop/D3D12Raytracing/src/D3D12RaytracingOpacityMicromaps

framework/decode/dx12_acceleration_structure_builder.h
zongdu-arm: I have updated with you requested.

framework/format/format.h
zongdu-arm: I have updated with you requested.

framework/decode/block_parser.cpp
zongdu-arm: I have updated with you requested.

locke-lunarg:
@zongdu-arm Thank you for your patience. We're almost ready to merge. But I think the new functions and enum element shouldn't just take the original names. They should be kInitDx12AccelerationStructureCommand2, SetupBuild2(), GetAccelerationStructureInputsBufferEntries2() or something similar to avoid confusion. Thanks for understanding!

zongdu-arm:
@locke-lunarg ,
Sorry, I may not have understood your previous comment. I have updated it. Please review it again.

locke-lunarg (approved):

framework/encode/dx12_state_writer.cpp
```diff
@@ -1780,26 +1780,19 @@ void Dx12StateWriter::WriteAccelerationStructuresState(
...
-            }
-            else if (as_build.inputs.Type == D3D12_RAYTRACING_ACCELERATION_STRUCTURE_TYPE_TOP_LEVEL)
-            {
-                cmd.inputs_num_instance_descs = as_build.inputs.NumDescs;
-            }
-            else
-            {
-                GFXRECON_ASSERT(false && "Invalid D3D12_RAYTRACING_ACCELERATION_STRUCTURE_TYPE.");
-            }
+            // Store the serialized build-inputs struct size in inputs_num_geometry_descs.
```
bradgrantham-lunarg: Since there's a new command, I think inputs_num_geometry_descs should be geometry_descs_size or something, since it's bytes and not a count.  (Is that correct?)

bradgrantham-lunarg (changes requested):

framework/encode/dx12_state_writer.cpp
zongdu-arm: I have updated it.

bradgrantham-lunarg (approved):

PR #2935, opened 2026-05-07
[Frame Looping] Rework solution for looped descriptor set allocations

nickdriscoll-lunarg:
In my original solution, the code tracks which pools have been allocated from during the looping frame, and resets them after `vkQueuePresent()`. This solution has a flaw.

Example: Descriptor Pool 1 is used to allocate set A before the looping frame, then it allocates set B during the looping frame. Pool 1 will have been marked for reset because of set B's allocation, but when it is reset, set A goes along with it. If set A is used during the looping frame, this is invalid usage.

I realized that I was overthinking this, and we can simply skip allocating descriptor sets during repetitions of the looping frame in order to solve this issue in a simpler way, and without the flaw described above.

That's also to say, `Process_vkAllocateDescriptorSets()` can eventually just be generated like the rest of the resource allocating functions.

### DIFF
--- a/framework/decode/vulkan_replay_frame_loop_consumer.cpp
+++ b/framework/decode/vulkan_replay_frame_loop_consumer.cpp
@@ -65,8 +65,8 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateDescriptorPool(
 
     if (frame_loop_info_.IsRepetition())
     {
-        // Skip allocation of dangling descriptor pools
-        if (dangling_descriptor_pools_.contains(pool_id))
+        // Skip allocation of descriptor pools with a dangling creation
+        if (dangling_create_descriptor_pools_.contains(pool_id))
         {
             return;
         }
@@ -78,7 +78,8 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateDescriptorPool(
     if (frame_loop_info_.IsLooping() && !frame_loop_info_.IsRepetition())
     {
         // Gather descriptor pools that are created during the loop range
-        dangling_descriptor_pools_.insert(pool_id);
+        // We will delete any pools from this set that are destroyed during the loop range
+        dangling_create_descriptor_pools_.insert(pool_id);
     }
 }
 
@@ -90,7 +91,8 @@ void VulkanReplayFrameLoopConsumer::Process_vkDestroyDescriptorPool(
 {
     if (frame_loop_info_.IsRepetition() && !frame_loop_info_.IsFinalIteration())
     {
-        if (dangling_descriptor_pools_.contains(descriptorPool))
+        // Skip destruction of descriptor pools with a dangling destruction
+        if (dangling_destroy_descriptor_pools_.contains(descriptorPool))
         {
             return;
         }
@@ -98,29 +100,29 @@ void VulkanReplayFrameLoopConsumer::Process_vkDestroyDescriptorPool(
 
     if (frame_loop_info_.IsLooping() && !frame_loop_info_.IsRepetition())
     {
-        if (!dangling_descriptor_pools_.contains(descriptorPool))
+        if (!dangling_create_descriptor_pools_.contains(descriptorPool))
         {
             // If this pool was not created during the loop range, ignore destroying it.
-            dangling_descriptor_pools_.insert(descriptorPool);
+            dangling_destroy_descriptor_pools_.insert(descriptorPool);
             return;
         }
         else
         {
-            // Created and destroyed during the loop range
-            dangling_descriptor_pools_.erase(descriptorPool);
+            // Created and destroyed during the loop range, not dangling
+            dangling_create_descriptor_pools_.erase(descriptorPool);
 
             // Check if this was the pool for any heretofore dangling descriptors
-            RemovePoolDanglingDescriptors(descriptorPool);
+            RemovePoolDanglingCreateDescriptors(descriptorPool);
         }
     }
     VulkanReplayConsumer::Process_vkDestroyDescriptorPool(call_info, device, descriptorPool, pAllocator);
 }
 
-void VulkanReplayFrameLoopConsumer::RemovePoolDanglingDescriptors(format::HandleId descriptorPool)
+void VulkanReplayFrameLoopConsumer::RemovePoolDanglingCreateDescriptors(format::HandleId descriptorPool)
 {
     std::vector<format::HandleId> handles_to_delete;
-    handles_to_delete.reserve(dangling_descriptor_sets_.size());
-    for (format::HandleId handle : dangling_descriptor_sets_)
+    handles_to_delete.reserve(dangling_create_descriptor_sets_.size());
+    for (format::HandleId handle : dangling_create_descriptor_sets_)
     {
         VulkanDescriptorSetInfo* info = GetObjectInfoTable().GetVkDescriptorSetInfo(handle);
         if (info != nullptr && info->pool_id == descriptorPool)
@@ -130,7 +132,7 @@ void VulkanReplayFrameLoopConsumer::RemovePoolDanglingDescriptors(format::Handle
     }
     for (format::HandleId handle : handles_to_delete)
     {
-        dangling_descriptor_sets_.erase(handle);
+        dangling_create_descriptor_sets_.erase(handle);
     }
 }
 
@@ -143,7 +145,7 @@ void VulkanReplayFrameLoopConsumer::Process_vkResetDescriptorPool(const ApiCallI
     if (frame_loop_info_.IsLooping() && !frame_loop_info_.IsFinalIteration())
     {
         // If any of the sets in this pool are dangling, skip pool reset
-        for (format::HandleId set_id : dangling_descriptor_sets_)
+        for (format::HandleId set_id : dangling_create_descriptor_sets_)
         {
             VulkanDescriptorSetInfo* info = GetObjectInfoTable().GetVkDescriptorSetInfo(set_id);
             if (info != nullptr && info->pool_id == descriptorPool)
@@ -154,7 +156,10 @@ void VulkanReplayFrameLoopConsumer::Process_vkResetDescriptorPool(const ApiCallI
     }
 
     VulkanReplayConsumer::Process_vkResetDescriptorPool(call_info, returnValue, device, descriptorPool, flags);
-    RemovePoolDanglingDescriptors(descriptorPool);
+
+    // For pools that contain dangling descriptor sets, this code will only be reached once,
+    // during the final iteration of the loop range.
+    RemovePoolDanglingCreateDescriptors(descriptorPool);
 }
 
 void VulkanReplayFrameLoopConsumer::Process_vkAllocateDescriptorSets(
@@ -169,7 +174,7 @@ void VulkanReplayFrameLoopConsumer::Process_vkAllocateDescriptorSets(
         // Skip allocation of dangling descriptor sets
         for (format::HandleId set_handle : pDescriptorSets->GetSpan())
         {
-            if (dangling_descriptor_sets_.contains(set_handle))
+            if (dangling_create_descriptor_sets_.contains(set_handle))
             {
                 return;
             }
@@ -182,9 +187,10 @@ void VulkanReplayFrameLoopConsumer::Process_vkAllocateDescriptorSets(
     if (frame_loop_info_.IsLooping() && !frame_loop_info_.IsRepetition())
     {
         // During first iteration of looping range, record which descriptor sets are allocated
+        // They will be removed from the set if they are freed during the loop range
         for (format::HandleId set_handle : pDescriptorSets->GetSpan())
         {
-            dangling_descriptor_sets_.insert(set_handle);
+            dangling_create_descriptor_sets_.insert(set_handle);
         }
     }
 }
@@ -202,7 +208,7 @@ void VulkanReplayFrameLoopConsumer::Process_vkFreeDescriptorSets(const ApiCallIn
         // then we want to omit their destruction
         for (format::HandleId set_handle : pDescriptorSets->GetSpan())
         {
-            if (dangling_descriptor_sets_.contains(set_handle))
+            if (dangling_destroy_descriptor_sets_.contains(set_handle))
             {
                 return;
             }
@@ -213,15 +219,15 @@ void VulkanReplayFrameLoopConsumer::Process_vkFreeDescriptorSets(const ApiCallIn
     {
         for (format::HandleId set_handle : pDescriptorSets->GetSpan())
         {
-            if (dangling_descriptor_sets_.contains(set_handle))
+            if (dangling_create_descriptor_sets_.contains(set_handle))
             {
                 // Any descriptor set that was freed during the loop range is not dangling
-                dangling_descriptor_sets_.erase(set_handle);
+                dangling_create_descriptor_sets_.erase(set_handle);
             }
             else
             {
                 // Descriptor set freed during loop range but created before
-                dangling_descriptor_sets_.insert(set_handle);
+                dangling_destroy_descriptor_sets_.insert(set_handle);
                 return;
             }
         }
--- a/framework/decode/vulkan_replay_frame_loop_consumer.cpp
+++ b/framework/decode/vulkan_replay_frame_loop_consumer.cpp
@@ -234,6 +234,10 @@ void VulkanReplayFrameLoopConsumer::Process_vkFreeDescriptorSets(const ApiCallIn
         }
     }
 
+    // For pools that contain dangling descriptor sets, this code will only be reached once,
+    // during the final iteration of the loop range.
+    RemovePoolDanglingCreateDescriptors(descriptorPool);
+
     VulkanReplayConsumer::Process_vkFreeDescriptorSets(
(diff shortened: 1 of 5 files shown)


antonio-lunarg (approved):

antonio-lunarg:
I wonder.. what if descriptors are originally created and freed within the same frame?

I think we should have some way of tracking whether descriptors are still alive before skipping creation after the first iteration.

antonio-lunarg:
Same reasoning for "destroy" calls. If a frame uses a resource (that has been created in previous frames), we might want to skip any destroy call otherwise the next iteration of the loop would attempt to use a resource that has been destroyed.

nickdriscoll-lunarg:
So I think I have a new implementation that covers what you brought up:

Here are the possible cases for descriptor pool/set creation/destruction
 - Created before loop range but destroyed in the loop range
   - Omit destruction enitrely
 - Created before loop range but destroyed after loop range
   - No special handling
 - Created during loop range, destroyed after
   - Only create on first iteration
 - Created during loop range, destroyed during
   - No special handling

The solution here is to track all "dangling" pools/sets, then ignore their destruction entirely, and only create them on the first iteration of the loop range.

A "dangling" resource is one that was either
	- Created during the loop range but destroyed after it
	or
	- Created before the loop range but destroyed during it

Please let me know what you think @antonio-lunarg

framework/decode/vulkan_replay_frame_loop_consumer.h
```diff
@@ -148,20 +148,49 @@ class VulkanReplayFrameLoopConsumer : public VulkanReplayConsumer
...
                                    format::HandleId                                queue,
                                    StructPointerDecoder<Decoded_VkPresentInfoKHR>* pPresentInfo) override;
 
+  private:
+    void DeleteDanglingPoolDescriptorSets(format::HandleId descriptorPool);
+
   private:
     graphics::FrameLoopInfo&             frame_loop_info_;
-    std::unordered_set<VkDescriptorPool> active_descriptor_pools_;
+    std::unordered_set<format::HandleId> dangling_descriptor_pools_;
```
antonio-lunarg: Worth adding documentation on top of these fields:
```
/// A "dangling" resource is one that was either
/// - created during the loop range but destroyed after it
/// - or created before the loop range but destroyed during it
```

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -302,19 +303,191 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateCommandPool(
...
+    }
+}
+
+void VulkanReplayFrameLoopConsumer::Process_vkDestroyDescriptorPool(
+    const ApiCallInfo&                                   call_info,
+    format::HandleId                                     device,
+    format::HandleId                                     descriptorPool,
+    StructPointerDecoder<Decoded_VkAllocationCallbacks>* pAllocator)
+{
+    if (frame_loop_info_.IsRepetition())
```
antonio-lunarg: How about checking if this is the last iteration of the loop range? In this case we might want to actually destroy the resource.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -302,19 +303,191 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateCommandPool(
...
+                return;
+            }
+        }
+    }
+
+    if (frame_loop_info_.IsLooping() && !frame_loop_info_.IsRepetition())
+    {
+        // Reset descriptor pool means that the pool is still valid, but the allocated descriptor sets are not
+        // Check if this was the pool for any heretofore dangling descriptors
+        DeleteDanglingPoolDescriptorSets(descriptorPool);
```
antonio-lunarg: We might want to skip reset if any of the sets in this pool are dangling. That is probably required if the sets are going to be used again in the next iteration of the loop range.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -302,19 +303,191 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateCommandPool(
...
     const ApiCallInfo&                                         call_info,
     VkResult                                                   returnValue,
     format::HandleId                                           device,
     StructPointerDecoder<Decoded_VkDescriptorSetAllocateInfo>* pAllocateInfo,
     HandlePointerDecoder<VkDescriptorSet>*                     pDescriptorSets)
 {
+    if (frame_loop_info_.IsRepetition())
+    {
+        // Skip allocation of dangling descriptor sets
+        const format::HandleId* set_handles = pDescriptorSets->GetPointer();
```
antonio-lunarg: nit: Is there a `->GetSpan()` method to iterate `HandleId`s directly?

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -302,19 +303,191 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateCommandPool(
...
+}
+
+void VulkanReplayFrameLoopConsumer::Process_vkFreeDescriptorSets(const ApiCallInfo& call_info,
+                                                                 VkResult           returnValue,
+                                                                 format::HandleId   device,
+                                                                 format::HandleId   descriptorPool,
+                                                                 uint32_t           descriptorSetCount,
+                                                                 HandlePointerDecoder<VkDescriptorSet>* pDescriptorSets)
+{
+    if (frame_loop_info_.IsRepetition())
```
antonio-lunarg: We might still want to free them if this is the last iteration of the loop.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -302,19 +303,191 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateCommandPool(
...
+        for (int i = 0; i < descriptorSetCount; ++i)
+        {
+            if (!dangling_descriptor_sets_.contains(handles[i]))
+            {
+                return;
+            }
+        }
+    }
+
+    VulkanReplayConsumer::Process_vkFreeDescriptorSets(
```
antonio-lunarg: We might not want to free the descriptor sets on the first iteration of the loop, they might still be used in the next iteration of the loop.

antonio-lunarg (changes requested):
That's a great approach, I wonder if we would be able to unit test this logic in some way.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
nickdriscoll-lunarg: So funnily enough, `StructPointerDecoder` has a `GetSpan()` method, but `HandlePointerDecoder` does _not_, so I guess I'll add it.

nickdriscoll-lunarg: I tried adding GetSpan() to `HandlePointerDecoder` but am now battling against this completely nonsensical compiler error. Just posting it here because I have to step away from my desk for a bit and maybe the problem will be obvious to you.

```
framework\decode\handle_pointer_decoder.h(63,57): error C2440: 'return': cannot convert from 'std::span<c
       onst unsigned __int64,18446744073709551615>' to 'std::span<gfxrecon::format::HandleId,18446744073709551615>'
```

antonio-lunarg: Try cherry-picking this: https://github.com/antonio-lunarg/gfxreconstruct/commit/4ea49613dc9dc3f42f7fa51e459743252c6430ff

nickdriscoll-lunarg: Thank you. I can see I was missing a `const` in the type specifier.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
nickdriscoll-lunarg: I added `FrameLoopInfo::IsFinalIteration()` for this purpose.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
nickdriscoll-lunarg: I added `FrameLoopInfo::IsFinalIteration()` for this purpose.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -302,20 +303,177 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateCommandPool(
...
+{
+    if (frame_loop_info_.IsRepetition() && !frame_loop_info_.IsFinalIteration())
     {
-        VkDescriptorPool pool = pAllocateInfo->GetPointer()->descriptorPool;
-        active_descriptor_pools_.insert(pool);
+        // If any of the descriptor sets are _not_ in the dangling list,
+        // then we want to omit their destruction
+        for (format::HandleId set_handle : pDescriptorSets->GetSpan())
+        {
+            if (!dangling_descriptor_sets_.contains(set_handle))
```
antonio-lunarg: I think we should remove `!` here, right?
If it was allocated in this loop, free it.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -302,20 +303,177 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateCommandPool(
...
+            // Created and destroyed during the loop range
+            dangling_descriptor_pools_.erase(descriptorPool);
+
+            // Check if this was the pool for any heretofore dangling descriptors
+            std::vector<format::HandleId> handles_to_delete;
+            handles_to_delete.reserve(dangling_descriptor_sets_.size());
+            for (format::HandleId handle : dangling_descriptor_sets_)
+            {
+                VulkanDescriptorSetInfo* info = GetObjectInfoTable().GetVkDescriptorSetInfo(handle);
+                if (info->pool_id == descriptorPool)
```
antonio-lunarg: I would check info as well: `if (info != nullptr && ...)`

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -302,20 +303,177 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateCommandPool(
...
+        for (format::HandleId set_id : dangling_descriptor_sets_)
+        {
+            VulkanDescriptorSetInfo* info = GetObjectInfoTable().GetVkDescriptorSetInfo(set_id);
+            if (info->pool_id == descriptorPool)
+            {
+                return;
+            }
+        }
+    }
+
```
antonio-lunarg, line 157: `vkResetDescriptorPool` frees all sets in a pool, We should erase the affected sets from `dangling_descriptor_sets_`.

Also, reset should happen on final iteration.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -302,20 +303,177 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateCommandPool(
...
+            if (!dangling_descriptor_sets_.contains(set_handle))
+            {
+                return;
+            }
+        }
     }
+
+    if (frame_loop_info_.IsLooping() && !frame_loop_info_.IsRepetition())
+    {
+        bool any_dangling = false;
```
antonio-lunarg: Dead

framework/decode/vulkan_replay_frame_loop_consumer.cpp
antonio-lunarg: Or you might track the sets and their corresponding pool with: `std::unordered_map<(descriptorPool), std::vector<(descriptorSet)>>`

antonio-lunarg (commented):
It's starting to look good!

framework/decode/vulkan_replay_frame_loop_consumer.cpp
antonio-lunarg: It's a bit complicated, I am figuring this case:
- Descriptor sets freed during loop range but created before -> got inserted in the dangling list.
- In the middle of the loop range, we indeed find them in the dangling list, so we skip the return, and we Free them, which is a potential bug if the sets were used in the loop.

I am starting to consider two different containers:
- One for resources created during the loop (we do not want to recreate them)
- One for resources created before the loop and destroyed within (we do not want to destroy them - only on last iteration)

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -81,16 +252,6 @@ void VulkanReplayFrameLoopConsumer::Process_vkQueuePresentKHR(
     GFXRECON_ASSERT(device_table);
```
antonio-lunarg: Now unused.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -50,22 +53,190 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateCommandPool(
...
+            if (dangling_create_descriptor_sets_.contains(set_handle))
+            {
+                // Any descriptor set that was freed during the loop range is not dangling
+                dangling_create_descriptor_sets_.erase(set_handle);
+            }
+            else
+            {
+                // Descriptor set freed during loop range but created before
+                dangling_destroy_descriptor_sets_.insert(set_handle);
+                return;
```
antonio-lunarg: Inserting only the first set handle and then `return`. Do we want to check all handles before deciding to return?

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -50,22 +53,190 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateCommandPool(
...
+            }
+        }
+    }
+
+    VulkanReplayConsumer::Process_vkResetDescriptorPool(call_info, returnValue, device, descriptorPool, flags);
+
+    // For pools that contain dangling descriptor sets, this code will only be reached once,
+    // during the final iteration of the loop range.
+    RemovePoolDanglingCreateDescriptors(descriptorPool);
```
antonio-lunarg: Do we want to invert this line with the previous one?
`VulkanReplayConsumer::Process_vkResetDescriptorPool` is going to delete all `VkDescriptorSetInfo`, and `RemovePoolDanglingCreateDescriptors` is not going to find them anymore.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -50,22 +53,190 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateCommandPool(
...
 
+void VulkanReplayFrameLoopConsumer::Process_vkCreateDescriptorPool(
+    const ApiCallInfo&                                        call_info,
+    VkResult                                                  returnValue,
+    format::HandleId                                          device,
+    StructPointerDecoder<Decoded_VkDescriptorPoolCreateInfo>* pCreateInfo,
+    StructPointerDecoder<Decoded_VkAllocationCallbacks>*      pAllocator,
+    HandlePointerDecoder<VkDescriptorPool>*                   pDescriptorPool)
+{
+    format::HandleId pool_id = *pDescriptorPool->GetPointer();
```
antonio-lunarg, line 63: `GFXRECON_ASSERT(pDescriptorPool != nullptr && pDescriptorPool->GetPointer() != nullptr)`

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -20,6 +20,9 @@
 ** DEALINGS IN THE SOFTWARE.
 */
 
+#include "decode/vulkan_replay_frame_loop_consumer.h"
```
antonio-lunarg: That's already included.

antonio-lunarg (approved):
Once last comments are addressed, this is good to go IMO. 👍

framework/decode/vulkan_replay_frame_loop_consumer.cpp
nickdriscoll-lunarg: I'm actually thinking that call can just be removed, since it's only reached during the final iteration of the loop range anyway.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
nickdriscoll-lunarg: I changed this so that all handles are checked before returning, but I'm wondering if the actual proper way of handling this call is to compute the subset of descriptor sets that are not dangling and only pass that down to `VulkanReplayConsumer::Process_vkFreeDescriptorSets()`

antonio-lunarg: Possibly. If you think it's worth it, go ahead, but in any case you have my approval.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
antonio-lunarg: I think it makes sense, but it's better to be sure not to leak anything, just in case the logic changes in the future.

bradgrantham-lunarg (approved):

PR #2936, opened 2026-05-07
[Frame Looping] Manage fences that are signaled/waited-on during the looping frame

nickdriscoll-lunarg:
This PR is based on top of https://github.com/LunarG/gfxreconstruct/pull/2935 so it will need to be merged first.

In the case of naive looping, it is easy to see how a fence can be double-signaled or double-waited-on, so we need to handle those situations.

Regarding a fence that is used during the looping frame, there are three cases:

1. The fence is signaled as many times as it is waited upon
2. The fence is signaled more times than it is waited upon
3. The fence is signaled fewer times than it is waited upon

In the first case, no special handling is needed, as the final fence operation will put the fence in the state that the first fence operation expects.

In the second case, we need to synthetically wait-on the fence so that it's in the unsignaled state at the start of the next iteration of the looping frame.

In the third case, we need to synthetically signal the fence so that it's in the signaled state at the start of the next iteration of the looping frame.

### DIFF
--- a/framework/decode/vulkan_replay_frame_loop_consumer.cpp
+++ b/framework/decode/vulkan_replay_frame_loop_consumer.cpp
@@ -419,25 +419,6 @@ void VulkanReplayFrameLoopConsumer::FixupDeviceFences(format::HandleId device, f
     }
 }
 
-void VulkanReplayFrameLoopConsumer::RemovePoolDanglingCreateDescriptors(format::HandleId descriptorPool)
-{
-    std::vector<format::HandleId> handles_to_delete;
-    handles_to_delete.reserve(dangling_create_descriptor_sets_.size());
-    for (format::HandleId handle : dangling_create_descriptor_sets_)
-    {
-        VulkanDescriptorSetInfo* info = GetObjectInfoTable().GetVkDescriptorSetInfo(handle);
-        if (info->pool_id == descriptorPool)
-        {
-            handles_to_delete.push_back(handle);
-        }
-    }
-
-    for (format::HandleId handle : handles_to_delete)
-    {
-        dangling_create_descriptor_sets_.erase(handle);
-    }
-}
-
 void VulkanReplayFrameLoopConsumer::Process_vkQueuePresentKHR(
     const ApiCallInfo&                              call_info,
     VkResult                                        returnValue,
--- a/framework/decode/vulkan_replay_frame_loop_consumer.cpp
+++ b/framework/decode/vulkan_replay_frame_loop_consumer.cpp
@@ -420,15 +420,14 @@ void VulkanReplayFrameLoopConsumer::FixupDeviceFences(format::HandleId device, f
     }
 }
 
-void VulkanReplayFrameLoopConsumer::Process_vkAcquireNextImageKHR(
-    const ApiCallInfo&                          call_info,
-    VkResult                                    returnValue,
-    format::HandleId                            device,
-    format::HandleId                            swapchain,
-    uint64_t                                    timeout,
-    format::HandleId                            semaphore,
-    format::HandleId                            fence,
-    PointerDecoder<uint32_t>*                   pImageIndex)
+void VulkanReplayFrameLoopConsumer::Process_vkAcquireNextImageKHR(const ApiCallInfo&        call_info,
+                                                                  VkResult                  returnValue,
+                                                                  format::HandleId          device,
+                                                                  format::HandleId          swapchain,
+                                                                  uint64_t                  timeout,
+                                                                  format::HandleId          semaphore,
+                                                                  format::HandleId          fence,
+                                                                  PointerDecoder<uint32_t>* pImageIndex)
 {
     if (frame_loop_info_.IsLooping() && !frame_loop_info_.IsRepetition())
     {
--- a/framework/decode/vulkan_replay_frame_loop_consumer.cpp
+++ b/framework/decode/vulkan_replay_frame_loop_consumer.cpp
@@ -170,7 +170,10 @@ void VulkanReplayFrameLoopConsumer::Process_vkResetDescriptorPool(const ApiCallI
     }
 
     VulkanReplayConsumer::Process_vkResetDescriptorPool(call_info, returnValue, device, descriptorPool, flags);
-    RemovePoolDanglingDescriptors(descriptorPool);
+
+    // For pools that contain dangling descriptor sets, this code will only be reached once,
+    // during the final iteration of the loop range.
+    RemovePoolDanglingCreateDescriptors(descriptorPool);
 }
 
 void VulkanReplayFrameLoopConsumer::Process_vkAllocateDescriptorSets(
--- a/framework/decode/vulkan_replay_frame_loop_consumer.cpp
+++ b/framework/decode/vulkan_replay_frame_loop_consumer.cpp
@@ -170,6 +170,7 @@ void VulkanReplayFrameLoopConsumer::Process_vkResetDescriptorPool(const ApiCallI
     }
 
     VulkanReplayConsumer::Process_vkResetDescriptorPool(call_info, returnValue, device, descriptorPool, flags);
+    RemovePoolDanglingDescriptors(descriptorPool);
 }
 
 void VulkanReplayFrameLoopConsumer::Process_vkAllocateDescriptorSets(
--- a/framework/decode/vulkan_replay_frame_loop_consumer.cpp
+++ b/framework/decode/vulkan_replay_frame_loop_consumer.cpp
@@ -29,6 +29,20 @@
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
+#define CHECK_VK_RESULT(_res_, _func_)                                        \
+    {                                                                         \
+        VkResult _RES = (_res_);                                              \
+        if (_RES != VK_SUCCESS)                                               \
+        {                                                                     \
+            GFXRECON_LOG_ERROR("[%s:%u] synthetic call to %s failed with %s", \
+                               __FILE__,                                      \
+                               __LINE__,                                      \
+                               _func_,                                        \
+                               util::ToString(_RES).c_str());                 \
+            std::exit(-1);                                                    \
+        }                                                                     \
+    }
+
 void VulkanReplayFrameLoopConsumer::Process_vkCreateCommandPool(
     const ApiCallInfo&                                     call_info,
     VkResult                                               returnValue,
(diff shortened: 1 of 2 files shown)


framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -324,24 +387,85 @@ void VulkanReplayFrameLoopConsumer::Process_vkQueuePresentKHR(
...
-    VkDevice               device     = queue_info->parent;
-    GFXRECON_ASSERT(device);
-    const graphics::VulkanDeviceTable* device_table = GetDeviceTable(device);
-    GFXRECON_ASSERT(device_table);
-
     VulkanReplayConsumer::Process_vkQueuePresentKHR(call_info, returnValue, queue, pPresentInfo);
 
     if (frame_loop_info_.IsLooping())
     {
+        // Get device
```
antonio-lunarg, line 473: I think extracting this logic into a private helper function would increase the readability. Other than that, LGTM!

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -324,24 +387,85 @@ void VulkanReplayFrameLoopConsumer::Process_vkQueuePresentKHR(
...
+        GFXRECON_ASSERT(device_table);
+
+        GFXRECON_LOG_DEBUG("Waiting for device to idle...");
         device_table->DeviceWaitIdle(device);
-        for (VkDescriptorPool pool : active_descriptor_pools_)
+
+        // Gather fences that were signaled but not waited upon during this frame
+        std::vector<VkFence> manual_wait_fences;
+        manual_wait_fences.reserve(signaled_fences_.size());
+        for (std::pair<format::HandleId, uint32_t> pair : signaled_fences_)
```
antonio-lunarg: I would use `for (auto [fence_id, signal_count] : signaled_fences_)` here

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -324,24 +387,85 @@ void VulkanReplayFrameLoopConsumer::Process_vkQueuePresentKHR(
...
+            {
+                GFXRECON_LOG_DEBUG("Will synthetically wait on fence %" PRIu64, fence_id);
+                VulkanFenceInfo* fence_info = table.GetVkFenceInfo(fence_id);
+                manual_wait_fences.push_back(fence_info->handle);
+            }
+        }
+
+        std::vector<VkFence> manual_signal_fences;
+        manual_signal_fences.reserve(signaled_fences_.size());
+        for (std::pair<format::HandleId, uint32_t> pair : waited_upon_fences_)
```
antonio-lunarg: ditto

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -324,24 +387,85 @@ void VulkanReplayFrameLoopConsumer::Process_vkQueuePresentKHR(
...
+            if (wait_count > signal_count)
+            {
+                GFXRECON_LOG_DEBUG("Will synthetically signal fence %" PRIu64, fence_id);
+                VulkanFenceInfo* fence_info = table.GetVkFenceInfo(fence_id);
+                manual_signal_fences.push_back(fence_info->handle);
+            }
+        }
+
+        GFXRECON_LOG_DEBUG("Synthetically waiting on fences...");
+        device_table->WaitForFences(device,
```
antonio-lunarg: Need to check `VkResult`.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -324,24 +387,85 @@ void VulkanReplayFrameLoopConsumer::Process_vkQueuePresentKHR(
...
+                                    manual_wait_fences.size(),
+                                    manual_wait_fences.data(),
+                                    VK_TRUE,
+                                    std::numeric_limits<uint64_t>::max());
+        GFXRECON_LOG_DEBUG("Resetting synthetically waited on fences...");
+        device_table->ResetFences(device, manual_wait_fences.size(), manual_wait_fences.data());
```
antonio-lunarg: Check `VkResult`.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -324,24 +387,85 @@ void VulkanReplayFrameLoopConsumer::Process_vkQueuePresentKHR(
...
+
+        // Fences may have been waited on but not reset, so we reset the fences we're going
+        // to manually signal here just in case.
+        device_table->ResetFences(device, manual_signal_fences.size(), manual_signal_fences.data());
```
antonio-lunarg: `VkResult`.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -324,24 +387,85 @@ void VulkanReplayFrameLoopConsumer::Process_vkQueuePresentKHR(
...
+
+        GFXRECON_LOG_DEBUG("Synthetically signaling fences...");
+        for (VkFence fence : manual_signal_fences)
         {
-            device_table->ResetDescriptorPool(device, pool, 0);
+            VulkanQueueInfo* queue_info = table.GetVkQueueInfo(queue);
+            device_table->QueueSubmit(queue_info->handle, 0, nullptr, fence);
```
antonio-lunarg: `VkResult`.

antonio-lunarg (commented):
Mmh.. what if the collected fences belong to different devices?
Maybe it is not that common as a case, but the replay consumer is grouping various things per-device to avoid this kind of issue.

nickdriscoll-lunarg:
Good catch that my code assumes a single `VkDevice`. I should be able to extend it to handle multiple devices without too much suffering.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
nickdriscoll-lunarg: So this would be a helper function that takes the queue `HandleID` as a parameter, but then would it return the device table, the device handle, or both?

I'm not seeing the vision, and I would prefer for there to be a second call-site before such a private helper function is written.

antonio-lunarg: I mean the whole fence management code.

nickdriscoll-lunarg: Gotcha.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -309,12 +323,163 @@ void VulkanReplayFrameLoopConsumer::Process_vkAllocateDescriptorSets(
...
+        }
+    }
+
+    VulkanReplayConsumer::Process_vkQueueSubmit(call_info, returnValue, queue, submitCount, pSubmits, fence);
+}
+
+void VulkanReplayFrameLoopConsumer::FixupDeviceFences(format::HandleId device, format::HandleId queue)
+{
+    // Get fence tracking info associated with this VkDevice
+    GFXRECON_ASSERT(per_device_fence_tracking_.contains(device));
```
antonio-lunarg: There might be cases where this could not contain any value associated to `device`.
Imagine a frame with no fences in queue submissions and no wait for fences calls.
Better to just check and return if `contains()` returns `false`.

antonio-lunarg (approved):
Make sure to squash all commits together. Anyway, LGTM.

bradgrantham-lunarg (approved):

nickdriscoll-lunarg:
I had to go through a pretty nasty rebase with this branch, and I'm seeing validation messages I didn't before.

~~I'm going to check if these messages are the app's fault and are being exposed with a new VVL version, or if I broke the fence handling code somewhere, before merging.~~ No it's actually broken.

nickdriscoll-lunarg:
The validation errors being thrown don't really make sense. In the following example, there are three fences that need to be synthetically signaled. Before issuing the synthetic `vkQueueSubmit()`, we call `vkResetFences()` just to make sure none of the fences are already signaled, but when we do, three validation messages are issued that each say that pFences[2] is currently in use.

So my questions are
1. How can any fence be in use after a call to `vkDeviceWaitIdle()` and before any calls to `vkQueueSubmit()`?
2. Why is it reporting an error for one of the three queues three separate times?

I'm really at a loss here. The PR still allows the trace to loop without hanging, and the Vulkan usage is valid according to my understanding of Vulkan, and I'm just not seeing the discrepancy. My second question makes me suspect VVL itself, but of course that's the last assumption I would want to make.

```
06-11 14:44:35.140 24611 24627 D gfxrecon: [gfxrecon] DEBUG - Waiting for device to idle...
06-11 14:44:35.169 24611 24627 D gfxrecon: [gfxrecon] DEBUG - Will synthetically wait on fence 3670
06-11 14:44:35.169 24611 24627 D gfxrecon: [gfxrecon] DEBUG - Will synthetically signal fence 125
06-11 14:44:35.169 24611 24627 D gfxrecon: [gfxrecon] DEBUG - Will synthetically signal fence 137
06-11 14:44:35.169 24611 24627 D gfxrecon: [gfxrecon] DEBUG - Will synthetically signal fence 82
06-11 14:44:35.169 24611 24627 D gfxrecon: [gfxrecon] DEBUG - Synthetically waiting on 1 fences...
06-11 14:44:35.169 24611 24627 D gfxrecon: [gfxrecon] DEBUG - Resetting 1 synthetically waited on fences...
06-11 14:44:35.169 24611 24627 D gfxrecon: [gfxrecon] DEBUG - Resetting fences to synthetically signal...
06-11 14:44:35.169 24611 24627 E gfxrecon: [gfxrecon] ERROR - DEBUG MESSENGER: Frame #2: VUID-vkResetFences-pFences-01123: vkResetFences(): pFences[2] (VkFence 0xa000000000a) is in use.
06-11 14:44:35.169 24611 24627 E gfxrecon: The Vulkan spec states: Each element of pFences must not be currently associated with any queue command that has not yet completed execution on that queue (https://docs.vulkan.org/spec/latest/chapters/synchronization.html#VUID-vkResetFences-pFences-01123)
06-11 14:44:35.169 24611 24627 E gfxrecon: [gfxrecon] ERROR - DEBUG MESSENGER: Frame #2: VUID-vkResetFences-pFences-01123: vkResetFences(): pFences[2] (VkFence 0xa000000000a) is in use.
06-11 14:44:35.169 24611 24627 E gfxrecon: The Vulkan spec states: Each element of pFences must not be currently associated with any queue command that has not yet completed execution on that queue (https://docs.vulkan.org/spec/latest/chapters/synchronization.html#VUID-vkResetFences-pFences-01123)
06-11 14:44:35.169 24611 24627 E gfxrecon: [gfxrecon] ERROR - DEBUG MESSENGER: Frame #2: VUID-vkResetFences-pFences-01123: vkResetFences(): pFences[2] (VkFence 0xa000000000a) is in use.
06-11 14:44:35.169 24611 24627 E gfxrecon: The Vulkan spec states: Each element of pFences must not be currently associated with any queue command that has not yet completed execution on that queue (https://docs.vulkan.org/spec/latest/chapters/synchronization.html#VUID-vkResetFences-pFences-01123)
```

antonio-lunarg:
> @nickdriscoll-lunarg  The validation errors being thrown don't really make sense. In the following example, there are three fences that need to be synthetically signaled. Before issuing the synthetic `vkQueueSubmit()`, we call `vkResetFences()` just to make sure none of the fences are already signaled, but when we do, three validation messages are issued that each say that pFences[2] is currently in use.
> 
> So my questions are
> 
> 1. How can any fence be in use after a call to `vkDeviceWaitIdle()` and before any calls to `vkQueueSubmit()`?
> 2. Why is it reporting an error for one of the three queues three separate times?
> 
> I'm really at a loss here. The PR still allows the trace to loop without hanging, and the Vulkan usage is valid according to my understanding of Vulkan, and I'm just not seeing the discrepancy. My second question makes me suspect VVL itself, but of course that's the last assumption I would want to make.
> 
> ```
> 06-11 14:44:35.140 24611 24627 D gfxrecon: [gfxrecon] DEBUG - Waiting for device to idle...
> 06-11 14:44:35.169 24611 24627 D gfxrecon: [gfxrecon] DEBUG - Will synthetically wait on fence 3670
> 06-11 14:44:35.169 24611 24627 D gfxrecon: [gfxrecon] DEBUG - Will synthetically signal fence 125
> 06-11 14:44:35.169 24611 24627 D gfxrecon: [gfxrecon] DEBUG - Will synthetically signal fence 137
> 06-11 14:44:35.169 24611 24627 D gfxrecon: [gfxrecon] DEBUG - Will synthetically signal fence 82
> 06-11 14:44:35.169 24611 24627 D gfxrecon: [gfxrecon] DEBUG - Synthetically waiting on 1 fences...
> 06-11 14:44:35.169 24611 24627 D gfxrecon: [gfxrecon] DEBUG - Resetting 1 synthetically waited on fences...
> 06-11 14:44:35.169 24611 24627 D gfxrecon: [gfxrecon] DEBUG - Resetting fences to synthetically signal...
> 06-11 14:44:35.169 24611 24627 E gfxrecon: [gfxrecon] ERROR - DEBUG MESSENGER: Frame #2: VUID-vkResetFences-pFences-01123: vkResetFences(): pFences[2] (VkFence 0xa000000000a) is in use.
> 06-11 14:44:35.169 24611 24627 E gfxrecon: The Vulkan spec states: Each element of pFences must not be currently associated with any queue command that has not yet completed execution on that queue (https://docs.vulkan.org/spec/latest/chapters/synchronization.html#VUID-vkResetFences-pFences-01123)
> 06-11 14:44:35.169 24611 24627 E gfxrecon: [gfxrecon] ERROR - DEBUG MESSENGER: Frame #2: VUID-vkResetFences-pFences-01123: vkResetFences(): pFences[2] (VkFence 0xa000000000a) is in use.
> 06-11 14:44:35.169 24611 24627 E gfxrecon: The Vulkan spec states: Each element of pFences must not be currently associated with any queue command that has not yet completed execution on that queue (https://docs.vulkan.org/spec/latest/chapters/synchronization.html#VUID-vkResetFences-pFences-01123)
> 06-11 14:44:35.169 24611 24627 E gfxrecon: [gfxrecon] ERROR - DEBUG MESSENGER: Frame #2: VUID-vkResetFences-pFences-01123: vkResetFences(): pFences[2] (VkFence 0xa000000000a) is in use.
> 06-11 14:44:35.169 24611 24627 E gfxrecon: The Vulkan spec states: Each element of pFences must not be currently associated with any queue command that has not yet completed execution on that queue (https://docs.vulkan.org/spec/latest/chapters/synchronization.html#VUID-vkResetFences-pFences-01123)
> ```

How about `vkAcquireNextImageKHR`? The fence used there should probably go into the `waited_upon_fences_`.
Then also try taking care of the fences in `vkQueueSubmit2` and `vkQueueSubmit2KHR`.

nickdriscoll-lunarg:
I agree that we need to handle those functions, but not handling them wouldn't have anything to do with _this_ error. It's saying that for the fences we _do_ track, the final one is being synthetically reset while in use, _despite_ waiting on all device work immediately before doing this fence fixup stuff.

antonio-lunarg:
> @nickdriscoll-lunarg  I agree that we need to handle those functions, but not handling them wouldn't have anything to do with _this_ error. It's saying that for the fences we _do_ track, the final one is being synthetically reset while in use, _despite_ waiting on all device work immediately before doing this fence fixup stuff.

> pFences[2] (VkFence 0xa000000000a) is in use

I suppose that corresponds to `Will synthetically signal fence 82`, and fence 82 got into the `waited_upon_fences_` because the app calls `vkWaitForFences`.

Then for some reason it's still used by something for which `vkDeviceWaitIdle` has no influence, and that something is probably the presentation engine.
> vkDeviceWaitIdle is equivalent to calling vkQueueWaitIdle for all queues owned by device.

My guess is that fence 82 is handed to `vkAcquireNextImageKHR` and it is not signaled by a queue submission, but it's signaled by the presentation engine when the image becomes available, and device idle does not guarantee that the presentation engine has done its job.

nickdriscoll-lunarg:
For this trace in particular, a fence is not used in the call to `vkAcquireNextImageKHR()`.

Here's the call:
```
{
  "index": 28244,
  "function": {
    "name": "vkAcquireNextImageKHR",
    "thread": 1,
    "return": "VK_SUCCESS",
    "args": {
      "device": 27,
      "swapchain": 12265,
      "timeout": 18446744073709551615,
      "semaphore": 12273,
      "fence": 0,
      "pImageIndex": 0
    }
  }
}
```

I do agree with you in general about needing to handle that entry point, but it really can't be the cause of this mysterious validation message.

nickdriscoll-lunarg:
You were right, I think the handling of `vkAcquireNextImageKHR()` got lost in the rebase.

PR #2937, opened 2026-05-07
replay: Refactor queue submit and fix imported semaphores

antonio-lunarg:


framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -1864,7 +1864,7 @@ void VulkanReplayDumpResourcesBase::OverrideCmdEndRenderingKHR(const ApiCallInfo
     OverrideCmdEndRendering(call_info, func, original_command_buffer);
 }
 
-VkResult VulkanReplayDumpResourcesBase::QueueSubmit(const std::vector<VkSubmitInfo>&   submit_infos,
+VkResult VulkanReplayDumpResourcesBase::QueueSubmit(const std::span<VkSubmitInfo>&     submit_infos,
```
fabian-lunarg: maybe better to use a span-value `std::span<const VkSubmitInfo> submit_infos`?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4384,81 +4383,79 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit(PFN_vkQueueSubmit
...
+
+    if (!altered_submits.empty())
+    {
+        // Make shallow copies of the VkSubmit info structures and change pWaitSemaphores to reference a copy of the
+        // original semaphore array with the imported semaphores omitted.
+        modified_submit_infos = std::vector<VkSubmitInfo>(current_submits_span.begin(), current_submits_span.end());
 
-            for (const auto& submit_iter : altered_submits)
+        for (const auto& submit_iter : altered_submits)
+        {
```
fabian-lunarg, line 4407: this looks like a pre-existing bug, but worth fixing in this PR anyway.

at the top of the loop we should clear the semaphore-vectors, they could accumulate semaphores from different submissions.

```
wait_semaphores.clear();
signal_semaphores.clear();
```

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4628,78 +4624,74 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit2(PFN_vkQueueSubmit2
...
+    if (!altered_submits.empty())
+    {
+        // Make shallow copies of the VkSubmit info structures and change pWaitSemaphores to reference a copy of the
+        // original semaphore array with the imported semaphores omitted.
+        modified_submit_infos = std::vector<VkSubmitInfo2>(submit_infos, std::next(submit_infos, submitCount));
 
-            for (const auto& submit_iter : altered_submits)
+        for (const auto& submit_iter : altered_submits)
+        {
+            // Shallow copy with filtered copy of pWaitSemaphores for submission info with imported semaphores.
```
fabian-lunarg, line 4648: and here:

```
wait_semaphore_infos.clear();
signal_semaphore_infos.clear()
```

framework/decode/vulkan_replay_consumer_base.cpp
antonio-lunarg: Clearing the vectors won't fix the issue. Vector memory should be alive until `func` is called. I think we need _per-submit_ vectors of semaphores.
This is a _big-fat-bug_ © in our sync management BTW, nice catch! 🕵🏻‍♂️

fabian-lunarg:
```
std::vector<VkSemaphore>& wait_semaphores   = per_submit_wait_semaphores[submit_info_key];
std::vector<VkSemaphore>& signal_semaphores = per_submit_signal_semaphores[submit_info_key];
```

nice, that works for me :)

fabian-lunarg (approved):
refactoring looks good, much cleaner and easier for me to read.
as extra the PR is fixing a longer-standing sync-bug, let's merge :)

PR #2939, opened 2026-05-07
convert: Update detection logic

MarkY-LunarG:
The detection logic was hard-coded and needed to be more flexible so that it can work with additional APIs.

### DIFF
--- a/tools/convert/convert_feature.h
+++ b/tools/convert/convert_feature.h
@@ -38,28 +38,36 @@ class ConvertFeatureBase
     virtual ~ConvertFeatureBase() = default;
 
     virtual void Initialize(decode::FileProcessor& file_processor, decode::JsonWriter* json_writer) = 0;
+    virtual bool WasDetected() const                                                                = 0;
     virtual void Destroy()                                                                          = 0;
 };
 
 // Template used by all API instantiations.
-template <class ConsumerT, class DecoderT>
+template <class ConsumerT, class DecoderT, class DetectConsumerT>
 class ConvertFeature : public ConvertFeatureBase
 {
   public:
+    // Allow forwarding of constructor arguments from children so we can properly initialize
+    // the detect_consumer_.
+    template <typename... DetectArgs>
+    ConvertFeature(DetectArgs&&... args) : detect_consumer_(std::forward<DetectArgs>(args)...)
+    {}
+
     ~ConvertFeature() { Destroy(); }
 
     void Initialize(decode::FileProcessor& file_processor, decode::JsonWriter* json_writer) final
     {
         json_consumer_.Initialize(json_writer);
+        decoder_.AddConsumer(&detect_consumer_);
         decoder_.AddConsumer(&json_consumer_);
         file_processor.AddDecoder(&decoder_);
     }
-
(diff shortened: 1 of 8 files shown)


MarkY-LunarG:
Also, before this change, we would actually go through all frames in a capture twice!  This incorporates the "was detected" logic into the processing loop, so should reduce execution time.

framework/decode/dx12_json_consumer_base.h
```diff
@@ -39,12 +39,24 @@ class Dx12JsonConsumerBase : public Dx12Consumer
     Dx12JsonConsumerBase();
     virtual ~Dx12JsonConsumerBase() override;
 
+    bool WasDetected() { return usage_detected_; }
```
mikes-lunarg:
```suggestion
    bool WasDetected() const { return usage_detected_; }
```

framework/decode/openxr_json_consumer_base.h
```diff
@@ -52,6 +52,8 @@ class OpenXrExportJsonConsumerBase : public OpenXrConsumer
 
     void Destroy();
 
+    bool WasDetected() { return usage_detected_; }
```
mikes-lunarg:
```suggestion
    bool WasDetected() const { return usage_detected_; }
```

framework/decode/vulkan_json_consumer_base.h
```diff
@@ -50,8 +50,19 @@ class VulkanExportJsonConsumerBase : public VulkanConsumer
 
     void Destroy();
 
+    bool WasDetected() { return usage_detected_; }
```
mikes-lunarg:
```suggestion
    bool WasDetected() const { return usage_detected_; }
```

tools/convert/convert_feature.h
```diff
@@ -38,6 +38,7 @@ class ConvertFeatureBase
     virtual ~ConvertFeatureBase() = default;
 
     virtual void Initialize(decode::FileProcessor& file_processor, decode::JsonWriter* json_writer) = 0;
+    virtual bool WasDetected()                                                                      = 0;
```
mikes-lunarg:
```suggestion
    virtual bool WasDetected() const                                                                = 0;
```

tools/convert/convert_feature.h
```diff
@@ -54,7 +55,7 @@ class ConvertFeature : public ConvertFeatureBase
         decoder_.AddConsumer(&json_consumer_);
         file_processor.AddDecoder(&decoder_);
     }
-
+    bool WasDetected() final { return json_consumer_.WasDetected(); }
```
mikes-lunarg:
```suggestion
    bool WasDetected() const final { return json_consumer_.WasDetected(); }
```

mikes-lunarg (approved):
I sprinkled some const suggestions throughout.

This certainly works for detecting API initialization in the first frame. Longer term, I wonder if a more generic solution exists by moving the check up to the ApiDecoder level. I think many of the tools would benefit from knowing if either 1) none of the attached decoders were used while processing the whole file or 2) the file contains any unhandled blocks

PR #2945, opened 2026-05-12
dump-resources: override query commands for replay dumps

dnlbaines:
Add dump-resources overrides for query-related Vulkan commands so generated replay code passes tracked object info into the dumper. The overrides replay query commands for dispatch/trace-rays dump contexts and avoid ending queries that were never begun, which fixes crashes during resource dumping.


framework/generated/khronos_generators/vulkan_generators/dump_resources_overrides.json
```diff
@@ -42,6 +42,16 @@
     "vkCmdExecuteCommands": "OverrideCmdExecuteCommands",
     "vkCmdBeginQuery": "OverrideCmdBeginQuery",
     "vkCmdEndQuery": "OverrideCmdEndQuery",
+    "vkCmdResetQueryPool": "OverrideCmdResetQueryPool",
+    "vkCmdWriteTimestamp": "OverrideCmdWriteTimestamp",
+    "vkCmdCopyQueryPoolResults": "OverrideCmdCopyQueryPoolResults",
```
fabian-lunarg, line 47: possibly missing `vkCmdCopyQueryPoolResultsToMemoryKHR` here?

framework/generated/generated_vulkan_replay_dump_resources.cpp
```diff
@@ -286,68 +267,30 @@ void VulkanReplayDumpResources::Process_vkCmdWriteTimestamp(
...
-        const std::vector<std::shared_ptr<DispatchTraceRaysDumpingContext>> dr_contexts = FindDispatchTraceRaysContexts(commandBuffer);
-        for (auto dr_context : dr_contexts)
-        {
-            VkCommandBuffer dispatch_rays_command_buffer = dr_context->GetDispatchRaysCommandBuffer();
-            if (dispatch_rays_command_buffer != VK_NULL_HANDLE)
-            {
-                func(dispatch_rays_command_buffer, pipelineStage, queryPool, query);
-            }
-        }
+        OverrideCmdWriteTimestamp(call_info, func, commandBuffer, pipelineStage, queryPool, query);
```
fabian-lunarg, line 275: I think there is a structural oversight, same for all added overrides I looked at.
before adding the overrides our generated code iterated all cmd-bufs for all `dc_contexts = FindDrawCallDumpingContexts(commandBuffer)`, recording/forwarding `func` into them.

this iteration got lost from code-gen when adding the json-override, 
but is also absent from `OverrideCmdWriteTimestamp`. 
it looked the same goes for all overrides, in addition to the `dr_context` (`DispatchTraceRaysDumpingContext`) also the `dc_contexts` (`DrawCallsDumpingContext`) should be handled.

framework/generated/khronos_generators/vulkan_generators/dump_resources_overrides.json
dnlbaines: Good catch, added.

framework/generated/generated_vulkan_replay_dump_resources.cpp
dnlbaines: I've restored this.

fabian-lunarg (approved):
I agree with the fix ("avoid ending queries that were never begun")
and the added overrides.
PR looks good to me, previous code-gen'ed behavior is
now handled in overrides with good code-reuse (e.g. `ForEachDrawCallCommandBuffer`).
I'm happy but @panos-lunarg might want to double-check

framework/decode/vulkan_replay_dump_resources.cpp
fabian-lunarg, line 3250: ok, one more thing :)

now thinking `OverrideCmdBeginQuery` / `OverrideCmdEndQuery` look a bit assymmetric vs.
their indexed flavours (`OverrideCmdBeginQueryIndexedEXT` ..).

I understand the crash-fix here was fixing `FinalizeCommandBuffer` ending queries it never began, but not forwarding of begin/end itself?

tldr
I think `OverrideCmdBeginQuery` / `OverrideCmdEndQuery` should
now invoke both `ForEachDrawCallCommandBuffer` / `ForEachDispatchTraceRaysCommandBuffer`,
same like the other overrides.

framework/decode/vulkan_replay_dump_resources.h
```diff
@@ -701,6 +793,37 @@ class VulkanReplayDumpResourcesBase
     std::shared_ptr<DrawCallsDumpingContext>              FindDrawCallContext(VkCommandBuffer original_command_buffer,
                                                                               decode::Index   qs_index);
 
+    template <typename Callback>
+    void ForEachDrawCallCommandBuffer(VkCommandBuffer original_command_buffer, Callback callback)
+    {
+        const std::vector<std::shared_ptr<DrawCallsDumpingContext>> dc_contexts =
+            FindDrawCallDumpingContexts(original_command_buffer);
+        for (auto dc_context : dc_contexts)
```
fabian-lunarg: minor. could be `const auto&`, avoiding shared_ptr / ref-count bumps

framework/decode/vulkan_replay_dump_resources.h
```diff
@@ -701,6 +793,37 @@ class VulkanReplayDumpResourcesBase
...
+            }
+        }
+    }
+
+    template <typename Callback>
+    void ForEachDispatchTraceRaysCommandBuffer(VkCommandBuffer original_command_buffer, Callback callback)
+    {
+        const std::vector<std::shared_ptr<DispatchTraceRaysDumpingContext>> dr_contexts =
+            FindDispatchTraceRaysContexts(original_command_buffer);
+        for (auto dr_context : dr_contexts)
```
fabian-lunarg: and here

fabian-lunarg:
last but not least:

@dnlbaines could you maybe provide a repro-case (.gfxr + input .json that crashes without this PR, or instructions how-to reproduce ourselves), adding to our (internal) test-coverage?

dnlbaines:
> last but not least:
> 
> @dnlbaines could you maybe provide a repro-case (.gfxr + input .json that crashes without this PR, or instructions how-to reproduce ourselves), adding to our (internal) test-coverage?

I can't share it publicly as its from an internal demo but I have uploaded it to share.lunarg.com and referenced it in the internal doc.

panos-lunarg (approved):

PR #2950, opened 2026-05-14
Report failure reason for additional arguments

dfriederich:
When passing additional, unexpected arguments to the replay app, it did just show the command line help, but not actually issue any hint about the actual command line failure.

For example on Android, we treat a single space as argument separator.
Hence using 2 spaces, as in '--dump-resources-dir  /sdcard/output' did cause the issue, just showing the usage help without any hint of what detail failed.
Fixes the non reporting of the failure reason, now it fails with

Unexpected positional argument '/sdcard/output'

### DIFF
--- a/tools/replay/android_main.cpp
+++ b/tools/replay/android_main.cpp
@@ -125,6 +125,10 @@ void android_main(struct android_app* app)
     }
     else if (arg_parser.IsInvalid() || (arg_parser.GetPositionalArgumentsCount() > 1))
     {
+        if (arg_parser.GetPositionalArgumentsCount() > 1)
+        {
+            GFXRECON_LOG_ERROR("Unexpected positional argument \'%s\'", arg_parser.GetPositionalArguments()[1].c_str());
+        }
         PrintUsage(kApplicationName);
         run = false;
     }
--- a/tools/replay/android_main.cpp
+++ b/tools/replay/android_main.cpp
@@ -127,7 +127,8 @@ void android_main(struct android_app* app)
     {
         if (arg_parser.GetPositionalArgumentsCount() > 1)
         {
-            GFXRECON_LOG_ERROR("Unexpected positional argument \'%s\'", arg_parser.GetPositionalArguments()[1].c_str());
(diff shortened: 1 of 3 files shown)


tools/replay/android_main.cpp
```diff
@@ -125,6 +125,10 @@ void android_main(struct android_app* app)
     }
     else if (arg_parser.IsInvalid() || (arg_parser.GetPositionalArgumentsCount() > 1))
     {
+        if (arg_parser.GetPositionalArgumentsCount() > 1)
+        {
+            GFXRECON_LOG_ERROR("Unexpected positional argument \'%s\'", arg_parser.GetPositionalArguments()[1].c_str());
+        }
```
MarkY-LunarG, line 132: Should you add the `else if (Count == 0)` path here as well (like the desktop_main.cpp version)?

tools/replay/android_main.cpp
```diff
@@ -125,6 +125,10 @@ void android_main(struct android_app* app)
     }
     else if (arg_parser.IsInvalid() || (arg_parser.GetPositionalArgumentsCount() > 1))
     {
+        if (arg_parser.GetPositionalArgumentsCount() > 1)
+        {
+            GFXRECON_LOG_ERROR("Unexpected positional argument \'%s\'", arg_parser.GetPositionalArguments()[1].c_str());
```
MarkY-LunarG: I think a better message would be "Too many inputs provided to command line starting with...".

tools/replay/android_main.cpp
dfriederich: The condition in line 126 in android_main.cpp is different to the desktop_main.cpp, the control flow does not go here 'just' for count == 0.
I did not look into what it means for count == 0 on Android, only wanted to add the message :-)

tools/replay/android_main.cpp
dfriederich: changed.

tools/replay/android_main.cpp
dfriederich: For no positional argument, android_main.cpp is reading from a default location, 
'/sdcard/gfxrecon_capture.gfxr'

PR #2951, opened 2026-05-15
Vdr hash filenames

panos-lunarg:


framework/decode/vulkan_replay_dump_resources_delegate.cpp
```diff
@@ -132,23 +134,127 @@ static bool DumpBufferToFile(DumpedBuffer&           dumped_buffer,
...
+            util::hash::hash_combine_64(hash, dumped_buffer_desc.desc_tuple.array_index);
+        }
+        break;
+
+        case DumpResourceType::kInitBufferMetaCommand:
+        {
+            GFXRECON_ASSERT(region_index == NO_INDEX);
+
+            const auto& dumped_transfer_command = static_cast<const DumpedTransferCommand&>(dumped_resource);
+            auto*       dumped_init_buffer =
```
bradgrantham-lunarg: Should all these be `const auto`?

bradgrantham-lunarg (changes requested):

framework/decode/vulkan_replay_dump_resources_delegate.cpp
bradgrantham-lunarg: Submit as a 2nd PR.

bradgrantham-lunarg (approved):

framework/decode/vulkan_replay_dump_resources_delegate.cpp
panos-lunarg: Was very straight forward. Done in this pr

Issue #2949, opened 2026-05-14
[replay] Support replaying with `-m rebind` of VK captures from vkd3d-proton


werman:
**Describe the replay bug:**

Replaying Vulkan captures obtained from translated D3D12 workload through vkd3d-proton runs into issues when replacing BDA addresses (`-m rebind`):
- https://github.com/LunarG/gfxreconstruct/issues/2941
- Assert on NULL DestroyBuffer and DestroyImage in VulkanRebindAllocator
- `offset` in `CmdPushConstants` not being taken into account
- `firstSet` in `CmdBindDescriptorSets` not being taken into account
- vkd3d-proton uses uvec2 for BDA values
- Those uvec2 accesses are not determined in all cases vkd3d-proton do them.

I'll open PR to fix most of these shortly. But the last one I'm not that sure about the fix, the branch with all fixes is at https://github.com/werman/gfxreconstruct/commits/feature/better-BDA-rebind/, which allows correctly replaying a capture of fairly complex game with `-m rebind`.

**Build Environment:**
b7eae6b8231fb444f3e32f607f798d978c0546df

**To Reproduce**

I made a demo in my vkd3d-proton branch to reproduce most usages of BDA in vkd3d, which reproduces issues described above: https://github.com/werman/vkd3d-proton/tree/test/gfxreconstruct-bda-demo

mikes-lunarg:
Thanks! I'll keep an eye out for the PR. I took a quick look at your branch and agree most of these are pretty straightforward, but I will also need to spend more time with that last commit.

werman:
The last one is 100% vibe-coded and I'm not confident to say that it is good, only that it solves the issue at hand for me.

mikes-lunarg:
@werman I've taken another try at the remaining uvec2 part of your original request in #3019. The demo app you provided was very helpful, I've confirmed it is now working with `-m rebind`. Let me know if you get a chance to try it out.

```
./build/tools/replay/gfxrecon-replay bda-paths.gfxr -m rebind --paused
================== Start timer (Frame: 1) ==================
[gfxrecon] WARNING - Failed to track back the Function Variable OpStore, hit a OpIAdd
[gfxrecon] WARNING - Failed to track back the Function Variable OpStore, hit a OpIAdd
[gfxrecon] WARNING - Failed to track back the Function Variable OpStore, hit a OpIAdd
[gfxrecon] WARNING - Failed to track back the Function Variable OpStore, hit a OpIAdd
[...]

PR #2952, opened 2026-05-15
replay: Fixes to replay captures from vkd3d-proton with `-m rebind`

werman:
Addresses issues described in https://github.com/LunarG/gfxreconstruct/issues/2949:

- `Assertion failed: 'descriptor_set_binding_info.desc_type == write->descriptorType'` - For mutable descriptor we can write any supported descriptor type.
  - Though this change doesn't handle potential issue when app would write different descriptor type at the same target idx. 
- Passing NULL to `::DestroyBuffer` and `::DestroyImage` in `VulkanRebindAllocator` should not assert since it's valid from VK perspective.
- `offset` in `CmdPushConstants` should be taken into account - push consts could be updated piece-wise, though this doesn't cover a case when a single uint64 address is being set with two push consts updates, 32b at a time.
- `firstSet` in `CmdBindDescriptorSets` should be taken into account.

Beside the above, the only change left to make replay with `-m rebind` work is treat uvec2 accesses in SPIRV as if they can be addresses, there is a draft commit to do this at https://github.com/werman/gfxreconstruct/commits/feature/better-BDA-rebind/ but I'm absolutely not confident regarding it, but with that change I was able to replay a trace of fairly complex game.

Why do I care about replaying vkd3d-proton traces with `-m rebind`? Without that I cannot make a RenderDoc capture from gfxreconstruct trace.

Also here is a demo that can reproduce BDA replaying issues: https://github.com/werman/vkd3d-proton/tree/test/gfxreconstruct-bda-demo (made on top of vkd3d-proton demos). Was with the following vkd3d options: `VKD3D_DISABLE_EXTENSIONS=VK_EXT_descriptor_buffer,VK_EXT_mesh_shader,VK_EXT_device_generated_commands,VK_NV_device_generated_commands,VK_NV_device_generated_commands_compute VKD3D_CONFIG=force_raw_va_cbv,nodxr`

### DIFF
--- a/framework/decode/vulkan_rebind_allocator.cpp
+++ b/framework/decode/vulkan_rebind_allocator.cpp
@@ -312,7 +312,12 @@ void VulkanRebindAllocator::DestroyBuffer(VkBuffer                     buffer,
                                           ResourceData                 allocator_data)
 {
     GFXRECON_UNREFERENCED_PARAMETER(allocation_callbacks);
-    GFXRECON_ASSERT(buffer != VK_NULL_HANDLE);
+
+    if (buffer == VK_NULL_HANDLE)
+    {
+        GFXRECON_ASSERT(allocator_data == 0);
+        return;
+    }
 
     if (allocator_data != 0)
     {
@@ -364,7 +369,12 @@ void VulkanRebindAllocator::DestroyImage(VkImage                      image,
                                          ResourceData                 allocator_data)
 {
     GFXRECON_UNREFERENCED_PARAMETER(allocation_callbacks);
-    GFXRECON_ASSERT(image != VK_NULL_HANDLE);
+
+    if (image == VK_NULL_HANDLE)
+    {
+        GFXRECON_ASSERT(allocator_data == 0);
+        return;
+    }
 
     if (allocator_data != 0)
     {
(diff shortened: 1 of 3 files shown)


framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -312,7 +312,15 @@ void VulkanRebindAllocator::DestroyBuffer(VkBuffer                     buffer,
...
 {
     GFXRECON_UNREFERENCED_PARAMETER(allocation_callbacks);
-    GFXRECON_ASSERT(buffer != VK_NULL_HANDLE);
+
+    if (buffer == VK_NULL_HANDLE)
+    {
+        if (allocator_data != 0)
+        {
+            delete reinterpret_cast<ResourceAllocInfo*>(allocator_data);
+        }
```
mikes-lunarg:
```suggestion
        GFXRECON_ASSERT(allocator_data == 0);
```
It shouldn't be possible for allocator_data to be non-zero here, we can just assert and return.

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -364,7 +372,15 @@ void VulkanRebindAllocator::DestroyImage(VkImage                      image,
...
 {
     GFXRECON_UNREFERENCED_PARAMETER(allocation_callbacks);
-    GFXRECON_ASSERT(image != VK_NULL_HANDLE);
+
+    if (image == VK_NULL_HANDLE)
+    {
+        if (allocator_data != 0)
+        {
+            delete reinterpret_cast<ResourceAllocInfo*>(allocator_data);
+        }
```
mikes-lunarg:
```suggestion
        GFXRECON_ASSERT(allocator_data == 0);
```
Same

mikes-lunarg (commented):
Changes look good, just have a minor suggestion about the `VK_NULL_HANDLE` delete handling

mikes-lunarg (approved):

PR #2955, opened 2026-05-21
Remove 32-bit Jenkins pipeline configuration

beau-lunarg:


bradgrantham-lunarg:
While we really want to retire 32-bit soon and want to take the 32-bit testing burden out of CI, I think we still probably need 32-bit build support in for manual in case something goes sideways until then.  I think that argues that bits would still be available as a switch and needs to be plumbed.

ci/Jenkinsfile
```diff
@@ -74,29 +74,28 @@ pipeline {
...
-                            'Windows11 AMD 9070 Dbg32': runJob.gfxrTestWindows('Win11 AMD 9070 Dbg32', runJob.ReleaseMode, runJob.Win11AMD9070Label, runJob.Bit32, "commit-suite.json", scm.branches),
-                            'Windows11 Nvidia 5080 Rel64': runJob.gfxrTestWindows('Win11 Nvidia 5080 Rel64', runJob.ReleaseMode, runJob.Win11Nvidia50XXLabel, runJob.Bit64, "commit-suite.json", scm.branches),
-                            'Windows11 Nvidia 5080 Dbg32': runJob.gfxrTestWindows('Win11 Nvidia 5080 Dbg32', runJob.ReleaseMode, runJob.Win11Nvidia50XXLabel, runJob.Bit32, "commit-suite.json", scm.branches),
+                            'Android Rel64': runJob.gfxrTestAndroid('Android Rel', runJob.ReleaseMode, runJob.AndroidLabel, "commit-suite.json", scm.branches),
+                            'Linux Mesa Rel64': runJob.gfxrTestLinux('Linux Mesa Rel', runJob.ReleaseMode, runJob.LinuxMesaLabel, "commit-suite.json", scm.branches),
+                            'Linux Mesa Dbg32': runJob.gfxrTestLinux('Linux Mesa Dbg', runJob.ReleaseMode, runJob.LinuxMesaLabel, "commit-suite.json", scm.branches),
```
bradgrantham-lunarg: Will these and others will be Dbg64 in this change?

ci/runJob.bat
```diff
@@ -82,4 +82,4 @@ python --version
 python -m venv %WORKSPACE%\python-venv
 %WORKSPACE%\python-venv\Scripts\python -m pip install --no-cache-dir -r VulkanTests\requirements.txt > %WORKSPACE%\python-venv.txt 2>&1 || exit /b
 
-%WORKSPACE%\python-venv\Scripts\python VulkanTests\gfxrecontest.py --compiler vs2022 --build-mode %BUILD_MODE% --bits %BITS% %WINDOWS_TEST_ARGS% --suite "ci-gfxr-suites\%GFXRECON_TRACE_SUBDIR%\%TEST_SUITE%" --trace-dir "%GFXRECON_TRACE_DIR%" --result-dir "%RESULTS_DIR%"
+%WORKSPACE%\python-venv\Scripts\python VulkanTests\gfxrecontest.py --compiler vs2022 --build-mode %BUILD_MODE% --bits 64 %WINDOWS_TEST_ARGS% --suite "ci-gfxr-suites\%GFXRECON_TRACE_SUBDIR%\%TEST_SUITE%" --trace-dir "%GFXRECON_TRACE_DIR%" --result-dir "%RESULTS_DIR%"
```
bradgrantham-lunarg: If I try to run a 32-bit manual job, will this override my 32 bit specification?

bradgrantham-lunarg (changes requested):
Have you tried a manual run of 32 bits with this PR to verify that still works?  (I know this is a pain, just in case someone squawks at the last minute.)

(FYI I just have the one review comment; "If I try to run a 32-bit manual job, will this override my 32 bit specification?".  I have an old pending comment "Will these and others will be Dbg64 in this change?" that no longer applies and that you can ignore if it gets posted with this review.)

ci/Jenkinsfile.extended
```diff
@@ -137,15 +137,15 @@ pipeline {
...
-                            'Windows AMD Rel64': runJob.gfxrTestWindows('Win AMD Rel64 Ext', runJob.ReleaseMode, runJob.WinAMDExtendedLabel, runJob.Bit64, "extended-suite.json", requestedBranches),
-                            'Windows Nvidia Rel64': runJob.gfxrTestWindows('Win Nvidia Rel64 Ext', runJob.ReleaseMode, runJob.WinNvidiaExtendedLabel, runJob.Bit64, "extended-suite.json", requestedBranches),
-                            'Windows11 AMD 9070 Rel64': runJob.gfxrTestWindows('Win11 AMD 9070 Rel64 Ext', runJob.ReleaseMode, runJob.Win11AMD9070Label, runJob.Bit64, "extended-suite.json", requestedBranches),
-                            'Windows11 Nvidia 50XX Rel64': runJob.gfxrTestWindows('Win11 Nvidia 50XX Rel64 Ext', runJob.ReleaseMode, runJob.Win11Nvidia50XXLabel, runJob.Bit64, "extended-suite.json", requestedBranches),
+                            'Android Rel': runJob.gfxrTestAndroid('Android R Ext', runJob.ReleaseMode, runJob.AndroidLabel, "extended-suite.json", requestedBranches),
```
ncesario-lunarg:
```suggestion
                            'Android Rel': runJob.gfxrTestAndroid('Android Rel Ext', runJob.ReleaseMode, runJob.AndroidLabel, "extended-suite.json", requestedBranches),
```
?

ci/Jenkinsfile.extended-manual
```diff
@@ -86,63 +86,63 @@ pipeline {
                         def runJob = load 'ci/runJob.groovy'
 
                         tests = [
-                            'Android Rel64': runJob.gfxrTestAndroidManual('Android R64', runJob.AndroidLabel, runJob.ReleaseMode, runJob.Bit64, params.TEST_SUITE,
+                            'Android Rel': runJob.gfxrTestAndroidManual('Android Rel', runJob.AndroidLabel, runJob.ReleaseMode, '64', params.TEST_SUITE,
```
ncesario-lunarg, line 89: I'm not familiar with this pipeline/groovy syntax, but it looked like the `bits` parameter was removed in the `runJob` function; however, it looks like it's still being used here?

ncesario-lunarg (commented):
I'm probably missing some context, but my assumption for this change would be that only the 32 bit jobs would be removed, but it looks like this is removing the ability to specify anything other than 64 bits? If that's the design goal and I'm just ignorant to it (very possible), then this is probably good. If that's not the goal, it might be worth parring this back to just remove the 32 bit jobs from the appropiate `tests` arrays.

ci/Jenkinsfile.extended-manual
beau-lunarg: Bits parameter was removed for all fns, but added back into the manual ones at Brads request.  This is just a patch so that the extended ones keep running 64bit.

beau-lunarg:
> I'm probably missing some context, but my assumption for this change would be that only the 32 bit jobs would be removed, but it looks like this is removing the ability to specify anything other than 64 bits? If that's the design goal and I'm just ignorant to it (very possible), then this is probably good. If that's not the goal, it might be worth parring this back to just remove the 32 bit jobs from the appropiate `tests` arrays.

Yes, the purpose is to remove 32 bit jobs for all normal runs.  However, Brad has asked for the ability to run 32 bit manually, so that was added back in.

ci/runJob.bat
beau-lunarg: The runJob.bat files are old and shouldn't be here.  I though tthey had been taken out when I split build and test, but it looks like it either got missed or got rebased back in somehow.  If you look at runJob.groovy, you'll see it isn't called anywhere

beau-lunarg:
> Have you tried a manual run of 32 bits with this PR to verify that still works? (I know this is a pain, just in case someone squawks at the last minute.)
> 
> (FYI I just have the one review comment; "If I try to run a 32-bit manual job, will this override my 32 bit specification?". I have an old pending comment "Will these and others will be Dbg64 in this change?" that no longer applies and that you can ignore if it gets posted with this review.)

I created a special manual pipeline to test this out.  The original one picks up its Jenkinsfile script from dev, so it wouldn't have tested this change without a merge.
Here is a 32bit run
http://jenkins:8080/job/gfxr-pipelines/job/gfxr-manual-no-32bit/5/pipeline-overview/

Notice in the build logs that cmake is using `-DCMAKE_C_FLAGS=-m32 -DCMAKE_CXX_FLAGS=-m32 -DCMAKE_SHARED_LINKER_FLAGS=-m32`

Here is a 64bit run, that doesn't have those flags:
http://jenkins:8080/job/gfxr-pipelines/job/gfxr-manual-no-32bit/6/pipeline-overview/?selected-node=40

bradgrantham-lunarg (approved):

PR #2957, opened 2026-05-21
Fix debug build assertion error when advancing to next frame

wzchu-lunarg:
Fix issue on debug builds of GFXR where advancing to the next frame (using rightarrow or n) causes an assertion error. This is due to the contexts that process keyboard inputs to call `application`'s `PlaySingleFrame()` eventually checking an assertion for `ReplayEventSink`'s member variable `frame_active_` even though it was never set to true with `FrameBegin`. 

One recommendation taken from Claude is to do the same as seen in the `application.cpp`'s loop to call `FrameBegin` and `FrameEnd`, except we move these calls to inside `PlaySingleFrame`.

### DIFF
--- a/framework/application/application.cpp
+++ b/framework/application/application.cpp
@@ -170,12 +170,6 @@ void Application::Run()
                 fps_info_->BeginFrame(frame_number);
             }
 
-            if (replay_event_sink_)
-            {
-                // Replay event plugin uses a 0-based frame index.
-                replay_event_sink_->FrameBegin(file_processor_->GetCurrentFrameNumber());
-            }
-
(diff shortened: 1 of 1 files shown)


fabian-lunarg (approved):
looks good!

antonio-lunarg (approved):

charles-lunarg:
(Repeating here, for visibility)

The PR description is excellent, could you add it to the commit message? That way the commit message contains "why" things were changed and in the git repo itself instead of being only in the github PR.

bradgrantham-lunarg (approved):

PR #2961, opened 2026-05-26
Fix Android 16k alignment

panos-lunarg:


antonio-lunarg:
How about this? https://github.com/LunarG/gfxreconstruct/blob/dev/CMakeLists.txt#L315
Doesn't that apply to other targets?

panos-lunarg:
> How about this? https://github.com/LunarG/gfxreconstruct/blob/dev/CMakeLists.txt#L315 Doesn't that apply to other targets?

Doesn't seem to have any effect. I removed it.

Edit: Basically this is the root cmake file which doesn't seem to be included in the cmake files we use for android builds

antonio-lunarg:
> > How about this? https://github.com/LunarG/gfxreconstruct/blob/dev/CMakeLists.txt#L315 Doesn't that apply to other targets?
> 
> Doesn't seem to have any effect. I removed it.
> 
> Edit: Basically this is the root cmake file which doesn't seem to be included in the cmake files we use for android builds

I see. Good enough.

Off-topic: I am not really fond of some of the duplication of `CMakeLists.txt` in the Android folder. When introducing `framework/plugin/` I opted for a single `CMakeLists.txt` that would work for both desktop and Android.

antonio-lunarg (approved):

bradgrantham-lunarg (approved):

PR #2964, opened 2026-05-26
Fixes for external pipeline cache file

panos-lunarg:
When --load-pipeline-cache and --save-pipeline-cache point to the same
file then the file is not emptied. Logic attempts to incrementally
accumulate all caches when multiple incomplete runs are done by marking which caches have been loaded from the file and which not and only storing these in the file.


dfriederich:
Works in my local tests :-)

dfriederich (approved):
works for me :-)

MarkY-LunarG (approved):

PR #2967, opened 2026-05-27
Use code generator for frame looping functions

davidlunarg:


davidlunarg:
Here's an initial implementation of using code gen to implement frame looping. I leveraged the replay consumer to create the frame looping generator, so there's still lots of code in the the frame looping generators that needs to be removed.

But the basic implementation is correct, would appreciate high-level feedback on the implementation. In the meantime, I'll continue with cleaning up the new generators.

davidlunarg:
Implementation is complete. This branch will need to be rebased from latest dev. Will hold off on rebasing to avoid making changes and then throwing them away in case the implementation has to change significantly.

bradgrantham-lunarg:
LGTM but I would like a thumbs-up from @antonio-lunarg

antonio-lunarg:
I think it's a good start for now, but we'll need to improve the way we manage resources during the looping range: https://github.com/LunarG/gfxreconstruct/pull/2935#issuecomment-4565215091
```
Here are the possible cases for descriptor pool/set creation/destruction

Created before loop range but destroyed in the loop range
Omit destruction enitrely
Created before loop range but destroyed after loop range
No special handling
Created during loop range, destroyed after
Only create on first iteration
Created during loop range, destroyed during
No special handling
The solution here is to track all "dangling" pools/sets, then ignore their destruction entirely, and only create them on the first iteration of the loop range.

A "dangling" resource is one that was either
- Created during the loop range but destroyed after it
or
- Created before the loop range but destroyed during it
```

antonio-lunarg (approved):

davidlunarg:
Yes, we need to improve the way we handle allocation/freeing of resources in the frame we are looping in. With code gen, we can implement the solution once in the code generator, and address the issue for all functions that allocate/free.

I'll start rebasing to latest dev.

bradgrantham-lunarg (approved):

PR #2971, opened 2026-05-28
Remove `TRANSFER_SRC_BIT` from virtual swapchain images

marius-pelegrin-arm:
At replay time, the `VulkanVirtualSwapchain` creates it's underlying swapchain by adding `VK_IMAGE_USAGE_TRANSFER_SRC_BIT` and `VK_IMAGE_USAGE_TRANSFER_DST_BIT`.

`DST` is necessary because of the copy from the virtual image to the actual swapchain image.
However, `SRC` is not useful, as the swapchain image are never read back. In particular, I think this was set for screenshot handling, but screenshot is done on the virtual image, not the actual swapchain image.


fabian-lunarg (approved):
agreed

PR #2972, opened 2026-05-28
Fix renderpass image layout tracking in vulkan replay consumer

bartosz-muszarski-arm:
Currently replay tracks render pass attachment finalLayout at vkCmdBeginRenderPass(2) and updates command-buffer image layout state before the render pass had actually ended. 
This is problematic because:
* finalLayout only becomes valid when the render pass ends
* tracking it at begin made replay think attachments had already left their in-pass layout
* tracked image layout can be wrong while a render pass was still active

This showed up as incorrect layout state for render pass attachments and resulted in visual corruptions in rebind path. 

This change:
* Keeps track of the active render pass, framebuffer, and resolved attachment image views during vkCmdBeginRenderPass(2)
* Applies attachment finalLayout to tracked image state only on vkCmdEndRenderPass(2).

### DIFF
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -692,6 +692,9 @@ struct VulkanCommandBufferInfo : public VulkanPoolObjectInfo<VkCommandBuffer>
 {
     bool                                                      is_frame_boundary{ false };
     std::vector<format::HandleId>                             frame_buffer_ids;
+    format::HandleId                                          active_render_pass_id{ format::kNullHandleId };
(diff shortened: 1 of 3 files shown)


fabian-lunarg (approved):
agree with the fix.
added two small consolidation/cleanup commits (code-reuse, unreachable code).

also very nice:
the PR also fixes a latent bug in `OverrideCmdBeginRenderPass2`.                                                                                                                                                                                                                                                       
it was missing the pNext-lookup for `VK_FRAMEBUFFER_CREATE_IMAGELESS_BIT` (OverrideCmdBeginRenderPass was fine)

PR #2976, opened 2026-05-30
Vdr fixes for 3d images

panos-lunarg:
Fixes dumping of 3D images. Now all depth indices of 3D images are dumped in a separate file.


tools/replay/parse_dump_resources_cli.cpp
```diff
@@ -321,6 +322,49 @@ static void ExtractVulkanDumpResourcesParameters(const nlohmann::ordered_json
...
+                                                 const std::string&  alternative_end_range_name,
+                                                 uint32_t            cmd_index,
+                                                 uint32_t            set,
+                                                 uint32_t            binding,
+                                                 uint32_t            ai)
+{
+    uint32_t count;
+    if (!json_it.contains(range_name))
+    {
+        count = std::numeric_limits<uint32_t>::max();
```
MarkY-LunarG: You could just set this during the initialization of count on line 334.  Then you could remove the set here and in line 348.

framework/decode/vulkan_replay_dump_resources_common.cpp
```diff
@@ -386,18 +386,38 @@ VkResult DumpImage(DumpedImage&                         dumped_image,
...
+    graphics::AspectFlagsToFlagBits(modified_subresource_range.aspect_mask, aspects);
 
-    const uint32_t total_subresources = GFXRECON_NARROWING_CAST(
-        uint32_t, (aspects.size() * (modified_subresource_range.layerCount * modified_subresource_range.levelCount)));
+    const VkExtent3D scaled_extent = (scale != 1.0f && scaling_supported)
+                                         ? graphics::ScaleExtent3DNoDepth(image_info->extent, scale)
+                                         : image_info->extent;
 
-    data.resize(total_subresources);
+    const bool is_3d = image_info->type == VK_IMAGE_TYPE_3D;
```
MarkY-LunarG, line 399: Is it only 3d textures that behave like this or do layered cube-map textures also behave similarly?

MarkY-LunarG (approved):

framework/decode/vulkan_replay_dump_resources_common.cpp
panos-lunarg: No. The difficulty is with 3d images as each mip, in the same fashion as width and height halves in each level, the same happens with depth. So each mip has different number of depth indices. Calculating the total number requires a for loop.

Multi layered images are simpler as the number of layers remain the same in each mip level.

tools/replay/parse_dump_resources_cli.cpp
panos-lunarg: done

PR #2977, opened 2026-06-01
Fix bad `PointerDecoder` usage

marius-pelegrin-arm:
For Vulkan function that follows the following usage:
```
// Get objectCount value from the driver
uint32_t objectCount;
vkGetObject(..., &objectCount, nullptr);

// Get objects from the driver
VkObject objects[objectCount]
vkGetObject(..., &objectCount, objects);
```

The override in VulkanReplayConsumer base should ALWAYS use the corresponding `GetOutputPointer()` for the replay time count/object and `GetPointer()` to get the capture time count/object.

Some functions were badly using `GetPointer()` for the replay time count/object leading to errors when using the queried lists.


fabian-lunarg (approved):
> ... should ALWAYS use the corresponding GetOutputPointer() for the replay time count/object and GetPointer() to get the capture time count/object

agree and approve

PR #2978, opened 2026-06-01
Restore synchronization on `vkGetQueryPoolResults` and `vkGetEventStatus`

marius-pelegrin-arm:
PR #2723 removed busy-wait loops for cases where `vkGetQueryPoolResults` and `vkGetEventStatus` were successful at capture time but not at replay time.

It's true that busy looping is to be avoided, and that the former implementation was not even working anyway due to the "max retry" threshold.
However, we still need to make sure that the replayer doesn't break the synchronization scheme of the overall workload: if the original application was using `vkGetEventStatus` to make sure a Vulkan resource is not in use anymore before reusing it in further Vulkan calls, we need to make sure that any call to `vkGetEventStatus` that was successful at capture time is also successful at replay time (and the same for `vkGetQueryPoolResults`), the same way it is done for `vkGetFenceStatus`.

For `vkQueryPoolResults`, I propose to use `VK_QUERY_RESULT_WAIT_BIT`, which makes the request blocking.

For `vkGetEventStatus`, as there is no equivalent flag or `vkWaitForEvent` function, I propose to re-implement the busy-waiting with the following changes:
- No "max retry" anymore, it continues for as long as necessary
- A thread sleep of 100us between every "retry". At the time scale of a GPU, this is slow enough to avoid spamming the call millions of times, and fast enough to avoid stuttering the entire replay.

This is not a perfect solution, but this solution actually guarantees correct synchronization for single-threaded applications using these calls, and attempts to minimize the CPU impact of the busy-waiting.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -82,6 +82,10 @@ GFXRECON_BEGIN_NAMESPACE(decode)
 constexpr char kUnknownDeviceLabel[]  = "<Unknown>";
 constexpr char kValidationLayerName[] = "VK_LAYER_KHRONOS_validation";
 
+// vkGetEventStatus 'can' be used for synchronization, but cannot be (host-)waited on.
+// we poll until receiving VK_EVENT_SET, when the capture returned it.
+constexpr auto kGetEventStatusPollInterval = std::chrono::microseconds(100);
+
 const std::unordered_set<std::string> kSurfaceExtensions = {
     VK_KHR_ANDROID_SURFACE_EXTENSION_NAME,  VK_MVK_IOS_SURFACE_EXTENSION_NAME,
     VK_MVK_MACOS_SURFACE_EXTENSION_NAME,    VK_KHR_MIR_SURFACE_EXTENSION_NAME,
@@ -4311,17 +4315,18 @@ VkResult VulkanReplayConsumerBase::OverrideGetEventStatus(PFN_vkGetEventStatus  
 
     VkResult result = func(device, event);
 
-    util::BeginInjectedCommands();
-
-    // If the query was successful at capture time, it MUST be successful at replay time, to avoid sync issues
-    while (original_result == VK_EVENT_SET && result == VK_EVENT_RESET)
+    // If the query was successful at capture time, it MUST be successful at replay time, to avoid sync issues.
+    // There is no host wait-primitive for events, so poll.
+    if (original_result == VK_EVENT_SET && result == VK_EVENT_RESET)
     {
-        std::this_thread::sleep_for(std::chrono::microseconds(100)); // Avoid true busy waiting
-        result = func(device, event);
+        util::BeginInjectedCommands();
+        while (result == VK_EVENT_RESET)
+        {
+            std::this_thread::sleep_for(kGetEventStatusPollInterval);
+            result = func(device, event);
+        }
+        util::EndInjectedCommands();
     }
-
-    util::EndInjectedCommands();
-
     return result;
 }
 
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -69,8 +69,10 @@
 #include "Vulkan-Utility-Libraries/vk_format_utils.h"
 
 #include <algorithm>
+#include <chrono>
 #include <limits>
 #include <numeric>
+#include <thread>
 #include <unordered_set>
 #include <future>
 
@@ -93,7 +95,7 @@ const std::unordered_set<std::string> kSurfaceExtensions = {
 const std::unordered_set<std::string> kTrimStateSetupDeviceExtensions = { VK_EXT_SHADER_STENCIL_EXPORT_EXTENSION_NAME };
 
 const std::unordered_set<std::string> kFunctionsAllowedToReturnDifferentCodeThanCapture = {
-    "vkSetDebugUtilsObjectNameEXT", "vkSetDebugUtilsObjectTagEXT", "vkGetQueryPoolResults", "vkGetEventStatus"
+    "vkSetDebugUtilsObjectNameEXT", "vkSetDebugUtilsObjectTagEXT"
 };
 
 // LUT containing an allow-list of differing Vulkan return-types (mapping: capture -> replay)
@@ -104,7 +106,8 @@ const std::unordered_map<VkResult, VkResult> kResultValuesAllowedDifferentCodeTh
     { VK_ERROR_OUT_OF_DATE_KHR, VK_SUCCESS },
     { VK_SUBOPTIMAL_KHR, VK_SUCCESS },
     { VK_ERROR_FORMAT_NOT_SUPPORTED, VK_SUCCESS },
-    { VK_ERROR_OUT_OF_POOL_MEMORY, VK_SUCCESS }
+    { VK_ERROR_OUT_OF_POOL_MEMORY, VK_SUCCESS },
+    { VK_EVENT_RESET, VK_EVENT_SET }
 };
 
 static VKAPI_ATTR VkBool32 VKAPI_CALL DebugReportCallback(VkDebugReportFlagsEXT      flags,
@@ -4296,6 +4299,32 @@ VkResult VulkanReplayConsumerBase::OverrideGetFenceStatus(PFN_vkGetFenceStatus  
     return result;
 }
 
+VkResult VulkanReplayConsumerBase::OverrideGetEventStatus(PFN_vkGetEventStatus    func,
+                                                          VkResult                original_result,
+                                                          const VulkanDeviceInfo* device_info,
+                                                          const VulkanEventInfo*  event_info)
+{
+    GFXRECON_ASSERT(device_info != nullptr && event_info != nullptr);
+
+    VkDevice device = device_info->handle;
+    VkEvent  event  = event_info->handle;
+
+    VkResult result = func(device, event);
+
+    util::BeginInjectedCommands();
+
+    // If the query was successful at capture time, it MUST be successful at replay time, to avoid sync issues
+    while (original_result == VK_EVENT_SET && result == VK_EVENT_RESET)
+    {
+        std::this_thread::sleep_for(std::chrono::microseconds(100)); // Avoid true busy waiting
+        result = func(device, event);
+    }
+
+    util::EndInjectedCommands();
+
+    return result;
+}
+
 VkResult VulkanReplayConsumerBase::OverrideGetQueryPoolResults(PFN_vkGetQueryPoolResults  func,
                                                                VkResult                   original_result,
                                                                const VulkanDeviceInfo*    device_info,
@@ -4313,6 +4342,12 @@ VkResult VulkanReplayConsumerBase::OverrideGetQueryPoolResults(PFN_vkGetQueryPoo
 
     VkDevice    device     = device_info->handle;
     VkQueryPool query_pool = query_pool_info->handle;
+
+    // If the query was successful at capture time, it MUST be successful at replay time, to avoid sync issues
+    if (original_result == VK_SUCCESS)
+    {
+        flags |= VK_QUERY_RESULT_WAIT_BIT;
+    }
 
     const VkResult result =
         func(device, query_pool, firstQuery, queryCount, dataSize, pData->GetOutputPointer(), stride, flags);
--- a/framework/decode/vulkan_replay_consumer_base.h
+++ b/framework/decode/vulkan_replay_consumer_base.h
@@ -689,6 +689,11 @@ class VulkanReplayConsumerBase : public VulkanConsumer
                                     const VulkanDeviceInfo* device_info,
                                     const VulkanFenceInfo*  fence_info);
 
+    VkResult OverrideGetEventStatus(PFN_vkGetEventStatus    func,
+                                    VkResult                original_result,
+                                    const VulkanDeviceInfo* device_info,
+                                    const VulkanEventInfo*  event_info);
+
     VkResult OverrideGetQueryPoolResults(PFN_vkGetQueryPoolResults  func,
                                          VkResult                   original_result,
                                          const VulkanDeviceInfo*    device_info,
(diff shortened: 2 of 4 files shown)


fabian-lunarg:
> For vkQueryPoolResults, I propose to use VK_QUERY_RESULT_WAIT_BIT, which makes the request blocking.

there already was a back/forth with the exact change you propose.
in short, I thought the same, applied it, shortly after we pinned regressions with existing captures to the change and reverted. the symptom would be an infinite hang in `vkQueryPoolResults` during replay. 
it was my bad I didn't leave an explanatory comment pointing at [the previous attempt](https://github.com/LunarG/gfxreconstruct/pull/2533) (which partially got reverted [here](https://github.com/LunarG/gfxreconstruct/pull/2681)).

for `vkGetEventStatus`, wrt.
> No "max retry" anymore, it continues for as long as necessary

I would fear the same would happen, it could just busy-wait / block on the call infinitely.

the nature of those hanging captures also involved non-determinism, sometimes not showing problems at all.
it really strikes me the order of calls, when subject to multiple threads during capture, is crucial here.
replay is single-threaded/serial, so we basically circumvent waiting on 'signals' that were never (or not yet) emitted.

per-mathisen-arm:
I'm pretty sure that the standard disallows the possibility of infinite hangs for conformant code for both events and queries. That is, you are not allowed to wait for unsubmitted queries, fences or events, and once they are submitted, waiting for them will eventually either complete or hard-fail - either way it will not cause an infinite wait. In what cases have you seen otherwise?

per-mathisen-arm:
On the flip side, *not* waiting *will* cause issues for conformant code for sure and only luck with timings will prevent crashes and corruptions.

fabian-lunarg:
> In what cases have you seen otherwise?

the issue was with an existing capture of an AAA title and was reproduced on windows/AMD.
I don't know of other cases.
it's entirely possible that we were in fact dealing with an ICD issue. 

> not waiting will cause issues for conformant code for sure

I agree and would summarize what happened as:
a perceived regression was pinned to the original commit 
and I could not present an actual, existing problem I solved with `VK_QUERY_RESULT_WAIT_BIT`.
so I think a solid reproduction case would be very valuable.

per-mathisen-arm:
I created a microbenchmark test for this issue: https://github.com/ARM-software/tracetooltests/blob/main/src/vulkan_compute_wait.cpp - run with frame boundary, then screenshot frame 1 to get red image and frame 2 to get green image. If the last image is red, then you hit a synchronization issue.

fabian-lunarg:
> I created a microbenchmark test for this issue: https://github.com/ARM-software/tracetooltests/blob/main/src/vulkan_compute_wait.cpp - run with frame boundary, then screenshot frame 1 to get red image and frame 2 to get green image. If the last image is red, then you hit a synchronization issue.

ok, nice. I was able to confirm your point using this test. 

creating two captures like:

```
export BENCHMARKING_ENABLE_JSON='{"target":"vulkan_compute_wait","capabilities":{"frameless":false}}'
# event capture  → vkGetEventStatus path
GFXRECON_CAPTURE_FILE=event.gfxr ./cmake-build-release/vulkan_compute_wait -wt completion-event -D 0

# query capture  → vkGetQueryPoolResults + WAIT_BIT path
GFXRECON_CAPTURE_FILE=query.gfxr ./cmake-build-release/vulkan_compute_wait -wt query-pool -D 0
```
and replaying with dev vs. this PR proofs the synchronization is corrected here. 
I mean, we agreed on this already, but now it's so much better than my previous handwaving.

this leaves only trying to reproduce the above mentioned hang, but I hope we rule out that concern too.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -93,7 +95,7 @@ const std::unordered_set<std::string> kSurfaceExtensions = {
 const std::unordered_set<std::string> kTrimStateSetupDeviceExtensions = { VK_EXT_SHADER_STENCIL_EXPORT_EXTENSION_NAME };
 
 const std::unordered_set<std::string> kFunctionsAllowedToReturnDifferentCodeThanCapture = {
-    "vkSetDebugUtilsObjectNameEXT", "vkSetDebugUtilsObjectTagEXT", "vkGetQueryPoolResults", "vkGetEventStatus"
```
fabian-lunarg, line 96: removing `vkGetQueryPoolResults` here is fine (because we wait),
but removing `vkGetEventStatus` would reintroduce log-spam we already fixed,
e.g.
`WARNING - vkGetEventStatus returned ... VK_EVENT_SET that does not match the result from the capture file: VK_EVENT_RESET` is a legit+high-frequency case

so, we need to suppress at least that by adding to the `kResultValuesAllowedDifferentCodeThanCapture` LUT-below: `{VK_EVENT_RESET → VK_EVENT_SET}`

marius-pelegrin-arm: Actually I think even for vkGetQueryPoolResults it's not good, because we can be ready when it was not at capture time.
MB, I will fix that.

marius-pelegrin-arm: I will also add vkGetFenceStatus and vkWaitForFences, same reason

marius-pelegrin-arm: Oh, my bad again, I misunderstood what this list was and how the CheckResult function uses it.
Yes, these functions need to be removed from the list and `{VK_EVENT_RESET → VK_EVENT_SET}` needs to be added to the LUT

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4296,6 +4246,32 @@ VkResult VulkanReplayConsumerBase::OverrideGetFenceStatus(PFN_vkGetFenceStatus
...
+
+    VkDevice device = device_info->handle;
+    VkEvent  event  = event_info->handle;
+
+    VkResult result = func(device, event);
+
+    util::BeginInjectedCommands();
+
+    // If the query was successful at capture time, it MUST be successful at replay time, to avoid sync issues
+    while (original_result == VK_EVENT_SET && result == VK_EVENT_RESET)
```
fabian-lunarg: I'd like to suggest an addition here.

the unbounded loop is a risk and I suspect could cause hangs in the wild.
like possible reasons in my mind include:
- capture-thread ordering vs. serialized replay
- ICD quirks
- skipped / failed submissions at replay
- trimmed captures 'could' miss device-set events

short, we should have an escape/timeout here.
it doesn't hurt the normal case but assures we won't hang.
instead of hanging we accept a possible desync in this case and fall back to warn-and-continue.

that should be the best of both worlds.
also we should avoid the magic-numbers and add to the top e.g.:

```
constexpr auto kGetEventStatusPollInterval = std::chrono::microseconds(100);
constexpr auto kGetEventStatusTimeout      = std::chrono::seconds(5);
```
if you're happy with that, I can add a commit

per-mathisen-arm: I would like to challenge you to come up with a an actual scenario in which Vulkan can deadlock here. According to my reading of the spec, this is not supposed to be possible. You cannot submit work that relies on not yet submitted work. All work will eventually resolve, even if we serialize all threads and all work. (And I really wanted to prove that my work on enabling multithreading was necessary, not just a nice to have performance improvement. But I could not.)

I have already implemented submissions skipping (see my blackhole PR), and this also does not raise synchronization issues. And if a queue submit fails only during replay... you already have bigger problems (this is just not recoverable).

Finally, a 5 seconds timeout may sound like an eternity but there are platforms that run not frames per second but hours per frame. I would really want to avoid having to specify different timeout parameter sets for different platforms.

fabian-lunarg: > a 5 seconds timeout may sound like an eternity but there are platforms that run not frames per second but hours per frame

ok, I hadn't considered this case. it's a strong argument against the timeout.

fabian-lunarg: otherwise, alright, you got me convinced. 
the ordering when `vkGetEventStatus` returned `VK_EVENT_SET` during capture **has** to line up.
I was thinking in direction of potential missing/unfinished `vkCmdSetEvent` during trimming.
but that looks fine too in `VulkanStateWriter::WriteState`.

> I really wanted to prove that my work on enabling multithreading was necessary, not just a nice to have performance improvement. But I could not

that's interesting. I was still under the impression we should see issues here, under certain conditions.

finally, I guess my original concern was more with driver behavior on the WAIT_BIT / query-pool path,
and not here.

this leaves just two things for this code-location:
- a `constexpr auto kGetEventStatusPollInterval = std::chrono::microseconds(100);` instead of hardcoding
- util::BeginInjectedCommands(); / util::EndInjectedCommands(); should be avoided when we're not polling

fabian-lunarg:
ok, a new day, some additional input. back to the challenge around `vkGetEventStatus` and the unbounded loop termination.

I raised the concern that the loop 'could' hang infinitely in presence of inflight submissions of `vkCmdSetEvent`.
discussing with @davidd-lunarg he brought forward a similar concern and gave this example:

```
cmdBufSetEvent1 be a command buffer that sets VkEvent event1
cmdBufResetEvent1 be a command buffer that resets VkEvent event1

vkQueueSubmit(cmdBufSetEvent1)
...
vkQueueSubmit(cmdBufResetEvent1)
vkGetEventStatus
    Capture: VK_EVENT_SET (cmdBufResetEvent1 not yet completed)
    Replay: VK_EVENT_RESET (cmdBufResetEvent1 already completed)
```

thinking about @per-mathisen-arm 's statement, we do agree on:

> You cannot submit work that relies on not yet submitted work. All work will eventually resolve, even if we serialize all threads and all work

but that doesn't imply it will resolve on EVENT_SET. in the example we got `vkCmdSetEvent(E)` and `vkCmdResetEvent(E)` both in-flight. No later set (vkCmd/vkSetEvent) re-signals the event after that reset.
same/different cmd-buffers make no difference.
same/different queue make only a minor difference, affecting if the final state is deterministic.

on host-side we invoke `vkGetEventStatus` without any gpu-sync like fences, `vkDeviceWaitIdle`, etc.
its placement relative to the set -> reset transitions is timing-dependent, so 'can' differ in capture vs. replay.
this is a race-condition and could yield an in-between `VK_EVENT_SET` ... or not. 

-> whenever we read EVENT_SET during capture, but the replay sees (the terminal) EVENT_RESET we will loop forever.

fabian-lunarg:
to prove our point I went ahead and turned above scenario into a tracetooltest `vulkan_event_hang`:
-> https://github.com/fabian-lunarg/tracetooltests/blob/fabian_get_event_status_hang/src/vulkan_event_hang.cpp

I captured this test and replayed using this PR-branch, yielding the exact hang we predicted, at:

```
__syscall_cancel_arch
gfxrecon::decode::VulkanReplayConsumerBase::OverrideGetEventStatus vulkan_replay_consumer_base.cpp:4325
gfxrecon::decode::VulkanReplayConsumer::Process_vkGetEventStatus generated_vulkan_replay_consumer.cpp:1249
...
```
so, that leaves us with:
- Per was able to show that the wait/poll is required for correct synchronization.
- we demonstrate that an unbounded loop is unsafe and will hang under certain conditions.
- we already tried to bound the polling via max-retries, which we agree was a failed attempt.
- the only other option to bound it would be timeout, which, when fixed as e.g. 5s, would be problematic in an 'hours per frame' scenario

so, how do we want to bound this? are there other options?

fabian-lunarg:
> so, how do we want to bound this? are there other options?

and I think the answer to that is adding event-tracking during replay.
we need to track all event set/reset, host+device, 2ndary command-buffers and both QueueSubmit flavours.
we would guard the polling based on the tracked event's 'terminal state' (still vulnerable to unsynchronized cross-queue set/reset of same event + host GetEventStatus, but 'seems' very rare)

per-mathisen-arm:
Well, this is interesting. Nice to see a microbenchmark for it! I think this boils down to a fundamental spec issue where a capture layer simply does not have enough information to correctly replay (there should have been a vkWaitForEvent(device, event, state) command instead). Hopefully new code will use timeline semaphores rather than events for host synchronization. I am not sure how we can solve this for existing code. For maximum correctness, we would have a timeout that leads to a vkDeviceWaitIdle ... but that could lead to some very surprising performance issues. Will have to think about it.

fabian-lunarg:
> ... boils down to a fundamental spec issue where a capture layer simply does not have enough information to correctly replay

definitely a shortcoming here wrt. API-design (combined w. the practice to synchronize based on cpu-retrieved event-status).

but for the hang demonstrated above:

> I think the answer to that is adding event-tracking during replay

which I flashed out on another branch, tracking any/all event-set/reset API calls and guarding the loop
when the 'terminal-state' doesn't line up.
-> https://github.com/fabian-lunarg/gfxreconstruct/tree/fabian_event_tracking_poll_guard

using this branch I was able to replay my local capture of the [vulkan_event_hang.cpp](https://github.com/fabian-lunarg/tracetooltests/blob/fabian_get_event_status_hang/src/vulkan_event_hang.cpp) test, circumventing the hang,
issuing warnings instead (about wrong `vkGetEventStatus` return-type).
so I'm suggesting to pick the latest commit from that branch into the PR, but let me examine this a bit more before that.

last but not least, I'm now revisiting said capture that showed blocking-issues w. `vkGetQueryPoolResults`
so we got the complete picture here.

fabian-lunarg:
> last but not least, I'm now revisiting said capture that showed blocking-issues w. vkGetQueryPoolResults
so we got the complete picture here.

-> using a trimmed capture on win11/AMD Radeon (RX 9070 XT, driver: 25.30.17, vulkan-driver 2.0.373), `vkGetQueryPoolResults` w. proposed/added `VK_QUERY_RESULT_WAIT_BIT` hangs right after trim-state-setup.
replay using current dev shows no issues.

the driver is from 08.01.2026. 
I would be interested if the hang is also present on most current drivers, but I'm not able to switch drivers.

either way, this demonstrates the change wrt. `VK_QUERY_RESULT_WAIT_BIT` also introduces risk of hangs
under certain conditions.

fabian-lunarg:
closed in favor of https://github.com/LunarG/gfxreconstruct/pull/3235 (commit got carried over)

PR #2984, opened 2026-06-01
Update test_suite for reenabling traces in commit-suite for windows-nvidia

wzchu-lunarg:


wzchu-lunarg:
Already in commit suite master

PR #2986, opened 2026-06-02
Make GFXRECON_LOG_FATAL actually fatal on non-Android platforms

### LINKED ISSUE #1024: Clean up use of GFXRECON_LOG_FATAL
1. During replay or other tools, GFXRECON_LOG_FATAL should abort or exit the application with a non-success return code.
2. Determine what GFXRECON_LOG_FATAL should do during capture and implement it.
    a. Should it shutdown the captured application?
4. Clean up use of the GFXRECON_LOG_FATAL macro so it is only used in fatal situations where GFXR cannot continue.
5. Make behavior as consistent as possible between platforms.

fabian-lunarg:
- Add `Log::SetFatalCallback()`: on `kFatal`, invoke the callback, followed by a catch-all `std::abort()`.
- Remove the copy-pasted fatal_error_handler_/RaiseFatalError pattern from all four replay consumer classes
- tools set the callback once to throw std::runtime_error instead.

fixes #1024

### DIFF
--- a/framework/decode/custom_dx12_struct_decoders.cpp
+++ b/framework/decode/custom_dx12_struct_decoders.cpp
@@ -1182,14 +1182,14 @@ size_t DecodeStruct(const uint8_t* buffer, size_t buffer_size, Decoded_D3D12_PIP
 
                 if (type == format::kInvalidSubobjectType)
                 {
-                    GFXRECON_LOG_FATAL(
+                    GFXRECON_LOG_ERROR(
                         "A pipeline state subobject encoding indicates that the stream contained an unrecognized "
                         "subobject type during capture and the captured data is incomplete, which may cause replay to "
                         "fail.");
                 }
                 else
                 {
-                    GFXRECON_LOG_FATAL("Pipeline state subobject decoding encountered unrecognized subobject type "
+                    GFXRECON_LOG_ERROR("Pipeline state subobject decoding encountered unrecognized subobject type "
                                        "D3D12_PIPELINE_STATE_SUBOBJECT_TYPE = %d, which may cause replay to fail.",
                                        type);
                 }
--- a/framework/decode/custom_dx12_struct_decoders.cpp
+++ b/framework/decode/custom_dx12_struct_decoders.cpp
@@ -715,9 +715,8 @@ size_t DecodeStruct(const uint8_t*                                              
                 ValueDecoder::DecodeUInt32Value((buffer + bytes_read), (buffer_size - bytes_read), &(value->NumBlocks));
             break;
         default:
-            GFXRECON_LOG_FATAL(
-                "Unrecognized D3D12_SERIALIZED_RAYTRACING_ACCELERATION_STRUCTURE_HEADER1 union type %u",
-                value->HeaderPostambleType);
+            GFXRECON_LOG_FATAL("Unrecognized D3D12_SERIALIZED_RAYTRACING_ACCELERATION_STRUCTURE_HEADER1 union type %u",
+                               value->HeaderPostambleType);
             break;
     }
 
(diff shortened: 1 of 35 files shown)


fabian-lunarg:
fyi @per-mathisen-arm

MarkY-LunarG (approved):

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -921,7 +921,6 @@ void Dx12ReplayConsumerBase::CheckReplayResult(const char* call_name, HRESULT ca
                 call_name,
                 enumutil::GetResultValueString(replay_result).c_str(),
                 enumutil::GetResultValueString(capture_result).c_str());
-            RaiseFatalError(enumutil::GetResultDescription(replay_result));
```
fabian-lunarg, line 925: @davidd-lunarg it would be awesome if you could have a critical look here. 
we want to rule out any unforeseen consequences (e.g. for tools, log-parsing, ..) by changing either log-output or the exception-sequence.

framework/decode/custom_dx12_struct_decoders.cpp
```diff
@@ -638,7 +638,7 @@ size_t DecodeStruct(const uint8_t* buffer, size_t buffer_size, Decoded_D3D12_REN
             // These cases have no additional values to decode.
             break;
         default:
-            GFXRECON_LOG_FATAL_ONCE("Unrecognized D3D12_RENDER_PASS_ENDING_ACCESS union type %u", value->Type);
+            GFXRECON_LOG_ERROR("Unrecognized D3D12_RENDER_PASS_BEGINNING_ACCESS union type %u", value->Type);
```
davidd-lunarg: Should this (and below changes) use `LOG_ERROR_ONCE` to preserve existing behavior?

framework/decode/dx12_replay_consumer_base.cpp
davidd-lunarg: There are a few places in `encode` where `LOG_FATAL` is still called, which would cause the application being captured to exit. Should GFXR be allowed to force-quit the application it is capturing?

```
framework\encode\custom_dx12_struct_encoders.cpp:916
framework\encode\custom_openxr_api_call_encoders.cpp:58
framework\encode\d3d12_capture_manager.cpp:788
framework\encode\d3d12_capture_manager.cpp:3044
framework\encode\d3d12_capture_manager.cpp:3050
```

Also one in a `graphics` util which can be called from either capture or replay.

```
framework\graphics\dx12_resource_data_util.cpp:346
```

fabian-lunarg: > Should GFXR be allowed to force-quit the application it is capturing?

I think we can leave this question open, at least for this PR.

here I think we should prioritize to maintain existing behavior.
since none of the existing `LOG_FATAL` was followed by any abort/exception they now correspond to LOG_ERROR

fabian-lunarg: I added another commit (re-)visiting leftover locations.
in some cases needed to decide/understand what the original intention was (often it seems we just leaned into the fact FATAL was not fatal).
when in doubt I demoted to ERROR

framework/decode/custom_dx12_struct_decoders.cpp
fabian-lunarg: done

tools/replay/replay_feature.h
```diff
@@ -128,10 +128,7 @@ class ReplayFeature : public ReplayFeatureBase
         application_    = application;
     }
 
-    void FinalizeConsumer()
-    {
-        replay_consumer_->SetFatalErrorHandler([](const char* message) { throw std::runtime_error(message); });
-    }
+    void FinalizeConsumer() {}
```
davidd-lunarg, line 131: This removes `SetFatalErrorHandler` from the common replay consumer creation code path. Now it is set in `Log::SetFatalCallback` in `desktop_main.cpp`. Does it also need to be set in `android_main.cpp` to preserve Android behavior?

fabian-lunarg: you are right, this was overlooked. fixed now

davidd-lunarg (approved):

bradgrantham-lunarg (approved):

PR #2988, opened 2026-06-02
Add --blackhole feature to Vulkan replayer

per-mathisen-arm:
This feature allows going through all the API calls without submitting any work to the GPU. Useful for quick surface-level testing of trace files with heavy GPU work, and for quickly estimating how CPU-bound a trace might be.

Based on existing implementation I did of this feature in the lavatube project. Ported by codex first to gfxreconstruct-arm and then ported to upstream.


per-mathisen-arm:
Note that if #2978 is merged first, then vkGetEventStatus needs to be changed to _not_ be blocking when running in blackhole mode, as vkCmdSetEvent will then never be run and it will wait forever.

PR #2989, opened 2026-06-02
Remove Android-specific build files

MarkY-LunarG:
Merge Android and desktop build files where possible.


android/test/test_apps/common/CMakeLists.txt
```diff
@@ -33,14 +33,14 @@ target_include_directories(gfxrecon-testapp-base
...
-add_subdirectory(${FRAMEWORK_DIR}/graphics ${FRAMEWORK_DIR}/graphics/build/${ANDROID_ABI})
-add_subdirectory(${FRAMEWORK_DIR}/format ${FRAMEWORK_DIR}/format/build/${ANDROID_ABI})
-add_subdirectory(${FRAMEWORK_DIR}/application ${FRAMEWORK_DIR}/application/build/${ANDROID_ABI})
+set(FRAMEWORK_BUILD_DIR ${GFXRECON_ANDROID_DIR}/framework/build/test_apps)
+add_subdirectory(${GFXRECON_SOURCE_DIR}/framework/util ${FRAMEWORK_BUILD_DIR}/util/${ANDROID_ABI})
+add_subdirectory(${GFXRECON_SOURCE_DIR}/framework/plugin ${FRAMEWORK_BUILD_DIR}/plugin/${ANDROID_ABI})
+add_subdirectory(${GFXRECON_SOURCE_DIR}/framework/decode ${FRAMEWORK_BUILD_DIR}/decode/${ANDROID_ABI})
+add_subdirectory(${GFXRECON_SOURCE_DIR}/framework/graphics ${FRAMEWORK_BUILD_DIR}/graphics/${ANDROID_ABI})
+add_subdirectory(${GFXRECON_SOURCE_DIR}/framework/format ${FRAMEWORK_BUILD_DIR}/format/${ANDROID_ABI})
+add_subdirectory(${GFXRECON_SOURCE_DIR}/framework/application ${FRAMEWORK_BUILD_DIR}/application/${ANDROID_ABI})
```
charles-lunarg, line 43: This change gives me pause. Not because of the change itself, I believe it correctly maps the old `<root>/android/framework/<tool>/CMakeLists.txt` to the correct `<root>/framework/<tool>/CMakeListst.txt` but because it means we are putting binaries outside of the build directory. Not something we can/should fix right now, but something I wanted to bring up.

CMakeLists.txt
```diff
@@ -28,6 +28,8 @@
 cmake_minimum_required(VERSION 3.16.3)
 project(GFXReconstruct LANGUAGES CXX)
 
+set(GFXRECON_SOURCE_DIR "${PROJECT_SOURCE_DIR}")
```
charles-lunarg, line 31: So my gut instinct is that this line isn't necessary.

Looking at the project, android defines `GFXRECON_SOURCE_DIR` because GFXR's 'root' CMakeLists.txt isn't the initial CMakeLists.txt android uses, but that `android/layer/CMakeLists.txt` is. In order to find everything we had to define a variable to hold that root, hence GFXRECON_SOURCE_DIR.

And it would greatly reduce the number of lines changed in this PR, since we can use `PROJECT_SOURCE_DIR` everywhere instead of sometimes using it, sometimes not.

charles-lunarg (commented):
One question about putting build files in the source tree, and a comment about replacing GFXRECON_SOURCE_DIR with PROJECT_SOURCE_DIR.

CMakeLists.txt
MarkY-LunarG: Good point.  I'll try updating it to use PROJECT_SOURCE_DIR instead.

MarkY-LunarG: It's because each Android CMakeLists.txt file defines a new project (for example the layer is `project(VkLayer_gfxreconstruct)` and the multi-win-replay is `project(gfxrecon-replay`.  So PROJECT_SOURCE_DIR changes once you enter building those targets.  And we can't use CMAKE_SOURCE_DIR because that fails when GFXReconstruct is built inside of another project.  So we really do need a parent directory defined globally.

charles-lunarg: The docs state that PROJECT_SOURCE_DIR doesn't change when project() is called in add_subdirectory(). 

> Note, it is not affected by calls to project() made within a child directory scope (i.e. from within a call to add_subdirectory() from the current scope).

https://cmake.org/cmake/help/latest/variable/PROJECT_SOURCE_DIR.html

Edit: It's one of those weird cmake isms that is counterintuitive. No doubt about that.

android/test/test_apps/common/CMakeLists.txt
MarkY-LunarG: Android files are already built weird.  They are built under the Android folder and then the target (like layer).  So while I agree, it's not making major changes to the design.  In fact, if I just set `FRAMEWORK_BUILD_DIR` to "CMAKE_BINARY_DIR/android/test_apps" it produces `android/test/test_apps/launcher/.cxx/Debug/332t5q47/arm64-v8a/android/test_apps`.  So you see it's still building it under the `android` source folder.  It's weird and in my opinion wrong, but this keeps it close to the original result.

framework/encode/CMakeLists.txt
```diff
@@ -178,17 +178,18 @@ endif()
...
+                               ${GFXRECON_SOURCE_DIR}/framework
                                $<$<BOOL:${D3D12_SUPPORT}>:${D3D12_INCLUDE_DIRECTORIES}>)
 
 target_link_libraries(gfxrecon_encode
                       gfxrecon_graphics
                       gfxrecon_format
                       gfxrecon_util
                       vulkan_registry
-                      OpenXR
+                      $<$<NOT:$<BOOL:${ANDROID}>>:OpenXR>
```
bradgrantham-lunarg: Should this be ` $<$<AND:$<BOOL:${OPENXR_SUPPORT_ENABLED}>,$<NOT:$<BOOL:${ANDROID}>>>:OpenXR>` like it is for `decode`?

CMakeLists.txt
MarkY-LunarG: I know when I made the change to use PROJECT_SOURCE_DIR instead, it failed inside of multiple source folders.  I didn't spend a lot of time, but I remember it being an issue for ARM for the mentioned reason.  I can take a look at it again.

framework/encode/CMakeLists.txt
MarkY-LunarG: Also see this bug in the application CMakeLists.txt as well.  I'll update both.

CMakeLists.txt
charles-lunarg: Best not to spend a lot of time on it - its a nitpick after all.

Thank you for trying!

android/layer/CMakeLists.txt
```diff
@@ -27,10 +27,10 @@ else()
     message(STATUS "OpenXR support disabled!")
 endif()
 
-add_subdirectory(../framework/util ${PROJECT_SOURCE_DIR}/../framework/util/build/layer/${ANDROID_ABI})
```
antonio-lunarg, line 30: Need to delete `android/framework/decode` and `android/framework/util` directories?

framework/util/CMakeLists.txt
antonio-lunarg: `PROJECT_SOURCE_DIR`->`GFXRECON_SOURCE_DIR`

antonio-lunarg (approved):

CMakeLists.txt
per-mathisen-arm: PROJECT_SOURCE_DIR also allows gfxreconstruct to be compiled as a submodule of another git repo. Why use a separate android CMakeLists file at all, when you could use cmake presets instead?

marius-pelegrin-arm:
Tested the change internally and it works for us  (@per-mathisen-arm)

MarkY-LunarG:
Applied changes suggested by @antonio-lunarg

bradgrantham-lunarg (approved):

PR #2991, opened 2026-06-02
Add DispatchSetEnvironmentVariablesCommand to openxr decoder

wzchu-lunarg:
Openxr's decoder was missing the `DispatchSetEnvironmentVariablesCommand` which causes convert for `tutorial4_capture_frames_1_through_500.gfxr` to be missing block 2 completely, as it had meta function `SetEnvironmentVariablesCommand`.

Replicate the same `DispatchSetEnvironmentVariablesCommand` found in the vulkan decoder in the openxr decoder.

### DIFF
--- a/framework/decode/openxr_decoder_base.cpp
+++ b/framework/decode/openxr_decoder_base.cpp
@@ -74,6 +74,15 @@ void OpenXrDecoderBase::DispatchDisplayMessageCommand(format::ThreadId thread_id
     }
 }
 
+void OpenXrDecoderBase::DispatchSetEnvironmentVariablesCommand(const format::SetEnvironmentVariablesCommand& header,
+                                                               const char*                                   env_string)
+{
+    for (auto consumer : consumers_)
+    {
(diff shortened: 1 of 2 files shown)


MarkY-LunarG (approved):
Cool, good catch

mikes-lunarg:
What happens if a trace contains both Vulkan and OpenXR? Will we get duplicated blocks in the convert output from both dispatches running? @MarkY-LunarG this may be a side effect of the `gfxrecon-convert` API split where we now have to figure out how to handle non-API metacommands.

MarkY-LunarG:
@mikes-lunarg, the `SetEnvironmentVariables` meta-data command has an API ID associated with it.  So it would require two different meta-data commands one with each ID.  It really depends on which API was detected first to which one is used for encoding the environment variable meta-command API ID.

mikes-lunarg:
Ahh thanks, I see now where it gets encoded during capture file creation. Carry on!

PR #2992, opened 2026-06-02
Update test_suite for reenabling traces in commit-suite for android

wzchu-lunarg:


wzchu-lunarg:
Merged into master from https://github.com/LunarG/gfxreconstruct/pull/3061

PR #2994, opened 2026-06-03
gen: Parallelize Vulkan generation

antonio-lunarg:
Local test:

- Before: `time python3 framework/generated/generate_vulkan.py`  18.81s user 0.90s system 99% cpu **19.810 total**
- After: `time python3 framework/generated/generate_vulkan.py`  27.73s user 1.63s system 1145% cpu **2.564 total**

### DIFF
--- a/framework/generated/generate_vulkan.py
+++ b/framework/generated/generate_vulkan.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 #
 # Copyright (c) 2018 Valve Corporation
-# Copyright (c) 2018-2023 LunarG, Inc.
+# Copyright (c) 2018-2026 LunarG, Inc.
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to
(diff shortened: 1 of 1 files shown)


framework/generated/generate_vulkan.py
```diff
@@ -148,34 +178,11 @@
...
-        subprocess.check_call(
-            gencode_args,
-            shell=False,
-            env=env,
-            cwd=SCRIPT_DIR,
-        )
+
+    target_generator = TargetGenerator(registry_path, video_path, args.headers_dir, env)
+    with ThreadPoolExecutor() as executor:
+        executor.map(target_generator.generate_target, generate_targets)
```
fabian-lunarg: I think like this with `executor.map`, potential exceptions are stored in those futures and are never retrieved/re-raised. there could be generator-error and the script still exits 0 ?

the idiomatic way I stumbled upon 'should' be a simple change:

wrapping with `list(...)` forces to pull the iterator's result and makes exceptions surface 
`list(executor.map(target_generator.generate_target, generate_targets))`

framework/generated/generate_vulkan.py
```diff
@@ -148,34 +171,14 @@
...
-            gencode_args,
-            shell=False,
-            env=env,
-            cwd=SCRIPT_DIR,
-        )
+
+    gen = functools.partial(
+        generate_target, registry_path, video_path, args.headers_dir, env
+    )
+    with ThreadPoolExecutor() as executor:
```
fabian-lunarg: `ThreadPoolExecutor()` with no argument defaults to `cpu_count + 4`, 
but would be inefficient, since all threads are cpu-bound.

better we explicitly cap: `with ThreadPoolExecutor(max_workers=os.cpu_count()) as executor:`

we could make it user-controllable (consistent with `build.py`'s `-j`/`--parallel 0` convention):

```python
arg_parser.add_argument('-j', '--parallel', dest='jobs', type=int, default=0,
    help='Parallel jobs. 0 = all CPUs.')
# ...
max_workers = args.jobs or os.cpu_count()
with ThreadPoolExecutor(max_workers=max_workers) as executor:
    list(executor.map(gen, generate_targets))
```

the `os.cpu_count()` cap  is the more important part.

The `-j` form would be a nice-to-have for local use, but also useful to force sequential processing
(e.g. confusing when during dev multiple generator 'would' fail and the log-output is subject to threading. then I would welcome the option to turn mt off)

fabian-lunarg (approved):
the code-changes look good to me, CI passes. all concerns addressed.

PR #2997, opened 2026-06-03
extract: Split out API features

MarkY-LunarG:
Split out the API-specific functionality into separate feature classes (just like info and convert).

NOTE: This does not include OpenXR because it is not a graphics but a compositing API.  So that feature is intentionally left out.

### DIFF
--- a/tools/extract/CMakeLists.txt
+++ b/tools/extract/CMakeLists.txt
@@ -31,24 +31,44 @@ add_executable(gfxrecon-extract "")
 target_sources(gfxrecon-extract
                PRIVATE
                    ${CMAKE_CURRENT_LIST_DIR}/main.cpp
-                    ${CMAKE_CURRENT_LIST_DIR}/../platform_debug_helper.cpp
-                    $<$<BOOL:WIN32>:${PROJECT_SOURCE_DIR}/version.rc>
-                    $<$<BOOL:WIN32>:${GFXRECON_WIN_MANIFEST}>
+                   ${CMAKE_CURRENT_LIST_DIR}/extract_feature.h
+                   ${CMAKE_CURRENT_LIST_DIR}/extract_d3d12_feature.cpp
+                   ${CMAKE_CURRENT_LIST_DIR}/extract_vulkan_feature.cpp
+                   ${CMAKE_CURRENT_LIST_DIR}/../platform_debug_helper.cpp
+                   ${CMAKE_CURRENT_LIST_DIR}/../tool_settings.h
+                   $<$<BOOL:WIN32>:${PROJECT_SOURCE_DIR}/version.rc>
+                   $<$<BOOL:WIN32>:${GFXRECON_WIN_MANIFEST}>
               )
 
 if (MSVC)
     # Force inclusion of "gfxrecon_disable_popup_result" variable in linking.
     # On 32-bit windows, MSVC prefixes symbols with "_" but on 64-bit windows it doesn't.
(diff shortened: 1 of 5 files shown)


MarkY-LunarG:
@davidd-lunarg , I added you for reviewing the D3D12 side of things.  The visual comparison passes, but I would like you to verify.

tools/extract/extract_vulkan_feature.cpp
```diff
@@ -0,0 +1,207 @@
...
+        decode::HandlePointerDecoder<VkPipeline>*                                   pPipelines) override
+    {
+        if ((returnValue >= 0) && (pCreateInfos != nullptr) && !pCreateInfos->IsNull())
+        {
+            for (size_t i = 0; i < createInfoCount; i++)
+            {
+                auto& pipeline_create_info = pCreateInfos->GetPointer()[i];
+                for (size_t j = 0; j < pipeline_create_info.stageCount; j++)
+                {
+                    auto& stage_create_info = pipeline_create_info.pStages[i];
```
mikes-lunarg: Wasn't you, but this should be `pStages[j]`!!!

tools/extract/main.cpp
```diff
@@ -76,460 +69,16 @@ static void PrintUsage(const char* exe_name)
...
     }
 
-  private:
-    std::string extract_dir_;
-};
-#endif
-
-int main(int argc, const char** argv)
-{
     gfxrecon::util::Log::Init();
```
mikes-lunarg: `Log::Init()` should be done first thing in main in case any of the code above it needs to print an error

tools/extract/main.cpp
mikes-lunarg: `dir_location` is unsigned, should be `if (dir_location != std::string::npos)`

tools/extract/main.cpp
```diff
@@ -21,30 +21,23 @@
...
-#include "generated/generated_dx12_consumer.h"
-#include "generated/generated_dx12_decoder.h"
-#endif
+#include "extract_feature.h"
 #include "util/argument_parser.h"
+#include "util/feature_module_registry.h"
 #include "util/file_path.h"
 #include "util/logging.h"
 
 #include "vulkan/vulkan.h"
```
mikes-lunarg: Can this be removed?

tools/extract/CMakeLists.txt
```diff
@@ -46,7 +50,7 @@ if (MSVC)
     endif()
 endif()
 
-target_include_directories(gfxrecon-extract PUBLIC ${CMAKE_BINARY_DIR})
```
mikes-lunarg, line 49: Really curious why CMAKE_BINARY_DIR was in here, doesn't seem right at all

mikes-lunarg (changes requested):
Just a couple of small things, some of them aren't even your fault!

tools/extract/extract_vulkan_feature.cpp
MarkY-LunarG: This was actually updated by Antonio.  So the change is done his way in the split now.

MarkY-LunarG:
@mikes-lunarg , all changes made.

tools/extract/CMakeLists.txt
MarkY-LunarG: IDK, I removed it to see what happens, and nothing failed to build.  So I'm keeping it removed.

mikes-lunarg (approved):

PR #2998, opened 2026-06-04
Fix push-constants when pipeline is bound after vkCmdPushConstants

### LINKED ISSUE #2899: [replay] Buffer device addresses are broken
If an app uses buffer device addresses and does:
```
vkCmdBindPipeline()
vkCmdPushConstants()
```
it works, but if the app does:
```
vkCmdPushConstants()
vkCmdBindPipeline()
```
The address will not be correctly updated in the replay and will get a device lost. The issue is in `VulkanAddressReplacer::ProcessCmdPushConstants()`, this function iterates over pipelines and updates the addresses:
`for (const auto& [bind_point, pipeline_id] : command_buffer_info->bound_pipelines)`
But if the pipeline has not yet been bound, this function will do nothing

fabian-lunarg:
relevant when using '-m rebind'.
If an app calls vkCmdPushConstants before vkCmdBindPipeline, the address replacer had no pipeline to consult and emitted the original capture-time buffer-device-addresses verbatim. Fixed by adding ProcessCmdBindPipeline to VulkanAddressReplacer. re-process stored push-constant data against the newly bound pipeline and injects a corrective vkCmdPushConstants into the command buffer if anything got changed/corrected.

Fixes #2899

### DIFF
--- a/framework/decode/vulkan_address_replacer.cpp
+++ b/framework/decode/vulkan_address_replacer.cpp
@@ -641,6 +641,44 @@ void VulkanAddressReplacer::ProcessCmdPushConstants(const VulkanCommandBufferInf
     }
 }
 
+void VulkanAddressReplacer::ProcessCmdBindPipeline(VulkanCommandBufferInfo*                  command_buffer_info,
+                                                   const decode::VulkanDeviceAddressTracker& address_tracker)
+{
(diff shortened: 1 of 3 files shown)


ziga-lunarg (approved):
Fixes my issue!

MarkY-LunarG (approved):
Ugh, good catch.

PR #3000, opened 2026-06-04
Fix some command line parameters

olehkuznetsov:
Those don't have arguments:
--idle-before-submit
--serialize-render-passes
--serialize-queue-submissions

Those should be in kOptions otherwise next command line argument is ignored


bradgrantham-lunarg:
@davidlunarg Can you plan adding single short test cases to CI that exercise each of these options?  I think extended would be fine.

antonio-lunarg (approved):

bradgrantham-lunarg (approved):

PR #3004, opened 2026-06-04
Tighten wording around capture of API calls

bradgrantham-lunarg:


docs/GFXReconstruct Architectural Principles.md
```diff
@@ -58,9 +58,13 @@ If an exception is made to a principle, there should be a well-considered, docum
 
 ## RECONSTRUCT 
 
-**A capture file should contain as much as possible all of the unmodified graphics API calls made by the application for the captured, trimmed range of frames.**
+**A capture file should contain as much as possible all of the unmodified graphics API call context made by the application for the captured, trimmed range of frames.**
 
-It may contain more commands represented by Metadata commands but it should be possible to **reconstruct** the original app’s sequence of API commands from the captured “api call” blocks.
+The capture file should preserve the original application API call, parameters, object associations, and observed results as captured.  This data should be stored as an ApiCall or MethodCall block when possible, representing the API call in its original form.
```
MarkY-LunarG: Maybe "Each API call of the original application should be preserved inside of the capture file in the proper sequence it was made and contain properly recorded parameters, object associations and results."  The reason I suggest it is because when I start reading it, the sentence sounds like it's relative to all contents of the capture file, but you're identifying the API call specifically and how it's recorded.

docs/GFXReconstruct Architectural Principles.md
```diff
@@ -58,9 +58,13 @@ If an exception is made to a principle, there should be a well-considered, docum
...
+
+When necessary, Metadata commands may encode additional information as a modifier for a recorded API call or may substitute an abstraction.  If the Metadata substitutes an abstraction, that abstraction must represent the API call object, parameters, and result so that it should be possible to **reconstruct** the original app’s sequence of API commands from the captured data.
```
MarkY-LunarG, line 65: Perhaps give an example of such an "abstraction"?

bradgrantham-lunarg: An example abstraction is mentioned immediately in "Implications".

docs/GFXReconstruct Architectural Principles.md
bradgrantham-lunarg: I think this has to be read in context with the next couple of sentences, stating explicitly that ApiCall and MethodCall blocks are preferred, but Metadata commands are an alternative when appropriate.  I've changed the paragraph to:
```
The capture file should preserve the original application API calls and parameters, object associations, and observed results for each call.  This data should be stored as an ApiCall or MethodCall block when possible, representing each API call in its original form.
```
How about that?

bradgrantham-lunarg: @MarkY-LunarG

MarkY-LunarG (approved):

PR #3006, opened 2026-06-05
fix: Pipeline indexing in extract tool

antonio-lunarg:


MarkY-LunarG (approved):
I like this.  Thanks for not just fixing the i -> j but making the names cleaner too.

PR #3007, opened 2026-06-05
Decoder args

antonio-lunarg:
Large decoder/consumer API refactor: this PR introduces args::* aggregate parameter structs for Vulkan decoder callbacks and generated consumers, and rewrites a large swath of hand-written and generated decode/replay/resource-tracking code to pass a single mutable args object instead of long parameter lists. It also adds generated support for those structs via generated_vulkan_decoder_args.h, wires that file into generation/CMake, and updates manual handlers for descriptor update templates, ray tracing pipeline creation, queue submit/bind/memory requirement/resource tracking paths, frame-loop replay, stats, JSON export, and referenced-resource/block consumers.


framework/decode/vulkan_consumer_base.h
```diff
@@ -51,44 +53,26 @@ class VulkanConsumerBase : public CommonConsumerBase
 
     virtual ~VulkanConsumerBase() {}
 
-    virtual void Process_vkUpdateDescriptorSetWithTemplate(const ApiCallInfo&               call_info,
-                                                           format::HandleId                 device,
-                                                           format::HandleId                 descriptorSet,
-                                                           format::HandleId                 descriptorUpdateTemplate,
-                                                           DescriptorUpdateTemplateDecoder* pData)
+    virtual void Process_vkUpdateDescriptorSetWithTemplate(const ApiCallInfo&                   call_info,
```
jzulauf-lunarg: Here and all following, 

1) The naming is redundant. Process(ApiCallInfo, UpdateDescriptorSetWithTemplateArgs) is unambiguous, though one could argue, less readable.

2) The *Args seem to be insufficiently scoped in framework::decode, I'd be tempted to put them into a namespace.

framework/decode/vulkan_consumer_base.h
```diff
@@ -51,44 +53,26 @@ class VulkanConsumerBase : public CommonConsumerBase
...
-        const ApiCallInfo&                                               call_info,
-        VkResult                                                         returnValue,
-        format::HandleId                                                 device,
-        format::HandleId                                                 deferredOperation,
-        format::HandleId                                                 pipelineCache,
-        uint32_t                                                         createInfoCount,
-        StructPointerDecoder<Decoded_VkRayTracingPipelineCreateInfoKHR>* pCreateInfos,
-        StructPointerDecoder<Decoded_VkAllocationCallbacks>*             pAllocator,
-        HandlePointerDecoder<VkPipeline>*                                pPipelines)
+    virtual void Process_vkCreateRayTracingPipelinesKHR(const ApiCallInfo&                call_info,
```
jzulauf-lunarg: Interface design question: Why not include  the return value in the arg pack ... it's the expected return, correct?

framework/decode/vulkan_cpp_consumer_base.cpp
```diff
@@ -3705,61 +3705,50 @@ void VulkanCppConsumerBase::ProcessDestroyHardwareBufferCommand(uint64_t buffer_
...
 
 void VulkanCppConsumerBase::Process_vkUpdateDescriptorSetWithTemplate(const ApiCallInfo& call_info,
-                                                                      format::HandleId   device,
-                                                                      format::HandleId   descriptorSet,
-                                                                      format::HandleId   descriptorUpdateTemplate,
-                                                                      DescriptorUpdateTemplateDecoder* pData)
+                                                                     UpdateDescriptorSetWithTemplateArgs& args)
 {
-    Generate_vkUpdateDescriptorSetWithTemplate(device, descriptorSet, descriptorUpdateTemplate, pData);
+    Generate_vkUpdateDescriptorSetWithTemplate(args.device, args.descriptorSet, args.descriptorUpdateTemplate, &args.pData);
```
jzulauf-lunarg: Take a look at how the DispatchVistor::DispatchArgs used GetTuple and std::apply to avoid the manual unpacking.

As the *Args are code gen'd adding a tuple accessor(s) would be straight forward.

framework/decode/vulkan_decoder_args.h
```diff
@@ -0,0 +1,78 @@
...
+#include "vk_video/vulkan_video_codec_h264std_encode.h"
+#include "vk_video/vulkan_video_codec_h265std.h"
+#include "vk_video/vulkan_video_codec_h265std_decode.h"
+#include "vk_video/vulkan_video_codec_h265std_encode.h"
+#include "vk_video/vulkan_video_codecs_common.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(decode)
+
+struct UpdateDescriptorSetWithTemplateArgs
```
jzulauf-lunarg: Yeah... I think a nested namespace here...

framework/generated/khronos_generators/vulkan_generators/vulkan_decoder_args_header_generator.py
```diff
@@ -0,0 +1,114 @@
...
+            args_struct_name = VulkanBaseGenerator.make_args_struct_name(cmd)
+            params: list[ValueInfo] = self.all_cmd_params[cmd][2]
+
+            body = "\n"
+            body += f"struct {args_struct_name}\n"
+            body += "{\n"
+
+            for param in params:
+                decoded_type = self.make_decoded_param_type(param)
+                body += f"    {decoded_type} {param.name};\n"
```
jzulauf-lunarg, line 123: This is where the GetTuple -ish generation would go... or even a method to call the appropriate Processes_<foo>(consumer, <rest of of args>...)

jzulauf-lunarg (commented):
Looking really interesting.  A couple thoughts.  Doesn't handle the stack relative reference case patterns, but those only exist in D3D and OpenXR...

framework/decode/vulkan_consumer_base.h
antonio-lunarg: My rationale is that a return value is not an argument, but there is no real benefit with keeping it separate, on the contrary it is convenient to include it imo.

jzulauf-lunarg:
@antonio-lunarg  I didn't see the updated generate source associated with python you pushed for XR...

antonio-lunarg:
> @jzulauf-lunarg  I didn't see the updated generate source associated with python you pushed for XR...

I am only focusing on Vulkan to limit the scope of this PR.

framework/decode/vulkan_consumer_base.h
antonio-lunarg: How about `args::UpdateDescriptorSetWithTemplate`?

framework/decode/vulkan_consumer_base.h
antonio-lunarg: Included.

framework/decode/vulkan_cpp_consumer_base.cpp
antonio-lunarg: I opted for passing `args` directly.

framework/generated/khronos_generators/vulkan_generators/vulkan_decoder_args_header_generator.py
antonio-lunarg: I generated it, but its signature is slightly different than what functions exect.
While GetTuple returns references to the fields of the args structure, most functions expect pointers, rather than references.

framework/decode/vulkan_consumer_base.h
bradgrantham-lunarg: @jzulauf-lunarg

bradgrantham-lunarg:
@jzulauf-lunarg Did Antonio address your feedback?

framework/decode/vulkan_consumer_base.h
```diff
@@ -51,44 +53,24 @@ class VulkanConsumerBase : public CommonConsumerBase
 
     virtual ~VulkanConsumerBase() {}
 
-    virtual void Process_vkUpdateDescriptorSetWithTemplate(const ApiCallInfo&               call_info,
-                                                           format::HandleId                 device,
-                                                           format::HandleId                 descriptorSet,
-                                                           format::HandleId                 descriptorUpdateTemplate,
-                                                           DescriptorUpdateTemplateDecoder* pData)
+    virtual void Process_vkUpdateDescriptorSetWithTemplate(const ApiCallInfo&                     call_info,
+                                                           args::UpdateDescriptorSetWithTemplate& args)
```
jzulauf-lunarg, line 59: Namespace... vulkan::args or vulkan_args?  @bradgrantham-lunarg constraining API specific types to api specific namesspaces?

framework/decode/vulkan_cpp_consumer_base.cpp
```diff
@@ -592,247 +592,221 @@ void VulkanCppConsumerBase::AddHandles(const std::string& outputName, const form
...
         struct_info->imageExtent.height = GetProperWindowHeight(struct_info->imageExtent.height);
     }
     std::string pcreate_info_struct = GenerateStruct_VkSwapchainCreateInfoKHR(
-        stream_pcreate_info, struct_info, pCreateInfo->GetMetaStructPointer(), *this);
+        stream_pcreate_info, struct_info, args.pCreateInfo.GetMetaStructPointer(), *this);
     fprintf(file, "%s", stream_pcreate_info.str().c_str());
     std::string pswapchain_name = "pSwapchain_" + std::to_string(this->GetNextId(VK_OBJECT_TYPE_SWAPCHAIN_KHR));
-    AddKnownVariables("VkSwapchainKHR", pswapchain_name, pSwapchain->GetPointer());
-    if (returnValue == VK_SUCCESS)
+    AddKnownVariables("VkSwapchainKHR", pswapchain_name, args.pSwapchain.GetPointer());
```
jzulauf-lunarg, line 636: Struct member naming issue.  Leaving the name as "pSwapchain" when the args::<foo> doesn't store the HandlePointerDecoder<VkSwapchainKHR> as a pointer is very confusing to the maintainer.  Need to strip `p` off *all* struct member names that are no longer pointers.

Yes, this is a pain, sorry.

framework/generated/generated_vulkan_decoder_args.h
```diff
@@ -0,0 +1,7918 @@
...
+GFXRECON_BEGIN_NAMESPACE(args)
+
+struct CreateInstance
+{
+    VkResult result;
+    StructPointerDecoder<Decoded_VkInstanceCreateInfo> pCreateInfo;
+    StructPointerDecoder<Decoded_VkAllocationCallbacks> pAllocator;
+    HandlePointerDecoder<VkInstance> pInstance;
+
+    auto GetTuple() const { return std::tie(result, pCreateInfo, pAllocator, pInstance); }
```
jzulauf-lunarg, line 75: Since the interface has been changed to accept args:: structs directly the GetTuple methods are unneeded, untested, and require `<tuple>` more broadly.  We don't need it.

framework/decode/vulkan_decoder_args.h
```diff
@@ -0,0 +1,82 @@
+#ifndef GFXRECON_DECODE_VULKAN_DECODER_ARGS_H
```
jzulauf-lunarg, line 23: Doesn't appear to be codegen'd, also doesn't have copyright. 

1) why not code gen
2) needs copyright

framework/decode/vulkan_decoder_args.h
```diff
@@ -0,0 +1,82 @@
...
+
+struct UpdateDescriptorSetWithTemplate
+{
+    format::HandleId                device;
+    format::HandleId                descriptorSet;
+    format::HandleId                descriptorUpdateTemplate;
+    DescriptorUpdateTemplateDecoder pData;
+};
+
+struct UpdateDescriptorSetWithTemplateKHR
```
jzulauf-lunarg, line 58: Can we use `using` here to avoid duplication, especially these hand coded structs.

framework/generated/khronos_generators/vulkan_generators/vulkan_decoder_args_header_generator.py
jzulauf-lunarg: Given that you decided to change the interface, the gettuple isn't needed.

framework/generated/khronos_generators/vulkan_generators/vulkan_decoder_args_header_generator.py
```diff
@@ -0,0 +1,140 @@
...
+
+            body += "\n    auto GetTuple() const { return std::tie("
+            if return_type and return_type != 'void':
+                body += "result, "
+            for param in params:
+                body += f"{param.name}, "
+            body = body[:-2]  # Remove the trailing comma and space.
```
jzulauf-lunarg, line 130: Assumes there's something to trim without test.

However, GetTuple() is dead code — so. If the generator stops emitting GetTuple() entirely, this whole block goes away with it, there isn't an issue.

jzulauf-lunarg (changes requested):
A few things.  Naming and Deadcode reduction are the important ones

framework/decode/vulkan_cpp_consumer_base.cpp
antonio-lunarg: I agree with this, but I would push for handling this in a follow-up PR. Reason is that it is not trivial. The name comes from the XML registry and stripping the p naively has various unwanted side effects. I believe it can be done, but it's an entire new task.

framework/decode/vulkan_decoder_args.h
antonio-lunarg: The structures in this files were already not generated and scattered in other source files. I moved them here for consistency.

antonio-lunarg:
@jzulauf-lunarg rebased

jzulauf-lunarg (approved):
All comments addressed.

PR #3008, opened 2026-06-08
Fix async pipeline creation when using VkPipelineCache

fabian-lunarg:
- a regression caused all pipelines using VkPipelineCache to compile sequentially
- was only required for VK_PIPELINE_CACHE_CREATE_EXTERNALLY_SYNCHRONIZED_BIT
-> fixed by properly applying 'requires_external_synchronization' flag

- reduce log-spam by demoting two WARNING to WARNING_ONCE

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -12021,8 +12021,8 @@ bool VulkanReplayConsumerBase::SkipPipelineCreationForCompileRequired(VkResult  
 
     if (original_result == VK_PIPELINE_COMPILE_REQUIRED)
     {
-        GFXRECON_LOG_WARNING("Skipping execution of VkPipeline creation due to original result being "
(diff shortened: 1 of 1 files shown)


mikes-lunarg (approved):
Thank you for fixing the log spam too!

bradgrantham-lunarg:
@fabian-lunarg Could we exercise this in CI somehow?  Would we essentially check the runtime?

fabian-lunarg:
> @fabian-lunarg Could we exercise this in CI somehow? Would we essentially check the runtime?

we certainly do exercise runs using `--pcj X` but that's only a smoke-test.

I think what's missing includes:
- to be able to judge if multiple threads were in fact used we would need two identical runs, with / without --pcj X
- the more pipelines the capture compiles, the more we should see a clear diff in replay wall-clock time.
- driver pipeline-caches should be cold
- no other processes should skew the measurement

so, we would actually be benchmarking in order to test if correct code-paths were exercises.

maybe another approach is more feasible. 
we could instrument the async-code with simple debug-logging, run the trace with matching log-settings
and confirm their presence in output.

PR #3009, opened 2026-06-08
Add capture and replay support for VK_ARM_tensors and VK_ARM_data_graph

### LINKED ISSUE #2291: Add support for VK_ARM_tensors and VK_ARM_data_graph
we [updated the vulkan-header](https://github.com/LunarG/gfxreconstruct/pull/2284) for next SDK-release, but blacklisted two newly introduced extensions.
basic support (at least codegen) is expected to not cause a lot of work.
an example is given by arm's branch (https://github.com/arm-software/gfxreconstruct-arm).

fabian-lunarg:
Adds initial gfxreconstruct support for two ARM Vulkan extensions:

  - `VK_ARM_tensors`: VkTensorARM / VkTensorViewARM objects and their memory
    binding, descriptor writes (VK_DESCRIPTOR_TYPE_TENSOR_ARM), and trimmed
    capture state reconstruction.
  - `VK_ARM_data_graph`: VkDataGraphPipelineARM creation, VkDataGraphPipelineSessionARM
    creation/memory binding, and vkCmdDispatchDataGraphARM recording.

  Both extensions are removed from `kVulkanUnsupportedDeviceExtension` so the
  layer no longer strips them from device feature negotiation.

## Capture (encode)
  - Hand-written encoder for vkCreateDataGraphPipelinesARM to handle the
    FAIL_ON_COMPILE_REQUIRED early-out and the group-create pattern. 
  - PostProcess hooks for vkBindTensorMemoryARM,
    vkBindDataGraphPipelineSessionMemoryARM, and vkCreateTensorViewARM in
    VulkanCaptureManager.
  - State tracker additions: TrackTensorMemoryBinding,
    TrackDataGraphPipelineSessionMemoryBinding,
    VK_DESCRIPTOR_TYPE_TENSOR_ARM descriptor tracking, DestroyState
    overloads, and VkObjectType debug name/tag handling for all three new
    types.
  - kBindPoint_data_graph added to PipelineBindPoints.
  - InitializeState / InitializeGroupObjectState specialisations for
    VkTensorARM, VkTensorViewARM, and VkDataGraphPipelineSessionARM.

## Trimmed capture (state writer)
  
  - WriteTensorState / WriteTensorMemoryState emit vkCreateTensorARM and
    vkBindTensorMemoryARM for all live tensors at trim start.
  - WriteTensorSnapshotState / ProcessTensorMemory upload tensor contents
    via a new kInitTensorCommand metadata block
  - WriteDataGraphPipelineSessionMemoryState reconstructs all session memory
    bindings, including recreation of pipelines (and their shader module /
    pipeline layout dependencies) that were destroyed before trim start.
  - VK_DESCRIPTOR_TYPE_TENSOR_ARM descriptor write reconstruction in the
    descriptor set state writer.

## Replay (decode)
  
  - ProcessInitTensorCommand handler uploads tensor data via the resource
    initializer on trimmed-capture replay.
  - OverrideCreateTensorARM, OverrideBindTensorMemoryARM,
    OverrideCreateDataGraphPipelineSessionARM,
    OverrideBindDataGraphPipelineSessionMemoryARM,
    OverrideDestroyTensorARM, OverrideDestroyDataGraphPipelineSessionARM,
    OverrideCreateDataGraphPipelinesARM, and
    OverrideGetTensorMemoryRequirementsARM added to the replay consumer.
  - OverrideCreateDataGraphPipelinesARM checks optical-flow pipeline
    compatibility against the replay device's reported formats.
  - Trimmed-capture replay adds TRANSFER_SRC/DST usage bits to tensor
    create infos so the resource initializer can upload snapshot data.

  ## Allocator support
  
  - VulkanDefaultAllocator: pass-through implementations for all new
    resource and memory-binding entry points.
  - VulkanRebindAllocator: full rebind implementations for tensor and
    data graph session memory, including GetTensorMemoryUsage heuristic,
    AllocateMemoryForTensor (with VMA), and BindDataGraphPipelineSessionMemory
    (queries per-binding requirements from the driver at replay time).

  ## Code generator
  
  - Blacklists / replay_overrides updated for all hand-written entry points.
  - Generator fixes for VkBaseOutStructure encoder/decoder wiring and
    decoder header template.
  - Struct decoder, encoder, handle mapper, handle wrapper, deep-copy, and
    JSON generators extended for all new VK_ARM_tensors / VK_ARM_data_graph
    struct types.

  ## Limitations
  
  - **Memory aliasing**: Deferred to a follow-up PR.

 ## Reproduction

- the changes have been tested using a linux Release-build of ARM's ai-ml-emulation-layer:
  https://github.com/arm/ai-ml-emulation-layer-for-vulkan/releases/tag/v0.9.0

- capture/replay is fine for all vulkan_tensor samples from this repository:
  https://github.com/ARM-software/tracetooltests

fixes #2291


framework/decode/custom_vulkan_struct_decoders.cpp
```diff
@@ -756,5 +757,40 @@ size_t DecodeStruct(const uint8_t* buffer, size_t buffer_size, Decoded_VkDescrip
...
+            uint32_t pointer_attrib = 0;
+            bytes_read +=
+                ValueDecoder::DecodeUInt32Value(buffer + bytes_read, buffer_size - bytes_read, &pointer_attrib);
+            size_t size = 0;
+            bytes_read += ValueDecoder::DecodeSizeTValue((buffer + bytes_read), (buffer_size - bytes_read), &size);
+            wrapper->pConstantData = DecodeAllocator::Allocate<uint8_t>(size);
+            bytes_read += ValueDecoder::DecodeUInt8Array(
+                (buffer + bytes_read), (buffer_size - bytes_read), (void*)wrapper->pConstantData, size);
+
+            value->pConstantData = wrapper->pConstantData;
```
MarkY-LunarG: This seems odd.  It's under a "description != nullptr" check, but it's not usinng decription at all.  Seems to at least justify a comment explaining what's going on here.

framework/decode/vulkan_default_allocator.cpp
```diff
@@ -966,6 +966,145 @@ VkResult VulkanDefaultAllocator::Allocate(const VkMemoryAllocateInfo*  allocate_
     return result;
 }
 
+VkResult
+VulkanDefaultAllocator::CreateDataGraphPipelineSession(const VkDataGraphPipelineSessionCreateInfoARM* create_info,
+                                                       const VkAllocationCallbacks*   allocation_callbacks,
+                                                       format::HandleId               capture_id,
+                                                       VkDataGraphPipelineSessionARM* data_graph_pipeline_session,
+                                                       ResourceData*                  allocator_data)
```
MarkY-LunarG, line 974: Did you run clang-format on this?

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -3300,5 +3305,578 @@ void VulkanRebindAllocator::ClearStagingResources()
...
+    GFXRECON_UNREFERENCED_PARAMETER(allocation_callbacks);
+    GFXRECON_UNREFERENCED_PARAMETER(capture_id);
+
+    VkResult result = VK_ERROR_INITIALIZATION_FAILED;
+
+    if ((create_info != nullptr) && (tensor != nullptr) && (allocator_data != nullptr))
+    {
+        result = functions_.create_tensor(device_, create_info, allocator_->GetAllocationCallbacks(), tensor);
+
+        if (result >= 0)
```
MarkY-LunarG, line 3370: This doesn't require a cast?

framework/graphics/vulkan_resources_util.cpp
```diff
@@ -2199,6 +2199,20 @@ VkResult VulkanResourcesUtil::ReadFromBufferResource(
...
+VkResult VulkanResourcesUtil::ReadFromTensorResource(VkTensorARM                   tensor,
+                                                     const VkTensorDescriptionARM* desc,
+                                                     uint32_t                      queue_family_index,
+                                                     std::vector<uint8_t>&         data)
+{
+    GFXRECON_UNREFERENCED_PARAMETER(tensor);
+    GFXRECON_UNREFERENCED_PARAMETER(desc);
+    GFXRECON_UNREFERENCED_PARAMETER(queue_family_index);
+    GFXRECON_UNREFERENCED_PARAMETER(data);
+    GFXRECON_LOG_WARNING_ONCE("Tensor trim snapshot via staging copy is not yet implemented; "
```
MarkY-LunarG: We should create an issue to track this and mention it in a comment above the warning.

framework/generated/generated_vulkan_struct_decoders.h
```diff
@@ -12722,6 +13035,104 @@ typedef Decoded_VkPipelineShaderStageRequiredSubgroupSizeCreateInfo Decoded_VkSh
...
+        Decoded_VkBaseOutStructure* return_type = nullptr;
+
+        uint32_t peek_structure_type = 0;
+        ValueDecoder::DecodeUInt32Value(buffer, buffer_size, &peek_structure_type);
+        VkStructureType vk_type = static_cast<VkStructureType>(peek_structure_type);
+
+        switch (vk_type)
+        {
+            default:
+                return_type = DecodeAllocator::Allocate<Decoded_VkBaseOutStructure>(len, initialize);
```
MarkY-LunarG, line 13393: We should log a warning_once here.

framework/generated/generated_vulkan_struct_decoders.h
```diff
@@ -12722,6 +13035,104 @@ typedef Decoded_VkPipelineShaderStageRequiredSubgroupSizeCreateInfo Decoded_VkSh
...
+        size_t bytes_read = 0;
+
+        uint32_t peek_structure_type = 0;
+        ValueDecoder::DecodeUInt32Value(buffer, buffer_size, &peek_structure_type);
+        VkStructureType vk_type = static_cast<VkStructureType>(peek_structure_type);
+
+        switch (vk_type)
+        {
+            default:
+                bytes_read += DecodeStruct((buffer + bytes_read), (buffer_size - bytes_read), dest);
```
MarkY-LunarG, line 13417: Same here.

framework/generated/khronos_generators/vulkan_generators/vulkan_base_generator.py
```diff
@@ -229,6 +227,30 @@ class VulkanBaseGenerator(KhronosBaseGenerator):
     Base class for Vulkan API parameter encoding and decoding generators.
     """
 
+    BASE_OUT_STRUCTURE_TYPE_INFO_OVERRIDES = {
+        ('vkGetPhysicalDeviceQueueFamilyDataGraphEngineOperationPropertiesARM', 'pProperties'): (
+            'VkQueueFamilyDataGraphOpticalFlowPropertiesARM',
+            'VkQueueFamilyDataGraphProcessingEnginePropertiesARM',
+        ),
+    }
```
MarkY-LunarG, line 234: Ideally, we could get Arm to document this in the XML file. I don't know if that's possible and we wouldn't have to hard-code it since other companies will start adding their own structs.  Otherwise, we'll end up having to check  for EVERY struct that has a pNext  (just like the pNext handling code).

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -13604,5 +13660,574 @@ void VulkanReplayConsumerBase::OverrideCmdExecuteGeneratedCommandsEXT(
...
+        if ((result != VK_SUCCESS) || (data_graph_property_count == 0))
+        {
+            continue;
+        }
+
+        std::vector<VkQueueFamilyDataGraphPropertiesARM> queue_family_data_graph_properties(
+            data_graph_property_count,
+            VkQueueFamilyDataGraphPropertiesARM{ VK_STRUCTURE_TYPE_QUEUE_FAMILY_DATA_GRAPH_PROPERTIES_ARM, nullptr });
+
+        result = instance_table->GetPhysicalDeviceQueueFamilyDataGraphPropertiesARM(
```
MarkY-LunarG, line 14010: Do we need to label "injected" commands around these various calls which do the properties queries?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -13604,5 +13660,574 @@ void VulkanReplayConsumerBase::OverrideCmdExecuteGeneratedCommandsEXT(
...
+    auto                                  capture_id    = (*tensor->GetPointer());
+
+    auto                  replay_create_info   = pCreateInfo->GetPointer();
+    VkTensorCreateInfoARM modified_create_info = *replay_create_info;
+
+    auto* tensor_info = reinterpret_cast<VulkanTensorARMInfo*>(tensor->GetConsumerData(0));
+    GFXRECON_ASSERT(tensor_info != nullptr);
+
+    VkTensorDescriptionARM modified_description_storage;
+    if (replaying_trimmed_capture_)
```
MarkY-LunarG, line 14192: Why are you adding this?  At least a comment.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -13604,5 +13660,574 @@ void VulkanReplayConsumerBase::OverrideCmdExecuteGeneratedCommandsEXT(
...
+        tensor_info->allocator_data = allocator_data;
+        tensor_info->tiling         = replay_create_info->pDescription->tiling;
+        tensor_info->format         = replay_create_info->pDescription->format;
+        tensor_info->dimensionCount = replay_create_info->pDescription->dimensionCount;
+        tensor_info->usage          = replay_create_info->pDescription->usage;
+        tensor_info->pDimensions.resize(tensor_info->dimensionCount);
+        tensor_info->pStrides.clear();
+
+        for (uint32_t i = 0; i < tensor_info->dimensionCount; ++i)
+            tensor_info->pDimensions[i] = replay_create_info->pDescription->pDimensions[i];
```
MarkY-LunarG, line 14215: Nit: Braces here and on the next for loop as well.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -13604,5 +13660,574 @@ void VulkanReplayConsumerBase::OverrideCmdExecuteGeneratedCommandsEXT(
...
+
+        if (replay_create_info->pDescription->pStrides != nullptr)
+        {
+            tensor_info->pStrides.resize(tensor_info->dimensionCount);
+            for (uint32_t i = 0; i < tensor_info->dimensionCount; ++i)
+                tensor_info->pStrides[i] = replay_create_info->pDescription->pStrides[i];
+        }
+
+        VkTensorMemoryRequirementsInfoARM tensor_mem_req{};
+        tensor_mem_req.sType  = VK_STRUCTURE_TYPE_TENSOR_MEMORY_REQUIREMENTS_INFO_ARM;
```
MarkY-LunarG: Why not just init struct with sType above in initializer (below too).

framework/encode/custom_vulkan_api_call_encoders.cpp
```diff
@@ -950,5 +950,125 @@ VKAPI_ATTR VkResult VKAPI_CALL vkGetImageViewOpaqueCaptureDescriptorDataEXT(
...
+        pPipelines);
+
+    return result;
+}
+
+VKAPI_ATTR VkResult VKAPI_CALL vkGetTensorOpaqueCaptureDescriptorDataARM(VkDevice,
+                                                                         const VkTensorCaptureDescriptorDataInfoARM*,
+                                                                         void*)
+{
+    return VK_SUCCESS;
```
MarkY-LunarG, line 1067: Should these have a TODO/Issue associated with them?

framework/encode/vulkan_state_tracker_initializers.h
```diff
@@ -570,6 +570,97 @@ InitializeGroupObjectState<VkDevice,
...
+    GFXRECON_UNREFERENCED_PARAMETER(parent_handle);
+    GFXRECON_UNREFERENCED_PARAMETER(secondary_handle);
+
+    wrapper->create_call_id    = create_call_id;
+    wrapper->create_parameters = std::move(create_parameters);
+
+    const VkBaseInStructure* current = reinterpret_cast<const VkBaseInStructure*>(create_info->pNext);
+    while (current != nullptr)
+    {
+        if (current->sType == VK_STRUCTURE_TYPE_DATA_GRAPH_PIPELINE_SHADER_MODULE_CREATE_INFO_ARM)
```
MarkY-LunarG, line 598: Don't we have template  queries to get the struct if it's in the pNext chain?

MarkY-LunarG (commented):
Additional comments from Claude:

```
  Code Review Findings

  [
    {
      "file": "framework/decode/vulkan_rebind_allocator.cpp",
      "line": 3875,
      "summary": "GetTensorMemoryRequirementsARM stores capture requirements from uninitialized struct before calling the Vulkan function that populates it.",
      "failure_scenario": "resource_alloc_info->capture_mem_reqs is assigned from memory_requirements->memoryRequirements at line 3875, but the Vulkan function 
  get_tensor_memory_requirements (which fills memory_requirements) is not called until line 3878. The stored capture requirements are whatever garbage was in the caller's 
  stack-allocated VkMemoryRequirements2. During replay, VMA uses these wrong requirements to allocate backing memory, causing allocation failures or size mismatches."
    },
    {
      "file": "framework/encode/vulkan_capture_manager.h",
      "line": 1779,
      "summary": "PostProcess_vkBindDataGraphPipelineSessionMemoryARM dereferences GetWrapper() result without a null check.",
      "failure_scenario": "GetWrapper() at line 1775 returns the DataGraphPipelineSessionARMWrapper for pBindInfos[i].session. If the session handle is invalid or not yet wrapped (e.g.
  trimmed capture resumes mid-frame), wrapper is null and wrapper->memory_bindings.begin() at line 1779 crashes. The analogous PostProcess_vkBindBufferMemory includes an equivalent 
  null guard."
    },
    {
      "file": "framework/decode/vulkan_rebind_allocator.cpp",
      "line": 3812,
      "summary": "BindTensorMemory silently masks per-tensor failures when binding multiple tensors: the result variable is overwritten by each iteration, so an early failure is erased
  if a later bind succeeds.",
      "failure_scenario": "With two tensors: AllocateMemoryForTensor fails for tensor[0] (result = VK_ERROR_*). The loop does not return early; iteration i=1 calls 
  AllocateMemoryForTensor and succeeds (result = VK_SUCCESS). The function returns VK_SUCCESS, and the caller assumes all binds succeeded. Contrast with 
  BindDataGraphPipelineSessionMemory, which returns immediately on error."
    },
    {
      "file": "framework/decode/vulkan_replay_consumer_base.cpp",
      "line": 13680,
      "summary": "Hint format compatibility check uses create_info.flowVectorFormat against info.hint_formats instead of a hint-specific format field.",
      "failure_scenario": "VkDataGraphPipelineOpticalFlowCreateInfoARM has no hintFormat field. hint_formats is populated by querying 
  VK_DATA_GRAPH_OPTICAL_FLOW_IMAGE_USAGE_HINT_BIT_ARM formats, which can differ from the output flow vector formats. The check at line 13680 validates whether the flow vector output 
  format is in the hint format list — semantically wrong. A pipeline where the hint image format differs from flowVectorFormat will either incorrectly fail compatibility or incorrectly
  pass, leading to a creation error or a wrong-device fallback at replay time."
    },
    {
      "file": "framework/encode/vulkan_state_tracker_initializers.h",
      "line": 743,
      "summary": "pDimensions is dereferenced inside a dimensionCount loop without a null guard.",
      "failure_scenario": "When TrackStruct processes a VkTensorDescriptionARM where dimensionCount > 0 but pDimensions == nullptr (malformed capture or fuzzing), the access 
  desc->pDimensions[i] at line 743 is a null pointer dereference. Per spec conventions dimensionCount > 0 implies non-null pDimensions, but this is a decode/state-restore path that 
  processes potentially untrusted .gfxr file data."
    },
    {
      "file": "framework/decode/vulkan_replay_consumer_base.cpp",
      "line": 13734,
      "summary": "data_graph_optical_flow_initialized flag is read and set without synchronization, creating a latent TOCTOU race.",
      "failure_scenario": "Two threads both hit OverrideCreateDataGraphPipelinesARM simultaneously on the same physical device for the first time. Both see 
  data_graph_optical_flow_initialized == false, both enter InitializeReplayDataGraphOpticalFlowInfo, both call clear() and then populate data_graph_optical_flow_infos concurrently, 
  resulting in a data race on the vector. Currently single-threaded replay prevents this, but it is a latent bug with no protection against future multi-threaded replay dispatch."
    }
  ]
```

Severity:
 1) Critical
 2 - 3) High
4-5) Medium
6) Low

Bugs 1–3 are the most actionable: #1 will silently corrupt tensor memory allocation sizes on every rebind-allocator replay, #2 is an unconditional crash if a handle is invalid, and #3 means multi-tensor bind errors go unreported.

framework/decode/vulkan_default_allocator.cpp
fabian-lunarg: I thought so, CI-check looked fine. quickly checking ...

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: think just an obsolete habit (originally meant to dodge a msvc issue with aggregate-init),
so no particular reasoning here.

framework/encode/vulkan_state_tracker_initializers.h
fabian-lunarg: we do have `graphics::vulkan_struct_get_pnext`, which gives the first matching struct.  let me check why it's not used here.
maaaybe because multiple (duplicate type) pNext-structs are a possible here

framework/decode/custom_vulkan_struct_decoders.cpp
fabian-lunarg: right, we only require a check if it's present but don't need the particulars.
so drop description, simplify to `if(gfxrecon::graphics::vulkan_struct_get_pnext<VkTensorDescriptionARM>(value))`

fabian-lunarg: done

framework/decode/vulkan_default_allocator.cpp
fabian-lunarg: yes, clang-format just looks weird here :)

framework/decode/vulkan_rebind_allocator.cpp
fabian-lunarg: good question. 
I think no, `VkResult` has strong guarantee to be a `int32_t`.
firstly, it's an enum, but contains negative values and an enum-size trick forces it to be 32-bit wide (the
  _MAX_ENUM = 0x7FFFFFFF)

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: done

framework/decode/vulkan_replay_consumer_base.cpp

framework/generated/generated_vulkan_struct_decoders.h
fabian-lunarg: done, adjusted in codegen

framework/generated/generated_vulkan_struct_decoders.h
fabian-lunarg: done

framework/encode/vulkan_state_tracker_initializers.h
fabian-lunarg: yes, right. iteration is manual because of allowed/expected struct-type duplicates

fabian-lunarg:
> Additional comments from Claude:

went through the list, applied corrections where required (some false positives)

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: same pattern as OverrideCreateBuffer/Image. we inject TRANSFER_DST so we can upload data. 
will add a comment

fabian-lunarg: done

framework/encode/custom_vulkan_api_call_encoders.cpp
fabian-lunarg: no, we won't require any action here. those entry-points are specific to descriptor-buffers w. tensors.
defined in [VkPhysicalDeviceDescriptorBufferTensorFeaturesARM](https://docs.vulkan.org/refpages/latest/refpages/source/VkPhysicalDeviceDescriptorBufferTensorFeaturesARM.html).

this group of capture/replay query-routines make no sense to capture (we only require the create-info pNext, which will contain the data queried here).
added a log_warning_once, analog to existing descriptor-buffer entry-points like `vkGetBufferOpaqueCaptureDescriptorDataEXT`

framework/graphics/vulkan_resources_util.cpp
fabian-lunarg: I added the missing pieces (required additions to ResourceUtil, along with some consolidation-refactor).
so instead of an follow-up-issue there is now an additional commit which requires review

MarkY-LunarG (approved):
Your new commit looks good too.

Issue #2880, opened 2026-04-17
[capture] Counter Strike 2 on VK assertion fails and crashes


0xCCCbb:
**Describe the capture bug:**
Counter Strike 2 crashes soon after entering main menu (possibly while loading UI map) and prints 
```
/usr/include/c++/15.2.1/bits/unique_ptr.h:734: typename std::add_lvalue_reference<_Tp>::type std::unique_ptr<_Tp [], _Dp>::operator[](std::size_t) const [with _Tp = long unsigned int; _Dp = std::default_delete<long unsigned int []>; typename std::add_lvalue_reference<_Tp>::type = long unsigned int&; std::size_t = long unsigned int]: Assertion 'get() != pointer()' failed.
```

**Build Environment:**
```
pacman -Qi gfxreconstruct-git 
Name            : gfxreconstruct-git
Version         : 1.4.341.0.r193.g0e368c94f-1
```

**Capture Settings:**
```bash
export VK_INSTANCE_LAYERS=VK_LAYER_LUNARG_gfxreconstruct
export GFXRECON_CAPTURE_TRIGGER=F3
```

**To Reproduce:**
Install 730 appid in steam
Subscribe to animgraph 2 beta in appid settings
Set settings to low
Set Video mode for the game lower than the native
quit and reload the game and apply capture variables
Straightly launch and wait till it crashes

**Screenshots:**
Last moment before crash
<img width="1920" height="1200" alt="Image" src="https://github.com/user-attachments/assets/ce167264-c158-44e4-a88a-1b6994d43daa" />

**System environment:**
 gfx90c, Mesa 26.0.5, Arch Linux with 6.19.12 kernel

**Title configuration:**
  - Counter-Strike 2
  - Animgraph 2 Beta
  - Steam ID 730
  - Is DXR enabled? NO

**Additional information (optional):**
  - The last known commit when it worked is not older than 2 weeks.

ghost:
AG2 went out of beta. gfxrecon has been updated to the latest commit. Game crashes only when the capture trigger envvar is present.

wzchu-lunarg:
Can confirm this issue on my Arch Linux system with:
`Mesa 26.1.2`
`Radeon RX 6800 XT`

Same assertion failure in `unique_ptr.h` shortly after launch.

wzchu-lunarg:
https://github.com/LunarG/gfxreconstruct/pull/3012

PR #3010, opened 2026-06-08
Update TrackUpdateDescriptorSetWithTemplate assert

wzchu-lunarg:
Updates TrackUpdateDescriptorSetWithTemplate to instead check whether either `uniform_texel_buffer_views` or `storage_texel_buffer_views` is not `nullptr`.

While looking into [2880](https://github.com/LunarG/gfxreconstruct/issues/2880), running CS2 on debug mode on windows with NVIDIA RTX 5060 Laptop, I get caught on the assert where `uniform_texel_buffer_views` is `nullptr` even though it is currently processing the storage buffer instead as `immutable_buffer` was false, and `storage_texel_buffer_views` was valid.

It should not only be checking `uniform_texel_buffer_views`, but rather the one that will later be used in `dst_info` buffer view.  

This however still doesn't address issue 2880.

### DIFF
--- a/framework/encode/vulkan_state_tracker.cpp
+++ b/framework/encode/vulkan_state_tracker.cpp
@@ -1558,7 +1558,8 @@ void VulkanStateTracker::TrackUpdateDescriptorSetWithTemplate(VkDescriptorSet   
             {
                 auto& binding = wrapper->bindings[current_binding];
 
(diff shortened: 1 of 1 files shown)


MarkY-LunarG (approved):
Looks good!

PR #3011, opened 2026-06-08
Update test refs for async testing

jzulauf-lunarg:
test.ref isn't tip of Tests, but next most recent commit, supporting async testing.


jzulauf-lunarg:
@mikes-lunarg @bradgrantham-lunarg  -- this would complete the initial async feature and test.

bradgrantham-lunarg (approved):

PR #3016, opened 2026-06-11
Isolate render passes during replay

antonio-lunarg:
Add `VulkanCommandSplitter` that splits a command buffer at each render pass boundary and submits the segments separately, chained through a semaphore. Hook it into OverrideCmdBegin/EndRenderPass and OverrideQueueSubmit, and let injected submit jobs opt out by returning a null semaphore.


per-mathisen-arm:
I'm curious how this is supposed to work, as each commandbuffer has its own state that gets dropped once it is done executing? It would be a really useful feature, though.

antonio-lunarg:
> I'm curious how this is supposed to work, as each commandbuffer has its own state that gets dropped once it is done executing? It would be a really useful feature, though.

Yep, I am planning to resubmit those commands for restoring the state after each split. WIP.

framework/decode/vulkan_command_splitter.cpp
```diff
@@ -0,0 +1,358 @@
...
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(decode)
+
+VulkanCommandBufferSplitInfo::VulkanCommandBufferSplitInfo(const VulkanDeviceInfo*            device_info,
+                                                           const graphics::VulkanDeviceTable* device_table,
+                                                           CommonObjectInfoTable*             object_table,
+                                                           format::HandleId                   command_buffer_id) :
+    device_info_(device_info),
+    device_table_(device_table), object_table_(object_table), split_semaphore_(device_info, device_table),
+    original_handle(VK_NULL_HANDLE)
```
MarkY-LunarG: `original_handle` should end with an underscore if it's a member.

framework/decode/vulkan_command_splitter.h
```diff
@@ -0,0 +1,153 @@
...
+    const graphics::VulkanDeviceTable* device_table_ = nullptr;
+    CommonObjectInfoTable*             object_table_ = nullptr;
+
+    VulkanInjectedSemaphore split_semaphore_;
+
+    VkCommandBuffer original_handle;
+
+    /// Associated command buffer handles, namely command buffers created at split-time that are associated with the
+    /// original command buffer. This does not include the original handle itself.
+    std::vector<VkCommandBuffer> associated_handles;
```
MarkY-LunarG: Should end in _

MarkY-LunarG: Same for all members below as well.

framework/decode/vulkan_command_splitter.cpp
```diff
@@ -0,0 +1,358 @@
...
+        VulkanCommandPoolInfo* pool_info = object_table_->GetVkCommandPoolInfo(command_buffer_info->pool_id);
+
+        const VulkanDeviceInfo* device_info = object_table_->GetVkDeviceInfo(command_buffer_info->parent_id);
+        GFXRECON_ASSERT(device_info == device_info_);
+
+        // Create a new handle for the split.
+        VkCommandBufferAllocateInfo alloc_info = { VK_STRUCTURE_TYPE_COMMAND_BUFFER_ALLOCATE_INFO };
+        alloc_info.commandPool                 = pool_info->handle;
+        alloc_info.level                       = VK_COMMAND_BUFFER_LEVEL_PRIMARY;
+        alloc_info.commandBufferCount          = 1;
```
MarkY-LunarG: Why not just create a bunch if this mode is active?  Like 5-10 and add them directly to associated_handles_?

framework/decode/vulkan_command_splitter.cpp
```diff
@@ -0,0 +1,358 @@
...
+        // Update the command buffer info to use the original handle for subsequent calls.
+        command_buffer_info->handle = split_info->ResetSplitHandles();
+    }
+}
+
+void VulkanCommandSplitter::EndFrame()
+{
+    // TODO: Need to consider what happens on a subsequent frame:
+    // when an app that re-records the same primary command buffer each frame via an implicit reset
+    // (vkBeginCommandBuffer on a RESET_COMMAND_BUFFER_BIT pool) never triggers ResetCommandBuffer.
```
MarkY-LunarG: Maybe we should log a warning here stating that this is currently not supported or requires additional work?

framework/decode/vulkan_command_splitter.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+        // Store the new handle in the associated handles vector and update the index for next time.
+        associated_handles.push_back(next_handle);
+    }
+
+    next_handle = associated_handles[next_associated_index];
+    ++next_associated_index;
+
+    // Store current handle in split vector.
+    split_handles.push_back(command_buffer_info->handle);
```
MarkY-LunarG: Won't we write the same command buffer twice then?  Once in BeginRenderPass and once in EndRenderPass?

MarkY-LunarG (changes requested):
A few comments and questions

framework/decode/vulkan_replay_consumer_base.cpp
MarkY-LunarG, line 10689: Did you need to add the same changes  from `OverrideCmdEndRenderPass` to `OverrideCmdEndRenderPass2`?

framework/decode/vulkan_command_splitter.cpp
antonio-lunarg: Implemented.

framework/decode/vulkan_command_splitter.cpp
antonio-lunarg: Done

framework/decode/vulkan_replay_consumer_base.cpp
antonio-lunarg: Done, also for dynamic rendering.

MarkY-LunarG (approved):

PR #3019, opened 2026-06-12
Treat uvec2 as a potential BDA type

mikes-lunarg:
vkd3d can pass 64-bit buffer device addresses as a 2-component vectors of unsigned 32-bit integers (uvec2). This is apparently a common lowering and explicitly allowed by the GL_EXT_buffer_reference_uvec2 GLSL extension. Therefore, treat uvec2 values (and arrays) as potential BDAs in the forward SPIRV-reflect pass of
SpirVParsingUtil::ParseBufferReferences.

Add a test case to the SpirVParsingUtil for both uvec2 and uvec2[] paths.


fabian-lunarg (approved):
nice!

PR #3023, opened 2026-06-15
job: Handle empty submit array

antonio-lunarg:
Add a Submit() helper that runs the plan's jobs without injecting semaphores, and early-out to it when submit_infos is empty.

### DIFF
--- a/framework/decode/vulkan_submit_job.h
+++ b/framework/decode/vulkan_submit_job.h
@@ -339,6 +339,15 @@ class VulkanSubmitJobExecution
     void SerializeExecution(std::span<VkSubmitInfo2> submit_infos2);
 
   private:
+    /**
+     * @brief  Just submit the jobs in the plan without modifying any other submit infos.
+     *
+     * @param  plan             plan with submit-indexed jobs to submit.
+     *
+     * @note   Any signal semaphores returned by the jobs will be ignored.
+     */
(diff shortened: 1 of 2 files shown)


framework/decode/vulkan_submit_job.h
```diff
@@ -339,6 +339,13 @@ class VulkanSubmitJobExecution
     void SerializeExecution(std::span<VkSubmitInfo2> submit_infos2);
 
   private:
+    /**
+     * @brief  Just submit the jobs in the plan without modifying any other submit infos.
+     *
+     * @param  plan             plan with submit-indexed jobs to submit.
+     */
+    void Submit(VulkanSubmitJobPlan plan) const;
```
fabian-lunarg: minor comment. 
of course this is a private method, but since you already got a doc-string:
it's only called when `if (submit_infos.empty())` (related to frame-warmup?) and I think only then it's correct to drop the returned semaphores.
maybe you could leave a comment/doc making this more clear to readers
or person creating custom 'jobs' in the future.

fabian-lunarg (approved):
only had a minor comment

framework/decode/vulkan_submit_job.h
antonio-lunarg: Renamed to `SubmitStandalone` and added a note about discarded signal semaphores if any.

PR #3024, opened 2026-06-15
Add _WIN32 support for Windows platform detection

Daveshammer:
MSVC defines _WIN32 by default but not WIN32, causing Windows specific code to be excluded
Change #if defined(WIN32) to #if defined(WIN32) || defined(_WIN32) to support both macros"

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -363,6 +363,7 @@ endif (UNIX AND NOT APPLE)
 add_library(windows_specific INTERFACE)
 
 target_compile_definitions(windows_specific INTERFACE
+        WIN32
         WIN32_LEAN_AND_MEAN
         NOMINMAX
         VK_USE_PLATFORM_WIN32_KHR


bradgrantham-lunarg:
@Daveshammer Thank you for your PR!  We're probably getting `WIN32` from the Visual Studio setup or maybe CMake and works fine for us so this doesn't solve a problem anyone's reported to us.  (try putting `#error in win32` after any `#if defined(WIN32)` in unmodified `main` branch and it should error out for you.)  Are you building under a particularly interesting config like MINGW or something?

@davidd-lunarg  We have `def.*\<WIN32\>` (def followed by identifier WIN32 so it doesn't match _WIN32) across 55 files in 130 places.  Testing `_WIN32`  instead does seem like the *right* way to do this, so can you take a look and see if you agree with it?  If so, and if Daveshammer isn't available, I'll make an issue; this seems like something `sed -i` could fix for us very easily.

davidd-lunarg:
@bradgrantham-lunarg Yes, I agree with you. `_WIN32` seems to be the recommended check so I support making an issue for the conversion. Or if we don't want to touch all those files we could define `WIN32` in CMake when building on Windows or in a macro in `gfxrecon::util`

Daveshammer:
Thanks @bradgrantham-lunarg  and @davidd-lunarg  for the suggestions!

Based on the discussion, I've decided to add `WIN32` in CMake (windows_specific target) instead of bulk-replacing with `_WIN32` across the codebase. This seems to be the safest approach - it works alongside the existing `_WIN32` macro, avoids touching third-party libraries, requires only a one-line change, and ensures better compatibility across different Windows toolchains and build environments.

bradgrantham-lunarg:
> Thanks @bradgrantham-lunarg and @davidd-lunarg for the suggestions!
> 
> Based on the discussion, I've decided to add `WIN32` in CMake (windows_specific target) instead of bulk-replacing with `_WIN32` across the codebase. This seems to be the safest approach - it works alongside the existing `_WIN32` macro, avoids touching third-party libraries, requires only a one-line change, and ensures better compatibility across different Windows toolchains and build environments.

I think we'll lean towards formalizing the use of `_WIN32` in our codebase just for correctness.

But we aren't seeing any build issues with WIN32 / _WIN32.  @Daveshammer Can you detail the problem you're seeing?

Daveshammer:
Hi @bradgrantham-lunarg ,

The build issues we're seeing are related to the use of WIN32 vs _WIN32 macros in the platform detection logic, which ultimately leads to the dlfcn.h inclusion errors on Windows.

Specifically, the framework/util/platform.h file currently uses #if defined(WIN32) to detect Windows. However, when we configure the build using:

bash
```
cd gfxreconstruct
mkdir build
cmake . -Bbuild -G "Visual Studio 17 2022" -A x64 -DCMAKE_CXX_FLAGS="/EHsc /WX- /wd4819 /wd4530" -DCMAKE_C_FLAGS="/EHsc /WX- /wd4819"
```
The MSVC compiler (version 19.44.35221.0, Visual Studio 2022 17.12) does not define WIN32 automatically. It only defines _WIN32 by default, not WIN32. Because of this, the condition #if defined(WIN32) evaluates to false during our Windows builds. As a result, the code falls through to the #elif defined(__linux__) branch, which then attempts to include <dlfcn.h> — a Linux-specific header that does not exist on Windows. This causes the error C1083: Cannot open include file: 'dlfcn.h': No such file or directory that we've been encountering.

I’m not sure if this is due to something wrong in my setup — please let me know if I missed any configuration steps.

bradgrantham-lunarg:
@Daveshammer

What I normally do is no compiler flags: `cmake . -Bbuild -G "Visual Studio 17 2022" -A x64`

`/DWIN32` is provided in `CXXFLAGS` and `CFLAGS` by default; specifying `CMAKE_CXX_FLAGS` and `CMAKE_C_FLAGS` as `-D` options to CMake *overrides* (replaces) the defaults which is why you're getting WIN32-related errors.  I suggest you may want instead a site-wide CMake file that *adds* the flags you want, like this:
```
# site-msvc.cmake
string(APPEND CMAKE_CXX_FLAGS_INIT " /wd4819 /wd4530")
string(APPEND CMAKE_C_FLAGS_INIT   " /wd4819")
```
(I dropped `/EHsc` because that is also in the defaults for CXXFLAGS and has no meaning in CFLAGS and I removed `/WX-` to make your environment more consistent with ours in the event you submit a C++ PR.)

You can configure cmake with this file like this:
```
cmake . -Bbuild -G "Visual Studio 17 2022" -A x64 -DCMAKE_TOOLCHAIN_FILE=site-msvc.cmake
```

Alternatively you could set environment variables and CMake will use those plus the defaults, like:
```
set CXXFLAGS=/wd4819 /wd4530 
set CFLAGS=/wd4819
cmake . -Bbuild -G "Visual Studio 17 2022" -A x64 
```

Daveshammer:
Thanks a lot! May be my environment has some problem.

PR #3025, opened 2026-06-16
fix: Return VulkanSemaphore from injected submit jobs

antonio-lunarg:
Carry the timeline value alongside the handle so injected waits use the semaphore's actual value instead of a hardcoded 1.

### DIFF
--- a/framework/decode/vulkan_submit_job.cpp
+++ b/framework/decode/vulkan_submit_job.cpp
@@ -177,9 +177,9 @@ void VulkanSubmitJobExecution::InjectBefore(VulkanSubmitJobPlan plan, std::span<
             // Execute each job function and gather injected wait-semaphores.
             for (const auto& job : jobs)
             {
-                VkSemaphore submit_semaphore = job(original_wait_semaphores);
-                GFXRECON_ASSERT(submit_semaphore != VK_NULL_HANDLE);
-                injected_wait_semaphores.push_back(submit_semaphore);
+                graphics::VulkanSemaphore submit_semaphore = job(original_wait_semaphores);
+                GFXRECON_ASSERT(submit_semaphore.semaphore != VK_NULL_HANDLE);
+                injected_wait_semaphores.push_back(submit_semaphore.semaphore);
             }
 
             // Replace the original waits with waits for the injected jobs.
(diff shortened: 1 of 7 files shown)


framework/graphics/vulkan_semaphore_util.h
```diff
@@ -34,8 +34,13 @@ GFXRECON_BEGIN_NAMESPACE(graphics)
 
 struct VulkanSemaphore
 {
-    VkSemaphore semaphore;
-    uint64_t    timeline_value; // Only used for timeline semaphores, ignored for binary semaphores
+    VkSemaphore semaphore = VK_NULL_HANDLE;
+    /// Only used for timeline semaphores, ignored for binary semaphores
+    uint64_t timeline_value = 0;
+
+    VulkanSemaphore(VkSemaphore handle = VK_NULL_HANDLE, uint64_t timeline_value = 0) :
```
fabian-lunarg: either the ctor or the in-class initializers could/should be dropped

fabian-lunarg:
changes are clean and consistent and I agree. 
only, what is 'fixed' in this PR? current behavior 'should' be identical.
both existing job producers (WarmUp, UpdateBufferAddresses) use binary semaphores.
we replace hardcoded 1 with hardcoded 0, practically ignoring both.

is there a follow-up relying on the corrected timeline-values?

framework/decode/vulkan_submit_job.cpp
```diff
@@ -246,13 +246,13 @@ void VulkanSubmitJobExecution::InjectBefore(VulkanSubmitJobPlan plan, std::span<
...
-                GFXRECON_ASSERT(submit_semaphore != VK_NULL_HANDLE);
+                graphics::VulkanSemaphore submit_semaphore = job(original_wait_semaphores);
+                GFXRECON_ASSERT(submit_semaphore.semaphore != VK_NULL_HANDLE);
 
                 VkSemaphoreSubmitInfo semaphore_info = {};
                 semaphore_info.sType                 = VK_STRUCTURE_TYPE_SEMAPHORE_SUBMIT_INFO;
-                semaphore_info.semaphore             = submit_semaphore;
-                semaphore_info.value                 = 1;
+                semaphore_info.semaphore             = submit_semaphore.semaphore;
+                semaphore_info.value                 = submit_semaphore.timeline_value;
```
antonio-lunarg, line 255: @fabian-lunarg the fix is actually here: future submit jobs will have the opportunity to return a timeline semaphore rather than a binary only. And yes, there's a draft PR https://github.com/LunarG/gfxreconstruct/pull/3016 that will need this.

fabian-lunarg: nice, that clears it up for me :)

fabian-lunarg (approved):

PR #3029, opened 2026-06-16
Make Vulkan optional for encode libraries

ncesario-lunarg:
This is a first pass at making Vulkan encode support optional by adding a GFXRECON_ENABLE_VULKAN cmake option.


ncesario-lunarg:
I'll wait for #2989 to go in first to avoid extra churn.

MarkY-LunarG:
@ncesario-lunarg , #2989 is now in.

framework/encode/CMakeLists.txt
```diff
@@ -181,14 +191,21 @@ target_include_directories(gfxrecon_encode
                                ${PROJECT_SOURCE_DIR}/framework
                                $<$<BOOL:${D3D12_SUPPORT}>:${D3D12_INCLUDE_DIRECTORIES}>)
 
-target_link_libraries(gfxrecon_encode
-                      gfxrecon_graphics
+target_link_libraries(gfxrecon_encode PUBLIC
                       gfxrecon_format
                       gfxrecon_util
                       vulkan_registry
```
bradgrantham-lunarg: Should this also break out vulkan_registry and OpenXR?

bradgrantham-lunarg (changes requested):

framework/encode/CMakeLists.txt
ncesario-lunarg: Yep! The fact it was an interface-only target for pulling in includes, so I didn't notice anything functionally changing with/without it, but it should be conditionally included.

bradgrantham-lunarg (approved):

PR #3030, opened 2026-06-17
VDR: Handle vkQueueSubmit2[KHR]

panos-lunarg:
vkQueueSubmit2[KHR] calls were not being handled when dumping resources and as a result no resources were dumped at all from these queue submissions

### DIFF
--- a/framework/decode/vulkan_replay_dump_resources_common.cpp
+++ b/framework/decode/vulkan_replay_dump_resources_common.cpp
@@ -2121,5 +2121,57 @@ void CopyBufferAndBarrier(VkCommandBuffer                    command_buffer,
                                     nullptr);
 }
 
+VkResult SubmitInfo2OnQueue(const graphics::VulkanDeviceTable& device_table,
+                            VkQueue                            queue,
+                            const VkSubmitInfo2&               submit_info_2,
+                            VkFence                            fence)
+{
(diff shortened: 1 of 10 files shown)


framework/decode/vulkan_replay_dump_resources_common.cpp
```diff
@@ -2121,5 +2121,52 @@ void CopyBufferAndBarrier(VkCommandBuffer                    command_buffer,
                                     nullptr);
 }
 
+VkResult SubmitInfo2OnQueue(const graphics::VulkanDeviceTable& device_table,
+                            VkQueue                            queue,
+                            const VkSubmitInfo2&               submit_info_2,
+                            VkFence                            fence)
+{
+    // When the device supports synchronization2 submit natively, without any conversion.
```
MarkY-LunarG: sync2?  The code below only compares vkQueueSubmit2.  Where does sync2 tie into it?

MarkY-LunarG (approved):

framework/decode/vulkan_replay_dump_resources_common.cpp
panos-lunarg: `VK_KHR_synchronization2` introduces `vkQueueSubmit2KHR` and Vulkan 1.3 introduces `vkQueueSubmit2`.
You are right, I need to check for both in order to handle both cases

PR #3034, opened 2026-06-19
Silenve MSVC warning

panos-lunarg:
MSVC emits a warning when building gfxr-consumer-library because PageStatusTracker::SetAllBlocksActiveWrite() calls std::fill with the integer literal 1, which gets deduced as int and then narrowed to the vector's element type uint8_t.


MarkY-LunarG (approved):
1u doesn't work?  Otherwise looks fine.

panos-lunarg:
> 1u doesn't work? Otherwise looks fine.

Probably, but since I cannot reproduce this I am just copying the proposed fix from the person that reported this verbatim

PR #3036, opened 2026-06-22
Fix crash in `VulkanReplayConsumerBase::OverrideCmdBindDescriptorSets2`

marius-pelegrin-arm:
The function wrongly assumes that it can be called only after a call to `vkCmdBindPipeline`

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -5457,7 +5457,8 @@ void VulkanReplayConsumerBase::OverrideCmdBindDescriptorSets2(
     VkCommandBuffer           command_buffer            = in_commandBuffer->handle;
     VkBindDescriptorSetsInfo* bind_descriptor_sets_info = pBindDescriptorSetsInfo->GetPointer();
 
-    if (bind_descriptor_sets_info != nullptr && UseAddressReplacement(device_info))
+    if (bind_descriptor_sets_info != nullptr && UseAddressReplacement(device_info) &&
+        !in_commandBuffer->bound_pipelines.empty())
     {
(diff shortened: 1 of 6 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5457,7 +5457,8 @@ void VulkanReplayConsumerBase::OverrideCmdBindDescriptorSets2(
     VkCommandBuffer           command_buffer            = in_commandBuffer->handle;
     VkBindDescriptorSetsInfo* bind_descriptor_sets_info = pBindDescriptorSetsInfo->GetPointer();
 
-    if (bind_descriptor_sets_info != nullptr && UseAddressReplacement(device_info))
+    if (bind_descriptor_sets_info != nullptr && UseAddressReplacement(device_info) &&
+        !in_commandBuffer->bound_pipelines.empty())
```
marius-pelegrin-arm: Either I misunderstood your comment or you misunderstood the commit. I am not removing the guard, I am adding it.

fabian-lunarg: all good, I was too fast and misread. lines up now :)

fabian-lunarg:
I agree with your fix. but it looks there is more we can do while we're here.

`VkPipelineBindPoint pipelineBindPoint              = in_commandBuffer->bound_pipelines.begin()->first;`

remains problematic, only picking the first bind-point, which is wrong.
we can infer pipeline-bindpoints from the stage-flags (dump-resources already doing this via a `ShaderStageFlagsToPipelineBindPoints`).

let me look into if/how we could consolidate this further.

fabian-lunarg (approved):
added another 2 commits, keeping your fix intact,
adding missing handling for multiple bindpoints

PR #3038, opened 2026-06-23
VDR: Push descriptors

panos-lunarg:
Adds handling for vkCmdPushDescriptorSet(KHR) and vkCmdPushDescriptorSet2(KHR)


csteinberg-samsung:
do you have an example of what the output of this looks like for a draw call?

panos-lunarg:
There's no new specific output. This PR adds support for the push descriptor commands in dump resources so the descriptors that are pushed/updated are tracked and dumped. Otherwise they were not considered for dumping as only cmdBindDescriptorSets was being handled

csteinberg-samsung:
so the output should look the same as a regular descriptor? There is no additional marker to indicate it was a push descriptor? That works for me.

panos-lunarg:
>so the output should look the same as a regular descriptor?

Yes.

framework/decode/vulkan_descriptor_utils.cpp
```diff
@@ -0,0 +1,158 @@
+/*
```
bradgrantham-lunarg, line 1: This all looks coupled to `decode` structures, but can any of it be moved to `framework/graphics` or `framework/util`?

bradgrantham-lunarg (approved):

PR #3039, opened 2026-06-23
Make platform-specific features more explicit

ncesario-lunarg:
Add the following cmake options for more control over which platform-specific features are built:

- GFXRECON_ENABLE_USERFAULTFD
- GFXRECON_PLATFORM_SUPPORTS_GETTID
- GFXRECON_PLATFORM_SUPPORTS_TGKILL

This change also fixes builds without zlib.

### DIFF
--- a/framework/util/platform.h
+++ b/framework/util/platform.h
@@ -391,12 +391,14 @@ inline uint64_t GetCurrentThreadId()
     uint64_t thread_id;
     pthread_threadid_np(pthread_self(), &thread_id);
     return thread_id;
-#else
+#elif defined(GFXRECON_PLATFORM_SUPPORTS_GETTID)
     return static_cast<uint64_t>(syscall(SYS_gettid));
+#else
+    return static_cast<uint64_t>(pthread_self());
 #endif
 }
 
-#if !defined(__APPLE__)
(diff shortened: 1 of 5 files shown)


MarkY-LunarG (approved):

framework/util/platform.h
```diff
@@ -391,12 +391,14 @@ inline uint64_t GetCurrentThreadId()
     uint64_t thread_id;
     pthread_threadid_np(pthread_self(), &thread_id);
     return thread_id;
-#else
+#elif defined(GFXRECON_PLATFORM_SUPPORTS_GETTID)
     return static_cast<uint64_t>(syscall(SYS_gettid));
+#else
+    return static_cast<uint64_t>(pthread_self());
```
ncesario-lunarg, line 397: One thing to note is that this path is taken for 32 bit linux on internal CI. However, I cannot reproduce that locally building 32 bit linux (neither on Fedora 43 nor Ubuntu 24.04).

ncesario-lunarg: I've looked at this pretty much every way I can, and now can't reproduce it anywhere. Merging and will back out if necessary.

PR #3041, opened 2026-06-23
Check for the present mode being supported

charles-lunarg:
Validate that the present mode being used is supported by the surface regardless of the --present-mode option. If the --present-mode option is present, log the error and continue. If it isn't, fallback to using VK_PRESENT_MODE_FIFO_KHR as it is the only guaranteed format.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -8031,9 +8031,21 @@ VkResult VulkanReplayConsumerBase::OverrideCreateSwapchainKHR(
             }
         }
 
+        auto instance_table = GetInstanceTable(physical_device_info->parent);
+
+        uint32_t present_mode_count = 0;
+        instance_table->GetPhysicalDeviceSurfacePresentModesKHR(
+            physical_device_info->handle, modified_create_info.surface, &present_mode_count, nullptr);
+
+        std::vector<VkPresentModeKHR> supported_present_modes(present_mode_count);
+        instance_table->GetPhysicalDeviceSurfacePresentModesKHR(physical_device_info->handle,
+                                                                modified_create_info.surface,
+                                                                &present_mode_count,
+                                                                supported_present_modes.data());
+
+        VkPresentModeKHR present_mode = modified_create_info.presentMode;
         if (options_.present_mode_option != util::PresentModeOption::kCapture)
         {
-            VkPresentModeKHR present_mode = modified_create_info.presentMode;
 
             switch (options_.present_mode_option)
             {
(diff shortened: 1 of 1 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -8053,30 +8065,28 @@ VkResult VulkanReplayConsumerBase::OverrideCreateSwapchainKHR(
...
+        if (is_supported)
+        {
+            modified_create_info.presentMode = present_mode;
+        }
+        else
+        {
+            if (options_.present_mode_option != util::PresentModeOption::kCapture)
             {
-                modified_create_info.presentMode = present_mode;
+                GFXRECON_LOG_ERROR("Swapchain present mode '%s' is requested but not supported",
```
bradgrantham-lunarg: Should this be FATAL?

bradgrantham-lunarg:
@fabian-lunarg you've looked at this code lately - does this make sense to you?

framework/decode/vulkan_replay_consumer_base.cpp
charles-lunarg: Thats a good point - the code wasn't fatal before so I didn't change that. But for the case when the command line switch `--present-mode` is specifying a present mode, it makes sense to yell loudly that the chosen mode does not work with the trace.

charles-lunarg: I went ahead and turned it into a FATAL error, but only if `--present-mode` isn't `capture`.

charles-lunarg:
Welp: Seems that CI considers 'no traces' to be an invalid state to be in.

mikes-lunarg:
@charles-lunarg I suspect your Android run is crashing for the same reason mine was in https://github.com/LunarG/gfxreconstruct/pull/3048, the surface handle is invalid when replaying with `--swapchain offscreen`

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: I'm also not sure about the FATAL-log. it contradicts the description: 

_If the --present-mode option is present, log the error and continue. If it isn't, fallback to using VK_PRESENT_MODE_FIFO_KHR_

makes total sense to me. but would mean we just fallback to a supported mode and use `LOG_ERROR` ?

fabian-lunarg:
> @fabian-lunarg you've looked at this code lately - does this make sense to you?

it does. I agree with the change, imo only the FATAL-log needs to be confirmed or turned into an error.

charles-lunarg:
Closing as I was able to work around this issue in the traces by adding `--present-mode fifo` to the command line arguments. Making traces work on systems other than the capture isn't a design goal, and this PR goes against that.

PR #3042, opened 2026-06-23
Alias memory for suballocated buffers

wzchu-lunarg:
Motivated by https://github.com/LunarG/Projects/issues/1171, where we capture WickedEngine (https://github.com/turanszkij/WickedEngine) test samples and replay with the flag `-m rebind`.

When replaying using `-m rebind`, all the models/geometries disappear.

This engine does the following:
-Allocate buffer that is binded to the shader, and the shader reads
-Suballocate buffers, and read into them from staging buffers using `cmdCopyBuffers`

This is due to GFXR's handling of reusing overlapping buffers. Although we try to resolve these with `FindVmaMemoryInfo`, and `vulkan_rebind_allocator.h`'s `is_compatible`, since these suballocated buffers do not have a corresponding `vkGetBufferMemoryRequirements` to record captured requirement sizes and defaults to 0, GFXR thinks the already allocated buffer is not compatible and creates a separate allocation, resulting in the shader seeing none of the data that was supposed to be copied.

Change 0 check of capture size to not return early by adding flag to progress specifically for buffers.

Also set `prefers_dedicated_allocation` to `false` when allocating buffers, as they could be suballocated.

Needs to be addressed that this only fixes the case where suballocations are binded after the initial allocation. If suballocations are binded before, the same missing geometries issue would still occur as the buffer binded to the shader will be a separate allocation.

### DIFF
--- a/framework/decode/vulkan_rebind_allocator.cpp
+++ b/framework/decode/vulkan_rebind_allocator.cpp
@@ -640,6 +640,9 @@ VulkanRebindAllocator::AllocateMemoryForBuffer(VkBuffer                         
     allocator_->GetBufferMemoryRequirements(
         buffer, replay_req, requires_dedicated_allocation, prefers_dedicated_allocation);
 
+    // We don't know if this buffer will be sub-allocated, so we don't want a dedicated allocation
+    prefers_dedicated_allocation = false;
+
     replay_req.alignment = std::max<VkDeviceSize>(replay_req.alignment, min_buffer_alignment_);
 
     VmaAllocationCreateInfo create_info{};
@@ -661,6 +664,7 @@ VulkanRebindAllocator::AllocateMemoryForBuffer(VkBuffer                         
                           requires_dedicated_allocation,
                           prefers_dedicated_allocation,
                           create_info,
+                          true,
                           vma_mem_info))
     {
         return VK_SUCCESS;
@@ -956,6 +960,7 @@ VkResult VulkanRebindAllocator::AllocateMemoryForImage(VkImage                  
                           requires_dedicated_allocation,
                           prefers_dedicated_allocation,
                           create_info,
+                          false,
                           vma_mem_info))
     {
         return VK_SUCCESS;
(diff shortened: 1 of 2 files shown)


framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -640,6 +640,9 @@ VulkanRebindAllocator::AllocateMemoryForBuffer(VkBuffer
     allocator_->GetBufferMemoryRequirements(
         buffer, replay_req, requires_dedicated_allocation, prefers_dedicated_allocation);
 
+    // We don't know if this buffer will be sub-allocated, so we don't want a dedicated allocation
+    prefers_dedicated_allocation = false;
```
fabian-lunarg, line 644: this is unconditionally set for all buffers, which is rather crude.
so all buffer loose their original `prefers_dedicated_allocation` setting.
it affects both performance but also correctness on certain drivers 
(e.g. large RT-buffers might require dedicated alloc to function properly)

framework/decode/vulkan_rebind_allocator.h
```diff
@@ -454,19 +454,27 @@ class VulkanRebindAllocator : public VulkanResourceAllocator
...
             {
                 return false;
             }
 
             // memory offset and size is in the range. mem_req and create_info are the same.
             if ((offset >= offset_from_original_device_memory) &&
-                ((offset + capture_req.size) <= (offset_from_original_device_memory + capture_mem_req.size)) &&
+                // capture-space containment: additional tightening if we have capture sizes.
+                (!have_capture_sizes ||
+                 ((offset + capture_req.size) <= (offset_from_original_device_memory + capture_mem_req.size))) &&
```
fabian-lunarg, line 477: I think this is problematic.

`(offset + capture_req.size) <= (offset_from_original_device_memory + capture_mem_req.size)` was the only guarantee the resource fell within this sub-block's **captured** range.

the replay's memory-range 'might' be larger (different alignment/ HW-reqs during replay).
-> the check 'could' succeed and bind a sub-allocation where it didn't belong during capture,
yielding wrong aliasing by sharing the wrong sub-block (correct VkDeviceMemory, wrong offset/region within it).

fabian-lunarg:
the diagnosis of the issue is spot on and the repro-case is very valuable.

while this PR allows successful replay of the sub-allocation aliasing-pattern exhibited by wicked-engine,
it would introduce other potential aliasing issues with `-m rebind`, which would be even harder to detect.
we discussed offline and will present a slightly different approach, avoiding the two caveats I pointed out.

PR #3044, opened 2026-06-24
VDR: Handle timeline semaphores in QueueSubmit2

panos-lunarg:
Initial implementation of vkQueueSubmit2 ignored timeline values via
VkTimelineSemaphoreSubmitInfo when converting VkQueueSubmitInfo into
VkQueueSuubmitInfo2. This lead to hangs when timeline semaphores were
used.


framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -33,6 +33,7 @@
 #include "format/format_util.h"
 #include "generated/generated_vulkan_enum_to_string.h"
 #include "generated/generated_vulkan_struct_decoders.h"
+#include "generated/generated_vulkan_struct_deep_copy.cpp"
```
fabian-lunarg: you should be able to just include the header (it's only templated and not generated)
-> `#include "graphics/vulkan_struct_deep_copy.h"` (otherwise it recompiles the deep_copy routines into decode)

fabian-lunarg (approved):
good to go, only the `#include` statement needs correction

framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -1856,46 +1857,75 @@ VkResult VulkanReplayDumpResourcesBase::QueueSubmit(std::span<const VkSubmitInfo
...
     std::vector<std::vector<VkSemaphoreSubmitInfo>>     signal_semaphores(submit_infos.size());
+    std::vector<std::vector<uint8_t>>                   submit_infos_deep_copy_raw(submit_infos.size());
 
     for (size_t i = 0; i < submit_infos.size(); ++i)
     {
-        const VkSubmitInfo& submit_info = submit_infos[i];
+        // If VkTimelineSemaphoreSubmitInfo is detected in the submit info's pNext chain then we translate the struct's
+        // information into the VkSubmitInfo2's structure and we remove it from the chain. Because we interfere with the
+        // pNext chain we create a deep copy
+        const size_t num_bytes = graphics::vulkan_struct_deep_copy(&submit_infos[i], 1, nullptr);
```
antonio-lunarg: This is done even for submits with no VkTimelineSemaphoreSubmitInfo.
I would check first with `vulkan_struct_get_pnext<VkTimelineSemaphoreSubmitInfo>` and only deep-copy when present.

framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -1856,46 +1857,75 @@ VkResult VulkanReplayDumpResourcesBase::QueueSubmit(std::span<const VkSubmitInfo
...
+                timeline_semaphore_wait_values[s] = timeline_semaphore_submit_info->pWaitSemaphoreValues[s];
+            }
+
+            for (uint32_t s = 0; s < std::min(modified_submit_info->signalSemaphoreCount,
+                                              timeline_semaphore_submit_info->signalSemaphoreValueCount);
+                 ++s)
+            {
+                timeline_semaphore_signal_values[s] = timeline_semaphore_submit_info->pSignalSemaphoreValues[s];
+            }
+            graphics::vulkan_struct_remove_pnext<VkTimelineSemaphoreSubmitInfo>(modified_submit_info);
```
antonio-lunarg: I think it would be nice to move this VkSubmitInfo->VkSubmitInfo2 adapter out somewhere else, as it seems to me "generic" enough to not be closely tied to the concept of dumping resources.

On that note, I'd like to point out that while this is correctly stripping `VkTimelineSemaphoreSubmitInfo`, there are still 2 structure that might not be valid for `VkSubmitInfo2`:
- `VkProtectedSubmitInfo` (this should be easy -> `VkSubmitinfo2::flags & VK_SUBMIT_PROTECTED_BIT`)
- `VkDeviceGroupSubmitInfo`, this is a bit more convoluted, if you do not want to handle it right now in this PR, I would suggest to add some sort of warning at least to mention it's not supported yet.

framework/decode/vulkan_replay_dump_resources.cpp
panos-lunarg: done

framework/decode/vulkan_replay_dump_resources.cpp
panos-lunarg: done

framework/decode/vulkan_replay_dump_resources.cpp
panos-lunarg: - Adapter/converted moved in separate .h/.cpp files
- Added handling for `VkProtectedSubmitInfo` and `VkDeviceGroupSubmitInfo` too

framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -33,6 +33,8 @@
 #include "format/format_util.h"
 #include "generated/generated_vulkan_enum_to_string.h"
 #include "generated/generated_vulkan_struct_decoders.h"
+#include "graphics/vulkan_struct_deep_copy.h"
```
antonio-lunarg: Not needed anymore here.

framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -1852,60 +1854,10 @@ VkResult VulkanReplayDumpResourcesBase::QueueSubmit(std::span<const VkSubmitInfo
...
-                                           static_cast<uint32_t>(wait_semaphores[i].size()),
-                                           wait_semaphores[i].size() ? wait_semaphores[i].data() : nullptr,
-                                           static_cast<uint32_t>(command_buffers[i].size()),
-                                           command_buffers[i].size() ? command_buffers[i].data() : nullptr,
-                                           static_cast<uint32_t>(signal_semaphores[i].size()),
-                                           signal_semaphores[i].size() ? signal_semaphores[i].data() : nullptr };
-    }
-
-    return QueueSubmit2(submit_infos_2, device_table, queue, fence, index);
+    return QueueSubmit2(translated.GetSubmitInfos2(), device_table, queue, fence, index);
```
antonio-lunarg, line 1975: I think you want to wrap the code here:
```
    if (device_table.QueueSubmit2 != graphics::noop::vkQueueSubmit2)
    {
        ...
    }
    else if (device_table.QueueSubmit2KHR != graphics::noop::vkQueueSubmit2KHR)
    {
        ...
    }
```

Otherwise it will fall back in `SubmitInfo2OnQueue` with timeline semaphores stripped from the submit infos.

panos-lunarg: Good catch!
I am not sure I understand your suggestion though. What about make `SubmitInfo2OnQueue` undo `SubmitInfo2Translator`'s work and reintroduce the removed structs from the pNext chain?

antonio-lunarg: Yeah, why not, you can improve the translator and implement a v2->v1.
I would make sure to add some unit tests to guarantee that once this series of transformations is done `v1->v2->v1'` then `v1==v1'`.

panos-lunarg: done

antonio-lunarg (approved):

PR #3047, opened 2026-06-24
[Frame looping] [Prototype] [Not to Merge] Make it compatible with more games

olehkuznetsov:
Refactored fences
Modified some resources destruction
Ignore vkCmdWriteTimestamp
Update and optimize buffer state restoration (lazy buffer tracking; fixes unbound buffers crash caused by copying state of buffers created without bound memory)


framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -512,6 +413,21 @@ void VulkanReplayFrameLoopConsumer::Process_vkQueuePresentKHR(
     }
 }
 
+void VulkanReplayFrameLoopConsumer::Process_vkCmdWriteTimestamp(const ApiCallInfo&      call_info,
```
antonio-lunarg: Need to do the same for `Process_vkCmdWriteTimestamp2` and `Process_vkCmdWriteTimestamp2KHR`?

framework/decode/vulkan_decoder_base.h
```diff
@@ -57,6 +57,18 @@ class VulkanDecoderBase : public ApiDecoder
 
     virtual void WaitIdle() override;
 
+    virtual void OnLoopStart() override
+    {
+        for (auto consumer : consumers_)
+        {
+            auto listener = dynamic_cast<LoopStartListener*>(consumer);
```
antonio-lunarg, line 64: I think we can drop `LoopStartListener` by adding a virtual `OnLoopStart()` to `VulkanConsumer`.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -103,7 +100,7 @@ void VulkanReplayFrameLoopConsumer::Process_vkDestroyDescriptorPool(
     format::HandleId                                     descriptorPool,
     StructPointerDecoder<Decoded_VkAllocationCallbacks>* pAllocator)
 {
-    if (frame_loop_info_.IsRepetition() && !frame_loop_info_.IsFinalIteration())
+    if (frame_loop_info_.IsRepetition())
```
antonio-lunarg: If this crashes, there's a bug somewhere, but I believe we still want to honor the destroy call on the last iteration of the loop.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -256,166 +253,110 @@ void VulkanReplayFrameLoopConsumer::Process_vkFreeDescriptorSets(const ApiCallIn
...
+            auto device_info = table.GetVkDeviceInfo(fence_info->parent_id);
+            if (device_info != nullptr && device_info->handle != VK_NULL_HANDLE)
             {
-                t.waited_upon_fences_[fence] = 1;
+                VkDevice                           device       = device_info->handle;
+                const graphics::VulkanDeviceTable* device_table = GetDeviceTable(device);
+                if (device_table != nullptr)
+                {
+                    VkResult status = device_table->GetFenceStatus(device, fence_info->handle);
+                    initial_fence_states_[fence_info->capture_id] = status;
```
antonio-lunarg, line 748: Nice. Is there any other kind of resource that might benefit from a similar approach?

BTW I think this does not take into account resources created _within_ the loop range, which is fine as I believe that's a separate problem. Still worth adding a comment to the function docs about that.

framework/decode/preload_file_processor.cpp
```diff
@@ -37,8 +38,23 @@ void PreloadFileProcessor::PreloadLoopFrame()
...
+    }
+    else
+    {
+        loop_bookmark_ = preload_batch_iterator_.MakeBookmark(preload_block_iterator_);
+
+        // For N > 1, the state is already booted up, so we capture immediately!
+        WaitDecodersIdle();
+        for (auto decoder : decoders_)
+        {
+            decoder->OnLoopStart();
```
antonio-lunarg, line 55: I understand that in case of a trimmed capture, an end-marker block is emitted which will trigger `OnLoopStart` on the consumer, but at this point we could move this loop down out of the else, and remove the `ProcessStateEndMarker` override entirely.

antonio-lunarg: Also, looping over the first frame of an untrimmed capture would not fire `OnLoopStart` cause that kind of capture does not have a `ProcesStateEndMarker` block, which is a point in favour to moving the `OnLoopStart` out of the conditional branches.

antonio-lunarg:
Need to sync with this PR https://github.com/LunarG/gfxreconstruct/pull/3052

nickdriscoll-lunarg:
I think that maybe if `vkCreateFence` was overridden to reset the fence on subsequent iterations of the loop frame, then this PR can supersede #3052. Otherwise the case of "fence was created and used during the loop range" is the one thing my PR handles that this one doesn't.

bradgrantham-lunarg:
@olehkuznetsov , when possible it's helpful to us for review purposes to make multiple smaller PRs that each contain well-encapsulated changes to the code.  It this PR a series of changes that are accumulating as you test more and more titles?  If so, please take a moment to break it into smaller PRs or we can discuss outside GitHub whether LunarG should break the PR apart for you.

Also, if you believe you've resolved the feedback given in review, please either note that you've applied the change or respond why you think there's an issue with the feedback.

Thank you!

bradgrantham-lunarg:
@olehkuznetsov Making a change to .clang-format is project wide shouldn't be made lightly; I note your commit contains no details on the change.  Can you explain in more detail the benefit of the changes you're proposing in the .clang-format file?

olehkuznetsov:
> @olehkuznetsov Making a change to .clang-format is project wide shouldn't be made lightly; I note your commit contains no details on the change. Can you explain in more detail the benefit of the changes you're proposing in the .clang-format file?

Seems as clang-format-14 is not available for me, so, goal of those changes are to make newer versions of clang-format to be compatible with clang-format-14

olehkuznetsov:
> @olehkuznetsov , when possible it's helpful to us for review purposes to make multiple smaller PRs that each contain well-encapsulated changes to the code. It this PR a series of changes that are accumulating as you test more and more titles? If so, please take a moment to break it into smaller PRs or we can discuss outside GitHub whether LunarG should break the PR apart for you.
> 
> Also, if you believe you've resolved the feedback given in review, please either note that you've applied the change or respond why you think there's an issue with the feedback.
> 
> Thank you!

The idea here is to share prototype code which works with more cases, so, Antonio and Nick can get insights and apply ideas from this prototype. 

This one should not be submitted - I've updated title to make it clear.

PR #3048, opened 2026-06-24
Always query supported replay surface formats

mikes-lunarg:
In cases where the surface formats are either not queried by the capture file or queried in an incompatible way (such as VK_GOOGLE_surfaceless_query), do the query at swapchain creation time so the B8G8R8A8 fallback path still has a chance to run.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -7997,6 +7997,35 @@ VkResult VulkanReplayConsumerBase::OverrideCreateSwapchainKHR(
                 supported_extension_iterator == instance_info->util_info.enabled_extensions.end();
         }
 
+        // If supported surface formats were not queried before, query them now
+        if (!physical_device_info->surface_formats && options_.swapchain_option != util::SwapchainOption::kOffscreen)
+        {
+            const auto instance_table = GetInstanceTable(physical_device_info->handle);
+            util::BeginInjectedCommands();
+            uint32_t surface_format_count = 0;
+            auto     result               = instance_table->GetPhysicalDeviceSurfaceFormatsKHR(
+                physical_device_info->handle, modified_create_info.surface, &surface_format_count, nullptr);
+            if (result == VK_SUCCESS && surface_format_count > 0)
+            {
+                physical_device_info->surface_formats =
+                    std::make_optional<std::vector<VkSurfaceFormatKHR>>(surface_format_count);
+                result =
+                    instance_table->GetPhysicalDeviceSurfaceFormatsKHR(physical_device_info->handle,
+                                                                       modified_create_info.surface,
+                                                                       &surface_format_count,
+                                                                       physical_device_info->surface_formats->data());
+                if (result == VK_SUCCESS || result == VK_INCOMPLETE)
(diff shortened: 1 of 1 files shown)


mikes-lunarg:
CC @richard-lunarg

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7946,46 +7946,71 @@ VkResult VulkanReplayConsumerBase::OverrideCreateSwapchainKHR(
...
+                        const auto* view_formats     = format_list_pnext->pViewFormats;
+                        const auto* view_formats_end = view_formats + format_list_pnext->viewFormatCount;
+                        bool        mutable_srgb     = std::any_of(view_formats, view_formats_end, vkuFormatIsSRGB);
+
+                        // overwrite existing VkImageFormatListCreateInfo
+                        format_list_create_info                  = *format_list_pnext;
+                        format_list_create_info->viewFormatCount = mutable_srgb ? 2 : 1;
+                        format_list_create_info->pViewFormats    = fallback_color_formats;
+                        graphics::vulkan_struct_add_pnext<VkImageFormatListCreateInfo>(
+                            &modified_create_info, &format_list_create_info.value());
```
MarkY-LunarG: How does this overwrite the existing item?   It  seems to only add it to the start of the list.

mikes-lunarg: https://github.com/LunarG/gfxreconstruct/blob/10fd4d371e755f50383d9b168461ae09d5c51678/framework/graphics/vulkan_struct_get_pnext.h#L157-L161

mikes-lunarg: vulkan_struct_add_pnext is very sneaky and removes any duplicates first

MarkY-LunarG (approved):

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7946,46 +7946,71 @@ VkResult VulkanReplayConsumerBase::OverrideCreateSwapchainKHR(
...
             {
-                if (auto* format_list_pnext =
-                        graphics::vulkan_struct_get_pnext<VkImageFormatListCreateInfo>(&modified_create_info))
+                // If surface formats are not queried before, query them now to check if the format is supported.
+                const auto instance_table = GetInstanceTable(physical_device_info->handle);
+                util::BeginInjectedCommands();
+                uint32_t surface_format_count = 0;
+                instance_table->GetPhysicalDeviceSurfaceFormatsKHR(
+                    physical_device_info->handle, modified_create_info.surface, &surface_format_count, nullptr);
+                std::vector<VkSurfaceFormatKHR> surface_formats(surface_format_count);
```
fabian-lunarg: not 100% sure, but after the required rebase (onto #3051) this local vector 'could/should' be dropped in favor of initializing `physical_device_info->surface_formats` here.

mikes-lunarg: Good idea, using `physical_device_info->surface_formats` makes it a bit simpler. Now I just query before the existing code don't have to change anything else

PR #3051, opened 2026-06-25
swapchain: pick supported fallback format instead of hardcoded BGRA8

fabian-lunarg:
- OverrideCreateSwapchainKHR: when captured imageFormat is unsupported, select fallback from supported formats instead of just VK_FORMAT_B8G8R8A8_UNORM
- OverrideCreateImageView: match the swapchain image's actual format for non-BGRA8 fallbacks to keep swapchain-image views format-compatible
- VulkanVirtualSwapchain::PresentImageAdHoc: same supported-format selection
- factor the selection into decode::SelectFallbackSurfaceFormat (vulkan_swapchain_format.h), shared by both swapchain paths
- log adjusted formats via util::ToString<VkFormat> instead of raw integers


mikes-lunarg (approved):
I think I would prefer if this goes in first and then I rebase #3048 on top of it

PR #3052, opened 2026-06-25
Frame loop fence tracking fix

nickdriscoll-lunarg:
The previous approach to handling fences relied on the assumption that the app will always call `vkWaitForFences` on a fence before signaling it again. However, some apps choose to check `vkGetFenceStatus` first and will skip waiting on the fence if it is already signaled.

The new approach is a more comprehensive, heavy-handed way of tracking, where the first time a fence is observed in the first iteration of the loop range, its state is recorded, so that at the end of each looped frame the fences can be forced back into their original state.

### DIFF
--- a/framework/decode/api_decoder.h
+++ b/framework/decode/api_decoder.h
@@ -57,7 +57,7 @@ class ApiDecoder
 
     virtual void WaitIdle() = 0;
 
-    virtual void OnLoopStart() {}
+    virtual void StartLooping() {}
 
     virtual bool IsComplete(uint64_t block_index) = 0;
 
--- a/framework/decode/api_decoder.h
+++ b/framework/decode/api_decoder.h
@@ -57,6 +57,8 @@ class ApiDecoder
 
     virtual void WaitIdle() = 0;
 
+    virtual void OnLoopStart() {}
+
     virtual bool IsComplete(uint64_t block_index) = 0;
 
     virtual bool SupportsApiCall(format::ApiCallId id) = 0;
--- a/framework/decode/preload_file_processor.cpp
+++ b/framework/decode/preload_file_processor.cpp
@@ -39,6 +39,11 @@ void PreloadFileProcessor::PreloadLoopFrame()
     loop_replay_ = true;
     loop_bookmark_ =
         SkipStateBlocks(dispatch_frame_number_, preload_batch_iterator_.MakeBookmark(preload_block_iterator_));
+
+    for (auto decoder : decoders_)
+    {
+        decoder->OnLoopStart();
+    }
 }
 void PreloadFileProcessor::PreloadNextFrames(size_t count)
 {
--- a/framework/decode/preload_file_processor.cpp
+++ b/framework/decode/preload_file_processor.cpp
@@ -42,7 +42,7 @@ void PreloadFileProcessor::PreloadLoopFrame()
 
     for (auto decoder : decoders_)
     {
-        decoder->OnLoopStart();
+        decoder->StartLooping();
     }
 }
 void PreloadFileProcessor::PreloadNextFrames(size_t count)
--- a/framework/decode/vulkan_replay_frame_loop_consumer.cpp
+++ b/framework/decode/vulkan_replay_frame_loop_consumer.cpp
@@ -43,6 +43,27 @@ GFXRECON_BEGIN_NAMESPACE(decode)
         }                                                                     \
     }
 
+void VulkanReplayFrameLoopConsumer::ProcessStateEndMarker(uint64_t frame_number)
+{
+    VulkanReplayConsumer::ProcessStateEndMarker(frame_number);
+
+    // If trim state had to be loaded, call OnLoopStart() again
+    per_device_fence_tracking_.clear();
+    OnLoopStart();
+}
+
+void VulkanReplayFrameLoopConsumer::OnLoopStart()
+{
+    WaitDevicesIdle();
+    GFXRECON_LOG_DEBUG("VulkanReplayFrameLoopConsumer::OnLoopStart()");
+    CommonObjectInfoTable& table = GetObjectInfoTable();
+    table.VisitVkFenceInfo([this](const VulkanFenceInfo* fence_info) {
+        GFXRECON_LOG_DEBUG("Tracking fence state for fence %" PRIu64, fence_info->capture_id);
+        format::HandleId device_id = fence_info->parent_id;
+        this->TrackFenceState(device_id, fence_info->capture_id);
+    });
+}
+
 void VulkanReplayFrameLoopConsumer::Process_vkCreateCommandPool(
     const ApiCallInfo&                                     call_info,
     VkResult                                               returnValue,
@@ -256,60 +277,37 @@ void VulkanReplayFrameLoopConsumer::Process_vkFreeDescriptorSets(const ApiCallIn
         call_info, returnValue, device, descriptorPool, descriptorSetCount, pDescriptorSets);
 }
 
-void VulkanReplayFrameLoopConsumer::Process_vkWaitForFences(const ApiCallInfo&             call_info,
-                                                            VkResult                       returnValue,
-                                                            format::HandleId               device,
-                                                            uint32_t                       fenceCount,
-                                                            HandlePointerDecoder<VkFence>* pFences,
-                                                            VkBool32                       waitAll,
-                                                            uint64_t                       timeout)
+void VulkanReplayFrameLoopConsumer::Process_vkCreateFence(
+    const ApiCallInfo&                                   call_info,
+    VkResult                                             returnValue,
+    format::HandleId                                     device,
+    StructPointerDecoder<Decoded_VkFenceCreateInfo>*     pCreateInfo,
+    StructPointerDecoder<Decoded_VkAllocationCallbacks>* pAllocator,
+    HandlePointerDecoder<VkFence>*                       pFence)
 {
+    if (frame_loop_info_.IsRepetition())
+    {
+        // Reset the fence on loop repetitions instead of creating it
+        VkDevice device_handle = GetObjectInfoTable().GetVkDeviceInfo(device)->handle;
+        GFXRECON_ASSERT(device_handle != 0);
+        const graphics::VulkanDeviceTable* device_table = GetDeviceTable(device_handle);
+        VkResult res = device_table->ResetFences(device_handle, 1, pFence->GetHandlePointer());
+        CHECK_VK_RESULT(res, "vkResetFences");
+        return;
+    }
+
+    VulkanReplayConsumer::Process_vkCreateFence(call_info, returnValue, device, pCreateInfo, pAllocator, pFence);
+
     if (frame_loop_info_.IsLooping() && !frame_loop_info_.IsRepetition())
     {
-        for (format::HandleId fence : pFences->GetSpan())
+        // Record this fence as having an initial state of unsignaled
+        if (!per_device_fence_tracking_.contains(device))
         {
-            TrackFenceState(device, fence);
+            per_device_fence_tracking_[device] = {};
         }
+        FenceTracking& t                               = per_device_fence_tracking_[device];
+        t.initial_fence_states_[*pFence->GetPointer()] = false;
     }
-
-    VulkanReplayConsumer::Process_vkWaitForFences(
-        call_info, returnValue, device, fenceCount, pFences, waitAll, timeout);
-}
-
-void VulkanReplayFrameLoopConsumer::Process_vkResetFences(const ApiCallInfo&             call_info,
-                                                          VkResult                       returnValue,
-                                                          format::HandleId               device,
-                                                          uint32_t                       fenceCount,
-                                                          HandlePointerDecoder<VkFence>* pFences)
-{
-    if (frame_loop_info_.IsLooping() && !frame_loop_info_.IsRepetition())
-    {
-        for (format::HandleId fence : pFences->GetSpan())
-        {
-            TrackFenceState(device, fence);
-        }
-    }
-
-    VulkanReplayConsumer::Process_vkResetFences(call_info, returnValue, device, fenceCount, pFences);
-}
-
-void VulkanReplayFrameLoopConsumer::Process_vkQueueSubmit(const ApiCallInfo&                          call_info,
-                                                          VkResult                                    returnValue,
-                                                          format::HandleId                            queue,
-                                                          uint32_t                                    submitCount,
-                                                          StructPointerDecoder<Decoded_VkSubmitInfo>* pSubmits,
-                                                          format::HandleId                            fence)
-{
-    if (frame_loop_info_.IsLooping() && !frame_loop_info_.IsRepetition())
-    {
-        CommonObjectInfoTable& table      = GetObjectInfoTable();
-        VulkanQueueInfo*       queue_info = table.GetVkQueueInfo(queue);
-        format::HandleId       device     = queue_info->parent_id;
-        GFXRECON_ASSERT(device);
-        TrackFenceState(device, fence);
-    }
-
-    VulkanReplayConsumer::Process_vkQueueSubmit(call_info, returnValue, queue, submitCount, pSubmits, fence);
 }
 
 void VulkanReplayFrameLoopConsumer::TrackFenceState(format::HandleId device, format::HandleId fence)
@@ -325,14 +323,11 @@ void VulkanReplayFrameLoopConsumer::TrackFenceState(format::HandleId device, for
         VulkanDeviceInfo* device_info = GetObjectInfoTable().GetVkDeviceInfo(device);
         GFXRECON_ASSERT(device_info != nullptr);
         VulkanFenceInfo* fence_info = GetObjectInfoTable().GetVkFenceInfo(fence);
-        // fence_info could be nullptr
-        if (fence_info == nullptr)
-        {
-            return;
-        }
+        GFXRECON_ASSERT(fence_info != nullptr);
         const graphics::VulkanDeviceTable* device_table = GetDeviceTable(device_info->handle);
         GFXRECON_ASSERT(device_table != nullptr);
-        VkResult res                   = device_table->GetFenceStatus(device_info->handle, fence_info->handle);
+        VkResult res = device_table->GetFenceStatus(device_info->handle, fence_info->handle);
+        GFXRECON_LOG_DEBUG("Fence %" PRIu64 " signaled == %s", res == VK_SUCCESS ? "true" : "false");
         t.initial_fence_states_[fence] = res == VK_SUCCESS;
     }
 }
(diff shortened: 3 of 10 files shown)


framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -345,78 +351,36 @@ void VulkanReplayFrameLoopConsumer::FixupDeviceFences(format::HandleId device, f
...
-
-        GFXRECON_LOG_DEBUG("Synthetically signaling fences...");
-        for (VkFence fence : manual_signal_fences)
-        {
-            VulkanQueueInfo* queue_info = table.GetVkQueueInfo(queue);
-            result                      = device_table->QueueSubmit(queue_info->handle, 0, nullptr, fence);
-            CHECK_VK_RESULT(result, "vkDeviceWaitIdle");
-        }
+        VulkanQueueInfo* queue_info = table.GetVkQueueInfo(queue);
+        result                      = device_table->QueueSubmit(queue_info->handle, 0, nullptr, fence);
```
davidd-lunarg: Does there need to be a wait of some kind to ensure all these `vkQueueSubmit` calls finish before the next frame?

nickdriscoll-lunarg: There's a wait in `Process_vkQueuePresentKHR` before `FixupDeviceFences` is called.

davidd-lunarg (approved):

nickdriscoll-lunarg:
@antonio-lunarg This is a bigger change now, as I had to add a way to run some code right after trim state loading but before any functions in the loop range are called.

Fences are no longer tracked per entry point, instead all fence objects are queried for their state right before frame looping begins.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -256,79 +277,59 @@ void VulkanReplayFrameLoopConsumer::Process_vkFreeDescriptorSets(const ApiCallIn
...
+    StructPointerDecoder<Decoded_VkAllocationCallbacks>* pAllocator,
+    HandlePointerDecoder<VkFence>*                       pFence)
 {
+    if (frame_loop_info_.IsRepetition())
+    {
+        // Reset the fence on loop repetitions instead of creating it
+        VkDevice device_handle = GetObjectInfoTable().GetVkDeviceInfo(device)->handle;
+        GFXRECON_ASSERT(device_handle != 0);
+        const graphics::VulkanDeviceTable* device_table = GetDeviceTable(device_handle);
+        VkResult res = device_table->ResetFences(device_handle, 1, pFence->GetHandlePointer());
```
antonio-lunarg: I am not sure this is the right value:
- `pFence->GetHandlePointer()` is nullptr at this point.
- But the fence has already been created and you can find its handle with this:
```
GetObjectInfoTable().GetVkFenceInfo(*pFence->GetPointer())->handle
```

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -256,79 +277,59 @@ void VulkanReplayFrameLoopConsumer::Process_vkFreeDescriptorSets(const ApiCallIn
...
-            t.signaled_fences_[fence];
-            GFXRECON_LOG_DEBUG("VkFence with handle \"%" PRIu64 "\" has been signaled %" PRIu32 " times.",
-                               fence,
-                               t.signaled_fences_[fence]);
-        }
+        GFXRECON_ASSERT(fence_info != nullptr);
+        const graphics::VulkanDeviceTable* device_table = GetDeviceTable(device_info->handle);
+        GFXRECON_ASSERT(device_table != nullptr);
+        VkResult res = device_table->GetFenceStatus(device_info->handle, fence_info->handle);
+        GFXRECON_LOG_DEBUG("Fence %" PRIu64 " signaled == %s", res == VK_SUCCESS ? "true" : "false");
```
antonio-lunarg: Missing fence id as a second argument?

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -256,79 +277,59 @@ void VulkanReplayFrameLoopConsumer::Process_vkFreeDescriptorSets(const ApiCallIn
...
+        // Reset the fence on loop repetitions instead of creating it
+        VkDevice device_handle = GetObjectInfoTable().GetVkDeviceInfo(device)->handle;
+        GFXRECON_ASSERT(device_handle != 0);
+        const graphics::VulkanDeviceTable* device_table = GetDeviceTable(device_handle);
+        VkResult res = device_table->ResetFences(device_handle, 1, pFence->GetHandlePointer());
+        CHECK_VK_RESULT(res, "vkResetFences");
+        return;
+    }
+
+    VulkanReplayConsumer::Process_vkCreateFence(call_info, returnValue, device, pCreateInfo, pAllocator, pFence);
```
antonio-lunarg: Need to consider how this interacts with the `allocatedLoopResources` mechanism.
`vkDestroyFence` still uses that kind of bookeeping. I am not sure which is the best approach, maybe calling `VulkanReplayFrameLoopConsumerBase::Process_vkCreateFence()`?

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -256,79 +277,59 @@ void VulkanReplayFrameLoopConsumer::Process_vkFreeDescriptorSets(const ApiCallIn
...
-            {
-                t.waited_upon_fences_[fence] = 1;
-            }
-            GFXRECON_LOG_DEBUG("VkFence with handle \"%" PRIu64 "\" has been waited on %" PRIu32 " times.",
-                               fence,
-                               t.waited_upon_fences_[fence]);
+            per_device_fence_tracking_[device] = {};
         }
+        FenceTracking& t                               = per_device_fence_tracking_[device];
+        t.initial_fence_states_[*pFence->GetPointer()] = false;
```
antonio-lunarg: `false` means that all fences being created are expected to be unsignaled at creation time, but there's a create flag that we need to take into account. I would use:
```
pCreateInfo->GetPointer()->flags & VK_FENCE_CREATE_SIGNALED_BIT
```

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -345,78 +346,36 @@ void VulkanReplayFrameLoopConsumer::FixupDeviceFences(format::HandleId device, f
...
-    std::vector<VkFence> manual_wait_fences;
-    manual_wait_fences.reserve(t.signaled_fences_.size());
-    for (auto [fence_id, signal_count] : t.signaled_fences_)
+    // Reset all fences, then synthetically signal the fences that were signaled
+    // at the start of the loop range
+    std::vector<VkFence> all_fences;
+    std::vector<VkFence> fences_to_signal;
+    all_fences.reserve(t.initial_fence_states_.size());
+    fences_to_signal.reserve(t.initial_fence_states_.size());
+    for (auto [fence_id, was_initially_signaled] : t.initial_fence_states_)
```
antonio-lunarg, line 365: Entries in this map are never erased. I think we should make sure to remove them overriding `vkDestroyFence`, otherwise `fence_info->` down below is going to segfault.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -43,6 +43,27 @@ GFXRECON_BEGIN_NAMESPACE(decode)
         }                                                                     \
     }
 
+void VulkanReplayFrameLoopConsumer::ProcessStateEndMarker(uint64_t frame_number)
+{
+    VulkanReplayConsumer::ProcessStateEndMarker(frame_number);
+
+    // If trim state had to be loaded, call OnLoopStart() again
+    per_device_fence_tracking_.clear();
+    OnLoopStart();
```
antonio-lunarg: I am not sure this is the right place for calling `OnLoopStart`.
At this point a trimming state is loaded, but that does not mean that a loop is starting. Maybe I want to loop over the _second_ frame of a trimmed capture.
The `OnLoopStart` call in `PreloadLoopFrame` is the right place and should be enough IMHO.

antonio-lunarg: Ah, now I see what's going on.
In case of looping over the first frame of a trimmed capture, we DO want to call `TrackFenceState` from here, but only when `IsLooping() == true`.

framework/decode/preload_file_processor.cpp
```diff
@@ -39,6 +39,11 @@ void PreloadFileProcessor::PreloadLoopFrame()
     loop_replay_ = true;
     loop_bookmark_ =
         SkipStateBlocks(dispatch_frame_number_, preload_batch_iterator_.MakeBookmark(preload_block_iterator_));
+
+    for (auto decoder : decoders_)
+    {
+        decoder->OnLoopStart();
```
antonio-lunarg: Ok then, this needs to be adjusted in case we have a preloaded frame with state blocks.
First iteration, state blocks are processed, then `ProcessStateEndMarker` would trigger `OnLoopStart()`.

In case of a non-trimmed capture, there are no state blocks, so `OnLoopStart()` should be called here.

Practically speaking we want something like:
```cpp
auto start_bookmark = preload_batch_iterator_.MakeBookmark(preload_block_iterator_);
loop_bookmark_ =
        SkipStateBlocks(dispatch_frame_number_, start_bookmark);

// In case of a non-trimmed capture, there are no state blocks, so loop starts here.
if (start_bookmark == loop_bookmark_) {
    for (auto decoder : decoders_) decoder->OnLoopStart();
}
// Otherwise, loop starts at `ProcessStateEndMarker` time.
```

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -256,79 +277,59 @@ void VulkanReplayFrameLoopConsumer::Process_vkFreeDescriptorSets(const ApiCallIn
...
-                                                          VkResult                                    returnValue,
-                                                          format::HandleId                            queue,
-                                                          uint32_t                                    submitCount,
-                                                          StructPointerDecoder<Decoded_VkSubmitInfo>* pSubmits,
-                                                          format::HandleId                            fence)
+void VulkanReplayFrameLoopConsumer::TrackFenceState(format::HandleId device, format::HandleId fence)
 {
-    if (frame_loop_info_.IsLooping() && !frame_loop_info_.IsRepetition())
+    // If fence hasn't been seen yet, check and store the state it is in.
+    if (!per_device_fence_tracking_.contains(device))
```
antonio-lunarg: This if is redundant as `FenceTracking& t = per_device_fence_tracking_[device];` already default constructs.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -345,78 +346,36 @@ void VulkanReplayFrameLoopConsumer::FixupDeviceFences(format::HandleId device, f
...
-        CHECK_VK_RESULT(result, "vkResetFences");
-
-        GFXRECON_LOG_DEBUG("Synthetically signaling fences...");
-        for (VkFence fence : manual_signal_fences)
-        {
-            VulkanQueueInfo* queue_info = table.GetVkQueueInfo(queue);
-            result                      = device_table->QueueSubmit(queue_info->handle, 0, nullptr, fence);
-            CHECK_VK_RESULT(result, "vkDeviceWaitIdle");
-        }
+        VulkanQueueInfo* queue_info = table.GetVkQueueInfo(queue);
```
antonio-lunarg: You can move this line up out of the loop.

framework/decode/api_decoder.h
```diff
@@ -57,6 +57,8 @@ class ApiDecoder
 
     virtual void WaitIdle() = 0;
 
+    virtual void OnLoopStart() {}
```
antonio-lunarg: The name is a bit misleading. I would think `OnLoopStart()` is called repeatedly, but it's only called once when looping starts. How about `StartLooping()`?

framework/decode/preload_file_processor.cpp
antonio-lunarg: In general I do not have a good feeling about this logic being spread in different places, it's difficult to reason about it.
It would be really nice to have a `ProcessFrameBeginMarker`, just after `ProcessStateEndMarker`, that would signal the start of looping if `IsFirstIteration`.

I invoke the expert here, @jzulauf-lunarg what's your opinion, can we have that?

framework/decode/api_decoder.h
nickdriscoll-lunarg: Sounds good to me.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -256,79 +280,75 @@ void VulkanReplayFrameLoopConsumer::Process_vkFreeDescriptorSets(const ApiCallIn
...
     }
 
-    VulkanReplayConsumer::Process_vkWaitForFences(
-        call_info, returnValue, device, fenceCount, pFences, waitAll, timeout);
+    VulkanReplayFrameLoopConsumerBase::Process_vkCreateFence(
+        call_info, returnValue, device, pCreateInfo, pAllocator, pFence);
+
+    if (frame_loop_info_.IsLooping() && !frame_loop_info_.IsRepetition())
+    {
+        // Record this fence as having an initial state of unsignaled
```
antonio-lunarg: Stale comment.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -256,79 +280,75 @@ void VulkanReplayFrameLoopConsumer::Process_vkFreeDescriptorSets(const ApiCallIn
...
-                                                          uint32_t                                    submitCount,
-                                                          StructPointerDecoder<Decoded_VkSubmitInfo>* pSubmits,
-                                                          format::HandleId                            fence)
+void VulkanReplayFrameLoopConsumer::Process_vkDestroyFence(
+    const ApiCallInfo&                                   call_info,
+    format::HandleId                                     device,
+    format::HandleId                                     fence,
+    StructPointerDecoder<Decoded_VkAllocationCallbacks>* pAllocator)
 {
     if (frame_loop_info_.IsLooping() && !frame_loop_info_.IsRepetition())
```
antonio-lunarg: Initial fence states should only contain fences create before the loop range, right? If so we only actually destroy during last iteration, so this conditional should be `frame_loop_info_.IsFinalIteration()`.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -256,79 +280,75 @@ void VulkanReplayFrameLoopConsumer::Process_vkFreeDescriptorSets(const ApiCallIn
...
-            GFXRECON_LOG_DEBUG("VkFence with handle \"%" PRIu64 "\" has been waited on %" PRIu32 " times.",
-                               fence,
-                               t.waited_upon_fences_[fence]);
-        }
+        // Reset the fence on loop repetitions instead of creating it
+        VkDevice device_handle = GetObjectInfoTable().GetVkDeviceInfo(device)->handle;
+        GFXRECON_ASSERT(device_handle != 0);
+        const graphics::VulkanDeviceTable* device_table = GetDeviceTable(device_handle);
+        VkFence  fence_handle = GetObjectInfoTable().GetVkFenceInfo(*pFence->GetPointer())->handle;
+        VkResult res          = device_table->ResetFences(device_handle, 1, &fence_handle);
```
antonio-lunarg: So this is the case that a fence was created within the frame loop range and no destroy has been called yet.
I believe we should not blindly reset the fence, instead we should look at its `initial_fence_states_[fence]` and honor that value.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -492,20 +430,18 @@ void VulkanReplayFrameLoopConsumer::Process_vkQueuePresentKHR(
 {
     VulkanReplayConsumer::Process_vkQueuePresentKHR(call_info, returnValue, queue, pPresentInfo);
 
+    CommonObjectInfoTable& table      = GetObjectInfoTable();
+    VulkanQueueInfo*       queue_info = table.GetVkQueueInfo(queue);
+    format::HandleId       device_id  = queue_info->parent_id;
```
antonio-lunarg: Unused

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -345,78 +365,36 @@ void VulkanReplayFrameLoopConsumer::FixupDeviceFences(format::HandleId device, f
...
-                                             std::numeric_limits<uint64_t>::max());
-        CHECK_VK_RESULT(result, "vkWaitForFences");
-
-        GFXRECON_LOG_DEBUG("Resetting %" PRIu64 " synthetically waited on fences...", manual_wait_fences.size());
-        result = device_table->ResetFences(vk_device, manual_wait_fences.size(), manual_wait_fences.data());
-        CHECK_VK_RESULT(result, "vkResetFences");
-    }
+    // Reset all fences
+    GFXRECON_LOG_DEBUG("Synthetically resetting all %" PRIu64 " observed fences...", all_fences.size());
+    result = device_table->ResetFences(vk_device, all_fences.size(), all_fences.data());
```
antonio-lunarg: Make sure to call this only then `all_fences.size() > 0`

framework/decode/vulkan_replay_frame_loop_consumer.cpp
antonio-lunarg: Or you can rely on the Fixup function to do its job correctly after each present and make this if a simple no-op `return`.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -256,79 +280,75 @@ void VulkanReplayFrameLoopConsumer::Process_vkFreeDescriptorSets(const ApiCallIn
...
         {
-            format::HandleId device = GetObjectInfoTable().GetVkQueueInfo(queue)->parent_id;
-            if (!per_device_fence_tracking_.contains(device))
-            {
-                per_device_fence_tracking_[device] = {};
-            }
             FenceTracking& t = per_device_fence_tracking_[device];
-
-            if (t.signaled_fences_.contains(fence))
+            if (t.initial_fence_states_.contains(fence))
```
antonio-lunarg: No need to check `contains`, just `erase` which is already a no-op if absent.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
nickdriscoll-lunarg: Of course. I wasn't thinking about how FixupDeviceFences already covers this.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -256,79 +280,66 @@ void VulkanReplayFrameLoopConsumer::Process_vkFreeDescriptorSets(const ApiCallIn
...
-            {
-                t.signaled_fences_[fence] = 1;
-            }
-            t.signaled_fences_[fence];
-            GFXRECON_LOG_DEBUG("VkFence with handle \"%" PRIu64 "\" has been signaled %" PRIu32 " times.",
-                               fence,
-                               t.signaled_fences_[fence]);
+            t.initial_fence_states_.erase(fence);
         }
+        VulkanReplayFrameLoopConsumerBase::Process_vkDestroyFence(call_info, device, fence, pAllocator);
```
antonio-lunarg: This should go down, out of the if, as it should run unconditionally.

antonio-lunarg (approved):
One my last comment is addressed, good to go!

framework/decode/vulkan_replay_frame_loop_consumer.cpp
nickdriscoll-lunarg: What an oversight 😓 hopefully this allows CI to complete.

PR #3053, opened 2026-06-25
Fix -m rebind regression from spurious uvec2 BDA candidates

fabian-lunarg:
- Address-replacer: in ResolveBufferAddresses, only keep a candidate whose stored value maps to a tracked capture-address. The uvec2 BDA-heuristic (#3019) flags additional potential BDA candidates; Filtering them avoids entering the dispatch path for non-addresses
- Enable vulkanMemoryModelDeviceScope at replay device creation when the memory-model is enabled and address-replacement is active (injected hashmap shaders use Device-scope atomics; VUID-RuntimeSpirv-vulkanMemoryModel-06265)
- Create the AS-compacted-size query-pool lazily (only on the AS-compaction path) instead of unconditionally in init_queue_assets; BDA-only captures no longer require VK_KHR_acceleration_structure
- SpirVParsingUtil: check_type_potential_ref: only treat uvec2 (and uvec2 arrays) as a potential
  buffer-device-address when module actually converts any uvec2 to pointer (uvec2 -> OpBitcast -> OpConvertUToPtr).

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -3543,6 +3543,50 @@ void VulkanReplayConsumerBase::ModifyCreateDeviceInfo(
                                                      modified_create_info.pNext,
                                                      modified_create_info.pEnabledFeatures,
                                                      options_.remove_unsupported_features);
+
+    // the address-replacer's injected compute-shaders use Device-scope atomics, which require
+    // vulkanMemoryModelDeviceScope when the memory-model is enabled (VUID-RuntimeSpirv-vulkanMemoryModel-06265).
+    auto* memory_model_features =
+        graphics::vulkan_struct_get_pnext<VkPhysicalDeviceVulkanMemoryModelFeatures>(&modified_create_info);
+    auto* vulkan_1_2_features =
+        graphics::vulkan_struct_get_pnext<VkPhysicalDeviceVulkan12Features>(&modified_create_info);
+    const bool memory_model_enabled = (memory_model_features != nullptr && memory_model_features->vulkanMemoryModel) ||
+                                      (vulkan_1_2_features != nullptr && vulkan_1_2_features->vulkanMemoryModel);
+
+    if (memory_model_enabled)
+    {
+        // Mirror UseAddressReplacement(): address-replacement is active when the allocator uses non-opaque addresses.
+        std::unique_ptr<VulkanResourceAllocator> probe_allocator(options_.create_resource_allocator());
+        const bool use_address_replacement = probe_allocator && !probe_allocator->SupportsOpaqueDeviceAddresses();
+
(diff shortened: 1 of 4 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -3543,6 +3543,50 @@ void VulkanReplayConsumerBase::ModifyCreateDeviceInfo(
...
+        graphics::vulkan_struct_get_pnext<VkPhysicalDeviceVulkan12Features>(&modified_create_info);
+    const bool memory_model_enabled = (memory_model_features != nullptr && memory_model_features->vulkanMemoryModel) ||
+                                      (vulkan_1_2_features != nullptr && vulkan_1_2_features->vulkanMemoryModel);
+
+    if (memory_model_enabled)
+    {
+        // Mirror UseAddressReplacement(): address-replacement is active when the allocator uses non-opaque addresses.
+        std::unique_ptr<VulkanResourceAllocator> probe_allocator(options_.create_resource_allocator());
+        const bool use_address_replacement = probe_allocator && !probe_allocator->SupportsOpaqueDeviceAddresses();
+
```
mikes-lunarg: I would prefer if `UseAddressReplacement()` was callable earlier rather than duplicating the logic here. Maybe if you pass a null device_info then it does the probe_allocator inside the same function (so at least the logic co-located)? You could also do a probe as soon as the memory portability arg is parsed and save off the result to options_

framework/util/spirv_parsing_util.cpp
```diff
@@ -338,16 +339,68 @@ bool SpirVParsingUtil::ParseBufferReferences(const uint32_t* const spirv_code, s
...
+    bool allow_uvec2_ref = false;
+    {
+        std::unordered_map<uint32_t, const Instruction*> defs;
+        for (const Instruction& insn : instructions)
+        {
+            if (insn.resultId() != 0)
+            {
+                defs[insn.resultId()] = &insn;
+            }
+        }
```
mikes-lunarg: We do this again below, using `definitions_` as a scratch lookup table, can we build the map once?

mikes-lunarg (approved):
Just a couple of minor observations, up to you if you want to take them. I did some testing with my other vkd3d/uvec2 repro case and it worked as well.

Such a shame, I thought that original uvec2 patch was such a nice little change.

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: done, went with 'pass a null device_info'

framework/util/spirv_parsing_util.cpp
fabian-lunarg: yes, good point. done

fabian-lunarg:
thank you @mikes-lunarg !

PR #3054, opened 2026-06-25
optimize: Split into feature modules

MarkY-LunarG:
Perform the same work done to `gfxrecon-info`, `gfxrecon-convert`, `gfxrecon-replay` and `gfxrecon-extract`  in separating out the various APIs supported by our tools into seperate features for maintainability and future modification.

### DIFF
--- a/tools/optimize/main.cpp
+++ b/tools/optimize/main.cpp
@@ -1,5 +1,5 @@
 /*
-** Copyright (c) 2020 LunarG, Inc.
+** Copyright (c) 2020-2026 LunarG, Inc.
 ** Copyright (c) 2022-2023 Advanced Micro Devices, Inc. All rights reserved.
 **
 ** Permission is hereby granted, free of charge, to any person obtaining a
@@ -22,29 +22,21 @@
 */
 
 #include PROJECT_VERSION_HEADER_FILE
-#include "file_optimizer.h"
 
-#include "../tool_settings.h"
-
-#if defined(D3D12_SUPPORT)
-#include "dx12_optimize_util.h"
-#endif
+#include "optimize_feature.h"
+#include "tool_settings.h"
 
 #include "decode/decode_api_detection.h"
-#include "decode/dx12_optimize_options.h"
 #include "decode/file_processor.h"
-#include "format/format.h"
-#include "generated/generated_vulkan_decoder.h"
-#include "generated/generated_vulkan_referenced_resource_consumer.h"
-#include "generated/generated_vulkan_referenced_block_consumer.h"
 #include "util/argument_parser.h"
-#include "util/logging.h"
 #include "util/date_time.h"
+#include "util/feature_module_registry.h"
+#include "util/logging.h"
 
-#include <cassert>
+#include <algorithm>
+#include <memory>
 #include <stdexcept>
 #include <string>
-#include <unordered_set>
 #include <vector>
 
 #if defined(WIN32)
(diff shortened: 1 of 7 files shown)


MarkY-LunarG:
@davidd-lunarg, this touches DX12 changes as part of the optimize tool.  Please review that at least.

tools/optimize/main.cpp
```diff
@@ -94,167 +90,63 @@ static void PrintUsage(const char* exe_name)
...
 }
 
-// GetUnreferencedBlocksResult defines a simple aggregate return-type
-struct GetUnreferencedBlocksResult
+int32_t main(int32_t argc, const char** argv)
 {
-    uint64_t                     num_blocks = 0;
-    std::unordered_set<uint64_t> unreferenced_blocks;
-};
+    int32_t success    = -1;
```
fabian-lunarg: this is error-prone. could `success` be `bool` instead?
and later we `return success ? EXIT_SUCCESS : EXIT_FAILURE;`

we don't seem to use other return-codes and `success = -1` can slip unchanged through the introduced goto handling.

tools/optimize/main.cpp
```diff
@@ -279,93 +171,59 @@ int main(int argc, const char** argv)
...
             }
         }
-        // Manual mode. Follow user instructions.
-        else
+
+        if (!any_ran)
         {
-            RunDx12Optimizations(input_filename, output_filename, dx12_options);
+            GFXRECON_LOG_ERROR("Could not detect graphics API. Aborting optimization.");
+            goto exit;
```
fabian-lunarg: I'd prefer not using goto for control-flow. it would be more readable without.

the only purpose of the `exit` lable seems to be logger-cleanup.
this could be done with RAII like `struct LogGuard { ~LogGuard() { gfxrecon::util::Log::Release(); } } log_guard;` at the top.
any and all `goto exit` + `exit()` statements change to early returns, the return codes are explicit (return EXIT_FAILURE/SUCCESS;)

what do you think?

tools/optimize/optimize_dx12_feature.h
```diff
@@ -0,0 +1,72 @@
...
+#include "decode/dx12_optimize_options.h"
+#include "generated/generated_dx12_decoder.h"
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(optimize)
+
+class OptimizeDx12Feature : public OptimizeFeature
+{
+  public:
+    OptimizeDx12Feature() : detection_consumer_(decode::Dx12DetectionConsumer::kNoBlockLimit) {}
```
fabian-lunarg: this is a bit tricky but it looks to me :

`kNoBlockLimit` here makes just the API detection scan the whole file.

(in main() where at comment `// Run the detection pass so each feature can identify its API in the capture.`)
it's the same in the `OptimizeVulkanFeature` ctor.
that's a perf-regression and windows-only issue, because only there we run two consumers vulkan/dx12.

to match the old behavior this should be `kDefaultBlockLimit` (cap ~1000, early-exit).
but only changing here is still wrong, the old code  (`DetectAPIs`) handled that by being two-stage.
run with a cap first, re-run unbounded only if nothing was detected.

so, in order to achieve this requires creating the consumer-members with those limits
and pass them in, per pass. 
so basically defer contruction of `detection_consumer_` to `RegisterDetectionDecoder(decode::FileProcessor& fp, uint64_t block_limit)` <- adding the limit there and create consumer with that

MarkY-LunarG: @fabian-lunarg , I've modifided it to have to two-pass mechanics as before.  Hopefully this will resolve your concerns.  Let me know.

tools/optimize/optimize_dx12_feature.cpp
```diff
@@ -0,0 +1,146 @@
+/*
+** Copyright (c) 2026 LunarG, Inc.
+** Copyright (c) 2022-2023 Advanced Micro Devices, Inc. All rights reserved.
```
lunarpapillo, line 3: Seems like a strange copyright attribution for a new file created in 2026.

Often, LunarG and some other company co-own the copyright - not the case here?

MarkY-LunarG: Dang, good catch.  Will fix.

tools/optimize/main.cpp
```diff
@@ -94,179 +91,80 @@ static void PrintUsage(const char* exe_name)
...
-    gfxrecon::util::Log::SetFatalCallback([](const char* message) { throw std::runtime_error(message); });
 
-    gfxrecon::util::ArgumentParser arg_parser(argc, argv, kOptions, kArguments);
+    gfxrecon::util::ArgumentParser arg_parser(argc, argv, options, arguments);
 
     if (CheckOptionPrintUsage(argv[0], arg_parser) || CheckOptionPrintVersion(argv[0], arg_parser))
     {
-        gfxrecon::util::Log::Release();
-        exit(0);
+        exit(EXIT_SUCCESS);
```
fabian-lunarg: need `return EXIT_SUCCESS` here, so `LogGuard` dtor fires (`exit()` would circumvent it)
same for all other `exit()`

tools/optimize/optimize_dx12_feature.h
fabian-lunarg: yes exactly, `run_detection_pass(Feature::kDefaultDetectionBlockLimit);` nails it

fabian-lunarg (approved):

PR #3055, opened 2026-06-25
Index copy semaphores in virtual swapchain by replay image index

### LINKED ISSUE #2692: [replay] Virtual swapchain does not synchronize correctly when a present fence is used
This is very easy to reproduce, the scenario is the following:
For every swapchain image 2 semaphores and a fence are created. In each frame the steps are:

- wait for fence
- acquire image with semaphore 1
- submit with waiting on semaphore 1 and signaling semaphore 2
- present with waiting on semaphore 2 and signaling fence

The same fence is always used with the same 2 semaphores, and by waiting on the fence at the start of the frame it is guaranteed that the semaphores are not in use anymore.

But the virtual swapchain in the replay creates more semaphore and fences than the app did, and the connection between the semaphores and fence is not kept, meaning the semaphores can be reused while still in use by the driver. The validation layers report `VUID-vkQueueSubmit-pSignalSemaphores-00067`

wzchu-lunarg:
We commonly see validation errors `VUID-vkQueueSubmit-pSignalSemaphores-00067` during replay of a wide variety of well synchronized applications, and many are likely from the queue submit associated with copying images from the virtual swapchain to the presentation swapchain. This is due to the `copy_semaphore` not being unsignaled, which should happen by the time presentation is finished. However, our logic doesn't let us know it is unsignaled, or that presentation is finished.

Example:
Frame 1: signal on semaphore associated to `capture_index = 0`, present on `replay_index = 1` 
Frame 2: signal on semaphore associated to `capture_index = 1`, present on `replay_index = 2` 
Frame 3: signal on semaphore associated to `capture_index = 0`, present on `replay_index = 0`

On frame 3, we don't know if the semaphore with `capture_index = 0` is unsignaled. The semaphore could still be signaled, hence the validation error. However, `vkQueuePresent` doesn't tell or signal us that the semaphore can be reused. 

Traditionally (there are workaround extensions), Vulkan let's us know an image's presentation is done if `AcquireNextImage` gives that index, which in the example above is a perfectly valid and released `replay_index = 0`. However, as the semaphore is not associated with the replay/real indices we have no idea if that semaphore's presentation is finished.

It shouldn't matter whether capture has more swapchain images or replay has more swapchain images, as long as the connection between the presentation image and the semaphore are not mapped together, and that there are at least two images in both swapchains, this can be an issue. All other combinations of image counts can have this subcase.

Subcase example:
Frame 1: signal on semaphore associated to `capture_index = 0`, present on `replay_index = 0` 
Frame 2: signal on semaphore associated to `capture_index = 1`, present on `replay_index = 1` 
Frame 3: signal on semaphore associated to `capture_index = 0`, present on `replay_index = 1` 
Acquire on frame 3 gives index 1 (which is perfectly valid), but still unknown if semaphore at `capture_index = 0` is unsignaled yet.

Change copy semaphores to use replay indices.

Fixes https://github.com/LunarG/gfxreconstruct/issues/2692

### DIFF
--- a/framework/decode/vulkan_virtual_swapchain.cpp
+++ b/framework/decode/vulkan_virtual_swapchain.cpp
@@ -365,11 +365,11 @@ VkResult VulkanVirtualSwapchain::CreateSwapchainResourceData(const VulkanDeviceI
             // Create one command buffer per queue per swapchain image so that we don't reset a command buffer that
             // may be in active use.
             uint32_t command_buffer_count = static_cast<uint32_t>(copy_cmd_data.command_buffers.size());
-            if (command_buffer_count < capture_image_count)
+            if (command_buffer_count < *replay_image_count)
             {
-                copy_cmd_data.command_buffers.resize(capture_image_count);
+                copy_cmd_data.command_buffers.resize(*replay_image_count);
 
-                uint32_t                    new_count     = capture_image_count - command_buffer_count;
+                uint32_t                    new_count     = *replay_image_count - command_buffer_count;
                 VkCommandBufferAllocateInfo allocate_info = {
                     VK_STRUCTURE_TYPE_COMMAND_BUFFER_ALLOCATE_INFO, // sType
                     nullptr,                                        // pNext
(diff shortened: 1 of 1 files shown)


wzchu-lunarg:
We should actually change all the copy resources to use replay indices.

This is because the resources should be tied to our frames in flight, which is our hardware acquired images.

We can imagine it as the following based on the number of images in each swapchain:
1. Capture images = 3, Replay images = 2 => We can only have 2 acquired images and therefore only two copies in flight. No third image to even write to if we have a copy resource.

2. Capture images = 2, Replay images = 3 => Can have up to 3 acquired images at once, and would need 3 sets of copy resources. Even if it reads the same virtual image concurrently (which it shouldn't since queue present should happen after doing a replay of a synchronized frame in the captured app), concurrent reads are allowed anyways. If there were only 2, then some of the copys will be stalled by waiting for the fence even if we are supposed to have 3 frames in flight. Furthermore, `TransitionSwapchainImage` seems to pass the replay index and accesses those same copy resources based on replay index.

MarkY-LunarG:
Can someone other than me review this?  I kind of suggested changes as he went along already and would like someone elses feedback for him.  @fabian-lunarg , @davidd-lunarg , @bradgrantham-lunarg ?

fabian-lunarg (approved):
I agree with the fix, the invariant -> 'sync primitives follow replay-image lifecycle'
basically nails it.
worth pointing out the PR also fixes a latent UB (deferred-alloc codepath) in `TransitionSwapchainImage`:
it already indexed command_buffers/fences by the replay acquired index, but vectors were sized to `capture_image_count`

PR #3062, opened 2026-06-30
Fix offscreen swapchain implicit synchronization

marius-pelegrin-arm:
The offscreen swapchain needs to emulate the behavior of a WSI. In particular, this implies waiting for and signaling semaphores and fences on image acquisition and queue presentation.

The behavior until now was to submit an empty workload as a queue submission waiting and signaling the correct synchronization primitives.
While this works as expected for queue presentation, this introduces additional synchronizations for image acquisition.

Indeed, image acquisition is queue independent, which means it doesn't need to wait for previous queue submissions to begin. By implementing it as an empty queue submit, we encounter cases where the implicit ordering constraints of Vulkan introduce a slowdown, making offscreen slower than onscreen.

I have currently no solution for Windows and/or Mac (assuming they don't support `VK_KHR_external_semaphore_fd` and `VK_KHR_external_fence_fd`).
However, on systems implementing these extensions (mainly Linux and Android systems), I propose to solve this using a technique inspired by Mesa's Vulkan WSI Layer:
https://gitlab.freedesktop.org/mesa/vulkan-wsi-layer

The idea is to signal semaphores and fences by importing them as temporary file descriptor system handles using a predefined "invalid" value recognized as an "already signaled primitive".

This behavior is described in the Vulkan specification for the `VkImportSemaphoreFdInfoKHR` and `VkImportFenceFdInfoKHR` structures (search for file descriptor value "-1").

### DIFF
--- a/framework/decode/vulkan_offscreen_swapchain.cpp
+++ b/framework/decode/vulkan_offscreen_swapchain.cpp
@@ -23,9 +23,23 @@
 #include "decode/vulkan_offscreen_swapchain.h"
 #include "encode/vulkan_handle_wrapper_util.h"
 #include "decode/decoder_util.h"
+#include "generated/generated_vulkan_enum_to_string.h"
+#include "util/callbacks.h"
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
+
+void VulkanOffscreenSwapchain::CleanDeviceResources(VkDevice device, const graphics::VulkanDeviceTable* device_table)
+{
+    VulkanVirtualSwapchain::CleanDeviceResources(device, device_table);
+
+    external_sync_type_.erase(device);
+}
+
+void VulkanOffscreenSwapchain::SetExternalSyncType(VkDevice device, ExternalSyncType external_sync_type)
+{
+    external_sync_type_[device] = external_sync_type;
+}
 
 VkResult VulkanOffscreenSwapchain::CreateSurface(VkResult                             original_result,
                                                  VulkanInstanceInfo*                  instance_info,
@@ -174,26 +188,9 @@ VkResult VulkanOffscreenSwapchain::AcquireNextImageKHR(VkResult                 
     *image_index = capture_image_index;
     if (semaphore != VK_NULL_HANDLE || fence != VK_NULL_HANDLE)
     {
-        uint32_t           signal_semaphore_count = 0;
-        const VkSemaphore* signal_semaphores      = nullptr;
-
-        if (semaphore != VK_NULL_HANDLE)
-        {
-            signal_semaphore_count = 1;
-            signal_semaphores      = &semaphore;
-        }
-
-        VkResult result = SignalSemaphoresFence(nullptr, 0, nullptr, signal_semaphore_count, signal_semaphores, fence);
-        if (result != VK_SUCCESS)
-        {
-            GFXRECON_LOG_ERROR("Offscreen swapchain failed to singal semaphore (Handle = %" PRIu64
-                               ") or fence (Handle = %" PRIu64 ") for swapchain (ID = %" PRIu64
-                               ") on AcquireNextImageKHR",
-                               semaphore,
-                               fence,
-                               swapchain_info->capture_id);
-            return result;
-        }
+        auto it = external_sync_type_.find(device_info->handle);
+        GFXRECON_ASSERT(it != external_sync_type_.end());
+        return SignalAcquireNextImageSemaphoreFence(device_info, semaphore, fence, it->second);
     }
     return original_result;
 }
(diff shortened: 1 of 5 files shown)


framework/decode/vulkan_offscreen_swapchain.cpp
```diff
@@ -308,40 +277,97 @@ void VulkanOffscreenSwapchain::PresentImageAdHoc(const VulkanDeviceInfo*
...
+            submit_info.pSignalSemaphores    = (semaphore == VK_NULL_HANDLE ? nullptr : &semaphore);
+
+            VkResult result = device_table_->QueueSubmit(default_queue_, 1, &submit_info, fence);
+
+            if (result != VK_SUCCESS)
+            {
+                GFXRECON_LOG_FATAL("Offscreen swapchain failed to signal semaphore and fence by submitting and empty "
+                                   "queue submission. (%s)",
+                                   util::ToString<VkResult>(result).c_str());
+                std::abort();
```
fabian-lunarg: `std::abort();` became redundant, `GFXRECON_LOG_FATAL` is now actually fatal.
same below

framework/decode/vulkan_offscreen_swapchain.cpp
```diff
@@ -254,37 +238,22 @@ VkResult VulkanOffscreenSwapchain::QueuePresentKHR(VkResult
...
-            queue_info, present_info->waitSemaphoreCount, present_info->pWaitSemaphores, 0, nullptr, VK_NULL_HANDLE);
+        util::MarkInjectedCommandsHelper mark_injected_commands_helper;
+        result = device_table_->QueueSubmit(queue_info->handle, 1, &submit_info, VK_NULL_HANDLE);
+
         if (result != VK_SUCCESS)
         {
-            GFXRECON_LOG_ERROR("Offscreen swapchain failed to wait semaphore (Handle = %" PRIu64
-                               ") for queue (ID = %" PRIu64 ") on QueuePresentKHR",
-                               present_info->pWaitSemaphores[0],
+            GFXRECON_LOG_ERROR("Offscreen swapchain failed to QueueSubmit on QueuePresentKHR for queue " PRIu64,
```
fabian-lunarg: missing % before PRIu64 (-> "... for queue %" PRIu64)

framework/decode/vulkan_offscreen_swapchain.cpp
```diff
@@ -308,40 +277,97 @@ void VulkanOffscreenSwapchain::PresentImageAdHoc(const VulkanDeviceInfo*
...
+            submit_info.commandBufferCount   = 0;
+            submit_info.pCommandBuffers      = nullptr;
+            submit_info.signalSemaphoreCount = (semaphore == VK_NULL_HANDLE ? 0 : 1);
+            submit_info.pSignalSemaphores    = (semaphore == VK_NULL_HANDLE ? nullptr : &semaphore);
+
+            VkResult result = device_table_->QueueSubmit(default_queue_, 1, &submit_info, fence);
+
+            if (result != VK_SUCCESS)
+            {
+                GFXRECON_LOG_FATAL("Offscreen swapchain failed to signal semaphore and fence by submitting and empty "
```
fabian-lunarg: nit: typo -> "... an empty"

fabian-lunarg: and actually thinking this should not be fatal here, keeping current behavior intact.
the QueueSubmit fallback-path (so basically matching current dev) should instead just forward any errors. 
meaning we need to restore `SignalAcquireNextImageSemaphoreFence` returning a `VkResult` (it's void now) and propagate it.

framework/decode/vulkan_swapchain.cpp
```diff
@@ -46,6 +46,16 @@ void VulkanSwapchain::Clean()
     }
 }
 
+void VulkanSwapchain::CleanDeviceResources(VkDevice device, const graphics::VulkanDeviceTable* device_table)
```
fabian-lunarg: the placement in the base class is problematic.
this is in fact unreachable code for the offscreen swapchain code-paths (VulkanOffscreenSwapchain derives from VulkanVirtualSwapchain, which overrides CleanDeviceResources and doesn't chain to the base). So `external_sync_type_` is never erased on device destroy.

instead the `external_sync_type_` map and its full lifetime:
SetExternalSyncType / lookup / cleanup would be better placed in `VulkanOffscreenSwapchain`.
it's the only user. that fixes both the cleanup and keeps the base class free of offscreen-specific state.

framework/decode/vulkan_replay_consumer_base.h
```diff
@@ -286,6 +286,7 @@ class VulkanReplayConsumerBase : public VulkanConsumer
         std::vector<VkPhysicalDevice>             replay_device_group;
         graphics::VulkanDeviceUtil                device_util;
         graphics::VulkanDevicePropertyFeatureInfo property_feature_info;
+        VulkanSwapchain::ExternalSyncType         external_sync_type;
```
fabian-lunarg: this is only read on offscreen code-paths. but still would prefer we would initialize  `= ExternalSyncType::QueueSubmit` here

fabian-lunarg (changes requested):

marius-pelegrin-arm:
I don't understand, what changes are you requesting ? I updated the commit with all changes requested in comments.

fabian-lunarg:
> I don't understand, what changes are you requesting ? I updated the commit with all changes requested in comments.

I reopened because from my perspective I did not see any codechanges. maybe a wrong commit got pushed?

marius-pelegrin-arm:
Oh, apparently there has been a git mess up. I will re-push the changes.

fabian-lunarg:
> Oh, apparently there has been a git mess up. I will re-push the changes.

no worries. but please, before just closing my open concerns, give me opportunity to verify the changes.

marius-pelegrin-arm:
Oh, okay, I thought marking conversation as resolved was the proper way to signal that I pushed modifications that should solve the issue.

fabian-lunarg:
> Oh, okay, I thought marking conversation as resolved was the proper way to signal that I pushed modifications that should solve the issue.

yes, for trivial concerns or one-liners normally not an issue (if there are actual changes).
the base-class / cleanup concern was not that trivial, so nicer to apply the change and let me catch up first.

marius-pelegrin-arm:
Pushed.

fabian-lunarg (approved):
alright!

PR #3063, opened 2026-07-01
Refactor: use format::kName* constants directly instead of NameXxx() wrappers

### LINKED ISSUE #1364: Refactor VulkanExportJsonConsumerBase to Use Names of Json Fields Directly
Remove functions like `constexpr const char* NameFunction() const { return format::kNameFunction; }` and have their callers use constants like `format::kNameFunction` directly.

These functions date to a time that a possible requirement to support more than one JSON schema was discussed, but that never actually became concrete.

**Priority:** P∞, it is just a tidy-up with no functional or performance impact.

KossiSessou:
## What does this PR do?

This PR removes the thin `NameXxx()` wrapper methods (e.g. `NameReturn()`, `NameArgs()`) from `VulkanExportJsonConsumerBase` and `OpenXrExportJsonConsumerBase` and replaces every call site with the underlying `format::kName*` constant directly (`NameReturn()` → `format::kNameReturn`, and so on). It also updates the Khronos/Vulkan JSON-consumer code generators so regenerated output emits the constants directly and never reintroduces the wrappers. The result is the same field-name-constant pattern the DX12 JSON consumer already uses, applied consistently to the Vulkan and OpenXR consumers.

## Why was this PR needed?

As described in #1364, the `NameXxx()` methods are leftover indirection from an earlier design that would have supported multiple JSON schemas — a design that was never implemented. Each method does nothing but return a single constant, so every call site pays for a purposeless function-call hop that resolves to a compile-time constant anyway. The DX12 side of the codebase had already been migrated to reference the constants directly, leaving the Vulkan and OpenXR consumers as the inconsistent outliers.

Investigation revealed the scope is larger than the base-class methods alone: roughly 1,861 of the ~1,884 call sites live in **generated** files (`generated_vulkan_json_consumer.cpp`, `generated_openxr_json_consumer.cpp`). Editing only the generated output would be silently reverted on the next codegen run, so the two Python generators that emit those files had to be updated in the same change. Two wrappers — `NameCommandIndex()` and `NameSubmitIndex()` — returned raw string literals (`"cmd_index"`, `"sub_index"`) rather than existing constants, so two new constants (`format::kNameCommandIndex`, `format::kNameSubmitIndex`) are added to `format_json.h` first, giving those call sites something to reference.

This is a purely structural change: the JSON field-name strings emitted at runtime are unchanged, so the emitted output is byte-for-byte identical before and after.

## What are the relevant issue numbers?

Closes #1364

## Screenshots / Recordings (if applicable)

Not a UI change. Two checks confirm correctness:

**1. No wrapper calls remain anywhere** (hand-edited files, generated files, and generator templates):

```console
$ grep -rn "NameReturn()\|NameArgs()\|NameCommandIndex()\|NameSubmitIndex()" \
    framework/ --include="*.cpp" --include="*.h" --include="*.py"
$ echo "matches: $?"
matches: 1        # grep exit 1 = zero matches
```

**2. Generators and generated files agree (regeneration is deterministic).** Re-running the project's own generators against the pinned Vulkan/OpenXR registries and diffing against the committed files:

| Generated file | Diff vs committed |
|----------------|-------------------|
| `generated_vulkan_json_consumer.cpp` | 0 lines (byte-identical) |
| `generated_openxr_json_consumer.cpp` | 0 lines (byte-identical) |

Both regenerate byte-for-byte identical to what is committed, confirming the generators and generated code stay consistent across future codegen runs.

## Does this PR meet the acceptance criteria?

- [ ] Tests added for new/changed behavior — *N/A: pure mechanical refactor with zero behavioral change; no new symbols, no changed runtime strings. Validation is via the regeneration-determinism check above rather than a new test.*
- [x] All tests passing — no behavioral change; emitted JSON is unchanged. Full build/test coverage is provided by upstream CI on this PR.
- [x] Follows project style guide — hand-edited C++ files passed through `clang-format` per CONTRIBUTING.md; branch targets `dev`.
- [x] No breaking changes introduced — output is byte-for-byte identical before and after.
- [x] Documentation updated (if applicable) — N/A; no public API or user-facing docs affected.

---

*This is my first contribution to GFXReconstruct, submitted as part of the CodePath open-source program. Happy to adjust anything to match project conventions.*

### DIFF
--- a/framework/decode/openxr_json_consumer_base.cpp
+++ b/framework/decode/openxr_json_consumer_base.cpp
@@ -77,8 +77,8 @@ void OpenXrExportJsonConsumerBase::Process_xrInitializeLoaderKHR(
     StructPointerDecoder<Decoded_XrLoaderInitInfoBaseHeaderKHR>* loaderInitInfo)
 {
     nlohmann::ordered_json& jdata = WriteApiCallStart(call_info, "xrInitializeLoaderKHR");
-    jdata[NameReturn()]           = returnValue;
-    auto& args                    = jdata[NameArgs()];
+    jdata[format::kNameReturn]    = returnValue;
+    auto& args                    = jdata[format::kNameArgs];
     switch (loaderInitInfo->GetPointer()->type)
     {
         default:
@@ -100,8 +100,8 @@ void OpenXrExportJsonConsumerBase::Process_xrCreateApiLayerInstance(
     HandlePointerDecoder<XrInstance>*                   instance)
 {
     nlohmann::ordered_json& jdata = WriteApiCallStart(call_info, "xrCreateApiLayerInstance");
-    jdata[NameReturn()]           = returnValue;
-    auto& args                    = jdata[NameArgs()];
+    jdata[format::kNameReturn]    = returnValue;
+    auto& args                    = jdata[format::kNameArgs];
     FieldToJson(args["info"], info);
     FieldToJson(args["layerInfo"], layerInfo);
     HandleToJson(args["instance"], instance);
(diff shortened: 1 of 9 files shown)


KossiSessou:
Hi @andrew-lunarg — this is my first contribution to GFXReconstruct. This PR addresses #1364 by removing the `NameXxx()` JSON field-name wrappers from the Vulkan and OpenXR export consumers and referencing the `format::kName*` constants directly, matching the pattern the DX12 consumer already uses. The two code generators are updated in the same change so regenerated output stays consistent (verified byte-identical via regeneration). No behavioral change — emitted JSON is unchanged. Would appreciate a review when you have time, and happy to adjust anything to fit project conventions.

mikes-lunarg (approved):
The changes look good. Once it finishes running through CI we will get it merged. Thank you for your contribution!

KossiSessou:
Of course @mikes-lunarg. Looking forward to contribute more to this project!

PR #3066, opened 2026-07-06
replay / virtual swapchain: Fix two VVL issues

fabian-lunarg:
when replaying a capture using VVL via the `--validate` option I'm seeing issues that are not present
in the captured application.

per swapchain-image (3x in my case)
```
[gfxrecon] ERROR - DEBUG MESSENGER: Frame #1: UNASSIGNED-non-acquired-swapchain-image-used: vkQueueSubmit(): pSubmits[0] performs a layout transition on presentable VkImage 0x18b000000018b, but the image has not been acquired from VkSwapchainKHR 0x18a000000018a (either never or since the last present operation).
The Vulkan spec states: Use of a presentable image must occur only after the image is returned by vkAcquireNextImageKHR, and before it is released by vkQueuePresentKHR. This includes transitioning the image layout and rendering commands (https://docs.vulkan.org/refpages/latest/refpages/source/VkSwapchainKHR.html#_description)
```

and in addition for trimmed captures, also multiple times:

```
[gfxrecon] ERROR - DEBUG MESSENGER: Frame #1: VUID-vkCmdDraw-None-09600: vkQueueSubmit2(): pSubmits[0] command buffer VkCommandBuffer 0x2faaef20 expects VkImage 0x1950000000195 (subresource: aspectMask = VK_IMAGE_ASPECT_COLOR_BIT, mipLevel = 0, arrayLayer = 0) to be in layout VK_IMAGE_LAYOUT_PRESENT_SRC_KHR--instead, current layout is VK_IMAGE_LAYOUT_UNDEFINED.
The Vulkan spec states: If a descriptor with type equal to any of VK_DESCRIPTOR_TYPE_SAMPLE_WEIGHT_IMAGE_QCOM, VK_DESCRIPTOR_TYPE_BLOCK_MATCH_IMAGE_QCOM, VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE, VK_DESCRIPTOR_TYPE_STORAGE_IMAGE, or VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT is accessed as a result of this command, all image subresources identified by that descriptor must be in the image layout identified when the descriptor was written (https://vulkan.lunarg.com/doc/view/1.4.350.0/linux/antora/spec/latest/chapters/drawing.html#VUID-vkCmdDraw-None-09600)
```

both errors are coming from our virtual swapchain and are fixed in this PR.


framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -1124,6 +1145,138 @@ VkResult VulkanVirtualSwapchain::QueuePresentKHR(VkResult
...
+                             swapchain_info->capture_id,
+                             queue_family_index);
+        return;
+    }
+
+    // trim-state setup has already run, so mark as injected from here
+    util::MarkInjectedCommandsHelper mark_injected_commands_helper;
+
+    VkCommandPool   transition_pool    = VK_NULL_HANDLE;
+    VkCommandBuffer transition_command = VK_NULL_HANDLE;
```
panos-lunarg: Perhaps you can reuse `TemporaryCommandBuffer` located in `framework/decode/vulkan_replay_dump_resources_common.h` (after moving it into a different source file that does not imply that it is used only by VDR) and save some (admittedly not much)  of the boilerplate?

panos-lunarg (approved):
Apart from a minor comment looks good

framework/decode/vulkan_virtual_swapchain.cpp
fabian-lunarg: I'll look into it!

fabian-lunarg: yes, agreed. could reuse 1:1 here and good opportunity to source into an external helper. will add another commit

fabian-lunarg: done. the common utils now live in `decode/vulkan_temporary_objects.h` (only clients are in decode).
-> moves some boilerplate both from here and in dump_resources.

framework/decode/vulkan_temporary_objects.h
panos-lunarg, line 1: Perhaps `framework/graphics` folder is a better location for the new files?

fabian-lunarg: that was the first location I thought about. but the structs are decode-typed: `TemporaryCommandBuffer` holds a `decode::VulkanDeviceInfo&` and all current clients are in framework/decode. 
so this is a more local placement, relative to its clients. all tools can access it.
it would make sense to move it to framework/graphics when it wouldn't use `VulkanDeviceInfo` 
and have clients outside of `framework/decode`.

panos-lunarg: Ah i see. Guess it's fine then

PR #3069, opened 2026-07-06
Unify tool PrintUsage and PrintVersion funcs

MarkY-LunarG:
We had multiple versions of the PrintUsage and PrintVersion functions, this unifies them.
Additionally, this moves the "global" defines of these from tool_settings.h which resulted in requiring a #define to filter them out of some locations, and into a new header called tool_command_line.h

### DIFF
--- a/tools/compress/main.cpp
+++ b/tools/compress/main.cpp
@@ -22,6 +22,8 @@
 */
 
 #include PROJECT_VERSION_HEADER_FILE
+#include "tool_command_line.h"
+
 #include "compression_converter.h"
 
 #include "decode/file_processor.h"
@@ -30,15 +32,10 @@
 #include "util/compressor.h"
 #include "util/logging.h"
 
-#include "vulkan/vulkan_core.h"
-
 #include <cassert>
 #include <cstdlib>
 
-const char kHelpShortOption[] = "-h";
-const char kHelpLongOption[]  = "--help";
-const char kVersionOption[]   = "--version";
-const char kNoDebugPopup[]    = "--no-debug-popup";
+const char kNoDebugPopup[] = "--no-debug-popup";
 
 const char kOptions[] = "-h|--help,--version,--no-debug-popup";
 
@@ -50,12 +47,7 @@ const char kArgUnknown[] = "<Unknown>";
 
 static void PrintUsage(const char* exe_name)
 {
-    std::string app_name     = exe_name;
-    size_t      dir_location = app_name.find_last_of("/\\");
-    if (dir_location >= 0)
-    {
-        app_name.replace(0, dir_location + 1, "");
-    }
+    std::string app_name = GetApplicationName(exe_name);
     GFXRECON_WRITE_CONSOLE("\n%s - A tool to compress/decompress GFXReconstruct capture files.\n", app_name.c_str());
     GFXRECON_WRITE_CONSOLE("Usage:");
     GFXRECON_WRITE_CONSOLE("  %s [-h | --help] [--version] <input_file> <output_file> <compression_format>\n",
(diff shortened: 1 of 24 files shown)


tools/compress/main.cpp
```diff
@@ -22,6 +22,8 @@
 */
 
 #include PROJECT_VERSION_HEADER_FILE
+#include "../tool_command_line.h"
```
mikes-lunarg: Can we fix the include path in CMakeLists.txt so we don't have to do a relative include here? There are a couple more places in this PR that do the same thing.

tools/optimize/main.cpp
```diff
@@ -56,12 +57,8 @@ static std::vector<std::unique_ptr<gfxrecon::optimize::OptimizeFeature>> g_optim
...
 static void PrintUsage(const char* exe_name)
 {
-    std::string app_name     = exe_name;
-    size_t      dir_location = app_name.find_last_of("/\\");
-    if (dir_location >= 0)
-    {
-        app_name.replace(0, dir_location + 1, "");
-    }
+    std::string app_name = GetApplicationName(exe_name);
+    GFXRECON_WRITE_CONSOLE("\n%s - Produce new captures with enhanced performance characteristics", app_name.c_str());
```
mikes-lunarg: This new write `GFXRECON_WRITE_CONSOLE` duplicates a similar call below

tools/replay/CMakeLists.txt
```diff
@@ -31,6 +31,7 @@ add_executable(gfxrecon-replay "")
 target_sources(gfxrecon-replay
                PRIVATE
                     ${CMAKE_CURRENT_LIST_DIR}/../tool_settings.h
+                   ${CMAKE_CURRENT_LIST_DIR}/../tool_command_line.h
```
mikes-lunarg: nit: alignment

tools/tool_settings.h
```diff
@@ -1476,52 +1473,4 @@ static gfxrecon::decode::DxReplayOptions GetDxReplayOptions(const gfxrecon::util
 }
 #endif
 
-// Only provide the usage functions if a define is not present
-#if !defined(GFXR_TOOL_SETTINGS_NO_USAGE)
```
mikes-lunarg, line 1480: This guard was removed in the new version. Was that intentional and were there any users of it?

tools/tool_command_line.h
```diff
@@ -0,0 +1,113 @@
...
+{
+    std::string app_name = GetApplicationName(exe_name);
+
+    GFXRECON_WRITE_CONSOLE("%s version info:", app_name.c_str());
+    GFXRECON_WRITE_CONSOLE("  " GFXRECON_APP_NAME_PREFIX "GFXReconstruct Version %s", GetProjectVersionString());
+}
+
+// Only provide the option-check functions if a define is not present. Translation units that
+// only need the option string constants or GetApplicationName()/PrintVersionHeader() above (and
+// never call CheckOptionPrintUsage()) can define this first to avoid an unused static function.
```
mikes-lunarg: This comment seems to be an expanded version of the original comment about the now-removed `GFXR_TOOL_SETTINGS_NO_USAGE` guard

mikes-lunarg (changes requested):

tools/tool_settings.h
MarkY-LunarG: It was only needed because we had some internal (non-main.cpp files) which needed tool settings info and the hanging PrintUsage() function caused compilation issues.  Since this content is moved to another header, it is no longer needed since I include both headers in main.cpp and only the tool_settings.h file in the non-main.cpp files.

tools/optimize/main.cpp
MarkY-LunarG: Copy-paste error.

tools/tool_settings.h
mikes-lunarg: Ok makes sense. I see the only usage of that #define is in optimize_dx12_feature.cpp, that should probably just come out with this change.

MarkY-LunarG: Oops, good catch

mikes-lunarg (approved):

PR #3075, opened 2026-07-07
Hashing memory tracking method

panos-lunarg:
Introduces a new memory tracking method called hashing.
As the name implies it detects writes by hashing the memory content. Does not use shadow memory and does not rely on OS provided mechanisms like page_guard or userfaultfd.
To enable set `GFXRECON_MEMORY_TRACKING_MODE` / `debug.gfxrecon.memory_tracking_mode` to `hashing`


per-mathisen-arm:
We've tested it with some content that had conflicts with the page guard and while pretty slow, it worked fine and allowed captures to be taken that would otherwise not been possible, so this seems like a great addition!

panos-lunarg:
I am seeing visual artifacts during replay with some applications (solar bay).
I believe this happens because this method cannot tell apart cpu from gpu writes and both are detected and dumped into the file. But detecting gpu writes can be problematic as they can happen at the same time as the detection is performed and therefore be dumped in the file partially

PR #3076, opened 2026-07-07
[Frame Looping] Resubmit prerecorded command buffers on subsequent iterations of loop range

nickdriscoll-lunarg:
The PR changes frame looping such that command buffer commands are only recorded on the first iteration of the loop range, reusing the recorded command buffers on subsequent iterations.


framework/generated/generated_vulkan_replay_frame_loop_consumer_base.cpp
```diff
@@ -681,6 +681,21 @@ void VulkanReplayFrameLoopConsumerBase::Process_vkDestroyCommandPool(
...
+    const ApiCallInfo&                          call_info,
+    VkResult                                    returnValue,
+    format::HandleId                            device,
+    format::HandleId                            commandPool,
+    VkCommandPoolResetFlags                     flags)
+{
+    // Only record command buffer commands on first iteration of looping frame.
+    if (getFrameLoopInfo().IsRepetition())
+    {
+        return;
```
antonio-lunarg: Mmh, I am not sure we should skip `ResetCommandPool` and `ResetCommandBuffer`.
What if, within a frame, there's a: `Begin -> RecordA -> End -> Submit -> Reset -> Begin -> RecordB -> End -> Submit`?

framework/generated/khronos_generators/vulkan_generators/replay_frame_loop_overrides.json
```diff
@@ -185,5 +185,16 @@
...
     "vkReleaseSwapchainImagesEXT",
     "vkReleaseSwapchainImagesKHR"
-  ]
+  ],
+    
+  "comment" : "These are entry points for command buffer state commands",
+  "comment" : "that don't start with vkCmd",
+    "commandBufferState": [
+        "vkAllocateCommandBuffers",
+        "vkFreeCommandBuffers",
```
antonio-lunarg: Should `AllocateCOmmandBuffers` and `FreeCommandBuffers` be removed from `resourceAllocateNotFullyImplemented` and `resourceFreeNotFullyImplemented`?

framework/generated/khronos_generators/khronos_replay_frame_loop_consumer_base_header_generator.py
```diff
@@ -87,6 +89,9 @@ def write_class_contents(self):
             )
             write(cmddef, file=self.outFile)
 
+        # Write private data
```
antonio-lunarg: Dead comment?

framework/generated/generated_vulkan_replay_frame_loop_consumer_base.cpp
nickdriscoll-lunarg: I think I see what you mean. Maybe an approach could be to create a new command buffer whenever one, or its pool, is reset and then start recording to that. We'd have to keep track of which submission each command buffer clone is associated with, but I think that's doable.

antonio-lunarg: We already have something for that in `VulkanCommandBufferSplitter`. I believe that can be generalized to accomodate this case as well.

antonio-lunarg: https://github.com/LunarG/gfxreconstruct/pull/3083

antonio-lunarg: Merged!

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -67,6 +67,39 @@ void VulkanReplayFrameLoopConsumer::StartLooping()
...
+        // If we are looping, save the handle in allocatedLoopResources
+        if (getFrameLoopInfo().IsLooping())
+        {
+            allocatedLoopResources.insert(handle);
+        }
+
+        // Initialize this device's command buffer util
+        CommonObjectInfoTable& table = GetObjectInfoTable();
+        VulkanDeviceInfo* info = table.GetVkDeviceInfo(handle);
+        command_buffer_utils_[handle] = VulkanCommandBufferUtil(
```
antonio-lunarg: No need for this. The class inherits `GetDeviceCommandBufferUtil()` from `VulkanReplayConsumerBase`, so you just call that to get a ref when you need it.

nickdriscoll-lunarg: Oh ok. `GetDeviceCommandBufferUtil` is a private member of `VulkanReplayConsumerBase`, but since you're saying this I'll change it to protected.

framework/generated/generated_vulkan_replay_frame_loop_consumer_base.cpp
```diff
@@ -632,13 +632,182 @@ void VulkanReplayFrameLoopConsumerBase::Process_vkFreeCommandBuffers(
     const ApiCallInfo&                          call_info,
     args::FreeCommandBuffers&                   args)
 {
-    if (getFrameLoopInfo().IsLooping() && !getFrameLoopInfo().IsFinalIteration())
+    // Only record command buffer commands on first iteration of looping frame.
+    if (getFrameLoopInfo().IsRepetition())
```
antonio-lunarg: I am not sure this is correct. Command buffers get freed on first iteration, and that would be a problem for subsequent iterations, right?

tools/replay/replay_vulkan_feature.cpp
```diff
@@ -107,7 +107,7 @@ void ReplayVulkanFeature::RegisterDecodeComponents(graphics::FpsInfo* fps_info)
 {
     if (is_enabled_)
     {
-        if (replay_options_.isolate_render_passes)
+        if (replay_options_.isolate_render_passes || application_->GetFrameLoopInfo() != nullptr)
         {
             auto  decoder  = std::make_unique<decode::VulkanStateRecordingDecoder>();
```
antonio-lunarg, line 793: I noticed this is _always_ recording. So in the case of frame looping enabled, it is going to record all commands for no reason. This needs to be fixed, but it can be a different PR.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -114,7 +114,64 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateCommandPool(const ApiCallInf
...
+    if (frame_loop_info_.IsRepetition())
+    {
+        return;
+    }
+    if (frame_loop_info_.IsLooping() && !frame_loop_info_.IsRepetition())
+    {
+        VulkanCommandPoolInfo*   pool_info   = GetObjectInfoTable().GetVkCommandPoolInfo(args.commandPool);
+        VulkanDeviceInfo*        device_info = GetObjectInfoTable().GetVkDeviceInfo(pool_info->parent_id);
+        VulkanCommandBufferUtil& cbu         = GetDeviceCommandBufferUtil(device_info);
+        // cbu.FreeCommandBuffers(pool_info->handle)
```
antonio-lunarg: Reset should already be handled in `OverrideResetCommandPool`. Maybe we should check the condition for when it should run.

antonio-lunarg (commented):
The direction is good, but I think CommandBufferUtil needs more "generalization".
So far, it's been used to record some commands, split the command buffers, and restore the original handle on "reset". These capabilities should not be used/enabled here, in other words, "begin" should just work on the current command buffer handle (which rotates on reset), and reset should not restore the original command buffer handle, but should rotate..

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -830,6 +794,18 @@ void VulkanReplayFrameLoopConsumer::Process_vkBeginCommandBuffer(const ApiCallIn
...
+void VulkanReplayFrameLoopConsumer::Process_vkFreeCommandBuffers(
+    const ApiCallInfo&                          call_info,
+    args::FreeCommandBuffers&                   args)
+{
+    // Don't free any command buffers while inside the loop range
+    if (frame_loop_info_.IsLooping())
+    {
+        return;
+    }
+    VulkanReplayConsumer::Process_vkFreeCommandBuffers(call_info, args);
```
antonio-lunarg, line 806: It should be OK to free command buffers in the last iteration of the loop.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -805,12 +761,20 @@ void VulkanReplayFrameLoopConsumer::Process_vkDestroyDescriptorPool(const ApiCal
 void VulkanReplayFrameLoopConsumer::Process_vkBeginCommandBuffer(const ApiCallInfo&        call_info,
                                                                  args::BeginCommandBuffer& args)
 {
-    VulkanReplayConsumer::Process_vkBeginCommandBuffer(call_info, args);
-
-    if (frame_loop_info_.IsLooping())
+    if (frame_loop_info_.IsLooping() && !frame_loop_info_.IsRepetition())
```
antonio-lunarg, line 764: We should call `VulkanReplayConsumer::Process_vkBeginCommandBuffer` if not looping.

PR #3080, opened 2026-07-07
Add linux nvidia 5080 & update test_suite.ref

charles-lunarg:


mikes-lunarg (approved):
Looks pretty straightforward. Something is a little weird with the line endings in ci/Jenkinsfile but that is pre-existing. @beau-lunarg Maybe worth normalizing the line endings in ci/ at some point

PR #3081, opened 2026-07-07
Refactor async access such that cross thread access is isolated

jzulauf-lunarg:
Previously the async processing support was (generally) correct by implementation.  This change makes it correct much more by construction.
@antonio-lunarg  -- should address your concerns re: correctness by construction
@davidd-lunarg  -- tagged you because I eliminating the block skipping file processor sub class, encapsulating the functionality in a way that could have clear thread ownership.


davidd-lunarg:
I found a DX12 w/ DXR capture that fails to optimize with f64a433c0f but succeeds with 496722ba83e. I'll send repro steps to you directly.

`[gfxrecon] ERROR - Could not detect graphics API. Aborting optimization.`

framework/decode/file_processor_types.h
```diff
@@ -55,26 +61,66 @@ enum class ProcessBlockState : int32_t
...
 using FrameNumber = uint64_t;
 struct ProcessBlocksResult
 {
-    // NOTE: This is the frame_number of the *next* frame
-    // Snapshot of process_frame_number_ at return.
+    // Always meaningful.
+    ProcessBlockState state{ ProcessBlockState::kContinue };
+    BlockIOError      error{ BlockIOError::kErrorNone };
+
+    // Process-side snapshot fields; only valid when has_process_state is true.
```
antonio-lunarg: You can use `std::optional<Snapshot>` for this.

framework/decode/file_processor.h
```diff
@@ -59,321 +57,147 @@ class AsyncProcessPolicy;
...
-    static bool ContinueProcessing(ProcessBlockState state) { return static_cast<int32_t>(state) >= 0; }
-    static bool IsFrameBoundary(ProcessBlockState state) { return state == ProcessBlockState::kFrameBoundary; }
-
-    ProcessBlockState HandleBlockEof(const char* operation, bool report_frame_and_block);
-
-    // Async processing needs to be to queury this frame wise.
-    friend class AsyncProcessor;
-    FrameNumber  GetProcessFrameNumber() const noexcept { return process_frame_number_; }
-    BlockIOError GetProcessErrorState() const noexcept { return process_error_state_; }
+    // Async processing needs to query this frame-wise.
```
antonio-lunarg: Async processing now only uses a BlockProcessor, right? So this comment is misleading?

framework/decode/file_processor.h
```diff
@@ -59,321 +57,147 @@ class AsyncProcessPolicy;
...
-    bool EntireFileWasProcessed() const
-    {
-        if (file_stack_.empty())
-        {
-            return true;
-        }
-
-        return file_stack_.front().active_file->IsEof();
-    }
+    bool EntireFileWasProcessed() const { return dispatch_terminal_state_ == ProcessBlockState::kEndOfFile; }
```
antonio-lunarg, line 133: This should be unit-testable. I would introduce a simple test, call `ProcessAllFrames()` and assert `EntireFileWasProcessed() == true` for both sync and async mode.

framework/decode/block_processor.h
```diff
@@ -0,0 +1,238 @@
...
+    bool SeekActiveFile(const FileInputStreamPtr& file, int64_t offset, util::platform::FileSeekOrigin origin);
+    bool SeekActiveFile(int64_t offset, util::platform::FileSeekOrigin origin);
+    bool SetActiveFile(const std::string& filename, bool execute_till_eof);
+    bool SetActiveFile(const std::string&             filename,
+                       int64_t                        offset,
+                       util::platform::FileSeekOrigin origin,
+                       bool                           execute_till_eof);
+
+    bool               AtEof() const;
+    BlockIOError       CheckFileStatus() const;
```
antonio-lunarg: This has no callers?

framework/decode/block_processor.h
```diff
@@ -0,0 +1,238 @@
...
+GFXRECON_END_NAMESPACE(file_processor)
+
+// Owns all process-side (I/O and parse thread) state for FileProcessor.
+// In async mode, ownership transfers to AsyncProcessor at InitializeFrameProcessing time;
+// FileProcessor's block_processor_ becomes null — any residual block_processor_ access from the dispatch thread
+// will null-deref, making the bug immediately detectable.
+//
+// block_limit is NOT stored here. It lives on FileProcessor as an immutable protected
+// member and is propagated into each policy object as a scalar before ProcessBlocks is called.
+class BlockProcessor
```
antonio-lunarg, line 63: Is this more a "Producer" rather than a "Processor"? I see that this class is producing blocks but it does not do anything with them rather than delegating the processing step to something else.
I think, in general, calling it "process-side" might be a bit misleading since that side is not processing IMO, but it's actually reading->producing.

jzulauf-lunarg: Depending on policy, it could produce or produce and consume...

framework/decode/block_processor.h
jzulauf-lunarg: removed, next push

antonio-lunarg (approved):

PR #3083, opened 2026-07-08
util: Replace command buffer

antonio-lunarg:
Generalize VulkanCommandSplitter into VulkanCommandBufferUtil.

Rename split info to associated info and extract
ReplaceWithAssociatedCommandBuffer from SplitCommandBuffer, so a command buffer handle can be swapped without ending/beginning a split. Prepares for frame-loop handling of command buffers reset during the loop frame.

### DIFF
--- a/framework/decode/vulkan_command_buffer_util.cpp
+++ b/framework/decode/vulkan_command_buffer_util.cpp
@@ -19,7 +19,7 @@
 ** FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 ** DEALINGS IN THE SOFTWARE.
 */
-#include "decode/vulkan_command_splitter.h"
+#include "decode/vulkan_command_buffer_util.h"
 #include "decode/vulkan_submit_info_helper.h"
 #include "generated/generated_vulkan_enum_to_string.h"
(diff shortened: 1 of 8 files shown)


framework/decode/vulkan_command_buffer_util.cpp
```diff
@@ -75,15 +75,20 @@ VkCommandBuffer VulkanCommandBufferSplitInfo::GetNextHandle(const VulkanCommandB
     next_handle = associated_handles_[next_associated_index_];
     ++next_associated_index_;
 
+    command_buffer_info->handle = next_handle;
+}
+
+void VulkanCommandBufferAssociatedInfo::SplitCommandBuffer(VulkanCommandBufferInfo* command_buffer_info)
+{
     // Store current handle in split vector.
     split_handles_.push_back(command_buffer_info->handle);
```
fabian-lunarg: this is the only place populating `split_handles_` and it has no callers,
like something was overlooked?

antonio-lunarg: Yes, I forgot to save handles after a split. I introduced a `PushSplitHandle` function for that.

fabian-lunarg (approved):

PR #3084, opened 2026-07-08
fix junk fpsinfo output; use sized integer print macros

bradgrantham-lunarg:
integer types in Varargs didn't match the printf-style format types, so different sizes were consumed, causing subsequent specifiers to consume the wrong values.  Fix the printf types to match the passed parameters on 64-bit and 32-bit.


MarkY-LunarG (approved):
LGTM

mikes-lunarg (approved):
Nice
```
INFO: looking for FPS and warning lines in 2071 bytes of stdout and 0 bytes of stderr...
INFO: Measured FPS: 27.204958 fps, 0.367580 seconds, 10 frames, 1 loop, framerange [1-11)
```

PR #3085, opened 2026-07-08
Fix usage exe name

MarkY-LunarG:
My last commit was over-zealous and added a prefix in front of an exe name.  The usage exe name should have been the trimmed exe name, not the full app name.

This fixes that.

### DIFF
--- a/tools/compress/main.cpp
+++ b/tools/compress/main.cpp
@@ -47,11 +47,12 @@ const char kArgUnknown[] = "<Unknown>";
 
 static void PrintUsage(const char* exe_name)
 {
-    std::string app_name = GetApplicationName(exe_name);
+    std::string trimmed_exe_name = GetTrimmedExeName(exe_name);
+    std::string app_name         = GetApplicationName(exe_name);
     GFXRECON_WRITE_CONSOLE("\n%s - A tool to compress/decompress GFXReconstruct capture files.\n", app_name.c_str());
     GFXRECON_WRITE_CONSOLE("Usage:");
     GFXRECON_WRITE_CONSOLE("  %s [-h | --help] [--version] <input_file> <output_file> <compression_format>\n",
-                           app_name.c_str());
+                           trimmed_exe_name.c_str());
     GFXRECON_WRITE_CONSOLE("Required arguments:");
     GFXRECON_WRITE_CONSOLE("  <input_file>\t\tPath to the input file to process.");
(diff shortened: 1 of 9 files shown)


tools/tool_command_line.h
```diff
@@ -45,19 +45,25 @@ const char kHelpShortOption[] = "-h";
...
+    size_t      dir_location = actual_exe.find_last_of("/\\");
 
     if (dir_location != std::string::npos)
     {
-        app_name.replace(0, dir_location + 1, "");
+        actual_exe.replace(0, dir_location + 1, "");
     }
 
-    return GFXRECON_APP_NAME_PREFIX + app_name;
+    return actual_exe;
```
mikes-lunarg, line 59: Elsewhere in the codebase we use std;:filesystem for stuff like this:
```
std::filesystem::path(options.capture_filename).stem().string()
```
For this case you probable want `std::filesystem::path::filename` to preserve the .exe extension on Windows.

Does it still warrant a utility function if we can replace it with a standard library call?

tools/compress/main.cpp
```diff
@@ -47,11 +47,12 @@ const char kArgUnknown[] = "<Unknown>";
 
 static void PrintUsage(const char* exe_name)
 {
-    std::string app_name = GetApplicationName(exe_name);
+    std::string trimmed_exe_name = GetTrimmedExeName(exe_name);
```
mikes-lunarg, line 50: I think "trimmed" in `trimmed_exe_name` and `GetTrimmedExeName` is ambiguous (what is trimmed? whitespace? extensions? path?) I would prefer to just call it exe_name/ExecutableName to match app_name/ApplicationName

tools/tool_command_line.h
```diff
@@ -45,19 +45,25 @@ const char kHelpShortOption[] = "-h";
...
+}
+
+/// @return exe_name with any leading directory components removed, and the
+/// GFXRECON_APP_NAME_PREFIX configured at build time (see project_version.h.in) applied.
+inline std::string GetApplicationName(const char* exe_name)
+{
+    return GFXRECON_APP_NAME_PREFIX + GetTrimmedExeName(exe_name);
```
mikes-lunarg, line 66: Why settle for a prefix? You could 
```
#ifndef GFXRECON_APP_NAME
#define GFXRECON_APP_NAME "Replay"
#endif
```

Then whatever needs to override the name would have full control over the displayed name, e.g. `-DGFXRECON_APP_NAME="Replay (Taylor's Version)"`

PR #3086, opened 2026-07-08
Add screenshot enable query to capture manager

MarkY-LunarG:
Simply adds a query to determine if the screenshot is enabled.  This allows us to restrict code that is only used if screenshotting is even enbaled.


MarkY-LunarG:
Adding @jzulauf-lunarg since he is now Domain Lead in "encode".

ncesario-lunarg (approved):

jzulauf-lunarg (approved):
👍   LGTM

PR #3089, opened 2026-07-09
add CLAUDE.md

bradgrantham-lunarg:


CLAUDE.md
```diff
@@ -0,0 +1,127 @@
...
+## Build Commands
+
+### Quick build (Windows)
+```bash
+python scripts/build.py --skip-check-code-style --skip-tests
+```
+
+### Parallel build
+```bash
+CMAKE_BUILD_PARALLEL_LEVEL=8 python scripts/build.py --skip-check-code-style --skip-tests
```
panos-lunarg: Perhaps we can skip the `build.py` specifics (which can change) and just mention cmake?

panos-lunarg (approved):
lgtm!

per-mathisen-arm:
I've been reading some of the recent research papers on agents files, and the short summary is that since every character in this file will consume context window from every agent session, it is important to keep it short. Anything that the agent could easily figure out on by searching or that it would normally infer on its own is better omitted and will overall degrade its effectiveness. Agent-generated agents files have been found to be pretty bad compared to manually written files compiled from agent mistakes.

Also, you could name it `AGENTS.md` which works with every other agent, then add a `CLAUDE.md` that contains only `@AGENTS.md`.

bradgrantham-lunarg:
> I've been reading some of the recent research papers on agents files, and the short summary is that since every character in this file will consume context window from every agent session, it is important to keep it short. Anything that the agent could easily figure out on by searching or that it would normally infer on its own is better omitted and will overall degrade its effectiveness. Agent-generated agents files have been found to be pretty bad compared to manually written files compiled from agent mistakes.

Caching should make the agents file fairly low cost unless it's excessive.

Still, point taken, and I pruned down a lot; took out things a recent model will already have baked in and anything redundant after tokenization (e.g. the comment in `python3 generate_vulkan.py # Generate Vulkan` is pointless).

Done!

> Also, you could name it `AGENTS.md` which works with every other agent, then add a `CLAUDE.md` that contains only `@AGENTS.md`.

Done!

AGENTS.md
```diff
@@ -0,0 +1,90 @@
...
+- **Backward compatibility**: Existing capture files stay readable (deprecate block IDs, never repurpose them), frozen Consumer/base-class and `framework/util` interfaces keep their signatures, existing environment variables and command-line options are never deleted, default tool behavior does not change, and convert's JSON key names/value types are not changed.
+
+If a change must conflict with a principle, explain the reason in the commit message and PR description.
+
+## Build
+
+Standard CMake. Common options:
+- `-DD3D12_SUPPORT=OFF`
+- `-DGFXRECON_ENABLE_OPENXR=OFF`
+- `-DGFXRECON_INCLUDE_TEST_APPS=ON`
```
davidd-lunarg, line 19: Shouldn't the DX12 and OpenXR flags be ON to ensure their builds don't get broken.

Flags I use and recommend:
- `RUN_TESTS=ON` - enables unit tests
- `GFXRECON_ENABLE_RELEASE_ASSERTS=ON` - logs when an assert fails in release mode, otherwise release build asserts are silent

AGENTS.md
```diff
@@ -0,0 +1,90 @@
...
+**Never hand-edit** files in `framework/generated/`. Instead, edit the Python generator scripts and regenerate:
+
+```bash
+cd framework/generated
+python3 generate_vulkan.py
+python3 generate_dx12.py
+python3 generate_openxr.py
+```
+
+Generator hierarchy: `base_generators/` -> `khronos_generators/` -> API-specific (`khronos_generators/{vulkan,openxr}_generators/`, `dx12_generators/`).
```
davidd-lunarg, line 35: I would guess agents will discover this if they ever have to edit the generator files. Having it here fills context for all agents regardless of whether they're modifying generated code.

davidd-lunarg (commented):
My preference is that we do not add an AGENTS/CLAUDE.md file to GFXR.

PR #3091, opened 2026-07-09
Respect swapchain pre-rotation for screenshots

olehkuznetsov:
Added CPU-side rotation logic in ScreenshotHandler::WriteImage to orient screenshots based on the application's swapchain preTransform setting (e.g. ROTATE_90_BIT, ROTATE_270_BIT).

This feature is optional and can be enabled by passing the new command-line flag: --screenshot-apply-prerotation.

Changes:

- Added pre_transform field to VulkanSwapchainKHRInfo to track the orientation.

- Implemented pixel buffer rotation in ScreenshotHandler before BMP/PNG encoding.

- Plumbed --screenshot-apply-prerotation through the replay options.


MarkY-LunarG (dismissed):

framework/decode/screenshot_handler.cpp
```diff
@@ -111,6 +111,29 @@ void ScreenshotHandler::WriteImage(const std::string&                         fi
         flip_y      = copy_scale.value()[1] < 0.f;
     }
 
+    switch (pre_transform)
```
jzulauf-lunarg: Some explanation of the mapping logic would be useful for maintainers.  Some of these mappings do not appear to match the specification.

For example 
```
        case VK_SURFACE_TRANSFORM_ROTATE_90_BIT_KHR:
        case VK_SURFACE_TRANSFORM_HORIZONTAL_MIRROR_ROTATE_90_BIT_KHR:
            pre_transform = VK_SURFACE_TRANSFORM_ROTATE_90_BIT_KHR;
```

are treated the same but the descriptions vary.

"VK_SURFACE_TRANSFORM_ROTATE_90_BIT_KHR specifies that image content is rotated 90
degrees clockwise."

"VK_SURFACE_TRANSFORM_HORIZONTAL_MIRROR_ROTATE_90_BIT_KHR specifies that image
content is mirrored horizontally, then rotated 90 degrees clockwise."

jzulauf-lunarg (changes requested):
Please double check the transform mapping logic, and if non-obviously correct, document the no-obvious mutating mappings.

tools/tool_settings.h
```diff
@@ -1300,6 +1301,11 @@ GetVulkanReplayOptions(const gfxrecon::util::ArgumentParser&           arg_parse
         replay_options.screenshot_ignore_frameBoundaryAndroid = true;
     }
 
+    if (arg_parser.IsOptionSet(kScreenshotApplyPrerotationArgument))
+    {
+        replay_options.screenshot_apply_prerotation = true;
```
antonio-lunarg: This can be oneline: `replay_options.screenshot_apply_prerotation = arg_parser.IsOptionSet(kScreenshotApplyPrerotationArgument)`

jzulauf-lunarg: `replay_options.screenshot_apply_prerotation |= arg_parser.IsOptionSet(kScreenshotApplyPrerotationArgument)` would be the equivalent of the logic.

Though note the surrounding style... it would be nice if they all collapsed to `|=` assigns, but I'm not sure I favor a mixed style in context.

framework/decode/screenshot_handler.cpp
olehkuznetsov: I've updated this part to make it more clear. Now mirror logic is handled the same way as rotation. And I added more comments to describe some parts.

tools/tool_settings.h
olehkuznetsov: Seems as we already have some one liners below. I moved the option close to one liners to make code style more uniform

jzulauf-lunarg (approved):
LGTM -- is there any test coverage?

olehkuznetsov:
> Please double check the transform mapping logic, and if non-obviously correct, document the no-obvious mutating mappings.

Done

olehkuznetsov:
> LGTM -- is there any test coverage?

I added tests for image rotation utils

framework/decode/screenshot_handler.cpp
```diff
@@ -424,7 +426,78 @@ void ScreenshotHandler::WriteImage(const std::string&                         fi
...
+                        }
+                        else if (pre_transform == VK_SURFACE_TRANSFORM_ROTATE_270_BIT_KHR ||
+                                 pre_transform == VK_SURFACE_TRANSFORM_HORIZONTAL_MIRROR_ROTATE_270_BIT_KHR)
+                        {
+                            rotation = util::imagewriter::ImageRotation::DEG_270;
+                        }
+
+                        bool is_mirrored =
+                            (pre_transform == VK_SURFACE_TRANSFORM_HORIZONTAL_MIRROR_BIT_KHR ||
+                             pre_transform == VK_SURFACE_TRANSFORM_HORIZONTAL_MIRROR_ROTATE_90_BIT_KHR ||
```
antonio-lunarg, line 454: According to Vulkan spec, HORIZONTAL_MIRROR_ROTATE_90/270 specifies that image content is first mirrored, then rotated. So the inverse transform should be rotate-then-mirror, but as far as I inderstand `RotateAndMirrorPixelsInternal` only mirrors after rotation?

olehkuznetsov: Great catch. Fixed.

antonio-lunarg (approved):

bradgrantham-lunarg (approved):

PR #3092, opened 2026-07-09
Add listing modules to Tool output (where appropraite) and perform some cleanup 

MarkY-LunarG:
(Split into 3 commits)
- Add  RAII logging struct for auto cleanup (and unify on return values)
- Remove APP_PREFIX_NAME which really didn't work as intended
- Add --list-modules command-line where appropriate.

### DIFF
--- a/project_version.h.in
+++ b/project_version.h.in
@@ -47,11 +47,4 @@ const char* GetProjectVersionString();
 #define GFXRECON_PROJECT_OPENXR_LAYER_NAME  "XR_APILAYER_LUNARG_gfxreconstruct"
 #define GFXRECON_PROJECT_DESCRIPTION        "GFXReconstruct Capture Layer"
 
-// Optional prefix applied to the display name of the command-line tools (usage banner and
-// --version output). Downstream distributions that repackage these tools under a different
-// product name can set this at configure/compile time. Empty by default.
-#ifndef GFXRECON_APP_NAME_PREFIX
-#define GFXRECON_APP_NAME_PREFIX ""
-#endif
-
 #endif // GFXRECON_PROJECT_VERSION_H
--- a/tools/compress/main.cpp
+++ b/tools/compress/main.cpp
@@ -32,8 +32,10 @@
 #include "util/compressor.h"
 #include "util/logging.h"
 
+#include <algorithm>
 #include <cassert>
 #include <cstdlib>
+#include <filesystem>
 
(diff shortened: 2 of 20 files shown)


MarkY-LunarG:
@bradgrantham-lunarg, this is based on your design change suggestion.  Does this match what you desired?

mikes-lunarg:
Conceptually, I think this overlaps with --version, since that functionally already enumerates the features and lists their versions:
```
$ gfxrecon-info --version
gfxrecon-info version info:
  GFXReconstruct Version 1.0.5-dev (dev:479695d*) Debug
  OpenXR Header Version  1.1.40
  Vulkan Header Version  1.4.356
```

Do we really need a new flag?

MarkY-LunarG:
> Conceptually, I think this overlaps with --version, since that functionally already enumerates the features and lists their versions:
> 
> ```
> $ gfxrecon-info --version
> gfxrecon-info version info:
>   GFXReconstruct Version 1.0.5-dev (dev:479695d*) Debug
>   OpenXR Header Version  1.1.40
>   Vulkan Header Version  1.4.356
> ```
> 
> Do we really need a new flag?

It doesn't print everything.  For example,  D3d doesn't have a header output so doesn't output during --version.  If we can come up with something to print in those cases, I'm fine.  but we might have to hard-code that.

MarkY-LunarG:
@mikes-lunarg / @bradgrantham-lunarg , we could output something like:

```
gfxrecon-info version info:
  GFXReconstruct Version 1.0.5-dev (dev:479695d*) Debug
  OpenXR Header Version  1.1.40
  Vulkan Header Version  1.4.356
  D3D12 No Header info
```

That would fulfill the need.

mikes-lunarg:
> @mikes-lunarg / @bradgrantham-lunarg , we could output something like:
> 
> ```
> gfxrecon-info version info:
>   GFXReconstruct Version 1.0.5-dev (dev:479695d*) Debug
>   OpenXR Header Version  1.1.40
>   Vulkan Header Version  1.4.356
>   D3D12 No Header info
> ```
> 
> That would fulfill the need.

I was looking at this a bit more and I think the D3D12 module should list the Windows SDK version it was generated from (10.0.26100.0). That would match the usage of the header versions on the Khronos side. I'm happy to go wire up the DX codegen parts if you want to hardcode or use a placeholder for now.

MarkY-LunarG:
Closing to be replaced by #3103

PR #3095, opened 2026-07-09
Disable options file reading if Vulkan is not enabled

MarkY-LunarG:


framework/encode/capture_settings.cpp
```diff
@@ -366,6 +366,7 @@ void CaptureSettings::LoadOptionsEnvVar(OptionsMap* options, bool load_log_setti
 
 void CaptureSettings::LoadOptionsFile(OptionsMap* options)
 {
+#if defined(GFXRECON_ENABLE_VULKAN)
```
jzulauf-lunarg, line 369: Yep.  this is a vulkan specific file.

jzulauf-lunarg (approved):

bradgrantham-lunarg:
The current change would be a broader behavior change than the PR description suggests, because it would silently remove settings-file support for D3D12 as well as Vulkan.

davidd-lunarg:
For D3D12 I have not personally used the settings file. It is also not mentioned in the D3D12 usage doc: https://github.com/LunarG/gfxreconstruct/blob/dev/USAGE_desktop_D3D12.md#capture-options

I think there was an issue once where a GFXR layer settings file was getting picked up (maybe from the Vulkan SDK install?) and overriding default D3D12 capture options causing issues--D3D12 default is `GFXRECON_PAGE_GUARD_EXTERNAL_MEMORY=true`, Vulkan defaults to `false`. This resulted in a significant capture time slowdown.

> it would silently remove settings-file support for D3D12 as well as Vulkan

I agree this silently changes DX12 behavior depending on whether Vulkan is included in the build.

I am OK with removing the settings file for non-Vulkan builds, but I also think it should not be applied during D3D12 capture even if Vulkan is included in the build. However I don't know if that is currently possible due to the API-agnostic way capture is setup.

bradgrantham-lunarg (approved):

PR #3096, opened 2026-07-09
Move reading the "remove_unsupported_features"

MarkY-LunarG:
Move the reading of the "remove_unsupported_features" option from the Vulkan-specific section to the base options section. This option is not Vulkan-specific and in the base options struct.

Setting is defined in framework/decode/replay_options.h line 72 in the base "ReplayOptions" struct, so it should be in read in the generic Options code.


mikes-lunarg (approved):
LGTM

bradgrantham-lunarg (approved):

PR #3097, opened 2026-07-10
Add a CI workflow job that runs clang tidy

bradgrantham-lunarg:
Add a job that runs clang-tidy on C++ code _except generated code_.  The job currently runs at 30 minutes without generated code, nearly an hour with generated.  Another possibility is to run `clang-tidy-cache` which would generally reduce runtime somewhat.

### DIFF
--- a/framework/decode/async_processor.h
+++ b/framework/decode/async_processor.h
@@ -72,7 +72,7 @@ class AsyncProcessor
         explicit AsyncBatchIterator(AsyncProcessor& async_processor, AsyncBatchQueue& queue) :
             async_processor_(&async_processor), queue_(&queue)
         {
-            Advance();
+            AsyncBatchIterator::Advance();
         }
         void Advance() override;
 
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -314,7 +314,7 @@ VulkanReplayConsumerBase::~VulkanReplayConsumerBase()
     }
 
     // Idle all devices before destroying other resources.
-    WaitDevicesIdle();
+    VulkanReplayConsumerBase::WaitDevicesIdle();
 
     // free replacer internal vulkan-resources
     device_address_replacers_.clear();
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -319,7 +319,7 @@ VulkanReplayConsumerBase::~VulkanReplayConsumerBase()
     }
 
     // Idle all devices before destroying other resources.
-    WaitDevicesIdle();
+    VulkanReplayConsumerBase::WaitDevicesIdle();
 
     // free replacer internal vulkan-resources
     device_address_replacers_.clear();
--- a/tools/replay/replay_feature.h
+++ b/tools/replay/replay_feature.h
@@ -47,7 +47,7 @@ GFXRECON_BEGIN_NAMESPACE(replay)
 class ReplayFeatureBase : public util::FeatureBase
 {
   public:
-    virtual ~ReplayFeatureBase() { Destroy(); }
+    virtual ~ReplayFeatureBase() = default;
 
     // Options queries
     virtual void QueryOptions(util::ArgumentParser& arg_parser, const std::string& capture_filename) {}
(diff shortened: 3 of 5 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -314,7 +314,7 @@ VulkanReplayConsumerBase::~VulkanReplayConsumerBase()
     }
 
     // Idle all devices before destroying other resources.
-    WaitDevicesIdle();
+    VulkanReplayConsumerBase::WaitDevicesIdle();
```
bradgrantham-lunarg, line 322: @panos-lunarg  WaitDevicesIdle is virtual but virtual resolution is bypassed inside constructors and destructors so these end up calling VulkanReplayConsumerBase::WaitDevicesIdle.  @wzchu-lunarg's fix is just to explicitly resolve the base function which (squinting) seems to match what this destructor may have meant to do.  What was your intent here?

framework/decode/async_processor.h
```diff
@@ -74,7 +74,7 @@ class AsyncProcessor
         explicit AsyncBatchIterator(AsyncProcessor& async_processor, AsyncBatchQueue& queue) :
             async_processor_(&async_processor), queue_(&queue)
         {
-            Advance();
+            AsyncBatchIterator::Advance();
```
bradgrantham-lunarg, line 75: @jzulauf-lunarg Advance here is virtual but virtual resolution is bypassed inside constructors and destructors so this ends up calling AsyncBatchIterator::Advance.  clang-tidy calls this out.  @wzchu-lunarg's fix is just to explicitly resolve the base function which (squinting) seems to match what this constructor may have meant to do.  What was your intent here?

framework/decode/vulkan_replay_consumer_base.cpp
bradgrantham-lunarg: (Sorry, incorrectly @'d John in this comment initially.

framework/decode/async_processor.h
jzulauf-lunarg: The actual lint complaint is more subtle.  It's warning that if an override happened in a derived class that the derived override wouldn't be called in the base class constructor and the reader might assume that it was.

https://godbolt.org/z/Evhrb1h8c  shows this.  The f() in the C constructor is B::f() (though one could even specify it C::f() even though it doesn't exist), and the f() in the D constructor is the D::f... so one can understand assertion it is untidy.

In context Advance(): and  AsyncBatchIterator::Advance(); are defined to do exactly the same thing... except as to how the linter sees it.

So yes, I mean it to call AsyncBatchIterator::Advance, especially since the class virtual void Advance=0;

framework/decode/vulkan_replay_consumer_base.cpp
panos-lunarg: I see in that old commit where I added this line, I also implemented `VulkanReplayConsumerBase::WaitDevicesIdle()`, so I also believe William's fix align with my what I was trying to do back then

bradgrantham-lunarg:
@jzulauf-lunarg I rebased, should be same changes as before rebase.  I chose you arbitrarily; please review and approve.

bradgrantham-lunarg:
@fabian-lunarg what do you think about this?

tools/replay/replay_feature.h
```diff
@@ -47,7 +47,7 @@ GFXRECON_BEGIN_NAMESPACE(replay)
 class ReplayFeatureBase : public util::FeatureBase
 {
   public:
-    virtual ~ReplayFeatureBase() { Destroy(); }
+    virtual ~ReplayFeatureBase() = default;
```
davidd-lunarg, line 50: If I understand correctly, the destructor was calling `ReplayFeatureBase::Destroy()` which is a no-op, and the derived class's overrides were never called. So this doesn't change behavior but does this reveal that there is an existing bug for derived class destruction which expected their `Destroy()` overrides to be called? If so, is there an issue or fix logged for it?

wzchu-lunarg: I hid the fix logged in the commit https://github.com/LunarG/gfxreconstruct/pull/3097/changes/67fe84191ec844d119300213b52d64fbc787bf5. Should probably log it somewhere better. However, there didn't seem to be an issue as it will be called in `replay_main_common.cpp`'s `feature->Destroy();` anyways.

davidd-lunarg: I missed that `Destroy` was called in `replay_main_common.cpp` so this destructor change LGTM. Thanks

davidd-lunarg (approved):

jzulauf-lunarg (approved):
The one I know about looks correct.

PR #3098, opened 2026-07-10
Descriptor utils consecutive binding updates

wzchu-lunarg:
Issue occurs in dump resources where an application may update a binding with `descriptorCount` greater than the remaining elements in the descriptor array, which should overflow to consecutive bindings. It seemed dump resources is not able to track this overflow and loses them completely.

Instead of only iterating through the descriptor array, also check if it overflows and go to the next binding.

There might also need to be changes to `vkUpdateDescriptorSetWithTemplate`.


framework/decode/vulkan_descriptor_utils.cpp
```diff
@@ -68,9 +68,37 @@ void HandleDescriptorUpdate(CommonObjectInfoTable&
             continue;
         }
 
+        uint32_t binding_index = write->dstBinding;
```
panos-lunarg: `write->dstBinding` is used in line 47. Can you move this before this line and replace `write->dstBinding` there with the new variable `binding_index`

framework/decode/vulkan_descriptor_utils.cpp
```diff
@@ -68,9 +68,37 @@ void HandleDescriptorUpdate(CommonObjectInfoTable&
...
+        uint32_t arr_idx       = write->dstArrayElement;
+        auto     binding_it    = descriptor_bindings_info.find(binding_index);
```
panos-lunarg: The `descriptor_bindings_info` map is searched again in line  47 (but with the operator `[]`). Can you unify the two accesses? What happens if find returns `end()`? Reading the spec I understand that it invalid usage. Perhaps we can skip that `write`.

wzchu-lunarg: We can unify the accesses. However, even if it is invalid usage, it seems to me that find should never return `end()`. 
Since we use operator `[]` in 47, it creates the binding info if it is not there.

Unifying it and explicitly creating it would definitely be more obvious.

framework/decode/vulkan_descriptor_utils.cpp
```diff
@@ -68,9 +70,33 @@ void HandleDescriptorUpdate(CommonObjectInfoTable&
             continue;
         }
 
+        uint32_t arr_idx = write->dstArrayElement;
+
         for (uint32_t i = 0; i < write_meta.decoded_value->descriptorCount; ++i)
         {
-            const uint32_t arr_idx = write->dstArrayElement + i;
+            if (binding_it != descriptor_bindings_info.end() && binding_it->second.count != 0 &&
```
panos-lunarg: I guess with `try_emplace`, `binding_it` cannot be `descriptor_bindings_info.end()` so this check can be removed

framework/decode/vulkan_descriptor_utils.cpp
```diff
@@ -68,9 +70,32 @@ void HandleDescriptorUpdate(CommonObjectInfoTable&
...
+                } while (binding_it != descriptor_bindings_info.end() && binding_it->second.count == 0);
+
+                if (binding_it == descriptor_bindings_info.end())
+                {
+                    break;
+                }
+                arr_idx = 0;
+            }
+
+            auto& descriptor_set_binding_info = binding_it->second;
```
fabian-lunarg: name is shadowing another `descriptor_set_binding_info` from above.
should be renamed to e.g. `current_binding_info`

fabian-lunarg (approved):
looks good to me, PR also adds a capture-based CI test.
one minor comment is open, otherwise approved when @panos-lunarg is happy too

panos-lunarg (approved):
lgtm too. For me only remaining thing is to also merge the test case

PR #3099, opened 2026-07-11
Check memory before write

locke-lunarg:
Memory could have been freed before the buffer was destroyed, so only write the buffer device address if the memory still exists.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -10133,14 +10133,35 @@ VkDeviceAddress VulkanReplayConsumerBase::OverrideGetBufferDeviceAddress(
     // retrieve replay-time device-address
     VkDeviceAddress replay_device_address = func(device, address_info);
 
-    // if supported, opaque device-addresses are expected to match
-    GFXRECON_ASSERT(!device_info->allocator->SupportsOpaqueDeviceAddresses() ||
-                    original_result == replay_device_address)
-
-    // keep track of old/new addresses in any case
     format::HandleId  buffer      = pInfo->GetMetaStructPointer()->buffer;
     VulkanBufferInfo* buffer_info = GetObjectInfoTable().GetVkBufferInfo(buffer);
     GFXRECON_ASSERT(buffer_info != nullptr);
+
+    // if supported, opaque device-addresses are expected to match.
+    //
+    // According to VUID-vkGetBufferDeviceAddress-bufferDeviceAddress-03324,
+    // The buffer has to
+    // 1. be a sparse, or
+    // 2. bind a valid memory, or
+    // 3. be created with VK_BUFFER_CREATE_DEVICE_ADDRESS_CAPTURE_REPLAY_BIT.
+    //
+    // Although it's legal to GetBufferDeviceAddress,
+    // the replay address could be 0 and mismatch if the memory is invalid, so print warning, instead of ASSERT.
+    //
+    // For trimming case, the buffer could still exist, but the memory is freed when the trimming starts.
+    if (device_info->allocator->SupportsOpaqueDeviceAddresses() && original_result != replay_device_address)
+    {
+        GFXRECON_LOG_WARNING(
+            "Opaque device addresses are supported, so captured and replay buffer device addresses are "
+            "expected to match. Buffer (ID = %" PRIu64 ") was captured with address 0x%" PRIx64
+            ", but replay returned 0x%" PRIx64 ". This can happen when the buffer's memory was freed "
+            "before the trim range started.",
+            buffer_info->capture_id,
+            original_result,
+            replay_device_address);
+    }
+
+    // keep track of old/new addresses in any case
     buffer_info->capture_address = original_result;
     buffer_info->replay_address  = replay_device_address;
 
--- a/framework/encode/vulkan_state_writer.cpp
+++ b/framework/encode/vulkan_state_writer.cpp
@@ -1535,18 +1535,37 @@ void VulkanStateWriter::WriteBufferDeviceAddressState(const VulkanStateTable& st
         GFXRECON_ASSERT(wrapper != nullptr && wrapper->device != VK_NULL_HANDLE);
         if (wrapper->device != VK_NULL_HANDLE && wrapper->address != 0)
         {
-            auto physical_device_wrapper = wrapper->bind_device->physical_device;
-            auto call_id                 = physical_device_wrapper->parent_info.api_version >= VK_MAKE_VERSION(1, 2, 0)
-                                               ? format::ApiCall_vkGetBufferDeviceAddress
-                                               : format::ApiCall_vkGetBufferDeviceAddressKHR;
+            const vulkan_wrappers::DeviceMemoryWrapper* memory_wrapper =
+                state_table.GetVulkanDeviceMemoryWrapper(wrapper->bind_memory_id);
 
-            parameter_stream_.Clear();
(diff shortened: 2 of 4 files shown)


framework/encode/vulkan_state_writer.cpp
```diff
@@ -1450,18 +1450,28 @@ void VulkanStateWriter::WriteBufferDeviceAddressState(const VulkanStateTable& st
...
-            parameter_stream_.Clear();
-            encoder_.EncodeHandleIdValue(wrapper->bind_device->handle_id);
-            VkBufferDeviceAddressInfoKHR info{ VK_STRUCTURE_TYPE_BUFFER_DEVICE_ADDRESS_INFO, nullptr, wrapper->handle };
-            EncodeStructPtr(&encoder_, &info);
-            encoder_.EncodeVkDeviceAddressValue(wrapper->address);
-            WriteFunctionCall(call_id, &parameter_stream_);
-            parameter_stream_.Clear();
+            // Memory could have been freed before the buffer was destroyed, so only write the buffer device address
+            // if the memory still exists.
+            if (memory_wrapper != nullptr)
```
fabian-lunarg: the check is good, but this could be a regression for sparse-buffers: 
`wrapper->bind_memory_id` is never populated for sparse and we would wrongly skip them here.

so adding this should be fine:
`if (memory_wrapper != nullptr || wrapper->is_sparse_buffer)`

fabian-lunarg (approved):
approved when guard handles sparse-buffers.

framework/encode/vulkan_state_writer.cpp
locke-lunarg: Thanks! Fixed!

MarkY-LunarG (approved):

locke-lunarg:
@fabian-lunarg @MarkY-LunarG
I checked vkspec, https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#VUID-vkGetBufferDeviceAddress-bufferDeviceAddress-03324

According to VUID-vkGetBufferDeviceAddress-bufferDeviceAddress-03324,
The buffer has to
1. be a sparse, or
2. bind a valid memory, or
3. be created with VK_BUFFER_CREATE_DEVICE_ADDRESS_CAPTURE_REPLAY_BIT.

Although it's legal to GetBufferDeviceAddress, the replay address could be 0 and mismatch if the memory is invalid, so print warning, instead of ASSERT.

zmike:
This actually causes a regression when trimming some traces.

zmike:
Or maybe not.

locke-lunarg:
@zmike  Could you please explain it more detail for me? Thanks!

zmike:
Ignore my comments.

MarkY-LunarG (approved):

framework/encode/vulkan_state_writer.cpp
```diff
@@ -1535,18 +1535,36 @@ void VulkanStateWriter::WriteBufferDeviceAddressState(const VulkanStateTable& st
...
+            // The buffer has to
+            // 1. be a sparse, or
+            // 2. bind a valid memory, or
+            // 3. be created with VK_BUFFER_CREATE_DEVICE_ADDRESS_CAPTURE_REPLAY_BIT.
+            //
+            // Memory could have been freed before the buffer was destroyed, so only write the buffer device address
+            // if the memory still exists.
+            if (memory_wrapper != nullptr || wrapper->is_sparse_buffer ||
+                (wrapper->modified_flags & VK_BUFFER_CREATE_DEVICE_ADDRESS_CAPTURE_REPLAY_BIT) ==
+                    VK_BUFFER_CREATE_DEVICE_ADDRESS_CAPTURE_REPLAY_BIT)
```
fabian-lunarg: I think this third clause (`wrapper->modified_flags & VK_BUFFER_CREATE_DEVICE_ADDRESS_CAPTURE_REPLAY_BIT` ...) is always true for any buffer that reaches it, which would make the whole if-statement never reject anything:

- we only arrive here when `wrapper->address != 0`
- this was set by `TrackBufferDeviceAddress()` or `TrackAccelerationStructureKHRDeviceAddress()`
-> meaning we also have `SHADER_DEVICE_ADDRESS` or `ACCELERATION_STRUCTURE_STORAGE_BIT_KHR` usage
-> `OverrideCreateBuffer()` has added `VK_BUFFER_CREATE_DEVICE_ADDRESS_CAPTURE_REPLAY_BIT`
-> `wrapper->modified_flags` is a copy of those modified creation-flags, so it looks to me, the bit is always set

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -10133,14 +10133,34 @@ VkDeviceAddress VulkanReplayConsumerBase::OverrideGetBufferDeviceAddress(
...
+    // 2. bind a valid memory, or
+    // 3. be created with VK_BUFFER_CREATE_DEVICE_ADDRESS_CAPTURE_REPLAY_BIT.
+    //
+    // Although it's legal to GetBufferDeviceAddress,
+    // the replay address could be 0 and mismatch if the memory is invalid, so print warning, instead of ASSERT.
+    //
+    // For trimming case, the buffer could still exist, but the memory is freed when the trimming starts.
+    if (device_info->allocator->SupportsOpaqueDeviceAddresses() && original_result != replay_device_address)
+    {
+        GFXRECON_LOG_WARNING(
```
fabian-lunarg, line 10154: couple smaller things here:

- unclear what "It" is
- missing space
- "replay address is ID = ..." -> it's an address (not ID), so nicer to use hex for printing

e.g. 

```
GFXRECON_LOG_WARNING(
    "Opaque device addresses are supported, so captured and replay buffer device addresses are "
    "expected to match. Buffer (ID = %" PRIu64 ") was captured with address 0x%" PRIx64
    ", but replay returned 0x%" PRIx64 ". This can happen when the buffer's memory was freed "
    "before the trim range started.",
    buffer_info->capture_id,
    original_result,
    replay_device_address);
```

framework/encode/vulkan_state_writer.cpp
fabian-lunarg: basically this check would only make sense if the original/unchanged `pCreateInfo->flags` were checked here

fabian-lunarg: but it's a tooling-flag, so apps don't set the bit and GFXR's injected bit must not count.
-> the check should be just `if(memory_wrapper != nullptr || wrapper->is_sparse_buffer)`

locke-lunarg: Or should I just remove all check? It could write `GetBufferDeviceAddress` even if the `memory_wrapper == nullptr` since it's legal for vkspec. We just change the ASSERT to warning log during replay?

framework/decode/vulkan_replay_consumer_base.cpp
locke-lunarg: Thanks. Fixed.

framework/encode/vulkan_state_writer.cpp
fabian-lunarg: no, actually still think the check is sound. the whole motivation was that no memory could be bound, which does make sense. then in addition the consideration about sparse buffers. so for me, nothing has changed since my 1st review and the check should be `if(memory_wrapper != nullptr || wrapper->is_sparse_buffer)`

locke-lunarg: Thanks. Changed!

fabian-lunarg (approved):

PR #3100, opened 2026-07-13
rebind: detect memory aliasing by per-bind range overlap

fabian-lunarg:
- detect aliasing when a bind's captured range overlaps a resource already
   bound to the same memory; share the existing allocation at the relative
   offset instead of repacking
- use VkBufferCreateInfo::size as footprint fallback when capture_req.size==0
   (requirement query absent), so disjoint pools aren't misread as aliased
- default everything else to the per-resource path
- FindAliasedMemoryInfo guards fit/alignment -> warn, never mis-place
- track bound_ranges per memory. drop entries on resource destroy or bind-failure

this PR attempts to solve the same issue like https://github.com/LunarG/gfxreconstruct/pull/3042

### DIFF
--- a/framework/decode/vulkan_rebind_allocator.cpp
+++ b/framework/decode/vulkan_rebind_allocator.cpp
@@ -295,6 +295,7 @@ VkResult VulkanRebindAllocator::CreateBuffer(const VkBufferCreateInfo*    create
             auto resource_alloc_info         = new ResourceAllocInfo;
             resource_alloc_info->usage       = create_info->usage;
             resource_alloc_info->object_type = VK_OBJECT_TYPE_BUFFER;
+            resource_alloc_info->create_size = create_info->size;
             (*allocator_data)                = reinterpret_cast<uintptr_t>(resource_alloc_info);
 
             if (create_info->pNext != nullptr)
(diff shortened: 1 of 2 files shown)


fabian-lunarg:
fyi @wzchu-lunarg

wzchu-lunarg:
Tested `-m rebind` on the samples from https://github.com/LunarG/Projects/issues/1171 and looks like it fixes them!

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -620,6 +621,104 @@ void VulkanRebindAllocator::FreeMemory(VkDeviceMemory               memory,
...
+        {
+            continue;
+        }
+
+        VmaMemoryInfo*     existing = range.vma_mem_info;
+        const VkDeviceSize base     = existing->offset_from_original_device_memory;
+
+        // Dedicated-allocation requirements forbid sharing: the newcomer may not bind into shared
+        // memory, and the existing resource's dedicated memory may not back another resource.
+        if (requires_dedicated_allocation || existing->requires_dedicated_allocation)
```
per-mathisen-arm: If a dedicated allocation overlapped at capture time with some other object then something must have gone horribly wrong somewhere and we should probably LOG_FATAL. The first condition is more plausible.

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -620,6 +621,104 @@ void VulkanRebindAllocator::FreeMemory(VkDeviceMemory               memory,
...
+        {
+            GFXRECON_LOG_WARNING(
+                "Rebind aliasing: a dedicated allocation is required at replay; cannot reproduce the share.");
+            return nullptr;
+        }
+
+        // The existing allocation's memory type must be allowed for the newcomer at replay.
+        if ((replay_req.memoryTypeBits & (1u << existing->allocation_info.memoryType)) == 0)
+        {
+            GFXRECON_LOG_WARNING("Rebind aliasing: memory type %u is not supported by the resource at replay; cannot "
```
per-mathisen-arm, line 676: In general, I wonder what we should do here. My go to solution would be to LOG_FATAL in all cases where aliasing could not take place, because the result will be corruption, and I would rather we fail early.

mikes-lunarg (approved):

framework/decode/vulkan_rebind_allocator.cpp
fabian-lunarg: I agree that this case is already deeply screwed in itself. it's mostly coming from thinking about 'what could possibly go wrong'. I'd give a low probability we'll ever see it in practice.

framework/decode/vulkan_rebind_allocator.cpp
fabian-lunarg: > In general, I wonder what we should do here

yes agree, this is basically a case where a faithful replay is just not possible anymore.

compared to what we currently do (1:1 memory/resource mapping, no aliasing reproduction at all) 
`LOG_FATAL` might be too harsh and could feel like a regression for 'edge-cases'.
at least, the PR adds a warning in those cases, so we wouldn't 'silently' fallback to per-resource.

framework/decode/vulkan_rebind_allocator.cpp
fabian-lunarg: > something must have gone horribly wrong somewhere and we should probably LOG_FATAL. The first condition is more plausible

good point. it's two different things. it does make sense to `LOG_FATAL` for `existing->requires_dedicated_allocation`.

for the 1st case `requires_dedicated_allocation` I'd prefer the warning. it allows us to detect failed cases, where currently I cannot say if/where we would hit this in the wild

fabian-lunarg: added a commit adding `LOG_FATAL` for the capture-time overlap

fabian-lunarg:
thank you @mikes-lunarg!

PR #3103, opened 2026-07-13
Update tools that use features to use a shared "--version" output format

MarkY-LunarG:
This breaks up the changes into multiple commits which might be easier to individual review than the final complete review.

1) Add a new shared feature version infrastructure (no tool changes)
2) Individual tools switching to the new version infrastructure (replay, convert, extract, optimize and info) all in separate commits
3) Remove the GFXRECON_APP_NAME_PREFIX as it didn't satisfy our original need
4) Modify the D3D12 version to use the #define D3D12_SDK_VERSION instead of a hand-coded one which could be forgotten.


MarkY-LunarG:
NOTE: This replaces PR #3092 which I have closed.

framework/util/api_version_info.cpp
```diff
@@ -0,0 +1,64 @@
...
+#endif
+
+#if defined(D3D12_SUPPORT)
+#include "d3d12.h"
+#endif
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(util)
+
+std::string GetVulkanHeaderVersionString()
```
davidd-lunarg, line 41: Should this be guarded with `#if defined(GFXRECON_ENABLE_VULKAN)`?

tools/replay/desktop_main.cpp
```diff
@@ -73,27 +75,32 @@ const char kLayerEnvVar[] = "VK_INSTANCE_LAYERS";
 
 int main(int argc, const char** argv)
 {
-    int return_code = 0;
+    struct LogGuard
```
davidd-lunarg: This `LogGuard` struct seems to be duplicated across several tools. Is there a common place it could live?

tools/replay/desktop_main.cpp
```diff
@@ -73,27 +75,32 @@ const char kLayerEnvVar[] = "VK_INSTANCE_LAYERS";
...
-        gfxrecon::util::Log::Release();
-        exit(0);
+        return EXIT_SUCCESS;
     }
     else if (arg_parser.IsInvalid() || (arg_parser.GetPositionalArgumentsCount() != 1))
     {
         PrintUsage(argv[0]);
-        gfxrecon::util::Log::Release();
-        exit(-1);
+        return EXIT_FAILURE;
```
davidd-lunarg: I like the standardization of the exit codes, but this changes replay's behavior so the failure exit code is now `1` instead of `-1`. This has the potential to break downstream dependencies like automated testing. It seems like this change is orthogonal to the other changes in this PR so I wonder if it should be a separate PR?

tools/replay/desktop_main.cpp
MarkY-LunarG: Removed this change for now.

tools/replay/desktop_main.cpp
MarkY-LunarG: Removed this change for now.

external/AgilitySDK/Update_Agility_SDK.md
```diff
@@ -8,7 +8,8 @@
 # Install Agility SDK
 - Extract `microsoft.direct3d.d3d12.x.xxx.x.nupkg`. `7-Zip` could do it.
 - Copy `microsoft.direct3d.d3d12.x.xxx.x\build\native` folder to `gfxreconstruct\external\AgilitySDK`.
-- Modify `D3D12SDKVersion` in `tools\optimize\main.cpp` and `tools\replay\desktop_main.cpp`. Also `D3D12SDKPath` if needed.
+- `D3D12SDKVersion` in `tools\optimize\main.cpp` and `tools\replay\desktop_main.cpp` is sourced from
+  `D3D12_SDK_VERSION` in the new `d3d12.h`, so no manual update is needed. Update `D3D12SDKPath` if needed.
```
mikes-lunarg: If this turns into a no-op we can just remove this step

framework/util/api_version_info.cpp
```diff
@@ -0,0 +1,66 @@
...
+    return std::string("OpenXR Header Version ") + std::to_string(XR_VERSION_MAJOR(XR_CURRENT_API_VERSION)) + "." +
+           std::to_string(XR_VERSION_MINOR(XR_CURRENT_API_VERSION)) + "." +
+           std::to_string(XR_VERSION_PATCH(XR_CURRENT_API_VERSION));
+}
+#endif
+
+#if defined(D3D12_SUPPORT)
+std::string GetD3D12SdkVersionString()
+{
+    return std::string("D3D12 SDK Version ") + std::to_string(D3D12_SDK_VERSION);
```
mikes-lunarg, line 61: @davidd-lunarg Any thoughts if we should use the Agility SDK version or the Windows SDK version here? Or both? I was initially thinking the Windows SDK since that is what drives the codegen, but now I see a couple tools were already reporting the Agility version.

davidd-lunarg: `d3d12.h` comes from the Agility SDK. All the DXGI files come from the Windows SDK. So it might makes sense to list both. If I had to pick only one, I'd pick the Agility SDK version `D3D12_SDK_VERSION`.

davidd-lunarg: The contents of the Agility SDK's `d3d12.h` specifies the feature set of D3D12. DXGI changes less frequently and is mostly updated by the Windows SDK. The code generators pull in headers from both the Agility SDK (checked into GFXR's externals) and from the system installed Windows SDK. Besides the DXGI headers, some common Windows headers are also pulled in from the Windows SDK during generation.

mikes-lunarg: Ok thanks, I didn't realize we also generated from the Agility headers but that makes sense.

mikes-lunarg (approved):

davidd-lunarg (approved):

PR #3104, opened 2026-07-13
Fix Vulkan-disabled build

MarkY-LunarG:
The recent changes for ARM Tensors introduced a breakage in the Vulkan-disabled build of GFXR.  This resolves  it.


fabian-lunarg (approved):
thank you for fixing!

PR #3107, opened 2026-07-14
[Frame looping] Fix nullptr crash regarding fence handling

nickdriscoll-lunarg:
This PR changes the frame loop handling of vkDestroyFence such that it deletes the fence tracking info if destroy was called during the first iteration of the loop range. Before this change, in the case of a fence being created and destroyed in the same frame, replay would crash with a nullptr dereference.

### DIFF
--- a/framework/decode/vulkan_replay_frame_loop_consumer.cpp
+++ b/framework/decode/vulkan_replay_frame_loop_consumer.cpp
@@ -279,6 +279,11 @@ void VulkanReplayFrameLoopConsumer::Process_vkDestroyFence(const ApiCallInfo& ca
         if (per_device_fence_tracking_.contains(args.device))
         {
             FenceTracking& t = per_device_fence_tracking_[args.device];
+
+            // Remove fence tracking struct from map if
+            // fence was destroyed during the loop range.
+            // This is a no-op in the case that the fence
+            // was created before the loop range.
             t.initial_fence_states_.erase(args.fence);
         }
     }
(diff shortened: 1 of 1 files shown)


framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -274,11 +274,16 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateFence(const ApiCallInfo& cal
...
     {
         if (per_device_fence_tracking_.contains(args.device))
         {
             FenceTracking& t = per_device_fence_tracking_[args.device];
+
+            // Remove fence tracking struct from map if
+            // fence was destroyed during the loop range.
+            // This is a no-op in the case that the fence
+            // was created before the loop range.
             t.initial_fence_states_.erase(args.fence);
```
antonio-lunarg, line 285: Actually, this is not a no-op. Consider this:

1. `StartLooping()` calls `TrackFenceState` for all fences, effectively snapshotting the state of all fences created and alive before the loop.
2. We should not destroy such fences in `Process_vkDestroyFence`, because they might be used by the loop range. Indeed `VulkanReplayFrameLoopConsumerBase::Process_vkDestroyFence` only calls `VulkanReplayConsumer::Process_vkDestroyFence` in the last iteration.

I believe you need to check `if (allocatedLoopResources.contains(args.fence))` here, because we want to erase the correspoinding entry from the `initial_fence_states_` only when the fence is actually going to be destroyed, and that would be the case for fences that have been created _within_ the loop.

In other words, I would just replace `if (frame_loop_info_.IsLooping() && !frame_loop_info_.IsRepetition())` with `if (allocatedLoopResources.contains(args.fence))`

nickdriscoll-lunarg: This makes sense.

davidlunarg (approved):

antonio-lunarg (approved):

PR #3108, opened 2026-07-15
replay: Reissue command buffer state

antonio-lunarg:
- Add a new VulkanStateRecordingDecoder that records the raw parameter buffers of state-setting commands per command buffer.
- When a command buffer is split, it reissues those recorded state commands into the new split buffer so draws/dispatches still have valid state (with a recursion guard and state-clearing on submit).
- Wires this up by making the feature's decoder injectable, so this custom decoder is used only when isolate_render_passes is set.

### DIFF
--- a/framework/decode/vulkan_command_buffer_util.cpp
+++ b/framework/decode/vulkan_command_buffer_util.cpp
@@ -80,9 +80,10 @@ void VulkanCommandBufferAssociatedInfo::ReplaceWithNewHandle(VulkanCommandBuffer
 
 VulkanCommandBufferUtil::VulkanCommandBufferUtil(const VulkanDeviceInfo*            device_info,
                                                  const graphics::VulkanDeviceTable* device_table,
-                                                 CommonObjectInfoTable*             object_table) :
+                                                 CommonObjectInfoTable*             object_table,
+                                                 VulkanStateRecordingDecoder*       decoder) :
     device_info_(device_info),
-    device_table_(device_table), object_table_(object_table)
+    device_table_(device_table), object_table_(object_table), decoder_(decoder)
 {
     GFXRECON_ASSERT(device_info != nullptr);
     GFXRECON_ASSERT(device_table != nullptr);
@@ -155,6 +156,11 @@ void VulkanCommandBufferUtil::ReplaceWithAssociatedCommandBuffer(VulkanCommandBu
 
 void VulkanCommandBufferUtil::SplitCommandBuffer(VulkanCommandBufferInfo* command_buffer_info)
 {
+    if (reissuing_command_buffer_state_ == true)
+    {
+        return;
+    }
+
     device_table_->EndCommandBuffer(command_buffer_info->handle);
 
     GetOrCreateAssociatedInfo(command_buffer_info->capture_id).PushSplitHandle(command_buffer_info->handle);
@@ -162,6 +168,10 @@ void VulkanCommandBufferUtil::SplitCommandBuffer(VulkanCommandBufferInfo* comman
 
     VkCommandBufferBeginInfo begin_info = { VK_STRUCTURE_TYPE_COMMAND_BUFFER_BEGIN_INFO };
     device_table_->BeginCommandBuffer(command_buffer_info->handle, &begin_info);
+
+    reissuing_command_buffer_state_ = true;
+    decoder_->ReissueCommandBufferState(command_buffer_info->capture_id);
+    reissuing_command_buffer_state_ = false;
 }
 
 graphics::VulkanSemaphore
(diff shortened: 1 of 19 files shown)


MarkY-LunarG:
This feels similar to the trimming tracker.  What's the differences?

antonio-lunarg:
> @MarkY-LunarG  This feels similar to the trimming tracker. What's the differences?

Mh, yeah, both capture raw encoded commands. Differences are:
- trimming tracker is in the capture layer, my decoder in replay.
- trimming tracker serializes the blocks into the capture file, my decoder decodes and hands them to the consumer.
- trimming tracker captures all blocks, my decoder filters them.

Do you see any code that could be shared by both?

antonio-lunarg:
> > @MarkY-LunarG  This feels similar to the trimming tracker. What's the differences?
> 
> Mh, yeah, both capture raw encoded commands. Differences are:
> 
> * trimming tracker is in the capture layer, my decoder in replay.
> * trimming tracker serializes the blocks into the capture file, my decoder decodes and hands them to the consumer.
> * trimming tracker captures all blocks, my decoder filters them.
> 
> Do you see any code that could be shared by both?

Maybe I can extract a mechanism for recording API calls in memory which would provide a way to visit them later.

MarkY-LunarG:
> > @MarkY-LunarG  This feels similar to the trimming tracker. What's the differences?
> 
> Mh, yeah, both capture raw encoded commands. Differences are:
> 
> * trimming tracker is in the capture layer, my decoder in replay.
> * trimming tracker serializes the blocks into the capture file, my decoder decodes and hands them to the consumer.
> * trimming tracker captures all blocks, my decoder filters them.
> 
> Do you see any code that could be shared by both?

Not off hand.  Sorry, I forgot this was replay-side.  Thanks.

antonio-lunarg:
> > > @MarkY-LunarG  This feels similar to the trimming tracker. What's the differences?
> > 
> > 
> > Mh, yeah, both capture raw encoded commands. Differences are:
> > 
> > * trimming tracker is in the capture layer, my decoder in replay.
> > * trimming tracker serializes the blocks into the capture file, my decoder decodes and hands them to the consumer.
> > * trimming tracker captures all blocks, my decoder filters them.
> > 
> > Do you see any code that could be shared by both?
> 
> Not off hand. Sorry, I forgot this was replay-side. Thanks.

I extracted some common code in a mechanism to record and visit encoded API calls. This involves state trackers, state writers, and the new recording decoder.

framework/decode/vulkan_state_recording_decoder.h
```diff
@@ -0,0 +1,152 @@
...
+            case format::ApiCall_vkCmdClearAttachments:
+            case format::ApiCall_vkCmdClearColorImage:
+            case format::ApiCall_vkCmdClearDepthStencilImage:
+            case format::ApiCall_vkCmdCopyBuffer:
+            case format::ApiCall_vkCmdCopyBufferToImage:
+            case format::ApiCall_vkCmdCopyImage:
+            case format::ApiCall_vkCmdCopyImageToBuffer:
+            case format::ApiCall_vkCmdDispatch:
+            case format::ApiCall_vkCmdDispatchBaseKHR:
+            case format::ApiCall_vkCmdDispatchIndirect:
```
MarkY-LunarG: Should this be code generated?

framework/decode/vulkan_command_buffer_util.cpp
```diff
@@ -209,6 +219,9 @@ graphics::VulkanSemaphore VulkanCommandBufferUtil::SubmitPreviouslySplitCommandB
                 // The command buffer in this submit has associated command buffers.
                 auto command_buffer_id = original_command_buffer_id_[command_buffer];
 
+                // Make sure to clear the recorded state for this command buffer.
+                decoder_->ClearRecordedState(command_buffer_id);
```
MarkY-LunarG: Why are you clearing the recorded state when you submit a split command buffer?  Won't you need it longer?

framework/decode/vulkan_state_recording_decoder.h
antonio-lunarg: Code for generating this would probably end up being longer than this list. Also, I believe we'd need different list depending on the "mode". If we want to isolate render passes, we do not want to record BeginRenderPass, but if we want to isolate draw calls, then we do want to record that.

framework/decode/vulkan_command_buffer_util.cpp
antonio-lunarg: Imagine the case a command buffer is recorded and submitted twice within a frame. I think I want to make sure I do not record commands related to another recording and submit all of them by mistake.

MarkY-LunarG (approved):
This latest commit feels much cleaner to me.

PR #3109, opened 2026-07-15
rebind: Fix shadow AccelerationStructure tracking in VulkanAddressReplacer

fabian-lunarg:
A ray-tracing capture (-m rebind) replayed fine on NVIDIA but crashed on RADV after a
`VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03800`

Two RADV specifics exposed latent bugs in the shadow-tracking logic that NVIDIA never triggered:

- `minAccelerationStructureScratchOffsetAlignment` is 256 on RADV (128 on NVIDIA). 
   ->  Rebound scratch-addresses ended up misaligned, so always using shadow/replacement path
         (was not used on NVIDIA)
- RADV recycled freed buffer device-addresses almost immediately, 
   but `shadow_as_map_` is keyed by replay device-address, and entries were not invalidated on
   destruction.
   -> A new acceleration-structure landing on a recycled address used a stale entry
        and had its valid dst handle swapped for a destroyed one.

The bug(s) are driver-independent. RADV just happened to trigger them for me.

### DIFF
--- a/framework/decode/vulkan_address_replacer.cpp
+++ b/framework/decode/vulkan_address_replacer.cpp
@@ -310,6 +310,7 @@ VulkanAddressReplacer::~VulkanAddressReplacer()
     pipeline_context_map_.clear();
     shadow_sbt_map_.clear();
     shadow_as_map_.clear();
+    shadow_scratch_map_.clear();
     submit_asset_map_.clear();
     submit_asset_ = {};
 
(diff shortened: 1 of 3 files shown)


framework/decode/vulkan_address_replacer.cpp
```diff
@@ -1350,20 +1351,29 @@ void VulkanAddressReplacer::ProcessCmdBuildAccelerationStructuresKHR(
...
+                buffer_context_t* replacement_scratch = nullptr;
 
-                auto& replacement_as = shadow_as_map_[as_replay_address];
-
-                if (replacement_as.handle == VK_NULL_HANDLE)
+                if (!as_buffer_usable)
                 {
-                    if (as_buffer_usable)
+                    // now definitely requiring address-replacement
+                    force_replace = true;
```
antonio-lunarg, line 1360: If the scratch buffer is not usable, `force_replace` stays `false`. Is that intentional?

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -1603,7 +1621,7 @@ void VulkanAddressReplacer::ProcessUpdateDescriptorSets(uint32_t              de
                 {
                     auto acceleration_structure_it = shadow_as_map_.find(acceleration_structure_info->replay_address);
                     if (acceleration_structure_it != shadow_as_map_.end() &&
-                        acceleration_structure_it->second.storage.num_bytes > 0)
+                        acceleration_structure_it->second.origin_handle == write_as->pAccelerationStructures[j])
```
antonio-lunarg, line 1628: You removed `acceleration_structure_it->second.storage.num_bytes > 0`.
What if `num_bytes == 0`? That would guard against failed shadow-AS creation.

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: yes. it was in fact not necessary (or even problematic) to set `force_replace` 
when only the scratch-buffer could not be re-used. 
we ended up replacing perfectly usable AS-addresses.

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: this was also a consequence from changed behavior wrt. scratch-buffers.
I thought we could skip it here since we dropped 'scratch-only' pseudo-entries in `shadow_as_map_`,
but you're right about the failed shadow-AS creation.

I think instead of re-adding the check, we could cleaner remove shadow-AS that failed creation,
right where we already had a warning

fabian-lunarg: done, added a commit, dealing with failed shadow-AS creation

antonio-lunarg (approved):

PR #3110, opened 2026-07-15
Fix unnecessary synchronization in virtual swapchain image transition

marius-pelegrin-arm:
On virtual swapchain image transition, `vkQueueWaitIdle` is called when a `VkFence` associated to the submitted workload is already available.

I propose to replace this with a `vkWaitForFences` call.

### DIFF
--- a/framework/decode/vulkan_virtual_swapchain.cpp
+++ b/framework/decode/vulkan_virtual_swapchain.cpp
@@ -636,7 +636,7 @@ VkResult VulkanVirtualSwapchain::TransitionSwapchainImage(VkDevice              
         return result;
     }
 
-    result = device_table_->QueueWaitIdle(initial_copy_queue_[device]);
+    result = device_table_->WaitForFences(device, 1, &copy_fence, VK_TRUE, UINT64_MAX);
     if (result != VK_SUCCESS)
     {
         GFXRECON_LOG_ERROR("TransitionSwapchainImage: Virtual swapchain failed waiting for internal command buffer %d "


framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -522,8 +522,8 @@ VkResult VulkanVirtualSwapchain::TransitionSwapchainImage(VkDevice
     begin_info.pInheritanceInfo         = nullptr;
 
     auto& copy_cmd_data  = swapchain_resources->copy_cmd_data[copy_queue_family_index_[device]];
-    auto  command_buffer = copy_cmd_data.command_buffers[image_index];
-    auto  copy_fence     = copy_cmd_data.fences[image_index];
+    auto  command_buffer = copy_cmd_data.command_buffers[0];
+    auto  copy_fence     = copy_cmd_data.fences[0];
```
fabian-lunarg: there is no guarantee that there are image_index virtual
    images, as image_index is based on the replay-swapchain image
    count, and there are capture-swapchain image count virtual images

I agree with that statement, but it's different arrays here.

the change is not necessary. command-buffer/fences are from `copy_cmd_data`,
get populated in different spots, but are sized wrt. replay-image-count. :

```
copy_cmd_data.command_buffers.resize(*replay_image_count); // <- line 365
copy_cmd_data.fences.resize(*replay_image_count);          // line 417
```

So `[image_index]` is in range and `[0]` isn't fixing an overflow.

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -635,7 +635,7 @@ VkResult VulkanVirtualSwapchain::TransitionSwapchainImage(VkDevice
         return result;
     }
 
-    result = device_table_->QueueWaitIdle(initial_copy_queue_[device]);
+    result = device_table_->WaitForFences(device, 1, &copy_fence, VK_TRUE, UINT64_MAX);
```
fabian-lunarg, line 639: agreed, that's much nicer

fabian-lunarg:
@marius-pelegrin-arm I'd suggest we drop the 1st commit but keep the 2nd

framework/decode/vulkan_virtual_swapchain.cpp
marius-pelegrin-arm: This is due to #3055 changing the number of copy_cmd_data entries. We don't have this commit internally yet.

fabian-lunarg: ok, understand. another recent PR in virtualswapchain was here, also changing API of `TransitionSwapchainImage` -> https://github.com/LunarG/gfxreconstruct/pull/3066

fabian-lunarg: the first commit has been removed and I adjusted the title + description, as agreed.

fabian-lunarg (approved):

PR #3111, opened 2026-07-16
replay: Fix decoding for VK_ARM_data_graph_optical_flow

fabian-lunarg:
we added support for `VK_ARM_tensors` and `VK_ARM_data_graph` in 
https://github.com/LunarG/gfxreconstruct/pull/3009.

we also added support for `VK_ARM_data_graph_optical_flow` but an issue remained on decode-side,
affecting all decoders like replay/convert etc.

this PR makes small corrections to codegen, dealing with proper handling of `VkBaseOutStructure`, 
which the extension relies on.

we also port an existing decode/encode **unit-test** from https://github.com/ARM-software/gfxreconstruct-arm/
since the test is an encode/decode roundtrip, it didn't make sense to put in either decode/test or encode/test.
I suggest to introduce a cross-module testing location in framework/test/ 

verified using capture/replay/convert of:
https://github.com/ARM-software/tracetooltests/blob/main/src/vulkan_data_graph_optical_flow.cpp

requires an updated version of ARM's AI-emulation with support for optical_flow:
https://github.com/arm/ai-ml-emulation-layer-for-vulkan/releases/tag/v0.10.0

added a capture-convert regression-test for `VK_ARM_data_graph_optical_flow`:
- bumped test-suite-ref


fabian-lunarg:
fyi @bartosz-muszarski-arm  - most of it was there already :)

bradgrantham-lunarg (approved):

Issue #3014, opened 2026-06-10
Improve DX12 replay adapter selection


davidd-lunarg:
DX12 adapter selection is based on either adapter enumerated index or LUID. Both values are not stable for a machine, even if there are no hardware changes. LUIDs can be change per reboot. Make a more robust solution to finding an adapter that matches the capture-time adapter, similar to what Vulkan does.

Also, make sure GFXR replay logs a message if the replay time adapter does not match the capture time adapter.

davidd-lunarg:
When an applicaiton uses `ID3D12DeviceFactory::CreateDevice` to create the device, GFXR doesn't correctly track the adapter and capture issues the warning:
`[gfxrecon] WARNING - No active adapters were found`

Identified when verifying PR #3087: https://github.com/LunarG/gfxreconstruct/pull/3087#pullrequestreview-4665599649

Support for adapters created via the factory method should also be handled.

PR #3113, opened 2026-07-16
Improve D3D12 adapter selection to use device and vendor IDs

wzchu-lunarg:
Previously, adapter information was only guaranteed in blocks after device creation, which was an injected `DxgiAdapterInfo` command after device creation. However, since this information comes after, we could not utilize the device or vendor IDs to select the adapter, which is what the Vulkan GFXR device selection does.

Adapter selection was done by faithfully reusing the adapter indices if the app passed an explicit adapter https://github.com/LunarG/gfxreconstruct/issues/3014. Furthermore, the info for vendor and device IDs were only used for warnings as it came after the device creation.

This leads to issues in some traces selecting the wrong adapter over time, even on the same setup and devices.

Mainly the issue comes from not being able to access adapter desc during device creation time as `getDesc` calls are not required by the application (application could just decide to use first enumerated adapter without checking, or pass `nullptr` to createDevice).

**Approach**
We can instead inject an enumeration along with their descriptions before device creation. Since an explicitly passed adapter to device creation has to get that adapter from enumeration, an enumeration command is guaranteed before device creation which we could track, which lets us map adapter handles to their captured adapter indices, and ultimately their descriptions.

**Considerations**
The changes targets `EnumAdapters` and `EnumAdapters1`, while there are other enumerations, `EnumAdapterByLuid`, `EnumWarpAdapter`, and `EnumAdapterByGpuPreference`. These enumerations won't capture and map the adapter handles to their indices, and therefore will leave `capture_desc` as `nullptr`, resulting in falling back to the previous behavior of using the passed `adapter_info`.

If `nullptr` was genuinely passed to createDevice, try to match capture's default adapter which would be the first enumerated adapter at index 0.

Backwards compatibility might have issues where old traces won't have the injected enumeration of all adapters. If the app never enumerates, then `capture_desc` will be `nullptr` and again fall back using the passed `adapter_info` handle. However, these cases are all combined into one warning when we don't have `capture_desc`, which is not as verbose as the previous warnings from processing `DxgiAdapterInfo`.


framework/encode/d3d12_capture_manager.cpp
```diff
@@ -3130,6 +3130,29 @@ void D3D12CaptureManager::WriteDxgiAdapterInfo()
     }
 }
 
+void D3D12CaptureManager::InjectDxgiAdapterEnumeration()
```
davidd-lunarg: If I'm reading this right, this will inject the commands into the capture file so that they appear as if the application called them. Generally GFXR should avoid adding command that weren't application-made. Instead a new meta data block could be used that contains the enumerated adapter info. This is how Vulkan does it with `kSetDevicePropertiesCommand` and `kSetDeviceMemoryPropertiesCommand` meta data commands.

davidd-lunarg (changes requested):

wzchu-lunarg:
Changed the approach to use new meta data blocks, `D3D12CreateDeviceAdapterInfo`. During capture, we find adapter descriptions by calling `TrackAdapters` to populate `adapters_` if we haven't already. We can then use the current capture time adapter's LUID during device creation and find the corresponding adapter. If the adapter passed was `nullptr`, we instead enumerate to find the adapter at index 0.

There is a limitation where if an app doesn't use `DXGI` functions before device creation, such as not creating a factory while also passing a `nullptr` adapter, there doesn't seem to be a way to track the adapters, as we don't have the real DXGI functions loaded. `dxgi_dispatch_table_.CreateDXGIFactory1` will just be a `nullptr` and we will have no factory to enumerate over.

framework/format/format.h
```diff
@@ -690,6 +691,14 @@ struct DxgiAdapterInfoCommandHeader
     DxgiAdapterDesc adapter_desc;
 };
 
+struct D3D12CreateDeviceAdapterInfoCommandHeader
```
davidd-lunarg, line 694: Could this struct also include the device's `HandleId` to make the association between device and adapter explicit and position independent (as long as it comes any time before the device creation)?

Since the device's `HandleId` isn't known until after the device is created, this would have to be done after the device object gets wrapped but before its creation call is written to the capture file. For example, by modifying the generators to add the write call to `D3D12CreateDevice` and `ID3D12DeviceFactory_Wrapper::CreateDevice` in `framework\generated\generated_dx12_wrappers.cpp`:

```
        if (SUCCEEDED(result))
        {
            WrapObject(riid, ppDevice, nullptr);

+           // Write device adapter info to capture file. 
+           manager->WriteD3D12CreateDeviceAdapterInfoCommand(ppDevice);
        }

        Encode_D3D12CreateDevice(
            result,
            pAdapter,
            MinimumFeatureLevel,
            riid,
            ppDevice);
```

Modification to the generators could look something like this in `def gen_wrap_object` in `dx12_wrapper_body_generator.py`:

```
            for tuple in params_wrap_obejct:
                if not tuple[2]:
                    expr += indent + 'WrapObject({}, {}, nullptr);\n'.format(
                        tuple[0], tuple[1]
                    )
+                   # Post wrap, pre-encode
+                   if(tuple[1] == "ppDevice"):
+                      expr += indent + "manager->WriteD3D12CreateDeviceAdapterInfoCommand(...);\n"
                else:
```
Although it may be better to put the gen code in a new function to keep `gen_wrap_object` single-purpose for wrapping the object.

This would also allow use of `ID3D12Device::GetAdapterLuid` to help resolve the adapter.

tools/info/main.cpp
```diff
@@ -110,7 +110,7 @@ void WriteOutput(const char* format_string, ...)
         }
         else
         {
-            GFXRECON_WRITE_CONSOLE(result_string.c_str());
+            GFXRECON_WRITE_CONSOLE("%s", result_string.c_str());
```
davidd-lunarg, line 113: Nice catch.

framework/encode/d3d12_capture_manager.cpp
```diff
@@ -2503,6 +2502,8 @@ void D3D12CaptureManager::PreProcess_D3D12CreateDevice(IUnknown*         pAdapte
     {
         EnableDRED();
     }
+
+    WriteD3D12CreateDeviceAdapterInfoCommand(pAdapter);
```
davidd-lunarg: This would also need to be written before device creation in `Dx12StateWriter::WriteState`--see line `StandardCreateWrite<ID3D12Device_Wrapper>(state_table);`--to ensure trimmed captures get the new block.

jzulauf-lunarg (commented):
The FileProcessor side LGTM.

framework/format/format.h
wzchu-lunarg: Added `additional_commands` to the `gen_wrap_object` function to specify these additional commands.

wzchu-lunarg:
The mismatch message can still be confusing, especially if there are multiple create device calls, and it can be unclear which call has the mismatch, or if the mismatches shown are from different create device calls or the same one. Perhaps adding block indices to the warnings is a reasonable change?

PR #3117, opened 2026-07-17
[Frame Looping] Fix handling of descriptor pools and fence creation

nickdriscoll-lunarg:
`vkCreateDescriptorPool` can be generated, so it now is. `vkDestroyDescriptorPool` still calls `RemovePoolDanglingCreateDescriptors` as before.

The checks in `Process_vkCreateFence` and `Process_vkCreateCommandPool` were wrong and unnecessary, as we call the base class's implementations which handle the resources correctly.


davidlunarg:
LGTM. My test passes on Windows.

antonio-lunarg (approved):

PR #3121, opened 2026-07-22
dx12: Fix inconsistent capture and replay back buffer index

zehao-jing-arm:
If the capture present fails, add a flag DXGI_PRESENT_TEST to fake the replay present.

Change-Id: I599a947254b082b306fdad156d8eb1724c31a9a4

### DIFF
--- a/framework/decode/dx12_replay_consumer_base.cpp
+++ b/framework/decode/dx12_replay_consumer_base.cpp
@@ -1104,6 +1104,10 @@ HRESULT Dx12ReplayConsumerBase::OverridePresent(DxObjectInfo* replay_object_info
 {
     auto replay_object = static_cast<IDXGISwapChain*>(replay_object_info->object);
     PrePresent(replay_object_info, flags);
+    if (FAILED(original_result))
+    {
+        flags |= DXGI_PRESENT_TEST;
+    }
     auto result = replay_object->Present(sync_interval, flags);
     PostPresent();
 


davidd-lunarg:
What error result did the present call return during capture? What is the return during replay? What flags were used in the call?

zehao-jing-arm:
`[gfxrecon] FATAL - IDXGISwapChain_Present returned DXGI_ERROR_DEVICE_REMOVED, which does not match the value returned at capture S_OK. Replay cannot continue.
Replay has encountered a fatal error and cannot continue: The video card has been physically removed from the system, or a driver upgrade for the video card has occurred. The application should destroy and recreate the device. For help debugging the problem, call ID3D12Device::GetDeviceRemovedReason. `

and
`{
  "index": 4815112,
  "method": {
    "name": "Present",
    "thread": 10,
    "object": {
      "type": "IDXGISwapChain",
      "handle": 1364
    },
    "return": "S_OK",
    "args": {
      "SyncInterval": 0,
      "Flags": 512
    }
  }
}, `

davidd-lunarg:
`{ "index": 4815112, "method": { "name": "Present", "thread": 10, "object": { "type": "IDXGISwapChain", "handle": 1364 }, "return": "S_OK", "args": { "SyncInterval": 0, "Flags": 512 } } }, `

If I am reading this right, this `Present` call returned with `S_OK` success during capture so shouldn't be affected by this fix. What captured `Present` call failed that motivated this PR? I'd like to understand under what conditions it is valid for the `Present` call to fail during capture but might succeed during replay.

zehao-jing-arm:
You're correct that this particular Present returned S_OK during capture.
From what I remember, the replay crash was actually triggered by an earlier Present on the same swapchain that returned a non-S_OK result during capture. That left the swapchain state inconsistent, and replay later failed when it reached this subsequent S_OK Present.
Unfortunately, this issue is quite old and I no longer have access to the original trace, so I can't point to the exact failing Present call anymore or reproduce the original conditions that motivated the fix.
Given that I can't provide a concrete repro case or trace at this point, I'm fine with not merging this PR for now. We can revisit it if the issue is observed again or if a reproducible test case becomes available.

davidd-lunarg:
Sounds good, thanks.

PR #3122, opened 2026-07-22
Smart memory tracker

panos-lunarg:


framework/util/range_list.h
```diff
@@ -0,0 +1,170 @@
+/*
+** Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
```
per-mathisen-arm, line 2: Copy-pasta error. Please change to Arm :-)

framework/encode/vulkan_smart_memory_tracker.cpp
```diff
@@ -0,0 +1,341 @@
+/*
+** Copyright (c) 2026 LunarG, Inc.
```
per-mathisen-arm, line 2: Kindly add Arm here as well :-)

per-mathisen-arm (commented):
Nice. Would be good to get a nod for coming up with the original implementation, even if everything ends up re-implemented. I think if a buffer device address is retrieved for a buffer and a shader is dispatched that resolves such addresses, then all such buffers need to be considered dirty and checked. (It is possible that we could do some quick initial scan of the shader to see if we could deduce the target buffer before it is submitted, but that won't always be possible.)

panos-lunarg:
> Nice. Would be good to get a nod for coming up with the original implementation, even if everything ends up re-implemented.

This is still a draft but we will mention Arm.

>  I think if a buffer device address is retrieved for a buffer and a shader is dispatched that resolves such addresses, then all such buffers need to be considered dirty and checked.

Yes, I was thinking the same.

I have noticed that some applications (i.e. Wildlife and Solar bay) have have poor performance during capture (similar to unassisted) with smart. I am investigating the possibility to use some of the hashing logic and drop the shadow/baseline memory copies and compares.

framework/util/range_list.h
panos-lunarg: done

framework/encode/vulkan_smart_memory_tracker.cpp
panos-lunarg: done

per-mathisen-arm:
> I have noticed that some applications (i.e. Wildlife and Solar bay) have have poor performance during capture (similar to unassisted) with smart. I am investigating the possibility to use some of the hashing logic and drop the shadow/baseline memory copies and compares.

Yes, 3DMark seems to love using one massive buffer to store everything. Really hard to disentangle and check partially in a smart way - and things get very slow. One way could be to force allocations apart by manipulating memory types, but you would be changing what is traced that way. I will try to have a look.

panos-lunarg:
I pushed a few more commits in this PR where I use some hashing logic and removed the baseline memory.
This sped up wild life enough to uncover a problem: Gpu writes cannot be distinguished from cpu writes. Detecting gpu writes is racy as they can be taking place the same time as they are being detected and as a result get dumped partially in the capture file. During replay they can manifest as rendering artifacts.
More specifically wild life appears to implement deferred rending which splits it up in two submissions. In the first submission it renders into the g buffers, which are read in the second submission. During replay the FillMemory commands clobber the g buffer.
Before using hashing, capturing was so slow that it was not possible to capture more than one frame from the same scene. This effectively hide the problem.

To work around this I force optimal tilled images to not be considered for check from the memory tracker (https://github.com/LunarG/gfxreconstruct/pull/3122/changes#diff-2ffd12a8986454311af84604e3e5c3657ff9998ce0473fb89e78641f1a3bf30fR5310). However this is a work-around as theoretically the same scenario can trigger with buffers where we cannot do the same trick.

per-mathisen-arm:
But this is why I think you actually want smart tracking, not hashing: It should be able to distinguish between CPU and GPU writes. Between the first and second submission, we should have cleared the tracking and the GPU write should not appear in the dirty list. You should never encounter a situation where you both write from CPU and write from GPU to the same memory region for the same submission (although it may be unfeasible difficult to distinguish the two regions from each other).

panos-lunarg:
>  Between the first and second submission, we should have cleared the tracking and the GPU write should not appear in the dirty list.

I guess I am missing something.

The hashing logic does not replace the tracking logic, only the `memcpy`s.
The first submission has the g buffers marked as write only and are ignored from the tracker.
The second submission has them as read only (sampled images) so the tracker will check for changes.
The memory remains mapped all the time.

> You should never encounter a situation where you both write from CPU and write from GPU to the same memory region for the same submission (although it may be unfeasible difficult to distinguish the two regions from each other).

Why not? Host writes something to a buffer, buffer is first used to read from and then used again for write. Maybe not common in practice but sounds like valid usage.

per-mathisen-arm:
The basic logic is if a region of memory is referenced or bound in a command and has been mapped (or remains mapped), we have to be assume that it could have been written to by host and has to be checked. Excluding write-only would be an optimization added on top of this logic only if we are sure there are no other accesses that do reads and it is a full write (ie discard). Between host work and GPU work there has to be a submission, and this is where we do the checking - this should hold even if there are multiple queues, as the API user has to explicitly synchronize the work, and these would get recorded ensuring we get our updates in the correct order. Can you give me a more specific example of API+meta commands that could fail?

panos-lunarg:
> Can you give me a more specific example of API+meta commands that could fail

During capture:
- Command buffer with a render pass which  writes to render attachments.
- Command buffer is submitted to a queue. Tracker detects that the command buffer only writes to these images so does not check for changes.
- Another command buffer is recorded which references above render targets as sampled images.
- Second command buffer is submitted for execution. Tracker detects that the command buffer references those images as read only and checks for changes. The first command buffer is still executing. Tracker detects partial writes from the gpu to the render targets and writes FillMemoryCommand meta commands accordingly.

During replay:
- First QueueSubmit is issued.
- Second QueueSubmit is encountered in the file accompanied by the FillMemories that affect the render targets, but previous execution is still in progress. While the render pass from the first command buffer is executing, FillMemory commands from second QueueSubmit clobber the gpu writes.

panos-lunarg:
I am realizing now that perhaps detecting changes with hashing is more prone to this as each region is divided in blocks (512 bytes currently) in order to hash and compare. Original smart tracker checks by comparing original memory and baseline 8 bytes (uint64_t) each time.
However I think that the race is there either way.

per-mathisen-arm:
I think you are right. I am writing a reproducer microbenchmark now in tracetooltests, but so far unable to actually get it to fail a race. We could detect this (queue submission followed by queue submission with no intervening wait for the fence of the first submission + non-empty dirty list) and inject a dummy queue submit + wait for its fence before scanning memory. Other types of workarounds would be needed for other types of synchronization primitives, and when handling multiple queues, but I suppose binary semaphore on the same queue is the most likely case. I will do some more experiments.

per-mathisen-arm:
I have a working reproducer now at https://github.com/ARM-software/tracetooltests/blob/main/src/vulkan_memory_tracking_race.cpp - you will have to make sure the capture GPU is slow and replay GPU is fast to reliably reproduce, though. And it makes use of our new assert buffer command to trigger. The test can trigger using the four different sync mechanisms I can think of that could be used (binary semaphores, timeline semaphores, pipeline barriers, and events).

I also managed to speed up the memory scanning for smart tracking by approx 50% on average (on synthetic tests), with extreme edge cases up to 17x - https://github.com/ARM-software/lavatube/commit/843e87ddd72375924f6cea27a9aea56677ecac67

Next I will try out some ideas for fixing the race conditions.

per-mathisen-arm:
The memory scanning optimization gives me around 30% performance improvement with 3DMark Steel Nomad, but it is still incredibly slow. Not as impactful as I had hoped.

per-mathisen-arm:
I found that smart memory tracking is extremely slow when it encounters applications using host visible host coherent memory that is _not_ also host cached. Forcing use of host cached or external memory allocations made Steel Nomad on one system go from doing asset loading in 10 minutes to 3 seconds. I suspect this is a particular problem for hosts with dedicated VRAM.

per-mathisen-arm:
You did not keep my change to https://github.com/ARM-software/gfxreconstruct-arm/blame/f85cbd7f4ef6de373356553024ca5c1feefa49d5/framework/encode/vulkan_state_tracker_initializers.h#L703 which is needed to make Sascha Willem's Vulkan samples work rather than just show an empty window.

Also, gfxrecon-info incorrectly shows tracking method as page-guard even though it is set to smart, along with some spurious non-changes in capture settings. Not sure where this is coming from, but that does not happen in my branch.

panos-lunarg:
Isn't this the same assignment: https://github.com/panos-lunarg/gfxreconstruct/blob/smart_memory_tracker/framework/decode/vulkan_replay_consumer_base.cpp#L6523

per-mathisen-arm:
But that's in the replayer, the bug is in the capture.

PR #3125, opened 2026-07-22
replay: stop stripping VkExternalMemoryImageCreateInfo

### LINKED ISSUE #2813: stop stripping VkExternalMemoryImageCreateInfo in replay
https://github.com/LunarG/gfxreconstruct/blob/18f571a3984d69386acf5ddcad50e51417b0e2e5/framework/decode/vulkan_replay_consumer_base.cpp#L6307

Having external memory info here can affect replay, so it shouldn't be stripped unless it absolutely cannot be used.

zmike:
this alters behavior between capture and replay

fixes #2813

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -6679,23 +6679,6 @@ VulkanReplayConsumerBase::OverrideCreateImage(PFN_vkCreateImage                 
             has_external_format             = false;
         }
 
-        if (external_memory->handleTypes & VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD_BIT)
-        {
-            // If external memory exists and is for an Opaque FD, we need to strip out the structure
-            // since during replay we convert the allocate memory to a standard memory type.
-            if (external_memory->handleTypes == VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD_BIT)
-            {
-                GFXRECON_LOG_INFO("OverrideCreateImage removing VkExternalMemoryImageCreateInfo");
-                graphics::vulkan_struct_remove_pnext<VkExternalMemoryImageCreateInfo>(&modified_create_info);
-                external_memory = nullptr;
-            }
-            // Otherwise, just strip out the flag
-            else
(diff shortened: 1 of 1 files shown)


zmike:
I've been testing this for several months and it seems fine.

wzchu-lunarg:
This seems tough so far. It seems that as of https://github.com/LunarG/gfxreconstruct/commit/fed75c6c, the import FD info is dropped, as the FD information refers to memory that is being populated by some other process or API which we can't capture.

Then later in https://github.com/LunarG/gfxreconstruct/commit/7afd7538, the struct is stripped as otherwise it results in an external image having non-external memory. However, it is a little unclear why this **has** to be done @MarkY-LunarG, and doesn't seem like an obvious spec violation, although there might be some cases when the external memory prefers dedicated allocations.

However, it is difficult as it does seem to be a balance between writing good Vulkan but also replaying faithfully, in this case resulting in image creation more accurate to the original app if we don't strip it. Recapture would then also indicate some sort of usage of external as well. 

Even more difficult is that replaying faithfully here has even more cases. Currently, capture copies the external memory **contents** once. However, this external memory could be updated later by some process or API which we can't capture (unless we somehow figure out all the cases possible and capture those too) and therefore we lose aliasing of that memory.  

Furthermore, it seems that we only deal with the FD and AHB cases so far, which are relevant to Linux and Android. However, I believe there still might need to be a future effort to deal with the Windows and Mac cases, where we also need to me a similar decision of whether we want to treat this memory as external to be faithful to the app, or non-external as that is what our workaround to track and replay this memory has to do, and we don't have the processes and APIs that populate this memory.

wzchu-lunarg:
This might be a change we can look at later, but perhaps a more app faithful approach on top of not stripping would be to not drop the `VkImportMemoryFdInfoKHR` struct, and export a copy of that memory with a new handle to update into the `VkImportMemoryFdInfoKHR` struct.

zmike:
I don't particularly care about the interop. This is specifically for driver/layer interactions which only occur when allocating memory with this mechanic.

wzchu-lunarg:
Closing and bringing to https://github.com/LunarG/gfxreconstruct/pull/3134 with additional changes

PR #3127, opened 2026-07-23
present-override: compose multi-layer images into a single window

### LINKED ISSUE #3119: --present-override should always use a single window
For cases with multiple layers, this creates multiple per-layer windows for the present override, but this is problematic for replay environments without window management, as the surfaces may overlap and then be hidden. Rendering both layers onto the same window would solve this issue.

fabian-lunarg:
- tile array-layers into one window instead of one window per layer,
  window placement is a no-op on wayland/headless/display-wsi (https://github.com/LunarG/gfxreconstruct/issues/3119)
- near-square tile-grid, degenerates to a single row for 2 layers (VR)
- acquire/record/submit/present once per present, not once per array-layer
- clear swapchain-images once when the grid leaves unused tiles
- derive blit-destinations from the swapchain-extent instead of the current
  window-size, the two can drift apart
- re-create the swapchain only when that extent no longer matches the window,
  and request a resize only when the desired size changed. avoids re-creating
  on every frame when the window-manager grants a different size
- clamp the swapchain-extent against freshly queried surface-capabilities,
  they track the current window-size and cannot be cached
- add VK_IMAGE_USAGE_TRANSFER_SRC_BIT for --present-override images, blitting
  from them is otherwise invalid usage and yields undefined output
- bail out when the flag is still absent, e.g. transient attachments
- fix pre-synchronization2 srcStageMask on the initial layout-transition
- update --present-override help-text and USAGE_desktop_Vulkan.md

fixes #3119

### DIFF
--- a/framework/decode/vulkan_virtual_swapchain.cpp
+++ b/framework/decode/vulkan_virtual_swapchain.cpp
@@ -30,6 +30,7 @@
 #include "decode/vulkan_temporary_objects.h"
 #include "generated/generated_vulkan_enum_to_string.h"
 #include "vulkan/vulkan_core.h"
+#include <cmath>
 #include <sstream>
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
@@ -1387,6 +1388,15 @@ void VulkanVirtualSwapchain::PresentImageAdHoc(const VulkanDeviceInfo*          
         return;
     }
 
+    // we need to blit from the image, which requires VK_IMAGE_USAGE_TRANSFER_SRC_BIT.
+    if ((image_info->usage & VK_IMAGE_USAGE_TRANSFER_SRC_BIT) == 0)
+    {
+        GFXRECON_LOG_ERROR_ONCE("%s: image '%s' was created without VK_IMAGE_USAGE_TRANSFER_SRC_BIT, unable to present",
+                                __func__,
+                                image_info->debug_utils_name.c_str());
+        return;
+    }
+
     // retrieve device-context, create if necessary
     GFXRECON_ASSERT(device != VK_NULL_HANDLE);
     auto& ofb_data = adhoc_device_data_[device];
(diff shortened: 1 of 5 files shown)


framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -1413,33 +1420,57 @@ void VulkanVirtualSwapchain::PresentImageAdHoc(const VulkanDeviceInfo*
...
+    const uint32_t num_windows = kPresentOverrideSingleWindow ? 1 : layer_count;
+    const uint32_t num_cols =
+        kPresentOverrideSingleWindow ? static_cast<uint32_t>(std::ceil(std::sqrt(static_cast<float>(layer_count)))) : 1;
+    const uint32_t num_rows = kPresentOverrideSingleWindow ? (layer_count + num_cols - 1) / num_cols : 1;
+
+    const uint32_t window_width  = num_cols * tile_width;
+    const uint32_t window_height = num_rows * tile_height;
+
+    // a non-rectangular layer-count leaves unused tiles, those need to be cleared once to avoid showing garbage
+    const bool clear_unused_tiles = num_cols * num_rows != layer_count;
```
mikes-lunarg, line 1464: `clear_unused_tiles` will be `true` for multi-layer images in the legacy `kPresetOverrideSingleWindow==false` path because `num_cols` and `num_rows` are hardcoded to 1

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -1702,14 +1774,33 @@ void VulkanVirtualSwapchain::PresentImageAdHoc(const VulkanDeviceInfo*
...
-            blit_params.dst_layer  = 0;
-            blit_params.flip_axis  = { flip_x, flip_y, false };
-
-            ofb_data.copy_util->BlitImage(frame_data.command_buffer, blit_params);
+            blit_params.src_layout                                         = src_layout;
+            blit_params.dst_layout                                         = image_data.image_layout;
+            blit_params.dst_layer                                          = 0;
+            blit_params.flip_axis                                          = { flip_x, flip_y, false };
+
+            // derive tile-size from the actual window-size, it can slightly differ from the requested size
```
mikes-lunarg: Does the requested vs actual size mismatch mean that the swapchain will get re-created above on every frame?
```
        // Create/Re-create a swapchain if necessary
        if (window_width != current_window_size.width || window_height != current_window_size.height ||
            swapchain.handle == VK_NULL_HANDLE)
        {
```

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -30,11 +30,18 @@
...
+#include <cmath>
 #include <sstream>
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
+// Present multi-layer ad-hoc images (--present-override) tiled into a single window, instead of using one
+// window per array-layer. Placing multiple windows relies on Window::SetPosition, which is a no-op on wayland,
+// headless and display-wsi, so the layers can end up overlapping and hidden there.
+// NOTE: the previous behavior remains restorable, in case we decide to break out as CLI-option
```
mikes-lunarg: I get that this is a pretty obscure feature and still being developed, but we should probably pick a stable public behavior soon so we can remove the dead code and/or add a new flag.

mikes-lunarg (commented):
I tried it out with https://github.com/SaschaWillems/Vulkan/blob/master/examples/texturearray/texturearray.cpp and I'm getting an incomplete tiled image on replay:
```
./build/tools/replay/gfxrecon-replay texturearray.gfxr --present-override texturearray
```

Image

what is really odd is that it *is* correct with the debugger attached:

Image

Any ideas?

fabian-lunarg:
> I tried it out with https://github.com/SaschaWillems/Vulkan/blob/master/examples/texturearray/texturearray.cpp and I'm getting an incomplete tiled image on replay:
> 
> ```
> ./build/tools/replay/gfxrecon-replay texturearray.gfxr --present-override texturearray
> ```
...
> 
> what is really odd is that it _is_ correct with the debugger attached:
...
> Any ideas?

I compiled most recent samples from Sascha and had to add debug-util-names to the texturearray-sample via:

```
// set debug-utils name
VkDebugUtilsObjectNameInfoEXT name_info = {.sType =  VK_STRUCTURE_TYPE_DEBUG_UTILS_OBJECT_NAME_INFO_EXT};
name_info.objectType = VK_OBJECT_TYPE_IMAGE;
name_info.objectHandle = reinterpret_cast<uint64_t>(textureArray.image);
name_info.pObjectName = "texturearray";
auto set_debug_name_fn = reinterpret_cast<PFN_vkSetDebugUtilsObjectNameEXT>(
	vkGetInstanceProcAddr(instance, "vkSetDebugUtilsObjectNameEXT"));
VK_CHECK_RESULT(set_debug_name_fn(device, &name_info));
```
I checked debug/release/release-with-debug with/without the debugger but couldn't reproduce the imcomplete tiled image you posted.

some ideas:
- the sample is missing `TRANSFER_SRC_BIT` on `textureArray.image`. do you still see the issue when this is present?
-> adding the debug-name will cause GFXR to attempt blitting that image, so a missing usage-bit causes lot's of VVL-errors and 'could' explain the weird differences.
- which 'debugger' did you use? just gdb or a frame-debugger like Renderdoc?

edit:
- now pretty much convinced we ran into spec-violation / UB with Sascha's sample. 
-> my test-cases were basically all trimmed captures. for those we unconditionally slapped `VK_IMAGE_USAGE_TRANSFER_SRC_BIT` onto all images (we do this for trimming and dump-resources)
-> I adjusted the condition when to use/add `VK_IMAGE_USAGE_TRANSFER_SRC_BIT` to also include an active `--present-override`

framework/decode/vulkan_virtual_swapchain.cpp
fabian-lunarg: yes, you're right. this still needs some attention

framework/decode/vulkan_virtual_swapchain.cpp
fabian-lunarg: yes, agree. already having doubts we should actually keep the `constexpr` switch. 
cleaner to just drop the old behavior and `kPresetOverrideSingleWindow` entirely.

framework/decode/vulkan_virtual_swapchain.cpp
fabian-lunarg: corrected. we'll only recreate swapchains when required to

framework/decode/vulkan_virtual_swapchain.cpp
fabian-lunarg: `clear_unused_tiles` should now always behave correctly, legacy code-path doesn't exist anymore

framework/decode/vulkan_virtual_swapchain.cpp
fabian-lunarg: yes, agreed. actually I should have gone further than the `constexpr`-switch and just settle on the composed single window. the reason is we'll never have a guarantee the multi-window approach will work everywhere. 

-> this is now the case and we settle on the one-window-multiple-tiles behavior, dead code is removed.

mikes-lunarg:
> I compiled most recent samples from Sascha and had to add debug-util-names to the texturearray-sample via

Whoops, I forgot that I manually added the debug name from last time I tested this feature. My bad for not being more specific with the repro steps.

> which 'debugger' did you use? just gdb or a frame-debugger like Renderdoc?

Just gdb, I guess it alters the timing just enough to surface the issue?

> * now pretty much convinced we ran into spec-violation / UB with Sascha's sample.
>   -> my test-cases were basically all trimmed captures. for those we unconditionally slapped `VK_IMAGE_USAGE_TRANSFER_SRC_BIT` onto all images (we do this for trimming and dump-resources)
>   -> I adjusted the condition when to use/add `VK_IMAGE_USAGE_TRANSFER_SRC_BIT` to also include an active `--present-override`

Nice validation fix!

BUT...

I tracked my missing tiles issue down to a race condition in the XCB and X11 window implementations. Both of those wait for a notify event from the display server before they update their internal size so there is a chance to miss it if we call SetSize/GetSize in rapid succession: 

```c++
// position and size, tile multiple windows
swapchain.surface_info.window->SetSize(window_width, window_height); // <-- set to 768x768
swapchain.surface_info.window->SetPosition(window_offset_x, window_offset_y);
window_offset_x += window_width;

// NOTE: compensate sporadic 1px size-mismatches after window-resize (image_info->extent != window_size)
// mandatory to query window-size again
current_window_size = swapchain.surface_info.window->GetSize(); // <-- might still return old size 320x240

...

swapchain_create_info.imageExtent.width     = current_window_size.width; // <-- now the swapchain is created with the wrong size
swapchain_create_info.imageExtent.height    = current_window_size.height;

```

Even better, you actually fixed this issue in the latest version because of the new clamping behavior:
```c++
const VkExtent2D image_extent = { std::clamp(current_window_size.width, // <-- 320
                                             surface_capabilities.minImageExtent.width, // <-- 768
                                             surface_capabilities.maxImageExtent.width), // <-- 768
                                  std::clamp(current_window_size.height,
                                             surface_capabilities.minImageExtent.height,
                                             surface_capabilities.maxImageExtent.height) };
```

mikes-lunarg (approved):

fabian-lunarg:
> race condition in the XCB and X11

ok, it was something with XCB/X11 after all. that was the next bullet-item on the list of suspects.
happy the clamp against swapchain-bounds seems to cover it.

many thanks @mikes-lunarg!

PR #3134, opened 2026-07-24
Export external memory imports during replay

### LINKED ISSUE #2813: stop stripping VkExternalMemoryImageCreateInfo in replay

wzchu-lunarg:
Fixes https://github.com/LunarG/gfxreconstruct/issues/2813. This PR is a continuation of https://github.com/LunarG/gfxreconstruct/pull/3125.

Previously, external memory was replayed by tracking the external memory's contents, and creating a new allocation with those contents. This leads to dropping the external memory's FD struct `VkImportMemoryFdInfoKHR` https://github.com/LunarG/gfxreconstruct/commit/fed75c6c as the imported FD is meaningless during replay. 

Furthermore, since we dropped this info, we needed to strip corresponding flags and structs in `OverrideCreateImage` and `OverrideCreateBuffer` https://github.com/LunarG/gfxreconstruct/commit/7afd7538.

Instead, we can keep this external memory information by creating a new export and new FD with the same contents which we can modify `VkImportMemoryFdInfoKHR` with, and keeps the calls more faithful to the original app's. We then also don't need to strip info from `VkExternalMemoryBufferCreateInfo` which addresses https://github.com/LunarG/gfxreconstruct/issues/2813.


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5780,8 +5780,63 @@ VkResult VulkanReplayConsumerBase::OverrideAllocateMemory(
         bool     uses_import_memory     = false;
         uint64_t opaque_address         = 0;
 
-        // FD is not available at replay time
-        graphics::vulkan_struct_remove_pnext<VkImportMemoryFdInfoKHR>(modified_allocate_info);
+        // Since the file descriptor in VkImportMemoryFdInfoKHR is only valid in the capture process
+        // we create a new FD at replay by allocating an exportable allocation and importing that FD
+        // instead.
+        VkDeviceMemory external_fd_backing_memory = VK_NULL_HANDLE;
+        int            replacement_import_fd      = -1;
```
MarkY-LunarG, line 5823: Nit: use `int32_t` just to be consistent.

bradgrantham-lunarg: `int` is the type used in `vkGetMemoryFdKHR` and `VkImportMemoryFdInfoKHR`; why isn't that consistent?

MarkY-LunarG: I guess so, it's just we've used the other descriptive types in most situations throughout the code.  And `int` is effectively `int32_t`.  I did say  it was a nit.

bradgrantham-lunarg: I agree - in the absence of a base type in the captured API, we should use descriptive types.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5780,8 +5780,70 @@ VkResult VulkanReplayConsumerBase::OverrideAllocateMemory(
...
+        auto* import_fd_info = graphics::vulkan_struct_get_pnext<VkImportMemoryFdInfoKHR>(modified_allocate_info);
+
+        if (import_fd_info != nullptr && UseAddressReplacement(device_info))
```
antonio-lunarg: Maybe we can use `!CanPreserveExternalMemory(..)` here to be more descriptive.

antonio-lunarg (approved):

wzchu-lunarg:
Supporting external memory with reallocations should be possible, but likely a bit lengthy so will be saving that for a later PR.

wzchu-lunarg:
Failling android extended tests http://jenkins:8080/job/gfxr-pipelines/job/gfxr-extended-manual/213/

MarkY-LunarG (approved):

PR #3136, opened 2026-07-24
Cleanup decoder DispatchInitTensorCommand

MarkY-LunarG:
Remove the need for every decoder to have to implement the DispatchInitTensorCommand prototype.

### DIFF
--- a/framework/decode/api_decoder.h
+++ b/framework/decode/api_decoder.h
@@ -183,7 +183,8 @@ class ApiDecoder
                                            format::HandleId device_id,
                                            format::HandleId tensor_id,
                                            uint64_t         data_size,
-                                           const uint8_t*   data) = 0;
+                                           const uint8_t*   data)
+    {}
 
     virtual void DispatchInitImageCommand(format::ThreadId             thread_id,
                                           format::HandleId             device_id,
--- a/framework/decode/custom_ags_decoder.h
+++ b/framework/decode/custom_ags_decoder.h
@@ -152,12 +152,6 @@ class AgsDecoder : public ApiDecoder
                                            uint64_t         data_size,
                                            const uint8_t*   data) override
     {}
-    void DispatchInitTensorCommand(format::ThreadId thread_id,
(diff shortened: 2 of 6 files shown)


bradgrantham-lunarg (changes requested):
Changing this hook from pure virtual to a default no-op removes compile-time enforcement for decoders that are expected to handle tensor initialization. 

This feels like a change intended to make it convenient to implement new Decoders but at the cost of potentially missing an important override.  It also makes this decoder method inconsistent with the others.

bradgrantham-lunarg:
Probably this should evolve into a conversation about API-specific metadata, for which in general we haven't made concessions.  You might argue the TensorCommand is Vulkan-specific, and therefore shouldn't show up in the general Decoder but should be in VulkanDecoder and then we should have some way for other Decoders to just skip the block (or warn e.g. in the case of combined OpenXR + Vulkan capture).

bradgrantham-lunarg:
The derived Decoder is only called if it returns `SupportsMetaDataId` for the block.  So this is really only to get around a compilation error, is that right @MarkY-LunarG ?  In which case maybe an middle ground would be a FATAL in the base implementation.  That's a little unsatisfactory, because it opens the possibility that a CaptureManager for a new API writes the block (e.g. new OpenGL implementation writes a CreateHardwareBufferCommand on Android) , but then the author of Decoder for the new API forgets to implement the derived block handler (e.g. tested only on Desktop or in Android apps that don't use AHB so CHBC is never encoded), and that means that there could be a surprise fatal error on decode later when it *does* get encoded.  That's less good than just not compiling but at least would cover some ground.

fabian-lunarg:
was there another reason besides cleanup? 
I understand it's verbose and slightly annoying to spell out the implementations in all subclasses.
if in doubt I'd prefer explicit implementation over an implicitly derived one, but in this case have no strong opinion against it.
api_decoder already has a mishmash of pure virtual and stubs, so that lines up either way.

PR #3138, opened 2026-07-27
Remove frame number assert in block_processor

bartosz-muszarski-arm:
Currently block processor asserts on frame number mismatch between internal counter and frame number provided by frame end marker. This assert would make sense for replayer, but block processor is a core module used by all other tools, including convert and optimize. This makes it impossible to do any kind of processing on traces that captured unexpected application behavior. Change downgrades this assert to an error log allowing the execution to continue.

### DIFF
--- a/framework/decode/block_processor.cpp
+++ b/framework/decode/block_processor.cpp
@@ -344,7 +344,13 @@ bool BlockProcessor::ProcessFrameDelimiter(format::ApiCallId call_id)
 
 bool BlockProcessor::ProcessFrameDelimiter(const FrameEndMarkerArgs& end_frame)
 {
-    GFXRECON_ASSERT((!capture_uses_frame_markers_) || (frame_number_ == (end_frame.frame_number - first_frame_)));
+    if (capture_uses_frame_markers_ && (frame_number_ != (end_frame.frame_number - first_frame_)))
+    {
+        GFXRECON_LOG_ERROR("Frame end marker frame number (%" PRIu64 ") does not match current frame number (%" PRIu64
+                           ")",
+                           end_frame.frame_number - first_frame_,
+                           frame_number_);
+    }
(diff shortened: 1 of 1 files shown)


framework/decode/block_processor.cpp
```diff
@@ -344,7 +344,13 @@ bool BlockProcessor::ProcessFrameDelimiter(format::ApiCallId call_id)
 
 bool BlockProcessor::ProcessFrameDelimiter(const FrameEndMarkerArgs& end_frame)
 {
-    GFXRECON_ASSERT((!capture_uses_frame_markers_) || (frame_number_ == (end_frame.frame_number - first_frame_)));
+    if (capture_uses_frame_markers_ && (frame_number_ != (end_frame.frame_number - first_frame_)))
```
jzulauf-lunarg, line 347: Why would this be just an error as opposed to a logical fault in frame processing?  Is there otherwise valid content where this is true?

bartosz-muszarski-arm: This assert can be hit if captured app emitted frame boundaries from multiple threads at the same time. It is more common on android (and that's where we found it) but it is not android-specific.
The root cause is that capture is using shared locks by default. That works for most calls but not frame boundaries, since they increment global current_frame_ in EndFrame().
Typically we expect this sequence:
QueuePresent
FrameEndMarker(N)
...
QueuePresent
FrameEndMarker(N+1)
...
But because the lock in vkQueueSubmit capture is by default shared instead of exclusive the sequence may be altered and trip the assert later in BlockProcessor.
That is a separate issue this change does not fix - we already have such traces and need to be able to process them. With this assert in place we can't even gfxrecon-convert such traces.

jzulauf-lunarg: That's an encoding side bug, certainly, (is there an issue opened for it?), but it means this isn't an invariant to assert on, since the input stream can violate it.

jzulauf-lunarg (approved):

PR #3139, opened 2026-07-27
Add more reserved MetaDataType values.

MarkY-LunarG:


jzulauf-lunarg (approved):
LGTM

Commentary describing for whom or what the reservation is might be useful for maintainers.

bradgrantham-lunarg:
Needs special-casing in framework/decode/block_parser.cpp with the other reserved blocks.  Probably worth making that a compile-time or at least statically-initialized `std::set`.

MarkY-LunarG:
@bradgrantham-lunarg / @jzulauf-lunarg : Here are the changes to support overriding the meta-data parsing from an override area.  This allows an external implementation of reserved blocks so that they can be processed by  the upstream implementation.  Let me know what you think.

I tried to not make too much public, which is why there's an accessor struct helper.  I could remove that, but I would need to make each of those accessor functions public in the BlockParser itself.

framework/decode/block_parser.cpp
```diff
@@ -1718,6 +1719,20 @@ ParsedBlock& BlockParser::ParseMetaData(BlockBuffer& block_buffer)
     }
     else
     {
+        // The built-in parsers do not know this meta-data type. Give an extended parser for the type
+        // -- if this build defines one -- a chance to parse it before skipping the block.
+        for (const ExtendedMetaDataEntry& entry : kExtendedMetaDataParsers)
```
jzulauf-lunarg: this really should be a std::map<Extended...>::find instead of a linear walk,

Actual maybe encapsulated away from the BlockParser so. it doesn't need to know the details, with a registered parser object (some noop registered by default)

```
block = extended_meta_data_parser_->parse(this, id, block_buffer); 
if block return block

```

framework/decode/block_parser.h
```diff
@@ -183,14 +183,18 @@ class BlockParser
     }
 
   private:
+    // The parsing primitives below stay private; the parsers for extended meta-data types reach them
+    // through this gateway rather than through a widened public interface.
+    friend struct ExtendedMetaDataAccess;
```
jzulauf-lunarg: I think I'd rather have the BlockParser share some "interface" object with extended parsers than friend them.  The interface object would be a friend of course, but it would expose a known, restricted set and inform maintainers what within the BlockParser needs to be conserved

Also, do need commentary around the GetParserInterface call that BlockParser is operating on the BlockProcessing thread which may not be the block disspatch thread.

jzulauf-lunarg (changes requested):
Some encapsulation/information boundaries to cleanup.

After the async processing cleanup, I'm somewhat sensitive to this.

framework/decode/block_parser.cpp
MarkY-LunarG: There aren't that many new meta data fields (< 10).  So a map actually is more expensive than an array.  I do agree about encapsulating this a little further.

framework/decode/block_parser.h
MarkY-LunarG: Done.

framework/decode/block_parser.h
```diff
@@ -231,6 +241,80 @@ class BlockParser
...
+// ExtendedMetaDataInterface
+//
+// The restricted view of a BlockParser given to the parsers for extended meta-data types (see
+// block_parser_meta_data.h). Those parsers never see the BlockParser itself, so this class -- and
+// not BlockParser's public interface -- is what an extended parser can depend on, and what has to be
+// conserved for them.
+//
+// Only BlockParser can construct one, via GetParserInterface().
+// -----------------------------------------------------------------------------
+class ExtendedMetaDataInterface
```
jzulauf-lunarg: This should just be the BlockParser::Interface and not specific to meta data.  One can imagine other block parsing extension needing the restricted interface.

framework/decode/block_parser.cpp
MarkY-LunarG: Actually, this has a further problem with this change:

```
Our extended parser for kReserved41 gets called and must hand back a ParsedBlock. That block carries its payload in DispatchArgs (gfxreconstruct/framework/decode/api_payload.h:846), which is a fixed std::variant — and
  whichever alternative we pick determines which ApiDecoder virtual gets called, via DispatchTraits<Args>::kDecoderMethod.
```

Since the specific metadata type is not defined here, we will need some kind of generic handler for the decoding.  I'm working that out now.

framework/decode/block_parser.h
```diff
@@ -182,15 +183,24 @@ class BlockParser
         return EmplaceBlock(args);
     }
 
+    // Returns the restricted interface that parsers for extended meta-data types (see
+    // block_parser_meta_data.h) use to build blocks.
+    ExtendedMetaDataInterface GetParserInterface() noexcept;
```
jzulauf-lunarg: could be const return --- should be private -- BlockParser grants access....

Need to look at how much of BlockParser should be *made* private... will send a patch of minimal public interface.

would no_discard be useful?  It's pointless unless ref'd.

jzulauf-lunarg (changes requested):
Couple small changes.

Will send you the patch for the suggested privitization.

framework/decode/block_parser.h
MarkY-LunarG: Renamed it `ExtendedParserInterface`.

framework/decode/block_parser.h
MarkY-LunarG: Returning const.

MarkY-LunarG: This can't be const.  I tried switching it and it failed to work properly. This is because the methods inside the `ExtendedMetaDataInterface` (Now the `ExtendedParserInterface`) modify the contents (for example the various "Emplace", "ReadParameterBuffer" and "MakeCompressibleParsedBlock") by doing things such as allocating out of the current BlockBatch.

MarkY-LunarG:
@jzulauf-lunarg , this change is still waiting for your final review.

framework/decode/api_payload.h
```diff
@@ -544,6 +544,27 @@ struct InitializeMetaArgs
...
+// meta-data parser (see block_parser_meta_data.h). The block's parameter data is passed through
+// undecoded, along with the MetaDataId, so the decoder that claims the id can interpret it with
+// its own API-specific decoders. Unlike the payloads above, meta_data_id is part of GetTuple: it
+// both selects the decoder (via SupportsMetaDataId) and tells that decoder which extended type it
+// has been handed.
+struct ExtendedMetaDataArgs
+{
+    format::MetaDataId meta_data_id;
+
+    const uint8_t* data;
```
bradgrantham-lunarg, line 557: I feel like this should be a shared_ptr unless the contract is that it's a region of data in another block in which case I imagine it should be a span.

bradgrantham-lunarg: I see this is similar to the other structs, maybe we can discuss in our meeting

bradgrantham-lunarg:
As we discussed offline, I understand the reasoning behind this change.  After reading through it think there may be a couple of things we will need to address before it becomes default.  Okay for it to go in with #ifdef around it with a comment noting it is experimental.  In the meantime, code still has to detect reserved IDs and reject them if not #ifdef'd and I think that should look like current code morphed to a statically-initialized `std::unordered_set` with the old reserved and new reserved types in it

MarkY-LunarG:
#define in place, approve @jzulauf-lunarg?

jzulauf-lunarg (approved):
LGTM

PR #3140, opened 2026-07-27
Support optimize for frameless captures; add missing vkQueueSubmit2KHR handler

pemgithub:
Optimize requires at least one frame, so frameless compute workloads are rejected outright. Accept a zero-frame capture that has application command-buffer submissions, excluding synthetic ones from state restoration.

vkQueueSubmit2KHR had no handler, so its command buffers never reached ProcessUserSubmission and resources used only by them could be stripped, including in framed captures. Add the missing override.

Fixes android-graphics/gfxr-backlog#147.


framework/decode/vulkan_referenced_resource_consumer_base.cpp
```diff
@@ -59,6 +65,37 @@ void VulkanReferencedResourceConsumerBase::Process_vkQueueSubmit2(const ApiCallI
...
+
+            for (size_t j = 0; j < command_buffer_count; ++j)
+            {
+                table_.ProcessUserSubmission(command_buffers[j].commandBuffer);
+            }
+        }
+    }
+}
+
+void VulkanReferencedResourceConsumerBase::Process_vkQueueSubmit2KHR(const ApiCallInfo&     call_info,
```
fabian-lunarg, line 61: I agree with the override, but we shouldn't copy paste that.
a private, templated routine, shared by both implementations
would be cleaner and avoid duplication.

fabian-lunarg (approved):
left one comment about DRY/duplication, which should be consolidated.
approved otherwise

bradgrantham-lunarg:
@pemgithub - if you can fix that deduplication we can approve and merge.

pemgithub:
a84638d732890dbf9943a6aee3870648216ad78e

pemgithub:
The merge-queue Jenkins run failed in Android R64 (ci/runTestAndroid.sh, exit code 2), while all PR-head checks are passing. @fabian-lunarg, looks like it needs to be requeued.

PR #3141, opened 2026-07-28
rebind: Fixes for Aliasing

fabian-lunarg:
- VMA turns both requiresDedicatedAllocation, prefersDedicatedAllocation
   into a dedicated block, but also in cases when neither flag is set.
-> check with vmaGetAllocationInfo2().dedicatedMemory, reject sharing
- don't alias on page-guard inflated requirement sizes
- Additional check for range-overlap at replay-time, reject sharing

follow-up for https://github.com/LunarG/gfxreconstruct/pull/3100
(introduced reproduction of memory-aliasing when using `-m rebind`)


locke-lunarg (approved):
LGTM!

PR #3142, opened 2026-07-28
Add a screenshot hotkey trigger for DX12 capture

davidd-lunarg:
Also add missing documentation for DX12 capture-time screenshot options.


framework/encode/capture_manager.cpp
```diff
@@ -830,6 +832,20 @@ bool CommonCaptureManager::IsTrimHotkeyPressed()
...
 
+bool CommonCaptureManager::IsScreenshotHotkeyPressed()
+{
+    if (screenshot_key_.empty())
+    {
+        return false;
+    }
+
+    // Return true when GetKeyState() transitions from false to true
+    bool hotkey_state                 = keyboard_.GetKeyState(screenshot_key_);
```
jzulauf-lunarg, line 843: The current logic can miss key presses quicker than two polling loops (needs an extra iterator for a explicit "up" detection - note 1, which would require some reworking of util:Keyboard to reduce to one loop, but that behavior matches other keypress operations logic in any case.

note1 -- Since the 1's bit detects "press since previous call to `GetAsyncKeyState`, a particularly un;ucky set of key presses timed just faster than the polling rate would never return an "up" state.

jzulauf-lunarg (commented):
LGTM -- noting common behavioral quirk with extant keypress operations

PR #3143, opened 2026-07-28
Reduce ApiDecoder interface to single templated function

jzulauf-lunarg:
A single overloaded Dispatch template... to simplify callers work.


antonio-lunarg (approved):

jzulauf-lunarg:
Marking this WIP to avoid API churn...

PR #3145, opened 2026-07-29
tocpp: Various fixes

CreativeCylon:
The last three commits mirror some fixes that were done to the replay code. All other commits are tocpp specific fixes.

These were enough to get the generated code to compile, and for a couple of frames to run but there are still issues since the capture I was trying this on ends up hitting a gpu fault after a few frames while replay works fine.


CreativeCylon:
Btw, just for reference, I also needed to address https://github.com/LunarG/gfxreconstruct/issues/3146 in the generated code to get the capture to not segfault. I have not tried to fix it in tocpp though.

PR #3148, opened 2026-07-29
[Frame Looping] Handle VK_EXT_frame_boundary

nickdriscoll-lunarg:
At present, frame looping assumes frames will be delimited by `vkQueuePresentKHR`. This PR adds support for `VK_EXT_frame_boundary`, adding checks to the other queue operations so that end-of-frame cleanup work happens when a frame boundary struct defines the end of a frame.

### DIFF
--- a/framework/decode/vulkan_replay_frame_loop_consumer.cpp
+++ b/framework/decode/vulkan_replay_frame_loop_consumer.cpp
@@ -467,6 +467,29 @@ void VulkanReplayFrameLoopConsumer::Process_vkMapMemory(const ApiCallInfo& call_
     VulkanReplayConsumer::Process_vkMapMemory(call_info, args);
 }
 
+void VulkanReplayFrameLoopConsumer::Process_vkQueueSubmit(const ApiCallInfo& call_info, args::QueueSubmit& args)
+{
+    VulkanReplayConsumer::Process_vkQueueSubmit(call_info, args);
+
+    auto frame_boundary = GetPNextMetaStruct<Decoded_VkFrameBoundaryEXT>(args.pSubmits.GetMetaStructPointer()->pNext);
+    if (frame_boundary != nullptr)
+    {
+        // This submit is being used as a frame boundary
+        CommonObjectInfoTable& table      = GetObjectInfoTable();
+        VulkanQueueInfo*       queue_info = table.GetVkQueueInfo(args.queue);
+        VkDevice               device     = queue_info->parent;
+        GFXRECON_ASSERT(device != 0);
+        const graphics::VulkanDeviceTable* device_table = GetDeviceTable(device);
+        GFXRECON_ASSERT(device_table != nullptr);
+
+        GFXRECON_LOG_DEBUG("Waiting for device to idle...");
+        VkResult result = device_table->DeviceWaitIdle(device);
+        CHECK_VK_RESULT(result, "vkDeviceWaitIdle");
+
+        FixupDeviceFences(queue_info->parent_id, args.queue);
+    }
+}
+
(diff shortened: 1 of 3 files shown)


davidlunarg:
This commit seems to allow my test program to run successfully. However, I'm still enhancing my test program, so will be testing the commit some more over the next couple of days.

davidlunarg (approved):
LGTM. But Antonio should also review.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -467,6 +467,70 @@ void VulkanReplayFrameLoopConsumer::Process_vkMapMemory(const ApiCallInfo& call_
...
+            CHECK_VK_RESULT(result, "vkDeviceWaitIdle");
+
+            FixupDeviceFences(queue_info->parent_id, queue);
+        }
+    }
+}
+
+void VulkanReplayFrameLoopConsumer::Process_vkQueueBindSparse(const ApiCallInfo& call_info, args::QueueBindSparse& args)
+{
+    VulkanReplayConsumer::Process_vkQueueBindSparse(call_info, args);
```
antonio-lunarg: Should this call `VulkanReplayFrameLoopConsumerBase::`?

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -467,6 +467,70 @@ void VulkanReplayFrameLoopConsumer::Process_vkMapMemory(const ApiCallInfo& call_
...
+            VkDevice               device     = queue_info->parent;
+            GFXRECON_ASSERT(device != 0);
+            const graphics::VulkanDeviceTable* device_table = GetDeviceTable(device);
+            GFXRECON_ASSERT(device_table != nullptr);
+
+            GFXRECON_LOG_DEBUG("Waiting for device to idle...");
+            VkResult result = device_table->DeviceWaitIdle(device);
+            CHECK_VK_RESULT(result, "vkDeviceWaitIdle");
+
+            FixupDeviceFences(queue_info->parent_id, queue);
```
antonio-lunarg: Need to call `FixupDeviceEvents` as well? I suppose we can introduce a "Fixup" function to centralize these calls (soon we'll need to add a FixupDeviceSemaphores ..)

antonio-lunarg (changes requested):

framework/decode/vulkan_replay_frame_loop_consumer.cpp
nickdriscoll-lunarg: I added a `FixupDeviceObjects` function that calls both. We can stick a future `FixupDeviceSemaphores` call in there as well.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
nickdriscoll-lunarg: Unlike the other two added queue ops, yes! Thank you.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -487,11 +487,17 @@ void VulkanReplayFrameLoopConsumer::FrameBoundaryEndOfFrame(format::HandleId que
...
             CHECK_VK_RESULT(result, "vkDeviceWaitIdle");
 
-            FixupDeviceFences(queue_info->parent_id, queue);
+            FixupDeviceObjects(queue_info->parent_id, queue);
         }
     }
 }
 
+void VulkanReplayFrameLoopConsumer::FixupDeviceObjects(format::HandleId device, format::HandleId queue)
+{
```
antonio-lunarg, line 496: Fixup should not happen if not looping or if it's the last iteration. That can be an early return here.

antonio-lunarg (approved):
Once last comment is addressed, LGTM

PR #3150, opened 2026-07-29
file_processor: Add filename accessor

MarkY-LunarG:
Some of the components in tools don't have easy access to the filename that is initialized into the FileProcessor. So, simply add a filename accessor for them.

### DIFF
--- a/framework/decode/file_processor.h
+++ b/framework/decode/file_processor.h
@@ -96,6 +96,7 @@ class FileProcessor
     }
 
     bool Initialize(const std::string& filename);
+    const std::string& GetFilename() const { return filename_; }
 
     // Called implicitly by ProcessAllFrames() and directly by Application::Run()
     // Must be called at most once after Initialize() and before ProcessNextFrame()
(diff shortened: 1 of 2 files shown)


MarkY-LunarG:
Ping @jzulauf-lunarg

framework/decode/file_processor.h
```diff
@@ -96,6 +96,7 @@ class FileProcessor
     }
 
     bool Initialize(const std::string& filename);
+    const std::string& GetFilename() const { return filename_; }
```
jzulauf-lunarg, line 99: This will give you the name that started the replay.  If you want the name of the file currently being dispatched, that would have to be signalled from the BlockProcessor (potentially on a different thread).

But if all you want is the top level filename this should be fine.

jzulauf-lunarg (approved):
LGTM, not sure why the CI cases are failing.  Seems unrelated to the change.

bradgrantham-lunarg:
I would think the tools all know the filename; can you tell me why it needs to be passed back from FileProcessor?  (Also in the case of "asset files" what is currently being processed may also change from moment to moment.)

MarkY-LunarG:
>I would think the tools all know the filename; can you tell me why it needs to be passed back from FileProcessor? (Also in the case of "asset files" what is currently being processed may also change from moment to moment.)

I need the filename in a derived version of the info and extract tools.  It's either put a clean accessor in the file_processor (this PR) or update the info_feature and extract_feature classes to accept a filename input into their RegisterDecodeComponents or Intialize unctions.  This seems cleaner and since it's only an accessor that will get  compiled out if not used, more efficient.

framework/decode/file_processor.h
MarkY-LunarG: That's all I want.  It's for knowing what the file name is and where it is located to reference additional support files.

bradgrantham-lunarg:
> > I would think the tools all know the filename; can you tell me why it needs to be passed back from FileProcessor? (Also in the case of "asset files" what is currently being processed may also change from moment to moment.)
> 
> I need the filename in a derived version of the info and extract tools. It's either put a clean accessor in the file_processor (this PR) or update the info_feature and extract_feature classes to accept a filename input into their RegisterDecodeComponents or Intialize unctions. This seems cleaner and since it's only an accessor that will get compiled out if not used, more efficient.

What happens in the case of asset files?  What happens in the case of remote replay?

If it's for an extract feature, it seems cleaner to provide that information to the extract feature implementation.

bradgrantham-lunarg:
> This seems cleaner and since it's only an accessor that will get compiled out if not used, more efficient.

It will make an extra copy of the filename.  It's not that big a deal but that won't "get compiled out if not used".

Higher in the callchain should always be the source of truth if possible.  Lower in the callchain should retain as little state as possible.  If an extraction tool invokes FileProcessor, why doesn't it know the filename, create a base dir, etc?

PR #3151, opened 2026-07-29
Cmake use FetchContent for compression libraries

wzchu-lunarg:


bradgrantham-lunarg:
@johnzupin  @richard-lunarg @KarenGhavam-lunarG This would change GFXR's compression library policy from "it's whatever binary we checked in" to "build from pinned source revision by default".

Notably downstream distro packaging may balk at not using the system libraries, but this shouldn't be any worse than before, just requires `GFXRECON_COMPRESSION_FROM_SOURCE=OFF` to build with the system.

Take a moment please to consider if this affects the SDK build or packaging in downstream distros.

per-mathisen-arm:
Having worked with packagers before, I think they would be fine as long as they have a switch to disable included libraries and use system ones instead. I (and I assume package maintainers) would prefer that it tries to use system libraries first and only falls back to fetching its own library sources, as then it tries to do the arguably correct thing first and (possibly significantly) reduces initial compile time.

jzulauf-lunarg:
CI testing of a version change should catch interface changes at compile time, and hopefully cover any behavioral changes as well.

antonio-lunarg:
This means ci jobs can be updated, dropping lines like `apt install liblz4-dev/libzstd-dev`, right?

cmake/FindLZ4.cmake
```diff
@@ -23,6 +23,48 @@
 # ^^^^^
 #
 # The ``LZ4_ROOT`` value may be set to tell this module where to look.
+if(TARGET LZ4::LZ4)
+    set(LZ4_FOUND TRUE)
+    return()
+endif()
+
+option(GFXRECON_COMPRESSION_FROM_SOURCE
```
antonio-lunarg, line 31: Do we want to include CI steps for ensuring gfxreconstruct builds with both `GFXRECON_COMPRESSION_FROM_SOURCE=ON` and `GFXRECON_COMPRESSION_FROM_SOURCE=OFF`?

bradgrantham-lunarg: We'd be testing `GFXRECON_COMPRESSION_FROM_SOURCE=OFF` against whatever is on our CI machines, and it wouldn't tell us much about what would happen on any 3rd party builder's machine, but it would help to ensure that some random CMakeLists.txt change doesn't break the build when that option is `OFF`.  I do like the idea of maybe another Pipeline run that just builds with `OFF` without running VulkanTests.  Can @wzchu-lunarg  add that as a subsequent PR @antonio-lunarg so we can get this in?

antonio-lunarg: Sure

wzchu-lunarg:
Was under the assumption that we would get this approved before a subsequent PR with the `GFXRECON_COMPRESSION_FROM_SOURCE=OFF`, however made a [draft](https://github.com/LunarG/gfxreconstruct/pull/3241) so maybe we can get this through soon.

richard-lunarg:
This does affect SDK builds for Windows ARM. These packages last I checked did not support building a combined binary (Intel/ARM), so I have to build both sets and combine them. I then checked them into the repository for ARM builds. Whenever we update the version, I either need to leave it as is, or do a manual build to update them.

cmake/FindLZ4.cmake
charles-lunarg:
```suggestion
endif()
```
This code only exists because we committed the binaries and headers into the repository. Since they are just raw headers and precompiled binaries, we have to do the work of giving CMake all the necessary info. Now that we use `find_package()`/`FetchContent`, that info is handled by the library itself. 

Thus we can safely remove all this now dead code.

cmake/FindZLIB.cmake
```diff
@@ -0,0 +1,87 @@
...
+
+    add_library(ZLIB::ZLIB ALIAS zlibstatic)
+
+    set(ZLIB_FOUND TRUE)
+    set(ZLIB_VERSION "${GFXRECON_ZLIB_VERSION}")
+    set(ZLIB_VERSION_STRING "${GFXRECON_ZLIB_VERSION}")
+    set(ZLIB_LIBRARIES ZLIB::ZLIB)
+    set(ZLIB_INCLUDE_DIRS "${zlib_SOURCE_DIR}" "${zlib_BINARY_DIR}")
+
+    return()
```
charles-lunarg:
```suggestion
```
Can delete this code.

cmake/FindZLIB.cmake
```diff
@@ -0,0 +1,87 @@
...
+endif()
+
+include("${CMAKE_ROOT}/Modules/FindZLIB.cmake")
```
charles-lunarg: Where is this file defined? I don't see any `Modules` folder in our project nor in any of the dependencies.

Is this just calling the CMake installed `FindZLIB.cmake`? Because that is what `find_package(ZLIB)` is doing, we shouldn't be including that file ourselves.

cmake/FindZSTD.cmake
```diff
@@ -23,6 +23,64 @@
...
+    endif()
+
+    add_library(ZSTD::ZSTD ALIAS libzstd_static)
+
+    set(ZSTD_FOUND TRUE)
+    set(ZSTD_VERSION "${GFXRECON_ZSTD_VERSION}")
+    set(ZSTD_LIBRARIES ZSTD::ZSTD)
+    set(ZSTD_INCLUDE_DIRS "${zstd_SOURCE_DIR}/lib")
+
+    return()
```
charles-lunarg:
```suggestion
```

cmake/FindZLIB.cmake
```diff
@@ -0,0 +1,87 @@
...
+#
+#  ZLIB_FOUND          : True if zlib was found.
+#  ZLIB_INCLUDE_DIRS   : The location of the zlib header files.
+#  ZLIB_LIBRARIES      : List of the zlib libraries.
+#  ZLIB_VERSION        : The version of zlib that was found.
+#
+# Hints
+# ^^^^^
+#
+# The ``ZLIB_ROOT`` value may be set to tell this module where to look.
```
charles-lunarg: Stale documentation

cmake/FindLZ4.cmake
charles-lunarg: Stale documentation. Even without my proposed changes, this is stale because we are now using `FetchContent()`/`find_package()` to get the dependency, not creating an imported target.

cmake/FindZSTD.cmake
```diff
@@ -23,6 +23,64 @@
 # ^^^^^
 #
 # The ``ZSTD_ROOT`` value may be set to tell this module where to look.
```
charles-lunarg: More stale documentation

charles-lunarg (commented):
Its a good start - just needs some cleanup work to remove the code and documentation that became stale with the changes.

cmake/FindLZ4.cmake
wzchu-lunarg: That sounds good. Mainly left the preexisting code as a backup if `GFXRECON_COMPRESSION_FROM_SOURCE` is off. However, that wouldn't matter if we do decide to delete the precompiled binaries anyways. In that case, I'll remove this option as well.

cmake/FindZLIB.cmake
wzchu-lunarg: It is indeed just calling the CMake installed one, but mainly for the case of if `GFXRECON_COMPRESSION_FROM_SOURCE` was off.  It seemed to me it would never end up doing the logic in the CMake installed one as it shares the same name as our find ZLIB, but since I will remove the option, this should be fine to remove anyways.

charles-lunarg (approved):
Noticed something in the top level CMakeLists.txt - we add the precompiled libraries to CMAKE_PREFIX_PATH, but we are deleting those libraries (except for windows, which still has some). Meaning, we can delete those calls to `set(CMAKE_PREFIX_PATH ${CMAKE_PREFIX_PATH} ...)`

I think there are only three places which need removal, linked here.
https://github.com/LunarG/gfxreconstruct/blob/6cccab697579f2c23e559770ebb10d93b693e97b/CMakeLists.txt#L307
https://github.com/LunarG/gfxreconstruct/blob/6cccab697579f2c23e559770ebb10d93b693e97b/CMakeLists.txt#L311
https://github.com/LunarG/gfxreconstruct/blob/6cccab697579f2c23e559770ebb10d93b693e97b/CMakeLists.txt#L319

Other than that I think its good to go.

Note that we likely will need to amend building to allow using system libraries instead of fetch content. This could be as simple as a build option, like `GFXRECON_COMPRESSION_FROM_SYSTEM` or a similar name, that just uses `find_package()`.
E.G
```cmake
if (GFXRECON_COMPRESSION_FROM_SYSTEM)
    find_package(LZ4 REQUIRED) 
    return()
endif()
<rest of the FindLZ4.cmake file>
```

bradgrantham-lunarg (approved):

PR #3152, opened 2026-07-30
Detect Vulkan API on `vkCreateInstance` instead of `vkCreateDevice`

marius-pelegrin-arm:
The entry point to any Vulkan program is instance creation, not device creation, and a program that would create an instance, query supported physical devices and their properties/other instance level calls without ever creating a device would still be a valid Vulkan program and could still benefit from some sort of optimization.


fabian-lunarg:
I agree with the description and looked straight-forward first.
just two potential issues, even though it's semantically correct imo:

- failed instance creation now counts as detection -> should check the `VkResult`
- more observation / open question: Mixed-API captures could show a changed behavior down the line. 
  e.g. a D3D12 capture that also happened to load a `VkInstance` (but nothing else) would behave differently in `gfxrecon-optimize` (runs every feature whose `ShouldRun()` is true, each (over?)writing the same output_filename)

fabian-lunarg (approved):

PR #3153, opened 2026-07-30
Remove duplicate fields and track device earlier

naomipappe:
1. Remove the fields from BufferWrapper that are already present in superclass.
2. In AssetWrappers (BufferWrapper and ImageWrapper) track the parent device on creation (hence the rename to just "device")

### DIFF
--- a/framework/encode/vulkan_capture_manager.h
+++ b/framework/encode/vulkan_capture_manager.h
@@ -927,20 +927,6 @@ class VulkanCaptureManager : public ApiCaptureManager
                                         uint32_t                     bindInfoCount,
                                         const VkBindImageMemoryInfo* pBindInfos);
 
-    void PostProcess_vkCreateBuffer(VkResult                     result,
-                                    VkDevice                     device,
-                                    const VkBufferCreateInfo*    pCreateInfo,
-                                    const VkAllocationCallbacks* pAllocator,
-                                    VkBuffer*                    pBuffer)
-    {}
-
-    void PostProcess_vkCreateImage(VkResult                     result,
-                                   VkDevice                     device,
-                                   const VkImageCreateInfo*     pCreateInfo,
-                                   const VkAllocationCallbacks* pAllocator,
-                                   VkImage*                     pImage)
-    {}
-
     void PostProcess_vkCmdBeginRenderPass(VkCommandBuffer              commandBuffer,
                                           const VkRenderPassBeginInfo* pRenderPassBegin,
                                           VkSubpassContents)
--- a/framework/encode/vulkan_capture_manager.h
+++ b/framework/encode/vulkan_capture_manager.h
@@ -932,44 +932,14 @@ class VulkanCaptureManager : public ApiCaptureManager
                                     const VkBufferCreateInfo*    pCreateInfo,
                                     const VkAllocationCallbacks* pAllocator,
                                     VkBuffer*                    pBuffer)
-    {
-        if (IsCaptureModeTrack() && (result == VK_SUCCESS) && (pCreateInfo != nullptr))
-        {
-            GFXRECON_ASSERT(state_tracker_ != nullptr);
-
-            auto buffer_wrapper = vulkan_wrappers::GetWrapper<vulkan_wrappers::BufferWrapper>(*pBuffer);
-
-            if (buffer_wrapper->is_sparse_buffer)
-            {
-                // We will need to set the bind_device for handling sparse buffers. There will be no subsequent
-                // vkBindBufferMemory, vkBindBufferMemory2 or vkBindBufferMemory2KHR calls for sparse buffer, so we
-                // assign bind_device to the device that created the buffer.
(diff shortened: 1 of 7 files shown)


framework/encode/vulkan_capture_manager.h
```diff
@@ -932,44 +932,14 @@ class VulkanCaptureManager : public ApiCaptureManager
...
-            if (buffer_wrapper->is_sparse_buffer)
-            {
-                // We will need to set the bind_device for handling sparse buffers. There will be no subsequent
-                // vkBindBufferMemory, vkBindBufferMemory2 or vkBindBufferMemory2KHR calls for sparse buffer, so we
-                // assign bind_device to the device that created the buffer.
-                buffer_wrapper->bind_device = vulkan_wrappers::GetWrapper<vulkan_wrappers::DeviceWrapper>(device);
-            }
-        }
-    }
+    {}
```
fabian-lunarg: instead of empty bodies here, we should rollback the unused hooks `PostProcess_vkCreateBuffer` (and `PostProcess_vkCreateImage` below).
the only users are the `CustomEncoderPostCall` template specializations (`custom_vulkan_encoder_commands.h`), which should be rolled back too.

framework/encode/vulkan_state_tracker_initializers.h
```diff
@@ -694,12 +693,11 @@ inline void InitializeState<VkDevice, vulkan_wrappers::BufferWrapper, VkBufferCr
     assert(create_info != nullptr);
     assert(create_parameters != nullptr);
 
-    GFXRECON_UNREFERENCED_PARAMETER(parent_handle);
-
+    wrapper->device            = vulkan_wrappers::GetWrapper<vulkan_wrappers::DeviceWrapper>(parent_handle);
```
fabian-lunarg: this is redundant and already handled in OverrideCreateBuffer

framework/encode/vulkan_state_tracker_initializers.h
```diff
@@ -778,8 +777,7 @@ inline void InitializeState<VkDevice, vulkan_wrappers::ImageWrapper, VkImageCrea
...
```
fabian-lunarg: same as above, already handled in `OverrideCreateImage`

framework/encode/vulkan_state_tracker_initializers.h
```diff
@@ -694,12 +693,11 @@ inline void InitializeState<VkDevice, vulkan_wrappers::BufferWrapper, VkBufferCr
...
     wrapper->create_call_id    = create_call_id;
     wrapper->create_parameters = std::move(create_parameters);
 
-    wrapper->created_size = create_info->size;
+    wrapper->size = create_info->size;
```
fabian-lunarg: same

framework/encode/vulkan_state_writer.cpp
```diff
@@ -5164,7 +5165,7 @@ void VulkanStateWriter::WriteTensorSnapshotState(const VulkanStateTable& state_t
     state_table.VisitWrappers([&](vulkan_wrappers::TensorARMWrapper* wrapper) {
         GFXRECON_ASSERT(wrapper != nullptr);
 
-        const vulkan_wrappers::DeviceWrapper* device_wrapper = wrapper->bind_device;
+        const vulkan_wrappers::DeviceWrapper* device_wrapper = wrapper->device;
         if (device_wrapper == nullptr)
```
fabian-lunarg: device is now set at creation, so the null check never fires.
the warning below suggest this is a possible case, but shouldn't.
-> we can replace the branch here with a `GFXRECON_ASSERT`

framework/encode/vulkan_state_writer.cpp
```diff
@@ -1533,7 +1533,7 @@ void VulkanStateWriter::WriteBufferDeviceAddressState(const VulkanStateTable& st
 {
     state_table.VisitWrappers([&](const vulkan_wrappers::BufferWrapper* wrapper) {
         GFXRECON_ASSERT(wrapper != nullptr && wrapper->device != VK_NULL_HANDLE);
```
fabian-lunarg: should match the if-branch below (`... wrapper->device != nullptr`)

framework/encode/vulkan_capture_manager.h
fabian-lunarg: done

framework/encode/vulkan_state_writer.cpp
fabian-lunarg: done

framework/encode/vulkan_state_writer.cpp
fabian-lunarg: done

framework/encode/vulkan_state_tracker_initializers.h
fabian-lunarg: done

fabian-lunarg (approved):

PR #3156, opened 2026-07-30
[Capture/Convert]: fix HDR metadata leak and AS JSON assignment typos

tpower-amd:
## Problem

Two bugs identified through static analysis affecting capture state writing and convert output.

## Solution

- `framework/encode/dx12_object_wrapper_info.h`, `dx12_state_tracker.cpp`, `dx12_state_writer.cpp`: `TrackSetHDRMetaData` allocated a new `char` buffer on every `SetHDRMetaData` call without freeing the previous one; the owning struct had no destructor. Replace `void* hdr_metadata` with `std::vector<uint8_t>`, use `assign()` in the tracker so repeat calls replace the previous contents, `clear()` when `pMetaData` is null or `Size` is 0 to keep the vector and `hdr_metadata_size` in sync, and pass `nullptr` to `EncodeVoidArray` when the buffer is empty.

- `framework/util/json_util.cpp`: Two lines in `FieldToJson` for `InitDx12AccelerationStructureGeometryDesc` used the comma operator instead of `=` for `aabbs_count` and `aabbs_stride`, silently discarding both values and emitting `null` in `gfxrecon-convert` JSON output for DXR acceleration-structure geometry. Replace `,` with `=` on both lines.

## Testing

- ✅ All test suites passed: `gfxrecon_util_test`, `gfxrecon_encode_test`, `gfxrecon_decode_test`, `gfxrecon_graphics_test`, and `gfxrecon_replay_event_plugin_loader_test`.

⚠️ May cause CI tests involving golden files to fail and require them to be regenerated.

### DIFF
--- a/framework/encode/dx12_object_wrapper_info.h
+++ b/framework/encode/dx12_object_wrapper_info.h
@@ -253,9 +253,9 @@ struct IDXGISwapChainInfo : public DxgiWrapperInfo
     bool                   set_color_space{ false };
     bool                   set_hdr_metadata{ false };
     DXGI_COLOR_SPACE_TYPE  color_space_type{};
-    DXGI_HDR_METADATA_TYPE   hdr_metadata_type{};
-    UINT                     hdr_metadata_size{ 0 };
-    std::vector<uint8_t>     hdr_metadata;
+    DXGI_HDR_METADATA_TYPE hdr_metadata_type{};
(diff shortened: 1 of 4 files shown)


davidd-lunarg (approved):
Approved. I'll mark it ready to merge, ~then it just needs to be updated with the latest from dev.~ <- Nevermind, the merge queue handles that automatically.

PR #3157, opened 2026-07-30
[Replay]: fix unreachable eviction loop in ExecuteAndWaitForCommandList

tpower-amd:
## Problem

The `MakeResident`/`Evict` cleanup loop in `ExecuteAndWaitForCommandList` was positioned after the function's `return` statement and never executed. `MakeResident` and `Evict` are ref-counted — every `MakeResident` call must be paired with an `Evict` to release the residency reference. Without the eviction loop running, GPU residency accumulated without bound across
the replay resource initialization phase, and `resident_resources` carried stale pointers into subsequent calls.

## Solution

- `framework/graphics/dx12_resource_data_util.cpp`: Move the eviction loop before the `return` statement and clear `resident_resources` after so the vector does not carry stale pointers into subsequent calls.

## Performance

Tested on discrete GPU hardware across multiple titles. Performance improvements include a 2–9% speed increase in replay load times, with the improvement scaling with resource count. On high-resource traces the current binary accumulates unbounded residency overhead that this fix eliminates. Removing residency management entirely is not viable — cold start penalties become unpredictable and driver-dependent without `MakeResident`/`Evict`.

## Testing

- ✅ All test suites passed: `gfxrecon_util_test`, `gfxrecon_encode_test`, `gfxrecon_decode_test`, `gfxrecon_graphics_test`, and
  `gfxrecon_replay_event_plugin_loader_test`.
- ✅ Performance validated on discrete GPU hardware across multiple titles with varying resource counts between ~2K-20K.

### DIFF
--- a/framework/graphics/dx12_resource_data_util.cpp
+++ b/framework/graphics/dx12_resource_data_util.cpp
@@ -626,7 +626,7 @@ HRESULT Dx12ResourceDataUtil::ExecuteAndWaitForCommandList(ID3D12CommandQueue* q
         command_queue = command_queue_;
     }
     command_queue->ExecuteCommandLists(1, cmd_lists);
-    return dx12::WaitForQueue(command_queue, command_fence_, ++fence_value_);
+    HRESULT result = dx12::WaitForQueue(command_queue, command_fence_, ++fence_value_);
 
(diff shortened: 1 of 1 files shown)


davidd-lunarg (approved):
LGTM. Will merge once CI tests pass.

PR #3160, opened 2026-07-31
Improve debug util tracking

panos-lunarg:
This is an attempt to enforce replay to enclose any synthesized api calls between Begin/EndInjectedCommands() and also inject DebuUtil to mark commands injected in an application command buffer or command buffers that are fully synthesized by replay


framework/generated/khronos_generators/vulkan_generators/vulkan_injected_call_table_generator.py
```diff
@@ -0,0 +1,167 @@
...
+        write('GFXRECON_END_NAMESPACE(ScopeCheck)', file=self.outFile)
+        self.newline()
+
+        write('class VulkanInjectedDeviceCallsTableBase', file=self.outFile)
+        write('{', file=self.outFile)
+        write('  public:', file=self.outFile)
+        write('    // clang-format off', file=self.outFile)
+
+        for name in device_cmds:
+            write('    PFN_{} {}{{ ScopeCheck::{} }};'.format(name, name[2:], name), file=self.outFile)
```
fabian-lunarg, line 161: The `ScopeCheck::` entries are free functions — no `this`, so they can't reach the table they were called through.

-> so every call goes through the global `device_table_registry_g` (shared_mutex acquisition + hash lookup), per injected call. 

The registry isn't optional given the current shape — it's how the trampoline finds the table to forward to. But the shape itself only exists to host the `GFXRECON_ASSERT`, which is dead under `NDEBUG` by default.

Emitting member functions on the derived table instead of PFN members avoids all of it:

```
void CmdCopyBuffer(VkCommandBuffer cb, ...) const 
{
    GFXRECON_ASSERT(util::InsideInjectedCommands() && "...");
    raw_table_->CmdCopyBuffer(cb, ...);
}
```

-> Call sites stay identical, release compiles to a direct call, and the registry + mutex + noop fallback go away.
- that would also solve the question if replay-allocated command buffers need `copy_dispatch_table_from_device`, since nothing is keyed on the dispatch pointer any more

framework/decode/vulkan_object_cleanup_util.cpp
```diff
@@ -176,7 +177,9 @@ void FreeAllLiveObjects(CommonObjectInfoTable*
         &CommonObjectInfoTable::RemoveVkEventInfo,
         [&](const VulkanDeviceInfo* parent_info, const VulkanEventInfo* object_info) {
             assert((parent_info != nullptr) && (object_info != nullptr));
-            get_device_table(parent_info->handle)->DestroyEvent(parent_info->handle, object_info->handle, nullptr);
+            graphics::InjectedCommandScope injected_commands_scope;
```
fabian-lunarg: I agree marking cleanup as injection, but think this could be done with a single scope at the top of `FreeAllLiveObjects`.
as written that's one `BeginInjectedCommands`/`EndInjectedCommands` pair — and one external `SetInjectedCommandCallbacks` round-trip — per destroyed object, across the 30 lambdas.

-> a `MarkInjectedCommandsHelper` at the top 'should' be enough.

framework/generated/khronos_generators/vulkan_generators/vulkan_injected_call_table_generator.py
fabian-lunarg: spinning this further, it does raise other problems and appears to be a rabbit-hole. 
also I overweighted the mutex/lookup cost and should be negligible in practice.

framework/graphics/vulkan_injected_call_table.cpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+const VulkanDeviceTable* GetRegisteredDeviceTable(const void* handle)
+{
+    static const VulkanDeviceTable noop_table{};
+
+    if (handle == nullptr)
+    {
+        GFXRECON_ASSERT(false && "Injected call made with a null handle");
+        return &noop_table;
```
fabian-lunarg: only way to arrive here is basically programming-error.
in Release this would silently drop work, so we should terminate there via:
`GFXRECON_LOG_FATAL("Injected Vulkan call made with a null handle.");` 

same below, so `static const VulkanDeviceTable noop_table{};` could be dropped.

framework/decode/vulkan_command_buffer_util.cpp
```diff
@@ -70,6 +73,12 @@ void VulkanCommandBufferAssociatedInfo::ReplaceWithNewHandle(VulkanCommandBuffer
 
         VkResult result = device_table_->AllocateCommandBuffers(device_info_->handle, &alloc_info, new_handles_ptr);
         GFXRECON_ASSERT(result == VK_SUCCESS);
+
+        for (size_t i = previous_size; i < associated_handles_.size(); ++i)
+        {
+            // Because this command buffer was not allocated through the loader, it must be assigned a dispatch table.
+            graphics::copy_dispatch_table_from_device(device_info->handle, *(new_handles_ptr + i));
```
fabian-lunarg: I'm asking myself, is this needed? if yes, these look like they need it too: screenshot_handler.cpp:205, vulkan_captured_swapchain.cpp:406/607, vulkan_virtual_swapchain.cpp:378/1637, vulkan_resource_initializer.cpp:424

but I think they're actually not needed (on conformant loader at least. for me `vkGetDeviceProcAddr` returns its own trampoline for `vkAllocateCommandBuffers`/`vkGetDeviceQueue` and sets the dispatch key).
can you remember what purpose `copy_dispatch_table_from_device` had or which problem it fixes? (Android, non-loader paths, ..?)

panos-lunarg: No I don't think they are needed. Will remove

framework/decode/vulkan_command_buffer_util.cpp
```diff
@@ -134,12 +145,19 @@ void VulkanCommandBufferAssociatedInfo::FreeCommandBuffers(VkCommandPool pool)
 {
     GFXRECON_ASSERT(!associated_handles_.empty());
 
+    auto injected_command_scope = device_table_->MarkScope();
     device_table_->FreeCommandBuffers(
         device_info_->handle, pool, static_cast<uint32_t>(associated_handles_.size()), associated_handles_.data());
+
+    // The freed handles must not be reused.
+    associated_handles_.clear();
```
fabian-lunarg, line 147: nice! this fixes a real, pre-existing use-after-free bug. 
you 'could' mention it in commit/Pr-desc somewhere

framework/decode/vulkan_command_buffer_util.cpp
fabian-lunarg: some digging: the pattern goes back to a4730319b (2019), "Add utilities for resource memory snapshot"
- 056559ab5 added the same trick for a VkQueue three weeks later.

-> 'might' have been necessary with 2019's loader, then got copied across later. 
-> the two remaining calls in vulkan_address_replacer.cpp:2295 / vulkan_resources_util.cpp:1297 are 
    probably dead as well, and the entire helper with them.

panos-lunarg: done

framework/decode/vulkan_object_cleanup_util.cpp
panos-lunarg: Done

framework/graphics/vulkan_injected_call_table.cpp
panos-lunarg: done

fabian-lunarg (approved):
looking good, let's merge

framework/graphics/vulkan_injected_call_table.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include <shared_mutex>
+#include <string>
+#include <vulkan/vulkan_core.h>
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(graphics)
+
+static std::atomic<bool> annotate_injected_commands{ false };
+static std::unordered_map<graphics::VulkanDispatchKey, const graphics::VulkanDeviceTable*> device_table_registry_g;
+static std::shared_mutex                                                                   registry_mutex_g;
```
jzulauf-lunarg, line 36: Currently replay is singly threaded, is it not?  Also, is there any reason for a thread safe global scope as opposed to a thread local table or stack?  The injection operations appear to be all thread local and RAII scoped.

framework/decode/vulkan_command_buffer_util.cpp
```diff
@@ -197,8 +213,8 @@ graphics::VulkanSemaphore VulkanCommandBufferUtil::SubmitPreviouslySplitCommandB
     const std::span<std::vector<VkCommandBuffer>> current_submits_cmdbufs,
     const std::span<graphics::VulkanSemaphore>    wait_semaphores)
 {
-    util::MarkInjectedCommandsHelper mark_injected_commands_helper;
```
jzulauf-lunarg, line 200: The naming of RAII things really should be clear.  I had to chase this through several layers to see that it's a scoped Guard like operator, even in default construction

framework/decode/vulkan_frame_warm_up.cpp
```diff
@@ -26,23 +26,23 @@
...
+                                     const graphics::VulkanInjectedDeviceCallsTable* device_table,
+                                     const graphics::VulkanInstanceTable*            instance_table,
+                                     CommonObjectInfoTable&                          object_table,
+                                     const std::string&                              spirv_path,
+                                     uint32_t                                        warm_up_load) :
     device_table_(device_table),
     spirv_path_(spirv_path), warm_up_load_(warm_up_load)
 {
-    util::MarkInjectedCommandsHelper mark_injected_commands_helper;
+    auto injected_command_scope = device_table->MarkScope();
```
jzulauf-lunarg, line 45: This just wraps MarkInjectedCommandsHelper() (by way of default constructed InjectedCommandScope), but there are several places where MarkInjectedCommandsHelper  is still used why are some changed and not others.

framework/generated/khronos_generators/vulkan_generators/vulkan_injected_call_table_generator.py
```diff
@@ -0,0 +1,167 @@
...
+                    'inline VKAPI_ATTR {} VKAPI_CALL {}({}) {{ GFXRECON_ASSERT(util::InsideInjectedCommands() && "{} called outside an InjectedCommandScope"); return GetRegisteredDeviceTable({})->{}({}); }}'
+                    .format(return_type, name, params, name, values[0].name, name[2:], args),
+                    file=self.outFile
+                )
+
+        write('// clang-format on', file=self.outFile)
+        write('GFXRECON_END_NAMESPACE(ScopeCheck)', file=self.outFile)
+        self.newline()
+
+        write('class VulkanInjectedDeviceCallsTableBase', file=self.outFile)
```
jzulauf-lunarg, line 155: Why generate a different base instead inheriting from VulkanDeviceTable?  Seems redundant.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -1533,8 +1531,12 @@ void VulkanReplayConsumerBase::AddInstanceTable(VkInstance instance)
 
 void VulkanReplayConsumerBase::AddDeviceTable(VkDevice device, PFN_vkGetDeviceProcAddr gpa)
 {
-    graphics::VulkanDeviceTable& table = device_tables_[graphics::GetVulkanDispatchKey(device)];
+    const auto key = graphics::GetVulkanDispatchKey(device);
+
+    graphics::VulkanDeviceTable& table = device_tables_[key];
     graphics::LoadVulkanDeviceTable(gpa, device, &table);
+
+    injected_device_tables_.insert_or_assign(key, graphics::VulkanInjectedDeviceCallsTable(device, &table));
```
jzulauf-lunarg, line 1539: Not sure why the injected_device_tables_ have to be inserted into a global mapped when the exist 1-1 with the device_tables_   ... instead of just pushing a pointer to the real table (or even this injected table) into a global
...

but as all injected operations are scoped, the guard can push this onto a TLS pointer for the duration, and some pointer to an oops table otherwise.

jzulauf-lunarg (commented):
A lot here... initial comments only.  Not sure why the "injected table" needs to be as distinct as it is.

panos-lunarg:
Taking another stab at this here: https://github.com/LunarG/gfxreconstruct/pull/3172

PR #3162, opened 2026-07-31
Fix Android Virtual Swapchain ANativeWindow collision after AdHoc presentation

olehkuznetsov:
This is reproducible if application uses swapchain, but also at the same time uses vkFrameBoundaryANDROID.

Fix Android Virtual Swapchain ANativeWindow collision after AdHoc presentation
When replaying Android traces, the replay runtime sometimes utilizes an AdHoc surface presentation layer to visualize headless loading phases. However, in the original architecture, this AdHoc surface held onto system OS Window resources permanently.

When the trace finally attempted to create its genuine swapchain surface (e.g. vkCreateAndroidSurfaceKHR), it would fail with VK_ERROR_NATIVE_WINDOW_IN_USE_KHR because the window was still bound to the AdHoc surface.

Changes:

- Overrode CreateSurface in VulkanVirtualSwapchain to intercept the application's genuine hardware surface request.

- When a real surface is requested, we now preemptively clear the internal adhoc_device_data_ and gracefully release the underlying ANativeWindow before delegating instantiation back to the base class.

- Introduced application_surface_created_ latch to permanently halt artificial AdHoc FrameBoundary renders once the application trace has successfully transitioned to rendering its own native pipeline....

### DIFF
--- a/framework/decode/vulkan_virtual_swapchain.cpp
+++ b/framework/decode/vulkan_virtual_swapchain.cpp
@@ -1383,7 +1383,7 @@ void VulkanVirtualSwapchain::PresentImageAdHoc(const VulkanDeviceInfo*          
     VkSemaphore semaphore = semaphore_info == nullptr ? VK_NULL_HANDLE : semaphore_info->handle;
 
     // If there is no image to present, do nothing
-    if (image == VK_NULL_HANDLE)
+    if (image == VK_NULL_HANDLE || application_surface_created_)
     {
         return;
     }
@@ -1483,7 +1483,7 @@ void VulkanVirtualSwapchain::PresentImageAdHoc(const VulkanDeviceInfo*          
         // empty -> automatic wsi deduction
         std::string wsi_extension;
 
-        result = CreateSurface(
+        result = VulkanSwapchain::CreateSurface(
             VK_SUCCESS, instance_info, wsi_extension, 0, &swapchain.surface_ptr, instance_table, application);
         GFXRECON_ASSERT(result == VK_SUCCESS);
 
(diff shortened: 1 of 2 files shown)


framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -1383,7 +1383,7 @@ void VulkanVirtualSwapchain::PresentImageAdHoc(const VulkanDeviceInfo*
     VkSemaphore semaphore = semaphore_info == nullptr ? VK_NULL_HANDLE : semaphore_info->handle;
 
     // If there is no image to present, do nothing
-    if (image == VK_NULL_HANDLE)
+    if (image == VK_NULL_HANDLE || application_surface_created_)
```
fabian-lunarg: this early return is wrong for all platforms but Android,
due to the assumption that only one window exists at a time.

this requires an additional guard to restrict to Android only

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -1916,5 +1916,23 @@ VkResult VulkanVirtualSwapchain::CreateVirtualSwapchainImage(const VulkanDeviceI
...
+                                               VkFlags                              flags,
+                                               HandlePointerDecoder<VkSurfaceKHR>*  surface,
+                                               const graphics::VulkanInstanceTable* instance_table,
+                                               application::Application*            application)
+{
+    // The application trace is attempting to create a genuine Surface, so it will need to attach to
+    // the single ANativeWindow. If we have used an AdHoc surface to bypass android splash screens,
+    // we must destroy that surface immediately so we release the NativeWindow cleanly before the trace tries.
+    application_surface_created_ = true;
+    adhoc_device_data_.clear();
```
fabian-lunarg, line 1943: `AdhocDeviceData` has no destructor, so this leaks a `VkCommandPool` causing VVL-errors later during cleanup.

framework/decode/vulkan_virtual_swapchain.h
```diff
@@ -177,6 +177,14 @@ class VulkanVirtualSwapchain : public VulkanSwapchain
 
     void CleanSwapchainResourceData(const VulkanDeviceInfo* device_info, const VulkanSwapchainKHRInfo* swapchain_info);
 
+    virtual VkResult CreateSurface(VkResult                             original_result,
+                                   VulkanInstanceInfo*                  instance_info,
+                                   const std::string&                   wsi_extension,
+                                   VkFlags                              flags,
+                                   HandlePointerDecoder<VkSurfaceKHR>*  surface,
+                                   const graphics::VulkanInstanceTable* instance_table,
+                                   application::Application*            application) override;
```
fabian-lunarg: override should be grouped under public (currently protected)

fabian-lunarg:
@olehkuznetsov I went ahead and added a 2nd commit, fixing all issues I flagged above. 
I hope that's fine with you :)

framework/decode/vulkan_virtual_swapchain.cpp
fabian-lunarg: added a dtor for `AdhocDeviceData`, so this is now fine

framework/decode/vulkan_virtual_swapchain.cpp
fabian-lunarg: added a preprocessor-guard for android

framework/decode/vulkan_virtual_swapchain.h
fabian-lunarg: done

fabian-lunarg (approved):
I took the liberty to add a 2nd commit to prevent a regression 
w. `--present-override` and `vkFrameBoundary` on Desktop.

olehkuznetsov:
> I took the liberty to add a 2nd commit to prevent a regression w. `--present-override` and `vkFrameBoundary` on Desktop.

Sure, thanks a lot for a great review and the fix :)

PR #3169, opened 2026-08-04
Fix the spirv-header path for macOS when SDK is installed

johnzupin:
Due to a quirk in AppleClang, the header include order was incorrect and used SPIRV headers in /usr/local/include from the system global SDK install instead of the local copy in external/SPIRV-Headers/include.


charles-lunarg:
Closing as an alternate PR #3179 has been merged

PR #3170, opened 2026-08-04
Enable openxr tutorial trace

wzchu-lunarg:
https://github.com/LunarG/ci-gfxr-suites/pull/176


MarkY-LunarG:
How are we replaying this if we don't have an OpenXR instance running?  Do the test apps have Monado installed?  Or is this only the Vulkan side of commands?

bradgrantham-lunarg:
> How are we replaying this if we don't have an OpenXR instance running? Do the test apps have Monado installed? Or is this only the Vulkan side of commands?

It's just the convert subtest

wzchu-lunarg:
Already merged into master

PR #3171, opened 2026-08-04
Erase last iteration fence when destroyed during frame looping

### LINKED ISSUE #3168: [replay] Allocate/free fence when replaying in frame loop specified using VK_EXT_frame_boundary crashes
Replay crashes in frame loop when looping is specified using VK_EXT_frame_boundary. See PR https://github.com/LunarG/ci-gfxr-suites/pull/166.

To reproduce: On windows nvidia 5080: remove the commented out fence allocate/frees in the replay loop in cube.c in the above PR, rebuild cube, capture ~200 frames of cube, and replay the capture file using args specified in the commit-suite.json file for in the above PR.

wzchu-lunarg:
`VulkanReplayFrameLoopConsumer::FixupDeviceFences` seems to crash as there are cases where it is trying to reset a fence that is already destroyed, but is still in `initial_fence_states_`.

The fence in question is a fence that was created in a previous frame, and will be destroyed in the current frame at the last iteration. This is due to `VulkanReplayFrameLoopConsumer::Process_vkDestroyFence`, where we remove the fence from `initial_fence_states_` if `allocatedLoopResources.contains(args.fence)`.

However, `allocatedLoopResources` is only populated by fences that are created within the current frame, so fences created before aren't erased from the map, which results in GFXR later trying to reset a fence that was destroyed.

Since `VulkanReplayFrameLoopConsumerBase::Process_vkDestroyFence` seems to destroy the fence correctly:
```
// Call Process_vkDestroyFence if:
//    We are not looping
//    We are looping and args.fence is in allocatedLoopResources
//    We are looping and this is the last iteration
```
we can reuse this logic by first calling `VulkanReplayFrameLoopConsumerBase::Process_vkDestroyFence(call_info, args);` (currently at the end) and then checking if it was actually removed with `GetObjectInfoTable().GetVkFenceInfo(args.fence) == nullptr`. 

Fixes https://github.com/LunarG/gfxreconstruct/issues/3168

### DIFF
--- a/framework/decode/vulkan_replay_frame_loop_consumer.cpp
+++ b/framework/decode/vulkan_replay_frame_loop_consumer.cpp
@@ -282,7 +282,8 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateFence(const ApiCallInfo& cal
 
 void VulkanReplayFrameLoopConsumer::Process_vkDestroyFence(const ApiCallInfo& call_info, args::DestroyFence& args)
 {
-    if (allocatedLoopResources.contains(args.fence))
+    VulkanReplayFrameLoopConsumerBase::Process_vkDestroyFence(call_info, args);
+    if (GetObjectInfoTable().GetVkFenceInfo(args.fence) == nullptr)
     {
         if (per_device_fence_tracking_.contains(args.device))
         {
@@ -293,7 +294,6 @@ void VulkanReplayFrameLoopConsumer::Process_vkDestroyFence(const ApiCallInfo& ca
             t.initial_fence_states_.erase(args.fence);
         }
     }
-    VulkanReplayFrameLoopConsumerBase::Process_vkDestroyFence(call_info, args);
 }
 
 void VulkanReplayFrameLoopConsumer::TrackFenceState(format::HandleId device, format::HandleId fence)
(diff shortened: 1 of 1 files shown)


framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -282,7 +282,8 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateFence(const ApiCallInfo& cal
 
 void VulkanReplayFrameLoopConsumer::Process_vkDestroyFence(const ApiCallInfo& call_info, args::DestroyFence& args)
 {
-    if (allocatedLoopResources.contains(args.fence))
+    VulkanReplayFrameLoopConsumerBase::Process_vkDestroyFence(call_info, args);
+    if (GetObjectInfoTable().GetVkFenceInfo(args.fence) == nullptr)
```
antonio-lunarg: nit: For a readability standpoint, I would introduce something like:
```cpp
bool actually_destroyed = (GetObjectInfoTable().GetVkFenceInfo(args.fence) == nullptr);
if (actually_destroyed)
```

antonio-lunarg (approved):

antonio-lunarg (approved):

PR #3172, opened 2026-08-05
Scope-based marking for Vulkan api calls synthesized by replay

panos-lunarg:
Adds graphics::VulkanInjectedDeviceCalls as a wrapper around the VulkanDeviceTable. The table's function pointer are reachable only by VulkanInjectedDeviceCalls::Open() which automates the calls to BeginInjectedCommands and EndInjectedCommands.
Also provides funcitonality to mark commands injected to existing command buffers or fully synthesized command buffers using the Vulkan DebutUtils functionality

### DIFF
--- a/framework/decode/vulkan_command_buffer_util.cpp
+++ b/framework/decode/vulkan_command_buffer_util.cpp
@@ -28,15 +28,15 @@
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
-VulkanCommandBufferAssociatedInfo::VulkanCommandBufferAssociatedInfo(const VulkanDeviceInfo*            device_info,
-                                                                     const graphics::VulkanDeviceTable* device_table,
-                                                                     CommonObjectInfoTable*             object_table,
-                                                                     format::HandleId command_buffer_id) :
+VulkanCommandBufferAssociatedInfo::VulkanCommandBufferAssociatedInfo(
+    const VulkanDeviceInfo*                    device_info,
+    const graphics::VulkanInjectedDeviceCalls& device_table,
+    CommonObjectInfoTable*                     object_table,
+    format::HandleId                           command_buffer_id) :
     device_info_(device_info),
     device_table_(device_table), object_table_(object_table), split_semaphore_(device_info, device_table)
 {
     GFXRECON_ASSERT(device_info != nullptr);
-    GFXRECON_ASSERT(device_table != nullptr);
     GFXRECON_ASSERT(object_table != nullptr);
 
     // Store the original command buffer handle and map it to the command buffer ID for future reference.
@@ -68,7 +68,8 @@ void VulkanCommandBufferAssociatedInfo::ReplaceWithNewHandle(VulkanCommandBuffer
         alloc_info.level                       = VK_COMMAND_BUFFER_LEVEL_PRIMARY;
         alloc_info.commandBufferCount          = NEW_COMMAND_BUFFER_COUNT;
 
-        VkResult result = device_table_->AllocateCommandBuffers(device_info_->handle, &alloc_info, new_handles_ptr);
+        auto     injected = device_table_.Open();
+        VkResult result   = injected->AllocateCommandBuffers(device_info_->handle, &alloc_info, new_handles_ptr);
         GFXRECON_ASSERT(result == VK_SUCCESS);
     }
 
@@ -78,15 +79,14 @@ void VulkanCommandBufferAssociatedInfo::ReplaceWithNewHandle(VulkanCommandBuffer
     command_buffer_info->handle = next_handle;
 }
 
-VulkanCommandBufferUtil::VulkanCommandBufferUtil(const VulkanDeviceInfo*            device_info,
-                                                 const graphics::VulkanDeviceTable* device_table,
-                                                 CommonObjectInfoTable*             object_table,
-                                                 VulkanStateRecordingDecoder*       decoder) :
+VulkanCommandBufferUtil::VulkanCommandBufferUtil(const VulkanDeviceInfo*                    device_info,
+                                                 const graphics::VulkanInjectedDeviceCalls& device_table,
+                                                 CommonObjectInfoTable*                     object_table,
+                                                 VulkanStateRecordingDecoder*               decoder) :
     device_info_(device_info),
     device_table_(device_table), object_table_(object_table), decoder_(decoder)
 {
     GFXRECON_ASSERT(device_info != nullptr);
-    GFXRECON_ASSERT(device_table != nullptr);
     GFXRECON_ASSERT(object_table != nullptr);
 }
 
@@ -119,7 +119,8 @@ VkCommandBuffer VulkanCommandBufferAssociatedInfo::ResetAssociatedHandles()
     // Reset all the associated command buffers.
     for (VkCommandBuffer handle : associated_handles_)
     {
-        VkResult result = device_table_->ResetCommandBuffer(handle, 0);
+        auto     injected = device_table_.Open();
+        VkResult result   = injected->ResetCommandBuffer(handle, 0);
         GFXRECON_ASSERT(result == VK_SUCCESS);
     }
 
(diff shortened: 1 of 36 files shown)


panos-lunarg:
@jzulauf-lunarg @fabian-lunarg 
Created this PR in place of https://github.com/LunarG/gfxreconstruct/pull/3160.

tools/replay/replay_settings.h
```diff
@@ -344,6 +343,10 @@ static void PrintUsage(const char* exe_name)
     GFXRECON_WRITE_CONSOLE("          \t\tForce wait on completion of queue operations for all queues");
     GFXRECON_WRITE_CONSOLE("          \t\tbefore calling Present. This is needed for accurate acquisition");
     GFXRECON_WRITE_CONSOLE("          \t\tof instrumentation data on some platforms.");
+    GFXRECON_WRITE_CONSOLE("  --annotate-injected-commands");
```
MarkY-LunarG, line 346: Make sure to update replay docs with new setting.

framework/decode/vulkan_object_cleanup_util.cpp
jzulauf-lunarg, line 361: This cleanup path is still replay-synthesized, but unlike the direct `injected->Destroy*` paths nearby it is not under a `VulkanInjectedDeviceCalls::Open()` scope. Should this allocator-backed cleanup path open a `VulkanInjectedDeviceCalls::Open()` scope so any Vulkan work
  performed by `allocator->DestroyBuffer(...)` will be bracketed by `BeginInjectedCommands()` / `EndInjectedCommands()`?

framework/decode/vulkan_virtual_swapchain.cpp
```diff
@@ -47,13 +47,24 @@ void VulkanVirtualSwapchain::CleanDeviceResources(VkDevice device, const graphic
...
+    if (!injected_calls.has_value())
+    {
+        // Never populated; there is nothing to destroy.
+        GFXRECON_ASSERT(swapchains.empty() && (copy_util == nullptr) && (command_pool == VK_NULL_HANDLE));
+        return;
+    }
+
+    // Keep the injected-commands window open for all cleanup, including
+    // copy_util's destructor, which destroys its own staging resources.
+    auto injected = injected_calls->Open();
```
jzulauf-lunarg, line 59: Should `copy_util` be explicitly reset here to ensure its cleanup happens within the
  `VulkanInjectedDeviceCalls::Open()` scope?

jzulauf-lunarg (commented):
A couple observations from reviewing with Codex

framework/decode/vulkan_virtual_swapchain.cpp
panos-lunarg: done

framework/decode/vulkan_object_cleanup_util.cpp
panos-lunarg: I haven't migrated the allocators yet. I plan doing this in a separate PR

tools/replay/replay_settings.h
panos-lunarg: done

jzulauf-lunarg (commented):
Two comments from review with Codex

framework/decode/vulkan_command_buffer_util.cpp
```diff
@@ -161,13 +167,16 @@ void VulkanCommandBufferUtil::SplitCommandBuffer(VulkanCommandBufferInfo* comman
...
+        ReplaceWithAssociatedCommandBuffer(command_buffer_info);
 
-    VkCommandBufferBeginInfo begin_info = { VK_STRUCTURE_TYPE_COMMAND_BUFFER_BEGIN_INFO };
-    device_table_->BeginCommandBuffer(command_buffer_info->handle, &begin_info);
+        VkCommandBufferBeginInfo begin_info = { VK_STRUCTURE_TYPE_COMMAND_BUFFER_BEGIN_INFO };
+        injected->BeginCommandBuffer(command_buffer_info->handle, &begin_info);
+    }
 
     reissuing_command_buffer_state_ = true;
     decoder_->ReissueCommandBufferState(command_buffer_info->capture_id);
```
jzulauf-lunarg, line 185: Why isn't the Reissue treated as injected?

framework/decode/vulkan_replay_consumer_base.cpp
jzulauf-lunarg, line 12179: Should this be injection wrapped as well?

jzulauf-lunarg, line 2741: There are several calls not injection wrapped in this section.

jzulauf-lunarg, line 11652: Vulkan calls that are not injection wrapped.

jzulauf-lunarg (commented):
Lots of allocator based operation not injection scoped -- but out of scope for this PR

A few places which appear to be injected calls that are not being scoped.  But I wouldn't claim to have found all of them.

framework/decode/vulkan_replay_consumer_base.cpp
panos-lunarg: I haven't migrated the instance table yet. I plan to do it in a separate PR

panos-lunarg: Same as previous comment. I plane to migrate the instance table in a separate PR

framework/decode/vulkan_command_buffer_util.cpp
panos-lunarg: done

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -10152,8 +10165,11 @@ VulkanReplayConsumerBase::OverrideDeferredOperationJoinKHR(PFN_vkDeferredOperati
     VkDevice               device             = device_info->handle;
     VkDeferredOperationKHR deferred_operation = deferred_operation_info->handle;
 
+    auto device_table = GetInjectedDeviceCalls(device);
+    auto injected     = device_table.Open();
```
jzulauf-lunarg: The semaphore on mark injected is TLS, the async threads won't see it.  Does each instance of the lambda need to open?

jzulauf-lunarg (commented):
Looks like there's an interaction between Scope::mark_helper_ and std::async...

Please check.

framework/decode/vulkan_replay_consumer_base.cpp
panos-lunarg: ~~Right. I guess `Open()` needs to move inside the lamda? Will fix.
Since we are talking about the semaphore, `thread_local` _and_ `std::atomic<uint32_t>` seem a bit redundant. I think converting into a simple `uint32_t` will work the same without the cost  of atomic. Thoughts?~~
Sorry, taking this back as I need to look into it a bit more. Will come back

panos-lunarg: The lambda here is not making any synthesized calls. `func` in `result = func(device, deferred_operation);` is not synthesized.
On the other hand the call to `vkGetDeferredOperationMaxConcurrencyKHR` outside of the lambda is synthesized and the call to `Open() ` leaves the Begin/End scope open from a process wide perspective while the lambdas living in their own thread don't have an open scope (which is correct) so that can be confusing. I enclosed the call to `vkGetDeferredOperationMaxConcurrencyKHR` in brackets so that the scope is closed by the time the async kicks off the lambdas.

Since we are talking about the semaphore, thread_local and std::atomic<uint32_t> seem a bit redundant. I think converting into a simple uint32_t will work the same without the cost of atomic.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -10162,11 +10175,18 @@ VulkanReplayConsumerBase::OverrideDeferredOperationJoinKHR(PFN_vkDeferredOperati
     VkDevice               device             = device_info->handle;
     VkDeferredOperationKHR deferred_operation = deferred_operation_info->handle;
 
-    PFN_vkGetDeferredOperationMaxConcurrencyKHR vkGetDeferredOperationMaxConcurrencyKHR =
-        GetDeviceTable(device)->GetDeferredOperationMaxConcurrencyKHR;
+    uint32_t max_concurrency;
+    {
```
jzulauf-lunarg, line 10474: The scoped Open is much clearer. Thanks.

jzulauf-lunarg (commented):
LGTM.

jzulauf-lunarg (approved):

bradgrantham-lunarg:
@MarkY-LunarG @jzulauf-lunarg If your feedback has been resolved to your satisfaction., please mark those threads "resolved" and mark the PR "approved"

MarkY-LunarG (approved):

bradgrantham-lunarg (approved):

PR #3174, opened 2026-08-05
Tensor/datagraph fixes for trim and rebind

bartosz-muszarski-arm:
Change fills gaps in tensor and datagraph support across the codebase
* Support for staging snapshots for tensor trimming
* Better support for multi-bind datagraph pipeline sessions in trim and rebind replay
* Add tensor support for convert and optimize tools

### DIFF
--- a/framework/decode/vulkan_rebind_allocator.cpp
+++ b/framework/decode/vulkan_rebind_allocator.cpp
@@ -3646,9 +3646,23 @@ VulkanRebindAllocator::CreateDataGraphPipelineSession(const VkDataGraphPipelineS
     ResourceAllocInfo* resource_alloc_info = new ResourceAllocInfo();
     resource_alloc_info->object_type       = VK_OBJECT_TYPE_DATA_GRAPH_PIPELINE_SESSION_ARM;
     resource_alloc_info->uses_extensions   = (create_info->pNext != nullptr);
-    resource_alloc_info->usage             = 0;
+    resource_alloc_info->usage             = create_info->flags;
 
-    *allocator_data = static_cast<ResourceData>(reinterpret_cast<uintptr_t>(resource_alloc_info));
+    ResourceData session_allocator_data = reinterpret_cast<ResourceData>(resource_alloc_info);
+
+    // Replay may require session memory even when capture required none and therefore emitted no bind command.
+    // Initialize directly from replay requirements as part of creating the replay session.
+    result = InitializeDataGraphPipelineSessionMemory(*data_graph_pipeline_session, resource_alloc_info);
+    if (result != VK_SUCCESS)
+    {
+        functions_.destroy_data_graph_pipeline_session(
+            device_, *data_graph_pipeline_session, allocator_->GetAllocationCallbacks());
+        delete resource_alloc_info;
+        *data_graph_pipeline_session = VK_NULL_HANDLE;
+        return result;
+    }
+
+    *allocator_data = session_allocator_data;
     return VK_SUCCESS;
 }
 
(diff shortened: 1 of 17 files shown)


tools/optimize/file_optimizer.cpp
```diff
@@ -124,6 +124,24 @@ decode::FileTransformer::VisitResult FileOptimizer::FilterMetaData(const decode:
     return kNeedsPassthrough;
 }
 
+decode::FileTransformer::VisitResult FileOptimizer::FilterMetaData(const decode::InitTensorArgs& args)
+{
+    GFXRECON_ASSERT(format::GetMetaDataType(args.meta_data_id) == format::MetaDataType::kInitTensorCommand);
+
+    // If the tensor is in the unused list, omit its initialization data from the file.
+    if (unreferenced_ids_.find(args.tensor_id) != unreferenced_ids_.end())
```
fabian-lunarg, line 132: this is currently (still) dead code. `unreferenced_ids_` never got populated,
we're missing the override for `Process_vkCreateTensorARM`
let me add a commit, looks like not much is actually missing.

fabian-lunarg: done, needed to add missing bits for `VkTensor`/`VkTensorView` to codegen, 
then add the missing `Process_vkCreateTensorARM` / `Process_vkCreateTensorViewARM` 
to the referenced resource-consumer. 
verified on small adhoc trim-testcase using ARM's tracetooltests (https://github.com/fabian-lunarg/tracetooltests/blob/fabian_unused_tensor/src/vulkan_tensors_unused.cpp) 
-> `InitTensorCommand` is now being filtered by `gfxrecon-optimize` when unused

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -3646,9 +3646,23 @@ VulkanRebindAllocator::CreateDataGraphPipelineSession(const VkDataGraphPipelineS
...
     resource_alloc_info->uses_extensions   = (create_info->pNext != nullptr);
-    resource_alloc_info->usage             = 0;
+    resource_alloc_info->usage             = create_info->flags;
 
-    *allocator_data = static_cast<ResourceData>(reinterpret_cast<uintptr_t>(resource_alloc_info));
+    ResourceData session_allocator_data = reinterpret_cast<ResourceData>(resource_alloc_info);
+
+    // Replay may require session memory even when capture required none and therefore emitted no bind command.
+    // Initialize directly from replay requirements as part of creating the replay session.
+    result = InitializeDataGraphPipelineSessionMemory(*data_graph_pipeline_session, resource_alloc_info);
```
fabian-lunarg, line 3655: I got trouble verifying/reproducing such case using the [ML-emulation-layer](https://github.com/arm/ai-ml-emulation-layer-for-vulkan) and tensor-samples from [tracetooltests](https://github.com/ARM-software/tracetooltests)

-> `vkGetDataGraphPipelineSessionBindPointRequirementsARM` always returns `pBindPointRequirementCount` = 0 for me

bartosz-muszarski-arm: vkGetDataGraphPipelineSessionBindPointRequirementsARM should return 1 for the optical flow tracetooltest.
Still, in order to break current implementation with valid vulkan you'd need a replay environment that requires different number of bindings for the datagraph session than the environment used for capture.
Currently the only way to do that would be to modify emulation layer, introduce a fake bind requirement and use that version for either capture or replay - then use unmodified layer for the other part.
Current implementation works with the assumption the number of bindings will stay the same; this change re-queries bind point requirements on DG session creation instead and turns later binds from the trace into no-ops.

fabian-lunarg: > vkGetDataGraphPipelineSessionBindPointRequirementsARM should return 1 for the optical flow tracetooltest

yes, can confirm, only checked the tensor-samples. running the optical-flow sample I see `requirement_count = 2` using `-m rebind`.
the code-changes look good and you confirmed you tested against the cases I cannot easily reproduce using an unmodified emulation-layer.

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -3782,263 +3796,206 @@ VulkanRebindAllocator::BindDataGraphPipelineSessionMemory(uint32_t bind_info_cou
...
+            }
 
-        VkDataGraphPipelineSessionMemoryRequirementsInfoARM mem_req_info{
-            VK_STRUCTURE_TYPE_DATA_GRAPH_PIPELINE_SESSION_MEMORY_REQUIREMENTS_INFO_ARM
-        };
-        mem_req_info.session     = session;
-        mem_req_info.bindPoint   = bind_point;
-        mem_req_info.objectIndex = object_idx;
+            VmaAllocationCreateInfo allocation_create_info{};
+            allocation_create_info.usage          = VMA_MEMORY_USAGE_UNKNOWN;
```
fabian-lunarg: setting `VMA_MEMORY_USAGE_UNKNOWN` here, without setting `preferredFlags`,
will make VMA return the lowest compatible memory-type-index.

e.g. on my machine (linux/nvidia), replaying the optical-flow test:
the datagraph-session's memory 'could' be device-local, according to `compatible_memory_type_bits` 
(`0x3` -> type 1 is `DEVICE_LOCAL`), but it lands on type 0, which has no property-flags at all), but falls back to host-visible.
-> recommend setting `allocation_create_info.usage = VMA_MEMORY_USAGE_GPU_ONLY;` instead.
-> if not possible, VMA would fallback anyway

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -3782,263 +3796,206 @@ VulkanRebindAllocator::BindDataGraphPipelineSessionMemory(uint32_t bind_info_cou
...
-        VkDataGraphPipelineSessionBindPointTypeARM bind_point_type = {};
-        uint32_t                                   num_objects     = 0;
-    };
+    // Data graph session memory is created entirely from replay-device requirements when the replay session is
+    // created. Captured bind commands describe capture-device requirements, so consuming them must have no effect.
+    GFXRECON_UNREFERENCED_PARAMETER(bind_info_count);
+    GFXRECON_UNREFERENCED_PARAMETER(bind_infos);
+    GFXRECON_UNREFERENCED_PARAMETER(allocator_session_datas);
+    GFXRECON_UNREFERENCED_PARAMETER(allocator_memory_datas);
+    GFXRECON_UNREFERENCED_PARAMETER(bind_memory_properties);
```
fabian-lunarg, line 3805: `bind_memory_properties` is left untouched here, but later we write it into the object-table
unconditionally (`vulkan_replay_consumer_base.cpp:14568`) 
-> under `-m rebind` every session ends up with `memory_property_flags == 0`. 
we don't currently read it, but this would be silently wrong data there.

-> we should match captured bind-infos against the replay-generated bindings.
let me add a commit.

fabian-lunarg (approved):
approved from my side, 
glad if you could double-check my proposed additions/change @bartosz-muszarski-arm

framework/decode/vulkan_rebind_allocator.cpp
fabian-lunarg: ~~done~~ not done, not worth it

bartosz-muszarski-arm:
Looks good, my only concern is [Report data graph session memory properties under rebind](https://github.com/LunarG/gfxreconstruct/pull/3174/changes/b3e891f6b571e7913f744c15b052623b070b2890)
Looks like it implements additional tracking, handling and processing of data that in the end does not provide much value.
The added value would be debug-level cross-check of capture-time bindings against replay-time bindings, but:
1. Bindings are expected to differ between environments and rebind is meant to cover that
2. Even with all that bookkeeping in place it will still miss differences if:
 * replay env reports memory bind point the capture env never had
 * replay env reports the same bind point but larger numObjects
 * capture env had no bind points at all while replay env does
Even if its meant just for consistency - it makes sense for other resources in rebind, but DG pipeline follows a different bind model.
Looks to me like extra bookkeeping for weak diagnostics, not functional value.

fabian-lunarg:
> Looks good, my only concern is [Report data graph session memory properties under rebind](https://github.com/LunarG/gfxreconstruct/pull/3174/changes/b3e891f6b571e7913f744c15b052623b070b2890) Looks like it implements additional tracking, handling and processing of data that in the end does not provide much value. The added value would be debug-level cross-check of capture-time bindings against replay-time bindings, but:
> 
>     1. Bindings are expected to differ between environments and rebind is meant to cover that
> 
>     2. Even with all that bookkeeping in place it will still miss differences if:
> 
> 
>     * replay env reports memory bind point the capture env never had
> 
>     * replay env reports the same bind point but larger numObjects
> 
>     * capture env had no bind points at all while replay env does
>       Even if its meant just for consistency - it makes sense for other resources in rebind, but DG pipeline follows a different bind model.
>       Looks to me like extra bookkeeping for weak diagnostics, not functional value.

agreed, I'll drop the commit.

> data that in the end does not provide much value

true, way too much effort only to have an accurate value in a write-only field (`memory_property_flags`).

PR #3176, opened 2026-08-05
[Frame Looping] Basic timestamp query handling

nickdriscoll-lunarg:
This PR allows frame looping to handle `vkCmdWriteTimestamp` by omitting calls to it during loop repetitions.

### DIFF
--- a/framework/decode/vulkan_replay_frame_loop_consumer.cpp
+++ b/framework/decode/vulkan_replay_frame_loop_consumer.cpp
@@ -297,6 +297,18 @@ void VulkanReplayFrameLoopConsumer::Process_vkDestroyFence(const ApiCallInfo& ca
     }
 }
 
+void VulkanReplayFrameLoopConsumer::Process_vkCmdWriteTimestamp(const ApiCallInfo&       call_info,
+                                                                args::CmdWriteTimestamp& args)
+{
+    if (frame_loop_info_.IsRepetition())
+    {
+        // Skip writing timestamps during repetition to avoid "query not reset" errors
+        // if the capture doesn't reset them in the frame.
+        return;
+    }
+    VulkanReplayConsumer::Process_vkCmdWriteTimestamp(call_info, args);
+}
+
 void VulkanReplayFrameLoopConsumer::TrackFenceState(format::HandleId device, format::HandleId fence)
 {
     // If fence hasn't been seen yet, check and store the state it is in.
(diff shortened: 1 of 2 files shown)


antonio-lunarg (approved):

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -297,6 +297,18 @@ void VulkanReplayFrameLoopConsumer::Process_vkDestroyFence(const ApiCallInfo& ca
...
 }
 
+void VulkanReplayFrameLoopConsumer::Process_vkCmdWriteTimestamp(const ApiCallInfo&       call_info,
+                                                                args::CmdWriteTimestamp& args)
+{
+    if (frame_loop_info_.IsRepetition())
+    {
+        // Skip writing timestamps during repetition to avoid "query not reset" errors
+        // if the capture doesn't reset them in the frame.
+        return;
```
antonio-lunarg: Not sure if this is the right fix. 

I suppose there might issues with `vkGetQueryPoolResults`/`VK_QUERY_RESULT_WAIT_BIT`, waiting for a cmd write that has been skipped.

I think a better direction would be to reset the query pool at loop boundary, like we do for fences and events.

nickdriscoll-lunarg: I changed it so that now it records pool reset commands on `vkBeginCommandBuffer`.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -155,6 +155,31 @@ void VulkanReplayFrameLoopConsumer::Process_vkDestroyDescriptorPool(const ApiCal
...
+        format::HandleId         device        = cb_info->parent_id;
+        VkDevice                 replay_device = GetObjectInfoTable().GetVkDeviceInfo(device)->handle;
+        GFXRECON_ASSERT(replay_device != 0);
+        GetObjectInfoTable().VisitVkQueryPoolInfo([this, replay_device, cb_info](const VulkanQueryPoolInfo* info) {
+            GFXRECON_ASSERT(query_pool_sizes_.contains(info->capture_id));
+            const graphics::VulkanDeviceTable* device_table = GetDeviceTable(replay_device);
+            GFXRECON_ASSERT(device_table != nullptr);
+            VkQueryPool pool_handle = info->handle;
+            uint32_t    pool_size   = query_pool_sizes_[info->capture_id];
+            GFXRECON_LOG_INFO(
```
antonio-lunarg: I think it's better to use `GFXRECON_LOG_DEBUG` here.

antonio-lunarg (approved):

PR #3177, opened 2026-08-06
Support VK_EXT_device_memory_report

marius-pelegrin-arm:
The extension allows to set up a callback at device creation time. This commit propose to simply remove the callback at replay time, as it is not possible to restore the same callback, and a NULL callback is not allowed.

The current behavior is a SEGFAULT when the driver tries to call the NULL callback.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -3485,6 +3485,17 @@ void VulkanReplayConsumerBase::ModifyCreateDeviceInfo(
         }
     }
 
+    // Remove VK_EXT_device_memory_report callback structures if set.
+    // The callbacks cannot be restored anyway, so the easiest is to remove them entirely
+    if (graphics::vulkan_struct_remove_pnext<VkPhysicalDeviceDeviceMemoryReportFeaturesEXT>(&modified_create_info))
+    {
+        GFXRECON_LOG_WARNING("VkPhysicalDeviceDeviceMemoryReportFeaturesEXT was removed at device creation");
+    }
+    while (graphics::vulkan_struct_remove_pnext<VkDeviceDeviceMemoryReportCreateInfoEXT>(&modified_create_info))
+    {
+        GFXRECON_LOG_WARNING("VkDeviceDeviceMemoryReportCreateInfoEXT callback was removed at device creation");
+    }
+
     // Sanity checks depending on extension availability
     std::vector<VkExtensionProperties> available_extensions;
     if (graphics::feature_util::GetDeviceExtensions(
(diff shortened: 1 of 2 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -3485,6 +3485,13 @@ void VulkanReplayConsumerBase::ModifyCreateDeviceInfo(
         }
     }
 
+    // Remove VK_EXT_device_memory_report callback structure if set.
+    // The callback cannot be restored anyway, so the easiest is to remove it entirely
+    if (graphics::vulkan_struct_remove_pnext<VkDeviceDeviceMemoryReportCreateInfoEXT>(&modified_create_info))
```
fabian-lunarg: it appears `VkDeviceDeviceMemoryReportCreateInfoEXT` is one the rare cases that allow to be duplicated in the `pNext`-chain by spec.
meaning we should loop to purge all occurrences.

marius-pelegrin-arm: Ah yeah, didn't see this. I will push a fix.

fabian-lunarg:
now with the callback always stripped at replay ... `VK_EXT_device_memory_report` could/should be added to **ignorable** extensions:
-> add `VK_EXT_DEVICE_MEMORY_REPORT_EXTENSION_NAME` to `kIgnorableExtensions` (framework/graphics/vulkan_feature_util.cpp:45)

reason is: a capture from a supporting device would still fail in `vkCreateDevice` on devices not supported the EXT

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -3485,6 +3485,13 @@ void VulkanReplayConsumerBase::ModifyCreateDeviceInfo(
         }
     }
 
+    // Remove VK_EXT_device_memory_report callback structures if set.
+    // The callbacks cannot be restored anyway, so the easiest is to remove them entirely
+    while (graphics::vulkan_struct_remove_pnext<VkDeviceDeviceMemoryReportCreateInfoEXT>(&modified_create_info))
+    {
+        GFXRECON_LOG_WARNING("VkDeviceDeviceMemoryReportCreateInfoEXT callback was removed at device creation");
```
fabian-lunarg, line 3496: this would log duplicates for each removed struct.

bool removed_memory_report_info = false;
while(graphics::vulkan_struct_remove_pnext<VkDeviceDeviceMemoryReportCreateInfoEXT>(&modified_create_info))
{
    removed_memory_report_info = true;
}
if(removed_memory_report_info){GFXRECON_LOG_WARNING(...)}

fabian-lunarg: not sure how relevant though, maybe logging multiple times is fine/preferred

marius-pelegrin-arm: Or I can also just use "GFXRECON_LOG_WARNING_ONCE"
But I thought it would be better to let the multiple messages so that it can be found that there were multiple callbacks.

fabian-lunarg: in this case I think it's fine

marius-pelegrin-arm:
I will also remove `VkPhysicalDeviceDeviceMemoryReportFeaturesEXT` from the pNext chain, that was forgotten as well.

fabian-lunarg (approved):

PR #3180, opened 2026-08-06
android: Remove deprecated code

ncesario-lunarg:
Use `ALooper_pollOnce` instead of `ALooper_pollAll`.
`ALooper_pollAll` is removed in NDK versions >26.
Removing the deprecated code allows gfxreconstruct to be built with
newer compilers in newer NDKs.

This also contains a change from @MarkY-LunarG to consolidate which NDK version gfxreconstruct uses, rather than having the same NDK version copy/pasted for each gradle project.

This change was built locally with NDK 30.0.14904198, but not heavily tested.

### DIFF
--- a/framework/application/android_context.cpp
+++ b/framework/application/android_context.cpp
@@ -43,38 +43,40 @@ void AndroidContext::ProcessEvents(bool wait_for_input)
 {
     assert(application_);
     assert(android_app_);
-    // Process all pending events.
+
+    // Process all pending events. The first poll blocks when waiting for input; subsequent polls only drain events
+    // that are already pending.
     for (;;)
     {
-        int                         result = 0;
-        int                         events = 0;
-        struct android_poll_source* source = nullptr;
+        int                         events  = 0;
+        struct android_poll_source* source  = nullptr;
+        const int                   timeout = wait_for_input ? -1 : 0;
 
-        if (wait_for_input)
+        const int result = ALooper_pollOnce(timeout, nullptr, &events, reinterpret_cast<void**>(&source));
+        wait_for_input   = false;
+
+        // Unlike the deprecated ALooper_pollAll(), ALooper_pollOnce() returns to the caller after dispatching an event
+        // to a registered looper callback.
+        if (result == ALOOPER_POLL_CALLBACK)
         {
-            result         = ALooper_pollAll(-1, nullptr, &events, reinterpret_cast<void**>(&source));
-            wait_for_input = false;
-        }
-        else
-        {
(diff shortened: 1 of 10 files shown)


charles-lunarg (approved):
Can't say I fully grok the android changes, but there doesn't seem to be much to do to use `ALooper_pollOnce` instead of `ALooper_pollAll` (which just called `ALooper_pollOnce` internally anyhow).

ncesario-lunarg:
> Can't say I fully grok the android changes, but there doesn't seem to be much to do to use `ALooper_pollOnce` instead of `ALooper_pollAll` (which just called `ALooper_pollOnce` internally anyhow).

Yep, that's basically my understanding as well (i.e., pollAll was just calling pollOnce in a loop). If you have suggestions to make this code easier to understand, I'm happy to make them!

charles-lunarg:
> Yep, that's basically my understanding as well (i.e., pollAll was just calling pollOnce in a loop). If you have suggestions to make this code easier to understand, I'm happy to make them!

I found this github issue which has a nice summary:

> The fix, aiui, is to call ALooper_pollOnce, and break out of your loop for any of the results, not just ALOOPER_POLL_WAKE, since any other return value might also imply ALOOPER_POLL_WAKE.

It seems we do that already so there isn't much to do. Of course, if android starts deadlocking we have a good place to start looking

framework/application/android_context.cpp
```diff
@@ -43,38 +43,40 @@ void AndroidContext::ProcessEvents(bool wait_for_input)
...
+    // Process all pending events. The first poll blocks when waiting for input; subsequent polls only drain events
+    // that are already pending.
     for (;;)
     {
-        int                         result = 0;
-        int                         events = 0;
-        struct android_poll_source* source = nullptr;
+        int                         events  = 0;
+        struct android_poll_source* source  = nullptr;
+        const int                   timeout = wait_for_input ? -1 : 0;
```
antonio-lunarg, line 53: What if the application is paused?

ncesario-lunarg: Aha, I think that is a problem. Good catch! IIUC,

1. `ALooper_pollOnce(timeout = -1, ...)` returns `ALOOPER_POLL_CALLBACK `
2. `wait_for_input` = `false`
3.  `ALooper_pollOnce(timeout = 0, ...)` <- pretty sure `timeout` should be `-1` here in the paused case

I think a simple fix is to just move `wait_for_input = false`  to _after_  the `result == ALOOPER_POLL_CALLBACK` check.

Thanks for the review @antonio-lunarg!

antonio-lunarg (approved):

PR #3186, opened 2026-08-10
Combined fixes for replay and -m rebind

fabian-lunarg:
An RTX 5080 capture of a path-tracer failed to replay on an AMD Radeon 890M (RADV) under `-m rebind --remove-unsupported`. Fixing each failure exposed the next, four total, all of the same kind

-> capture-side data used directly against the replay device. 
-> With all four fixed the capture replays end-to-end, validation-clean.

commits are in the order the failures appeared and are independently reviewable.

1. `vkCreateDevice` — The captured `pQueueCreateInfos` were forwarded unmodified, then capture-side family indices indexed replay-sized arrays (used_queue_families[] in `VulkanRebindAllocator::Initialize`, `phys_dev_queue_props[]` in `PostCreateDeviceUpdateState`). 
-> ModifyCreateDeviceInfo now drops families the device lacks and clamps `queueCount`, reporting all adjustments in one warning.

 2. `vkQueueSubmit2` — segfault. With the create-info corrected, seven of the capture's nine `vkGetDeviceQueue` calls returned `VK_NULL_HANDLE`
 -> a queue-family is substituted when it was not created or when its replay `queueFlags` carry no graphics/compute/transfer support (-> was video-decode in this case). preserving the index would have been wrong. Covers `vkGetDeviceQueue`, `vkGetDeviceQueue2`, `vkCreateCommandPool`.

 3. Acceleration-structure build — assert on memory-type selection. VulkanAddressReplacer::create_buffer searched the capture device's memory-property using the replay device's memory_requirements -> fixed.
 
 4. Ray-generation shader — GPU page fault, VK_ERROR_DEVICE_LOST. 
 -> `spirv_parsing_util` recomputed block member offsets from a hand-rolled layout model that assumed std430.
-> offsets now come from the module's `OpMemberDecorate`, which is correct for `std140`, `std430` and `scalar` alike.
-> added unit-test covering a `std140` block with a vec3-wrapping struct to prevent regressions

also in this PR (review feedback, thanks @ziga-lunarg @MarkY-LunarG @wzchu-lunarg):
- track queue-counts per (family, flags) -> VUID-VkDeviceCreateInfo-queueFamilyIndex-02802.
- a substituted queue-family now also has to provide the requested `VkDeviceQueueCreateFlags`
- mark the injected `vkGetPhysicalDeviceQueueFamilyProperties` queries, log an error when returning a null queue
- record capture-time queue-family properties (new replay-overrides + codegen), try to match substitute against the captured capabilities
 
follow-up:
queue-family remapping here covers queue retrieval and command-pool creation, not concurrent-sharing pQueueFamilyIndices, ownership-transfer barriers, or surface-support queries.
 -> out of scope and not easily verifiable here, tracking-issue: https://github.com/LunarG/gfxreconstruct/issues/3187


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -3764,6 +3830,8 @@ VkResult VulkanReplayConsumerBase::PostCreateDeviceUpdateState(VulkanPhysicalDev
         device_info->enabled_queue_family_flags.queue_family_index_enabled[queue_create_info->queueFamilyIndex] = true;
         device_info->enabled_queue_family_flags.queue_family_properties_flags[queue_create_info->queueFamilyIndex] =
             phys_dev_queue_props[queue_create_info->queueFamilyIndex].queueFlags;
+        device_info->enabled_queue_family_flags.queue_family_queue_counts[queue_create_info->queueFamilyIndex] =
```
ziga-lunarg: This looks suspicious, and similar to something I've dealt with in #2795
`queue_create_info->queueFamilyIndex` is not a unique identifier, you can have multiple `VkDeviceQueueCreateInfo` with the same `queueFamilyIndex`, but different `flags`

fabian-lunarg: yes, agreed. that's here in the [spec](https://registry.khronos.org/vulkan/specs/latest/man/html/VkDeviceCreateInfo.html#VUID-VkDeviceCreateInfo-queueFamilyIndex-02802):
"The combination of the values in the queueFamilyIndex and flags members of each element of pQueueCreateInfos must be unique within pQueueCreateInfos"

I basically copied the behavior above, but the introduced assign into `queue_family_queue_counts`  and existing `queue_family_creation_flags` need to be keyed differently.
maybe a good time to fix that, when we're here already

fabian-lunarg: lookup/keying is fixed now, added another commit. 
also found another wrong assert and sprinkled some explanatory comments.
thanks @ziga-lunarg!

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4110,6 +4178,102 @@ VkResult VulkanReplayConsumerBase::OverrideEnumeratePhysicalDeviceGroups(
     return result;
 }
 
+uint32_t VulkanReplayConsumerBase::MapQueueFamilyIndex(const VulkanDeviceInfo* device_info, uint32_t queue_family_index)
```
ziga-lunarg: Should `MapQueueFamilyIndex` also take `flags`?

framework/decode/decoder_util.h
```diff
@@ -56,15 +56,24 @@ static VkQueue GetDeviceQueue(const graphics::VulkanDeviceTable* device_table,
...
-    const auto queue_family_flags =
-        device_info->enabled_queue_family_flags.queue_family_creation_flags.find(queue_family_index);
-    assert(queue_family_flags != device_info->enabled_queue_family_flags.queue_family_creation_flags.end());
+    const auto& queue_counts = device_info->enabled_queue_family_flags.queue_family_queue_counts;
+    const auto  family_entry = queue_counts.find(queue_family_index);
+    GFXRECON_ASSERT(family_entry != queue_counts.end());
+
+    if (family_entry == queue_counts.end() || family_entry->second.empty())
+    {
+        return queue;
```
MarkY-LunarG, line 66: Should we log a warning or info on this?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -3445,6 +3445,72 @@ void VulkanReplayConsumerBase::ModifyCreateDeviceInfo(
         replay_next          = replay_next->pNext;
     }
 
+    // The captured queue-create-infos refer to the capture-device's queue-families, which may not exist on the
+    // replay-device. Drop out-of-range families and clamp queue-counts, so that no invalid device is created.
+    {
+        uint32_t replay_queue_family_count = 0;
```
MarkY-LunarG: Do we need to wrap these with the injected call identifiers?

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4110,6 +4178,102 @@ VkResult VulkanReplayConsumerBase::OverrideEnumeratePhysicalDeviceGroups(
...
+
+    for (uint32_t i = 0; i < family_flags.queue_family_index_enabled.size(); ++i)
+    {
+        if (is_usable(i))
+        {
+            if (substitute == std::numeric_limits<uint32_t>::max())
+            {
+                substitute = i;
+            }
+            if (queue_flags(i) & VK_QUEUE_GRAPHICS_BIT)
```
MarkY-LunarG: Why only match graphics?  Shouldn't we match what the user wants?

framework/decode/vulkan_address_replacer.cpp
```diff
@@ -2028,9 +2028,13 @@ bool VulkanAddressReplacer::create_buffer(VulkanAddressReplacer::buffer_context_
...
-    // derive the capture-time(!) memory-index from memory_property_flags
-    const uint32_t memory_type_index = graphics::GetMemoryTypeIndex(
-        capture_memory_properties_, memory_requirements.memoryTypeBits, memory_property_flags);
+    // derive the capture-time(!) memory-index from memory_property_flags.
+    // memory_requirements.memoryTypeBits is a replay-device mask and must not be used to filter capture-device
+    // memory-types. this index is only used to look up its property-flags, the actual replay-device memory-type
+    // is picked by the rebind-allocator when the buffer is bound.
+    constexpr uint32_t any_memory_type = std::numeric_limits<uint32_t>::max();
+    const uint32_t     memory_type_index =
+        graphics::GetMemoryTypeIndex(capture_memory_properties_, any_memory_type, memory_property_flags);
```
wzchu-lunarg, line 2037: Since `memory_requirements.memoryTypeBits` is swapped for `any_memory_type`, would it be possible that `GetMemoryTypeIndex` chooses a memory type index that is over the memory type count of the replay device?

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: first I was thinking no, because the usability-test doesn't require the flags.
but further down there is a real gap: when we substitute a queue-family we could/should
check if the requested flags are supported there too. adding a commit for that

framework/decode/decoder_util.h
fabian-lunarg: yes, you're right. actually a LOG_ERROR would be justified here:

vulkan_captured_swapchain.cpp:384 uses that queue for a layout transition,
screenshot_handler.cpp:152 for a copy.

so that would already be VVL-error and/or a crash further down the line.
-> adding a LOG_ERROR here

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: yes, we should. at least for newly added calls this should be done here in this PR. 
there are also close, pre-existing places that would require it, which would be best patched in a dedicated PR.

framework/decode/decoder_util.h
fabian-lunarg: done

framework/decode/vulkan_replay_consumer_base.cpp

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: yes, agree. that would require just some additional wiring:
matching "what the user wants" boils down to recording the capture's `vkGetPhysicalDeviceQueueFamilyProperties[2]` which we basically have in the capture but don't forward here. 
-> it 'would' be useful though.

the reasoning wrt. "Why only match graphics"? was this:
it's a heuristic based on this spec-guarantee that at least one family supports both graphics and compute.
and graphics/compute implies transfer. it's not exactly matching intent, rather compatibility atm.

-> looking into how to use the information we got from `vkGetPhysicalDeviceQueueFamilyProperties[2]` here,
seems this requires additional override(s) for `vkGetPhysicalDeviceQueueFamilyProperties[2][KHR]`

fabian-lunarg: done. added another commit:

- overrides for `vkGetPhysicalDeviceQueueFamilyProperties[2][KHR]`
-> keep track of captured queue-family properties
-> "match what the user wants", fallback to `VK_QUEUE_GRAPHICS_BIT` when unknown

framework/decode/vulkan_address_replacer.cpp
fabian-lunarg: no, that value does never reach the replay device. 
since this code-path is `-m rebind` only, the resulting `memory_type_index`
ends up as `memory_alloc_info->original_index` ... in RebindAllocator [here](https://github.com/LunarG/gfxreconstruct/blob/ccbf158dff240fd0e201bd6929546adb52df79d9/framework/decode/vulkan_rebind_allocator.cpp#L580)
so, that's fine and replay memory type correctly handled by VMA down the line.

MarkY-LunarG (approved):

PR #3188, opened 2026-08-10
loop: Track and fix up semaphore states

antonio-lunarg:
## Problem

The frame loop replays one frame range many times. Semaphore state does not stay inside the range. Three failures arise on repetition.

- A submit in the range signals a binary semaphore. The matching wait comes after the range, so the loop never replays it. The semaphore stays signaled. The next iteration signals it again.
- A submit in the range waits on a binary semaphore. The matching signal came before the range. Iteration 1 consumes the signal. Iteration 2 waits again, but no signal comes. The replay stops.
- Timeline counters only increase. The capture recorded absolute signal values. In iteration 2 these values are not larger than the counter.

### Binary semaphores

Vulkan has no function that reads binary semaphore state. Because of this, the consumer learns the state from usage during the first iteration only.

- A signal puts the semaphore in the "signaled in loop" set.
- A wait removes the semaphore from that set. If the set does not contain the semaphore, the signal came from before the loop. The consumer then puts the semaphore in the "signaled before loop" set.

The two sets do not change after the first iteration. At each loop boundary, the consumer resets their state as it was when starting the loop.

### Timeline semaphores

A timeline counter cannot decrease, so the consumer cannot restore a value.
It replaces the semaphore object instead:

1. At loop start, the consumer reads each counter with vkGetSemaphoreCounterValue and stores the value. For a semaphore created inside the range, it stores the initial value from the create info.
2. At each boundary, the consumer replaces the semaphore with a new semaphore with the appropriate initial value as stored from step 1.

### DIFF
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -9602,6 +9602,32 @@ VulkanReplayConsumerBase::OverrideQueuePresentKHR(PFN_vkQueuePresentKHR         
     return result;
 }
 
+VkResult VulkanReplayConsumerBase::OverrideCreateSemaphore(
+    PFN_vkCreateSemaphore                                      func,
+    VkResult                                                   original_result,
+    const VulkanDeviceInfo*                                    device_info,
+    const StructPointerDecoder<Decoded_VkSemaphoreCreateInfo>* pCreateInfo,
+    const StructPointerDecoder<Decoded_VkAllocationCallbacks>* pAllocator,
+    HandlePointerDecoder<VkSemaphore>*                         pSemaphore)
+{
+    auto* semaphore_info = reinterpret_cast<VulkanSemaphoreInfo*>(pSemaphore->GetConsumerData(0));
+
+    const auto* type_info =
+        GetPNextMetaStruct<Decoded_VkSemaphoreTypeCreateInfo>(pCreateInfo->GetMetaStructPointer()->pNext);
+
+    if (type_info != nullptr && type_info->decoded_value != nullptr &&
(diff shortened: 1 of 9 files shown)


nickdriscoll-lunarg (approved):

davidlunarg (approved):
LGTM

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -9602,6 +9602,32 @@ VulkanReplayConsumerBase::OverrideQueuePresentKHR(PFN_vkQueuePresentKHR
...
+    const StructPointerDecoder<Decoded_VkSemaphoreCreateInfo>* pCreateInfo,
+    const StructPointerDecoder<Decoded_VkAllocationCallbacks>* pAllocator,
+    HandlePointerDecoder<VkSemaphore>*                         pSemaphore)
+{
+    auto* semaphore_info = reinterpret_cast<VulkanSemaphoreInfo*>(pSemaphore->GetConsumerData(0));
+
+    const auto* type_info =
+        GetPNextMetaStruct<Decoded_VkSemaphoreTypeCreateInfo>(pCreateInfo->GetMetaStructPointer()->pNext);
+
+    if (type_info != nullptr && type_info->decoded_value != nullptr &&
```
jzulauf-lunarg, line 9618: Why are we only looking at the first link in the pnext chain?

jzulauf-lunarg (commented):
One open queston.

framework/decode/vulkan_replay_consumer_base.cpp
antonio-lunarg: We are not: `GetPNextMetaStruct()` follows the `pNext` chain until it finds the right `sType` or `nullptr`.

jzulauf-lunarg (approved):
LGTM

PR #3189, opened 2026-08-10
Fix DxgiAdapterDesc layout mismatch causing gfxrecon-info to fail on Linux

tpower-amd:
## Problem

`gfxrecon-info` fails to correctly read D3D12 trace metadata when run on Linux. `DxgiAdapterDesc::Description` is declared as `wchar_t[kAdapterDescriptionSize]`, but `wchar_t` is 4 bytes on Linux and 2 bytes on Windows. When reading a trace captured on Windows, the Description field occupies twice the expected bytes on disk, corrupting the layout of all subsequent fields in the struct (`VendorId`, `DeviceId`, memory sizes, LUID, etc.).

## Solution

Guard the field declaration with a platform `#ifdef`: use `char16_t` on Linux and retain `wchar_t` on Windows. Change is in `framework/format/format.h` only — no changes to serialization or consumer code are required as the on-disk representation is unchanged.

## Testing

- ✅ Test suites all passed: `gfxrecon_util_test`, `gfxrecon_encode_test`, `gfxrecon_decode_test`, `gfxrecon_format_test`, `gfxrecon_framework_test`, `gfxrecon_graphics_test`, `gfxrecon_replay_event_plugin_loader_test`.

### DIFF
--- a/USAGE_android.md
+++ b/USAGE_android.md
@@ -822,16 +822,16 @@ options:
                         by <file>
   --version             Print version information and exit (forwarded to
                         replay tool)
-  --log-level LEVEL     Specify highest level message to log. Options are:
-                        debug, info, warning, error, and fatal. Default is
+  --log-level LEVEL     Specify lowest level message to log. Options are:
+                        fatal, error, warning, info, and debug. Default is
                         info. (forwarded to replay tool)
   --log-timestamps      Output a timestamp in front of each log message.
   --log-file DEVICE_FILE
                         Write log messages to a file at the specified path
                         instead of logcat (forwarded to replay tool)
   --debug-messenger-level LEVEL
-                        Specify highest debug messenger severity level.
-                        Options are: debug, info, warning, and error. Default
+                        Specify lowest debug messenger severity level.
+                        Options are: error, warning, info, and debug. Default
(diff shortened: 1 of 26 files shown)


framework/format/format.h
```diff
@@ -665,7 +665,13 @@ struct InitDx12AccelerationStructureGeometryDesc
 
 struct DxgiAdapterDesc
 {
+#if defined(__linux__)
+    // wchar_t is 4 bytes on Linux but 2 bytes on Windows. Use char16_t to match
+    // the on-disk format of traces captured on Windows and keep struct layout consistent.
+    char16_t Description[kAdapterDescriptionSize];
```
davidd-lunarg: Let's use `char16_t` for all platforms and skip the preprocessor `#if`. GFXR needs to ensure this struct is always serialized with the exact number of bytes.

davidd-lunarg (changes requested):

tpower-amd:
@davidd-lunarg PR was created incorrectly on fork's `dev` branch, pushed to feature branch with requested changes and recreated PR here: https://github.com/LunarG/gfxreconstruct/pull/3236

PR #3191, opened 2026-08-11
Complete migration to vulkan injected device calls

panos-lunarg:


framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -2336,6 +2374,9 @@ VkResult VulkanRebindAllocator::UpdateMappedMemoryRanges(
                 VkDeviceSize range_start = memory_ranges[i].offset;
                 VkDeviceSize range_end   = range_start + size;
 
+                // Mark vma's api calls as synthesized
+                auto injected = device_table_.Open();
```
fabian-lunarg: better placed at the top of this scope (avoid repeated open/close inside the loop).

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -3934,7 +3985,8 @@ VkResult VulkanRebindAllocator::InitializeDataGraphPipelineSessionMemory(VkDataG
             memory_info.alc_create_info                    = allocation_create_info;
             memory_info.offset_from_original_device_memory = 0;
 
-            result = vmaAllocateMemory(allocator_,
+            auto injected = device_table_.Open();
```
fabian-lunarg: this should be moved up, for two reasons:

- inside loop: causes repeated open/close
- it needs to be above 3924, so it covers the Dtors of `std::vector<std::unique_ptr<PendingBinding>>             pending_bindings;`

framework/graphics/vulkan_injected_calls.cpp
```diff
@@ -67,14 +68,21 @@ void VulkanInjectedDeviceCalls::Scope::InsertLabel(VkCommandBuffer command_buffe
 }
 
 VkResult VulkanInjectedDeviceCalls::Scope::BeginCommandBuffer(VkCommandBuffer                 command_buffer,
-                                                              const VkCommandBufferBeginInfo* begin_info) const
+                                                              const VkCommandBufferBeginInfo* begin_info,
+                                                              const char*                     label) const
 {
     VkResult result = table_->BeginCommandBuffer(command_buffer, begin_info);
 
     if (result == VK_SUCCESS)
```
fabian-lunarg: this block now creates a couple (tmp) string-objects (mem-allocations),
even when the feature is disabled.

we should check for: `if (result == VK_SUCCESS && GetAnnotateInjectedCommands()){...}`

framework/graphics/vulkan_injected_calls.h
```diff
@@ -72,12 +72,14 @@ class VulkanInjectedDeviceCalls
 
         // Emits a single "GFXR Replay: <category>" label into command_buffer.
         // No-op when annotation is disabled or debug utils is unavailable.
-        void InsertLabel(VkCommandBuffer command_buffer, const char* category) const;
+        void InsertLabel(VkCommandBuffer command_buffer, const std::string& category) const;
```
fabian-lunarg: the old signature was better. 
now it is required to create a string, before `InsertLabel` can check for `GetAnnotateInjectedCommands()`. overall this causes unnecessary memory-allocations on hot code-path (state-init!)

framework/decode/vulkan_replay_dump_resources_common.cpp
```diff
@@ -1595,7 +1517,7 @@ static VkResult SerializeAccelerationStructure(AccelerationStructureDumpResource
                                                  acceleration_structure->serialized_data.buffer };
 
         VkDeviceOrHostAddressKHR device_address;
-        device_address.deviceAddress = device_table.GetBufferDeviceAddressKHR(device, &bdai);
+        device_address.deviceAddress = injected->GetBufferDeviceAddressKHR(device, &bdai);
```
fabian-lunarg: I know this was pre-existing, but here it's missing the proper branching:

```
const PFN_vkGetBufferDeviceAddress get_buffer_device_address =
    phys_dev_info->parent_info.api_version >= VK_API_VERSION_1_2
        ? injected->GetBufferDeviceAddress
        : injected->GetBufferDeviceAddressKHR;
```
(equivalent to e.g. `framework/decode/vulkan_replay_dump_resources_as.cpp:499`)

framework/decode/vulkan_temporary_objects.h
```diff
@@ -145,13 +145,34 @@ struct TemporaryCommandBuffer
...
+        query_pool(VK_NULL_HANDLE), device(dev), device_table(dev_table)
+    {}
+
+    ~TemporaryQueryPool();
+
+    VkResult Create(uint32_t query_count);
+
+    VkQueryPool                                query_pool;
+    VkDevice                                   device;
+    const graphics::VulkanInjectedDeviceCalls& device_table;
```
fabian-lunarg: the others (`TemporaryFence` and `TemporaryCommandBuffer`) store by value:
`graphics::VulkanInjectedDeviceCalls device_table;`

better to be consistent here and avoid potential ownership issues

framework/decode/vulkan_rebind_allocator.cpp
panos-lunarg: Done

framework/decode/vulkan_rebind_allocator.cpp
panos-lunarg: Done

framework/graphics/vulkan_injected_calls.cpp
panos-lunarg: Done

framework/graphics/vulkan_injected_calls.h
panos-lunarg: Done

framework/decode/vulkan_replay_dump_resources_common.cpp
panos-lunarg: Same problem appears in multiple places. I'll try to address it

framework/decode/vulkan_temporary_objects.h
panos-lunarg: Done

fabian-lunarg (approved):
@panos-lunarg I had nothing else to brag here
and the remaining (pre-existing) issue isn't blocking for me.

framework/decode/vulkan_replay_dump_resources_common.cpp
panos-lunarg: Pushed one more commit which handles this.

PR #3194, opened 2026-08-11
Fix Wayland replay size when fractional scaling is enabled on a monitor

MarkY-LunarG:
The replay window used wl_surface::set_buffer_scale with the integer scale from wl_output. On a fractionally scaled desktop (125%, 150%) compositors report 2 to clients that do not speak wp_fractional_scale_v1, so the window claimed twice the logical size it needed and appeared at roughly half size.
    
Bind wp_viewporter and wp_fractional_scale_manager_v1, leave the buffer at scale 1, and set a viewport destination of pixels divided by the preferred scale. One buffer pixel then lands on one physical pixel. Compositors that advertise neither protocol keep the old integer-scale path unchanged.

To reproduce issue, boot into Gnome with Wayland enabled (my system was Fedora 44), under Settings, Display, select your monitor and set "Scale" to some value other than 100%.  I tested with 125%.  Before this fix, the replay swapchain was 1/2 the size= it should have been compared to the capture itself.

### DIFF
--- a/framework/generated/generated_wayland_xdg_shell.h
+++ b/framework/generated/generated_wayland_xdg_shell.h
@@ -24,6 +24,7 @@
 #ifndef GFXRECON_GENERATED_WAYLAND_XDG_SHELL_H
 #define GFXRECON_GENERATED_WAYLAND_XDG_SHELL_H
 
+#include <cstdint>
 #include <vector>
 
 #include <wayland-client.h>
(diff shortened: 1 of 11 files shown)


charles-lunarg:
This isn't working like I would expect.

Here is vkcube being captured with the monitor at 100% scaling
image

And here is the same vkcube being captured with the monitor at 125% scaling
image

We can clearly see vkcube is larger.

However, when I *replay* vkcube at 125% scaled, it is the same size at the 100% scaled monitor output.
image

I went up to 200% scaled monitor and gfxrecon-replay of vkcube still renders at the 100% scaled output.
image

MarkY-LunarG:
For me, the main issue is if you have scaling enabled, then when you capture the replay should look similar.  The problem was that when you replayed, we didn't take into account the scaling so it drew without the scaling so the native resolution.  However, all other apps treat the scaling like an adjustment in resolution so we actually needed to change the rendered resolution result for the replay.

The goal is what it looks like in capture, it looks like in replay.

charles-lunarg:
After looking into it further, I understand the confusion.

The "bug" is that replay would render to a sub-100% scaled image. With the change, it renders to a 100% scaled output, regardless of the scaling factor of the monitor used during capture.

My initial assumption was that if I captured on a 200% scaled monitor, I should see replay with a 200% scaled output. Instead I got 100% when using the PR. Hence the confusion. Without the PR, the output is at 50% like you described in the reproduction steps.

So the real question is: Should gfxrecon-replay try to 'behave' like a normal application when replaying on a scaled monitor, such that the output appears bigger? Or should we try to replay at 100% (like the PR currently does) so that the compositor doesn't scale the image before presentation? I think the latter is fine, we just need to be a bit more clear that we are intentionally negating the scaling that wayland is telling us is occurring.

MarkY-LunarG:
<So the real question is: Should gfxrecon-replay try to 'behave' like a normal application when replaying on a scaled monitor, such that the output appears bigger? Or should we try to replay at 100% (like the PR currently does) so that the compositor doesn't scale the image before presentation? I think the latter is fine, we just need to be a bit more clear that we are intentionally negating the scaling that wayland is telling us is occurring.

@bradgrantham-lunarg , thoughts?

charles-lunarg (changes requested):
Running the code locally, I see that the changes fix the issue as presented.
My initial confusion (and all those screenshots) is from an assumption about what gfxrecon-replay *should* do, which isn't what we ultimately want to do.

> so the window occupies the same area of the display that it did on the machine the capture came from.

This is what confused me. What I see during capture isn't what is recorded into the gfxr file, because what I see is the compositor output, not the gfxr file contents. While it _may_ be possible to capture the scale factors into the gfxr, that isn't really useful for replay. We are capturing what Vulkan rendered, not how wayland upscaled or downscaled the image. 

So my request is to simply clarify that we are aiming to render the output at 100% scale of what Vulkan captured, not the 'area of the display it was on'. 

Maybe there should be a replay option that lets the compositor scale the output as it sees fit, but we definitely need this PR to at least get the output to be 1-1 from what is in the gxfr file.

charles-lunarg (approved):

framework/generated/generated_wayland_xdg_shell.h
```diff
@@ -31,477 +31,477 @@
 #include "util/defines.h"
 #include "util/wayland_loader.h"
 
-struct xdg_wm_base;
-struct xdg_positioner;
-struct xdg_surface;
-struct xdg_toplevel;
-struct xdg_popup;
+struct GfxrXdgWmBase;
```
bradgrantham-lunarg: I think you're going the right direction with this.

In my experience having a whole bunch of structs and enums and types with a leading prefix is a signal that they should be in a namespace with the prefix removed.  Since you're renaming these ostensibly with the goal of avoiding collisions, the way to do that in C++ is to use a namespace, so please put these in a namespace, like `gfxrecon::util` and remove the `Gfxr`.  You'll also need to make the enums "enum class blah" so they don't escape the namespace and with ": uint32_t" if they're going to be passed to uint32_t

framework/generated/generated_wayland_xdg_shell.h
```diff
@@ -31,477 +31,477 @@
...
+enum GfxrXdgWmBaseError
 {
-	XDG_WM_BASE_ERROR_ROLE = 0,
-	XDG_WM_BASE_ERROR_DEFUNCT_SURFACES = 1,
-	XDG_WM_BASE_ERROR_NOT_THE_TOPMOST_POPUP = 2,
-	XDG_WM_BASE_ERROR_INVALID_POPUP_PARENT = 3,
-	XDG_WM_BASE_ERROR_INVALID_SURFACE_STATE = 4,
-	XDG_WM_BASE_ERROR_INVALID_POSITIONER = 5,
-	XDG_WM_BASE_ERROR_UNRESPONSIVE = 6,
+    GFXR_XDG_WM_BASE_ERROR_ROLE = 0,
```
bradgrantham-lunarg: I think you're going the right direction with this.

In my experience having a whole bunch of enumerants with the a prefix that is the same as the enum type is a signal to remove the type from the enumerant (e.g. if you use a namespace like I request above, a segment of code `using gfxrecon::util` will have `xdg::BaseError::DEFUNCT_SURFACES`), so I request also extracting the type from the enumerant as well as removing the leading `GfxrXdg` from the enum type itself.

bradgrantham-lunarg (changes requested):

bradgrantham-lunarg:
> <So the real question is: Should gfxrecon-replay try to 'behave' like a normal application when replaying on a scaled monitor, such that the output appears bigger? Or should we try to replay at 100% (like the PR currently does) so that the compositor doesn't scale the image before presentation? I think the latter is fine, we just need to be a bit more clear that we are intentionally negating the scaling that wayland is telling us is occurring.
> 
> @bradgrantham-lunarg , thoughts?

Best I can tell this is a cosmetic PR, and as that it seems fine.

We don't capture Wayland, so we can't replicate the capture environment during replay, so anything we do here is essentially arbitrary as far as I can tell.  This just helps someone not be jarred by capturing vkcube and non-scaling-aware apps and then immediately replaying that content on the same device.  If the app is scaling-aware, then replay will be incorrect for some cases no matter what we do as long as we don't know what the original app specified to the compositor.  I think dealing with scaling, screen resolution, etc is complicated and is out of scope for us at this time; we have many other higher priorities.

framework/generated/generated_wayland_xdg_shell.h
MarkY-LunarG: Moved enumerates to enum class, also moved to use `gfxrecon::util` namespace.

MarkY-LunarG:
> We don't capture Wayland, so we can't replicate the capture environment during replay, so anything we do here is essentially arbitrary as far as I can tell. This just helps someone not be jarred by capturing vkcube and non-scaling-aware apps and then immediately replaying that content on the same device. If the app is scaling-aware, then replay will be incorrect for some cases no matter what we do as long as we don't know what the original app specified to the compositor. I think dealing with scaling, screen resolution, etc is complicated and is out of scope for us at this time; we have many other higher priorities.

Ok, I'll just close this for now then.

MarkY-LunarG:
@bradgrantham-lunarg , I am actually seeing a crash now without this change.  I think it has to do with certain desktop resolutions + scaling in Fedora 44 with Wayland at least.  I've re-opened this because this is no longer just a cosmetic change.

PR #3195, opened 2026-08-12
Support VkImageFormatListCreateInfo for vkCreateSwapchainKHR

locke-lunarg:
If the `CreateSwapchainKHR` has `VkImageFormatListCreateInfo`, the swapchain could support multiple view formats. The view format doesn't have to be the same as the swapchain image format.

### DIFF
--- a/framework/decode/vulkan_object_info.h
+++ b/framework/decode/vulkan_object_info.h
@@ -469,7 +469,7 @@ struct VulkanImageInfo : public VulkanObjectInfo<VkImage>
 {
     std::unordered_map<uint32_t, size_t> array_counts;
 
-    bool is_swapchain_image{ false };
+    format::HandleId swapchain_id{ format::kNullHandleId };
 
     // The following values are only used for memory portability.
     VulkanResourceAllocator::ResourceData allocator_data{ 0 };
@@ -639,6 +639,7 @@ struct VulkanSwapchainKHRInfo : public VulkanObjectInfo<VkSwapchainKHR>
     uint32_t             width{ 0 };
     uint32_t             height{ 0 };
     VkFormat             format{ VK_FORMAT_UNDEFINED };
+    std::vector<VkFormat> supported_view_formats; // Based on VkImageFormatListCreateInfo
     std::vector<VkImage> images; // This image could be virtual or real according to if it uses VirtualSwapchain.
     std::unordered_map<uint32_t, size_t> array_counts;
 
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -8688,6 +8688,13 @@ VkResult VulkanReplayConsumerBase::OverrideCreateSwapchainKHR(
     swapchain_info->height             = modified_create_info.imageExtent.height;
     swapchain_info->format             = modified_create_info.imageFormat;
 
+    if (const auto* format_list = graphics::vulkan_struct_get_pnext<VkImageFormatListCreateInfo>(replay_create_info);
+        format_list != nullptr && format_list->viewFormatCount > 0)
+    {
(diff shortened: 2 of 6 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -8688,6 +8688,26 @@ VkResult VulkanReplayConsumerBase::OverrideCreateSwapchainKHR(
     swapchain_info->height             = modified_create_info.imageExtent.height;
     swapchain_info->format             = modified_create_info.imageFormat;
 
+    swapchain_info->supported_view_formats.clear();
+    auto* pnext = modified_create_info.pNext;
+    while (pnext != nullptr)
```
fabian-lunarg: I checked and it looks `VkImageFormatListCreateInfo` must not appear multiple times in any pNext-chain.
-> that means we can safely use our helper `if (auto* format_list_ci =
                graphics::vulkan_struct_get_pnext<VkImageFormatListCreateInfo>(modified_create_info)){...}`

framework/decode/vulkan_object_info.h
```diff
@@ -469,7 +469,8 @@ struct VulkanImageInfo : public VulkanObjectInfo<VkImage>
 {
     std::unordered_map<uint32_t, size_t> array_counts;
 
-    bool is_swapchain_image{ false };
+    bool             is_swapchain_image{ false };
+    format::HandleId swapchain_id{ format::kNullHandleId };
```
fabian-lunarg, line 472: this is slightly redundant+error-prone now.
`is_swapchain_image` can now be deduced from checking `swapchain_id != format::kNullHandleId`,
possibly as simple getter. this would avoid the hazard to keep both variables in sync.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -11456,19 +11478,38 @@ VkResult VulkanReplayConsumerBase::OverrideCreateImageView(
...
         }
     }
-    else if (img_info->is_swapchain_image && img_info->format != modified_create_info.format)
+    else if (img_info->is_swapchain_image)
     {
-        // for swapchain-images set image-view to a fallback format, avoid issues with distorted HDR/SRGB colors.
-        // when the swapchain fell back to a BGRA8 format we preserve the captured view's sRGB-ness; for any other
-        // fallback format we match the swapchain image format exactly to keep the view format-compatible.
-        if (img_info->format == VK_FORMAT_B8G8R8A8_UNORM || img_info->format == VK_FORMAT_B8G8R8A8_SRGB)
+        GFXRECON_ASSERT(img_info->swapchain_id != format::kNullHandleId);
```
fabian-lunarg: already mentioned above, but here it shows exactly what I mean:

we check `if (img_info->is_swapchain_image)` and still need a `GFXRECON_ASSERT(img_info->swapchain_id != format::kNullHandleId);`

`is_swapchain_image` could be dropped (== `swapchain_id != format::kNullHandleId`) (or become an inline getter)

framework/decode/vulkan_object_info.h
fabian-lunarg: ok I see, didn't consider the overlap, no concern

framework/decode/vulkan_replay_consumer_base.cpp
locke-lunarg: Thanks. Fixed.

framework/decode/vulkan_object_info.h
locke-lunarg: Thanks. Fixed.

framework/decode/vulkan_replay_consumer_base.cpp
locke-lunarg: Thanks. Fixed.

fabian-lunarg (approved):

PR #3196, opened 2026-08-12
util: Fix Android path variables for multi-user and add ExternalDataPath

olehkuznetsov:
### Problem
The `${InternalDataPath}` variable expansion on Android was hardcoded to `/data/data/<package_name>`. In multi-user environments (e.g., when running under a secondary user profile like User 10), `/data/data` (which symlinks to `/data/user/0`) is not writeable by the secondary user due to permission constraints. This causes file creation failures when trying to save captures to `${InternalDataPath}`.

### Solution
1.  **Dynamic `${InternalDataPath}`:** Changed the expansion to dynamically resolve the current Android user ID using `getuid() / 100000` (matching Android's internal `UserHandle.getUserId` logic). The path now resolves to `/data/user/<user_id>/<package_name>`, which correctly handles secondary users.
2.  **Added `${ExternalDataPath}`:** Introduced a new `${ExternalDataPath}` variable that resolves to `/storage/emulated/<user_id>/Android/data/<package_name>/files`. This allows users to configure capture output to external storage, making it easily accessible/pullable without root permissions.

### Verification
*   Updated `TEST_CASE("ExpandPathVariables")` in `framework/util/test/main.cpp` to verify both dynamic paths.
*   Verified that the project builds successfully for Android (`arm64-v8a` and `x86_64`).

### DIFF
--- a/framework/util/file_path.cpp
+++ b/framework/util/file_path.cpp
@@ -599,7 +599,6 @@ enum PathVariable : uint32_t
 {
     kAppName,
     kInternalDataPath,
-    kExternalDataPath,
 
     kMaxValue,
 };
@@ -611,7 +610,6 @@ std::vector<const char*> GetPathVariables()
 
     path_variables[PathVariable::kAppName]          = "${AppName}";
     path_variables[PathVariable::kInternalDataPath] = "${InternalDataPath}";
-    path_variables[PathVariable::kExternalDataPath] = "${ExternalDataPath}";
 
     return path_variables;
 }
@@ -619,9 +617,9 @@ std::vector<const char*> GetPathVariables()
 #ifdef __ANDROID__
 // Android user ID is determined by dividing UID by 100,000 (PER_USER_RANGE).
 // We resolve this dynamically to support multi-user profiles.
-static uid_t GetAndroidUserId()
+uid_t GetAndroidUserId()
 {
-    return getuid() / 100000;
+    return getuid() / kPerUserRange;
 }
 #endif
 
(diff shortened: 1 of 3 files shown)


framework/util/file_path.cpp
```diff
@@ -636,7 +647,20 @@ std::string ExpandPathVariables(const FileInfo& info, const std::string& path)
...
+                }
+#else
+                    GFXRECON_LOG_WARNING(
+                        "Unimplemented path variable pattern: %s. This pattern is only supported on Android.", pattern);
+#endif
+                break;
+                case PathVariable::kExternalDataPath:
+#ifdef __ANDROID__
+                {
+                    replacement = "/storage/emulated/" + std::to_string(GetAndroidUserId()) + "/Android/data/" +
```
antonio-lunarg: What if this directory does not exist? I guess we either create the parent directory chain or fail with a message.

framework/util/file_path.cpp
```diff
@@ -610,10 +611,20 @@ std::vector<const char*> GetPathVariables()
 
     path_variables[PathVariable::kAppName]          = "${AppName}";
     path_variables[PathVariable::kInternalDataPath] = "${InternalDataPath}";
+    path_variables[PathVariable::kExternalDataPath] = "${ExternalDataPath}";
```
antonio-lunarg: It's cool to see a new path variable here! Needs to be added to the docs (`USAGE_android.md`).

framework/util/file_path.cpp
```diff
@@ -610,10 +611,20 @@ std::vector<const char*> GetPathVariables()
...
 
     return path_variables;
 }
 
+#ifdef __ANDROID__
+// Android user ID is determined by dividing UID by 100,000 (PER_USER_RANGE).
+// We resolve this dynamically to support multi-user profiles.
+static uid_t GetAndroidUserId()
+{
+    return getuid() / 100000;
```
antonio-lunarg: `100000` can be replaced with a constant named `kPerUserRange`, right?

antonio-lunarg (approved):

PR #3197, opened 2026-08-12
Fix use-after-free bug in state tracker

panos-lunarg:
resource_utils_ needs to be maintained when devices are destroyed.


dfriederich (approved):
Fixes the reported crash :-)

fabian-lunarg (approved):

PR #3198, opened 2026-08-13
optimize: Fix removing creation calls of referenced pipelines

panos-lunarg:
Command buffers (CB) that were never submitted but freed were retained in the capture file because their ids were removed from the live objects hence they were never been scanned for whethere they were referenced or not.

On the other hand pipeline objects that were referenced by such a CB were considered as unreferenced (the CB was never submitted) and their creation blocks were removed. This caused crashes as the vkCmdBindPipeline commands were replayed.

This PR fixes this issue. Note that this does not apply to other tracked objects (images, buffers, etc) as the optimizer removes their Init metacommands, not their creation blocks.

### DIFF
--- a/framework/decode/referenced_resource_table.cpp
+++ b/framework/decode/referenced_resource_table.cpp
@@ -135,6 +135,13 @@ void ReferencedResourceTable::AddResourceToUser(format::HandleId user_id, format
                 {
                     user_info->resource_infos.emplace(child_resource_id, child_resource_info);
                 }
+
+                // A user that has already been submitted keeps all of its recorded blocks in the optimized file,
+                // even if the current recording is never submitted, so the resource must be treated as referenced.
+                if (user_info->used)
+                {
+                    MarkResourceUsed(resource_info.get());
+                }
             }
         }
     }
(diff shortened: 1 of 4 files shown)


framework/decode/referenced_resource_table.cpp
```diff
@@ -135,6 +135,13 @@ void ReferencedResourceTable::AddResourceToUser(format::HandleId user_id, format
                 {
                     user_info->resource_infos.emplace(child_resource_id, child_resource_info);
                 }
+
+                // A user that has already been submitted keeps all of its recorded blocks in the optimized file,
+                // even if the current recording is never submitted, so the resource must be treated as referenced.
+                if (user_info->used)
+                {
+                    MarkResourceUsed(resource_info.get());
```
fabian-lunarg, line 143: I think the same should be done in `AddContainerToUser`.

-> resources reached through a descriptor set still get classified unreferenced (when bound into an already-submitted CB that isn't resubmitted).

panos-lunarg: Don't think so. Optimize removes only pipeline creation blocks. Descriptors (sets, and buffers/images) don't  have their creation block removed by optimize so they are not affected.

fabian-lunarg: I agree that only pipeline-**creation** blocks are removed.

my thinking was `unreferenced_ids` is also used by `FileOptimizer::FilterMetaData` where it drops `kInitBufferCommand` / `kInitImageCommand` for unreferenced resources.
so, in a trimmed capture we still got the creation-blocks for images/buffer but not the init meta-commands.
but that's fine in this case, we won't use it

fabian-lunarg (approved):

PR #3203, opened 2026-08-13
Upgrade Agility SDK from 1.618.5 to 1.619.5

tpower-amd:
## Summary

- Upgrades the bundled Agility SDK from `1.618.5` to `1.619.5`
- Fixes DX12 code generator handling of function pointer typedefs as struct members

### Agility SDK 1.619.5

Updates all binaries, headers, and generated code across the `1.619.x` release series. Notable changes:

**1.619.0 — new features**
- Shader Model 6.9: Shader Execution Reordering (DXR 1.2), Opacity Micromaps (DXR 1.2), long vectors, 16/64-bit shader and wave ops now required, 16-bit float specials, revised resource view creation APIs, periodic trim notifications, increased dispatch grid limit, CPU timeline query
- Advanced Shader Delivery supplemental APIs for developer workflow usability and statistics

**1.619.1**
- Handle root signature objects in `ID3D12StateObjectDatabase` setters
- Fixed Opacity Micromap Array address alignment validation from 256B to correct 128B — debug layer only

**1.619.2**
- `D3D12StateObjectCompiler.exe`: added `--priority` and `--max-threads` parameters
- SDK Layers fixes for enhanced barriers: fixed NO_ACCESS state leaks and updated `LayoutAfter` validation with `TEXTURE_BARRIER_FLAG_DISCARD`
- SODB capture performance improved (defaults to write-ahead logging mode)

**1.619.3**
- Fixed `ID3D12StateObjectDatabase` store with `ID3D12RootSignature1::GetSerializedData`
- `ShaderCacheControl` CLEAR operations now allowed outside developer mode

**1.619.4**
- Tight alignment (`USE_TIGHT_ALIGNMENT`) fixes: lifted `Alignment==0` requirement; extended to textures
- Revised view creation fixes: implicit size (size = 0) for SRVs/UAVs; `CreateUnorderedAccessView` counter offset validation

**1.619.5**
- ETW/telemetry logging when D3D12 Video Encode is unavailable due to exclusive engine access
- Debug layer: `UpdateTileMappings`/`CopyTileMappings` simultaneous-write error reporting fixed; fixed race condition causing subresource states/layouts for subresources > 0 to incorrectly validate against subresource 0's state/layout; fixed Opacity Micromap Array alignment validation
- `D3D12Events.h` now published in the NuGet package and DirectX-Headers repo

**New interfaces in 1.619**

| Interface | Methods |
|---|---|
| `ID3D12Device15` | `RegisterTrimNotificationCallback`, `UnregisterTrimNotificationCallback`, `TryCreateShaderResourceView`, `TryCreateUnorderedAccessView`, `TryCreateConstantBufferView`, `TryCreateSampler2`, `TryCreateRenderTargetView`, `TryCreateDepthStencilView`, `TryCreateSamplerFeedbackUnorderedAccessView`, `CreateQueryHeap1`, `ResolveQueryData` |
| `ID3D12RootSignature1` | `GetSerializedSize`, `GetSerializedData` |
| `ID3D12RuntimeValidationControl` | `DisableFailuresFromStricterValidationInAppLocalRuntime`, `FailuresFromStricterValidationInAppLocalRuntimeDisabled` |
| `ID3D12ApplicationIdentity` | `SetApplicationIdentity` |
| `ID3D12DeviceStatistics` | `GetStateObjectStatistics` |

**New header:** `D3D12Events.h` — ETW event definitions

### DX12 generator fix: function pointer typedefs in structs

`D3D12_REGISTER_TRIM_NOTIFICATION` (new in `1.619.5`) contains a `D3D12_PFN_TRIM_NOTIFICATION_CALLBACK` member. The generator did not recognize `_PFN_`-namespaced typedefs as function pointers when they appeared as struct members, producing missing encoder output and an invalid `DecodeD3D12_PFN_TRIM_NOTIFICATION_CALLBACKValue` call that doesn't exist. Fixed by extending `is_function_ptr()` to match types containing `_PFN_`, falling through to it in `convert_function()` for types not in the hardcoded list, and routing function pointer members through the decoded meta struct (`uint64_t`) in the `to_json` generator.

## Testing

- ✅ All test suites passed: `gfxrecon_util_test`, `gfxrecon_replay_event_plugin_loader_test`, `gfxrecon_format_test`, `gfxrecon_graphics_test`, `gfxrecon_encode_test`, `gfxrecon_framework_test`, and `gfxrecon_decode_test`.


external/AgilitySDK/include/d3dx12/d3dx12.h
```diff
@@ -1,36 +1,34 @@
...
-
-
+//*********************************************************
+//
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License (MIT).
+//
+//*********************************************************
+
+#ifndef __D3DX12_H__
```
davidd-lunarg: The line endings on these Agility SDK source files are changed from Unix `lf` to Windows style `crlf`. Although the SDK ships with `crlf`, please switch these to `lf` to reduce the diff/history impact from line ending change, and to keep it consistent with other externals that are checked in with `lf`. `autocrlf = input` is a git config option that I use and it replaces `crlf` with `lf` on commit, which is what we want for all source files in GFXR.

framework/format/api_call_id.h
```diff
@@ -1436,6 +1436,25 @@ enum ApiCallId : uint32_t
...
+    ApiCall_ID3D12Device15_UnregisterTrimNotificationCallback                                     = MakeApiCallId(ApiFamily_D3D12, 0x117d),
+    ApiCall_ID3D12Device15_TryCreateShaderResourceView                                            = MakeApiCallId(ApiFamily_D3D12, 0x117e),
+    ApiCall_ID3D12Device15_TryCreateUnorderedAccessView                                           = MakeApiCallId(ApiFamily_D3D12, 0x117f),
+    ApiCall_ID3D12Device15_TryCreateConstantBufferView                                            = MakeApiCallId(ApiFamily_D3D12, 0x1180),
+    ApiCall_ID3D12Device15_TryCreateSampler2                                                      = MakeApiCallId(ApiFamily_D3D12, 0x1181),
+    ApiCall_ID3D12Device15_TryCreateRenderTargetView                                              = MakeApiCallId(ApiFamily_D3D12, 0x1182),
+    ApiCall_ID3D12Device15_TryCreateDepthStencilView                                              = MakeApiCallId(ApiFamily_D3D12, 0x1183),
+    ApiCall_ID3D12Device15_TryCreateSamplerFeedbackUnorderedAccessView                            = MakeApiCallId(ApiFamily_D3D12, 0x1184),
```
davidd-lunarg, line 1450: This Agility SDK update adds [D3D12 Revised Create Views](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/D3D12RevisedCreateViews.md)

Descriptors require special handling in GFXR and these would need to include that handling to be properly supported. Some examples include:

- Custom post processing for each `Create...View` call. This is primarily for descriptor state tracking when trimming. For example, for `CreateShaderResourceView`:

https://github.com/LunarG/gfxreconstruct/blob/04afa15886b3ebb115408759e60e6a35385175b7/framework/encode/custom_dx12_wrapper_commands.h#L605-L613

- Replay also has custom post processing for descriptor creation calls:
https://github.com/LunarG/gfxreconstruct/blob/04afa15886b3ebb115408759e60e6a35385175b7/framework/decode/custom_dx12_replay_commands.h#L157-L165

- The original methods don't return an `HRESULT` and so there's no check for success. The new methods can return an unsuccessful result so the handling should account for that. Besides the post processing commands above, there is `D3D12CaptureManager::EndCreateDescriptorMethodCallCapture`.

framework/format/api_call_id.h
```diff
@@ -1436,6 +1436,25 @@ enum ApiCallId : uint32_t
...
+    ApiCall_ID3D12Device15_CreateQueryHeap1                                                       = MakeApiCallId(ApiFamily_D3D12, 0x1185),
+    ApiCall_ID3D12Device15_ResolveQueryData                                                       = MakeApiCallId(ApiFamily_D3D12, 0x1186),
+    ApiCall_ID3D12RuntimeValidationControl_DisableFailuresFromStricterValidationInAppLocalRuntime = MakeApiCallId(ApiFamily_D3D12, 0x1187),
+    ApiCall_ID3D12RuntimeValidationControl_FailuresFromStricterValidationInAppLocalRuntimeDisabled = MakeApiCallId(ApiFamily_D3D12, 0x1188),
+    ApiCall_ID3D12ApplicationIdentity_SetApplicationIdentity                                      = MakeApiCallId(ApiFamily_D3D12, 0x1189),
+    ApiCall_ID3D12DeviceStatistics_GetStateObjectStatistics                                       = MakeApiCallId(ApiFamily_D3D12, 0x118a),
+
```
davidd-lunarg, line 1457: I can't leave a review comment on d3d12.h because that file is too big, so I'm putting it here. It is not releated to the lines of code above.

This PR adds some new enums to `D3D12_FEATURE`: `D3D12_FEATURE_BARRIER_LAYOUT` and `D3D12_FEATURE_D3D12_OPTIONS22`. GFXR has a few places with custom handling for those that need to include the new values:

https://github.com/LunarG/gfxreconstruct/blob/04afa15886b3ebb115408759e60e6a35385175b7/framework/encode/custom_dx12_struct_encoders.cpp#L748
https://github.com/LunarG/gfxreconstruct/blob/04afa15886b3ebb115408759e60e6a35385175b7/framework/decode/dx12_decoder_base.cpp#L370
https://github.com/LunarG/gfxreconstruct/blob/04afa15886b3ebb115408759e60e6a35385175b7/framework/format/platform_types_d3d_overrides.h#L2388
https://github.com/LunarG/gfxreconstruct/blob/04afa15886b3ebb115408759e60e6a35385175b7/framework/decode/dx12_json_consumer_base.cpp#L184

framework/format/api_call_id.h
```diff
@@ -1436,6 +1436,25 @@ enum ApiCallId : uint32_t
     ApiCall_ID3D12StateObjectDatabase_FindObjectVersion                                           = MakeApiCallId(ApiFamily_D3D12, 0x1178),
     ApiCall_ID3D12StateObjectDatabaseFactory_CreateStateObjectDatabaseFromFile                    = MakeApiCallId(ApiFamily_D3D12, 0x1179),
 
+    // Agility SDK 1.619.5
+    ApiCall_ID3D12RootSignature1_GetSerializedSize                                                = MakeApiCallId(ApiFamily_D3D12, 0x117a),
```
davidd-lunarg, line 1440: Adding a separate comment regarding `D3D12 Revised Create Views`:

This adds the ability to specify a byte offset of a descriptor within a heap. They were previously indexed by element within the heap, something like `descriptor_heap_address = sizeof(descriptor_type) * descriptore_index + heap_start`. GFXR does some bookkeeping that IIRC relies on index-based behavior, so we need to take a closer look at whether the byte offset usage is supported.

framework/format/api_call_id.h
```diff
@@ -1436,6 +1436,25 @@ enum ApiCallId : uint32_t
...
+    ApiCall_ID3D12Device15_TryCreateDepthStencilView                                              = MakeApiCallId(ApiFamily_D3D12, 0x1183),
+    ApiCall_ID3D12Device15_TryCreateSamplerFeedbackUnorderedAccessView                            = MakeApiCallId(ApiFamily_D3D12, 0x1184),
+    ApiCall_ID3D12Device15_CreateQueryHeap1                                                       = MakeApiCallId(ApiFamily_D3D12, 0x1185),
+    ApiCall_ID3D12Device15_ResolveQueryData                                                       = MakeApiCallId(ApiFamily_D3D12, 0x1186),
+    ApiCall_ID3D12RuntimeValidationControl_DisableFailuresFromStricterValidationInAppLocalRuntime = MakeApiCallId(ApiFamily_D3D12, 0x1187),
+    ApiCall_ID3D12RuntimeValidationControl_FailuresFromStricterValidationInAppLocalRuntimeDisabled = MakeApiCallId(ApiFamily_D3D12, 0x1188),
+    ApiCall_ID3D12ApplicationIdentity_SetApplicationIdentity                                      = MakeApiCallId(ApiFamily_D3D12, 0x1189),
+    ApiCall_ID3D12DeviceStatistics_GetStateObjectStatistics                                       = MakeApiCallId(ApiFamily_D3D12, 0x118a),
+
     // AGS API
```
davidd-lunarg, line 1458: Adding another comment unrelated to the source line its on:

The change log says:
> lifted Alignment==0 requirement

I wonder if that means we can drop the special handling in `graphics::dx12::RobustGetCopyableFootprint`:

https://github.com/LunarG/gfxreconstruct/blob/04afa15886b3ebb115408759e60e6a35385175b7/framework/graphics/dx12_util.cpp#L1593-L1597

davidd-lunarg (changes requested):
See comments below

framework/format/api_call_id.h
davidd-lunarg: On second thought, even if it is true that `GetCopyableFootprints` now works when `Alignment != 0` for tightly aligned resources, it would require that all applications are using this Agility SDK to avoid capture-time issues with `GetCopyableFootprints` so I don't think `RobustGetCopyableFootprint` can be changed.

framework/format/api_call_id.h
davidd-lunarg: I misread the spec. It doesn't add the ability to specify byte offset of a descriptor. Instead, it adds new SRV/UAV view dimensions that specify byte offsets into the descriptor's buffer. So GFXR's descriptor tracking is OK, but there are several places where the new descriptor dimensions should be added to GFXR handling. A search in the GFXR code for `D3D12_UAV_DIMENSION*` and `D3D12_SRV_DIMENSION*` finds several places where descriptor dimensions are processed and may need added support for `D3D12_UAV_DIMENSION_BUFFER_BYTE_OFFSET` and `D3D12_SRV_DIMENSION_BUFFER_BYTE_OFFSET`.

tpower-amd: Thanks for the clarification. Did a deeper investigation and arrived at the same conclusion myself. The generated `EncodeStruct`/`DecodeStruct` overloads for the new structs already exist, they just need wiring into those switches.

framework/format/api_call_id.h
tpower-amd: Agreed. Even if `1.619.4` lifts the alignment restriction, the workaround needs to stay in place to support backwards compatibility.

framework/format/api_call_id.h
tpower-amd: @bradgrantham-lunarg @davidd-lunarg

A few questions on approach for the `TryCreate*` implementation.

Replay-side `PostCall_` hooks: currently hand-writing the implementations that delegate to the corresponding `PostCall_ID3D12Device_Create*` on `SUCCEEDED`. Another option would be extracting the body of each into a shared helper so both the original and `TryCreate*` variant call it, rather than one calling the other. Also, we could potentially have the generator could produce this boilerplate automatically for new descriptor-creating APIs. Is there a preference on either of those approaches, or is hand-writing the delegation acceptable for now?

`EndCreateDescriptorMethodCallCapture`: I noticed the generated `Encode_ID3D12Device15_TryCreate*` functions call this unconditionally regardless of the `HRESULT`, which would incorrectly record a failed descriptor creation in the trim snapshot. Our suggestion is fixing this through the generator rather than custom encoder overrides. Does that seem like the right approach to y'all?

davidd-lunarg: Since the descriptor APIs don't seem to change too often I think hand-writing the descriptor delegation is acceptable. And as long as each pair of `Create` and `TryCreate` functions share the same code & logic, I don't have a preference on whether one function calls the other or they both call a helper.

`EndCreateDescriptorMethodCallCapture` needs to be called even if the call failed because it also calls `EndMethodCallCapture();` which, among other things, is how the API call gets written to the capture file, and GFXR's goal is to capture all API calls made by the application even if they weren't successful. Could `EndCreateDescriptorMethodCallCapture` add an argument `HRESULT result` that the `Create` methods always pass `S_OK` to? EDIT: So to answer your question, yes I think this can be mostly a fix in the generator with a tweak to the function signature.

framework/decode/dx12_dump_resources.cpp
```diff
@@ -828,6 +828,9 @@ void Dx12DumpResources::GetDescriptorSubresourceIndices(DHShaderResourceViewInfo
             }
             break;
         }
+        case D3D12_SRV_DIMENSION_BUFFER_BYTE_OFFSET:
+            // Byte-offset buffer SRC - no subresource indices applicable.
```
davidd-lunarg, line 832: In order to dump the data pointed to by this SRV, this should still call `info.subresource_indices.emplace_back(0);`--same as the non-offset `D3D12_SRV_DIMENSION_BUFFER`. Additional bookkeeping would be needed to accurately process the specified bytes.

`case D3D12_UAV_DIMENSION_BUFFER_BYTE_OFFSET:` at line 904 below has a similar issue.

framework/decode/custom_dx12_struct_decoders.h
```diff
@@ -89,20 +89,22 @@ struct Decoded_D3D12_SHADER_RESOURCE_VIEW_DESC
...
-    Decoded_D3D12_TEX3D_UAV*          Texture3D{ nullptr };
+    D3D12_UNORDERED_ACCESS_VIEW_DESC*     decoded_value{ nullptr };
+    Decoded_D3D12_BUFFER_UAV*             Buffer{ nullptr };
+    Decoded_D3D12_TEX1D_UAV*              Texture1D{ nullptr };
+    Decoded_D3D12_TEX1D_ARRAY_UAV*        Texture1DArray{ nullptr };
+    Decoded_D3D12_TEX2D_UAV*              Texture2D{ nullptr };
+    Decoded_D3D12_TEX2D_ARRAY_UAV*        Texture2DArray{ nullptr };
+    Decoded_D3D12_TEX2DMS_UAV*            Texture2DMS{ nullptr };
+    Decoded_D3D12_TEX2DMS_ARRAY_UAV*      Texture2DMSArray{ nullptr };
+    Decoded_D3D12_BUFFER_UAV_BYTE_OFFSET* BufferByteOffset{ nullptr };
```
davidd-lunarg, line 106: For these new UAV/SRV dimensions, there are some additional places that need them.

1. The switches in the json generator python file for the new SRV/UAV dimensions:
https://github.com/LunarG/gfxreconstruct/blob/bec62a501f6061cb70cc1340258ecd732d0e7910/framework/generated/dx12_generators/dx12_struct_decoders_to_json_body_generator.py#L239
https://github.com/LunarG/gfxreconstruct/blob/bec62a501f6061cb70cc1340258ecd732d0e7910/framework/generated/dx12_generators/dx12_struct_decoders_to_json_body_generator.py#L311

2. `platform_types_d3d_overrides.h`:
https://github.com/LunarG/gfxreconstruct/blob/bec62a501f6061cb70cc1340258ecd732d0e7910/framework/format/platform_types_d3d_overrides.h#L4275
https://github.com/LunarG/gfxreconstruct/blob/bec62a501f6061cb70cc1340258ecd732d0e7910/framework/format/platform_types_d3d_overrides.h#L4427

framework/generated/generated_dx12_api_call_encoders.cpp
```diff
@@ -6608,6 +6673,269 @@ void Encode_ID3D12Device14_CreateRootSignatureFromSubobjectInLibrary(
...
+        bool omit_output_data = false;
+        if (return_value != S_OK)
+        {
+            omit_output_data = true;
+        }
+        encoder->EncodeObjectValue(pQueryHeap);
+        encoder->EncodeEnumValue(Type);
+        encoder->EncodeUInt32Value(StartIndex);
+        encoder->EncodeUInt32Value(NumQueries);
+        encoder->EncodeVoidPtr(pResolvedQueryData);
```
davidd-lunarg, line 6933: `encoder->EncodeVoidPtr(pResolvedQueryData);` encodes a pointer value but doesn't save the results of the queries to the capture file. 

And on the replay side `Dx12ReplayConsumer::Process_ID3D12Device15_ResolveQueryData` calls `ResolveQueryData` with a `nullptr` for `in_pResolvedQueryData`.

The in/out `void* pResolvedQueryData` argument is kind of unique for the API so the generators don't handle it automatically. A few steps would be required to fully capture and replay:

Implement a helper `graphics::dx12::GetResolveQueryDataSize(...)` in `dx12_util.h/cpp` and then add an entry to `ARRAY_SIZE_LIST` in `dx12_base_generator.py`, something like:
```
        [
            'ID3D12Device15_ResolveQueryData',
            'pResolvedQueryData',
            'graphics::dx12::GetResolveQueryDataSize(...)'
        ],
```

And finally, modify the relevant generator python so that `Dx12ReplayConsumer::Process_ID3D12Device15_ResolveQueryData` calls `AllocateOutputData` to actually allocate the data behind the argument for `pResolvedQueryData`.

davidd-lunarg (changes requested):

framework/graphics/dx12_util.cpp
```diff
@@ -1719,6 +1719,38 @@ uint64_t GetSubresourceSizeTex3D(uint32_t depth, uint32_t mip_levels, uint32_t d
...
 
+size_t GetResolveQueryDataSize(D3D12_QUERY_TYPE type, UINT num_queries)
+{
+    size_t element_size = 0;
+    switch (type)
+    {
+        case D3D12_QUERY_TYPE_OCCLUSION:
+        case D3D12_QUERY_TYPE_BINARY_OCCLUSION:
+        case D3D12_QUERY_TYPE_TIMESTAMP:
+        case D3D12_QUERY_TYPE_VIDEO_DECODE_STATISTICS:
```
davidd-lunarg, line 1730: `D3D12_QUERY_TYPE_VIDEO_DECODE_STATISTICS` queries return a `D3D12_QUERY_DATA_VIDEO_DECODE_STATISTICS` so the size should be based on the size of that struct to ensure enough memory is allocated. However, the [CPU resolve spec](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/D3D12CpuTimelineQueryResolution.md#id3d12deviceresolvequerydata) doesn't mention `D3D12_QUERY_TYPE_VIDEO_DECODE_STATISTICS` at all, so maybe this isn't supported for CPU resolve.

But in any case, GFXR doesn't currently support the video decode API, so I think this case should log a warning and set `element_size = sizeof(D3D12_QUERY_DATA_VIDEO_DECODE_STATISTICS);`.

framework/decode/dx12_dump_resources.cpp
davidd-lunarg, line 1237: When `ViewDimension` is `D3D12_UAV_DIMENSION_BUFFER_BYTE_OFFSET`, the `desc.Buffer` union member is not valid. Instead `desc.BufferByteOffset.CounterOffsetInBytes` should be read.

davidd-lunarg, line 1159: This should have a case for `D3D12_SRV_DIMENSION_BUFFER_BYTE_OFFSET` that sets
```
offset = desc.BufferByteOffset.Offset;
size   = desc.BufferByteOffset.Size;
```

Same goes for the switch under `D3D12_DESCRIPTOR_RANGE_TYPE_UAV` below.

davidd-lunarg (changes requested):

PR #3205, opened 2026-08-13
Forward if PresentImageAdHoc early returns

wzchu-lunarg:
In `OverrideFrameBoundaryANDROID`, it seems that when doing `PresentImageAdHoc` while we have `disable_adhoc_presentation_` set to true, it early returns and never gets forwarded to the driver, resulting in missing frames. This caused the CI's Pixel 8 traces to fail, as screenshots during recapture looked reasonable, yet was the wrong frame when comparing to the references in the same frame index. 

Change `PresentImageAdHoc` to return a bool of whether it actually succeeded. If not, we forward the captured call.


wzchu-lunarg:
Should fix https://github.com/LunarG/Projects/issues/1212. Extended run: http://jenkins:8080/job/gfxr-pipelines/job/gfxr-manual/695/

wzchu-lunarg:
Suites PR to reenable Pixel 8 traces: https://github.com/LunarG/ci-gfxr-suites/pull/179

bradgrantham-lunarg (approved):

PR #3207, opened 2026-08-17
Implement multi handle allocation in frame looping

davidlunarg:
Support for:
    vkCreateShadersEXT
    vkCreateGraphicsPipelines
    vkCreateComputePipelines
    vkCreateRayTracingPipelinesNV
    vkCreateDataGraphPipelinesARM
    vkCreateSharedSwapchainsKHR

### DIFF
--- a/framework/generated/generated_vulkan_replay_frame_loop_consumer_base.cpp
+++ b/framework/generated/generated_vulkan_replay_frame_loop_consumer_base.cpp
@@ -930,14 +930,9 @@ void VulkanReplayFrameLoopConsumerBase::Process_vkCreateComputePipelines(
     VkComputePipelineCreateInfo* raw_infos  = args.pCreateInfos.GetPointer();
     Decoded_VkComputePipelineCreateInfo* meta_infos = args.pCreateInfos.GetMetaStructPointer();
 
-    // args.pCreateInfos/args.pPipelines retain their original (full-batch) decoded length
-    // regardless of args.createInfoCount, so Process_vkCreateComputePipelines() will still run
-    // MapStructArrayHandles() over the *entire* args.pCreateInfos array once per to_create entry
-    // below (not just the single slot-0 entry being replayed that iteration). That
-    // repeated work does no harm since MapStructHandles() writes mapped handles into
-    // each struct's live fields while leaving the wrapper's original capture-id
-    // fields untouched.
     const uint32_t original_count_value = args.createInfoCount;
+    const size_t original_pCreateInfos_length = args.pCreateInfos.GetLength();
+    const size_t original_pPipelines_length = args.pPipelines.GetLength();
 
     for (uint32_t i : to_create)
     {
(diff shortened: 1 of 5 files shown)


davidlunarg:
I'm still in the process of creating a test program for these changes. But I'm putting the current implementation out here for initial feedback.

framework/generated/generated_vulkan_replay_frame_loop_consumer_base.cpp
```diff
@@ -883,36 +890,112 @@ void VulkanReplayFrameLoopConsumerBase::Process_vkCreateComputePipelines(
...
+    VkComputePipelineCreateInfo* raw_infos  = args.pCreateInfos.GetPointer();
+    Decoded_VkComputePipelineCreateInfo* meta_infos = args.pCreateInfos.GetMetaStructPointer();
+
+    // Process_vkCreateComputePipelines() reads the capture-id array via args.pPipelines, not through a
+    // separate parameter, so that array has to be kept in sync with the
+    // args.pCreateInfos/args.pPipelines swap below in order for its internal handle registration
+    // (AddHandles) to associate the new VkPipeline with the correct capture id.
+    // args.pPipelines.GetPointer() only exposes a const array so the const_cast
+    // below is required to reorder it in place.
+    format::HandleId* mutable_capture_ids = const_cast<format::HandleId*>(capture_ids);
```
antonio-lunarg: I am not a fan of const casting. How about introducing a `GetPointerMut()`?

jzulauf-lunarg: I don't think there are any existence proofs for const casting.  We tend to copy and modify to a local store.

davidlunarg: GetPointerMutable() implemented and used in this PR.

framework/generated/khronos_generators/khronos_replay_frame_loop_consumer_base_body_generator.py
```diff
@@ -63,36 +63,119 @@ def make_replay_frame_loop_consumer_func_body(self, api_data, return_type, name,
...
+            body += '        // Move index i into slot 0 so Process_{}/driver code (which\n'.format(name)
+            body += '        // always starts at index 0) operates on the {} we actually want.\n'.format(handle_base_type)
+            body += '        std::swap(raw_infos[0], raw_infos[i]);\n'
+            body += '        std::swap(meta_infos[0], meta_infos[i]);\n'
+            body += '        meta_infos[0].decoded_value = &raw_infos[0];\n'
+            body += '        meta_infos[i].decoded_value = &raw_infos[i];\n'
+            body += '        std::swap(capture_ids[0], capture_ids[i]);\n'
+            body += '\n'
+            body += '        // Restrict this call to a single create info/handle: Process_{}()\n'.format(name)
+            body += '        // uses {} (not GetLength()) to size the output handle array and to\n'.format(count_name)
```
antonio-lunarg: Some functions do use `GetLength()` and the result would be a mismatch between the number of create infos and handles. (SkipPipelineCreationForCompileRequired, AddHandlesAsync, populate_shader_stages)
I think it would be better to narrow the decoded lengths too.

davidlunarg: Not sure what you mean by "narrow the decoded lengths". 

PR 26e241a6564f717a347abd667bed9b729ee20b4b changes generator to allow GetLength() to be used on create infos.

antonio-lunarg: "narrow the decoded lengths" -> `SetLength(1)`, especially for handle arrays, which your commit addresses correctly. Thanks!

antonio-lunarg (approved):

PR #3208, opened 2026-08-18
Improve/Fix toCpp

MarkY-LunarG:
This fixes multiple issues with the toCpp path including:

 * Invalid compilation including max window dimensions
 * Correct image/memory/push descriptor calls
 * Fix crashes
 * Treat VK_SUBOPTIMAL_KHR as a valid return code from the swapchain
 * Fix hangs when generating code in several situations
 * Improve pNext chain running out of stack space (still a problem, but a big step forward)
 * Use std::vectors in place of hard-coded arrays
 * Write strings correctly for several API calls
 * Split up the consumer base code as a step towards maintainence as complexity increases
 * Actually choose correct device and memory based on capture info instead of hard-coding offsets (just like replay)
 * Add an option to force a GPU
 * Rewrite handling of how swapchain surfaces are selected
 * Clean up multiple windows when done with them
 * Remove Android AHBs when replaying on desktop
 * Fix Android build


MarkY-LunarG:
NOTE: This is only my first step towards cleaning up and improving toCpp.  This increases the number of test applications I have run that now work with toCpp on both Desktop and Android.

MarkY-LunarG:
NOTE 2: An additional goal of mine is to introduce screenshot functionality that can be used to compare a trace running through replay (at a given frame) with that of the generated toCpp version at the same given frame.  But this is in the future, but would be the best way to test the resulting cpp code since that can go in so many different ways.  Of course, it should also compile as well.  So step 1 might be to add compilation of toCpp content on a trace, and step 2 might be doing the screensthos.

bradgrantham-lunarg:
> NOTE 2: An additional goal of mine is to introduce screenshot functionality that can be used to compare a trace running through replay (at a given frame) with that of the generated toCpp version at the same given frame. But this is in the future, but would be the best way to test the resulting cpp code since that can go in so many different ways. Of course, it should also compile as well. So step 1 might be to add compilation of toCpp content on a trace, and step 2 might be doing the screensthos.

I encourage you to go more in the direction of saving off and comparing the Vulkan commands since that's where we've tilted with test/test_cases and also work @beau-lunarg has been doing for internal CI.  Screenshot comparisons are problematic because driver versions and GPU families produce slightly different pixel output.

CreativeCylon:
I had some fixes in https://github.com/LunarG/gfxreconstruct/pull/3145 maybe those can be pulled in here

MarkY-LunarG:
> I had some fixes in https://github.com/LunarG/gfxreconstruct/pull/3145 maybe those can be pulled in here

Pulled over fixes from #3145, although some introduced issues and I had to revert it.  A large desktop capture I use as a baseline now works better and cleaner than before.

PR #3210, opened 2026-08-18
[Frame looping] Image layout restoration for looped frames

wzchu-lunarg:
During frame looping, we want to restore image layouts as otherwise we might get validation errors on the next looped frame.

Add and modify changes from https://github.com/LunarG/gfxreconstruct/pull/3047/commits/3e23b0bd43ec5619e6073227ec20a801b98b5fa9 and https://github.com/LunarG/gfxreconstruct/pull/3047/commits/1fea71d7cfad9e012605390ddf2e4c02d43438a9 to support image layout restoration when frame looping.

The changes tries to address the layout mismatch issue by tracking the layouts before frame looping, and then doing a "Fixup" to the layouts at the end of the looped frame. 

At the end of the frame, the changes are then applied by updating the image barriers in the command buffer (applying), updating each image's layout (propagating), and then a fixup transitioning each layout to the initial one with a new command pool.

Multiple refactors from Oleh's original code to match current code in `vulkan_replay_frame_loop_consumer`: 

1. Rename `ExtractSubpass0Layouts` -> `TrackSubpass0Layouts` 

2. Rename `PropagateRenderPassInitialLayouts` -> `ApplySubpass0Layouts` 
3. Rename `RecordInitialLayouts` -> `TrackImageLayouts`
4. Add `Process_vkDestroyRenderPass`
5. Add `Process_vkCmdBeginRenderPass2KHR`
6. Add `Process_vkCmdBeginRenderingKHR`
7. Move layout update logic in `Process_vkCmdBeginRendering` to helper `ApplyRenderingLayouts` 
8. `RestoreImageLayouts` -> `FixupImageLayouts`, change const_cast to a `GetVkImageInfo` call. Change `barrier.srcAccessMask` and `barrier.dstAccessMask` to use `VK_ACCESS_MEMORY_READ_BIT || VK_ACCESS_MEMORY_WRITE_BIT`
9. `InitializeRestorationResources` -> `SubmitImageLayoutBarriers`, moved command buffer creation and submission to be together here, rather than having submit logic in `FixupImageLayouts`, use `CHECK_VK_RESULT` for these calls 
10. `PropagateImageLayouts` instead takes a command buffer to update image layouts, and is called during queue submission 

**Limitation**
The current limitation that Oleh [mentions](https://github.com/olehkuznetsov/gfxreconstruct/blob/bf2a84926f0316160d688ecdbd5957a9aa0d110f/frame_loop_layouts.md) is that GFXR currently only stores a single `VkImageLayout` for each image. However, depth/stencil images and mipmap often have multiple layouts for each subresource. This affects frame looping, as restoring the entire layout of these multiple subresource images can result in validation errors. 

As the fix will involve a decently sized refactor, these changes will be pulled in on a later PR.


wzchu-lunarg:
Removing tracking for subpass0 layouts. It seems that originally, it would try to patch barriers to use the image's current layout for `oldLayout` This makes it necessary to track these intermediate transitions, and the only way to track the implicit transitions such as in `vkCmdBeginRenderPass` is to use subpass0's layout.

However, it seems this is no longer necessary, as one of the original commit's changes https://github.com/LunarG/gfxreconstruct/pull/3047/changes/1fea71d7cfad9e012605390ddf2e4c02d43438a9 removed processing `vkCmdPipelineBarrier` and related commands, as we don't patch barriers, and instead restore the final barrier to be the initial one.

A concern I might need more insight on is if we need to keep track of these implicit transitions still, as perhaps this `vkCmdBeginRenderPass`'s implicit transition could be the last transition, such as if `vkEndCommandBuffer` and submit happens in the next frame. 
However, it seems to me that if submit is not done in the same frame, then the image doesn't actually transition so we don't need to track it. If submit does happen, then that means the implicit transition to the final layout eventually happens which we track in `VulkanReplayConsumerBase::ApplyRenderPassFinalLayouts`.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -62,6 +62,73 @@ void VulkanReplayFrameLoopConsumer::StartLooping()
     GFXRECON_LOG_DEBUG("VulkanReplayFrameLoopConsumer::StartLooping()");
     TrackFenceStates();
     TrackEventStates();
+    TrackImageLayouts();
+}
+
+void VulkanReplayFrameLoopConsumer::ApplyRenderingLayouts(format::HandleId                               command_buffer,
```
antonio-lunarg: I have the feeling that the name of this function is a bit misleading.
At first I imagine that it is going to transition images to some layout for rendering, but reading the code I realize it is actually updating the tracked image layout in a command buffer info structure. So, how about something like "UpdateTrackedImageLayouts"?

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -493,13 +560,158 @@ void VulkanReplayFrameLoopConsumer::FrameBoundaryEndOfFrame(format::HandleId que
...
+}
+
+void VulkanReplayFrameLoopConsumer::SubmitImageLayoutBarriers(const VulkanDeviceInfo*                  device_info,
+                                                              const VulkanQueueInfo*                   queue_info,
+                                                              const std::vector<VkImageMemoryBarrier>& barriers)
+{
+    const graphics::VulkanDeviceTable* device_table = GetDeviceTable(device_info->handle);
+    GFXRECON_ASSERT(device_table != nullptr);
+
+    VkCommandPoolCreateInfo pool_create_info = { VK_STRUCTURE_TYPE_COMMAND_POOL_CREATE_INFO };
```
antonio-lunarg: How about using `TemporaryCommandBuffer` from `decode/vulkan_temporary_objects.h`?
Maybe improving its `CreateAndBegin` to accept an optional queue parameter rather than just defaulting to queue 0?

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -520,9 +732,14 @@ void VulkanReplayFrameLoopConsumer::Process_vkQueueSubmit(const ApiCallInfo& cal
 {
     VulkanReplayConsumer::Process_vkQueueSubmit(call_info, args);
 
-    if (frame_loop_info_.IsLooping())
+    for (const Decoded_VkSubmitInfo& submit : args.pSubmits.GetMetaStructSpan())
     {
-        for (Decoded_VkSubmitInfo submit : args.pSubmits.GetMetaStructSpan())
+        for (format::HandleId command_buffer : submit.pCommandBuffers.GetSpan())
+        {
+            PropagateImageLayouts(command_buffer);
```
antonio-lunarg: I do not think we need to call this outside of the frame loop range.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -533,9 +750,17 @@ void VulkanReplayFrameLoopConsumer::Process_vkQueueSubmit2(const ApiCallInfo& ca
 {
     VulkanReplayConsumer::Process_vkQueueSubmit2(call_info, args);
```
antonio-lunarg, line 1125: I wonder if we need to do the same thing for `Process_vkQueueSubmit2KHR`..

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: I agree, the original intention was that I wanted a different name than "Track", and in a way it does apply the layout but to the command buffer infos. Furthermore, there were functions in `replay_consumer_base` that uses the word `Apply` similarly, such as `ApplyRenderPassFinalLayouts`.

However, I think `UpdateTrackedImageLayout` would be a better name and won't be as misleading. Adding a commit for this.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Wow had no idea this existed, and looks like it will be a lot cleaner to use it. Will update `CreateAndBegin` to take in `queue_index`.

Adding a commit for this.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: The intention was to keep tracking the layouts of the live images even outside of frame looping. This gives a previous layout that we can use when doing an image transition, as we will have the correct `old_layout`.

However, this brings up a good point, and that maybe this image propagation and tracking of an image's current layout should be in `replay_consumer_base` instead.

wzchu-lunarg: Going to look into this a little more. It seems to me that we aren't tracking image layouts correctly even in the consumer base. This incorrect tracking seemed to never affect replay as we simply replay tracked barriers. However, when we inject commands that needed an image's current layout, there might be issues. Might make a separate PR for this.

wzchu-lunarg: https://github.com/LunarG/gfxreconstruct/pull/3217 will address this and does all image tracking and propagation in consumer base. Going to cherry pick that commit over for now as well.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: With the changes in https://github.com/LunarG/gfxreconstruct/pull/3217, the image propagation won't be in the frame loop consumer's queue submits. However, it does seem that we are missing the frame boundary handling for `Process_vkQueueSubmit2KHR`.

wzchu-lunarg: > With the changes in #3217, the image propagation won't be in the frame loop consumer's queue submits. However, it does seem that we are missing the frame boundary handling for `Process_vkQueueSubmit2KHR`.

Addressing this in a separate PR https://github.com/LunarG/gfxreconstruct/pull/3219.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -875,13 +877,183 @@ void VulkanReplayFrameLoopConsumer::FrameBoundaryEndOfFrame(format::HandleId que
...
+        if ((initial != current) && IsRestorableLayout(initial))
+        {
+            barriers.push_back(MakeLayoutRestoreBarrier(
+                image_info->handle, current, initial, { aspects, 0, mip_levels, 0, array_layers }));
+        }
+        return;
+    }
+
+    // Otherwise restore each subresource individually.
+    for (VkImageAspectFlagBits aspect : graphics::kLayoutMapAspects)
```
antonio-lunarg: There might be a problem with combined depth/stencil images. I think at the current state, the code is going to generate two separate barriers, a `DEPTH` only and a `STENCIL` only. We should detect such cases and emit one barrier with the OR of those aspects.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -875,13 +877,183 @@ void VulkanReplayFrameLoopConsumer::FrameBoundaryEndOfFrame(format::HandleId que
...
+void VulkanReplayFrameLoopConsumer::SubmitImageLayoutBarriers(const VulkanDeviceInfo*                  device_info,
+                                                              const VulkanQueueInfo*                   queue_info,
+                                                              const std::vector<VkImageMemoryBarrier>& barriers)
+{
+    const graphics::VulkanDeviceTable* device_table = GetDeviceTable(device_info->handle);
+    GFXRECON_ASSERT(device_table != nullptr);
+
+    TemporaryCommandBuffer temp_command_buffer(*device_info, *device_table);
+
+    VkResult result = temp_command_buffer.CreateAndBegin(queue_info->family_index, queue_info->queue_index);
```
antonio-lunarg, line 1012: Since we already have a `queue_info->handle`, can we overload `CreateBegin` and avoid passing `queue_index` here?
Refinement for a follow-up PR.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Interesting, didn't notice this case. 
> VUID-VkImageMemoryBarrier-image-03320
If image has a depth/stencil format with both depth and stencil and the [separateDepthStencilLayouts](https://docs.vulkan.org/spec/latest/chapters/features.html#features-separateDepthStencilLayouts) feature is not enabled, then the aspectMask member of subresourceRange must include both [VK_IMAGE_ASPECT_DEPTH_BIT](https://docs.vulkan.org/refpages/latest/refpages/source/VkImageAspectFlagBits.html) and [VK_IMAGE_ASPECT_STENCIL_BIT](https://docs.vulkan.org/refpages/latest/refpages/source/VkImageAspectFlagBits.html)

Going to look into a way to restructure things so we can group aspects if they share layouts.

wzchu-lunarg: Adding a commit that groups subresource aspects if they share the same transition, so a single barrier will do multiple aspects. 

Could potentially look into coalescing mip levels and layers if they are contiguous as well, but I don't see any cases that would benefit from this.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Sounds good for a follow-up.

antonio-lunarg (approved):

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: https://github.com/LunarG/gfxreconstruct/pull/3254

PR #3212, opened 2026-08-19
VDR: Fix crash when dumping secondaries

panos-lunarg:
This PR fixes a crash that happened when the same secondary is executed multiple times in the same CmdExecuteCommands and is requested for dumping.

The described case is not fully supported for dumping resources as there is no way to describe which execution of the repeated secondary to be dumped. However the first ocurence should be possible to be dumped.

Additionally taking the opportunity to simplify some of the render pass logic and fix some other bugs in the logic that handles secondaries


framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
```diff
@@ -1168,18 +1165,18 @@ VkResult DrawCallsDumpingContext::DumpDrawCalls(VkQueue              queue,
...
+        //     // This map is updated in UpdateSecondaries accordingly
+        //     const auto entry = dc_params.secondary_identifiers.find(cb_absolute);
+        //     GFXRECON_ASSERT(entry != dc_params.secondary_identifiers.end());
+        //     if (entry != dc_params.secondary_identifiers.end())
+        //     {
+        //         secondary_indices = entry->second;
+        //         GFXRECON_ASSERT(secondary_indices.execute_cmds_index != UNDEFINED_INDEX);
+        //         GFXRECON_ASSERT(secondary_indices.execute_cmds_cmd_buf_index != UNDEFINED_INDEX);
+        //     }
+        // }
```
fabian-lunarg: that's the only blocker I found, what is the intention here?

now, `SecondaryIdentifiers secondary_indices;` remains as default, which looks it changes hashes/filenames down the line. 
but also: one draw call dumped from two different `vkCmdExecuteCommands` now hashes identically.

it looks this might be trying to avoid a problematic assert, but what did the asserts actually catch?

panos-lunarg: Doh, I pushed the commented out lines.
Yes I discovered an assert failing while testing.
It seemed that it was out of scope for this PR but I'll take another look and try to fix it

framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
fabian-lunarg, line 1167: wrt. to this block. it looks good now, but following the breadcrumbs I think
some resource are not cleared, for repeatedly-executed draws. 
( second dump of a repeatedly-executed draw call re-emits the first dump's entries)

please have a look at :1155, we could/should clear there like: 

```
if (is_before_command)
{
      // Each dump of this draw call starts from an empty resource record
      dc_params.dumped_resources.dumped_render_targets.clear();
      dc_params.dumped_resources.dumped_vertex_index_buffers.clear();
      dc_params.dumped_resources.dumped_descriptors.clear();
...
```

atm this is visible in the newly added testsuite-case:
imageId 23 appears twice at "location": 0 on the second draw-246 record

framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
fabian-lunarg: looks good now

framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
panos-lunarg: Good catch. I believe it should be addressed now.

framework/decode/vulkan_replay_dump_resources_compute_ray_tracing.cpp
fabian-lunarg, line 1352: is this missing

`dispatch_dumped_descriptors_.acceleration_structures.clear();`
...
`trace_rays_dumped_descriptors_.acceleration_structures.clear();` ?

panos-lunarg: Done

fabian-lunarg (approved):

PR #3213, opened 2026-08-20
Add `BlockProcessor::ProcessBlock`

marius-pelegrin-arm:
Add function in block processor to process a single block, instead of processing an entire frame at once.

### DIFF
--- a/framework/decode/block_processor.h
+++ b/framework/decode/block_processor.h
@@ -123,6 +123,9 @@ class BlockProcessor
     // -------------------------------------------------------------------------
 
     template <typename Policy>
+    file_processor::ProcessBlockState ProcessBlock(Policy& policy, BlockBuffer& block_buffer);
+
+    template <typename Policy>
     file_processor::ProcessBlockState ProcessBlocks(Policy& policy);
 
     // -------------------------------------------------------------------------
--- a/framework/decode/block_processor.cpp
+++ b/framework/decode/block_processor.cpp
@@ -116,67 +116,70 @@ file_processor::ProcessBlocksResult BlockProcessor::MakeResult(file_processor::P
 }
 
 template <typename Policy>
+file_processor::ProcessBlockState BlockProcessor::ProcessBlock(Policy& policy, BlockBuffer& block_buffer)
+{
+    PrintBlockInfo();
+
+    if (!policy.ContinueBlockProcessing(block_index_))
+    {
+        return file_processor::ProcessBlockState::kEndProcessing;
+    }
+
+    BlockParser& block_parser = *block_parser_.get();
+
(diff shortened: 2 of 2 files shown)


jzulauf-lunarg:
@marius-pelegrin-arm -- did you run any before after performance testing.  Changes to block processing can be touchy.

marius-pelegrin-arm:
No, I did not. I just assumed re-using the block buffer was very important peformance-wise (that's why it's an argument) and that the process visitor - being only a few bytes - could be re-created for each block.
I wanted to have the process visitor as an argument as well, but the header inclusion scheme is pretty messy and I didn't want to touch all that just for sparing a few bytes of stack allocation per block.

Now, feel free to change the commit if you see performance degradations =)

framework/decode/block_processor.h
```diff
@@ -122,6 +122,9 @@ class BlockProcessor
     // Main processing loop (moved from FileProcessor)
     // -------------------------------------------------------------------------
 
+    template <typename Policy>
+    file_processor::ProcessBlockState ProcessBlock(Policy& policy, BlockBuffer& block_buffer);
```
fabian-lunarg, line 126: the function is public + templated, without explicit instantiations in the header.
it looks this PR is preparing the ground to expose this API and use in other TUs in the future.
this might be causing linker issues, depending on compiler/settings.

fabian-lunarg (approved):
I have no real concerns here, my comment wrt. linker-issues 'might become' relevant. 
the PR-description could better motivate/explain why the change is done. it can be understood by cross-referencing with [ARM's fork](https://github.com/ARM-software/gfxreconstruct-arm) (preparing ground to examine individual blocks) but the motivation is not expressed in the PR or commit-msg.

framework/decode/block_processor.h
marius-pelegrin-arm: It is the exact same context as the ProcessBlocks function. I am not sure to understand the concern, but the function needs to be public because it is accessed from the file processor in our fork

marius-pelegrin-arm:
@fabian-lunarg I can add an explanation in the commit message with reference to our fork

fabian-lunarg:
> @fabian-lunarg I can add an explanation in the commit message with reference to our fork

just a brief explanation why the change was necessary (what the problem was) would be great,
in case someone comes back in the future to understand the PR.

marius-pelegrin-arm:
@fabian-lunarg I think you missed it, I pushed the change to the commit message this morning =)
Unless you need something more descriptive ?

fabian-lunarg:
> @fabian-lunarg I think you missed it, I pushed the change to the commit message this morning =) Unless you need something more descriptive ?

yes, noticed it after writing the text above, I'm happy :)

PR #3214, opened 2026-08-20
Fix error message for block header read failure

bradgrantham-lunarg:


framework/decode/block_parser.cpp
```diff
@@ -495,7 +495,7 @@ ParsedBlock& BlockParser::ParseMetaData(BlockBuffer& block_buffer)
 
     if (!success)
     {
-        HandleBlockReadError(kErrorReadingBlockHeader, "Failed to read function call block header");
+        HandleBlockReadError(kErrorReadingBlockHeader, "Failed to read block header");
```
jzulauf-lunarg: "Failed to read metadata block header"

jzulauf-lunarg (changes requested):
Good catch, but be specific of the block header type.

PR #3215, opened 2026-08-21
[Frame Looping] Suppress queries 

wzchu-lunarg:
Add Oleh's changes to suppress queries when frame looping as [described](https://github.com/olehkuznetsov/gfxreconstruct/blob/d64f4ab205604d2f55ed964a7521b7e62fb05425/frame_loop_queries.md) in https://github.com/LunarG/gfxreconstruct/pull/3047/changes/e069c04677ff38f4e67d2a809c25a76703ad0cec.

In Vulkan, queries must be reset before it can be written to or activated. However, query resets and writes can be in different frames. When looping, this can result in writes without resets and validation errors.

The changes opted for here is simply to suppress these query commands while frame looping by adding a frame looping check in the generated `Process_*` commands.

Perhaps it is possible to somehow keep functionality during frame looping by resetting specific queries before writing, but it is mentioned that this is a simpler solution than "custom complex manual reset trackers for every query pool index." Furthermore, accessing query results such as `vkGetQueryPoolResults` and `vkCmdCopyQueryPoolResults` might be take more effort to support.

### DIFF
--- a/framework/generated/generated_vulkan_replay_frame_loop_consumer_base.cpp
+++ b/framework/generated/generated_vulkan_replay_frame_loop_consumer_base.cpp
@@ -412,6 +412,17 @@ void VulkanReplayFrameLoopConsumerBase::Process_vkDestroyQueryPool(
     }
 }
 
+void VulkanReplayFrameLoopConsumerBase::Process_vkGetQueryPoolResults(
+    const ApiCallInfo&                          call_info,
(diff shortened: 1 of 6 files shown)


wzchu-lunarg:
Seems to be handled in https://github.com/LunarG/gfxreconstruct/pull/3176, closing.

PR #3216, opened 2026-08-22
Tools: Move command-line processing into features

MarkY-LunarG:
Optimize was a first approach at supporting command-line arguments in the features, but I decided to tweak it and do it a different way and make it global.  Then I implemented it in the replay tool and replaced the original mechanism in optimize.   This will eventually be migrated to other tools.

### DIFF
--- a/framework/util/feature_base.h
+++ b/framework/util/feature_base.h
@@ -26,9 +26,38 @@
 #include "util/defines.h"
 
 #include <string>
+#include <vector>
 
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(util)
+
+// One command-line entry that a Feature adds to the tool that contains it. A Feature returns
+// these entries from FeatureBase::GetOptionDescs(). The tool collects them before it builds its
+// ArgumentParser, so an entry is accepted only when its Feature is built into the tool.
+//
+// See tools/tool_feature_options.h for the functions that collect, validate, and print these
+// entries.
+struct FeatureOptionDesc
+{
+    // The name of the value that the entry takes, for the usage text. Example: "<backend>".
+    std::string name;
+
+    // The usage text for the entry. Each string is one line.
+    std::vector<std::string> description;
+
+    // True when the entry takes a value.
+    bool has_argument{ true };
+
+    // The names that trigger the entry, in ArgumentParser syntax. Put a pipe character between
+    // alternative names, as in "-o|--option-to-do-something". Do not use a comma.
+    std::string trigger_names;
+
+    // The full set of accepted values, when that set is closed. The tool adds the set to the
+    // usage text of the entry. A value that is not in the set gets a warning, and this Feature
+    // then uses the default value for the entry. The comparison ignores case. An empty vector
+    // accepts any value, and the Feature checks the value itself.
+    std::vector<std::string> accepted_values;
+};
 
 // Common interface required of every per-tool Feature base class (e.g. ReplayFeatureBase,
 // ConvertFeatureBase, ExtractFeatureBase, OptimizeFeature, InfoFeature) so that a Feature
@@ -47,6 +76,11 @@ class FeatureBase
     // suitable for printing as part of a tool's "--version" output. Returns an empty string
     // if this feature has no meaningful compiled-header version to report.
     virtual std::string CompiledHeaderVersionString() const = 0;
+
+    // The command-line entries that this Feature adds to its tool. The tool calls this function
+    // on each loaded Feature before it builds the ArgumentParser. The default is an empty list.
+    // A Feature that has no command-line entries of its own does not override this function.
+    virtual std::vector<FeatureOptionDesc> GetOptionDescs() const { return {}; }
 };
 
 GFXRECON_END_NAMESPACE(util)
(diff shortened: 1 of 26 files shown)


tools/replay/replay_vulkan_feature.cpp
```diff
@@ -35,12 +37,480 @@
...
+    if (arg_parser.IsOptionSet(kVirtualSwapchainSkipBlitLongOption) ||
+        arg_parser.IsOptionSet(kVirtualSwapchainSkipBlitShortOption))
+    {
+        replay_options.virtual_swapchain_skip_blit = true;
+    }
+
+    const std::string debug_severity_string = arg_parser.GetArgumentValue(kDebugMessageSeverityArgument);
+    if (!debug_severity_string.empty())
+    {
+        if (gfxrecon::util::platform::StringCompareNoCase("debug", debug_severity_string.c_str()))
```
fabian-lunarg: comparison is inverted. should be `if(platform::StringCompareNoCase(...) == 0){...}`
here and couple more cases below

MarkY-LunarG: Wow, good catch.  Looks like these were wrong from before as well.

MarkY-LunarG: Fixed.

framework/util/feature_base.h
```diff
@@ -26,10 +26,37 @@
...
+    // True when the entry takes a value.
+    bool has_argument{ true };
+
+    // The names that trigger the entry, in ArgumentParser syntax. Put a pipe character between
+    // alternative names, as in "-o|--option-to-do-something". Do not use a comma.
+    std::string trigger_names;
+
+    // The full set of accepted values, when that set is closed, the tool will reject a value
+    // that is not in the set. An empty vector accepts any value.
+    std::vector<std::string> accepted_values;
```
fabian-lunarg, line 59: `accepted_values` is never used/populated anywhere but used in three places (`CheckFeatureOptionValues`, `JoinFeatureOptionValues` and `PrintFeatureUsage`)

it 'would' be useful e.g. for options `-m none|remap|rebind|...`, `--swapchain`, `--present-mode`.
those currently do manual checks w. `StringCompareNoCase`, warn and fall back to their defaults.

it looks the intention was so `accepted_values` would generate the "The accepted values are:" logging,
which is currently dead.
wiring up the existing logic would also enable checking via `CheckFeatureOptionValues` and have one other implication:
we'd move from `StringCompareNoCase` to `std::find` over `std::string`: 
-> `--swapchain VIRTUAL` would start failing, but currently works (less important imo, but e.g. wrapper-scripts 'could' break)

tools/replay/replay_d3d12_feature.cpp
```diff
@@ -23,24 +23,209 @@
...
+const char kDxTwoPassReplay[]                  = "--dx12-two-pass-replay";
+const char kDxOverrideObjectNames[]            = "--dx12-override-object-names";
+const char kDxAgsMarkRenderPasses[]            = "--dx12-ags-inject-markers";
+const char kBatchingMemoryUsageArgument[]      = "--batching-memory-usage";
+const char kDumpResourcesModifiableStateOnly[] = "--dump-resources-modifiable-state-only";
+const char kDumpResourcesBeforeDrawOption[]    = "--dump-resources-before-draw";
+
+// Nothing reads this entry. It stays declared and documented so that a command line that
+// carries it is still accepted.
+const char kDumpResourcesImageFormatArgument[] = "--dump-resources-image-format";
```
fabian-lunarg: this cli-option isn't available for D3D12, 
so should be moved to the vulkan-feature (or deleted)?
the comment is misleading, on a Vulkan-only build the D3D12 feature is absent and replay would fail using the option

tools/replay/replay_vulkan_feature.cpp
```diff
@@ -50,6 +520,217 @@ std::string ReplayVulkanFeature::CompiledHeaderVersionString() const
...
+               { "Parameters to forward to the replay event plugin. The plugin determines",
+                 "the format of the parameters, and replay does not interpret them." },
+               true,
+               kReplayEventPluginParams },
+             { "",
+               { "Isolate the render passes. Replay splits the command buffer into more", "than one submit." },
+               false,
+               kIsolateRenderPasses },
+             { "<level>",
+               { "The highest debug messenger severity level to report. The levels are",
```
fabian-lunarg, line 713: highest -> lowest

tools/replay/replay_d3d12_feature.cpp
MarkY-LunarG: Looks completely unused.  Asking around, if it's dead, I'll just trim it.

tools/replay/replay_vulkan_feature.cpp
MarkY-LunarG: Good catch.  Fixing the wording for this and --log-level.

framework/util/feature_base.h
MarkY-LunarG: Done

tools/replay/replay_d3d12_feature.cpp
MarkY-LunarG: Looks like it's used, but through a different interface.  I'll move it.

fabian-lunarg (approved):

PR #3217, opened 2026-08-24
Track image layouts in `replay_consumer_base`

wzchu-lunarg:
`VulkanImageInfo::current_layout` is supposed to record the layout an image is actually on the GPU. This is useful for transitioning images, and we need to know the current layout when injecting barriers to use as the `oldLayout`. However, it is not being updated and tracked correctly. 

We have not run into issues during replay as replay normally uses the `oldLayout` from captured barriers. However, for injected commands, such as layout restoration during frame looping https://github.com/LunarG/gfxreconstruct/pull/3210, as we only know the layout from tracking this becomes an issue.

**Main changes**
Create `PropagateImageLayouts` from logic already in queue submits when `screenshot_handler_ != nullptr`, and do this propagation when even when not handling screenshots.

Rename uses of "Apply" as it was misleading, rename `ApplyRenderPassFinalLayouts` to `UpdateTrackedRenderPassFinalLayouts`.

Create `UpdateTrackedRenderingLayouts` for tracking layouts in dynamic rendering by tracking their attachments and the corresponding image views. Currently will handle color, depth and stencil attachments.


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4919,9 +4912,22 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit(PFN_vkQueueSubmit
         completion_source = GFXR_REPLAY_QUEUE_SUBMIT_COMPLETION_SOURCE_QUEUE_IDLE;
     }
 
+    // Update layout on the image infos.
+    if ((result == VK_SUCCESS) && (submit_info_data != nullptr) && RequiresImageLayoutTracking())
+    {
+        for (uint32_t i = 0; i < submitCount; ++i)
```
antonio-lunarg: I think you can simplify various parts of this PR's code using spans:
```cpp
for (auto submit : pSubmits->GetMetaStructSpan()) {
  for (auto command_buffer : submit.pCommandBuffers.GetSpan()) {
    PropagateImageLayouts(GetObjectInfoTable().GetVkCommandBufferInfo(command_buffer));
  }
}
```

wzchu-lunarg: Sounds good. Originally was just pulling it out of the loops when screenshot handling, but this looks better. Adding a commit for this.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -11268,12 +11373,111 @@ void VulkanReplayConsumerBase::ApplyRenderPassFinalLayouts(VulkanCommandBufferIn
...
+}
+
+void VulkanReplayConsumerBase::PropagateImageLayouts(const VulkanCommandBufferInfo* command_buffer_info)
+{
+    if (command_buffer_info == nullptr)
+    {
+        return;
+    }
+
+    for (const auto& [image_id, layout] : command_buffer_info->image_layout_barriers)
```
antonio-lunarg: Should we clear `image_layout_barriers` in `ClearCommandBufferInfo()`?

antonio-lunarg: How about image_layout_barriers from secondary command buffers? Worth having a look at `OverrideCmdExecuteCommands` to see how it merges `addresses_to_replace` from secondary into the primary.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -7569,6 +7585,95 @@ void VulkanReplayConsumerBase::OverrideCmdPipelineBarrier2KHR(
...
+         srcStageMask,
+         dstStageMask,
+         memoryBarrierCount,
+         pMemoryBarriers->GetPointer(),
+         bufferMemoryBarrierCount,
+         pBufferMemoryBarriers->GetPointer(),
+         imageMemoryBarrierCount,
+         pImageMemoryBarriers->GetPointer());
+
+    for (uint32_t i = 0; i < imageMemoryBarrierCount; ++i)
```
antonio-lunarg: I see this for block is replicated in `OverrideCmdWaitEvents2`, `OcerrideCmdPipelineBarrier`, and `OverrideCmdPipelineBarrier2`.
Time to extract into a common function :)

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -4912,9 +4912,20 @@ VkResult VulkanReplayConsumerBase::OverrideQueueSubmit(PFN_vkQueueSubmit
         completion_source = GFXR_REPLAY_QUEUE_SUBMIT_COMPLETION_SOURCE_QUEUE_IDLE;
     }
 
+    // Update layout on the image infos.
+    if ((result == VK_SUCCESS) && (submit_info_data != nullptr) && RequiresImageLayoutTracking())
+    {
+        for (auto submit : pSubmits->GetMetaStructSpan())
```
antonio-lunarg, line 5101: `const auto&`

framework/generated/khronos_generators/vulkan_generators/replay_overrides.json
```diff
@@ -64,6 +64,9 @@
     "vkCmdPipelineBarrier": "OverrideCmdPipelineBarrier",
     "vkCmdPipelineBarrier2": "OverrideCmdPipelineBarrier2",
     "vkCmdPipelineBarrier2KHR": "OverrideCmdPipelineBarrier2KHR",
+    "vkCmdWaitEvents": "OverrideCmdWaitEvents",
+    "vkCmdWaitEvents2": "OverrideCmdWaitEvents2",
+    "vkCmdWaitEvents2KHR": "OverrideCmdWaitEvents2KHR",
```
antonio-lunarg: You can map this directory to `OverrideCmdWaitEvents2`

framework/decode/vulkan_replay_consumer_base.cpp
wzchu-lunarg: Yes we should be clearing `image_layout_barriers`, didn't pull in from my other branch. Adding this as a commit.

wzchu-lunarg: > How about image_layout_barriers from secondary command buffers? Worth having a look at `OverrideCmdExecuteCommands` to see how it merges `addresses_to_replace` from secondary into the primary.

We should definitely be doing this for secondary command buffers too. We can just keep overwriting the image with the latest transition from the secondary command buffer when `CmdExecuteCommands` runs. Adding a commit for this.

framework/decode/vulkan_replay_consumer_base.cpp
wzchu-lunarg: Ah should have done this! Adding a commit for this.

framework/decode/vulkan_replay_consumer_base.cpp
wzchu-lunarg: Adding a commit for this.

framework/generated/khronos_generators/vulkan_generators/replay_overrides.json
wzchu-lunarg: Removing `OverrideCmdWaitEvents2KHR` completely then.

framework/decode/vulkan_replay_consumer_base.h
```diff
@@ -1925,6 +1973,22 @@ class VulkanReplayConsumerBase : public VulkanConsumer
 
     void WriteScreenshots(const Decoded_VkPresentInfoKHR* meta_info) const;
 
+    // Returns true if any replay-injected operation needs to know the layout images are currently in.
+    bool RequiresImageLayoutTracking() const
+    {
+        return (screenshot_handler_ != nullptr) || options_.dumping_resources || requires_image_layout_tracking_;
```
fabian-lunarg: this doesn't cover `VulkanVirtualSwapchain::PresentImageAdHoc`, 
and not sure if `options_.dumping_resources` actually belongs there.

also the branches guarded by `RequiresImageLayoutTracking()` aren't expensive.
future consumer of `VulkanImageInfo::current_layout` would require to remember this function and update it.
overall it 'looks' cleaner to drop it entirely and propagate layouts unconditionally.

fabian-lunarg (approved):
looks good to me. just had one comment, but not blocking.
approved when @antonio-lunarg is happy too

framework/decode/vulkan_replay_consumer_base.h
wzchu-lunarg: Was trying to be more conscious of the performance so I added this later, but I agree it looks cleaner and easier to work with without the guard. 

Adding a commit for this.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -11198,12 +11277,111 @@ void VulkanReplayConsumerBase::ApplyRenderPassFinalLayouts(VulkanCommandBufferIn
     }
 }
 
+void VulkanReplayConsumerBase::UpdateTrackedImageViewLayout(VulkanCommandBufferInfo* command_buffer_info,
```
antonio-lunarg, line 11619: nit: I think you can use this function in `UpdateTrackedRenderPassFinalLayouts` immediately above:
```cpp
for (size_t i = 0; i < render_pass_info->attachment_description_final_layouts.size(); ++i)
{
    UpdateTrackedImageViewLayout(command_buffer_info,
                                 attachment_image_view_ids[i],
                                 render_pass_info->attachment_description_final_layouts[i]);
}
```

antonio-lunarg (approved):

framework/decode/vulkan_replay_consumer_base.cpp
wzchu-lunarg: Yep, we should also do it for the stencils too.

Adding a commit for this.

PR #3218, opened 2026-08-24
Add state setup begin/end events to ReplayEventSink

nickdriscoll-lunarg:
This PR adds `GFXR_REPLAY_EVENT_STATE_SETUP_BEGIN` and `GFXR_REPLAY_EVENT_STATE_SETUP_END` to `GfxrReplayEventType` so that in the case of trimmed capture/replay plugins can differentiate between work that occurs during state setup and work that occurs during the actual trim range.


framework/plugin/public/gfxr/replay_event_plugin.h
```diff
@@ -43,6 +43,8 @@ typedef enum GfxrReplayEventType
     GFXR_REPLAY_EVENT_QUEUE_SUBMIT_END   = 2,
     GFXR_REPLAY_EVENT_FRAME_BEGIN        = 3,
     GFXR_REPLAY_EVENT_FRAME_END          = 4,
+    GFXR_REPLAY_EVENT_STATE_SETUP_BEGIN  = 5,
```
antonio-lunarg, line 50: Adding new event types implies bumping up `GFXR_REPLAY_PLUGIN_ABI_VERSION` (`2u`), but we still want to support older versions, so let's introduce a `GFXR_REPLAY_PLUGIN_ABI_MIN_VERSION` (`1u`).

antonio-lunarg: At that point `replay_event_plugin_loader.cpp:IsValidPlugin` needs to be updated. We are going to verify that the loaded plugin ABI version is within the supported range. E.g.:
```
plugin->abi_version >= GFXR_REPLAY_PLUGIN_ABI_MIN_VERSION && plugin->abi_version <= GFXR_REPLAY_PLUGIN_ABI_VERSION && ...
```

framework/plugin/replay_event_sink.cpp
```diff
@@ -39,6 +39,10 @@ static size_t GetSizeOfStruct(GfxrReplayEventType type)
             return sizeof(GfxrReplayFrameBeginEvent);
         case GFXR_REPLAY_EVENT_FRAME_END:
             return sizeof(GfxrReplayFrameEndEvent);
+        case GFXR_REPLAY_EVENT_STATE_SETUP_BEGIN:
```
antonio-lunarg: Let's improve this thing by computing everything at compile time.

We can introduce something like `EventTraits { GfxrReplayEventType type; size_t size; uint32_t since_version; }` and then a table to hold everything we need:
```cpp
constexpr EventTraits kEventTraits[] = {
    { GFXR_REPLAY_EVENT_QUEUE_SUBMIT_BEGIN, sizeof(GfxrReplayQueueSubmitBeginEvent), 1 },
    ...
    { GFXR_REPLAY_EVENT_STATE_SETUP_END,    sizeof(GfxrReplayStateSetupEndEvent),    2 },
};
```

framework/plugin/replay_event_sink.cpp
```diff
@@ -151,6 +171,16 @@ PluginReplayEventSink::~PluginReplayEventSink()
...
+}
+
+void PluginReplayEventSink::EmitStateSetupEnd(const GfxrReplayStateSetupEndEvent& event)
+{
+    Forward(event.header);
+}
+
 void PluginReplayEventSink::EmitQueueSubmitBegin(const GfxrReplayQueueSubmitBeginEvent& event)
 {
     Forward(event.header);
```
antonio-lunarg, line 170: `Forward()` needs to be improved. Once we have a new ABI version, we still want to support v1 plugins, so we need to store somewhere the negotiated version at plugin load time, and then the `Forward()` function will skip new events:
```cpp
if (since_version > negotiated_version) {
    return;
}
```

framework/plugin/test/main.cpp
```diff
@@ -35,6 +35,22 @@ class TestReplayEventSink : public gfxrecon::plugin::ReplayEventSink
     GfxrReplayFrameEndEvent       last_frame_end_event  = {};
 
   protected:
+    void EmitStateSetupBegin(const GfxrReplayStateSetupBeginEvent& event) override
```
antonio-lunarg, line 38: We need to exercise these paths with a new `TEST_CASE`.

framework/plugin/public/gfxr/replay_event_plugin.h
antonio-lunarg: I think it also makes sense to add a comment for each new enumerant: `// since ABI 2`

nickdriscoll-lunarg: I added the comments, `GFXR_REPLAY_PLUGIN_ABI_MIN_VERSION`, and changed the IsValidPlugin logic.

framework/plugin/replay_event_sink.cpp
nickdriscoll-lunarg: Switch statements are optimized into jump tables, and sizeof() is constexpr, so will this really make a difference?

nickdriscoll-lunarg: Oh but I see how we do need to store a `since_version` anyway. Will make the change.

framework/plugin/replay_event_sink.cpp
nickdriscoll-lunarg: I modified `IsValidPlugin` to report the negotiated version back to the `PluginReplayEventSink` and added statements to the Emit* functions so that they early-return if the negotiated version is lower than the version that event was added in.

framework/plugin/test/main.cpp
nickdriscoll-lunarg: Added

framework/plugin/replay_event_sink.cpp
```diff
@@ -56,6 +44,22 @@ GfxrReplayEventHeader ReplayEventSink::CreateEventHeader(GfxrReplayEventType typ
     return header;
```
antonio-lunarg, line 50: I think we should update the `abi_version` set for this header to something like `GetEventAbiVersion(type)`, rather than always `GFXR_REPLAY_PLUGIN_ABI_VERSION`. In other words, let's use `since_version` from the even traits table.

framework/plugin/replay_event_sink.cpp
```diff
@@ -151,23 +157,57 @@ PluginReplayEventSink::~PluginReplayEventSink()
     }
 }
 
+void PluginReplayEventSink::EmitStateSetupBegin(const GfxrReplayStateSetupBeginEvent& event)
+{
+    if (abi_version_ < 2)
```
antonio-lunarg: You can avoid manually checking every Emit function by doing this check in the `Forward` function.
There you can early return `if (event.abi_version > abi_version_)`.

framework/plugin/replay_event_sink.h
```diff
@@ -93,6 +119,7 @@ class PluginReplayEventSink final : public ReplayEventSink
     void Forward(const GfxrReplayEventHeader& event);
 
     bool                          disabled_      = false;
+    uint32_t                      abi_version_   = 0;
```
antonio-lunarg: My suggestion here is to either add a comment mentioning that this holds a version negotiated between the loader and the plugin, or just rename it to `negotiated_abi_version_`, or both :D

antonio-lunarg (commented):
`docs/PLUGIN_replay_event.md` also needs to be updated with the new events and we need to reword the part that says: `the loader accepts only an exact match of the version today` which is no longer going to be true.

docs/PLUGIN_replay_event.md
```diff
@@ -163,6 +165,14 @@ This event is intended for:
...
 `FrameEnd` is emitted when replay reaches the frame boundary according to replay frame-boundary semantics. It carries the first and last submit indices associated with that replay frame, or the invalid submit sentinel when the frame contains no submits.
 
+### StateSetupBegin
+
+`StateSetupBegin` is emitted right before state setup commands start getting executed.
+
+### StateSetupEnd
+
+`StateSetupEnd` is emitted right after the last state setup command.
+
```
antonio-lunarg, line 180: There are few lines below that belong to the FrameEnd event:
```
This event is intended for:

* frame duration visualization  
* association of submit ranges with a replay frame
```
Once that's fixed, LGTM!

antonio-lunarg (approved):

PR #3220, opened 2026-08-24
Track image subresource layouts

wzchu-lunarg:
**Problem**
Currently, GFXR tracks image layouts as a single enum. We see this in `vulkan_object_info`'s `VulkanImageInfo` where we have `VkImageLayout` as the `current_layout`.

However, Vulkan image layouts aren't always a uniform layout representing the whole image. As mentioned in Oleh's branch's [notes](https://github.com/olehkuznetsov/gfxreconstruct/blob/4c2e534fe98b2ce0f6c3779482e8ae3ea6a7edc9/frame_loop_layouts.md), some common examples are depth/stencil images and mipmapped textures. It boils down to image having layouts for each subresource, where each subresource is accessed by it's (`aspect`, `mip_level`, `layer`) tuple.

This PR aims to pull in some of Oleh's changes as described, and also as a contained refactor of the existing uses of an image's `current_layout`.

This would allow accurate image layout restoration during frame looping, as we need to track the initial layout of an image properly in order to restore it.

**Main Changes**
Changes are based off a few commits in https://github.com/LunarG/gfxreconstruct/pull/3217 for image propagation and tracking as those changes also interact with image layouts.

Create class `ImageLayoutMap` in **vulkan_image_layout_map.h.** This class keeps track of the layout for each subresource. It has a setter and getter:
```
void ImageLayoutMap::SetLayout(const VkImageSubresourceRange& range, VkImageLayout layout)

VkImageLayout ImageLayoutMap::GetLayout(VkImageAspectFlagBits aspect, uint32_t mip_level, uint32_t array_layer)
```
Stub existing consumers of `current_layout` with this new structure with `subresource_layouts.GetLayout(aspect, 0, 0)` to read the first subresource. 

Writes to `current_layout` now uses the actual ranges.

**vulkan_object_info.h**

- Update `VulkanImageInfo::current_layout` from `VkImageLayout` to `ImageLayoutMap subresource_layouts`.
- Update `VulkanCommandBufferInfo::image_layout_barriers` from `unordered_map<HandleId, VkImageLayout>` to `std::unordered_map<format::HandleId, ImageLayoutMap>`. 
- Add `VulkanImageViewInfo::subresource_range`. An attachment only transitions the subresources its view covers, not the whole image, so tracking a render pass or dynamic rendering attachment needs the view's range.
- Add `VulkanRenderPassInfo::attachment_description_stencil_final_layouts`, so a depth/stencil attachment that ends a pass with split layouts can record both.

**vulkan_command_buffer_util.h**

- Add `InitializeCommandBufferImageLayouts`, which sizes a command buffer's map for an image from that image's own map if **not initialized** yet. Record sites call this before `SetLayout`.


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -11264,13 +11320,157 @@ void VulkanReplayConsumerBase::ApplyRenderPassFinalLayouts(VulkanCommandBufferIn
...
+    }
+
+    VulkanImageInfo* image_info = object_info_table_->GetVkImageInfo(image_view_info->image_id);
+    if (image_info == nullptr)
+    {
+        return;
+    }
+
+    InitializeCommandBufferImageLayouts(command_buffer_info, image_info);
+    command_buffer_info->image_layout_barriers[image_view_info->image_id].SetLayout(image_view_info->subresource_range,
```
fabian-lunarg: aspect is not honored here, causing issues for combined depth/stencil images.
we need to bail out here when the aspect does not match, like done at :11342

```
VkImageSubresourceRange range = image_view_info->subresource_range;
range.aspectMask &= graphics::GetLayoutAspects(layout);
if (range.aspectMask == 0){ return; } // -> not the aspect you're looking for :)
command_buffer_info->image_layout_barriers[image_view_info->image_id].SetLayout(range, layout);
```

framework/graphics/vulkan_image_layout_map.cpp
```diff
@@ -0,0 +1,237 @@
...
+            }
+        }
+    }
+}
+
+VkImageLayout ImageLayoutMap::GetLayout(VkImageAspectFlagBits aspect, uint32_t mip_level, uint32_t array_layer) const
+{
+    if (is_uniform_)
+    {
+        return uniform_layout_;
```
fabian-lunarg: header says the function returns `VK_IMAGE_LAYOUT_UNDEFINED` "if the image has no such aspect",
but the `is_uniform_` branch doesn't check that.

e.g. a depth-only image in `DEPTH_ATTACHMENT_OPTIMAL` would yield:
`GetLayout(VK_IMAGE_ASPECT_COLOR_BIT, 0, 0)` -> `DEPTH_ATTACHMENT_OPTIMAL` here.

recommend to move up the branch below and a minor addition to the check:
```
if (((aspect & aspects_) == 0) || (mip_level >= mip_levels_) || (array_layer >= array_layers_))
{
   return VK_IMAGE_LAYOUT_UNDEFINED;
}
```

like this the uniform vs. non-uniform code-paths always agree.

framework/graphics/vulkan_image_layout_map.h
```diff
@@ -0,0 +1,100 @@
...
+    {
+        VkImageSubresourceRange range;
+        VkImageLayout           layout;
+    };
+
+    ImageLayoutMap() = default;
+
+    void Initialize(uint32_t mip_levels, uint32_t array_layers, VkImageAspectFlags aspects);
+
+    bool               IsInitialized() const { return aspects_ != 0; }
```
fabian-lunarg: you 'could' sprinkle `[[nodiscard]]` for all newly added getters here.
it's not just decoration, but could prevent actual mistakes.

framework/decode/vulkan_object_info.h
```diff
@@ -727,12 +730,15 @@ struct VulkanCommandPoolInfo : public VulkanPoolInfo<VkCommandPool>
...
-    format::HandleId                                          active_render_pass_id{ format::kNullHandleId };
-    format::HandleId                                          active_framebuffer_id{ format::kNullHandleId };
-    std::vector<format::HandleId>                             active_render_pass_attachment_image_view_ids;
-    std::unordered_map<format::HandleId, VkImageLayout>       image_layout_barriers;
+    bool                          is_frame_boundary{ false };
+    std::vector<format::HandleId> frame_buffer_ids;
+    format::HandleId              active_render_pass_id{ format::kNullHandleId };
+    format::HandleId              active_framebuffer_id{ format::kNullHandleId };
+    std::vector<format::HandleId> active_render_pass_attachment_image_view_ids;
+    std::vector<format::HandleId> executed_secondary_command_buffers;
```
fabian-lunarg: `executed_secondary_command_buffers` appears to be unused.
if it's a leftover we should remove it, otherwise wire it up.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -11264,13 +11320,157 @@ void VulkanReplayConsumerBase::ApplyRenderPassFinalLayouts(VulkanCommandBufferIn
...
+                                      ? rendering_meta->pDepthAttachment->GetMetaStructPointer()
+                                      : nullptr);
+    UpdateTrackedAttachmentLayout(command_buffer_info,
+                                  rendering_info->pStencilAttachment,
+                                  (rendering_meta->pStencilAttachment != nullptr)
+                                      ? rendering_meta->pStencilAttachment->GetMetaStructPointer()
+                                      : nullptr);
+}
+
+void VulkanReplayConsumerBase::PropagateImageLayouts(const VulkanCommandBufferInfo* command_buffer_info)
```
fabian-lunarg, line 11710: this method is on a rather hot code-path (once per submitted command-buffer, when tracking is on)
and `GetSubresourceLayouts()` heap-allocates once per image, uniform maps included (pushes into
an empty vector).

this 'could' be written in terms of **merging** two layout-maps.
via an (suggestion) API like `void ImageLayoutMap::MergeFrom(const ImageLayoutMap& other)`

and the loop below would collapse to:

```
for (const auto& [image_id, command_buffer_layouts] : command_buffer_info->image_layout_barriers)
{
      VulkanImageInfo* image_info = object_info_table_->GetVkImageInfo(image_id);
      if (image_info != nullptr)
      {
          image_info->subresource_layouts.MergeFrom(command_buffer_layouts); // 'MergeFrom' should/would be free from heap-allocations
      }
}
```

framework/decode/vulkan_object_info.h
wzchu-lunarg: This was supposed to be for tracking image transitions in secondary command buffers. However, it is leftover, and tracking of these transitions should be done in `OverrideCmdExecuteCommands` in the consumer base. 

Adding a commit for this.

framework/decode/vulkan_replay_consumer_base.cpp
wzchu-lunarg: Adding a commit for this

framework/decode/vulkan_replay_consumer_base.cpp
wzchu-lunarg: This sounds good. This way we don't need to jump through these hurdles of creating all the ranges and it's associated layouts. Will create a method that will copy all the non undefined layouts (command buffer touched) from a given layout.

Adding a commit for this.

framework/graphics/vulkan_image_layout_map.cpp
wzchu-lunarg: Completely missed this. 

Adding a commit for this.

framework/graphics/vulkan_image_layout_map.h
wzchu-lunarg: Good point. This would definitely be helpful.

Adding a commit.

framework/graphics/vulkan_image_layout_map.h
```diff
@@ -0,0 +1,103 @@
...
+
+    void SetLayout(const VkImageSubresourceRange& range, VkImageLayout layout);
+
+    // Merges every subresource layout from `src` that is not VK_IMAGE_LAYOUT_UNDEFINED into this map.
+    void MergeFrom(const ImageLayoutMap& src);
+
+    // Layout of a single subresource, or VK_IMAGE_LAYOUT_UNDEFINED if the image has no such aspect.
+    [[nodiscard]] VkImageLayout GetLayout(VkImageAspectFlagBits aspect, uint32_t mip_level, uint32_t array_layer) const;
+
+    [[nodiscard]] std::vector<RangeLayout> GetSubresourceLayouts() const;
```
fabian-lunarg: nice, looks the addition of `MergeFrom` made `GetSubresourceLayouts()` 
and `struct RangeLayout` entirely obsolete now.
-> could/should be removed if there are no other plans for it.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -5931,6 +5955,15 @@ void VulkanReplayConsumerBase::OverrideCmdExecuteCommands(PFN_vkCmdExecuteComman
             in_commandBuffer->addresses_to_replace.insert(secondary_cmd_buffer_info->addresses_to_replace.begin(),
                                                           secondary_cmd_buffer_info->addresses_to_replace.end());
         }
+
+        if (RequiresImageLayoutTracking())
+        {
+            // Update the image's layout based on the recorded layout transitions in the secondary command buffer.
+            for (const auto& [image_id, layout] : secondary_cmd_buffer_info->image_layout_barriers)
+            {
+                in_commandBuffer->image_layout_barriers[image_id] = layout;
```
fabian-lunarg: this overwrites the primary barriers unconditionally but should merge (when already initialized) or initialize like:

```
for (const auto& [image_id, secondary_layouts] : secondary_cmd_buffer_info->image_layout_barriers){
    auto& primary_layouts = in_commandBuffer->image_layout_barriers[image_id];
    if (primary_layouts.IsInitialized()) { primary_layouts.MergeFrom(secondary_layouts); }
    else { primary_layouts = secondary_layouts; }
}
```

framework/graphics/vulkan_image_layout_map.cpp
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    if (is_uniform_)
+    {
+        ExpandUniform();
+    }
+
+    GFXRECON_ASSERT((mip_levels_ == src.mip_levels_) && (array_layers_ == src.array_layers_) &&
+                    (aspects_ == src.aspects_));
+
```
fabian-lunarg, line 181: the assert is good, but Release could still hit OOB (in `src.subresource_layouts_[i]`). we should check in any case and log an error.

```
if ((mip_levels_ != src.mip_levels_) || (array_layers_ != src.array_layers_) || (aspects_ != src.aspects_))
{
    GFXRECON_LOG_ERROR("ImageLayoutMap::MergeFrom: mismatched layout-map dimensions, skipping merge");
    return;
}
```

reasoning for `GFXRECON_LOG_ERROR`:

- it's a violated invariant here -> that's an error
- we don't expect this, but if we'd see it we're interested in the frequency too -> `GFXRECON_LOG_ERROR_ONCE` would brush information under the rug

fabian-lunarg:
not for this PR ... `vkCmdWaitEvents` / `vkCmdWaitEvents2` 'can' carry image memory barriers too
and have no overrides (yet), 
-> transitions done via events aren't tracked and fail silently: 
-> the map keeps the previous barrier's layout, and the `UNDEFINED` fallback in the consumers won't work.

this is not critical and/or blocking for this PR, but should be mentioned.
your call if you want to create a tracking issue or pick up in a follow-up.

framework/graphics/vulkan_image_layout_map.h
wzchu-lunarg: Thought I would keep it if we ever needed to get the a list of the layouts, but I'll add a commit to remove it since it does seem unnecessary.

framework/decode/vulkan_replay_consumer_base.cpp
wzchu-lunarg: Completely missed this. Was pulling in similar changes from a single layout branch. 

Adding a commit for this.

framework/graphics/vulkan_image_layout_map.cpp
wzchu-lunarg: That's a good point. It felt like to me the mismatched layouts should not be reachable anyways unless there was a bad capture. However, we still want to handle it better than just an assert.

Adding a commit for this.

wzchu-lunarg:
> not for this PR ... `vkCmdWaitEvents` / `vkCmdWaitEvents2` 'can' carry image memory barriers too and have no overrides 

This is going to be followed up in a separate PR https://github.com/LunarG/gfxreconstruct/pull/3217, where the main focus is to cover more cases of layout tracking as it didn't seem to be correct before.

fabian-lunarg (approved):

framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
```diff
@@ -3154,7 +3155,14 @@ static void UpdateOriginalCommandBufferWithNewImageLayouts(const VulkanRenderPas
         {
             att_img_view_info =
                 object_info_table.GetVkImageViewInfo(framebuffer_info->attachment_image_view_ids[att_ref.attachment]);
-            original_command_buffer_info->image_layout_barriers[att_img_view_info->image_id] = att_ref.layout;
+
+            VulkanImageInfo* att_img_info = object_info_table.GetVkImageInfo(att_img_view_info->image_id);
```
antonio-lunarg, line 3102: Let's assert `att_img_view_info != nullptr` before dereferencing it

framework/graphics/vulkan_image_layout_map.cpp
```diff
@@ -0,0 +1,232 @@
...
+void ImageLayoutMap::SetLayout(const VkImageSubresourceRange& range, VkImageLayout layout)
+{
+    if (!IsInitialized())
+    {
+        return;
+    }
+
+    // A range naming no aspect this image has tells us nothing about which subresources moved, so record nothing.
+    const VkImageAspectFlags aspect_mask = range.aspectMask & aspects_;
```
antonio-lunarg: I think there's an issue here. We can't assume barrier and format use the same names.

In case of a multi-planar image, `range.aspectMask` is going to be `COLOR` because the spec requires it in order to specify "all planes", but `aspects_` is going to be something like `PLANE_0 | PLANE_1`, so the result of this operation is `0`.

panos-lunarg (approved):
As discussed with @wzchu-lunarg, VDR need more work in order to benefit by the new tracking method. However I believe that this is out of scope for this PR but should be tracked by a separate issue.

framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
wzchu-lunarg: This looks good.

Adding a commit for this.

framework/graphics/vulkan_image_layout_map.cpp
wzchu-lunarg: Didn't realize this was a case for multi-planar images.

Adding a commit to address this case.

framework/graphics/vulkan_image_layout_map.cpp
```diff
@@ -0,0 +1,241 @@
...
+        if (layout != VK_IMAGE_LAYOUT_UNDEFINED)
+        {
+            subresource_layouts_[i] = layout;
+        }
+    }
+}
+
+VkImageLayout ImageLayoutMap::GetLayout(VkImageAspectFlagBits aspect, uint32_t mip_level, uint32_t array_layer) const
+{
+    if ((aspect & aspects_) == 0 || (mip_level >= mip_levels_) || (array_layer >= array_layers_))
```
antonio-lunarg, line 195: I think we have the same problem here, when calling `GetLayout(COLOR, ...)` and `aspects_` is multiplanar.

framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
```diff
@@ -3182,15 +3192,23 @@ static void UpdateOriginalCommandBufferWithNewImageLayouts(const VulkanRenderPas
...
+
+        if (att_img_view_info != nullptr)
+        {
+            VulkanImageInfo* att_img_info = object_info_table.GetVkImageInfo(att_img_view_info->image_id);
+            if (att_img_info != nullptr)
+            {
+                InitializeCommandBufferImageLayouts(original_command_buffer_info, att_img_info);
+                original_command_buffer_info->image_layout_barriers[att_img_view_info->image_id].SetLayout(
+                    att_img_view_info->subresource_range,
+                    original_render_pass_ci->pSubpasses[0].pDepthStencilAttachment->layout);
```
antonio-lunarg, line 3151: Should we use a value from `current_subpass_` instead of `0` to index `pSubpasses`, @panos-lunarg?

framework/graphics/vulkan_image_layout_map.cpp
wzchu-lunarg: I believe the difference here is that we shouldn't really be trying to get the `COLOR_BIT` subresource for a multiplanar image. Multiplanar images only use this `COLOR_BIT` as a convention that means "all planes" even though they don't actually have a subresource at that aspect. However, the intention of this function is to get the layout of a singular subresource indexed by it's tuple (aspect, mip, layer), which doesn't really apply to this case, as we shouldn't be trying to get the layouts of all the planes with this function. 

I believe usage of this function should be similar to `vkGetImageSubresourceLayout`:
>VUID-vkGetImageSubresourceLayout-tiling-08717
If the tiling of the image is [VK_IMAGE_TILING_LINEAR](https://docs.vulkan.org/refpages/latest/refpages/source/VkImageTiling.html) and has a [multi-planar format](https://docs.vulkan.org/spec/latest/chapters/formats.html#formats-multiplanar), then the aspectMask member of pSubresource must be a single valid [multi-planar aspect mask](https://docs.vulkan.org/spec/latest/chapters/formats.html#formats-multiplanar-image-aspect) bit

although similar confusions could arise and maybe we should have a better name instead of the generic `GetLayout` to make this more clear.

The `COLOR_BIT` as "all planes" does matter in certain other usages, such as `VkImageSubresourceRange` and transitions (our `SetLayout`).

I think changing the name from `GetLayout` to `GetSubresourceLayout` will make it a little clearer about the intentions. Going to add a commit for this.

framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
antonio-lunarg: Resolving here as out of scope.

antonio-lunarg (approved):

PR #3222, opened 2026-08-25
Adds nullchecks in dx12_replay_consumer

tellingt-amd:
## Problem
There are several instances in dx12_replay_consumer.cpp where one of the get functions returns a pointer that may be null. Most of these are properly checked before use, however in several places, the object is assumed to be valid and members of the object are directly accessed or checked. 

For example on line 459:
``` 
    auto object_info = GetObjectInfo(resource_info.second.resource_id);
    if (object_info->extra_info != nullptr)
```
object_info can be null here since `GetObjectInfo` can return null.

Further down `GetExtraInfo<>` can do the same thing.

## Fix
Added several if-checks to catch potential null pointer dereferences. 
Added warnings to make sure that if these new checks catch null pointers, they do not fail silently.

### DIFF
--- a/framework/decode/dx12_replay_consumer_base.cpp
+++ b/framework/decode/dx12_replay_consumer_base.cpp
@@ -456,7 +456,7 @@ void Dx12ReplayConsumerBase::ApplyBatchedResourceInitInfo(
         for (auto& resource_info : resource_infos)
         {
             auto object_info = GetObjectInfo(resource_info.second.resource_id);
-            if (object_info->extra_info != nullptr)
+            if (object_info != nullptr && object_info->extra_info != nullptr)
             {
                 auto extra_info = GetExtraInfo<D3D12ResourceInfo>(object_info);
                 if (extra_info->swap_chain_id != format::kNullHandleId)
@@ -467,6 +467,12 @@ void Dx12ReplayConsumerBase::ApplyBatchedResourceInitInfo(
                 {
                     others_resource_infos.insert(std::pair(resource_info.first, &resource_info.second));
                 }
+            }
+            else
+            {
+                GFXRECON_LOG_WARNING("ApplyBatchedResourceInitInfo: skipping resource_id %" PRIu64
+                                     " with missing object/extra info; its init data will not be applied.",
+                                     resource_info.second.resource_id);
             }
         }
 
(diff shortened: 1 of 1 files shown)


framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -533,9 +539,23 @@ void Dx12ReplayConsumerBase::ApplyBatchedResourceInitInfo(
         {
             auto object_info          = GetObjectInfo(resource_info.second->resource_id);
             auto extra_info           = GetExtraInfo<D3D12ResourceInfo>(object_info);
+            if (extra_info == nullptr)
```
davidd-lunarg, line 542: It shouldn't be possible for this resource `extra_info` to be nullptr because of the check at 459:
`if (object_info != nullptr && object_info->extra_info != nullptr)`
So the entries in `swapchain_resource_infos` should all have non-null `extra_info`.

Even if it is null, this `extra_info` object is already dereferenced above at line 493:
`auto swapchain_info        = GetObjectInfo(extra_info->swap_chain_id);`
so replay would crash before reaching this check.

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -468,6 +468,12 @@ void Dx12ReplayConsumerBase::ApplyBatchedResourceInitInfo(
                     others_resource_infos.insert(std::pair(resource_info.first, &resource_info.second));
                 }
             }
+            else
+            {
+                GFXRECON_LOG_WARNING("ApplyBatchedResourceInitInfo: skipping resource_id %" PRIu64
```
davidd-lunarg, line 473: Previously no warning was fired when `object_info->extra_info == nullptr` but now this warning will be emitted so this changes behavior in that case.

framework/decode/dx12_replay_consumer_base.cpp
```diff
@@ -533,9 +539,23 @@ void Dx12ReplayConsumerBase::ApplyBatchedResourceInitInfo(
...
+            {
+                GFXRECON_LOG_WARNING("ApplyBatchedResourceInitInfo: skipping swapchain resource_id %" PRIu64
+                                     "; missing D3D12ResourceInfo extra info",
+                                     resource_info.second->resource_id);
+                continue;
+            }
             auto swapchain_info       = GetObjectInfo(extra_info->swap_chain_id);
             auto swapchain_extra_info = GetExtraInfo<DxgiSwapchainInfo>(swapchain_info);
-
+            if (swapchain_extra_info == nullptr)
```
davidd-lunarg, line 551: This `swapchain_extra_info` object is already dereferenced above at line 527:
`resource_data_util_->ExecuteAndWaitForCommandList(swapchain_extra_info->command_queue);`
so replay would crash before reaching this check.

davidd-lunarg (changes requested):
Instead of adding null checks to the control flow and emitting warnings, I recommend using `GFXRECON_ASSERT(object_info/extra_info != nullptr)` at the point where the objects are first resolved (e.g., after lines 458, 492, 493, and 494). It isn't expected for these objects to be null and AFAIK we haven't encountered cases where that happened.

PR #3224, opened 2026-08-26
docs: Plugin ABI versioning

antonio-lunarg:
The replay event plugin ABI is a binary contract — a customer builds a plugin
once and never rebuilds it — but nothing writes down how that contract may grow.

- `docs/PLUGIN_replay_event.md`: new "ABI Versioning and Compatibility" section
- `docs/GFXReconstruct Architectural Principles.md`: new `PLUGIN-ABI` entry. A plugin `.so` must
  keep working unmodified.

The loader still requires an exact version match; the section notes this where it
introduces `GFXR_REPLAY_PLUGIN_ABI_MIN_VERSION`. A follow-up PR adds the range
check and the host-side event gating.

### DIFF
--- a/docs/GFXReconstruct Architectural Principles.md
+++ b/docs/GFXReconstruct Architectural Principles.md
@@ -24,6 +24,7 @@
 - [Backward Compatibility (major revision)](#backward-compatibility-within-a-major-revision)
   - [Format](#format)
   - [Interface](#interface)
+  - [Plugin ABI](#plugin-abi)
   - [Compatibility](#compatibility)
   - [Invocation](#invocation)
   - [Convert-JSON](#convert-json)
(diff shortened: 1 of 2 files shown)


nickdriscoll-lunarg (approved):

bradgrantham-lunarg:
@MarkY-LunarG relevant to versioning work you are doing?

docs/GFXReconstruct Architectural Principles.md
```diff
@@ -223,6 +224,20 @@ Existing function signatures should not be removed or altered for functions in 
...
 
+## PLUGIN-ABI
+
+**Replay will continue to load plugin binaries that customers built with an older public header.**
+
+The public plugin ABI in `framework/plugin/public/gfxr/replay_event_plugin.h` is a binary contract, not a source contract.  A customer builds a replay event plugin one time and gives replay a shared library.  When the customer updates GFXReconstruct and does not rebuild the plugin, that library must keep its behavior.
+
+This commitment is stronger than [INTERFACE](#interface).  INTERFACE limits compatibility to source code.  PLUGIN-ABI applies to the binary layout of the public structs, to the numbers of the enumerants, and to the names of the factory symbols.
+
+A change that only adds data increases `GFXR_REPLAY_PLUGIN_ABI_VERSION` by one.  Replay and the plugin negotiate one version at load time.  Replay then emits only the events that this version contains.  A plugin that a customer built with an older header receives the same event stream as before.
```
MarkY-LunarG, line 235: How does replay negotiate this version?  Can they disagree?  Like if GFXR replay stops supporting version 1, and the library only supports version 1 what happens?

MarkY-LunarG: If you look at the loader interface, it has a "Negotiate" function.

antonio-lunarg: The gfxreconstruct replay app has a plugin loader in `replay_event_plugin_loader.cpp`. The loader checks whether the plugin is valid in `IsValidPlugin()`, which is also where the negotiation should happen. Currently, the check lincludes: `plugin->abi_version == GFXR_REPLAY_PLUGIN_ABI_VERSION`. The idea is to make it like this:
```cpp
plugin->abi_version >= GFXR_REPLAY_PLUGIN_ABI_MIN_VERSION && plugin->abi_version <= GFXR_REPLAY_PLUGIN_ABI_VERSION
```

If that function fails, plugin loading fails, an error is logged, plugin is destroyed and the corresponding library unloaed. Gfxreconstruct does not treat this as a critical error, and proceeds with replay.

MarkY-LunarG (approved):

PR #3227, opened 2026-08-27
D3D12 trim-load staging buffers in a reusable UPLOAD heap

JerryAMD:
The GFXR create Committed Resource as stagging buffer when upload data from capture file over to resource (InitSubResource). But Committed Resource Create&Release will take some CPU&GPU time.

The solution is,
Create a big Heap when start InitSubResource, then create all batched stagging buffer as PlacedResource to the Heap. So that could save some time to allocate CPU&GPU memory.

### DIFF
--- a/framework/graphics/dx12_resource_data_util.cpp
+++ b/framework/graphics/dx12_resource_data_util.cpp
@@ -1,6 +1,6 @@
 /*
 ** Copyright (c) 2021 LunarG, Inc.
-** Copyright (c) 2025 Advanced Micro Devices, Inc. All rights reserved.
+** Copyright (c) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
 **
 ** Permission is hereby granted, free of charge, to any person obtaining a
 ** copy of this software and associated documentation files (the "Software"),
@@ -249,7 +249,8 @@ void Dx12ResourceDataUtil::GetResourceCopyInfo(ID3D12Resource*                  
 
 Dx12ResourceDataUtil::Dx12ResourceDataUtil(ID3D12Device* device, uint64_t min_buffer_size) :
     device_(device), staging_buffers_{ nullptr, nullptr }, staging_buffer_sizes_{ 0, 0 },
-    min_buffer_size_(min_buffer_size), fence_value_(0)
+    min_buffer_size_(min_buffer_size), fence_value_(0), batch_upload_heap_(nullptr), batch_heap_capacity_(0),
+    batch_heap_offset_(0), batch_heap_adapter_(nullptr), batch_heap_max_mem_usage_(0.0), batch_heap_is_uma_(false)
 {
     HRESULT result = E_FAIL;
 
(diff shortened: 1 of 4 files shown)


framework/graphics/dx12_resource_data_util.cpp
```diff
@@ -894,5 +895,156 @@ Dx12ResourceDataUtil::ExecuteTransitionCommandList(ID3D12Resource*
...
+    // at least the immediate request, capped at the max. Rounded up to placement alignment.
+    uint64_t capacity = std::max(kBatchHeapDefaultCapacity, min_buffer_size_);
+    capacity          = std::max(capacity, min_capacity);
+    capacity          = std::min(capacity, kBatchHeapMaxCapacity);
+    capacity          = util::platform::AlignValue<kBatchPlacementAlign>(capacity);
+
+    // Don't attempt a large heap allocation that the GPU/system memory budget can't accommodate. Check before
+    // touching any existing heap so a rejected grow leaves the current heap usable; the caller falls back to
+    // per-resource committed staging buffers. Skipped when no adapter was supplied (budget unknown).
+    if ((batch_heap_adapter_ != nullptr) &&
```
davidd-lunarg: A nullptr adapter here allows default batch size. If the adapter is nullptr, `IsMemoryAvailable` can't be called so GFXR can't know how much memory is available and batching should be disabled.

framework/graphics/dx12_resource_data_util.cpp
```diff
@@ -894,5 +895,156 @@ Dx12ResourceDataUtil::ExecuteTransitionCommandList(ID3D12Resource*
...
+    {
+        return S_OK;
+    }
+
+    // Choose a bounded capacity: at least the default and the largest single resource seen (min_buffer_size_),
+    // at least the immediate request, capped at the max. Rounded up to placement alignment.
+    uint64_t capacity = std::max(kBatchHeapDefaultCapacity, min_buffer_size_);
+    capacity          = std::max(capacity, min_capacity);
+    capacity          = std::min(capacity, kBatchHeapMaxCapacity);
+    capacity          = util::platform::AlignValue<kBatchPlacementAlign>(capacity);
```
davidd-lunarg: - `uint64_t capacity = std::max(kBatchHeapDefaultCapacity, min_buffer_size_);`

`min_buffer_size_` is initialized to be the size of the largest resource in the trim state, so a trim that contains a large (e.g., 1gb resource) will always pick that.

- `capacity          = std::max(capacity, min_capacity);`

`min_capacity` is the current resource size that wants to be batched, so this line is a no-op since the previous line sets capacity to be at least as large as the single largest resource.

framework/graphics/dx12_resource_data_util.cpp
```diff
@@ -894,5 +895,156 @@ Dx12ResourceDataUtil::ExecuteTransitionCommandList(ID3D12Resource*
...
+    {
+        // Treat as a fit failure; caller flushes/resets and falls back to a committed buffer if needed.
+        return nullptr;
+    }
+
+    batch_heap_offset_ = aligned_offset + required_buffer_size;
+    return placed;
+}
+
+void Dx12ResourceDataUtil::ResetBatchHeap()
```
davidd-lunarg: On creation ID3D12Heap is "almost always zeroed" according to the spec. So reset should probably also clear the heap with 0's to prevent possible garbage data ending up in a destination resource. For example, this could happen if the replay resource required size is larger than the size of the init data in the capture file.

PR #3229, opened 2026-08-27
Add option to present when trace lacks presentation but has frame boundaries

werman:
With certain VR games I ran into the issue of them only having frame boundaries, but not any presentation. So even `--present-override` did nothing since there is no presentation. Another issue I ran into is VR images having 2 layers, both of which are needed to be either screenshotted or presented.

This PR:
- Adds `--present-frame-boundary` to present frame-boundary images;
- Screenshots now dump all layers;


MarkY-LunarG:
The screenshot API has changed since you've created this PR.  It now has a ScreenshotController which works with an API Screenshot source.  Please rebase on this and ping us again once that's done.

werman:
@MarkY-LunarG Rebased and tested, should be good.

werman:
Hmm, was not enough to show rendering of another games, where game created several images with the same name. I have to track which image is actually used in the submission to display the right one. 

Would such tracking be acceptable addition?

PR #3230, opened 2026-08-27
Expose current frame number to sibling layers

hongSARC:
Adds `GetCurrentFrameGFXR` to the custom layer function table, so a sibling layer in the same chain can read the frame number that the capture layer is recording. This mirrors the existing `GetBlockIndexGFXR` entry point, which exists for the same reason one field over.

Any tool that wants to correlate its own per-frame data with the capture file has to either reimplement all of that detection or drift out of step with it. We hit that drift twice: once missing `VkFrameBoundaryEXT` and once missing the VR frame delimiter. In both cases the symptom was that our tooling reported one long frame while the capture file correctly contained many.
Reading the count directly removes the duplication and the drift.

Purely additive. Nothing inside the capture layer calls it, so there is no behavior change for existing users.


fabian-lunarg (approved):

hongSARC:
@fabian-lunarg Hello, the PR is approved, but the checks all show "waiting for status to be reported", looks like GitHub's holding the workflow run for maintainer approval since this is from a fork. Are there other steps that I need to do from my end? Thanks.

panos-lunarg:
Apparently the github runners were waiting for approval before running. I clicked the button and it seems that they have started

hongSARC:
> Apparently the github runners were waiting for approval before running. I clicked the button and it seems that they have started

Thank you!

hongSARC:
I only saw update branch button and pressing on that seems to cause another request needed. @panos-lunarg Can you give it another try?

hongSARC:
I'm guessing I don't have access to push any of the code, only update branch button for me.

PR #3231, opened 2026-08-27
Fix for crash on Android Emulator

richard-lunarg:
Games on the Android emulator on macOS that use the JAVA UI instead of their own embedded GUI framework are crashing on launch when I attempt to capture them. This change vastly improves the number of games I'm able to capture (tested in use already).

Why: The emulator's Vulkan driver returns VK_SUCCESS from vkGetAndroidHardwareBufferPropertiesANDROID with allocationSize == 0 for HWUI's window AHardwareBuffers, and ProcessHardwareBuffer() turned that into a fatal assert(ahb_size), aborting the captured process. This change treats a zero allocation size the same as a failed properties query — log an error and omit that buffer's data from the capture — so capture continues instead of killing the app. It also protects release builds, which compile the assert out and would have passed size 0 into CommonProcessHardwareBuffer().

### DIFF
--- a/framework/encode/vulkan_capture_manager.cpp
+++ b/framework/encode/vulkan_capture_manager.cpp
@@ -2165,10 +2165,8 @@ void VulkanCaptureManager::ProcessHardwareBuffer(format::ThreadId thread_id,
     }
     else
     {
-        GFXRECON_LOG_ERROR("GetAndroidHardwareBufferPropertiesANDROID returned %s with allocationSize %" PRIu64
-                           ": hardware buffer data will be omitted from the capture file",
-                           util::ToString(vk_result).c_str(),
-                           static_cast<uint64_t>(properties.allocationSize));
+        GFXRECON_LOG_ERROR("GetAndroidHardwareBufferPropertiesANDROID failed: hardware buffer data will be omitted "
+                           "from the capture file");
     }
 #else
     GFXRECON_UNREFERENCED_PARAMETER(thread_id);
--- a/framework/encode/vulkan_capture_manager.cpp
+++ b/framework/encode/vulkan_capture_manager.cpp
@@ -2158,6 +2158,9 @@ void VulkanCaptureManager::ProcessHardwareBuffer(format::ThreadId thread_id,
     VkResult vk_result =
         device_table->GetAndroidHardwareBufferPropertiesANDROID(device_unwrapped, hardware_buffer, &properties);
 
+    // There was originally an assert on properties.allocationSize inside the if(). It was causing a crash
+    // when trying to trace on the Android emulator in some circumstances. The failure is a nomral operating error
+    // not a cause for a crash in debug mode.
     if (vk_result == VK_SUCCESS && properties.allocationSize > 0)
     {
         const size_t ahb_size = properties.allocationSize;
--- a/framework/encode/vulkan_capture_manager.cpp
+++ b/framework/encode/vulkan_capture_manager.cpp
@@ -2158,17 +2158,17 @@ void VulkanCaptureManager::ProcessHardwareBuffer(format::ThreadId thread_id,
     VkResult vk_result =
         device_table->GetAndroidHardwareBufferPropertiesANDROID(device_unwrapped, hardware_buffer, &properties);
 
-    if (vk_result == VK_SUCCESS)
+    if (vk_result == VK_SUCCESS && properties.allocationSize > 0)
     {
         const size_t ahb_size = properties.allocationSize;
-        assert(ahb_size);
-
         CommonProcessHardwareBuffer(thread_id, device_wrapper, memory_id, hardware_buffer, ahb_size, this, nullptr);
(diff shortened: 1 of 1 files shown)


framework/encode/vulkan_capture_manager.cpp
```diff
@@ -2158,17 +2158,17 @@ void VulkanCaptureManager::ProcessHardwareBuffer(format::ThreadId thread_id,
     VkResult vk_result =
         device_table->GetAndroidHardwareBufferPropertiesANDROID(device_unwrapped, hardware_buffer, &properties);
 
-    if (vk_result == VK_SUCCESS)
+    if (vk_result == VK_SUCCESS && properties.allocationSize > 0)
```
spencer-lunarg, line 2164: image

spencer-lunarg: likely worth leaving "some" comment what `allocationSize == 0` is for the next person (or LLM) reading this code

bradgrantham-lunarg: ~~That sounds to me like we need to plumb through the possibility that allocationSize is 0 and still write this record.  So the real issue is the assert?~~

edit: Yes, the memory allocated should be 0 size, but I don't know what it means for GetAndroidHardwareBufferPropertiesANDROID to **return** 0 for the AHB size

richard-lunarg: Without the assert, everything behaves fine, and I'm able to get and replay captures no problem.

bradgrantham-lunarg: @spencer-lunarg I think I don't understand your pasted image.  This is a case where `GetAndroidHardwareBufferPropertiesANDROID` **returned** 0 `allocationSize` rather than passing 0 `allocationSize` to vkAllocateMemory.  What does it mean if the AHB memory size was reported to be 0?

bradgrantham-lunarg:
Here's my concern - if it is legal/valid to use an AHB as the backing for a VkImage and GetAndroidHardwareBufferPropertiesANDROID returned 0 for the `allocationSize` but VK_SUCCESS, what does that mean?  It sounds suspiciously like a valid sequence of commands that we should be writing to the file.

I think there's a possibility @richard-lunarg that skipping `CommonProcessHardwareBuffer` in your patch means that the capture isn't complete.

framework/encode/vulkan_capture_manager.cpp
spencer-lunarg: ... ok, sorry, I looked more, the `allocationSize` here [in the pasted image] is **NOT** `VkAndroidHardwareBufferPropertiesANDROID::allocationSize` , it is `VkMemoryAllocateInfo::allocationSize`

(this is super misleading, will fix in spec)

... so looking, I think this is a driver bug that should be reported... it should have returned a VkError, not zero

so I think what Richard has here is right and guarding against a bad implementation returning zero

spencer-lunarg: https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/8511

richard-lunarg:
@bradgrantham-lunarg 
"means that the capture isn't complete."

It seems to play back okay, but I've no way of knowing if it's dropped a frame at the end, etc. Would this "incompleteness" manifest in some way? I'm at 30+ of what appear to be good captures with this change locally that I could not get before.

kocdemir:
This is happening because the goldfish gralloc backend used in the emulator is setting the allocation size to zero for objects which do not have any CPU usage flags. The total allocation size available internally through separate mapper calls, but not exposed through `vkGetAndroidHardwareBufferPropertiesANDROID` right now. This is apparently CTS conformant and changing this behavior would require updating the older system images. 

In new system images however, we're actually transitioning away from this goldfish gralloc to new minigbm based gralloc with virtio-gpu, which directly asks virtgpu device for the allocation size and works regardless of the CPU usage flags. 

In current host emulator binaries, you can enable minigbm backend feature using `-feature Minigbm` command line argument and that should fix the invalid `allocationSize` values from the vulkan call. Note that the minigbm flag is only supported in API 37+ system images, and only expected to be used (i.e tested) together with `-feature GuestAngle` flag.

spencer-lunarg:
@kocdemir thanks for the clarification, what I want is inside https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/8511 to have this information documented in the Vulkan spec

kocdemir:
Right, just pointed out a workaround for the emulator, but I agree that the spec (or CTS) needs some work. On the other hand, when I look into `CommonProcessHardwareBuffer`, looks like the input size is only used for CPU readable buffers, so maybe passing the provided zero in there could actually work fine, ie. `properties.allocationSize > 0` is not needed?

richard-lunarg:
This has been superceeded by another PR that addresses this issue in a different way.

PR #3234, opened 2026-08-28
[Frame Looping] Buffer content restoration

wzchu-lunarg:
**Problem**
As described in Oleh's [notes](https://github.com/olehkuznetsov/gfxreconstruct/blob/d64f4ab205604d2f55ed964a7521b7e62fb05425/frame_loop_buffers_content.md), buffer contents can mutate during loop iterations. These mutations should be reverted otherwise things can be mismatched and seem incorrect.

**Main Changes**
Most of the intentions of the changes follow the design in Oleh's [notes](https://github.com/olehkuznetsov/gfxreconstruct/blob/d64f4ab205604d2f55ed964a7521b7e62fb05425/frame_loop_buffers_content.md) and the idea of using shadow buffers to record what was in the initial buffers, then restoring them at the end of the loop. However, changes are mainly followed conceptually rather than pulled in directly.

**vulkan_replay_frame_loop_consumer**
- Buffers need to be copyable and be copied to by modifying it with `VK_BUFFER_USAGE_TRANSFER_SRC_BIT` and `VK_BUFFER_USAGE_TRANSFER_DST_BIT`
- Create `BufferTracking` class to manage and utilize the shadow buffers that restore the contents
- Recording initial state doesn't limit shadow buffers to 32 mb as in Oleh's changes, but shows a warning if that allocation fails
- Create `RecordBufferStates` at loop start to record buffer contents per device
- Create `FixupDeviceBuffers` to restore contents at end of loop

**vulkan_util**
- Create function `FindGraphicsOrComputeQueueFamilyIndex` to get queue for recording initial state. This might not be perfect, as this might end up with a different queue than what was used for buffer creation, which could be a problem when using `VK_SHARING_MODE_EXCLUSIVE`.  Restoration uses the queue from submission.


framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -318,10 +321,36 @@ void VulkanReplayFrameLoopConsumer::ProcessStateEndMarker(uint64_t frame_number)
         per_device_fence_tracking_.clear();
         per_device_event_tracking_.clear();
         per_device_semaphore_tracking_.clear();
+        for (auto& [device_id, buffer_tracking] : per_device_buffer_tracking_)
+        {
+            buffer_tracking.DestroyShadowBuffers();
+        }
+        per_device_buffer_tracking_.clear();
```
antonio-lunarg: Maybe we can extract these few lines above into a `Reset()` function and call that from the destructor as well.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -403,6 +434,247 @@ void VulkanReplayFrameLoopConsumer::TrackSemaphoreStates()
...
+    std::unordered_map<format::HandleId, std::vector<format::HandleId>> device_buffers;
+    table.VisitVkBufferInfo([&device_buffers](const VulkanBufferInfo* buffer_info) {
+        if (buffer_info == nullptr || buffer_info->handle == VK_NULL_HANDLE || buffer_info->size == 0)
+        {
+            return;
+        }
+
+        // A buffer that was created but never bound to memory (vkBindBufferMemory never called, or
+        // never succeeded) has no backing memory at all.
+        if (buffer_info->memory_property_flags == 0)
```
antonio-lunarg, line 497: The comment associated to the `memory_property_flags` in `vulkan_object_info.h:454` is stale.
I would change it with `// non-zero implies a successful vkBindBufferMemory`, which would make this code here reasonable.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -403,6 +434,247 @@ void VulkanReplayFrameLoopConsumer::TrackSemaphoreStates()
...
+            continue;
+        }
+
+        const VulkanBufferInfo* buffer_info = object_table_.GetVkBufferInfo(buffer_id);
+        if (buffer_info == nullptr || buffer_info->handle == VK_NULL_HANDLE || buffer_info->size == 0)
+        {
+            continue;
+        }
+
+        VkBufferCreateInfo create_info = { VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO };
```
antonio-lunarg, line 556: Ding 4x hit! Time to refactor into something like `TemporaryBuffer` in `vulkan_temporary_objects.h`.
See:
- screenshot_handler.cpp:655
- vulkan_resource_initializer.cpp:1042
- vulkan_address_replacer.cpp:2031

antonio-lunarg: Non-blocking though. This can be done in a follow-up PR.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -403,6 +434,247 @@ void VulkanReplayFrameLoopConsumer::TrackSemaphoreStates()
...
+                                 buffer_id);
+            allocator_->DestroyBufferDirect(shadow.buffer, nullptr, shadow.alloc_data);
+            continue;
+        }
+
+        VkMemoryAllocateInfo alloc_info = { VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO };
+        alloc_info.allocationSize       = mem_reqs.size;
+        alloc_info.memoryTypeIndex      = memory_type_index;
+
+        result = allocator_->AllocateMemoryDirect(&alloc_info, nullptr, &shadow.memory, &shadow.mem_data);
```
antonio-lunarg, line 597: Ouch. Each shadow buffer gets its own allocation. This might fail if the number of allocation reaches `maxMemoryAllocationCount`.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -318,10 +321,36 @@ void VulkanReplayFrameLoopConsumer::ProcessStateEndMarker(uint64_t frame_number)
...
+}
+
+void VulkanReplayFrameLoopConsumer::Process_vkCreateBuffer(const ApiCallInfo& call_info, args::CreateBuffer& args)
+{
+    VkBufferCreateInfo* create_info = args.pCreateInfo.GetPointer();
+
+    // Ensure that buffers can be copied to and from so contents can be restored.
+    if (create_info != nullptr)
+    {
+        create_info->usage |= VK_BUFFER_USAGE_TRANSFER_SRC_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT;
```
antonio-lunarg, line 369: Mmh, I am not sure about editing `create_info` is the right call here. `OverrideCreateBuffer` works on a clone (`modified_create_info`).

framework/decode/vulkan_replay_frame_loop_consumer.cpp
antonio-lunarg: Also, this should happen at `DestroyDevice` time as well.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -403,6 +434,247 @@ void VulkanReplayFrameLoopConsumer::TrackSemaphoreStates()
...
+    if (shadow_buffers_.empty())
+    {
+        return;
+    }
+
+    VulkanDeviceInfo* device_info = object_table_.GetVkDeviceInfo(device_id_);
+    GFXRECON_ASSERT(device_info != nullptr);
+
+    TemporaryCommandBuffer temp_cmd_buff(*device_info, device_table_);
+    if (temp_cmd_buff.CreateAndBegin(queue_family_index) != VK_SUCCESS)
```
antonio-lunarg: What is `queue_family_index` here happens to be different thatn the one used in `RecordInitialState`?

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -403,6 +434,247 @@ void VulkanReplayFrameLoopConsumer::TrackSemaphoreStates()
...
+        GetBufferTracking(device_id).RecordInitialState(buffer_ids);
+    }
+}
+
+void VulkanReplayFrameLoopConsumer::FixupDeviceBuffers(format::HandleId device, format::HandleId queue)
+{
+    auto it = per_device_buffer_tracking_.find(device);
+    if (it == per_device_buffer_tracking_.end() || it->second.shadow_buffers_.empty())
+    {
+        return;
```
antonio-lunarg, line 520: Log warning

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: The problem that can occur is if ownership was exclusive to the queue for the original buffer. The main issue seems to be deciding the queue in `RecordInitialState`. Right now it is just making the best guess with  `graphics::FindGraphicsOrComputeQueueFamilyIndex`. The alternative would be tracking the queue the original buffer was created on.

Will probably look into this alternative for a bit first.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: This would just be an early return if there were no buffers to fixup. However, the second early return if queue is `nullptr` is something that shouldn't occur.

Adding a commit to add an error log if the `queue_info` check fails.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Was thinking of using suballocations, but thought it might be complex and worth as a followup later. I'll look into implementing one here then.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Completely missed this.

Adding a commit for this.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: I believe this should modify the usage flags as in `OverrideCreateBuffer`, it takes the already mutated pointer with `replay_create_info   = pCreateInfo->GetPointer();`, and then does the copy.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Seems like this would be better!

Adding a commit to create helpers for this. `DestroyDevice` would however need to do something slightly different, as it only destroys one device rather than clearing all of them.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Perhaps this is better as a followup PR along with a refactor into `TemporaryBuffer`. We can have something like a `TemporaryBufferPool` and do suballocations then.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: For more detail, the main problem is in the case a buffer is created with `VK_SHARING_MODE_EXCLUSIVE`, and in this case we should be using the same queue family the buffer is created with to do the "snapshot" of those buffers. If we try accessing from different queues, the contents could be undefined. However, I don't believe we track this sort of "current owning family". Although in `vulkan_object_info.h` there is `VulkanBufferInfo -> queue_family_index`, this is always 0 for `VK_SHARING_MODE_EXCLUSIVE` from creation time and is never updated. 

From what I am seeing, the queue that an exclusive sharing mode buffer belongs to is the one that first uses it. Ownership can then be transferred through some type of barriers, but we don't seem to track the changes from those either. 

GFXR seems to be able to get away with this since most of the time as it either:
- uses queue 0 which most apps use anyways
- guesses it by using the capability flags which also ends up with the queue an app is most likely to use
- reuses the actual queue index from the capture

There are then some cases this could break, although I don't have any specific traces or apps that do that, and they are probably not that common. Since most apps just use the queues with the most capabilities, most of the time guessing with capability flags also results in the same queue. However, it does seem to be doable to track a buffer's "current queue family" if we want better correctness. This would involve getting the queue that "first touched" this buffer by tracking commands that use buffers + submit, and additional ownership transfers with buffer memory barriers.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -318,10 +321,54 @@ void VulkanReplayFrameLoopConsumer::ProcessStateEndMarker(uint64_t frame_number)
...
+    {
+        it->second.DestroyShadowBuffers();
+        per_device_buffer_tracking_.erase(it);
+    }
+}
+
+void VulkanReplayFrameLoopConsumer::Process_vkDestroyDevice(const ApiCallInfo& call_info, args::DestroyDevice& args)
+{
+    VulkanReplayFrameLoopConsumerBase::Process_vkDestroyDevice(call_info, args);
+    ResetBufferTracking(args.device);
```
antonio-lunarg: Those two lines should be inverted:
```cpp
if (!frame_loop_info_.IsLooping() || frame_loop_info_.IsFinalIteration())
{
    ResetBufferTracking(args.device);
}
VulkanReplayFrameLoopConsumerBase::Process_vkDestroyDevice(call_info, args);
```

framework/decode/vulkan_replay_frame_loop_consumer.cpp
antonio-lunarg: Sure

framework/decode/vulkan_replay_frame_loop_consumer.cpp
antonio-lunarg: I've noticed [VkBufferUsageFlags2CreateInfo](https://docs.vulkan.org/refpages/latest/refpages/source/VkBufferUsageFlags2CreateInfo.html), which is also something we might need to take into account.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -403,6 +455,241 @@ void VulkanReplayFrameLoopConsumer::TrackSemaphoreStates()
...
+    for (const auto& [device_id, buffer_ids] : device_buffers)
+    {
+        GetBufferTracking(device_id).RecordInitialState(buffer_ids);
+    }
+}
+
+void VulkanReplayFrameLoopConsumer::FixupDeviceBuffers(format::HandleId device)
+{
+    auto it = per_device_buffer_tracking_.find(device);
+    if (it == per_device_buffer_tracking_.end() || it->second.shadow_buffers_.empty())
```
antonio-lunarg: No need to check `empty` here

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Missed this. Always forget about `pNext` cases.

Adding a commit for this.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Adding a commit for this.

antonio-lunarg (approved):

PR #3235, opened 2026-08-31
Restore synchronization on vkGetQueryPoolResults and vkGetEventStatus, bounded by state tracking

fabian-lunarg:
supersedes #2978, carry over @marius-pelegrin-arm's original commit.

#2978 showed two things:

1) Waiting is necessary for correctness. Replay must reproduce a captured `VK_SUCCESS` / `VK_EVENT_SET` before continuing, otherwise conformant workloads desync 
-> ARM's **tracetooltests**' `vulkan_compute_wait` replays corrupted on current dev, on both the `vkGetEventStatus` and the `vkGetQueryPoolResults` path.

2) But unconditional waiting can hang. A captured `VK_EVENT_SET` can be a transient set/reset race outcome. replay might (legitimately) never reproduce it (see `vulkan_event_hang`, backtrace in the #2978 thread), and trimmed captures can leave queries permanently unavailable (state-setup resets the entire pool and re-activation doesn't cover all query types) — reproduced with a trimmed capture hanging in `vkGetQueryPoolResults` + `VK_QUERY_RESULT_WAIT_BIT` on win11/AMD.

So additional tracking is required to safely bound the synchronization loops. 
This PR bounds them by what the replayed stream actually guarantees, **avoiding both timeouts and retry caps**:

events: 

- track each event's **terminal state** (host `vkSetEvent`/`vkResetEvent` immediately, `vkCmdSetEvent`/`vkCmdResetEvent(2/2KHR)` at submit-time, secondaries merged at `vkCmdExecuteCommands`).
- The `vkGetEventStatus` poll only runs when the tracked terminal state is `VK_EVENT_SET` 
-> the setter is then guaranteed to be replayed and the loop terminates.

query pools: 
- track per-query availability (`vkCmdEndQuery(IndexedEXT)`, `vkCmdWriteTimestamp(2/2KHR)`, `vkCmdWriteAccelerationStructuresPropertiesKHR` vs. host/device resets). `VK_QUERY_RESULT_WAIT_BIT` is injected only when the whole requested range is tracked-available. 
- an app-supplied `WAIT_BIT` the stream cannot satisfy is stripped.
- when a wait is declined, the call returns the replay-truthful value and the capture/replay mismatch is logged (`SET` -> `RESET` and `SUCCESS` -> `NOT_READY` are deliberately not allow-listed)
-> this is where synchronization could not be reproduced.
- The benign "replay is ahead" case `RESET` -> `SET` is allow-listed.
- **frame-loop replay keeps the tracking** in sync with its injected query-pool resets and event fixups.

Verified: `vulkan_compute_wait` replays correctly on both paths (corrupts on dev), `vulkan_event_hang` completes (hangs with an unbounded loop), and the trimmed capture that hung on win11/AMD replays through.


framework/decode/vulkan_object_info.h
```diff
@@ -259,12 +259,24 @@ struct VulkanPoolObjectInfo : public VulkanObjectInfo<T>
     format::HandleId pool_id{ format::kNullHandleId }; // ID of the pool that the object was allocated from.
 };
 
+struct VulkanEventInfo : public VulkanObjectInfo<VkEvent>
+{
+    // Terminal event-state in stream order: host set/reset apply immediately, device cmd set/reset at
+    // submit-time. Bounds the vkGetEventStatus poll to only wait when VK_EVENT_SET is the terminal state.
+    VkResult latched_state{ VK_EVENT_RESET };
```
antonio-lunarg: This only ever holds SET or RESET, right? Maybe a `bool` would be enough and safer.

antonio-lunarg (approved):

framework/decode/vulkan_object_info.h
fabian-lunarg: agreed, only ever SET or RESET here

fabian-lunarg: done, simpler/cleaner in couple places, thanks!

PR #3236, opened 2026-08-31
Fix DxgiAdapterDesc layout mismatch causing gfxrecon-info to fail on Linux

tpower-amd:
## Problem

`gfxrecon-info` fails to correctly read D3D12 trace metadata when run on Linux. `DxgiAdapterDesc::Description` is declared as `wchar_t[kAdapterDescriptionSize]`, but `wchar_t` is 4 bytes on Linux and 2 bytes on Windows. When reading a trace captured on Windows, the Description field occupies twice the expected bytes on disk, corrupting the layout of all subsequent fields in the struct (`VendorId`, `DeviceId`, memory sizes, LUID, etc.).

## Solution

~~Guard the field declaration with a platform `#ifdef`: use `char16_t` on Linux and retain `wchar_t` on Windows. Change is in `framework/format/format.h` only — no changes to serialization or consumer code are required as the on-disk representation is unchanged.~~ GFXR needs to ensure this struct is always serialized with the exact number of bytes, so use `char16_t` to ensure cross-platform consistency. Additionally, add new helper functions to support the type change:

- `util::strings::convert_u16string_to_utf8`: a cross-platform UTF-16 to UTF-8 converter analogous to the existing `convert_wstring_to_utf8`, used when serializing `DxgiAdapterDesc` to JSON.
- `util::FieldToJson(u16string_view)`: a `FieldToJson` overload for `char16_t` string data.
- `util::Char16ArrayToString`: a Windows-only helper analogous to `WCharArrayToString`, used at call sites where `DxgiAdapterDesc::Description` is converted to a display string.

## Testing

- ✅ Test suites all passed: `gfxrecon_util_test`, `gfxrecon_encode_test`, `gfxrecon_decode_test`, `gfxrecon_format_test`, `gfxrecon_framework_test`, `gfxrecon_graphics_test`, `gfxrecon_replay_event_plugin_loader_test`.

## Notes

PR using incorrect branch in fork was previously created here: https://github.com/LunarG/gfxreconstruct/pull/3189
Made [requested changes](https://github.com/LunarG/gfxreconstruct/pull/3189#discussion_r3816077752) and recreated this PR using correct feature branch.


framework/util/to_string.h
```diff
@@ -157,6 +157,27 @@ inline std::string HandleIdToString(format::HandleId handleId)
 }
 
 #if defined(_WIN32)
+inline std::string Char16ArrayToString(const char16_t* pStr)
+{
+    // wchar_t and char16_t are both 2-byte UTF-16 on Windows.
```
davidd-lunarg: A `static_assert(sizeof(wchar_t) == sizeof(char16_t));` would help guard against this code being compiled on an unsupported platform.

framework/util/to_string.h
```diff
@@ -157,6 +157,27 @@ inline std::string HandleIdToString(format::HandleId handleId)
...
+    }
+    if (required_size > 0)
+    {
+        std::string result_str;
+        result_str.resize(required_size);
+        WideCharToMultiByte(
+            CP_UTF8, 0, wPtr, -1, result_str.data(), static_cast<int>(result_str.size()), nullptr, nullptr);
+        return result_str;
+    }
+    return "";
```
davidd-lunarg: This code section looks like it duplicates the logic of `WCharArrayToString` so could it just call that with `wPtr`?

tpower-amd: Fair point, trying to preemptively avoid coupling the two won't prevent us from refactoring in the future, so will refactor to passing through to `WCharArrayToString`.

davidd-lunarg (approved):

PR #3239, opened 2026-09-01
VulkanStateWriter annotate omitted creation calls

panos-lunarg:
Image views, buffer views and frame buffers that are detected as orphaned have their creation call omitted from the capture file. VulkanStateWriter will now put an annotation in the place of the culled call

Additionally optimizer will now include the size of the omitted kInitBufferCommand, kInitImageCommand, and kInitTensorCommand

### DIFF
--- a/framework/format/format_util.cpp
+++ b/framework/format/format_util.cpp
@@ -66,6 +66,18 @@ bool ValidateFileHeader(const FileHeader& header)
     return valid;
 }
 
+AnnotationHeader MakeAnnotationHeader(AnnotationType type, size_t label_length, size_t data_length)
+{
+    AnnotationHeader annotation{};
+    annotation.block_header.size = GetAnnotationBlockBaseSize() + label_length + data_length;
+    annotation.block_header.type = BlockType::kAnnotation;
+    annotation.annotation_type   = type;
+    annotation.label_length      = GFXRECON_NARROWING_CAST(uint32_t, label_length);
+    annotation.data_length       = data_length;
+
+    return annotation;
+}
+
 util::Compressor* CreateCompressor(CompressionType type)
 {
     util::Compressor* compressor = nullptr;
--- a/framework/format/format_util.cpp
+++ b/framework/format/format_util.cpp
@@ -72,8 +72,8 @@ AnnotationHeader MakeAnnotationHeader(AnnotationType type, size_t label_length, 
     annotation.block_header.size = GetAnnotationBlockBaseSize() + label_length + data_length;
     annotation.block_header.type = BlockType::kAnnotation;
     annotation.annotation_type   = type;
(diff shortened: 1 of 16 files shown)


framework/format/format_util.cpp
```diff
@@ -66,6 +66,18 @@ bool ValidateFileHeader(const FileHeader& header)
     return valid;
 }
 
+AnnotationHeader MakeAnnotationHeader(AnnotationType type, size_t label_length, size_t data_length)
+{
+    AnnotationHeader annotation{};
+    annotation.block_header.size = GetAnnotationBlockBaseSize() + label_length + data_length;
+    annotation.block_header.type = BlockType::kAnnotation;
+    annotation.annotation_type   = type;
+    annotation.label_length      = GFXRECON_NARROWING_CAST(uint32_t, label_length);
```
jzulauf-lunarg, line 75: Prefer:

`GFXRECON_NARROWING_ASSIGN(annotation.label_length, label_length);`

as the destination length is derived from the assignment target, and not prone to copy paste error.

jzulauf-lunarg (changes requested):
Use the safer narrowing assign.  Otherwise LGTM.

csteinberg-samsung:
I'd like if the args parameters could be retained. This doesn't give me much to work from. I get that you want to skip the allocation during write state restoration, but I don't think writing the args would be that expensive. None of these object classes actually store a data blob/binary, so the args are mostly just integers. The most important here is probably the pView handle, where otherwise we have nothing to reference this removed buffer view by. The allocator is probably not relevant in this case. 

```
{
  "index": 91,
  "annotation": {
    "type": "kText",
    "label": "removed-resource",
    "data": "Removed vkCreateBufferView for orphaned buffer view 100"
  }
},
```

as opposed to
```
{
  "index": 91,
  "annotation": {
    "type": "kText",
    "label": "removed-resource",
    "data": "Removed vkCreateBufferView for orphaned buffer view 100",
    "args": {
      "device": 14,
      "pCreateInfo": {
        "sType": "VK_STRUCTURE_TYPE_BUFFER_VIEW_CREATE_INFO",
        "flags": "0x00000000",
        "buffer": 100,
        "format": "VK_FORMAT_R32G32B32A32_SFLOAT",
        "offset": 0,
        "range": 18446744073709551615,
        "pNext": null
      },
      "pAllocator": null,
      "pView": 115
    }
  }
},

panos-lunarg:
Done. I don't store the whole create info struct to minimize the memory footprint a bit.

framework/format/format_util.cpp
panos-lunarg: Done

csteinberg-samsung (approved):
Meets my needs, thanks!

bradgrantham-lunarg (approved):

PR #3242, opened 2026-09-02
Add unit tests for vulkan rebind allocator

bartosz-muszarski-arm:
* add an injectable VMA backend to VulkanRebindAllocator so allocator policy can be unit tested without calling into VMA directly
* route image allocation and data-graph session allocation paths through the backend while keeping the default production behavior unchanged
* add a dedicated gfxrecon_rebind_allocator_test target and test build plumbing for GoogleTest/gMock
* add unit coverage for
  * image memory usage selection,
  * memory-usage fallback,
  * image allocation request shaping,
  * cached allocation reuse,
  * data-graph session binding synthesis

Change-Id: I81072fd6f68dd0f210654cfdef7f026fac1b140c


framework/decode/test/vulkan_rebind_allocator_test.cpp
```diff
@@ -0,0 +1,909 @@
...
+** AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+** FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+** DEALINGS IN THE SOFTWARE.
+*/
+
+#define CATCH_CONFIG_MAIN
+#include <catch2/catch.hpp>
+
+#include <gmock/gmock.h>
```
fabian-lunarg, line 27: this requires some attention here.
it's a **Catch2 test** (`#define CATCH_CONFIG_MAIN`) that uses gmock only for mocks.
it looks there is no `InitGoogleMock`, no gtest main. 
meaning gtest's failure reporter is never installed and we should be missing reported errors.
let me confirm ...

bartosz-muszarski-arm: yes only catch2 checks are reported here.
gmock does mocking only.
I can extend error reporting to gmock as well.
Do you prefer catch2 based tests+custom gmock installer or just go straight for gtest+gmock?

framework/decode/vulkan_rebind_allocator.cpp
```diff
@@ -3987,11 +4054,12 @@ VkResult VulkanRebindAllocator::InitializeDataGraphPipelineSessionMemory(VkDataG
             memory_info.alc_create_info                    = allocation_create_info;
             memory_info.offset_from_original_device_memory = 0;
 
-            result = vmaAllocateMemory(allocator_,
-                                       &memory_requirements.memoryRequirements,
-                                       &allocation_create_info,
-                                       &memory_info.allocation,
-                                       &memory_info.allocation_info);
+            util::MarkInjectedCommandsHelper injected;
```
fabian-lunarg, line 4057: not required, in fact shadowing `auto injected = device_table_.Open();` from above

framework/decode/test/vulkan_rebind_allocator_test.cpp
fabian-lunarg: > catch2 based tests+custom gmock installer or just go straight for gtest+gmock

it's a standalone test-executable, it 'could' just become a **gtest** entirely to enable reporting from gmock.
easier+safer than having two test-frameworks in one test (using some 'bridge', which sounds messy).

-> "go straight for gtest+gmock" in my opinion

MarkY-LunarG:
FYI: I'm doing some cleanup and improvements to testing.  The first changes are in PR #3270.

PR #3245, opened 2026-09-03
vulkan: Fix crash during async pipeline compilation and handle destruction

olehkuznetsov:
When replaying traces with parallel pipeline creation enabled (`--pipeline-creation-jobs` / `--pcj`), background worker threads compile pipelines asynchronously. Several edge cases in async handle tracking, handle destruction, and pipeline cache lifecycle caused driver crashes and race conditions.

### Key issues and fixes:

1. **Reference-counted handle tracking across concurrent tasks:**
   Multiple async pipeline creation tasks frequently share common dependencies (`VkPipelineLayout`, `VkRenderPass`, `VkShaderModule`). Previously, dependencies were tracked globally without task reference counting. When the first task finished, it cleared the handle, allowing premature destruction and use-after-free in subsequent tasks still compiling on background threads (e.g. `SIGSEGV` in `SPIR_Parser::parse`).
   - **Fix**: Track dependencies by async task ID sets and deferred callbacks in `struct AsyncTrackedHandle` (`async_tracked_handles_`) and per-task sync callbacks in `async_task_sync_fns_`. A handle is only cleared when all referencing tasks have finished. In `DestroyAsyncHandle`, all referencing tasks are synchronized before destroying the handle.

2. **Handle destruction lifecycle and consolidation:**
   Consolidate handle destruction across `VkPipeline`, `VkPipelineLayout`, `VkRenderPass`, and `VkShaderModule` via `DestroyTrackedHandle`. Defer destruction when handles are tracked by active async tasks via `DestroyAsyncHandle`.

3. **Intercept `VkPipelineLayout` destruction:**
   During async pipeline compilation, the main thread previously destroyed `VkPipelineLayout` handles immediately via `vkDestroyPipelineLayout` because GFXReconstruct lacked an override, crashing background compilation threads accessing the layout.
   - **Fix**: Add `OverrideDestroyPipelineLayout` to `replay_overrides.json` and defer layout destruction via `DestroyTrackedHandle` until referencing async tasks complete.

4. **Shared pipeline cache lifecycle and dispatch ordering:**
   In batched creations or shared caches, multiple pipelines reference the same pipeline cache ID. `OverrideDestroyPipeline` previously destroyed the cache on the first pipeline destruction, leading to assertion failures (`tracked_pipeline_caches_`) or missed cache saves. Additionally, capturing `out_pipelines.data()` by raw pointer into the `MainThreadQueue` completion callback caused dangling pointers because the result vector in the future's shared state is freed once all handles in the batch have synced their futures (`object->future = {}`).
   - **Fix**: Extract `DestroyAssociatedPipelineCache` and track remaining correspondences (`sameIdFound`) so `vkDestroyPipelineCache` and `SavePipelineCache` execute only after the last pipeline is destroyed. Capture `pipeline_handles` by value into completion closures. In `DestroyAsyncHandle`, drain `main_thread_queue_.poll()` after worker synchronization so cache registrations dispatch before destruction, and assert all referencing tasks have erased their IDs.

5. **Exception handling and future synchronization invariants:**
   In `sync_handle`, unhandled worker thread exceptions could escape during handle synchronization.
   - **Fix**: Enforce the `object->future_handle_index < async_handles.size()` invariant via `GFXRECON_ASSERT`, catch `const std::exception&` with `GFXRECON_LOG_FATAL` to fail fast with diagnostic information, and clear `object->future = {}` upon resolution to release the future's shared state and prevent redundant synchronization.

### DIFF
--- a/framework/decode/vulkan_object_info_table_base.h
+++ b/framework/decode/vulkan_object_info_table_base.h
@@ -27,6 +27,7 @@
 #include "decode/vulkan_object_info.h"
 #include "format/format.h"
 #include "util/defines.h"
+#include "util/logging.h"
 
 #include "vulkan/vulkan.h"
 
@@ -62,8 +63,21 @@ static inline void sync_handle(T* object)
     static_assert(has_handle_future_v<T>);
     if (object != nullptr && object->handle == VK_NULL_HANDLE && object->future.valid())
     {
-        const auto& [result, async_handles] = object->future.get();
-        object->handle                      = async_handles[object->future_handle_index];
+        try
+        {
+            const auto& [result, async_handles] = object->future.get();
+            GFXRECON_ASSERT(object->future_handle_index < async_handles.size());
+            object->handle = async_handles[object->future_handle_index];
+        }
+        catch (const std::exception& e)
+        {
+            // Catch std::exception to prevent worker-thread exceptions from escaping silently
+            // or causing undefined behavior during subsequent replay.
+            GFXRECON_LOG_FATAL("Future error while synchronizing handle: %s", e.what());
+        }
+
+        // Release shared state and prevent redundant future.get() synchronization on subsequent lookups.
+        object->future = {};
     }
 }
 
--- a/framework/decode/vulkan_replay_consumer_base.cpp
+++ b/framework/decode/vulkan_replay_consumer_base.cpp
@@ -8555,13 +8555,12 @@ VkResult VulkanReplayConsumerBase::OverrideSetDebugUtilsObjectNameEXT(
         size_t               info_size   = graphics::vulkan_struct_deep_copy(info, 1, nullptr);
         std::vector<uint8_t> info_copy(info_size);
         graphics::vulkan_struct_deep_copy(info, 1, info_copy.data());
-        auto& async_handle_asset         = async_tracked_handles_[pipeline_id];
-        async_handle_asset.post_build_fn = [this,
-                                            device = device_info->handle,
-                                            pipeline_id,
-                                            func,
-                                            info_copy = std::move(info_copy),
-                                            original_result]() mutable {
+        async_tracked_handles_[pipeline_id].post_build_fn = [this,
+                                                             device = device_info->handle,
+                                                             pipeline_id,
+                                                             func,
+                                                             info_copy = std::move(info_copy),
+                                                             original_result]() mutable {
             auto* info = reinterpret_cast<VkDebugUtilsObjectNameInfoEXT*>(info_copy.data());
 
             // correct referenced handle in info. this would sync, but task is already done
@@ -13596,70 +13595,69 @@ VkResult VulkanReplayConsumerBase::OverrideCreateShadersEXT(
     return replay_result;
 }
 
+void VulkanReplayConsumerBase::DestroyAssociatedPipelineCache(const VulkanDeviceInfo* device_info, VkPipeline pipeline)
+{
+    if (pipeline == VK_NULL_HANDLE)
+    {
+        return;
+    }
+
+    auto itCorresp = pipeline_cache_correspondances_.find(pipeline);
+    if (itCorresp != pipeline_cache_correspondances_.end())
+    {
+        format::HandleId id = itCorresp->second;
+        pipeline_cache_correspondances_.erase(itCorresp);
+
+        // For batched pipeline creations or shared pipeline caches, multiple pipelines
+        // reference the same pipeline cache ID. Do not destroy the driver cache until the last pipeline is destroyed.
+        bool sameIdFound = false;
+        for (const auto& elt : pipeline_cache_correspondances_)
+        {
+            if (elt.second == id)
+            {
+                sameIdFound = true;
+                break;
+            }
+        }
+
+        // If this is the only remaining pipeline bound to the pipeline cache, save and destroy the pipeline cache
+        if (!sameIdFound)
+        {
+            auto itTracked = tracked_pipeline_caches_.find(id);
+            GFXRECON_ASSERT(itTracked != tracked_pipeline_caches_.end());
+            GFXRECON_ASSERT(itTracked->second.device_info != nullptr);
+            GFXRECON_ASSERT(itTracked->second.vk_cache != VK_NULL_HANDLE);
+
+            if (save_pipeline_caches_to_file)
+            {
+                SavePipelineCache(id, itTracked->second);
+            }
+            auto device_table = GetInjectedDeviceCalls(device_info->handle);
+            auto injected     = device_table.Open();
+            injected->DestroyPipelineCache(itTracked->second.device_info->handle, itTracked->second.vk_cache, nullptr);
+            tracked_pipeline_caches_.erase(itTracked);
+        }
+    }
+}
+
 void VulkanReplayConsumerBase::OverrideDestroyPipeline(
     PFN_vkDestroyPipeline                                      func,
     const VulkanDeviceInfo*                                    device_info,
     const VulkanPipelineInfo*                                  pipeline_info,
     const StructPointerDecoder<Decoded_VkAllocationCallbacks>* pAllocator)
 {
-    GFXRECON_ASSERT(device_info != nullptr);
-    VkDevice                     in_device     = device_info->handle;
-    VkPipeline                   in_pipeline   = VK_NULL_HANDLE;
-    const VkAllocationCallbacks* in_pAllocator = GetAllocationCallbacks(pAllocator);
+    DestroyTrackedHandle(func, device_info, pipeline_info, pAllocator, [this, device_info](VkPipeline p) {
+        DestroyAssociatedPipelineCache(device_info, p);
+    });
+}
 
-    if (pipeline_info != nullptr)
-    {
-        in_pipeline =
-            MapHandle<VulkanPipelineInfo>(pipeline_info->capture_id, &VulkanObjectInfoTable::GetVkPipelineInfo);
-
-        if (IsUsedByAsyncTask(pipeline_info->capture_id))
-        {
-            // schedule deletion
-            DestroyAsyncHandle(pipeline_info->capture_id, [func, in_device, in_pipeline, in_pAllocator]() {
-                func(in_device, in_pipeline, in_pAllocator);
-            });
-            return;
-        }
-
-        // Check if the pipeline has been created with a specially created pipeline cache
-        auto itCorresp = pipeline_cache_correspondances_.find(pipeline_info->handle);
-        if (itCorresp != pipeline_cache_correspondances_.end())
-        {
-            format::HandleId id = itCorresp->second;
-            pipeline_cache_correspondances_.erase(itCorresp);
-
-            // Find if other pipelines have been created with the same pipeline cache
-            bool sameIdFound = false;
-            for (const auto& elt : pipeline_cache_correspondances_)
-            {
-                if (elt.second == id)
-                {
-                    sameIdFound = true;
-                    break;
-                }
-            }
-
-            // If this is the only remaining pipeline bound to the pipeline cache, save and destroy the pipeline cache
-            if (!sameIdFound)
-            {
-                auto itTracked = tracked_pipeline_caches_.find(id);
-                GFXRECON_ASSERT(itTracked != tracked_pipeline_caches_.end());
-                GFXRECON_ASSERT(itTracked->second.device_info != nullptr);
-                GFXRECON_ASSERT(itTracked->second.vk_cache != VK_NULL_HANDLE);
-
-                if (save_pipeline_caches_to_file)
-                {
-                    SavePipelineCache(id, itTracked->second);
-                }
-                auto device_table = GetInjectedDeviceCalls(device_info->handle);
-                auto injected     = device_table.Open();
-                injected->DestroyPipelineCache(
-                    itTracked->second.device_info->handle, itTracked->second.vk_cache, nullptr);
-                tracked_pipeline_caches_.erase(itTracked);
-            }
-        }
-    }
-    func(in_device, in_pipeline, in_pAllocator);
+void VulkanReplayConsumerBase::OverrideDestroyPipelineLayout(
+    PFN_vkDestroyPipelineLayout                                func,
+    const VulkanDeviceInfo*                                    device_info,
+    VulkanPipelineLayoutInfo*                                  pipeline_layout_info,
+    const StructPointerDecoder<Decoded_VkAllocationCallbacks>* pAllocator)
+{
+    DestroyTrackedHandle(func, device_info, pipeline_layout_info, pAllocator);
 }
 
 void VulkanReplayConsumerBase::OverrideDestroyRenderPass(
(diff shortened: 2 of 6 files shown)


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -13564,18 +13564,19 @@ void VulkanReplayConsumerBase::OverrideDestroyPipeline(
...
-
         if (IsUsedByAsyncTask(pipeline_info->capture_id))
         {
-            // schedule deletion
-            DestroyAsyncHandle(pipeline_info->capture_id, [func, in_device, in_pipeline, in_pAllocator]() {
-                func(in_device, in_pipeline, in_pAllocator);
+            auto capture_id = pipeline_info->capture_id;
+            DestroyAsyncHandle(capture_id, [this, func, in_device, capture_id, in_pAllocator]() {
+                VkPipeline resolved_pipeline =
+                    MapHandle<VulkanPipelineInfo>(capture_id, &VulkanObjectInfoTable::GetVkPipelineInfo);
```
antonio-lunarg: No need to call `MapHandle` from this callback.

Going up the stack, `OverrideDestroyPipeline` is called by `Process_vkDestroyPipeline`, which does this:
```cpp
GetVkPipelineInfo(P)
OverrideDestroyPipeline(...)
RemoveHandle(P)
```

`GetVkPipelineInfo(P)` already blocks un case of a sync handle, therefore `pipeline_info->handle` is already correct within `OverrideDestroyPipeline`.

Then, a possible pitfall, in case we enable `async_defer_deletion_`, the callback register by `DestroyAsyncHandle` would be deferred, would be called at a point where `RemoveHandle(P)` was already called, and `MapHandle(P)` would not find the pipeline resulting in a `Failed to map handle`.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -13617,6 +13618,34 @@ void VulkanReplayConsumerBase::OverrideDestroyPipeline(
...
+{
+    VkDevice                     in_device          = device_info->handle;
+    VkPipelineLayout             in_pipeline_layout = VK_NULL_HANDLE;
+    const VkAllocationCallbacks* in_pAllocator      = GetAllocationCallbacks(pAllocator);
+
+    if (pipeline_layout_info != nullptr)
+    {
+        in_pipeline_layout = pipeline_layout_info->handle;
+
+        if (IsUsedByAsyncTask(pipeline_layout_info->capture_id))
```
antonio-lunarg: What if the same layout is used to create two different pipelines and one pipeline finishes while the other is still compiling? @fabian-lunarg

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -13633,9 +13662,11 @@ void VulkanReplayConsumerBase::OverrideDestroyRenderPass(
 
         if (IsUsedByAsyncTask(renderpass_info->capture_id))
```
antonio-lunarg: The whole body is a repeated pattern that could be "templated-out". Worth mentioning for a follow-up PR.

framework/decode/vulkan_replay_consumer_base.cpp
olehkuznetsov: Good catch. Done

framework/decode/vulkan_replay_consumer_base.cpp
olehkuznetsov: Good catch. Done.

framework/decode/vulkan_replay_consumer_base.cpp
olehkuznetsov: Great idea. Added template.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -14124,28 +14121,93 @@ void VulkanReplayConsumerBase::DestroyAsyncHandle(format::HandleId handle, std::
...
+    GFXRECON_ASSERT(device_info != nullptr);
+    VkDevice                      in_device     = device_info->handle;
+    typename InfoType::HandleType in_handle     = VK_NULL_HANDLE;
+    const VkAllocationCallbacks*  in_pAllocator = GetAllocationCallbacks(pAllocator);
+
+    if (object_info != nullptr)
+    {
+        in_handle = object_info->handle;
+
+        if constexpr (std::is_same_v<InfoType, VulkanPipelineInfo>)
```
antonio-lunarg: I think you can safely delete the whole `if constexpr` block here.

Same rationale applies: `object_info->handle` has already been mapped (correctly with `sync_handle()`) by `Process_vkDestroyPipeline`. A `VK_NULL_HANDLE` in this case means that creation failed.

antonio-lunarg (approved):

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -14169,28 +14121,93 @@ void VulkanReplayConsumerBase::DestroyAsyncHandle(format::HandleId handle, std::
...
+        auto post_it = async_tracked_handles_.find(handle);
+        if (post_it != async_tracked_handles_.end())
+        {
+            if (post_it->second.post_build_fn)
             {
-                handle_asset.post_build_fn();
+                post_it->second.post_build_fn();
             }
-            if (destroy_fn)
+            async_tracked_handles_.erase(post_it);
```
fabian-lunarg: this is erasing tracked handles un-conditionally, ignoring the introduced ref-counting

-> missing a check for empty `async_task_ids`, analog to `ClearAsyncHandles` above

but adding the check would skip deletion of handles, which are still in use, causing leaking of undestroyed handles.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -14169,28 +14121,93 @@ void VulkanReplayConsumerBase::DestroyAsyncHandle(format::HandleId handle, std::
...
-            if (handle_asset.post_build_fn)
+        }
+
+        // Synchronizing tasks above signals completion, but post-creation tasks (TrackNewPipelineCache,
+        // ClearAsyncHandles) were dispatched to MainThreadQueue(). Draining the queue guarantees that
+        // pipeline-to-cache correspondences are registered before destroy_fn() executes.
+        main_thread_queue_.poll();
+
+        auto post_it = async_tracked_handles_.find(handle);
+        if (post_it != async_tracked_handles_.end())
```
fabian-lunarg: I think, after syncing via `main_thread_queue_.poll();` the entire `if (post_it != async_tracked_handles_.end()){...}` block is unreachable.
we should instead just drop this block and assert via:

```
// every synced task has drained and erased its ids by now
GFXRECON_ASSERT(async_tracked_handles_.find(handle) == async_tracked_handles_.end());
```

framework/decode/vulkan_object_info_table_base.h
```diff
@@ -62,8 +63,24 @@ static inline void sync_handle(T* object)
...
+            if (object->future_handle_index < async_handles.size())
+            {
+                object->handle = async_handles[object->future_handle_index];
+            }
+        }
+        catch (const std::exception& e)
+        {
+            // Catch std::exception (not just std::future_error) to prevent std::bad_alloc or other
+            // worker-thread exceptions from escaping during FreeAllLiveObjects in destructors.
+            GFXRECON_LOG_ERROR("Future error while synchronizing handle: %s", e.what());
```
fabian-lunarg: I think it's problematic to try recover here from errors, like e.g. `std::bad_alloc`.
we would carry on with a wrong `VK_NULL_HANDLE` and run into errors later, somewhere else.

so, I suggest this should be `GFXRECON_LOG_FATAL` here

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: (also makes my comment below obsolete)

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -13909,16 +13829,19 @@ std::function<decode::handle_create_result_t<VkPipeline>()> VulkanReplayConsumer
...
             // asynchronous operation is done. clear tracked handles, call deferred deletes
-            ClearAsyncHandles(handle_deps);
+            ClearAsyncHandles(async_task_id, handle_deps);
 
             // if a pipeline cache was created, track it to know when to destroy it/save it to file
             if (cache_pipeline_id != format::kNullHandleId && replay_result == VK_SUCCESS)
             {
-                TrackNewPipelineCache(device_info, cache_pipeline_id, pipeline_cache, pipeline_handles, num_pipelines);
+                // Capture pipeline_handles by value to avoid referencing out_pipelines
+                // after it is moved into the future return value.
```
fabian-lunarg: comment is misleading, moving keeps the buffer alive.
the actual reason to copy is clearing `future = {}` at vulkan_object_info_table_base.h:83

comment could/should say e.g.:
`// copy: the result vector is freed once every info in the batch has synced its future`

framework/decode/vulkan_object_info_table_base.h
```diff
@@ -62,8 +63,24 @@ static inline void sync_handle(T* object)
     static_assert(has_handle_future_v<T>);
     if (object != nullptr && object->handle == VK_NULL_HANDLE && object->future.valid())
     {
-        const auto& [result, async_handles] = object->future.get();
-        object->handle                      = async_handles[object->future_handle_index];
+        try
+        {
+            const auto& [result, async_handles] = object->future.get();
+            // Guard against bounds overrun if pipeline creation failed or returned fewer handles than expected.
+            if (object->future_handle_index < async_handles.size())
```
fabian-lunarg: the guard is misleading and not required. 
`future_handle_index` and `async_handles` are both sized from the original createInfoCount.

here:
framework/decode/vulkan_replay_consumer_base.cpp:13811
`std::vector<VkPipeline> out_pipelines(createInfoCount);`

and here:
framework/decode/common_handle_mapping_util.h:203-210
```
  const auto ids_len_32 = GFXRECON_NARROWING_CAST(uint32_t, ids_len);
  for (uint32_t i = 0; i < ids_len_32; ++i)
  {
      ...
      initial_info.future_handle_index = i;
```
(ids_len comes from the generated caller)

so, instead of the guard, we should state our assumption and assert:
`GFXRECON_ASSERT(object->future_handle_index < async_handles.size());`

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: (see other comment. we prob. shouldn't be here in the first place)

fabian-lunarg:
I agree with the code-changes (minus the few open comments), but some claims in the description are not accurate:

- "handles were mapped before checking IsUsedByAsyncTask ... resolved to VK_NULL_HANDLE"
-> used `MapHandle`, which syncs. `VK_NULL_HANDLE` could only come from failed or skipped creation

- "MapHandle before IsUsedByAsyncTask caused unnecessary stalls"
-> the generated caller had already synced that. and this hasn't changed by the reordering.

- "dangling pointers when out_pipelines was moved": 
-> moving keeps the buffer. the problem was caused by clearing the future

- "Catching only std::future_error allowed other worker exceptions to escape"
-> we didn't catch anything there, just forwarded

meaning, just in case somebody comes back to the PR, we could/should clean the description a bit.

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -14169,28 +14121,93 @@ void VulkanReplayConsumerBase::DestroyAsyncHandle(format::HandleId handle, std::
...
+            return;
+        }
+
+        if constexpr (!std::is_same_v<PreDestroyHook, std::nullptr_t>)
+        {
+            pre_destroy_hook(in_handle);
+        }
+    }
+
+    if (in_handle != VK_NULL_HANDLE)
```
fabian-lunarg: the check should be removed. 
according to [spec](https://docs.vulkan.org/refpages/latest/refpages/source/vkDestroyPipeline.html#VUID-vkDestroyPipeline-pipeline-parameter) passing `VK_NULL_HANDLE` is explicitly allowed.

wrt. "SIGBUS in pipeline::~pipeline()": 
in other places the same nulls already reach the driver without `--pcj` (`SkipPipelineCreationForCompileRequired`) and at teardown (vulkan_object_cleanup_util.cpp:411)
 
also mentioned in other comment: a synced in-flight pipeline is not null

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -14026,16 +13950,19 @@ std::function<handle_create_result_t<VkPipeline>()> VulkanReplayConsumerBase::As
...
             // asynchronous operation is done. clear tracked handles, call deferred deletes
-            ClearAsyncHandles(handle_deps);
+            ClearAsyncHandles(async_task_id, handle_deps);
 
             // if a pipeline cache was created, track it to know when to destroy it/save it to file
             if (cache_pipeline_id != format::kNullHandleId && replay_result == VK_SUCCESS)
             {
-                TrackNewPipelineCache(device_info, cache_pipeline_id, pipeline_cache, pipeline_handles, num_pipelines);
+                // Capture pipeline_handles by value to avoid referencing out_pipelines
+                // after it is moved into the future return value.
```
fabian-lunarg: same like other comment above

framework/decode/vulkan_replay_consumer_base.cpp
olehkuznetsov: Done

framework/decode/vulkan_replay_consumer_base.cpp
olehkuznetsov: Done

framework/decode/vulkan_replay_consumer_base.cpp
olehkuznetsov: Done

framework/decode/vulkan_replay_consumer_base.cpp
olehkuznetsov: Done

framework/decode/vulkan_object_info_table_base.h
olehkuznetsov: Done

framework/decode/vulkan_replay_consumer_base.cpp
olehkuznetsov: Done

framework/decode/vulkan_replay_consumer_base.cpp
olehkuznetsov: Done

framework/decode/vulkan_object_info_table_base.h
olehkuznetsov: Done

framework/decode/vulkan_replay_consumer_base.h
```diff
@@ -2166,6 +2172,34 @@ class VulkanReplayConsumerBase : public VulkanConsumer
...
+    /**
+     * @brief   DestroyTrackedHandle handles synchronous or deferred destruction of Vulkan handles
+     *          (e.g., VkPipeline, VkPipelineLayout, VkRenderPass, VkShaderModule) tracked across replay.
+     *
+     * If the handle is referenced by active asynchronous tasks (e.g. concurrent pipeline compilation),
+     * destruction is deferred via DestroyAsyncHandle until all referencing tasks complete. Otherwise,
+     * it is destroyed immediately.
+     *
+     * For pipelines, if compilation completed asynchronously in the background, the handle is resolved
+     * via MapHandle prior to destruction. An optional PreDestroyHook can be provided to execute cleanup
```
antonio-lunarg: `the handle is resolved via MapHandle prior to destruction` should be removed

antonio-lunarg (approved):

fabian-lunarg (approved):
all my points have been handled. 
happy here when Antonio is happy

PR #3249, opened 2026-09-04
Modify Screenshot code to be more universal

MarkY-LunarG:
We had duplicate functionality and settings spread out across several APIs and some of this would end up being duplicated in certain areas.  Also fixes a few holes we had in screenshot support (bad alignment/format packing issues).

This introduces a `ScreenshotController` which knows everything about a screenshot that is not an API call. It knows no graphics API; it includes only replay_options.h, image_writer.h, and options.h.  Its whole job is policy plus one file write:

 * Frame selection: `IsScreenshotFrame()`, `EndFrame()`, the range/interval state.
 * Naming: `FilenameFor(index, count)`.
 * Sizing policy: `ResolveScale()` — the one place that arbitrates --screenshot-scale vs. --screenshot-size (scale wins).
 * Orientation policy: `ApplyPreRotation()`.
 * Output: `Finish()` — rotate/mirror if asked, then write the file. It owns the reusable rotated_pixels_ buffer.

Ideally, each API would pay attention to the controller, and then separately define how the screenshot occurs and grab the source before using the write image call.  For now, only Vulkan was updated to use a new `VulkanScreenshotSource` class.


MarkY-LunarG:
**NOTE** Also includes several new tests to make sure we don't regress.

framework/graphics/vulkan_util.h
```diff
@@ -132,6 +135,27 @@ static constexpr VkExtent3D ScaleExtent3DNoDepth(const VkExtent3D& extent, float
...
+ *
+ * A negative factor mirrors that axis, thus the size uses the size of the
+ * factor only.  This is the form a screenshot asks for, where the sign of
+ * --screenshot-scale selects a flip.
+ *
+ * @param[in]   extent    The VkExtent3D to scale
+ * @param[in]   scale     The scaling factor for width and for height
+ * @return  The scaled VkExtent3D
+ */
+static constexpr VkExtent3D ScaleExtent3DNoDepth(const VkExtent3D& extent, const std::array<float, 2>& scale)
```
panos-lunarg, line 133: Can the two `ScaleExtent3DNoDepth` be consolidated into a single function?

framework/decode/vulkan_screenshot_handler.cpp
```diff
@@ -0,0 +1,188 @@
...
+#include <cstring>
+#include <memory>
+
+GFXRECON_BEGIN_NAMESPACE(gfxrecon)
+GFXRECON_BEGIN_NAMESPACE(decode)
+
+namespace
+{
+
+bool IsSrgbFormat(VkFormat image_format)
```
panos-lunarg: I think an implementation of this exists in `vk_format_utils.h`: `vkuFormatIsSRGB`

framework/graphics/vulkan_util.h
```diff
@@ -29,6 +29,9 @@
 
 #include "vulkan/vulkan.h"
 
+#include <algorithm>
```
panos-lunarg, line 32: is the `algorithm` include necessary?

tools/tool_settings.h
```diff
@@ -490,7 +490,7 @@ static std::optional<std::array<float, 2>> GetScreenshotScale(const gfxrecon::ut
             }
             else
             {
-                // single value provided — apply uniformly
+                // single value provided -- apply uniformly
```
panos-lunarg: nitpicking: Can the em dash be replaced with a single normal dash instead of 2, or a comma?

framework/graphics/vulkan_resources_util.h
```diff
@@ -123,8 +124,13 @@ class VulkanResourcesUtil
...
         VkImageAspectFlagBits aspect         = VK_IMAGE_ASPECT_NONE;
         bool                  dump_resources = false;
-        float                 scale          = 1.0f;
         VkFormat              dst_format     = VK_FORMAT_UNDEFINED;
+
+        //! A factor for width and one for height.  A negative factor mirrors that axis.
+        std::array<float, 2> scale = { 1.0f, 1.0f };
+
+        //! The first layer to read.  A screenshot reads one layer of an image that has many.
+        uint32_t base_layer = 0;
```
panos-lunarg, line 179: I believe this is not the right approach.

`ImageResource` struct holds information about the image being dumped. This new member variable does not match the pattern as it is basically a request _which_ layer to dump. This is probably not documented well but `ReadImageResources` dumps _all_ subresources of an image to host memory. The order in which subresources are placed in host memory is:
M(`0)L(`0), M(`0`)L(`1`), ... , M(`0`)L(`l-1`)
M(`1`)L(`0`), M(`1`)L(`1`), ... , M(`1`)L(`l-1`)
...
M(`m - 1`)L(`0`), M(`m - 1`)L(`1`), ... , M(`m - 1`)L(`l-1`)
Where `m` is the number of mip levels and `l` is the number of layers (or z indices if it is a 3d image)

`GetImageResourceSizesOptimal` and `GetImageSubresourceSizesDumpResources` return the sizes and offsets in the host memory of each subresource.

From what I understand the purpose of `base_layer` is to place the desired layer first in the host memory and not have to deal with the calculations to locate it in the host memory.

Perhaps a helper function to handle this calculation and return the desired pointer would be better.

framework/decode/vulkan_screenshot_handler.h
```diff
@@ -0,0 +1,96 @@
+/*
+** Copyright (c) 2020 LunarG, Inc.
```
panos-lunarg: 2026

framework/decode/vulkan_screenshot_handler.h
```diff
@@ -0,0 +1,96 @@
...
+** Copyright (c) 2021 Advanced Micro Devices, Inc. All rights reserved.
```
panos-lunarg, line 3: This line is probably not necessary

framework/decode/vulkan_screenshot_handler.cpp
```diff
@@ -0,0 +1,188 @@
+/*
+** Copyright (c) 2020 LunarG, Inc.
```
panos-lunarg: 2026

framework/decode/vulkan_screenshot_handler.cpp
```diff
@@ -0,0 +1,188 @@
...
+** Copyright (c) 2021 Advanced Micro Devices, Inc. All rights reserved.
```
panos-lunarg, line 3: This line is probably not necessary

framework/decode/screenshot_controller.h
```diff
@@ -0,0 +1,201 @@
...
+#include "decode/replay_options.h"
+#include "util/defines.h"
+#include "util/image_writer.h"
+#include "util/options.h"
+
+#include <array>
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <vector>
```
panos-lunarg: I think only `vector` is needed from these #includes

framework/decode/screenshot_controller.h
```diff
@@ -0,0 +1,201 @@
+/*
+** Copyright (c) 2026 LunarG, Inc.
+** Copyright (c) 2021 Advanced Micro Devices, Inc. All rights reserved.
```
panos-lunarg: This line is probably not necessary

framework/decode/screenshot_controller.cpp
```diff
@@ -0,0 +1,184 @@
+/*
+** Copyright (c) 2026 LunarG, Inc.
+** Copyright (c) 2021 Advanced Micro Devices, Inc. All rights reserved.
```
panos-lunarg: This line is probably not necessary

framework/decode/screenshot_controller.h
```diff
@@ -0,0 +1,201 @@
...
+    uint32_t GetCurrentFrame() const { return current_frame_; }
+
+    /**
+     * @brief The name of the file for one of this frame's images, no extension.
+     *
+     * "<prefix>_frame_<n>" for a single image, and
+     * "<prefix>_swapchain_<i>_frame_<n>" when a frame presents more than one,
+     * so the names stay distinct without changing for the common case.
+     */
+    std::string FilenameFor(uint32_t index, uint32_t count) const;
```
panos-lunarg: Nitpick: From what I understand the two arguments are only relevant when multiple swapchains are used. A comment saying so or adjusting the argument names would help with this.
Also 3 out of 4 callers pass (0, 1). Perhaps making 0 and 1 for each argument respectively the default value would be help with readability. Let only the `WriteScreenshots` caller which handles swaphains pass the swapchain index and the total swapchain count.

framework/decode/vulkan_screenshot_handler.cpp
```diff
@@ -0,0 +1,188 @@
...
+                        (pre_transform == VK_SURFACE_TRANSFORM_HORIZONTAL_MIRROR_ROTATE_90_BIT_KHR) ||
+                        (pre_transform == VK_SURFACE_TRANSFORM_HORIZONTAL_MIRROR_ROTATE_180_BIT_KHR) ||
+                        (pre_transform == VK_SURFACE_TRANSFORM_HORIZONTAL_MIRROR_ROTATE_270_BIT_KHR);
+
+    return rotation;
+}
+
+bool VulkanScreenshotSource::Readback(const ScreenshotRequest& request, CpuImage* out)
+{
+    GFXRECON_ASSERT(out != nullptr);
```
panos-lunarg: Making `out` a reference would render the assert unnecessary

framework/graphics/vulkan_util.h
MarkY-LunarG: Yes, for std::max in the ScaleExtend3DNoDepth function.

framework/graphics/vulkan_util.h
MarkY-LunarG: Yes, I just have to repeat the value, but I can reduce it down to the one func.

framework/decode/vulkan_screenshot_handler.h
MarkY-LunarG: This is pulling code from a file that AMD contributed to.  I'd rather keep it just in case some of the code pulled originated from them.

framework/decode/vulkan_screenshot_handler.cpp
MarkY-LunarG: See comment two above.

framework/graphics/vulkan_resources_util.h
MarkY-LunarG: Dumping out each layer is still supported, it just does it from the `VulkanReplayConsumer` now (line 2660) and `base_layer = 0` is only the initializer.  It is overridden in vulkan_screenshot_handler.cpp:117.

So `ReadImageResources` reads levels `[0, level_count)` and layers `[base_layer, base_layer + layer_count)` of each image. 

Your output order is still what should be generated.

I did create some helper functions to determine the offset of a given level/layer within a texture.

panos-lunarg: Yes functionality does not change but there are a couple of issues if you modify the logic to not dump all layers from layer 0.
1. `VulkanResourcesUtil::ResolveImage` and `VulkanResourcesUtil::RenderPassResolve` need to have their logic also adapted to this change since they still assume base layer to be 0.
2. `GetImageResourceSizesOptimal` and `GetImageSubresourceSizesDumpResources` stop working as expected when base_layer != 0. This is a discrepancy although minor.
3. The `for (uint32_t layer = 0; layer < num_layers; ++layer)` for loop in framework/decode/vulkan_replay_consumer_base.cpp#2660 calls `Readback` per layer, each iteration will allocate and deallocate resources and will do a separate queue submit and wait for the roundtrip. Dumping all layers in one go would be more efficient

MarkY-LunarG: 1. Yes, your correct
2. I'm not sure how.  Can you provide more details on what you think is happening?
3. I'll make the change.  However, the biggest downside to this is that we are allocating much larger resources to perform the copy, but if we're more concerned about the number of submits I agree this is a better approach.  We might want to have both paths eventually (since mobile might want the other mechanism where submitting multilple times is a priority over consuming much more memory, especially as framebuffer sizes grow).

MarkY-LunarG: Changes pushed up in last two commits

panos-lunarg (approved):

PR #3251, opened 2026-09-07
Miscellaneous replay settings fixes

marius-pelegrin-arm:
This PR contains 2 unrelated commits:

### Add some forgotten semicolons
It does not change the behavior of GFXReconstruct, but fix some review tools.

### Move `--onhb` option to Vulkan-only settings
This option is only implemented for Vulkan and, as far as I know, only makes sense for Vulkan (I don't see why you could import Android buffers on DirectX).

This commit also adds a description for the option, which was missing.

### DIFF
--- a/tools/replay/replay_vulkan_feature.cpp
+++ b/tools/replay/replay_vulkan_feature.cpp
@@ -112,6 +112,8 @@ const char kReplayEventPluginPath[]                 = "--replay-event-plugin-pat
 const char kReplayEventPluginParams[]               = "--replay-event-plugin-params";
 const char kIsolateRenderPasses[]                   = "--isolate-render-passes";
 const char kSerializeComputeAndTransfer[]           = "--serialize-compute-and-transfer";
+const char kOmitNullHardwareBuffersShortOption[]    = "--onhb";
+const char kOmitNullHardwareBuffersLongOption[]     = "--omit-null-hardware-buffers";
 
 const char kMemoryTranslationNone[]    = "none";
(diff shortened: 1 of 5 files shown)


tools/replay/replay_vulkan_feature.cpp
```diff
@@ -714,6 +718,11 @@ std::vector<util::FeatureOptionDesc> ReplayVulkanFeature::GetOptionDescs() const
                  "debug, info, warning, and error. The default is warning." },
                true,
                kDebugMessageSeverityArgument },
+             { "",
+               { "Skip calls to `vkGetAndroidHardwareBufferPropertiesANDROID` when the provided Android",
+                 "AHardwareBuffer was NULL at capture time." },
```
fabian-lunarg: capture time -> replay time

marius-pelegrin-arm: Updated.

fabian-lunarg:
wrt. `--onhb` ... "This option is only implemented for Vulkan"

agreed. we could complement your change and move `decode::ReplayOptions::omit_null_hardware_buffers`,
it belongs into `decode::VulkanReplayOptions`

marius-pelegrin-arm:
> agreed. we could complement your change and move `decode::ReplayOptions::omit_null_hardware_buffers`, it belongs into `decode::VulkanReplayOptions`

Yes. Actually that's how I implemented it internally but I forgot this part of the code while porting it for this PR. I just updated it =)

fabian-lunarg (approved):
alright!

PR #3253, opened 2026-09-07
Add replay block index skipping

zhangzhousuper:
Add `--skip-index` to skip selected capture blocks during replay, for example `--skip-index 7,10-20`. Indices are zero-based physical block positions and ranges include both endpoints. Function-call, method-call, and metadata blocks can be skipped; structural markers and annotations continue to be processed.

It's for debug purpose to detect which api(e.g. drawcall) introduce failure.


framework/decode/file_processor.cpp
```diff
@@ -63,6 +63,15 @@ void FileProcessor::SetBlocksToSkip(const std::unordered_set<uint64_t>& blocks_t
     }
 }
 
+void FileProcessor::SetSkipBlockIndices(const std::vector<util::UintRange>& ranges)
+{
+    GFXRECON_ASSERT(!frame_processing_initialized_);
+    if (!frame_processing_initialized_)
+    {
+        skip_block_indices_ = ranges;
```
jzulauf-lunarg, line 71: These are never used directly by the `FileProcessor`. Rename to "pending" to match the other forwarded settings

framework/decode/file_processor_types.h
```diff
@@ -330,6 +332,51 @@ struct ContinueProcessingPolicy
     };
 };
 
+// Replay ranges are kept compact, including when a range covers the entire uint32_t domain.
+class BlockSkipRanges
```
jzulauf-lunarg, line 336: This seems more generic... maybe in the util:: namespace with a more neutral naming as opposed to stuffing into the file_processor namespace

framework/decode/block_processor.cpp
```diff
@@ -136,7 +136,12 @@ file_processor::ProcessBlockState BlockProcessor::ProcessBlock(Policy& policy, B
     file_processor::ProcessVisitor    process_visitor(*this);
 
     const bool skip_block = block_skip_ && (*block_skip_)(block_index_);
-    if (!skip_block)
+    const bool skip_index = skip_block_ranges_(block_index_, block_buffer.Header().type);
```
jzulauf-lunarg, line 139: if (skip_block) why check for skip index?  Why the separate state.

```
if (!skip_block) {
    skip_block = skip_block_ranges(...
    if (skip_block) { ...LOG...)
}
```
no wasted lookups... no extra state.

jzulauf-lunarg (changes requested):

framework/decode/file_processor.cpp
zhangzhousuper: thx,updated

framework/decode/file_processor_types.h
zhangzhousuper: updated

framework/decode/block_processor.cpp
zhangzhousuper: updated

PR #3254, opened 2026-09-07
Add overload for TemporaryCommandBuffer's `CreateAndBegin` with `VulkanQueueInfo`

wzchu-lunarg:


framework/decode/vulkan_temporary_objects.cpp
```diff
@@ -41,6 +41,32 @@ VkResult TemporaryCommandBuffer::CreateAndBegin(graphics::FindQueueFamilyIndex_f
...
+VkResult TemporaryCommandBuffer::CreateAndBegin(const VulkanQueueInfo& queue_info)
+{
+    GFXRECON_ASSERT(queue_info.handle != VK_NULL_HANDLE);
+
+    queue = queue_info.handle;
+
+    return CreateAndBeginForQueue(queue_info.family_index);
+}
+
+VkResult TemporaryCommandBuffer::CreateAndBeginForQueue(uint32_t queue_family_index)
```
antonio-lunarg: You can avoid the need to call another function by accepting a `VkQueue` parameter here. Then the other two `CreateAndBegin` will just forward their arguments.

antonio-lunarg (changes requested):

framework/decode/vulkan_temporary_objects.cpp
wzchu-lunarg: Ah now I see the intention. We can just pass the `VkQueue` so we won't need to `GetDeviceQueue` with the queue index unless we don't have that handle.

antonio-lunarg (approved):

PR #3256, opened 2026-09-08
Temporary buffers for injected buffers

wzchu-lunarg:
Create `TemporaryBuffer` to simplify creation of injected buffers.


framework/decode/vulkan_temporary_objects.h
```diff
@@ -130,14 +132,7 @@ struct TemporaryCommandBuffer
...
 
-    ~TemporaryCommandBuffer()
-    {
-        if (command_pool != VK_NULL_HANDLE)
-        {
-            auto injected = device_table.Open();
-            injected->DestroyCommandPool(device_info.handle, command_pool, nullptr);
-        }
-    };
+    ~TemporaryCommandBuffer();
```
antonio-lunarg: Not destroying the command pool because we expect the command pool to create other command buffers and be destroyed somewhere else?

framework/decode/vulkan_temporary_objects.h
```diff
@@ -175,6 +170,53 @@ struct TemporaryQueryPool
...
+        device(dev),
+        allocator(alloc), device_table(injected_calls)
+    {}
+
+    TemporaryBuffer(VkDevice dev, VulkanResourceAllocator* alloc, const graphics::VulkanDeviceTable& dev_table) :
+        TemporaryBuffer(dev, alloc, graphics::VulkanInjectedDeviceCalls(&dev_table))
+    {}
+
+    // TemporaryBuffer can be a member of structs that are used in containers, so ownership of the buffer has to
+    // transfer rather than be duplicated.
```
antonio-lunarg: Being a member of a struct used in containers does not sound as a valid reason to forbid copy of a resource. I would just forbid copy because by design we do not expect to duplicate this resource via C++ copy semantics.

framework/decode/vulkan_temporary_objects.h
```diff
@@ -175,6 +170,53 @@ struct TemporaryQueryPool
...
+    TemporaryBuffer(const TemporaryBuffer&)            = delete;
+    TemporaryBuffer& operator=(const TemporaryBuffer&) = delete;
+
+    // Moving lets a buffer live in a container that reallocates, and lets a default-constructed member be
+    // configured later by assigning a fully constructed buffer over it.
```
antonio-lunarg: This comment is redundant, imo, we have no reason to explain C++ in docs of our classes

framework/decode/vulkan_temporary_objects.h
```diff
@@ -175,6 +170,53 @@ struct TemporaryQueryPool
...
+    TemporaryBuffer(TemporaryBuffer&& other) noexcept;
+    TemporaryBuffer& operator=(TemporaryBuffer&& other) noexcept;
+
+    void swap(TemporaryBuffer& other) noexcept;
+
+    // Destroys the buffer.  Any memory the caller bound to it has to outlive this and be freed afterwards.
```
antonio-lunarg: I am confused by this comment. This class manages its own memory via `allocator` and frees its memory within the `Destroy()` function.

framework/decode/vulkan_temporary_objects.h
```diff
@@ -34,6 +34,8 @@
 GFXRECON_BEGIN_NAMESPACE(gfxrecon)
 GFXRECON_BEGIN_NAMESPACE(decode)
 
+struct VulkanDeviceInfo;
```
antonio-lunarg, line 39: Is this unused?

framework/decode/vulkan_temporary_objects.cpp
```diff
@@ -211,5 +224,92 @@ VkResult TemporaryQueryPool::Create(uint32_t query_count)
...
+    }
+    return *this;
+}
+
+void TemporaryBuffer::swap(TemporaryBuffer& other) noexcept
+{
+    std::swap(handle, other.handle);
+    std::swap(size, other.size);
+    std::swap(requirements, other.requirements);
+    std::swap(resource_data, other.resource_data);
```
antonio-lunarg, line 311: What about device, allocator and device_table?

framework/decode/vulkan_temporary_objects.cpp
```diff
@@ -211,5 +224,92 @@ VkResult TemporaryQueryPool::Create(uint32_t query_count)
     return res;
 }
 
+VkResult TemporaryBuffer::Create(VkDeviceSize buffer_size, VkBufferUsageFlags usage)
```
antonio-lunarg: I am not a fun of semi-initialized objects: this results in constructing objects with an intermediate/invalid/uninitialized state that can be avoided IMO. Just move this logic in the constructor.

framework/decode/vulkan_temporary_objects.h
wzchu-lunarg: The main reasoning I should have highlighted is that if it is duplicated, we will call `vkDestroyBuffer` on a buffer that is already destroyed.

framework/decode/vulkan_temporary_objects.h
wzchu-lunarg: `VulkanDeviceInfo` was from including `vulkan_temporary_objects.h`. However, with the changes to these temporary buffers, we end up getting a circular include from `vulkan_temporary_objects` -> `vulkan_resource_initializer` -> `vulkan_temporary_objects`.

This is to forward declare it without including `vulkan_temporary_objects`.

framework/decode/vulkan_temporary_objects.h
wzchu-lunarg: Adding a commit to make this more obvious.

framework/decode/vulkan_temporary_objects.h
wzchu-lunarg: Sounds good, deleting this.

framework/decode/vulkan_temporary_objects.cpp
wzchu-lunarg: The original intent I had here was to follow the "Create" functions similar to the other temporary objects such as `TemporaryCommandBuffer::CreateAndBegin` and `TemporaryFence::Create`, which ends up returning a `VkResult`. This `VkResult` is sometimes used by the callers as well to see if the buffer creation failed. 

However, I do agree that it is better to move everything in the constructor if we can. Adding a commit for this.

framework/decode/vulkan_temporary_objects.cpp
wzchu-lunarg: The original intent was specifying these members in the constructor, while the swapped members were from the `Create` function. 

Changing this since everything is now in the constructor.

antonio-lunarg (approved):

PR #3257, opened 2026-09-09
Replay pre-frame blocks before the first FrameBegin

antonio-lunarg:
`BlockProcessor::ProcessBlocks()` returns `kPreFrameBoundary` once, before the first block that belongs to frame 0. It peeks the next block header and treats leading meta, annotation, and state marker blocks, and every block inside a state block, as pre-frame. `FileProcessor::ProcessPreFrame()` replays up to that boundary, and `Application::Run()` calls it once before the frame loop.

A trimmed capture's state setup now runs before FrameBegin 0, so its queue submits emit no plugin events and the first frame interval holds only frame work.

Full captures replay the same blocks in the same order, only the two leading meta blocks move ahead of FrameBegin 0.

### DIFF
--- a/framework/decode/block_processor.cpp
+++ b/framework/decode/block_processor.cpp
@@ -109,6 +109,39 @@ bool BlockProcessor::IsFileValid() const
     return false;
 }
 
+bool BlockProcessor::NextBlockIsFrame() const
+{
+    if (file_stack_.empty())
+    {
+        return false;
+    }
+
+    // Within the state setup, next block might be any type, but that does not mean they belong to a frame.
+    if (in_state_setup_)
+    {
+        return false;
+    }
+
+    // Outside of state setup, a meta data, annotation, or state marker block type,
+    // means that the next block is not part of a frame.
+    format::BlockHeader header{};
+    const size_t        peeked = GetCurrentFile().active_file->PeekBytes(&header, sizeof(header));
+    if (peeked < sizeof(header))
+    {
+        return false;
+    }
+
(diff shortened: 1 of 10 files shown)


framework/decode/block_processor.cpp
```diff
@@ -553,6 +601,12 @@ BlockProcessor::ActiveFileContext& BlockProcessor::GetCurrentFile()
     return file_stack_.back();
 }
 
+const BlockProcessor::ActiveFileContext& BlockProcessor::GetCurrentFile() const
+{
+    GFXRECON_ASSERT(file_stack_.size());
```
nickdriscoll-lunarg: Should explicitly compare against zero e.g. `file_stack_.size() > 0`

framework/decode/block_processor.h
```diff
@@ -150,6 +151,14 @@ class BlockProcessor
     bool HasPendingBlocksToSkip() const noexcept { return !pending_blocks_to_skip_.empty(); }
 
   private:
+    /// This is true until the first block that belongs to frame 0 is encountered.
+    /// Meta-data, annotation, and state marker blocks are considered pre-frame.
+    bool pre_frame_boundary_pending_{ true };
+
+    /// This is true as once the block processor encounters a state begin marker block,
```
nickdriscoll-lunarg: Remove "as" from this sentence.

nickdriscoll-lunarg (commented):
Just some nitpicks. Looks pretty good overall!

PR #3260, opened 2026-09-09
[Frame Looping] Image content restoration

wzchu-lunarg:
**Problem**
Similarly to https://github.com/LunarG/gfxreconstruct/pull/3234, image contents can also mutate during loop iterations, and we would need to restore these when frame looping.

**Main Changes**
Create an `ImageTracking` class for tracking image layout and content. We then track images in `vkCreateImage` and inject transfer usage flags.

Copying images requires transitioning image layouts of both the buffer we want to restore, and the shadow buffers with `VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL` and `VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL`. This involves doing transitions to and from these layouts on the initial record and each end of frame restoration.

### DIFF
--- a/framework/decode/vulkan_replay_frame_loop_consumer.cpp
+++ b/framework/decode/vulkan_replay_frame_loop_consumer.cpp
@@ -1768,8 +1768,6 @@ void VulkanReplayFrameLoopConsumer::ImageTracking::RecordInitialState(
 
     // Puts every copyable subresource of each source image into VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL.
     std::vector<VkImageMemoryBarrier> pre_barriers;
-    // Transitions those subresources back out of VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL.
-    std::vector<VkImageMemoryBarrier> post_barriers = pre_barriers;
 
     for (format::HandleId image_id : recorded_ids)
     {
@@ -1779,6 +1777,9 @@ void VulkanReplayFrameLoopConsumer::ImageTracking::RecordInitialState(
         AppendTransferLayoutBarriers(
             image_info, state.excluded_ranges, VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL, pre_barriers);
     }
+
+    // Transitions those subresources back out of VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL.
+    std::vector<VkImageMemoryBarrier> post_barriers = pre_barriers;
 
     ReverseLayoutBarriers(post_barriers);
 
--- a/framework/decode/vulkan_replay_frame_loop_consumer.cpp
+++ b/framework/decode/vulkan_replay_frame_loop_consumer.cpp
@@ -1725,6 +1725,8 @@ void VulkanReplayFrameLoopConsumer::ImageTracking::DestroyShadowImage(ImageState
 void VulkanReplayFrameLoopConsumer::ImageTracking::RecordInitialState(
     const std::vector<format::HandleId>& image_ids, const std::vector<format::HandleId>& restorable_image_ids)
 {
+    restore_commands_ = {};
+
     // Record layouts
     for (format::HandleId image_id : image_ids)
     {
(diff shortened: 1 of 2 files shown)


framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -1271,100 +1348,633 @@ static void AppendImageLayoutRestoreBarriers(const VulkanImageInfo*
...
+    VulkanDeviceInfo* device_info = object_table_.GetVkDeviceInfo(device_id_);
+    GFXRECON_ASSERT(device_info != nullptr);
+
+    VkImageCreateInfo create_info = { VK_STRUCTURE_TYPE_IMAGE_CREATE_INFO };
+    create_info.imageType         = image_info->type;
+    create_info.format            = image_info->format;
+    create_info.extent            = image_info->extent;
+    create_info.mipLevels         = image_info->level_count;
+    create_info.arrayLayers       = image_info->layer_count;
+    create_info.samples           = VK_SAMPLE_COUNT_1_BIT;
```
antonio-lunarg: I think sample count should match the original image for vkCmdCopyImage to work.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -377,6 +382,60 @@ void VulkanReplayFrameLoopConsumer::Process_vkCreateBuffer(const ApiCallInfo& ca
...
+    return false;
+}
+
+void VulkanReplayFrameLoopConsumer::Process_vkCreateImage(const ApiCallInfo& call_info, args::CreateImage& args)
+{
+    VkImageCreateInfo* create_info = args.pCreateInfo.GetPointer();
+
+    if ((create_info != nullptr) && !args.pImage.IsNull() && !CanRestoreImage(create_info))
+    {
+        create_info->usage |= VK_IMAGE_USAGE_TRANSFER_SRC_BIT | VK_IMAGE_USAGE_TRANSFER_DST_BIT;
```
antonio-lunarg, line 443: We can to something useful here before setting these usages: check whether the VkFormatProperties for this image permit transfer usage.

wzchu-lunarg: Added a `IsFormatSupported` check for transfer bits.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Missed this. Adding a commit to use the original image's sample count.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
antonio-lunarg: I noticed `IsFormatSupported()` only accepts optimal and linear tiling. Its implementation looks _specific_ to me. It does not seem that it's checking whether a certain format is supported by the current device. @panos-lunarg what was your intention when you implemented it?
IMO, it needs to be properly documented and its name should convey more info. Something like `IsCombinationOfFormatTilingAndUsage()` is probably too much, but  clearer :D

wzchu-lunarg: It looked to me that it does check whether a certain format _with certain features_ is supported by the device. It does the `GetPhysicalDeviceFormatProperties` call to get features, but then checks based on if the image's tiling is linear or optimal. 

For example, on my device, when checking this with a `VK_FORMAT_R8G8B8A8_SRGB` format image, I get `0x0001dd81` for `optimalTilingFeatures` and `0x0001dc01` for `linearTilingFeatures`, both of which support the transfer bits of `0x00008000` and `0x00004000` if we check the bits for `0x0000d000`.

antonio-lunarg: Yes, and `IsFormatSupported()` expects its tiling parameter to be either optimal or linear. But these are not the only valid tiling modes for an image, therefore passing any other tiling mode is going to violate the contract of that function: in debug mode is going to crash, in release mode is going to do something unexpected by the designer of the function.

Anyway I would expect an `IsFormatSupported()` to just accept a VkFormat parameter as input and a return value that would tell me whether the device supports that format.
With its current shape, the name of this function is misleading as it's actually doing something more specific that involves checking more parameters in combination with a Vulkan image.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -577,12 +577,19 @@ void VulkanReplayFrameLoopConsumer::RecordBufferStates()
...
 
     std::unordered_map<format::HandleId, std::vector<format::HandleId>> device_buffers;
-    table.VisitVkBufferInfo([&device_buffers](const VulkanBufferInfo* buffer_info) {
+    table.VisitVkBufferInfo([&table, &device_buffers](const VulkanBufferInfo* buffer_info) {
         if (buffer_info == nullptr || buffer_info->handle == VK_NULL_HANDLE || buffer_info->size == 0)
         {
             return;
         }
 
+        // A buffer can outlive the device that created it in the object info table, in which case
```
antonio-lunarg: This does not sound good, surely VVL would complain in this case, but some defensive programming does not hurt. I am a bit unsure about the right strategy for handling this condition, but in general I prefer `GFXRECON_ASSERT` here as I expect the device to be available in the context of restoring the state for frame looping.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -1601,6 +1608,11 @@ void VulkanReplayFrameLoopConsumer::TrackImageStates()
                 return;
             }
 
+            if (GetObjectInfoTable().GetVkDeviceInfo(image_info->parent_id) == nullptr)
```
antonio-lunarg: Same reasoning for the image handle and device info, both are expected to be valid when restoring their state at frame looping time, hence I would prefer GFXRECON_ASSERT rather than ignoring .

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Sounds good. Adding a commit for this.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Sounds good, adding a commit for this.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Yea after looking again that does seem to be an issue.

After a quick search, I believe the only other tiling mode is `VK_IMAGE_TILING_DRM_FORMAT_MODIFIER_EXT`? Perhaps we can do a check for it, or just ignore this extension?

wzchu-lunarg: Adding a commit to check this tiling mode.

antonio-lunarg: Last but not least, changing the usage might change an image's memory requirements and might break replay. I think it's worth creating an issue demanding thorough testing for this case (to be addressed in a subsequent PR).

antonio-lunarg (approved):

PR #3261, opened 2026-09-10
Reduce log spam when unnecessary

marius-pelegrin-arm:
When `vkGetFenceStatus`/`vkWaitForFences` returns `VK_SUCCESS` at replay time when it returned `VK_NOT_READY`/`VK_TIMEOUT` at capture time, the result is accepted but still displayed in the logs, which can lead to bad log spam.
This is because even if a result code difference is explicitly stated as accepted, it is still logged in debug logs.

I think this is unnecessary and this log should be removed.


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -1698,11 +1698,6 @@ void VulkanReplayConsumerBase::CheckResult(const char*                func_name,
                 GFXRECON_LOG_WARNING("%s.", log_str);
             }
         }
-        else
-        {
-            // in case we accept a mismatching return-code, just log a debug-message
-            GFXRECON_LOG_DEBUG("%s.", log_str);
```
fabian-lunarg, line 1704: this is not specific to `vkGetFenceStatus` / `vkWaitForFences` and unconditionally removes debug-logging for all calls returning a mismatching but accepted code. 
I'm happy to hear somebody is actually reading those but not sure I can agree here.
the logging here is deliberately  demoted to debug and debug-logging is opt-in already.
imo, we could/would loose potentially valuable debug-information in some (future) cases.

per-mathisen-arm: Time to introduce a debug log level? :-)

We usually run with debug log on in our CI and this shows up enough times that it drowns out everything else and causes a slowdown. I don't think this particular log message is particularly useful.

marius-pelegrin-arm: If you prefer, another possible solution could be *NOT* to return `VK_SUCCESS` at replay time when the replay time result is `VK_SUCCESS` and the capture time result was `VK_NOT_READY` for this particular call (`vkGetFenceStatus`). This would stop the log spam without "hiding" logs for all other calls.
This could also be done for `vkGetEventStatus` which has a similar issue.

fabian-lunarg:
I think, in order to achieve your goal (which we can agree on, in case of fence/event those logs aren't helpful),
but not cripple other logging we actually want, we should refine our grouping of cases:

`VK_TIMEOUT` -> `VK_SUCCESS`, `VK_NOT_READY` -> `VK_SUCCESS`, `VK_EVENT_RESET` -> `VK_EVENT_SET`
-> benign, caused by replay-**timing**, and basically un-interesting to log.

the other candidates not so much:

`VK_ERROR_OUT_OF_DATE_KHR`, `VK_SUBOPTIMAL_KHR`, `VK_ERROR_OUT_OF_POOL_MEMORY`, `VK_ERROR_FORMAT_NOT_SUPPORTED` etc. 
-> they relate more to differences in replay-**environment** we must not brush under the rug.

we need a different/additional category, meaning we should split the `kResultValuesAllowedDifferentCodeThanCapture` LUT

framework/decode/vulkan_replay_consumer_base.cpp
fabian-lunarg: > ... for this particular call

I think it's not just this one call, we'd need several spot-solutions. 
see my comment below, my favorite would be a different 'flavour' of `kResultValuesAllowedDifferentCodeThanCapture`, which could achieve a better weighting/granularity of log-messages.

fabian-lunarg: LUT split done

fabian-lunarg:
@marius-pelegrin-arm I added a commit on top, doing the LUT-split I described above.
-> no more debug-logging for `VK_TIMEOUT` -> `VK_SUCCESS`, `VK_NOT_READY` -> `VK_SUCCESS`, `VK_EVENT_RESET` -> `VK_EVENT_SET`

fabian-lunarg (approved):

PR #3265, opened 2026-09-11
generators: take static array lengths from the registry len attribute

jzulauf-lunarg:
Replace the '<name>Count' sibling heuristic with the registry len. Heuristic was incomplete, and no longer needed.
Future captures of VkQueueFamilyGlobalPriorityProperties::priorities and VkShaderModuleIdentifierEXT::identifier will encode only up to length. 

Decoder compatibility is maintained.

Encoder clamps the count to capacity; decoder warns on a length matching neither count nor capacity.

### DIFF
--- a/framework/generated/generated_vulkan_struct_encoders.cpp
+++ b/framework/generated/generated_vulkan_struct_encoders.cpp
@@ -1092,9 +1092,11 @@ void EncodeStruct(ParameterEncoder* encoder, const VkPhysicalDeviceLimits& value
 void EncodeStruct(ParameterEncoder* encoder, const VkPhysicalDeviceMemoryProperties& value)
 {
     encoder->EncodeUInt32Value(value.memoryTypeCount);
-    EncodeStructArray(encoder, value.memoryTypes, value.memoryTypeCount);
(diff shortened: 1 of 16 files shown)


framework/generated/generated_vulkan_struct_encoders.cpp
```diff
@@ -1092,9 +1092,11 @@ void EncodeStruct(ParameterEncoder* encoder, const VkPhysicalDeviceLimits& value
 void EncodeStruct(ParameterEncoder* encoder, const VkPhysicalDeviceMemoryProperties& value)
 {
     encoder->EncodeUInt32Value(value.memoryTypeCount);
```
wzchu-lunarg, line 1094: Do we also need to encode the clamped value `memoryTypes_count` instead of just `value.memoryTypeCount`?

jzulauf-lunarg: @wzchu-lunarg -- I think we what to record as is.  The replay side allocates/copies based on the clamped values.

wzchu-lunarg (approved):

PR #3266, opened 2026-09-14
optimize: Add modifier framework, use for aliasing detection/recovery

fabian-lunarg:
this PR introduces two new things:

1. A modifier framework for `gfxrecon-optimize`
- Adds `CallModifierBase`+`BlockEdit` abstractions for passes that rewrite, insert or delete blocks
- 'unreferenced-resource scan' and 'modifier scan' share one pass
- blocks which no modifier rewrote are passed through verbatim (no re-encoding)
- deleted blocks are replaced with annotations
- first modifier: aliasing-group detection, **on by default**
-> tracks binds, groups overlapping resources in device memory, 
   records the groups in a new meta-command block into the optimized capture (after `vkCreateDevice`)
   
2. A new meta-command (useful to correct `-m rebind` replay wrt. aliasing reproduction)
- introduce `kResourceAliasingGroupsCommand` (Block ID 48)
-> contains per device memory objects, aliasing groups and members' create-infos and bind offsets
-> replay decodes it and queries replay memory requirements
-> rebind allocator can now size one allocation per group and place members at correct offsets
- "-m rebind" now requests `VK_KHR_maintenance4` (when Vulkan version <1.3), used to query memory-requirements


MarkY-LunarG:
Review comments:

```
Verdict: the design is sound and the refactors are behavior-preserving, but there is one crash-class bug that affects every replay, and three medium issues in the new group placement path.

  Findings, most severe first

  1. High. Empty optional dereferenced in device creation. framework/decode/vulkan_replay_consumer_base.cpp:3578 reads the replay device's properties optional unconditionally. That optional is only filled if the
     captured app called vkGetPhysicalDeviceProperties, which is why the post-create code at line 3697 fills it lazily. This runs on every replay, not just rebind. Undefined behavior in release, abort with
     libstdc++ assertions.
  2. Medium. Group path drops the unaliased-overlap guard. The #3100 per-bind path calls OverlapsUnaliasedResource before sharing an allocation. The new group path does not. Groups are transitive, so two members
     that were disjoint at capture can overlap in the shared block once replay sizes inflate. Both live, both written, silent corruption. The old path refused this case.
  3. Medium. Group block can leak. The shared block is allocated before the per-member size, type and alignment guards run. If the first member fails a guard, the block stays at reference count zero and is never
     freed until allocator teardown. Alignment mismatch on a different GPU is the likely trigger.
  4. Medium. Usage-bit mismatch between requirement query and creation. The meta block is processed during state setup, so members get the TRANSFER_DST bit added. Members created inside the frame range do not.
     Their requirements can differ from the group layout, so they fall back to per-resource placement with a warning. Silent loss of the feature, not a crash.
  5. Medium, inherited. UpdateAllocInfo copies host-written original content over the whole shared block on every member bind, clobbering GPU-written data of already-bound members. Pre-existing in #3100, but
     groups make it much more frequent.
  6. Low. Decoder reserves from unvalidated counts and can throw on a corrupt block. Optimizer and replayer size buffer footprints by different rules. Re-optimizing an optimized file writes a duplicate block. The
     base-class doc for SetResourceAliasingGroups says replace, the code merges.

  What checked out: the buffer, image and tensor create-info refactor is equivalent to the old inline code. Mapped memory writes reach group members correctly. Group pointers stay valid across map growth. Struct
  packing is consistent between writer and reader. No new threading. The three dev commits underneath the branch are clean.

  Tests: gfxrecon_framework_test, gfxrecon_optimize_test, and the [rebind] cases in gfxrecon_decode_test all pass on the current build.

  Item 1 needs fixing before merge. Items 2 and 3 are worth fixing now while the code is fresh. Items 4 and 5 are design choices Fabian should weigh in on.
```

fabian-lunarg:
thank you @MarkY-LunarG, went through the list, 
and added couple commits, which should handle points 2,3 and most of 6.

1) was checked, the guard tightened and verified the optional is never empty, also added a note.

I left out 4) "Usage-bit mismatch between requirement query and creation"
the mismatch is benign for the derived size, it could only ever over-allocate.

and suggest to follow-up on 5) "UpdateAllocInfo copies host-written original content over the whole shared block"
-> was pre-existing and is a general thing to improve for both aliasing code-paths (runtime + new meta-command).

MarkY-LunarG (approved):

PR #3267, opened 2026-09-14
Synch access to VkQueue from AHB readback code

panos-lunarg:
CommonProcessHardwareBuffer access to the application's queue objects is
now guarded by a mutex in order to synchronize accesses to the same
queue objects by other API calls made by the application or the capture
layer itself


framework/encode/vulkan_capture_manager.h
```diff
@@ -1691,6 +1691,25 @@ class VulkanCaptureManager : public ApiCaptureManager
...
+    VkResult OverrideQueueSubmit2KHR(VkQueue queue, uint32_t submitCount, const VkSubmitInfo2* pSubmits, VkFence fence);
+
+    VkResult OverrideQueuePresentKHR(VkQueue queue, const VkPresentInfoKHR* pPresentInfo);
+
+    VkResult OverrideQueueWaitIdle(VkQueue queue);
+
+    VkResult OverrideDeviceWaitIdle(VkDevice device);
+
+    VkResult
+    OverrideQueueBindSparse(VkQueue queue, uint32_t bindInfoCount, const VkBindSparseInfo* pBindInfo, VkFence fence);
```
fabian-lunarg, line 1711: `OverrideQueueBindSparse` is not reachable:
- missing in capture_overrides.json -> codegen doesn't forward here

panos-lunarg: Fixed

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -4537,5 +4532,179 @@ void VulkanCaptureManager::PostProcess_vkTransitionImageLayout(VkResult
...
+                                                         const VkDeviceQueueInfo2* pQueueInfo,
+                                                         VkQueue*                  pQueue)
+{
+    if (pQueue != nullptr && *pQueue != VK_NULL_HANDLE && device != VK_NULL_HANDLE && pQueueInfo != nullptr)
+    {
+        auto* device_wrapper = vulkan_wrappers::GetWrapper<vulkan_wrappers::DeviceWrapper>(device);
+        auto* queue_wrapper  = vulkan_wrappers::GetWrapper<vulkan_wrappers::QueueWrapper>(*pQueue);
+
+        std::lock_guard<std::mutex> lock(device_wrapper->queues_map_mutex);
+        device_wrapper->child_queues[pQueueInfo->queueFamilyIndex][pQueueInfo->queueIndex] = queue_wrapper;
```
fabian-lunarg: this needs attention (regression vs. dev) 

the key(s) are not unique and we need to take [VUID-02802](https://vulkan.lunarg.com/doc/view/1.4.304.1/linux/antora/spec/latest/chapters/devsandqueues.html#VUID-VkDeviceCreateInfo-queueFamilyIndex-02802) into account

a device may legally have both `{family = 0, flags = 0}` and `{family = 0, flags = VK_DEVICE_QUEUE_CREATE_PROTECTED_BIT}`. 
`vkGetDeviceQueue(0, 0)` and `vkGetDeviceQueue2({family = 0, index = 0, flags = PROTECTED})` 
are two different `VkQueue` handles (VUID-vkGetDeviceQueue-flags-01841, VUID-VkDeviceQueueInfo2-flags-06225), and the second registration silently overwrites the first.

-> `child_queues` should instead be `unordered_map<VkQueue, QueueWrapper*>`

framework/encode/vulkan_state_writer.cpp
```diff
@@ -515,15 +515,23 @@ void VulkanStateWriter::WriteSemaphoreState(const VulkanStateTable& state_table)
...
-            const vulkan_wrappers::QueueWrapper* queue_wrapper = entry.first->child_queues.front();
-            WriteCommandExecution(queue_wrapper->handle_id,
-                                  0,
-                                  nullptr,
-                                  static_cast<uint32_t>(entry.second.size()),
-                                  entry.second.data(),
-                                  0,
-                                  nullptr,
-                                  nullptr);
+            const auto device_queues_entry = entry.first->child_queues.begin();
```
fabian-lunarg, line 518: `child_queues` is an `unordered_map` now, so this is now selecting a random queue

framework/encode/vulkan_state_writer.cpp
```diff
@@ -3323,7 +3331,8 @@ void VulkanStateWriter::WriteImageMemoryState(const VulkanStateTable& state_tabl
                                                             *device_wrapper->physical_device->layer_table_ref,
                                                             device_wrapper->property_feature_info,
                                                             device_wrapper->version_extension_info,
-                                                            device_wrapper->physical_device->memory_properties);
+                                                            device_wrapper->physical_device->memory_properties,
+                                                            MakeQueueLockFn(device_wrapper));
```
fabian-lunarg: think locking not required here, this scope is already protected (under `ActivateTrimming`)

framework/encode/vulkan_state_writer.cpp
```diff
@@ -3467,7 +3476,8 @@ void VulkanStateWriter::WriteResourceMemoryState(const VulkanStateTable& state_t
                                                     *device_wrapper->physical_device->layer_table_ref,
                                                     device_wrapper->property_feature_info,
                                                     device_wrapper->version_extension_info,
-                                                    device_wrapper->physical_device->memory_properties);
+                                                    device_wrapper->physical_device->memory_properties,
+                                                    MakeQueueLockFn(device_wrapper));
```
fabian-lunarg: same, seems redundant here (under `ActivateTrimming`)

framework/encode/vulkan_capture_common.cpp
```diff
@@ -1001,5 +1256,23 @@ void CommonProcessHardwareBuffer(format::ThreadId                      thread_id
...
 }
 
+graphics::VulkanResourcesUtil::QueueLockFn MakeQueueLockFn(const vulkan_wrappers::DeviceWrapper* device_wrapper)
+{
+    GFXRECON_ASSERT(device_wrapper != nullptr);
+
+    return [device_wrapper](VkQueue queue) -> std::unique_lock<std::mutex> {
+        // FindQueueWrapper releases queues_map_mutex before queue_mutex is taken. This is the same order used by
+        // OverrideDeviceWaitIdle and no code path takes queues_map_mutex while holding a queue_mutex.
+        auto* queue_wrapper = vulkan_wrappers::FindQueueWrapper(device_wrapper, queue);
```
fabian-lunarg: the comment is misleading, or I'm confused here ("This is the same order used by OverrideDeviceWaitIdle")

but it's not (`OverrideDeviceWaitIdle` takes all `queue_mutex` while holding `queues_map_mutex`)

framework/encode/vulkan_handle_wrapper_util.h
```diff
@@ -302,6 +304,27 @@ inline void CreateWrappedHandle<DeviceWrapper, NoParentWrapper, DeviceMemoryWrap
     memory_wrapper->parent_device = GetWrapper<DeviceWrapper>(device);
 }
 
+// Returns the wrapper of the queue the application retrieved for this driver (unwrapped) handle, or nullptr if
+// the application never retrieved it.
+inline QueueWrapper* FindQueueWrapper(const DeviceWrapper* device_wrapper, VkQueue handle)
+{
+    assert(device_wrapper != nullptr);
```
fabian-lunarg: `GFXRECON_ASSERT`

framework/encode/vulkan_capture_common.cpp
fabian-lunarg: also thinking this could/should be made more robust (and simpler).
it's not required to call `FindQueueWrapper` at all (involving the nested loops, mutex-locking etc.)
-> we could call `GetWrapper<QueueWrapper>(queue, false)` directly (and so does `OverrideQueueSubmit`, so the two should agree)

framework/encode/vulkan_capture_manager.cpp
panos-lunarg: Good catch. Fixed

framework/encode/vulkan_state_writer.cpp
panos-lunarg: Selecting a random queue from the device is the intention here even before this patch. It triggers a submission just for semaphore signaling (no command buffers)

framework/encode/vulkan_state_writer.cpp
panos-lunarg: removed

framework/encode/vulkan_state_writer.cpp
panos-lunarg: removed

framework/encode/vulkan_capture_common.cpp
panos-lunarg: Good point. Simplified by just calling GetWrapper

framework/encode/vulkan_handle_wrapper_util.h
panos-lunarg: Fixed

framework/encode/vulkan_state_writer.cpp
fabian-lunarg: alright, makes sense

framework/encode/vulkan_capture_manager.cpp
```diff
@@ -4537,5 +4532,176 @@ void VulkanCaptureManager::PostProcess_vkTransitionImageLayout(VkResult
...
+    auto*                        queue_wrapper = vulkan_wrappers::GetWrapper<vulkan_wrappers::QueueWrapper>(queue);
+    std::unique_lock<std::mutex> queue_lock;
+    if (queue_wrapper != nullptr)
+    {
+        queue_lock = std::unique_lock<std::mutex>(queue_wrapper->queue_mutex);
+    }
+
+    auto                    handle_unwrap_memory = GetHandleUnwrapMemory();
+    const VkBindSparseInfo* pBindInfo_unwrapped =
+        vulkan_wrappers::UnwrapStructPtrHandles(pBindInfo, handle_unwrap_memory);
```
fabian-lunarg: the generated code we had before (correctly) used:

`const VkBindSparseInfo* pBindInfo_unwrapped = vulkan_wrappers::UnwrapStructArrayHandles(pBindInfo, bindInfoCount, handle_unwrap_memory);`

I think like this we would pass data for only one `VkBindSparseInfo` to the driver,
and `bindInfoCount` > 1 would cause reading past that.

panos-lunarg: fixed

fabian-lunarg (approved):

PR #3270, opened 2026-09-15
Test harness: run through ctest, remove false passes, add reruns and reports

MarkY-LunarG:
## What was wrong

  - The run scripts called gfxrecon-testapp-runner directly. ctest knew the cases but started them in the build tree, where the launcher, the tools and the known-good files do not exist, so ctest was unusable,
    and no case had a label or a timeout.
  - A case converted <name>.gfxr after the app ran without deleting the previous one. When the app wrote no capture, the case compared a stale file and passed. Seven of the fourteen cases passed with no
    capture layer loaded.
  - The JSON normalizer that drops volatile fields was a chain of string compares on quoted keys, with no comment on any key and no test.
  - Three test apps were built and launchable but had no test case. Two of them cannot run on the mock ICD today, and nothing recorded why.
  - One test app indexed its per-image fence list by frame instead of by image index. It aborts on any driver with fewer swapchain images than frames in flight. It also rendered 1000 identical frames, so its
    capture was 1.47 MB and its case took 10 s.
  - The Linux CI test step ran as root, and the CI job showed a failure only as a log to download.
  - scripts/test.py had a no-argument mode that searched a directory where no test binary is ever installed, and its list lacked three of the nine Catch2 targets.

  ## Result

 | | Before | After |
 | ---| :-----: | :----: |
 | Cases under ctest             | 0        | 63 listed: 62 run, 1 disabled with its reason in the file |
 | Labels and timeouts          | none    | smoke and unit, 120 s each                                |
 | Runs per capture app         | 1        | 3                                                         |
 | False pass on a stale capture | yes      | fixed, with a proof case                                  |
 | CI report                     | log only | job summary page plus the log                             |
 | Suite time on a debug build   | 2 s      | 6 s                                                       |

##  How to run

  python3 scripts/build.py --skip-check-code-style --test-apps
  ctest --test-dir build/linux/x64 --output-on-failure        # everything
  ctest --test-dir build/linux/x64 -L smoke                   # the app cases
  ctest --test-dir build/linux/x64 -L unit                    # the Catch2 tests

  test/README.md has the rest.

 ## Notes for review

  - The two known-good files that this series adds total 474 KB. known_good/ is 801 KB.
  - HostImageCopy.CorrectGFXR is disabled. The app compares bytes that it copied with VK_EXT_host_image_copy, and the mock ICD moves no bytes. A later series gives the mock real memory.
  - The Windows and macOS run scripts changed in the same way as the Linux one. This pull request is their first run.
  - The test-summary step is new to this repository. It reads one XML file and writes to the job summary. If the team prefers no third-party action there, the XML alone still serves the later coverage work,
    and the step can go.
 - The current design required us to use "sudo" to run the tests on Linux, that should not be needed.


bradgrantham-lunarg:
- [should-fix] Parallel ctest races on shared output files. Paths builds capture_path as <install>/test/<app>.gfxr, so    Triangle.CorrectGFXR, Serialized.CorrectGFXR/triangle, and CaptureDisabled.ProducesNoFile/triangle all write, read, and now    delete the same file (verify-gfxr.cpp:305    (https://github.com/LunarG/gfxreconstruct/blob/4f7120199a3f957a2af4c8a9c2097d7ee8185c4a/test/verify-gfxr.cpp#L305)). The old    in-process runner was serial; ctest runs one process per case, and the script headers advertise -j 4 (run-tests.sh.in:6    (https://github.com/LunarG/gfxreconstruct/blob/4f7120199a3f957a2af4c8a9c2097d7ee8185c4a/test/run-tests.sh.in#L6)). With -j,    verify_no_capture can see a sibling's file and fail, or a CorrectGFXR case can fail with "capture file was not roduced" after    a sibling deletes it. Fix options: include the gtest suite/test name in the output file names    (testing::UnitTest::GetInstance()->current_test_info()), or set RESOURCE_LOCK per app in test_environment.cmake.in, or remove    -j from the docs and mark the cases RUN_SERIAL.
  - [should-fix] function_definitions.h:2331 (https://github.com/LunarG/gfxreconstruct/blob/4f7120199a3f957a2af4c8a9c2097d7ee8185c4a/test/icd/generated/function_definitions.h#L2331): GetDeviceImageMemoryRequirements still returns alignment = 1 and an unrounded size. The PR comments say every requirement is in 64 KiB units; an app using vkGetDeviceImageMemoryRequirements(KHR)  would reintroduce the page-size dependence the series removes. Route it through RoundToMockGranularity and set the alignment.
  - [should-fix] verify-gfxr.cpp:457
    (https://github.com/LunarG/gfxreconstruct/blob/4f7120199a3f957a2af4c8a9c2097d7ee8185c4a/test/verify-gfxr.cpp#L457):    capture_and_replay now deletes stale outputs but did not get the exists(capture_path) assertion that verify_gfxr and    run_trimming_app got. A missing capture surfaces as a replay exit code instead of the clear message.
  - [should-fix] README.md:47   (https://github.com/LunarG/gfxreconstruct/blob/4f7120199a3f957a2af4c8a9c2097d7ee8185c4a/test/README.md#L47) says ctest -L unit    every CaptureDisabled case pass. The layer behavior itself is right (capture_manager.cpp:496 sets kModeDisabled on mismatch).    The CorrectGFXR case for the same app in the same ctest run does prove the environment, so this is a documentation caveat    unless you want to assert on a layer log line via GFXRECON_LOG_FILE.
  - [should-fix] README.md:47    (https://github.com/LunarG/gfxreconstruct/blob/4f7120199a3f957a2af4c8a9c2097d7ee8185c4a/test/README.md#L47) says ctest -L unit    runs the Catch2 tests, but the add_test at Test.cmake:78    (https://github.com/LunarG/gfxreconstruct/blob/4f7120199a3f957a2af4c8a9c2097d7ee8185c4a/cmake/Test.cmake#L78) is inside if    (${RUN_TESTS}). A build.py --skip-tests configure registers none. Say so, or register unconditionally.
  - [nit] verify-gfxr.cpp:49
    (https://github.com/LunarG/gfxreconstruct/blob/4f7120199a3f957a2af4c8a9c2097d7ee8185c4a/test/verify-gfxr.cpp#L49):
    key.rfind(prefix, 0) == 0 → key.starts_with(prefix) (project is C++20). Line 84: break after setting skip_next_buffer, or
    std::any_of.
  - [nit] remove_previous_outputs is a non-static free function with no header declaration; make it static.
  - [nit] reruns.cpp:53 (https://github.com/LunarG/gfxreconstruct/blob/4f7120199a3f957a2af4c8a9c2097d7ee8185c4a/test/test_cases/reruns.cpp#L53):    std::replace(name.begin(), name.end(), '-', '_').
  - [nit] json-normalizer.cpp:38 (https://github.com/LunarG/gfxreconstruct/blob/4f7120199a3f957a2af4c8a9c2097d7ee8185c4a/test/test_cases/json-normalizer.cpp#L38), 77, 107: size() compared with a signed literal; use 1u/0u.
  - [nit] The app-name lists in reruns.cpp duplicate GFXRECON_TESTCASES in CMake, and the #if defined(__linux__) guard does not match CMake's Linux|BSD|GNU. Consider generating the list from CMake or a shared header.

PR #3271, opened 2026-09-15
[Frame Looping] Shadow buffer suballocation

wzchu-lunarg:
Shadow buffers were previously created by allocating memory for each buffer in https://github.com/LunarG/gfxreconstruct/pull/3234.

We can instead allocate blocks of memory, and bind the buffers to offsets. 

We do this in two passes:
First create an estimate by creating buffers for each shadow, and obtain the memory requirements with `GetBufferMemoryRequirements`. We can then estimate a total size with `aligned_value(requirements.size, requirements.alignment)`

The second pass allocates blocks based on this total size, and adds blocks if needed while binding the memory.


framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -529,6 +530,45 @@ void VulkanReplayFrameLoopConsumer::FixupDeviceBuffers(format::HandleId device)
     it->second.Restore();
 }
 
+// Spec required minimum for maxMEmoryAllocationSize.
+constexpr VkDeviceSize kMaxMemoryBlockSize = 1024ull * 1024ull * 1024ull;
+
+constexpr size_t kInvalidBlockIndex = std::numeric_limits<size_t>::max();
```
antonio-lunarg: Should these two constants go into a header with similar configurations?

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -529,6 +530,45 @@ void VulkanReplayFrameLoopConsumer::FixupDeviceBuffers(format::HandleId device)
...
+
+VkDeviceSize VulkanReplayFrameLoopConsumer::BufferTracking::MaxBlockSize(uint32_t memory_type_index) const
+{
+    GFXRECON_ASSERT(memory_type_index < memory_properties_->memoryTypeCount);
+
+    // An allocation can never be larger than the heap that it comes from.
+    const uint32_t heap_index = memory_properties_->memoryTypes[memory_type_index].heapIndex;
+
+    return std::min(kMaxMemoryBlockSize, memory_properties_->memoryHeaps[heap_index].size);
```
antonio-lunarg: if `kMaxMemoryBlockSize` is the required minimum by spec, then this function si always returning `kMaxMemoryBlockSize`?

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -529,6 +530,45 @@ void VulkanReplayFrameLoopConsumer::FixupDeviceBuffers(format::HandleId device)
...
+}
+
+size_t VulkanReplayFrameLoopConsumer::BufferTracking::AddMemoryBlock(uint32_t     memory_type_index,
+                                                                     VkDeviceSize preferred_size,
+                                                                     VkDeviceSize minimum_size)
+{
+    VkDeviceSize block_size = std::min(std::max(preferred_size, minimum_size), MaxBlockSize(memory_type_index));
```
antonio-lunarg: This line is a bit too complex, it required my a few seconds to understand the logic.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -529,6 +530,45 @@ void VulkanReplayFrameLoopConsumer::FixupDeviceBuffers(format::HandleId device)
...
+    block_size              = std::max(block_size, minimum_size);
```
antonio-lunarg: What if `minimum_size` is greater than `MaxBlockSize(..)`?

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -529,6 +530,45 @@ void VulkanReplayFrameLoopConsumer::FixupDeviceBuffers(format::HandleId device)
...
+    VkMemoryAllocateInfo alloc_info = { VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO };
+    alloc_info.allocationSize       = block_size;
+    alloc_info.memoryTypeIndex      = memory_type_index;
+
+    if (allocator_->AllocateMemoryDirect(&alloc_info, nullptr, &block.memory, &block.mem_data) != VK_SUCCESS)
+    {
+        return kInvalidBlockIndex;
+    }
+
+    block.size = block_size;
```
antonio-lunarg: This can be encapsulated in `MemoryBlock::Allocate(...)`. Also the size checks above can go in there.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -578,66 +624,116 @@ void VulkanReplayFrameLoopConsumer::BufferTracking::RecordInitialState(const std
...
+        // The shadow buffers are all created with the same flags and usage, and should use the same memory type index.
         if (memory_type_index == std::numeric_limits<uint32_t>::max())
         {
-            memory_type_index = graphics::GetMemoryTypeIndex(*memory_properties_, mem_reqs.memoryTypeBits, 0);
+            memory_type_index = graphics::GetMemoryTypeIndex(
+                *memory_properties_, pending.requirements.memoryTypeBits, VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT);
+            if (memory_type_index == std::numeric_limits<uint32_t>::max())
+            {
+                memory_type_index =
+                    graphics::GetMemoryTypeIndex(*memory_properties_, pending.requirements.memoryTypeBits, 0);
```
antonio-lunarg, line 685: I think there should be some comment here explaining why we look for device local memory first and then fallback to any kind of memory.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -578,66 +624,116 @@ void VulkanReplayFrameLoopConsumer::BufferTracking::RecordInitialState(const std
...
+                                     buffer_id);
+                allocator_->DestroyBufferDirect(pending.shadow.buffer, nullptr, pending.shadow.alloc_data);
+                continue;
+            }
         }
-        if (memory_type_index == std::numeric_limits<uint32_t>::max())
+        GFXRECON_ASSERT((pending.requirements.memoryTypeBits & (1u << memory_type_index)) != 0);
+
+        // The padding that aligning each suballocation adds is part of what a block has to hold.
+        remaining_size += util::aligned_value(pending.requirements.size, pending.requirements.alignment);
```
antonio-lunarg: This looks odd to me. Reading `remaining_size` makes me think that it represents memory that is still available to be used by the buffers, but you're accumulating here, so is that `total_memory_size` or something instead?

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -578,66 +624,116 @@ void VulkanReplayFrameLoopConsumer::BufferTracking::RecordInitialState(const std
...
                                  util::ToString(result).c_str());
-            allocator_->FreeMemoryDirect(shadow.memory, nullptr, shadow.mem_data);
-            allocator_->DestroyBufferDirect(shadow.buffer, nullptr, shadow.alloc_data);
+            allocator_->DestroyBufferDirect(pending.shadow.buffer, nullptr, pending.shadow.alloc_data);
             continue;
         }
 
-        VkBufferCopy region = { 0, 0, buffer_info->size };
-        device_table_.CmdCopyBuffer(temp_cmd_buff.command_buffer, buffer_info->handle, shadow.buffer, 1, &region);
+        block.next_offset = offset + pending.requirements.size;
```
antonio-lunarg: This also looks like something that can be encapsulated in a MemoryBlock::function

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -578,66 +624,116 @@ void VulkanReplayFrameLoopConsumer::BufferTracking::RecordInitialState(const std
...
+        remaining_size -=
+            std::min(remaining_size, util::aligned_value(pending.requirements.size, pending.requirements.alignment));
 
-        shadow_buffers_[buffer_id] = shadow;
+        const VulkanBufferInfo* buffer_info = object_table_.GetVkBufferInfo(pending.buffer_id);
+        GFXRECON_ASSERT(buffer_info != nullptr);
+
+        VkBufferCopy region = { 0, 0, pending.shadow.size };
+        device_table_.CmdCopyBuffer(
+            temp_cmd_buff.command_buffer, buffer_info->handle, pending.shadow.buffer, 1, &region);
```
antonio-lunarg: I think this can be a `PendingShadowBuffer` consuming-transform function. Possible a `PendingShadowBuffer::CopyBuffer(...)` that would consume the `PendingShadowBuffer` `this` instance and would return a `ShadowBuffer` ready to be stored.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -687,13 +783,18 @@ void VulkanReplayFrameLoopConsumer::BufferTracking::DestroyShadowBuffers()
...
+
+        for (auto& block : memory_blocks_)
+        {
+            if (block.memory != VK_NULL_HANDLE)
             {
-                allocator_->FreeMemoryDirect(shadow.memory, nullptr, shadow.mem_data);
+                allocator_->FreeMemoryDirect(block.memory, nullptr, block.mem_data);
             }
         }
     }
```
antonio-lunarg: I think if `allocator_` is stored as an attribute in both `ShadowBuffer` and `MemoryBlock`, you can move all this logic in their destructors.

framework/decode/vulkan_replay_frame_loop_consumer.h
```diff
@@ -193,22 +193,43 @@ class VulkanReplayFrameLoopConsumer : public VulkanReplayFrameLoopConsumerBase
         struct ShadowBuffer
```
antonio-lunarg, line 199: I forgot to mention that it would be beneficial to add documentation to this class. Something to explain what is a ShadowBuffer, what it might be used for, why is it called "Shadow".

antonio-lunarg (changes requested):

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: `kMaxMemoryBlockSize` is the 1 Gib that we know no allocation should exceed no matter the device, although some devices do allow more. However, the memory type's heap might be greater or lower than this. It is mainly for cases where the heap doesn't even have a capacity of 1 Gib, such as resizable bar being in chunks of 256 Mib.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Yep, that seems much cleaner.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Breaking down into clearer names

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Missed this, adding a guard to drop buffers that are bigger than the max block size, as well as an assert.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Encapsulating into the Allocate function

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Adding a comment

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: The goal here is to obtain the size of all shadow buffers and their alignments. The reasoning for `remaining` is we are consuming it on the second pass when we allocate memory to bind the buffers to. 

I think `unbound_size` will make this clearer.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Encapsulating this too

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Adding this encapsulation

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: Adding a commit to access these types with unique pointers, as we need move operations to destruct properly and not get a double free.

framework/decode/vulkan_replay_frame_loop_consumer.h
wzchu-lunarg: Adding a comment for the struct

framework/decode/vulkan_replay_frame_loop_consumer.cpp
antonio-lunarg: I am not sure I agree with the updated comment. How can you tell that the buffers we shadow are likely device local? I don't think we have guarantees to say that. Buffers can also be host visible or have other memory properties.
I think the goal here is to ensure that the copy operation runs as fast as it can, which might be true if it works with local device memory.

framework/decode/vulkan_replay_frame_loop_consumer.h
antonio-lunarg: Nice. I would say that the creation strategy prefers device local memory when possible.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
antonio-lunarg: Brilliant, make sure to include your answer as a comment in that function :)

framework/decode/vulkan_replay_frame_loop_consumer.cpp
antonio-lunarg: In general I prefer proper implemented constructors and copy/move operations. I would use unique pointers only when I want to effectively express unique ownership of a resource, rather than a conveniency to avoid implementing copy/move.

That said, the practice of using unique pointers as a workaround is widely used, so I would say: your call. If you're fine with using unique pointers this way, I'm fine too.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -578,66 +671,108 @@ void VulkanReplayFrameLoopConsumer::BufferTracking::RecordInitialState(const std
...
                                  util::ToString(result).c_str());
-            allocator_->FreeMemoryDirect(shadow.memory, nullptr, shadow.mem_data);
-            allocator_->DestroyBufferDirect(shadow.buffer, nullptr, shadow.alloc_data);
             continue;
         }
 
-        VkBufferCopy region = { 0, 0, buffer_info->size };
-        device_table_.CmdCopyBuffer(temp_cmd_buff.command_buffer, buffer_info->handle, shadow.buffer, 1, &region);
+        unbound_size -=
+            std::min(unbound_size, util::aligned_value(pending.requirements.size, pending.requirements.alignment));
```
antonio-lunarg, line 759: I think we can extract block bookeeping in its own class. Ok if done in a subsequent PR.

Something like this:
```cpp
/// Suballocates shadow buffers from a few large allocations of one memory type.
class ShadowMemoryPool
{
  public:
    /// Sizes the next block.
    void Reserve(VkDeviceSize bytes);

    /// Checks the newest block and adds a block when needed.
    VkResult Bind(ShadowBuffer& shadow, const VkMemoryRequirements& requirements);

  private:
    std::vector<std::unique_ptr<MemoryBlock>> blocks_;
    VkDeviceSize                              reserved_{ 0 };
};
```

framework/decode/vulkan_replay_frame_loop_consumer.cpp
antonio-lunarg: I think the comment I was referring to has gone :D

wzchu-lunarg: Updated the comment.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
wzchu-lunarg: In that case I will add a commit that does the move constructors for the shadow buffers and memory blocks.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -578,66 +677,113 @@ void VulkanReplayFrameLoopConsumer::BufferTracking::RecordInitialState(const std
...
+                GFXRECON_LOG_WARNING("No suitable memory type for shadow buffer for buffer %" PRIu64
+                                     "; its contents will not be restored across loop repetitions.",
+                                     buffer_id);
+                continue;
+            }
         }
-        if (memory_type_index == std::numeric_limits<uint32_t>::max())
+        GFXRECON_ASSERT((pending.requirements.memoryTypeBits & (1u << memory_type_index)) != 0);
+
+        if (pending.requirements.size > MaxBlockSize(*memory_properties_, memory_type_index))
```
antonio-lunarg: What if we have a buffer bigger than `kMaxMemoryAllocationSize`? That would get skipped, even if there would be enough memory in the corresponding heap, right? Just thinking out loud in case we want to address this case in a follow-up PR.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
antonio-lunarg: This and a fake allocator should enable us to create unit tests to:
- make sure packing works as expected
- large buffers are handled correctly (after checking the largest buffers in our gfxr files)
- allocation failures are emitted as expected
- allocation and destroy happens correctly

antonio-lunarg (approved):

PR #3272, opened 2026-09-16
ci: quote Windows batch parameters to prevent shell injection

beau-lunarg:
cloneTests.bat, cloneSuites.bat, buildGfxr.bat, and runTest.bat expanded TEST_REPO, TEST_BRANCH, TEST_SUITE_REPO, TEST_SUITE_BRANCH, BUILD_MODE, and BITS unquoted. cmd.exe reparses an unquoted %VAR% expansion for shell metacharacters (&, |, etc.), the same vector confirmed exploitable against the legacy Windows Jenkins jobs. Quote each expansion so the gfxr and gfxr-extended pipeline jobs are hardened the same way.

Refs LunarG/Projects#1223


ncesario-lunarg (approved):
Yikes, glad nothing worse happened and you guys caught it!

PR #3273, opened 2026-09-16
[Frame Looping] Handle vkBindBufferMemory and vkBindImageMemory

nickdriscoll-lunarg:
This PR adds generated code for handling the original memory biding entry points.

A future PR will handle the modern entry points that can bind multiple objects to memory at once.


framework/generated/khronos_generators/khronos_replay_frame_loop_consumer_base_body_generator.py
```diff
@@ -267,6 +268,33 @@ def make_replay_frame_loop_consumer_func_body(self, api_data, return_type, name,
             body += '        // Since it might still be in use during the loop range, ONLY free it in the last iteration.\n'
             body += '        ' + self.genCallReplayConsumer(return_type, name, values)
             body += '    }\n'
+            body += '    // If this resource binds to memory, remove it from bound memory set\n'
+            body += '    boundMemory.erase(' + values[-2].prefixed_name + ');\n'
```
antonio-lunarg: This should happen only when the resource is actually destroyed, meaning it needs to go in each individual branches above:
- when not looping
- when the resource has been allocated within the loop
- when the resource has been allocated before and loop and loop iteration is the last one

framework/generated/generated_vulkan_replay_frame_loop_consumer_base.cpp
```diff
@@ -203,32 +207,64 @@ void VulkanReplayFrameLoopConsumerBase::Process_vkFreeMemory(
         // Since it might still be in use during the loop range, ONLY free it in the last iteration.
         VulkanReplayConsumer::Process_vkFreeMemory(call_info, args);
     }
+    // If this resource binds to memory, remove it from bound memory set
+    boundMemory.erase(args.memory);
```
antonio-lunarg: This is a special case: we're in `Process_vkFreeMemory`. There is no memory bound to itself, but there might be other resources that use this memory, so the code here should remove all entries whose value equals `args.memory`.

antonio-lunarg (changes requested):

framework/generated/khronos_generators/khronos_replay_frame_loop_consumer_base_body_generator.py
nickdriscoll-lunarg: Moved the code to the end of each branch.

framework/generated/generated_vulkan_replay_frame_loop_consumer_base.cpp
nickdriscoll-lunarg: Made vkFreeMemory a special case in VulkanReplayFrameLoopConsumer

framework/generated/generated_vulkan_replay_frame_loop_consumer_base.cpp
```diff
@@ -172,63 +190,60 @@ void VulkanReplayFrameLoopConsumerBase::Process_vkAllocateMemory(
     }
 }
 
-void VulkanReplayFrameLoopConsumerBase::Process_vkFreeMemory(
+void VulkanReplayFrameLoopConsumerBase::Process_vkBindBufferMemory(
```
antonio-lunarg, line 193: Do we need to implement also `Process_vkBindBufferMemory2` and `Process_vkBindImageMemory2`?

framework/decode/vulkan_replay_frame_loop_consumer.cpp
```diff
@@ -359,6 +359,76 @@ void VulkanReplayFrameLoopConsumer::Process_vkDestroyDevice(const ApiCallInfo& c
...
+            }
+            else
+            {
+                ++it;
+            }
+        }
+    }
+    else if (allocatedLoopResources.contains(args.memory))
+    {
+        // Looping special case:
```
antonio-lunarg, line 392: I think this might a special-special case.
Imagine a buffer created before the loop range, then memory created within the loop range and bound to that buffer. Then the memory is deallocated and the frame repeats. At that point a new memory gets created and bound to the same buffer, but that's a problem because the buffer is still bound to the previous deallocated memory. https://registry.khronos.org/VulkanSC/specs/1.0-extensions/man/html/vkBindBufferMemory.html#VUID-vkBindBufferMemory-buffer-07459

antonio-lunarg (changes requested):

framework/generated/generated_vulkan_replay_frame_loop_consumer_base.cpp
nickdriscoll-lunarg: Code generation for the modern endpoints is more complicated which is why I opted to split the work into two PRs.

framework/decode/vulkan_replay_frame_loop_consumer.cpp
nickdriscoll-lunarg: With the exception of sparsely bound buffers, it's invalid usage to free the memory to which a buffer is bound before the buffer has itself been destroyed.

From [the docs](https://docs.vulkan.org/spec/latest/chapters/resources.html#resources-association): `Once bound, the memory binding is immutable for the lifetime of the resource.`

Since frame looping doesn't handle sparse binding at the moment, and since sparse binding is very rarely used by apps to begin with (poor driver support, many engines simply emulate the functionality in software), I think we can ignore this case until we go back to support sparse binding.

antonio-lunarg (approved):

PR #3280, opened 2026-09-17
Add API call reader tool interface

per-mathisen-arm:
This adds a callback-based external tool interface for gfxreconstruct.

This allows external tools to read gfxreconstruct files, including both API commands and meta-commands. For now, only a limited number of meta-commands and only Vulkan API commands are supported. Two meta-commands that only exist in the Arm fork (device address and shader group handle fixups) have support that is non-functional in this patch, pending possible support for these in future patches.

The pull request looks big, but it is mostly just autogenerated code.

This interface is currently being used by lavatube to import and replay gfxreconstruct traces. You can test this by building lavatube with `cmake -DGFXRECON_VULKAN_API_READER_INCLUDE_DIR=<gfxr path>/framework -DGFXRECON_VULKAN_API_READER_LIBRARY=<path to>/libgfxrecon_vulkan_api_reader.so` and after building you can run `ctest -R gfxr` to check that the relevant tests pass.

This is still very much a work in progress, but feedback on the design and direction would be much welcome.

The second commit adds minimal support for traces produced Arm's fork by recognizing the header flag that we set. Here I think we probably should least reserve a few more bits for other vendors to have their own flags.


bradgrantham-lunarg:
Chaged to draft so we don't accidentally merge it before it's ready

PR #3282, opened 2026-09-17
WIP: Add support for VK_LUNARG_direct_driver_loading

MarkY-LunarG:
This adds support for the extension structs during both capture and replay, but also removes them during replay if the underlying HW doesn't support it.


framework/encode/vulkan_capture_manager.cpp
```diff
@@ -292,6 +293,105 @@ void VulkanCaptureManager::WriteSetDevicePropertiesCommand(format::HandleId
...
+        for (uint32_t i = 0; i < driver_list->driverCount; ++i)
+        {
+            DirectDriverRecord record;
+            record.driver_index = i;
+            record.driver_count = driver_list->driverCount;
+            record.mode         = static_cast<uint32_t>(driver_list->mode);
+
+            // A function pointer is not convertible to an object pointer directly. Go through an integer.
+            const auto address     = reinterpret_cast<uintptr_t>(driver_list->pDrivers[i].pfnGetInstanceProcAddr);
+            record.capture_address = static_cast<uint64_t>(address);
```
charles-lunarg:
```suggestion
            const auto address    = reinterpret_cast<const void*>(driver_list->pDrivers[i].pfnGetInstanceProcAddr);
            record.capture_address = reinterpret_cast<uint64_t>(address);
```
In C, it isn't allowed but in C++11 and onwards, it is conditionally allowed based on the architecture of the platform. On all the platforms we support, it is allowed (confirmed by compiling with gcc just now). By architecture, I mean Von Neumann vs Harvard, not the ISA.

test/test_apps/common/test_app_base.h
```diff
@@ -357,6 +357,15 @@ class InstanceBuilder
...
     InstanceBuilder& set_headless(bool headless = true);
 
+    // Add a structure to the pNext chain of VkInstanceCreateInfo.
+    // The structure must be valid when InstanceBuilder::build() is called.
+    template <typename T>
+    InstanceBuilder& add_pNext(T* structure)
+    {
+        info.pNext_elements.push_back(reinterpret_cast<VkBaseOutStructure*>(structure));
+        return *this;
+    }
```
charles-lunarg, line 367: Oh, I was going to comment about how we shouldn't be using VkBaseOutStructure*, and reinterpret casting structs to VkBaseOutStructure* is invalid C++, and some other not-so-nice statements. Then I saw that you are just following the existing code's jank, and remembered that the existing jank is code **I wrote** from vk-bootstrap. I have fixed it in upstream vk-bootstrap, but haven't propagated that here. Asking you to fix it would be pointless since it isn't your mistakes to fix. Just thought you'd like to know! (and so you can remind me to go port over the improvements sometime).

charles-lunarg (approved):
I checked this out locally, tried out a few things and found no issues with using it. Was able to replace the driver using the replay command line args, which was fun. Glad to see this relatively obscure extension work seamlessly with gfxr.

PR #3287, opened 2026-09-18
Update header to 1.4.363

wzchu-lunarg:
No new commands this time 😁


wzchu-lunarg:
It seems test case PipelineBinaries failed due to changes in the header and 4ae835c00017bbf96239c8b93b3f9e680b83af55. This was due to a registry change where `VkPipelineBinaryKeyKHR` now has a `len="keySize"` added.

This results in the encoder clamping the count for the struct, while the original capture didn't, which resulted in the failed test case. 

Updated the trace.

bradgrantham-lunarg:
> It seems test case PipelineBinaries failed due to changes in the header and [4ae835c](https://github.com/LunarG/gfxreconstruct/commit/4ae835c00017bbf96239c8b93b3f9e680b83af55). This was due to a registry change where `VkPipelineBinaryKeyKHR` now has a `len="keySize"` added.
> 
> This results in the encoder clamping the count for the struct, while the original capture didn't, which resulted in the failed test case.
> 
> Updated the trace.

Does this mean that any traces in the wild that used that struct are now invalid?

wzchu-lunarg:
> Does this mean that any traces in the wild that used that struct are now invalid?

They will still run fine. In this example, the original was encoding some extra unused data. But since our tests here check based on the converted output, it sees the difference and fails.

However, they will indeed be invalid in the sense that comparing convert output fails.

bradgrantham-lunarg (approved):

Issue #3013, opened 2026-06-10
VDR: VK_ERROR_DEVICE_LOST when dumping draws from secondary command buffers


fabian-lunarg:
- Resource dumping crashes with `VK_ERROR_DEVICE_LOST` on 'some' GPUs.
- Replay without dumping completes successfully
- bisecting pointed to 50207f9 (PR #2607)
- Crash probability scales with dump volume
- Splitting the dump range or re-triggering from the failing draw call succeeds
- Not reproducible on all GPU vendors/generations

current 'best guess' why above VDR-commit causes issues:
- each primary cmd-buffer re-executes all previously seen secondaries, causing per-submission GPU work to grow quadratically. growth is unbounded and eventually hits a device/driver-timeout/limit

bradgrantham-lunarg:
@panos-lunarg is this still an issue to the best of your knowledge?

panos-lunarg:
> [@panos-lunarg](https://github.com/panos-lunarg) is this still an issue to the best of your knowledge?

Yes, most likely it is

PR #3289, opened 2026-09-21
dump-resources: execute each target draw once by resuming render passes

### LINKED ISSUE #3258: [replay] Vulkan dump resources command-buffer clones reuse mutated attachments
**Describe the replay bug:**
If an attachment is mutated by a prior-dumped draw and is used by the current draw, then the current draw being dumped reads those mutations, and its attachments are corrupted. For example

```
// original API calls
CmdBuf
 RenderPass
  DrawOne // blend
  DrawTwo // blend
 RenderPass
CmdBuf
```
becomes
```
// dump attachments of DrawOne
CmdBuf
 RenderPass
  DrawOne      <- dump attachments
 RenderPass
CmdBuf

// dump attachments of DrawTwo
CmdBuf
 RenderPass
  DrawOne
  DrawTwo      <- dump attachments
 RenderPass
CmdBuf
```
The execution of "dump attachments of DrawTwo" reads the results of the prior command buffer replay and blends into those results, instead of the original attachment state.

fabian-lunarg:
this PR addresses the problem surfacing in #3013  :                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                    
- Dump-resources builds a clone per target draw, replaying its render pass from `vkCmdBeginRenderPass`                                                                                                                                                                                                                                         
   up to that draw,  then submits it standalone.                                                                                                                                                                                                                                                                                                   
-> for a pass with M **target draws** (i.e. draws to be **dumped**) that is O(M²) draws, independent of secondaries.                                                                                                                                                                                                                                                              
- `OverrideCmdExecuteCommands` nests it: the accumulator appends each secondary's full set (of already-cumulative clones)                                                                                                                                                                                                                         
->  peak submission is O(S * M^2). e.g. (S=25 secondaries, M=1000 draws): 
     ~12.5M draw re-executions in one submission, against ~25K when the targets are primary draws.                                                                                                                                                                    
-> those very large submissions are suspected to trigger the `VK_ERROR_DEVICE_LOST` detailed in #3013.                                                                                                                                                                                                                                           
-> it explains why the failure scales with dump volume and why chunking the dump can avoid device-loss.                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                    
**proposed solution:**                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                    
- "incremental LOAD-chaining" :                                                                                                                                                                                                                                                                                                                   
- partition the command stream into windows between consecutive target draws.                                                                                                                                                                                                                                                                     
-> a clone executes exactly the work in its window, plus the state needed to do so.                                                                                                                                                                                                                                                             
-> resumes the render pass it did not begin with `LOAD_OP_LOAD`.                                                                                                                                                                                                                                                                                  
-> each draw runs once, and each submission is one window.                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                    
**also (partially) fixes #3258  :**                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                    
- cumulative clones applied earlier dumped draws a second time whenever the attachment cannot be rebuilt from inside the command buffer.                                                                                                                                                                                                          
-> confirmed with a [constructed repro-case](https://github.com/fabian-lunarg/Vulkan/blob/gfxr_repro_issue_3258/examples/triangle_loadblend/triangle_loadblend.cpp)/capture: current dev reproduces the issue                                                                                                                                                                                                                                                                                                                          
-> fixed in this PR by construction. each draw is executed only once.


framework/decode/vulkan_replay_dump_resources_draw_calls.h
```diff
@@ -207,6 +213,30 @@ class DrawCallsDumpingContext
 
     uint32_t GetDrawCallActiveCommandBuffers(CommandBufferIterator& first, CommandBufferIterator& last) const;
 
+    // The clone a work command is recorded into: the current one.
+    uint32_t GetWorkCommandBuffers(CommandBufferIterator& first, CommandBufferIterator& last) const;
```
panos-lunarg: `GetWorkCommandBuffers` now returns just 1 command buffer, correct? Can we make some changes to better reflect that? Rename to `GetWorkCommandBuffer` or similar. The function can return the command buffer instead of  `current_cb_index_` since I don't think any caller uses it. Also the callers drop the `for (CommandBufferIterator it = first; it < last; ++it)`?

framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
```diff
@@ -3653,20 +3867,179 @@ uint32_t DrawCallsDumpingContext::GetDrawCallActiveCommandBuffers(CommandBufferI
...
+    // An instance this context did not begin itself lives entirely inside the current clone's window
+    if (!inside_renderpass_ || render_pass_contexts_.empty())
+    {
+        return GetWorkCommandBuffers(first, last);
+    }
+
+    const RenderPassContext& render_pass_context = *render_pass_contexts_.back();
+
+    const size_t begin = std::max(render_pass_context.first_clone, current_cb_index_);
+    const size_t end   = std::max(begin, render_pass_context.last_clone);
```
panos-lunarg, line 3900: `last_clone` is set to `first_clone + 1` in `BeginRenderPass` and `BeginRendering` so `GetRenderPassCommandBuffers`, similarly to `GetWorkCommandBuffers`, also returns 1 command buffer, correct? Can we apply similar changes as `GetWorkCommandBuffers` here as well?

framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
```diff
@@ -3644,6 +3816,48 @@ size_t DrawCallsDumpingContext::CmdBufToDCVectorIndex(size_t cmd_buf_index) cons
     }
 }
 
+void DrawCallsDumpingContext::GetCloneWindow(size_t cmd_buf_index, uint64_t& lo, uint64_t& hi) const
+{
+    // A slot's position in the primary's stream is where the vkCmdExecuteCommands that merged it sits,
+    // or the draw call block index for the primary's own draws.
+    const auto slot_position = [this](size_t slot_index) {
+        const DrawCallSlot& slot = dc_slots_[slot_index];
```
panos-lunarg, line 3833: Can you put a `GFXRECON_ASSERT(slot_index < dc_slots_.size());` here?

framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
panos-lunarg: my mistake. I see that it can actually return a larger range than just 1 command buffer

panos-lunarg (changes requested):

framework/decode/vulkan_replay_dump_resources_draw_calls.h
fabian-lunarg: > The function can return the command buffer instead of current_cb_index_ since I don't think any caller uses it.

yes, you're right. changed

framework/decode/vulkan_replay_dump_resources_draw_calls.cpp
fabian-lunarg: done

framework/decode/vulkan_replay_dump_resources_draw_calls.h
panos-lunarg: Actually I see the return value _is_ used by some callers but as a boolean which is wrong (preexisting bug). Perhaps `GetWorkCommandBuffers` can return a `VkCommandBuffer` instead of `current_cb_index_` and the callers check if it is `!= VK_NULL_HANDLE`

fabian-lunarg: it should be fine now. we define `VkCommandBuffer GetWorkCommandBuffer() const` 
and the checks basically don't exist in their previous form. 
also we can guarantee we'll return a valid cmd-buffer, never `VK_NULL_HANDLE` (except some `vkAllocateCommandBuffers` failed)

framework/decode/vulkan_replay_dump_resources.cpp
```diff
@@ -2767,7 +2810,7 @@ void VulkanReplayDumpResourcesBase::OverrideCmdCopyBufferToImage(
     for (auto transf_context : transf_contexts)
     {
         transf_context->HandleCmdCopyBufferToImage(call_info,
-                                                   commandBuffer,
+                                                   TransferRecordingCommandBuffer(commandBuffer),
```
panos-lunarg: What's the purpose of `TransferRecordingCommandBuffer`? Transfer contexts don't have a clone command buffer but inject their commands directly in the original command buffer

PR #3290, opened 2026-09-21
replay: Replace debug callbacks in the realign pre-pass

MarkY-LunarG:
gfxrecon-replay -m realign faulted in vkCreateInstance on any capture from an app that put a VkDebugUtilsMessengerCreateInfoEXT in the instance pNext chain. The resource tracking pre-pass passed the decoded VkInstanceCreateInfo to the loader as is, so the messenger still held the app's callback pointer from the capture process, and the loader called it for its first message.

Replace pfnUserCallback of every debug utils messenger and pfnCallback of every debug report callback in the chain with a function that returns VK_FALSE before the pre-pass creates its

Replace pfnUserCallback of every debug utils messenger and pfnCallback of every debug report callback in the chain with a function that returns VK_FALSE before the pre-pass creates its instance. The pre-pass reports nothing, so a silent callback is right. The replay consumer already does the same with its own logging callbacks.


arno-lunarg (dismissed):

arno-lunarg:
With those changes I still get debug messenger messages - LLM suggested this patch instead, and it does the job for me:
[gfxr_no_debug_callback.patch](https://github.com/user-attachments/files/32506482/gfxr_no_debug_callback.patch)

MarkY-LunarG:
@arno-lunarg , I added the patch you suggested.  Is this what you want?

arno-lunarg (approved):
Thank you Mark!!

framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -8431,8 +8455,18 @@ VkResult VulkanReplayConsumerBase::OverrideCreateDebugUtilsMessengerEXT(
...
-        modified_create_info.pfnUserCallback = DebugUtilsCallback;
-        modified_create_info.pUserData       = this;
+        modified_create_info.pfnUserCallback = NoopDebugUtilsCallback;
+        modified_create_info.pUserData       = nullptr;
+
+        auto* pnext_callback_info =
+            graphics::vulkan_struct_get_pnext<VkDebugUtilsMessengerCreateInfoEXT>(&modified_create_info);
+        while (pnext_callback_info != nullptr)
+        {
+            pnext_callback_info->pfnUserCallback = NoopDebugUtilsCallback;
```
jzulauf-lunarg: Why not just walk the list and update the pnext chain to walk around the nodes you want to Noop?

jzulauf-lunarg (commented):
I see what your doing, but wouldn't rewiring the pnext chain around the blocks you're Nooping be cleaner/less code?

PR #3291, opened 2026-09-21
A few fixes need to replay traces from vkd3d-proton

werman:
With these fixes I'm able to replay traces captured from D3D12 games running via vkd3d-proton, both with normal swapchain and with offscreen.


framework/decode/vulkan_replay_consumer_base.cpp
```diff
@@ -13669,6 +13753,43 @@ VkResult VulkanReplayConsumerBase::OverrideGetPastPresentationTimingGOOGLE(
...
+    if ((pPastPresentationTimingInfo == nullptr) || pPastPresentationTimingInfo->IsNull() ||
+        (pPastPresentationTimingInfo->GetPointer() == nullptr))
+    {
+        return original_result;
+    }
+
+    const VkPastPresentationTimingInfoEXT* in_past_presentation_timing_info = pPastPresentationTimingInfo->GetPointer();
+
+    GFXRECON_UNREFERENCED_PARAMETER(pPastPresentationTimingProperties);
+    return DrainPastPresentationTimingEXT(func, device_info->handle, *in_past_presentation_timing_info);
```
fabian-lunarg, line 13773: this makes sense but will probably trigger log-warn spamming for your traces.
(e.g. `VK_INCOMPLETE` in capture now gets `VK_SUCCESS` in replay)

best to add a { `VK_INCOMPLETE`, `VK_SUCCESS` } entry 
to the `kResultValuesAllowedByReplayEnvironment`-LUT at the top

fabian-lunarg (approved):
@werman  I have no concerns, only that I think you'll want
that entry in `kResultValuesAllowedByReplayEnvironment` I pointed out

framework/decode/vulkan_replay_consumer_base.cpp
werman: Good idea, added it.

